PR #12804, opened 2023-05-29
refactor(app): add style to truncate protocol name in splash screen

koji:
# Overview
add style to truncate protocol name in splash screen.

Screenshot 2023-05-29 at 4 35 30 PM

close RAUT-464

# Test Plan
- run a long name protocol

# Changelog
- add css props to SplashBody styled-component in RunSummary 

# Review requests
low

# Risk assessment

### DIFF
--- a/app/src/pages/OnDeviceDisplay/RunSummary.tsx
+++ b/app/src/pages/OnDeviceDisplay/RunSummary.tsx
@@ -272,6 +272,11 @@ const SplashHeader = styled.h1`
   color: ${COLORS.white};
 `
 const SplashBody = styled.h4`
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 4;
+  overflow: hidden;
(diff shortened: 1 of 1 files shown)


app/src/pages/OnDeviceDisplay/RunSummary.tsx
```diff
@@ -272,6 +272,11 @@ const SplashHeader = styled.h1`
   color: ${COLORS.white};
 `
 const SplashBody = styled.h4`
+  display: -webkit-box;
```
ewagoner, line 275: This works, but I'm curious why you didn't use then truncate() method instead?

ewagoner (approved):

app/src/pages/OnDeviceDisplay/RunSummary.tsx
koji: basically truncate() is for a case that needs to truncate text with a specific point.

PR #12807, opened 2023-05-30
refactor(app): 96 channel assets and instruments dashboard updates

jerader:
closes RLIQ-405 and RLIQ-406

# Overview

With a 96-channel attached on ODD, the correct info should be displayed on the instruments dashboard and when you detach it, it should show the correct flow.

Going through the flows, the assets should be updated except for the calibration flow (assets are not complete yet)

# Test Plan

on a robot, with a 96-channel attached, look at the instruments dashboard on the touchscreen, it should display only 1 item for left + right mount and clicking to detach it should launch the 96-channel instead of the single mount pipette flows. Also, going through the flows, the assets should be updated (except for in calibration flow)

# Changelog

- correct logic throughout pipette work in instruments dashboard to search for the 96 channel model name instead of name. Add logic to only display 1 labeled mount instead of 2 when 96 channel is attached
- add assets in the form of `webm`, create `getPipetteAnimations96` util that grabs the animations and update tests
- create `get96ChannelFromModel` util that returns boolean if you have a 96 channel attached from the different model names.
- write test for instrument info

# Review requests

see test plan

# Risk assessment

low


smb2268 (approved):
I tested this out on ODD and desktop and it works as expected!

PR #12809, opened 2023-05-30
chore: Regex fixup

DerekMaggio:
# Overview

There were some issues with our serial number regexes (regexs?) (regi?) (who knows). Fixed them.
Specifically,

- Our character matches were not strict enough is some areas, allowing matches on special characters
- Over-defined quantifiers which, although they worked, were confusing
- Unnecessary non-matching group

Also broke down the Python regex string itself and commented on what it does. 
Reading regex from scratch usually sucks

# Test Plan

- [x] Make sure tests still pass

# Changelog

- Change `{2,2}` and `{3,3}` quantifiers to `{2}` and `{3}` respectively. They mean the same thing but took some searching to make sure I wasn't missing anything. 
- Change usage of `.` to `[\w\d]`. The `.` character allows all characters except newlines to be matched. Going off the `SERIAL_FORMAT_MSG` inside the files, we want everything to be alphanumeric. The only one I am not %100 sure about is the `code` portion. I think we would want to only support alphanumeric but could use confirmation.
- Change usage of `[0-9]` to `\d`. Both function same but `\d` is specifically meant to replace `[0-9]`. 
- Use a multi-line string to break up regex into its logical parts and comment what each one is doing.
- Add test cases to cover fixed cases
- Remove non-capturing group. You only need to use non-capturing groups when you want to use quantifiers. So for instance, `(?:C){3}(something)` would match `cccsomething` but only capture something.

# Review requests

- [x] Can someone confirm that we want the `code` portion to be alphanumeric?

Confirmed by @sfoster1 

# Risk assessment

Low, tested with regex validation software, also tests will cover this

### DIFF
--- a/hardware/opentrons_hardware/instruments/gripper/serials.py
+++ b/hardware/opentrons_hardware/instruments/gripper/serials.py
@@ -5,7 +5,20 @@ import struct
 
 from opentrons_hardware.instruments.serial_utils import ensure_serial_length
 
-SERIAL_RE = re.compile("^(?:GRPV)(?P<model>[0-9]{2,2})(?P<code>.{,12})$")
+# Separate string into 2 groups
+#  - model
+#  - code
+
+GRIPPER_REGEX_STRING = (
+    "^"  # start of string
+    "GRPV"  # The characters GRPV
+    "(?P<model>\d{2})"  # "model" group contains exactly 2 digits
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Looks good to me, glad to have a regexpert fix these up

PR #12810, opened 2023-05-30
feat(api): randomize tip dropping location in trash

sanni-t:
Closes RQA-822

# Overview

Adds tip drop randomization to protocol engine `dropTip` command and wires up PAPI v2.15+ to use randomization when dropping tips without a location specified.

# Test Plan

- Run a v2.15 protocol with a few `instrument_context.drop_tip()` methods on both Flex & OT2. The pipette should drop tips at different, random locations in the trash well every time. This location should always be along the center line of the well in x-direction and should be within 3/4th area of the well from its center.
- Also run a v2.15 protocol with a drop_tip location specified. Verify that this does NOT randomize the drop location

# Changelog

- added `randomizeDropLocation` param to engine's `dropTip` command, which when true, drops tip in random trash location
- added a `get_random_drop_tip_location` getter to labware state
- added `randomize_drop_locatoin` arg to `core.drop_tip`
- updated `instrument_context.drop_tip` to set `core.random_drop_location` to True if API >= 2.15 & location is not specified.

# Review requests

- Is this a good approach? I've talked with people from SW and science about the best approach for both PAPI users as well as for engine and landed on this implementation. It avoids the problem of versioning in engine by making `randomizeDropLocation` an optional param.
- The PAPI method also still allows for overriding randomization by specifying a drop location so protocol authors will still have an option.

# Risk assessment

Medium. It changes the default behavior of `drop_tip()` when no location is specified. But after talking to science, this sounds like a welcome change.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/instrument.py
+++ b/api/src/opentrons/protocol_api/core/engine/instrument.py
@@ -325,6 +325,7 @@ class InstrumentCore(AbstractInstrument[WellCore]):
         location: Optional[Location],
         well_core: WellCore,
         home_after: Optional[bool],
+        randomize_drop_location: Optional[bool] = False,
     ) -> None:
         """Move to and drop a tip into a given well.
 
@@ -333,6 +334,8 @@ class InstrumentCore(AbstractInstrument[WellCore]):
                 Used to calculate the relative well offset for the drop command.
             well_core: The well we're dropping into
             home_after: Whether to home the pipette after the tip is dropped.
+            randomize_drop_location: Whether to randomize the exact location to drop tip
+                within the specified well.
         """
         well_name = well_core.get_name()
         labware_id = well_core.labware_id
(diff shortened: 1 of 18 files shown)


sfoster1 (approved):
Looks good to me!

jbleon95 (approved):
Looks good. I am in favor of, moving forward, going the route of making it cycle through predefined locations in a consistent pattern. Not only just for deterministic protocols, but also for deterministic tests (the one for random location, while running many iterations within the test to ensure boundary checking, could theoretically suddenly fail after many iterations the same way a coin could theoretically land on heads 1000 times in a row. Though for this PR the code is straightforward enough and the test is exhaustive enough that I'm not worried about this instance, though its a design pattern I wouldn't want to expand on)

Soapbox aside, this looks good and moves us forward with a path towards a more holistic solution to this issue.

PR #12811, opened 2023-05-30
feat(shared-data): adapter definitions and stacking overlaps

jbleon95:
# Overview

Coses RLAB-340 and RLAB-341

This PR covers two additions to shared data that are required for labware split.

The first is adding new definitions for adapters. It was decided to use the existing labware definitions for this task. To support and differentiate these from existing labware, a new `allowed_roles` property was added to the schema. This is implemented as an array of enums: `labware` that covers existing labware/tipracks, `adapter` which has been newly utilized in this PR, `fixture` for deck items that are affixed to the deck, and `maintenance` for deck items that would not be used in a normal protocol, such a calibration adapters. If the `allowed_roles` field is not present in a definition, it will be treated as `labware` being the only allowed role (no difference in behavior).

To support stacking labware on top of these adapters (as well as future stacking features), two new properties were also added to the schema, `stackingOffsetWithLabware` and `stackingOffsetWithModule`. These are both implemented as JSON objects, where the key is the load name of the labware or model name of the module, and the value is the overlap x, y and z offsets for that deck item. Only labware in this property can be stacked upon, and if the property is not present the labware cannot be stacked on any other labware.

`stackingOffsetWithLabware` will be used to determine z-height via the equation `labware z height + adapter z height - overlap`. `stackingOffsetWithModule` will have the similar equation `module labware offset z + labware z - overlap` to determine z height for pipetting operations. If this property or a module is not present in the definition, default behavior of module z height calculations will be used.

# Test Plan

```
metadata = {
    'protocolName': 'Labware Adapter Split load updated and new labware'
}

requirements = {
    "robotType": "OT-3",
    "apiLevel": "2.15"
}

ADAPTERS = [
    'opentrons_96_deep_well_adapter', 
    'opentrons_96_flat_bottom_adapter', 
    'opentrons_96_pcr_adapter', 
    'opentrons_universal_flat_adapter'
]

UPDATED_LABWARE = [
    'armadillo_96_wellplate_200ul_pcr_full_skirt',
    'corning_384_wellplate_112ul_flat',
    'nest_96_wellplate_100ul_pcr_full_skirt',
    'nest_96_wellplate_200ul_flat',
    'nest_96_wellplate_2ml_deep'
]


def run(context):
    location = 1
    for a in ADAPTERS:
        # When this epic is complete you will not be able to load adatpers via load_labware
        # BUT for purposes of this PR, this proves adatpers adhere to the updated schema
        adapter = context.load_labware(a, location)
        assert adapter._core._definition.allowedRoles == ['adapter']
        location += 1

    for l in UPDATED_LABWARE:
        labware = context.load_labware(l, location)
        assert labware._core._definition.version == 2
        location += 1
```

# Changelog

- Added three new fields to schema, `allowed_roles`, `stackingOffsetWithLabware` and `stackingOffsetWithModule`
- Added `opentrons_96_deep_well_adapter`, `opentrons_96_flat_bottom_adapter`, `opentrons_96_pcr_adapter`, `opentrons_universal_flat_adapter` to definitions
- Updated relevant labware definitions with stacking overlaps
- Updated python schema objects with new properties
- Fixed labware definition code to ignore unset properties to preserve existing hashing functionality

# Review requests

The big one is if we need to bump up the schema. There are three paths forward for how we can handle these new properties:
- Update everything in place, knowing that no existing information is being changed or removed, and remain on schema v2
- Remain on schema v2, but for any labware definition that has new properties added to it, bump up the _definition_ version (i.e make a `2.json` for those that only have a `1.json`
- Bump up the schema to v3

For bumping up the schema to v3, existing patterns would have all the definitions copied over to a new `labware/definitions/3` folders, which is a lot of copying of definitions that have not been changed at all. Discussion has been had about potentially only creating new v3 definitions for those that have changed, and defaulting back to v2 for those that haven't (though since it is a subset of v3, treating them for all intensive purposes as v3 definitions)

# Risk assessment
For going with updating the schema in place (but bumping versions of changed labware), medium-high. Anything that is using the old schema and using the new versions will not be validated, since the schema's `additionalProperties` field is set to False. Since the new properties are optional, old definitions will be fine with the new schema.


sfoster1 (changes requested):
This looks really good and I love it! I think we should try and future-proof by adding some more data, or at least slots for more data.
- I think we should get the offset values to be objects with axis names like `offset: z: 2.01` so if we need y and x we can add them
- i'm a little uncomfortable with `stackingOverlapWithLabware` and `stackingOverlapWithModule` as things that are defined by their name; I wonder if it might be better to have `stackingOverlaps` be a map of like, `role: offset` and add `module` as a role (😬) or maybe have those keys allow modules in addition to roles
- let's timebox the no-schema-version-bump-yes-def-version-bump path

shared-data/labware/schemas/2.json
```diff
@@ -40,13 +40,30 @@
...
+        "adapter",
         "other"
       ]
     },
     "safeString": {
       "description": "a string safe to use for loadName / namespace. Lowercase-only.",
       "type": "string",
       "pattern": "^[a-z0-9._]+$"
+    },
+    "coordinates": {
```
sfoster1, line 52: shocked this didn't exist honestly

sfoster1 (approved):
I think this is awesome especially if everything handles the definition version change well!

api/src/opentrons/protocol_api/core/engine/load_labware_params.py
```diff
@@ -15,6 +15,11 @@
     # easily cause collisions. (Jira RSS-197.)
     "opentrons_24_aluminumblock_generic_2ml_screwcap": 2,
     "opentrons_96_aluminumblock_generic_pcr_strip_200ul": 2,
+    "armadillo_96_wellplate_200ul_pcr_full_skirt": 2,
+    "corning_384_wellplate_112ul_flat": 2,
+    "nest_96_wellplate_100ul_pcr_full_skirt": 2,
+    "nest_96_wellplate_200ul_flat": 2,
+    "nest_96_wellplate_2ml_deep": 2,
```
sanni-t, line 24: We should specify that these are bumped to 2 because of the overlaps info and not because of wrong zDimension.

labware-library/src/labwareInference.ts
```diff
@@ -33,8 +33,9 @@ export function getUniqueWellProperties(
             }
     )
 
-    const xStart = wellProps[0].x
-    const yStart = wellProps[0].y
+    console.log(wellProps[0])
```
TamarZanzouri: remove this

TamarZanzouri (approved):
I like it! just one print to remove but overall I like this approach and I think our path forward is much more clear now

shared-data/python/opentrons_shared_data/labware/dev_types.py
```diff
@@ -116,3 +124,9 @@ class LabwareDefinition(TypedDict):
     dimensions: LabwareDimensions
     wells: Dict[str, WellDefinition]
     groups: List[WellGroup]
+
+
+class LabwareDefinition(_RequiredLabwareDefinition, total=False):
+    stackingOffsetWithLabware: Dict[str, NamedOffset]
+    stackingOffsetWithModule: Dict[str, NamedOffset]
+    allowedRoles: List[LabwareRoles]
```
sanni-t, line 132: Nice!

sanni-t (approved):
Looks really good! And now I'm fully on-board with @sfoster1 's idea of including all 3 axes' offsets instead of just the z overlap. It's going to be critical for calculating the x,y offsets for labware on adapters with no wells- like the tiprack adapter!

PR #12814, opened 2023-05-30
fix(api, app): thermocycler hold time on API v2.14

sanni-t:
Closes RESC-132.

# Overview

Fixes the bug in python API v2.14 that discarded thermocycler temperature hold time passed in `set_block_temperature`. 

# Test Plan

- Run a protocol with a large hold time set using `thermocycler_context.set_block_temperature` and verify that the protocol waits the full specified time before moving onto the next step.
- For definite proof that hold time is sent to the thermocycler, check the serial logs and verify that the set temperature gcode (M104) is sent with a hold time specified by the arg `H: ..`. Also verify that the hold time is allowed to get to `H=0` before next protocol command is executed.

# Changelog

- added an optional param `holdTimeSeconds` to `thermocycler/setTargetBlockTemperature` and wired up engine core's Thermocycler module core to use that param.

# Review requests

- I ran into a small dilemma related to hold time validation- the firmware accepts floating point value as an arg to any gcode, but this value gets type casted to an unsigned int when it gets passed to the thermocycler's timer. Should we be converting a float value to int in the engine itself? Finally decided to not do any conversion because the older API doesn't do it and we don't gain anything really. To make it truly useful, we'll have to update the command response to return the truncated float so that users can be informed of the conversion. But that might(?) need app-side changes which would make this fix a bit more involved. Another dilemma was whether we should validate that a given number fits into a 32-bit integer size but it kinda seemed silly and probably not worth it.
-  There were a couple of different ways to implement having a hold time- like setting it via a new, separate engine command or using a `waitForDuration` command. But the former would require a change to hardware controller in order to only set the hold time, while the latter would create a difference between the actual hold time duration in previous PAPI versions vs the new version since the exact point at which we start counting down in firmware would be different from the point we starting counting down the `waitForDuration`'s timer. In the firmware, we start the hold_time timer as soon as the temperature overshoot is over, and this is hard to reliably determine in API.

# Risk assessment

Low. Fixes a thing that was broken. Doesn't change anything else in 6.3.0.


sfoster1 (approved):
Looks great to me (haven't hardware tested though).

api/src/opentrons/protocol_engine/commands/thermocycler/set_target_block_temperature.py
```diff
@@ -71,13 +77,19 @@ async def execute(
             )
         else:
             target_volume = None
+        hold_time: Optional[float]
+        if params.holdTimeSeconds is not None:
```
jbleon95, line 81: Do we want this to be a check if holdTimeSeconds is zero instead of None? I know that down the stack, when the G-Code is calculated, a hold time of None will have no `H` prefixed command sent to the module, whereas including 0 will send an `H0`. I don't know if this functionally different in any way, but this would be a g-code change for PAPI calls that don't use hold time.

sanni-t: Interesting point! But I think we are safe here because the public API method does a hold time validation where it converts `None`s to `0`s so everything downstream gets a `seconds=0`.
Technically there still will be a difference of g-code between engine command and PAPI when `hold_time` is not specified but we don't claim to have a 1:1 relationship between PAPI & engine commands anyway so this is totally acceptable.

ecormany (commented):
Looks fine from the docs perspective. There appear to be no changes to public docstrings, and the internal docstrings and descriptions are clear.

api/src/opentrons/protocol_engine/commands/thermocycler/set_target_block_temperature.py
```diff
@@ -25,6 +25,12 @@ class SetTargetBlockTemperatureParams(BaseModel):
         description="Amount of liquid in uL of the most-full well"
         " in labware loaded onto the thermocycler.",
     )
+    holdTimeSeconds: Optional[float] = Field(
+        None,
+        description="Amount of time, in seconds, to hold the temperature for."
```
fsinapi, line 30: Probably want a space at the end of this line or beginning of the next. Same for the line below.

fsinapi (approved):
This LGTM, one string nitpick aside!

I do want to make sure that testing for this includes testing hold times with:
* A setting for `hold_time_minutes` but no `hold_time_seconds`
* A setting for `hold_time_seconds` but no `hold_time_minutes`
* A setting for both `hold_time_seconds` and `hold_time_minutes`

sanni-t:
Tested this on an OT2 with the above test params by @fsinapi as well and verified that hold time is now set correctly in 2.14. 
Will merge once CI clears.

sanni-t:
So because of the way the `set_block_temperature` method gets divided into a command that sets the temp & hold time and another that waits for the temp and hold time, our app's run log is not able to say anything about the hold time. 
Here's a screenshot of the run log steps when the protocol has a hold time specified in each of the set temperature commands-

Screenshot 2023-05-30 at 4 57 56 PM

~Do temperature profile execution steps show up the same way?~ Update. Nope. They are shown like this-
Screenshot 2023-05-30 at 5 23 59 PM

 I'm fine with keeping this difference as is for now. Maybe we can add this caveat to release notes.

ecormany:
> Do temperature profile execution steps show up the same way? I'm fine with keeping this for now. Maybe we can add this caveat to release notes.

Oof, profiles are quite broken too:

image

(It keeps going. If I specified 100 reps it would output 300 lines there.)


Profiles used to show this in 2.13:
`
Thermocycler starting 20 repetitions of cycle composed of the following steps: [{'temperature': 95, 'hold_time_seconds': 30}, {'temperature': 57, 'hold_time_seconds': 30}, {'temperature': 72, 'hold_time_seconds': 60}]
`

sanni-t:
So I talked with @b-cooper yesterday and he has provided a fix for the first issue that it doesn't show hold time in the run log. I'm going to test it and push to this branch.

As for the issue about profile steps showing each atomic steps, we will have to defer to next release since it's actually a profile command architecture problem in the backend which is not a simple fix.

PR #12571, opened 2023-04-27
feat(shared-data,api,robot-server): Use new OT-3 deck slot naming style

SyntaxColoring:
# Overview

This is a chunk of back-end support for OT-3 style deck slot names, like `D1` instead of `1`.

Closes RLAB-250, RLAB-251, RLAB-253, RLAB-256, RLAB-257, RLAB-339, and RLAB-346.

# Changelog

In general, the robot will now accept deck slots of *either style* (`1` or `D1`) as input. It will normalize them to the  device's preferred style (`1` for OT-2s, `D1` for OT-3s) in its internal state, and in its output. In protocol analyses, the style in the output will follow the protocol's declared `robotType`.

This affects:

* `loadLabware` commands
* `moveLabware` commands
* `loadModule` commands
* Labware offsets `POST`ed over HTTP
* Run summaries visible over HTTP

The underlying OT-3 deck definition in `shared-data` is changed to use the OT-3-style names. The module definitions are updated to match.

# Review requests

* [ ] Scrutinize my updates in `shared-data`.
  * [ ] Make sure I'm updating things to the *correct* new values. [Here's a handy quick reference table](https://gist.github.com/SyntaxColoring/9139aa7420e3993d20e9028501f65755), but double-check this with a physical OT-3.
  * [ ] Make sure I didn't accidentally leave any of the OT-2-style numbers behind. **It's very important that the module definitions and deck definitions are consistent with each other.**
* [ ] Can you think of any other places that might need to normalize deck slots before comparing them? For example, any conflict-checking code hiding out somewhere?

# Risk assessment

High.

This unfortunately has to change a bunch of stuff all at once, because it's all cross-dependent on each other.

Code that does this is in danger of breaking silently:

```
if some_slot == some_other_slot:
    ...
```

We do this in places such as:

* Determining the offset of a module. Modules have per-slot transformations, which are matched by slot ID.
* Determining the Labware Position Check offset to apply to a labware. Labware Position Check offsets are defined by a labware URI and a location, and the location includes the slot ID.

# Test plan

[Here's the protocol file that I'm using.](https://gist.github.com/SyntaxColoring/f4419bdf85eb73acb951a5ff45e811e8)

After pushing this branch to an OT-3, you'll need to factory-reset runs history before the robot is usable. You'll also need to factory-reset runs history when pushing `edge` back.

* [x] OT-3-only: Make sure you can still detach and attach pipettes.
* [x] OT-3-only: Make sure you can still calibrate pipettes.
[...]

PR #12815, opened 2023-05-30
fix(app): render correct deck in LPC

shlokamin:
# Overview

This PR changes LPC to use the deck definition for the robot type provided by the robot server's protocol endpoint. This fixes a bug where we were accessing deck slot names that don't exist during LPC.

closes RAUT-456

# Test Plan

Verified that LPC works for Flex protocols on a Flex.

# Changelog

- Render correct deck in LPC

# Risk assessment

Low

### DIFF
--- a/app/src/organisms/Devices/hooks/useProtocolMetadata.ts
+++ b/app/src/organisms/Devices/hooks/useProtocolMetadata.ts
@@ -16,6 +16,8 @@ export function useProtocolMetadata(): ProtocolMetadata {
   const author = protocolMetadata?.author
   const description = protocolMetadata?.description
   const lastUpdated = protocolMetadata?.lastModified
+  // TODO: remove robot type from this hook to align on server definition of metadata
+  // see https://github.com/Opentrons/opentrons/pull/12815#discussion_r1210780976
   const robotType = protocolRecord?.data.robotType
 
   return {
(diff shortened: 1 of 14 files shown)


app/src/organisms/Devices/hooks/useProtocolMetadata.ts
```diff
@@ -1,10 +1,11 @@
 import { useCurrentProtocol } from '../../ProtocolUpload/hooks'
-
+import type { RobotType } from '@opentrons/shared-data'
 interface ProtocolMetadata {
   author?: string
   lastUpdated?: number | null
   description?: string | null
   creationMethod?: 'json' | 'python'
+  robotType?: RobotType
```
SyntaxColoring, line 8: I don't think we should consider `robotType` part of "metadata."

On the server and in the analysis engine, we consider a protocol's metadata to be stuff that's not meaningful to the behavior of the protocol or its requirements. So, the author and title are metadata, but `apiLevel` and `robotType` are not.

Happy for this to be merged as-is if it's meaningfully easier, but can we call it out with a `// todo` or something to avoid terminological confusion down the line?

SyntaxColoring (approved):
Tested in-person with @shlokamin, atop #12571. Thank you!

app/src/organisms/Devices/hooks/useProtocolMetadata.ts
shlokamin: in classic fashion i ran out of time and i dont wanna block you so im gonna merge this for now with a todo and come back to this next week to refactor. thx max!

SyntaxColoring (approved):
Yay!

PR #12817, opened 2023-05-30
chore: do not build the app xplatform all the time

sfoster1:
It's good to run app builds to test if peoples' PRs work, but it's bad if those app builds sporadically fail or take an hour even if they succeed. Let's cut the number of times we do a full cross-platform app build down by only doing the build if
- This is a tag
- This is edge
- This is release
- This is internal-release
- This push was to a branch that ends in `app-build`

In other cases, we'll only do a linux build. On Linux, we don't have any signing or attestation, so the builds are reliable and fast. That lets us get the benefit (be sure we didn't break something obvious) and still have the builds not take forever.

## Note
This may not be a good idea! It means we won't be getting many app builds, and something breaking could surprise us. It also means we might have a rough first couple alphas since we can't really test the "we get builds on tags" part without pushing a matching tag. But then again, that's what the alphas are for.

### DIFF
--- a/.github/workflows/app-test-build-deploy.yaml
+++ b/.github/workflows/app-test-build-deploy.yaml
@@ -98,6 +98,10 @@ jobs:
     strategy:
       matrix:
         os: ['windows-2022', 'ubuntu-22.04', 'macos-11']
+        exclude:
+          - os: 'macos-11'${{!startsWith(github.ref, 'refs/tags') || !startsWith(github.ref, 'refs/heads/edge') || !startsWith(github.ref, 'refs/heads/internal-release') || !startsWith(github.ref, 'refs/heads/release') || null}}
+          - os: 'windows-2022'${{!startsWith(github.ref, 'refs/tags') || !startsWith(github.ref, 'refs/heads/edge') || !startsWith(github.ref, 'refs/heads/internal-release') || !startsWith(github.ref, 'refs/heads/release') || null}}
+
     name: 'opentrons app backend unit tests and build'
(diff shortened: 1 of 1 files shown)


.github/workflows/app-test-build-deploy.yaml
```diff
@@ -193,6 +194,10 @@ jobs:
     strategy:
       matrix:
         os: ['windows-2022', 'ubuntu-22.04', 'macos-latest']
+        exclude:
+          - os: format('macos-latest{0}', startsWith(github.ref, 'refs/tags') || startsWith(github.ref, 'refs/heads/edge') || startsWith(github.ref, 'refs/heads/internal-release') || startsWith(github.ref, 'refs/heads/release') || endsWith(github.ref, 'app-build') || null)
+          - os: format('windows-2022{0}', startsWith(github.ref, 'refs/tags') || startsWith(github.ref, 'refs/heads/edge') || startsWith(github.ref, 'refs/heads/internal-release') || startsWith(github.ref, 'refs/heads/release') || endsWith(github.ref, 'app-build') || null)
+
```
shlokamin, line 200: i'd add a comment that we're leaving ubuntu cuz its cheap and still good to get some sort of feedback

shlokamin (approved):
🐎

.github/workflows/app-test-build-deploy.yaml
sfoster1: Oh I'm uh... not anymore. Now it will build it on nothing, but it runs the backend tests.

PR #12818, opened 2023-05-30
fix(app): odd network setup live qa fixes

b-cooper:
# Overview

 Fix various visual UI bugs from a live QA session on the ODD network setup screens
 
 - make the border radius constants more specific. `size1` -> ` borderRadiusSize1`
 - add useWifiList poll of 5 seconds to the connect via wifi screen
 - remove the isSearching state and icon according to updated designs
 - add a new `spacing60` constant and use it for padding on a couple of setup screens


jerader (approved):
cool

PR #12781, opened 2023-05-25
feat(api): Support new Flex name as a Python robotType

SyntaxColoring:
# Overview

Python protocols declare the type of robot that they're meant to run on.

```python
requirements = {"apiLevel": "2.15", "robotType": "OT-3"}
```

When we added this, we only added support for the strings "OT-2" and "OT-3", because the OT-3's official marketing name, "Flex," didn't exist yet. This PR adds support for the string "Flex" so protocol authors don't have to be unduly exposed to an internal technical name.

```python
requirements = {"apiLevel": "2.15", "robotType": "Flex"}
```

Closes RCORE-756.

# Test plan

None needed. Existing automated `api` and `robot-server` tests should be sufficient for this.

# Changelog

* Support `"robotType": "Flex"`.
* Continue to support `"robotType": "OT-3"` to avoid disrupting our internal users.
* Update all the protocols that I could find to use `"Flex"`. We don't need to do this, but if we're saying `"OT-3"` is deprecated, it's nice to be consistent.
* Wipe the internal release notes, because v0.9.0 has already been branched off of `edge`.

# Review requests

Are we still okay with this user-facing spelling and capitalization? `"robotType": "Flex"` and `"robotType": "OT-2"`, not `"flex"` or `"ot-2"` or `"ot2"`.

Is my wipe of the release notes correct?

# Risk assessment

Low.

PR #12822, opened 2023-05-31
fix(app): Make labware filtering in Labware Position Check robust to different deck types

SyntaxColoring:
# Overview

The app has some code to filter out the trash bin from the labware to visit during Labware Position Check. This code was implemented in a way that depended on details of the deck definition.

This was a latent bug that manifests now that the OT-2 and OT-3 deck definitions are going to have different slot IDs. To fix it, we now bypass deck definitions entirely for this logic. Instead, we look at the labware definition's `params.format` and filter it out if it's equal to `"trash"`.

This fixes RLAB-350.

# Test plan

I've tested this as part of #12781. I clicked through Labware Position Check and made sure it visited all labware, and that it didn't visit the trash.

# Review requests

Is `format === 'trash'` a good way to implement this?

# Risk assessment

I'm gonna go with a pessimistic "medium" because I'm unfamiliar with the app code and there might be deeper implications that I'm missing.

### DIFF
--- a/app/src/organisms/ApplyHistoricOffsets/hooks/getLabwareLocationCombos.ts
+++ b/app/src/organisms/ApplyHistoricOffsets/hooks/getLabwareLocationCombos.ts
@@ -20,7 +20,7 @@ export function getLabwareLocationCombos(
   return commands.reduce<LabwareLocationCombo[]>((acc, command) => {
     if (command.commandType === 'loadLabware') {
       if (command.result?.definition == null) return acc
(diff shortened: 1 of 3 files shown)


b-cooper (approved):
Full Stack 🥞 !

SyntaxColoring:
Tested after changes by clicking through Labware Position Check on a `robot-server` OT-2 dev server (`make -C robot-server dev`).

PR #12823, opened 2023-05-31
feat(app): design qa from ADC

shlokamin:
# Overview

This PR makes adjustments that came out of design QA on 5/31.

- Do not redirect to runs page inside protocol details
- Clear runs query cache after successfully creating run
- Move LPC above labware in protocol setup screen
- Correct in LPC confirm modal
- Fix header on generic wizard title

# Risk assessment

Low


b-cooper (approved):
✅

PR #12824, opened 2023-05-31
feat(hardware): add limit switch backoff to home sequence

ahiuchingau:
# Overview

This PR implements backing off away from the limit switch right after a home request to ensure the origins of the stepper motor axes are consistent every time we home. 

See firmware-side changes https://github.com/Opentrons/ot3-firmware/pull/676 

# Test Plan
Test on all stepper motor axes and verify the stepper and encoder positions

# Changelog
* Adding a backoff step to the stepper motor home move group
* Halving the homing velocity to backoff to increase precision
<!--
List out the changes to the code in this PR. Please try your best to categorize your changes and describe what has changed and why.

# Changes from last review
* we only use the axis boundary distance as the home request distance if the stepper position status is not okay
* add 1.0 mm as the over-travel distance so we don't pass in a home request with 0 duration if we home an axis that's already in its home position

### DIFF
--- a/hardware/opentrons_hardware/firmware_bindings/constants.py
+++ b/hardware/opentrons_hardware/firmware_bindings/constants.py
@@ -341,7 +341,7 @@ class MoveStopCondition(int, Enum):
     gripper_force = 0x8
     stall = 0x10
     ignore_stalls = 0x20
-    limit_switch_backoff = 0x4C0
+    limit_switch_backoff = 0x40
 
 
 @unique
--- a/hardware/opentrons_hardware/firmware_bindings/constants.py
+++ b/hardware/opentrons_hardware/firmware_bindings/constants.py
@@ -341,7 +341,7 @@ class MoveStopCondition(int, Enum):
     gripper_force = 0x8
     stall = 0x10
     ignore_stalls = 0x20
-    limit_switch_backoff = 0x30
+    limit_switch_backoff = 0x40
 
 
 @unique
--- a/hardware/opentrons_hardware/firmware_bindings/constants.py
+++ b/hardware/opentrons_hardware/firmware_bindings/constants.py
@@ -341,6 +341,7 @@ class MoveStopCondition(int, Enum):
     gripper_force = 0x8
     stall = 0x10
     ignore_stalls = 0x20
+    limit_switch_backoff = 0x30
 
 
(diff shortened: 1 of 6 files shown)


hardware/opentrons_hardware/firmware_bindings/constants.py
```diff
@@ -341,6 +341,7 @@ class MoveStopCondition(int, Enum):
     gripper_force = 0x8
     stall = 0x10
     ignore_stalls = 0x20
+    limit_switch_backoff = 0x4C0
```
sfoster1:
```suggestion
    limit_switch_backoff = 0x40
```
?

sfoster1 (changes requested):
errant `C` I think

hardware/opentrons_hardware/firmware_bindings/constants.py
ahiuchingau: ugh

sfoster1 (approved):
Looks good to me!

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -401,15 +403,22 @@ def _handle_move_completed(
...
         group_id = message.payload.group_id.value - self._start_at_index
+        seq_id = message.payload.seq_id.value
         ack_id = message.payload.ack_id.value
         node_id = arbitration_id.parts.originating_node_id
         try:
-            stop_cond = self._stop_condition[group_id]
+            stop_cond = self._stop_condition[group_id][seq_id]
             if (
-                stop_cond == MoveStopCondition.limit_switch
+                stop_cond
```
sfoster1, line 412: Let's factor this into an `acceptable_stop_condition()` predicate function that checks this and can be extended to other stuff IMO

sfoster1 (approved):
Looks good but would love to see the stop condition checking be its own function

hardware/opentrons_hardware/hardware_control/move_group_runner.py
ahiuchingau: Yes! That's the next step i was gonna take, to do a refactor on the move runner completion message and make sure we're actually responding timeout correctly, but will be a next PR

PR #12828, opened 2023-06-01
feat(shared-data): update flex deck svg layers

shlokamin:
# Overview

This PR updates the Flex deck def to match the design team's designs. This will update how the flex deck gets rendered in all of our frontend products. This does NOT change any geometry of the flex.

# Changelog

- Update flex deck svg layers

# Review requests

Open up the app/storybook at look at the deck map to make sure labware + modules fit properly on it.

# Risk assessment

Low


b-cooper (approved):
🗺️

PR #12831, opened 2023-06-01
fix(shared-data): p1000_multi_v3 return tip height

sfoster1:
Used the wrong height for the multi variants; they need to be lower for the exact same reason.


fsinapi (approved):
yea

PR #12671, opened 2023-05-10
feat(api): add extension mount and axes to top level types

sanni-t:
Closes RLAB-219 & RLAB-231

# Overview

Inside `opentrons.types`,
- Adds Extension mount to `Mount`, `MountType` 
- Adds extension mount's axes to `Axis`.

# Test Plan

- Test homing an OT3's gripper mount with home command over http
- Test that doing the above raises an error on OT2
- Test that `moveToMaintenancePosition` with gripper mount, moves the gantry to the given x, y location, ignoring the z offset. (this will potentially fail until point2 in review requests is addressed).

# Changelog

The only user-facing changes are:
- Addition of `EXTENSION` type to Mount and MountType
- Addition of `Z_G` and `G` axes to `Axis` type
- `protocol_context.load_instruments` now raises error when `Mount.EXTENSION` is used for `mount` arg.

Client-facing change for HTTP layer is:
- You can now specify `extension_z` and `extension_jaw` for axis in `home` command

Non-user/client facing changes mostly fall under these categories:
- updating conversion from one type to another, now includes conversion of extension mount & axes types
- hardware control layer changes to expect `Mount` and `Axis` to include extension mount and axes
- for places that iterate over `Mount` type, restrict the iteration to OT2 mount types when using OT2, or restrict the iteration to pipette mounts only when the code is relevant to pipette mounts only.

# Review requests

1. The main reason to add Extension mount's _axes_ was to implement homing the gripper axes using HTTP `home` command. This simple requirement trickled down into needing to update the top level `Axis` type, which required updating the `Mount` & `MountType` types for keeping their inter-dependency valid. Do you think the chain of changes is justified and in the end is a change we want to make- i.e., we want `types.Axis` to reflect all the axes available- or would it be instead better to not have to update the top level types and keep the change restricted to just protocol engine? (basically by updating only `engine.types.MotorAxis` and handling homing the gripper axes with a special call to ot3api.home with axes specified in `OT3Axis` type)
2. The main reason to add Extension _mount type_ was to allow `moveToMaintenancePosition` to move the gripper mount to maintenance position during attach/detach. But this cannot be fully supported without a change in hardware controller since there is no 'gripper mount' available to move when there is no gripper attached. So in essence we want to just move the gantry to a specified x, y location in order to perform gripper attachment. So one can argue that we don't necessarily need the extension mount type specified in order to do such a move. 
[...]

PR #12834, opened 2023-06-01
refactor(api): merge OT3Axis into Axis 

sanni-t:
Closes RLAB-356

# Overview

This is a follow-up PR to #12671 and merges the two different axis types- Axis and OT3Axis into one. 
Please don't be daunted by the large diff. Most of it is just changing the names from 'OT3Axis' to 'Axis'. I'm pointing out the critical stuff in changelog below.

# Test Plan

Test a couple of protocol flows- including running attach/detach flows and LPC on both OT2 & OT3 to make sure nothing breaks.

# Changelog

- Made `OT3Axis` enum the new `Axis` with aliases for the old OT2 Axis members. So the true members of `Axis` are now: `X, Y, Z_L, Z_R, Z_G, P_L, P_R, Q, G` with aliases for OT2 axes of `Z, A, B, C`.
- Any and all changes relating to making the above change and making sure OT2 code still works and the smoothie driver gets the correct 'translated' old Axis.
- Most of the changes are just name updates following the merging of the `OT3Axis`.
- The critical changes are:
    - replacing the `.to_mount` method on `Axis` (and OT3Axis) with separate `.to_ot2_mount` and `.to_ot3_mount` for each robot type
    - making sure the OT2's axes still translate to `Z, A, B, C` for the smoothie driver. For this I created an `ot2_axis_to_string` util function and replaced all calls to `axis.name` or `str(axis)` or any equivalent with that function.
    - removing almost all `AxisType` TypeVars that specified the type as combination of `Axis` & `OT3Axis`
    - removing all overloaded functions relating to the two different axis types in `motion_utilities.py`
    - changes to models and router for `/motors`

# Review requests

- Check if there's any silly auto-refactor changes like `Union[OT3Axis, Axis]` -> `Union[Axis, Axis]`
- Any changes I've missed or made incorrectly? esp. in `motion_utilities.py`
- Are there any pipette settings & related files that might get affected by this?
- Look out for questions in comments
- Test if the OT3 and OT2 still work correctly! ~(working on this)~ Tested successfully!

# Risk assessment

High. Changes very low level code for OT2. 
Not much detrimental for OT3 other than making sure all scripts are updated correctly (ones in this repo and any testing/ factory scripts in use).


api/src/opentrons/hardware_control/motion_utilities.py
```diff
@@ -51,200 +54,39 @@ def offset_for_mount(
...
     left_mount_offset: Point,
     right_mount_offset: Point,
-    gripper_mount_offset: None = None,
+    gripper_mount_offset: Optional[Point] = None,
 ) -> "OrderedDict[Axis, float]":
-    ...
-
-
-@overload
-def target_position_from_absolute(
```
sanni-t, line 169: I am not sure if I should have kept the overloads for the `target_position_from_absolute` function since the TODO at the top says that `offset_for_mount` should be an overloaded function in the first place. But it didn't seem like the implementation, as it stands right now, is utilizing any overload features so I decided to remove it. Let me know if I should revert this change.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1032,7 +1037,7 @@ async def move_rel(
         )
         if max_speeds:
             checked_max: Optional[OT3AxisMap[float]] = {
-                OT3Axis.from_axis(k): v for k, v in max_speeds.items()
+                k: v for k, v in max_speeds.items()  # (spp): is this even needed?
```
sfoster1: uhhhh probably not

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -157,7 +166,7 @@ class OT3API(
     # of methods that are present in the protocol will call the (empty,
     # do-nothing) methods in the protocol. This will happily make all the
     # tests fail.
-    HardwareControlInterface[OT3Transforms, OT3Axis],
+    HardwareControlInterface[OT3Transforms, Axis],
```
sfoster1: Can we remove the parametrization over axis types entirely?

sfoster1: I know it's a lot to add but it's The Thing

api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py
```diff
@@ -50,7 +49,7 @@
 
 MOD_LOG = logging.getLogger(__name__)
 
-AxisType = TypeVar("AxisType", Axis, OT3Axis)
+AxisType = TypeVar("AxisType", bound=Axis)
```
sfoster1: Can we remove this type parametrization entirely?

api/src/opentrons/hardware_control/motion_utilities.py
```diff
@@ -51,200 +54,39 @@ def offset_for_mount(
     return offsets[primary_mount]
 
 
-@overload
```
sfoster1, line 54: ahhh im so glad to see this

robot-server/robot_server/service/legacy/models/motors.py
```diff
@@ -24,16 +25,19 @@ class EngagedMotor(BaseModel):
...
-    z: EngagedMotor
-    a: EngagedMotor
-    b: EngagedMotor
-    c: EngagedMotor
+    z_l: EngagedMotor
+    z_r: EngagedMotor
+    p_l: EngagedMotor
+    p_r: EngagedMotor
+    q: typing.Optional[EngagedMotor]  # Optional since OT2 doesn't have these axes
+    g: typing.Optional[EngagedMotor]
```
sanni-t, line 42: This is a breaking change for expected response of `/motors/engaged` endpoint. But the app doesn't use this endpoint anywhere so I think it's okay?

robot-server/tests/service/legacy/routers/test_motors.py
```diff
@@ -3,24 +3,26 @@
...
+        Axis.Z_L: True,
+        Axis.Z_R: False,
+        Axis.P_L: True,
+        Axis.P_R: True,
     }
 
     res0 = api_client.get("/motors/engaged")
     result0 = res0.json()
-    assert res0.status_code == 200
+    # assert res0.status_code == 200
```
sanni-t, line 16:
```suggestion
    assert res0.status_code == 200
```

api/src/opentrons/hardware_control/protocols/motion_controller.py
```diff
@@ -2,10 +2,10 @@
 from typing_extensions import Protocol
 
 from opentrons.types import Mount, Point
-from ..types import Axis, CriticalPoint, MotionChecks, OT3Axis
+from ..types import Axis, CriticalPoint, MotionChecks
 
 
-AxisType = TypeVar("AxisType", Axis, OT3Axis)
+AxisType = TypeVar("AxisType", bound=Axis)
```
sfoster1: Can we get rid of it?

api/src/opentrons/hardware_control/types.py
```diff
@@ -153,58 +99,40 @@ def by_mount(cls, mount: Union[top_types.Mount, OT3Mount]) -> "OT3Axis":
...
-    def mount_axes(cls) -> Tuple["OT3Axis", "OT3Axis", "OT3Axis"]:
+    def mount_axes(cls) -> None:
         """The axes which are used for moving instruments up and down."""
+        raise NotImplementedError(
+            "`Axis.mount_axes` has been removed. Use `Axis.ot2_mount_axes` or "
+            "`Axis.ot3_mount_axes` instead."
+        )
+
+    @classmethod
+    def ot2_mount_axes(cls) -> Tuple["Axis", "Axis"]:
```
sfoster1, line 115: Doesn't have to be in this pr, but we probably want to get `Axis` out of this business and drop these methods on the hardware controller classes or something. They're properties of the machine, and the hardware controller ultimately represents that.

sfoster1 (changes requested):
- Would love to see us get rid of the parametrization on `Axis` in the hardware controller protocols
- Doesn't have to be now but we should move the "machine-shape" accessors (like ot2_mounts) to the hardware controller
- Can we switch the way this works in `hardware-testing` to be changign what backs up `hardware_testing.opentrons_api.types.OT3Axis` to just be `Axis`? It would keep the diff down.

robot-server/tests/service/legacy/routers/test_pipettes.py
```diff
@@ -31,16 +31,16 @@ def test_get_pipettes(api_client, hardware, attached_pipettes):
...
             "id": attached_pipettes[types.Mount.LEFT]["pipette_id"],
         },
         "right": {
             "model": attached_pipettes[types.Mount.RIGHT]["model"],
             "name": attached_pipettes[types.Mount.RIGHT]["name"],
             "tip_length": 123,
-            "mount_axis": "a",
-            "plunger_axis": "c",
+            "mount_axis": "z_r",
+            "plunger_axis": "p_r",
```
sanni-t: Shoot! Forgot about this. I think this should return the old names of axes to prevent a breaking change for the client. I'm going to make that change unless anyone disagrees.

sanni-t:
> * Would love to see us get rid of the parametrization on `Axis` in the hardware controller protocols

Done

> * Doesn't have to be now but we should move the "machine-shape" accessors (like ot2_mounts) to the hardware controller

Left todo comments on all such accessors.

> * Can we switch the way this works in `hardware-testing` to be changign what backs up `hardware_testing.opentrons_api.types.OT3Axis` to just be `Axis`? It would keep the diff down.

That is certainly a smart idea that I did not think of when doing the refactor the first time. But now that they're all changed to Axis, it's a bit of a pain to go back and rename all the references, specially since some files in hardware-testing import Axis directly from hardware control. Can we just keep the change now? 😬

sanni-t:
Tested simple transfer protocols on both OT2 & Flex to confirm I didn't break anything. Also tested calibration on OT-2 to make sure the pipette offset endpoints return expected responses.

sfoster1 (approved):
This is awesome, thank you!

PR #12835, opened 2023-06-01
fix(api): handle absent motor status

sfoster1:
We added a latch to implement disabling stall detection that works by latching in the first time the motors respond OK (aka after they home for the first time). But for that, it has to access the motor status; and we do it in the function where we handle motor status; and so the very first response we get, we don't have a motor status to handle, and we crash.

The fix is to treat a missing motor status equivalently to a bad motor status.

Closes RQA-906


fsinapi (approved):
Yeah for sure

sfoster1:
Tested on the QA bot showing the error

PR #12837, opened 2023-06-01
fix(app): show most recent protocol run data in carousel

b-cooper:
# Overview

Fixes an issues where the oldest run of a given protocol would be displayed on the robot dashboard of the ODD. Now the most recent run of that protocol will be displayed.  This will also lead to the expected offsets being included.

### DIFF
--- a/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCarousel.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCarousel.tsx
@@ -69,9 +69,9 @@ export function RecentRunProtocolCarousel({
     >
       <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing8}>
         {sortedProtocols.map((protocol: ProtocolResource) => {
-          const run = runs.data?.data.findLast(
-            run => run.protocolId === protocol.id
-          )
+          const run = runs.data?.data
+            .reverse()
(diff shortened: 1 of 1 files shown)


shlokamin (approved):

app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCarousel.tsx
```diff
@@ -69,7 +69,7 @@ export function RecentRunProtocolCarousel({
     >
       <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing8}>
         {sortedProtocols.map((protocol: ProtocolResource) => {
-          const run = runs.data?.data.find(
+          const run = runs.data?.data.findLast(
```
koji: probably the version of ts we use doesn't support.

shlokamin: im confused by this, `findLast` is part of es6, and esnext is what our target is set to, which is the latest ECMAScript version under development. ill just reverse this list for now, but yeah this error baffles me

shlokamin: https://github.com/Opentrons/opentrons/blob/e65750c35fcbcea08a73bb2a2db045fdee79c5b6/tsconfig-base.json#L3

PR #12839, opened 2023-06-02
build(api,shared-data): replace static versioning with intervals

marcosgdiaz:
# Overview

Change static dependencies versioning with intervals. This allows opentrons to be compatible with third-party software. 

# Test Plan

We are waiting for the OT-2 to arrive and therefore we could not test it on the hardware yet.  On the other hand, test will be run on the CI when creating the pull request.

# Changelog

- Replace static dependencies versions with intervals 
# Review requests

# Risk assessment

The dependency jsonschema follows semantic versioning and not breaking changes were introduced in 4.X.X. Hereby we included up until version 5 in `jsonschema>=3.0.2,<5`

### DIFF
--- a/api/setup.py
+++ b/api/setup.py
@@ -60,12 +60,12 @@ DESCRIPTION = (
 PACKAGES = find_packages(where="src")
 INSTALL_REQUIRES = [
     f"opentrons-shared-data=={VERSION}",
-    "aionotify==0.2.0",
-    "anyio==3.3.0",
-    "jsonschema==3.0.2",
+    "aionotify>=0.2.0,<1",
+    "anyio>=3.3.0,<4",
+    "jsonschema>=3.0.2,<5",
     "numpy>=1.15.1,<2",
-    "pydantic==1.8.2",
-    "pyserial==3.5",
+    "pydantic>=1.8.2,<2",
+    "pyserial>=3.5,<4",
     "typing-extensions>=4.0.0,<5",
     "click>=8.0.0,<9",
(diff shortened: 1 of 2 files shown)


sfoster1 (changes requested):
@marcosgdiaz Thank you for the pull request!

I think we need to modify this PR slightly to add the specific versions that were previously in `INSTALL_REQUIRES` to the [pipenv files](https://github.com/Opentrons/opentrons/blob/edge/api/Pipfile) for this module; for `shared-data`; and for `robot-server`. We really want to use exact versions to test because those exact versions are the ones we use on the OT-2, as set in [buildroot](https://github.com/opentrons/buildroot). Currently, the reason the exact versions are in the setup.py here is that we delegate installing those dependencies from pipenv to here.

If we move the exact dependency specifications to the pipfiles for these modules and for the modules that use these as a dependency, I think we should be able to get the setup pys to specify ranges.

sfoster1:
@marcosgdiaz Hey, I'm sorry, I know it's been a really long time. Do version ranges like this: https://github.com/Opentrons/opentrons/pull/14420/files#diff-7bac1606de7de80231cd062b4bee4ba9e7874be65ce0dbdbfc20f31e4cc0dc31 work for you?

sfoster1:
We just merged https://github.com/Opentrons/opentrons/pull/14420 and will release it in 7.2.0, so I'm going to close this PR. Thank you for bringing this to our attention, and sorry we took so long to fix it.

PR #12843, opened 2023-06-02
feat(motor-control): use um/tick/tick for acceleration

ahiuchingau:
# Overview
We want to increase the resolution of the stepper motor acceleration so we can work with much slower acceleration. So instead of passing values as mm/tick/tick to the firmware, we're using um/tick/tick so we don't lose precision for small values.

Firmware-side changes https://github.com/Opentrons/ot3-firmware/pull/678


sfoster1 (approved):
ummmmm

PR #12844, opened 2023-06-02
feature(api, hardware-controller): allow hardware to move both mounts at the same time

TamarZanzouri:
# Overview

Move both mounts for 96c attach/detach flow.
`move_axes` in hardware controller.
Allow `machine_from_deck` to use more than 3 axis. 
Use point instead of slot for `move_to_maintenance_position` https://opentrons.atlassian.net/browse/RSS-244

# Test Plan

- Use the app on at OT-3 to attach/detach a 96c. make sure both mounts come down the z to the same location and allow easy attaching.
- Use app to test attach/detach flows with 1-8c. make sure works as expected. 
- Test protocols with movement commands and make sure it's all working as expected. we made a change to the coordinates machine calculations and we need to make sure nothing has changed. 

# Changelog

- Changed `move_to_maintenance_position`:

1. use 2 modes - AttachPlate for 96c and AttachInstrument for all other instruments.
2. make 2 calls to move_axes for both mounts when using 96c. 
3. use point instead of slot with offset for the attach/detach position. 

- added hardware `move_axes` to move a list of axis instead of a mount that move_to uses. 
- refactored `machine_from_deck` to accept more than 3 axes (to be able to move both mounts at the same time - currently not supported by head firmware)

# Review requests

does changes make sense? did I forget anything?

# Risk assessment

Medium. changes how we calculate `machine_from_deck` and this should affect all movement commands. `move_axes` should only affect `move_to_maintenance_poisition` but need to make sure nothing in the attach/detach flows is broken.


api/src/opentrons/hardware_control/motion_utilities.py
```diff
@@ -322,30 +322,42 @@ def machine_from_deck(
...
-    to_transform = Point(*(tp for ax, tp in deck_pos.items() if ax in ax.gantry_axes()))
+    try:
+        point_for_z_axis = {
+            axe: Point(x=deck_pos[type(axe).X], y=deck_pos[type(axe).Y], z=pos)
+            for axe, pos in deck_pos.items()
+            if axe in axe.mount_axes()
+        }
+    except KeyError:
+        raise ValueError(
+            "Moves must specify either exactly an " "x, y, and (z or a) or none of them"
```
TamarZanzouri:
```suggestion
            "Moves must specify either exactly an x, y, and (z or a) or none of them"
```

api/src/opentrons/hardware_control/motion_utilities.py
```diff
@@ -322,30 +322,42 @@ def machine_from_deck(
...
     # need to transform
     machine_pos = {
         ax: pos for ax, pos in deck_pos.items() if ax not in ax.gantry_axes()
     }
-    if len(to_transform) != 3:
-        raise ValueError(
-            "Moves must specify either exactly an " "x, y, and (z or a) or none of them"
-        )
 
     # Type ignored below because linal.apply_transform (rightly) specifies
```
TamarZanzouri: do we still need this comment?

api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
```diff
@@ -59,18 +75,35 @@ async def execute(
...
+            await ot3_api.move_axes(
+                {
+                    OT3Axis.Y: _ATTACH_POINT.y,
+                    OT3Axis.X: _ATTACH_POINT.x,
+                    OT3Axis.Z_L: _PLATE_ATTACH_Z_POINT,
+                }
+            )
+            await ot3_api.move_axes(
+                {
+                    OT3Axis.Y: _ATTACH_POINT.y,
```
TamarZanzouri, line 102: should we leave this explicit of let the `move_axes` use `current_position` ?

api/src/opentrons/hardware_control/motion_utilities.py
sfoster1: don't think so

sfoster1 (approved):
Looks good to me!

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+        The effector of the pipette mount axis are the mount critical points but only in z.
+        """
+        assert position, "move_axes requires a none empty position"
+
+        if not self._current_position:
+            await self.refresh_positions()
+
+        for axis in position.keys():
+            if not self._backend.axis_is_present(axis):
+                # create a custom error
```
Laura-Danielle: Is this comment relevant still?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+        if not self._backend.check_encoder_status(list(position.keys())):
+            await self.home()
+
+        valid_motor = self._current_position and self._backend.check_motor_status(
+            list(position.keys())
+        )
+        if not valid_motor:
+            raise MustHomeError("Current position is invalid; please home motors.")
+
+        absolute_positions: "OrderedDict[OT3Axis, float]" = OrderedDict()
```
Laura-Danielle, line 1009: Curious why we can't rely solely on `current_position` here

api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
```diff
@@ -1,31 +1,44 @@
...
 MoveToMaintenancePositionCommandType = Literal["calibration/moveToMaintenancePosition"]
 
 
+class MaintenancePosition(str, enum.Enum):
+    """Maintenance position options."""
+
+    AttachPlate = "attachPlate"
+    AttachInstrument = "attachInstrument"
+
+
```
SyntaxColoring: For future reference, [I think](https://opentrons.atlassian.net/l/cp/2abqhnJ1) we should generally avoid subclassing `Enum`s from `str` unless we really need to, and I don't think we need to here.

(Also, less importantly, I think our usual convention is to capitalize the Python attributes like `ATTACH_PLATE` instead of `AttachPlate`).

Both of these are easy to change later, so no need to do it in this PR.

SyntaxColoring (approved):
The Protocol Engine side of this looks good to me. 🔧  :shipit:

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+        current_position = self._current_position
+        if OT3Axis.X in position:
+            absolute_positions[OT3Axis.X] = position[OT3Axis.X]
+        else:
+            absolute_positions[OT3Axis.X] = current_position[OT3Axis.X]
+        if OT3Axis.Y in position:
+            absolute_positions[OT3Axis.Y] = position[OT3Axis.Y]
+        else:
+            absolute_positions[OT3Axis.Y] = current_position[OT3Axis.Y]
+
```
sanni-t, line 1019: Hmm, so we will always be moving in X -> Y -> Z, ya? Although the current usage of `move_axes` for moving to maintenance position looks fine, we should verify whether, and under which circumstances can this sequence result in the gripper jaws colliding with the thermocycler lid's pins. We usually don't run into this problem in protocols because we are moving the pipettes only to access labware. And this labware is not close enough to the lid to have the pipette crash into it.

I imagine if `move_axes` gets a position of `{'OT3Axis.X': 0, ' OT2Axis.Y': 0}` when the gantry is in home location *and* the thermocycler is in the caddy, it might run into the lid clips. But if we do Y -> X then the gripper will collide with clips if the starting point is (x=0, y=0). I know we ask users to clear the deck before doing any instrument calibration but I think we should assume that some users would still leave them in. I don't have an answer for this issue right now but we should at test this out and document if the collision happens.

api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
```diff
@@ -59,18 +77,27 @@ async def execute(
...
+        ot3_api = ensure_ot3_hardware(
+            self._hardware_api,
         )
+        if params.maintenancePosition == MaintenancePosition.AttachInstrument:
+            mount_to_axis = OT3Axis.by_mount(params.mount.to_hw_mount())
+            await ot3_api.move_axes(
+                {
+                    OT3Axis.Y: _ATTACH_POINT.y,
+                    OT3Axis.X: _ATTACH_POINT.x,
+                    mount_to_axis: _INSTRUMENT_ATTACH_Z_POINT,
```
sanni-t: Ah I see this is going in order Y -> X -> Z_L/ Z_R/ Z_G but I think `move_axes` changes that order to become X -> Y -> Z_L/ Z_R/ Z_G ? Unless I am misreading something..

api/src/opentrons/hardware_control/ot3api.py
sfoster1: > Hmm, so we will always be moving in X -> Y -> Z, ya? 

No, this code is just prepping the position dict for later - it doesn't imply order of motion. This is a single motion from one point in the state space to another, just like `move_to`.

> Although the current usage of `move_axes` for moving to maintenance position looks fine, we should verify whether, and under which circumstances can this sequence result in the gripper jaws colliding with the thermocycler lid's pins. 

This is true but I think it's a concern of the protocol engine command that calls this - the hardware layer doesn't really know or care about that.

api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
sfoster1: No, this is just like `move_to` - this is moving to the single specified point in one motion.

api/src/opentrons/hardware_control/ot3api.py
sanni-t: Ah! ok. My bad

sanni-t (approved):
LGTM

api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
```diff
@@ -80,20 +80,35 @@ async def execute(
         ot3_api = ensure_ot3_hardware(
             self._hardware_api,
         )
-        if params.maintenancePosition == MaintenancePosition.AttachInstrument:
+        current_position = await ot3_api.gantry_position(Mount.LEFT)
+        max_travel_z = ot3_api.get_instrument_max_height(Mount.LEFT)
+        way_points = self._state_view.motion.get_movement_waypoints_to_coords(
+            origin=current_position,
+            dest=_ATTACH_POINT,
```
SyntaxColoring, line 87: Is this correct? Won't this plan an arc to (0, 100, 0), trying to drive the mount to z=0, which is too low?

TamarZanzouri: I tested this and the z does not move down. If I supply the current z I'm getting an exception that the z is out of bound? something like that

TamarZanzouri: this moves just the x, y

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
             expect_stalls=_expect_stalls,
         )
 
+    async def move_axes(  # noqa: C901
```
Laura-Danielle, line 983: Overall, I'm not a huge fan of this solution. However, I am fine with keeping it in here for now as we only have a use-case for `move_to` point for the mounts. Could we create a spike ticket for utilizing a `BOTH` mount in the future?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+        if OT3Axis.X in position:
+            absolute_positions[OT3Axis.X] = position[OT3Axis.X]
+        else:
+            absolute_positions[OT3Axis.X] = current_position[OT3Axis.X]
+        if OT3Axis.Y in position:
+            absolute_positions[OT3Axis.Y] = position[OT3Axis.Y]
+        else:
+            absolute_positions[OT3Axis.Y] = current_position[OT3Axis.Y]
+
+        have_z = False
```
Laura-Danielle, line 1020: imo I think we should require this command to be called with a Z axis. Was there a reason for not doing that?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+        self,
+        position: Mapping[OT3Axis, float],
+        speed: Optional[float] = None,
+        max_speeds: Optional[Dict[OT3Axis, float]] = None,
+    ) -> None:
+        """Moves the effectors of the specified axis to the specified position.
+        The effector of the x,y axis is the center of the carriage.
+        The effector of the pipette mount axis are the mount critical points but only in z.
+        """
+        assert position, "move_axes requires a none empty position"
```
Laura-Danielle: Can we remove this assert?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+            absolute_positions[OT3Axis.Y] = position[OT3Axis.Y]
+        else:
+            absolute_positions[OT3Axis.Y] = current_position[OT3Axis.Y]
+
+        have_z = False
+        for axis in [OT3Axis.Z_L, OT3Axis.Z_R, OT3Axis.Z_G]:
+            if axis in position:
+                have_z = True
+                if OT3Axis.Z_L:
+                    carrige_effectors_offset = self._robot_calibration.left_mount_offset
```
Laura-Danielle: Do we expect there to only ever be one Z position? Otherwise this would change the carriage effector offset while we loop through.

Laura-Danielle (commented):
I have a few comments/questions.

Not a huge fan of the `move_axes` implementation in general, but I am fine with it for now since we only have 1 use case of moving both mounts simultaneously.

api/src/opentrons/hardware_control/ot3api.py
sfoster1: I guess we could use a `NoPipetteAttached` or something similar

api/src/opentrons/hardware_control/ot3api.py
sfoster1: We could `deepcopy()` it to initialize `absolute_positions`, if that's what you mean? This method's general-purpose enough to use for gantry movement if you want, so it has to handle somebody specifying `X` and not `Y` or vice versa, and since the machine position translation needs either all of (x, y, z) or none of them we have to have values for the missing ones.

api/src/opentrons/hardware_control/ot3api.py
sfoster1: honestly not having a command where you could tell just one axis to go to a position and leave the others where they were was something that's been bugging me for a while, I like having this capability

api/src/opentrons/hardware_control/ot3api.py
sfoster1: `carriage_effectors_offset` looks like it's only used on line 1037 which is inside the loop body, the if block is determining what it is and then we get a new value next iteration

```suggestion
                    carriage_effectors_offset = self._robot_calibration.left_mount_offset
```

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+                elif OT3Axis.Z_R:
+                    carrige_effectors_offset = (
```
```suggestion
                    carriage_effectors_offset = (
```

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -979,6 +980,70 @@ async def move_to(
...
+                        self._robot_calibration.right_mount_offset
+                    )
+                else:
+                    carrige_effectors_offset = (
```
```suggestion
                    carriage_effectors_offset = (
```

api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
```diff
@@ -1,31 +1,44 @@
...
 if TYPE_CHECKING:
     from opentrons.hardware_control import HardwareControlAPI
     from ...state import StateView
 
 
-# Question (spp): Does this offset work for gripper mount too?
-# These offsets are based on testing attach flows with 8/1 channel pipettes
-_INSTRUMENT_ATTACH_OFFSET = Point(y=10, z=400)
+# These offsets supplied from HW
+_ATTACH_POINT = Point(x=0, y=100)
```
sfoster1: I wonder if we can drop these in the machine config file or deck def at some point

Laura-Danielle: hm not sure if that makes much sense since we'll only be using these locations in the protocol engine command.

api/src/opentrons/hardware_control/ot3api.py
Laura-Danielle: This check feels redundant a bit to me, unless we only check instrument axes.

TamarZanzouri: oops forgot toi implement this

api/src/opentrons/hardware_control/ot3api.py
Laura-Danielle: sorry I meant to put this comment one 1014 originally, but you can ignore it.

SyntaxColoring (approved):
Protocol Engine changes since my last review look good!

PR #12845, opened 2023-06-02
fix(app): Protocol Run DQA updates

jgbowser:
# Overview

Fixes some issues noted during design qa on the protocol run screens of the ODD including:
**Header**
- Copy is overflowing the page
- should be a fixed height of around 108px 
- Looks like there’s a jump in header between first and second page
- Missing controls from second page

**Content**
- Update the close button to match designs (stroke not thick enough) 
- protocol step: the blue background should hug the content
- when you swipe back from the second page (run log) onto the first page, the current step text fades in from bottom. Need to nix! 

closes RAUT-434

Screenshot 2023-06-02 at 12 26 40 PM
Screenshot 2023-06-02 at 12 26 48 PM
Screenshot 2023-06-02 at 12 27 31 PM

# Test Plan

Ran through a couple of protocols via the sim that allowed me to pause the execution of the protocol and verify that the noted issues were no longer present.

# Changelog

- updated some styling to the headers in the protocol command and list views for consistency
- updated styling on the protocol command list view to ensure the text didn't overflow the page
- added a sessionStorage check to ensure the command text animation only occurred on first render and not everytime the screen is swiped between the current command and the command list
- updated the cancel run button's icon to match the designs by adding the `ot-close-thick` icon

# Review requests

Ensure the changes seem appropriate and run a quick test to ensure the noted issues in the ticket have all been resolved

# Risk assessment

Low, mostly style updates


app/src/organisms/OnDeviceDisplay/RunningProtocol/CurrentRunningProtocolCommand.tsx
```diff
@@ -113,6 +130,20 @@ export function CurrentRunningProtocolCommand({
   const currentCommand = robotSideAnalysis?.commands.find(
     (c: RunTimeCommand, index: number) => index === currentRunCommandIndex
   )
+
+  let shouldAnimate = true
+  const lastAnimatedCommand = sessionStorage.getItem('lastAnimatedCommand')
```
brenthagen: i'm not sure about the sessionStorage approach - would it work to track last animated command with `useRef` on the parent `RunningProtocol`?

jgbowser: Yeah let me take a look at that

jgbowser: updated, works just the same, working on a new test setup now

app/src/organisms/OnDeviceDisplay/RunningProtocol/CurrentRunningProtocolCommand.tsx
```diff
@@ -66,17 +67,33 @@ const RUN_TIMER_STYLE = css`
   color: ${COLORS.darkBlackEnabled};
 `
 
-const COMMAND_ROW_STYLE = css`
+const COMMAND_ROW_STYLE_ANIMATED = css`
```
brenthagen, line 70: css template syntax suggestion: if the non-animation values are identical, can reference like this to avoid duplication:
```
const COMMAND_ROW_STYLE_ANIMATED = css`
  ${COMMAND_ROW_STYLE}
  animation: ${fadeIn} 1.5s ease-in-out;
`
```

brenthagen (approved):
LGTM !

PR #12846, opened 2023-06-02
feat(api): Enforce apiLevel 2.15 for Flex-style deck slot names

SyntaxColoring:
# Overview

Closes RLAB-338.

# Test plan

Existing automated unit tests should be sufficient for this.

# Changelog

* Raise `APIVersionError` if you provide an OT-3–style deck slot name like `"A1"` to methods like `load_labware()`, and your protocol's `apiLevel` is less than `2.15`.
* Public-facing documentation (docs.opentrons.com) is *not* done in this PR. I'm doing that in RTC-286.

# Review requests

See comments.

# Risk assessment

Low.


api/src/opentrons/protocol_api/validation.py
```diff
@@ -32,6 +34,9 @@
     from .labware import Well
 
 
+_COORDINATE_STYLE_DECK_LABELS_ADDED_IN = APIVersion(2, 15)
```
SyntaxColoring: Is there a better name for this? I feel like I'm inconsistent. "OT-3–style," "coordinate-style," "chess-style," ...

jbleon95: Battleship-style 🛥️ 

But seriously, I usually prefer including `VERSION_GATE` just to be more clear what this is being used for, so something like `_COORDINATE_DECK_LABEL_VERSION_GATE` would be good, imo

jbleon95 (approved):
LGTM, besides any name changes you want I think its good to go

jbleon95 (approved):
LGTM, besides any name changes you want I think its good to go

PR #12847, opened 2023-06-02
feat(hardware): Added a driver to read and write data from the Flex eeprom

vegano1:
# Overview

We need a way to read/write to the eeprom on the carrier board, this python module will let us do that from userspace. 

Closes: [RET-1346](https://opentrons.atlassian.net/browse/RET-1346)

**Property Format**
Data is written to the eeprom in the form of serialized property byte string with the following format
`prop_id(1b) + data_length(1b) + data(1-253b)`

- The `prop_id` is a 1-byte 1-0xFE value and denotes the unique property being stored: **NOTE: 0xFF is invalid**
- The `data_length` is a 1-byte 1-0xFD value and denotes the length of the data 
- The `data` is a 1-253b long hexadecimal value representing some data that corresponds to a prop_id.

**Properties**
Properties represent some piece of data stored on the eeprom, and are categorized by a `prop_id`, a `data_type`, and a `value`. When we find a valid serialized property byte string, which means we have a bytestring matching our format denoted above where the `length` byte corresponds to the actual size of the data. We deserialize this bytestring based on the specific prop_id, the type of data that prop_id is associated with, and how long the data should be. We use the data_type to encode/decode the data accordingly and create a Property object with the parsed data type.

**API**
In order to interact with the eeprom we have the concept of properties, and we use the methods `self.property_write` and `self.property_read` to read and write these properties to the eeprom.

The `self.property_write` method takes a set of property and value tuples `Set[Tuple[property_id, Any]])` which are serialized, chunked up, and written to the eeprom. This method returns a set of PropId for the properties that were successfully written to the eeprom.

The `self.property_read` method takes a set of PropId enums and returns a set of Property objects which represent the eeprom data.

# Test Plan

- [x] Make sure we can read/write data to/from the eeprom on the flex

# Changelog

- Added eeprom module that lets us read/write data to/from any eeprom device in `/sys/bus/i2c/devices`

# Review requests

- Does this approach make sense?
- Am I missing something?

# Risk assessment
Low, flex work

[RET-1346]: https://opentrons.atlassian.net/browse/RET-1346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hardware/opentrons_hardware/drivers/eeprom/types.py
```diff
@@ -0,0 +1,71 @@
...
+    PropId.UNIT_NUMBER: PropType.INT,
+}
+
+PROP_TYPE_SIZE = {
+    PropType.BYTE: 1,
+    PropType.CHAR: 1,
+    PropType.SHORT: 2,
+    PropType.INT: 4,
+    PropType.STR: 254,
+    PropType.BIN: 254,
```
sfoster1: same?

sfoster1 (changes requested):
I think if we're doing TLV, we should do TLV all the way. We can drop the start and end tokens, because we have the type, length and value. Require the type to not be 255 and you can't be confused with erased data; then, you can read out `size-1` bytes, nice and easy.

hardware/opentrons_hardware/drivers/eeprom/eeprom.py
```diff
@@ -0,0 +1,252 @@
...
+    def _write(self, data: bytes, address: int = 0) -> int:
+        """Write data to the eeprom at the given address."""
+        if self._eeprom_fd == -1:
+            raise RuntimeError(
+                "Could not read from eeprom, file descriptor is unavailable"
+            )
+
+        try:
+            logger.debug(
+                f"Writting {len(data)} bytes to address {hex(address)} for device {self.name} - {data.hex()}"
```
sfoster1:
```suggestion
                f"Writing {len(data)} bytes to address {hex(address)} for device {self.name} - {data.hex()}"
```

sfoster1 (approved):
Looks good to me!

PR #12848, opened 2023-06-02
fix(app): scrape historic offsets when start run setup on ODD, fix horizontal scroll

b-cooper:
# Overview

Fix Scraping and cloning historic offsets when creating a run from the ODD. Also fix the horizontal
scroll behavior on the pinned protocol and run carousels

- Add historic offset scraping to the start setup button on the protocol detail page of the ODD
- replace useSwipe usage with horizontal scrolling containers in carousels and run monitoring screen


app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
```diff
@@ -78,6 +100,19 @@ export function RecentRunProtocolCard({
...
+    [RUN_STATUS_SUCCEEDED]: t('completed'),
+    [RUN_STATUS_FAILED]: t('failed'),
+  }
+  const formattedLastRunTime = formatDistance(
+    new Date(runData.createdAt),
+    new Date(),
+    {
+      addSuffix: true,
+    }
+  ).replace('about ', '')
```
shlokamin, line 115: rather than doing a string replace can we use built in options? i think if you set ` addSuffix: false` (or omit because i think it defaults to `false`) that might work

https://date-fns.org/v2.30.0/docs/formatDistance

shlokamin (approved):
you've got some linting errors but code lgtm. will test on my dev kit soon if u have not already

app/src/pages/ProtocolDashboard/PinnedProtocolCarousel.tsx
```diff
@@ -66,36 +27,42 @@ export function PinnedProtocolCarousel(props: {
...
+              setShowDeleteConfirmationModal={setShowDeleteConfirmationModal}
+            />
+          )
+        })}
+      </Flex>
+    </CarouselWrapper>
   )
 }
+
+const CarouselWrapper = styled.div`
```
ewagoner, line 57: Man, this works like a champ. Why did I think we had to use the swipe gesture?

ewagoner: Ah... scrolling with a mouse doesn't work on the simulated ODD. We were trying to make sure both ways work. I think natural scrolling with the finger is more important than using a mouse though. This is aces.

b-cooper: I was able to scroll with the mouse using simulated odd (`make dev-odd -C app`) 🤷

PR #12853, opened 2023-06-05
fix(hardare-testing): Add blowout to jog-ot3 script

andySigler:
# Overview

The blow-out command was mistakenly left out of the input parser in the jog-ot3 script.

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
makes sense

PR #12854, opened 2023-06-05
fix(hardware-testing): DVT-2 diagnostics script fixes

andySigler:
# Overview

Some things that were blocking DVT-2 builds (mostly to do with machines both running 0.7.0 and 0.8.0):
1. Stall-detection feature-flag settings needed to be allowed to fail
2. The imported default-deck-transform variable name was changed on edge, so a try/except was added to the import
3. No longer moving pipette to bottom at the beginning of jog script (took too much time away from SZ team)

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
cool cool

PR #12857, opened 2023-06-05
fix(api): home plunger & mount axis properly after drop tip

ahiuchingau:
# Overview
Pipette plunger and mount were not homed properly in `ot3api.drop_tip()` function.


Laura-Danielle (changes requested):
Why are we defaulting to homing again after drop tip? Can we not rely on the head encoders?

andySigler:
we should discuss with HWE before merging this

Laura-Danielle (approved):
ok looks fine to me

sfoster1 (approved):
okay. we should really make this not required tho

PR #12858, opened 2023-06-05
feat(api, shared-data): add pipette plunger backlash compensation

ryanthecoder:
# Overview
adding backlash compensation in line with the recommendations from the hardware team. whenever the pipette plunger moves upwards we are going to move slightly past the target position then back down to the target position. 
We don't compensate on downward movements since it could create a situation where we aspirate at the destination well.

Since the gen1 and gen2 pipettes backlash is handled internally in the ot2 code we are setting their value in the shared data to 0.0 

# Test Plan

# Changelog

# Review requests

# Risk assessment


shared-data/pipette/definitions/2/general/single_channel/p300/1_3.json
```diff
@@ -32,5 +32,6 @@
   "channels": 1,
   "shaftDiameter": 5.0,
   "shaftULperMM": 19.635,
-  "backCompatNames": []
+  "backCompatNames": [],
+  "backlashDistance": 0.1
```
Laura-Danielle: We want backlash values for the GEN1 pipettes as well? (those are the v1 versions)

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1421,6 +1421,14 @@ async def _move_to_plunger_bottom(
             # save time by using max speed
             max_speeds = self.config.motion_settings.default_max_speed
             speed = max_speeds[self.gantry_load][OT3AxisKind.P]
+        if current_pos > target_pos[pip_ax]:
```
Laura-Danielle, line 1420: Shouldn't this check happen after the original move to bottom `_move` on line 1432?

api/tests/opentrons/hardware_control/test_ot3_api.py
```diff
@@ -1160,7 +1160,7 @@ async def test_move_to_plunger_bottom(
     ot3_hardware._current_position[pip_ax] = target_pos[pip_ax] + 1
     mock_move.reset_mock()
     await ot3_hardware.prepare_for_aspirate(mount)
-    mock_move.assert_called_once_with(
+    mock_move.assert_called_with(
```
Laura-Danielle, line 1173: Can we add a more comprehensive test that checks the distance moved and also adds in checks for when the current position is not more than the target position?

Laura-Danielle (commented):
Had some comments/questions.

api/src/opentrons/hardware_control/ot3api.py
ryanthecoder: So the behavior we want is, whenever the plunger is moving up, we travel past the target distance, then back down to where we want to be. This was at the advice of hardware as the best way to compensate this.

andySigler (approved):

PR #12860, opened 2023-06-05
feat(api,robot-server): expose module calibration offsets via /modules endpoint.

vegano1:
# Overview

The clients need to know if there is any module calibration data for the attached modules, this pr will expose this data over the /modules endpoint.

# Test Plan

- [x] Test on the flex and make sure modules are reporting their calibration data via the `/modules` endpoint
- [x] Make sure the module offset data is both re-loaded after a re-calibration and that the `/modules` endpoint contains the new data.
- [x] Make sure the OT2 does not respond with any module calibration offset data

# Changelog

- added `module_offset: ModuleCalibrationData` to the `/modules` endpoint response
- added `@lru_cache` to load calibration offset from disk functions so we limit IO operations. Note, the cache is cleared if a new calibration is stored, this way if the `/modules` endpoint is hit it will retrieve the latest data.
- made the `slot: int` arg for calibrations optional as it's not required for loading from disk

# Review requests

- Anything I'm missing?
- We changed deck slot names to a coordinate system like "A1, D2, etc", we still report them as integers here, maybe change in a follow-up pr?

# Risk assessment

Med, Flex only work but this code path is also hit by the OT2 although it has no effect then.


sfoster1 (approved):
Looks right to me, and I think we'd be ok with the ot2 stuff as long as the robot server ot2-only tests pass

PR #12864, opened 2023-06-06
feat(api): Monitor for overpressure on pipetting commands

Laura-Danielle:
# Overview

Although a low chance, we should monitor whether the pressure sensor ever exceeds its max rated value.

# TODO

- [x] Finish tests for `tool_sensors`
- [x] Finish tests for `ot3controller`
- [x] Add a check in the blow out function

# Test Plan

- [x] Execute a protocol with a few pipetting functions (i.e. aspirate/dispense/blowout/drop tip) and verify via the serial log or CAN logs that a bind request for max threshold is sent. To ease testing, you can also tell the request to log so you can see the pressure values being print to screen.
- [x] Using monitor sensors, determine a sufficient occlusion material (maybe a piece of cloth?) to get it to exceed 3.9kpa. This should cause the firmware to throw and error and stop all movement.

# Changelog

- Added tests at all levels from scheduler, tool_sensors and up to ot3controller
- Added a context manager that will throw an error if it receives an error message from firmware
- This currently only monitors 1 pressure sensor for a given pipette, we should figure out a nice way to monitor both sensors and can maybe tackle it as part of the partial tip pick up project.

# Review requests

Not a huge fan of the partial function workaround, but I wanted the ot3controller to have an actual context manager so we could nest specific moves inside the pipetting functions without pushing it down to `tool_sensors.py`. If anyone has better suggestions, I am definitely open.

Also please let me know if I missed any high risk pipetting functions!

# Risk assessment

Ideally low, but this could have unforeseen consequences regarding anything utilizing the i2c bus (for example the eeprom). We should closely monitor in the coming weeks.

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -939,12 +939,23 @@ class OT3Controller:
         self, mount: OT3Mount, sensor_id: SensorId = SensorId.S0
     ) -> AsyncIterator[None]:
         tool = sensor_node_for_mount(OT3Mount(mount.value))
-        # TODO we should switch the sensor type based on the channel
+        # FIXME we should switch the sensor type based on the channel
         # used when partial tip pick up is implemented.
-        # TODO we should also monitor pressure in all available channels
+        # FIXME we should also monitor pressure in all available channels
         provided_context_manager = check_overpressure(self._messenger, tool, sensor_id)
-        async with provided_context_manager():
+        async with provided_context_manager() as handle_overpressure:
             yield
+
+        def _pop_queue() -> Optional[ErrorCode]:
+            try:
+                return handle_overpressure.get_nowait()
+            except asyncio.QueueEmpty:
+                return None
+
+        if _pop_queue():
+            raise RuntimeError(
+                f"The pressure sensor on the {mount} mount has exceeded operational limits."
+            )
 
     async def liquid_probe(
         self,
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -34,6 +34,7 @@ from .ot3utils import (
     create_home_group,
     node_to_axis,
     sensor_node_for_mount,
+    sensor_node_for_pipette,
     sensor_id_for_instrument,
     create_gripper_jaw_grip_group,
     create_gripper_jaw_home_group,
@@ -938,17 +939,21 @@ class OT3Controller:
     async def monitor_overpressure(
         self, mount: OT3Mount, sensor_id: SensorId = SensorId.S0
     ) -> AsyncIterator[None]:
-        tool = sensor_node_for_mount(OT3Mount(mount.value))
+        tool = sensor_node_for_pipette(OT3Mount(mount.value))
         # FIXME we should switch the sensor type based on the channel
         # used when partial tip pick up is implemented.
         # FIXME we should also monitor pressure in all available channels
-        provided_context_manager = check_overpressure(self._messenger, tool, sensor_id)
-        async with provided_context_manager() as handle_overpressure:
+        provided_context_manager = await check_overpressure(
+            self._messenger, tool, sensor_id
+        )
+        errors: asyncio.Queue[ErrorCode] = asyncio.Queue()
+
+        async with provided_context_manager() as errors:
             yield
 
         def _pop_queue() -> Optional[ErrorCode]:
             try:
-                return handle_overpressure.get_nowait()
+                return errors.get_nowait()
             except asyncio.QueueEmpty:
                 return None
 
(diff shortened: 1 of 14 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1063,6 +1065,33 @@ def _axis_map_to_present_nodes(
...
+        # FIXME we should switch the sensor type based on the channel
+        # used when partial tip pick up is implemented.
+        # FIXME we should also monitor pressure in all available channels
+        provided_context_manager = await check_overpressure(
+            self._messenger, tool, sensor_id
+        )
+        errors: asyncio.Queue[ErrorCode] = asyncio.Queue()
+
+        async with provided_context_manager() as errors:
+            yield
```
sfoster1: this probably wants to have a try/finally so the cleanup always (at least tries to) happen

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1063,6 +1065,33 @@ def _axis_map_to_present_nodes(
...
+
+        def _pop_queue() -> Optional[ErrorCode]:
+            try:
+                return errors.get_nowait()
+            except asyncio.QueueEmpty:
+                return None
+
+        if _pop_queue():
+            raise RuntimeError(
```
sfoster1: let's get a custom exception for this

sfoster1 (changes requested):
I think we should add a try/finally and, bowing to the wisdom of history, I think we might want to add a feature flag that lets you turn this off juuuust in case

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -933,6 +936,42 @@ def _axis_map_to_present_nodes(
...
+            # FIXME we should also monitor pressure in all available channels
+            provided_context_manager = await check_overpressure(
+                self._messenger, tool, sensor_id
+            )
+            errors: asyncio.Queue[ErrorCode] = asyncio.Queue()
+
+            async with provided_context_manager() as errors:
+                try:
+                    yield
+                finally:
```
sfoster1, line 956: Not quite, I think. As implemented, we'll now correctly do whatever cleanup we need but we'll also swallow and blackhole the exception. We need to put the `_pop_queue` definition and call _inside_ the `finally` block so that it both runs even if there's an exception _and_ lets that exception bubble up afterward.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -933,6 +936,42 @@ def _axis_map_to_present_nodes(
...
+                except asyncio.QueueEmpty:
+                    return None
+
+            if _pop_queue():
+                raise OverPressureDetected(
+                    f"The pressure sensor on the {mount} mount has exceeded operational limits."
+                )
+        else:
+            try:
+                yield
```
sfoster1: conversely here we don't need this `try`/`finally` at all

sfoster1 (changes requested):
One more step to error handling, I think it'll be pretty quick.

hardware/opentrons_hardware/sensors/scheduler.py
```diff
@@ -469,3 +469,74 @@ def _filter(arbitration_id: ArbitrationId) -> bool:
...
+                    binding=SensorOutputBindingField(
+                        SensorOutputBinding.max_threshold_sync.value
+                    ),
+                )
+            ),
+            expected_nodes=[target_sensor.node_id],
+        )
+        if error != ErrorCode.ok:
+            log.error(
+                f"recieved error {str(error)} trying to bind sensor output on {str(target_sensor.node_id)}"
```
sfoster1, line 518: shouldn't we raise here?

hardware/opentrons_hardware/sensors/scheduler.py
```diff
@@ -469,3 +469,74 @@ def _filter(arbitration_id: ArbitrationId) -> bool:
...
+                        sensor_id=SensorIdField(target_sensor.sensor_id),
+                        binding=SensorOutputBindingField(
+                            SensorOutputBinding.none.value
+                        ),
+                    )
+                ),
+                expected_nodes=[target_sensor.node_id],
+            )
+            if error != ErrorCode.ok:
+                log.error(
```
sfoster1, line 540: shouldn't we raise here?

sfoster1 (approved):
Ok looks good now if it passes tests though shouldn't we raise on sensor config errors?

hardware/opentrons_hardware/sensors/scheduler.py
Laura-Danielle: We don't raise anywhere else that we receive an error from the sensor tasks so I decided not to add that here. Can do that in a follow-up PR if you want. I will not be merging this until the andys have a chance to test it.

hardware/opentrons_hardware/sensors/scheduler.py
Laura-Danielle: See above.

hardware/opentrons_hardware/sensors/scheduler.py
sfoster1: ah ok

andySigler (approved):
Test on a bot and works as expected 👍 . Couple of comments for a followup PR:

There are two scenarios where pressure should be monitored but it is not yet as currently implemented:

- prepare for aspirate
- api.home()

Might be easier to manage all these cases if the pressure monitoring was instead happening inside `OT3Controller.home()` and `OT3Controller.move()`, and it's called whenever a plunger axis is passed in as an argument, but we could save that for a followup PR.

PR #12865, opened 2023-06-06
fix(api): Slower XY speeds/accelerations for DVT2

andySigler:
# Overview

Better reliability means slowing things down, it seems. Values were found through cycle testing both LOW and HIGH throughput gantry-loads.

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
😢

sfoster1 (approved):
suh weet

PR #12866, opened 2023-06-06
fix(api): set the module firmware dir to /usr/lib/firmware

vegano1:
# Overview

We are adding module binaries to the flex filesystem in [this](https://github.com/Opentrons/oe-core/pull/80) oe-core pull request. The binaries are installed in `/usr/lib/firmware` along with the rest of the firmware, so let's move the module binary dir to `/usr/lib/firmware` for consistency.

# Test Plan

- [x] Make sure that the robot server is able to find the module firmware binaries in `/usr/lib/firmware`
- [x] Make sure that the `hasAvailableUpdate`  flag in the `/modules` endpoint response changes if the binary firmware version is greater than whats running on the module.
- [x] Make sure that there is no functionality change on the OT2 

# Changelog

- Added `firmware_binaries_dir` ConfigElement that points to `/usr/lib/firmware`
- Removed `ROBOT_FIRMWARE_DIR` constant as we are now using the config index

# Review requests

# Risk assessment

Low, flex only work

### DIFF
--- a/api/src/opentrons/__init__.py
+++ b/api/src/opentrons/__init__.py
@@ -18,7 +18,7 @@ from opentrons.config import (
     name,
     robot_configs,
     IS_ROBOT,
-    ROBOT_FIRMWARE_DIR,
+    get_opentrons_path,
 )
 from opentrons.util import logging_config
 from opentrons.protocols.types import ApiDeprecationError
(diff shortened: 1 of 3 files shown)


sfoster1 (changes requested):
can we maybe put this in the config index?

sfoster1 (approved):
Nice, thank you!

PR #12867, opened 2023-06-07
feat(hardware-testing): Dispense liquid-class settings are configurable per test volume

andySigler:
# Overview

This PR expands the dispense liquid-class configurations dictionary so that dispense settings can be set per test volume. This allows, for example, the dispense flow-rate to be different at 5uL vs 50uL.

Also updates all pipette models' leading-air-gap to the latest/greatest from SZ testing.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
wow this is a lot

PR #12870, opened 2023-06-07
fix(hardware): increase move group timeout

fsinapi:
# Overview

We've had at least a few crashes due to move group timeouts because the hardware controller caches the motor positions based on the _last_ move group stage completion message from each axis. Therefore, if an axis only sends a completion for the 1st or 2nd stage out of 3, the hardware controller calculates the _next_ move with the wrong starting position (and often smacks into something).

This PR takes the easy way out by bumping up the timeout by a bit. It still tracks if the move group takes longer than the _original_ timeout (1.1x move group length) to help in debugging when/why this happens, but hopefully this change will reduce the number of legitimate crashes in the near term.

# Test Plan

Tested on a robot by uploading changes but with the "expected" time using a 0.8x multiplier so every move should time out. Confirmed that all successful movements would log the time it took vs the "expected" time as a warning.

# Changelog

- Increase move group timeout to 2x move group estimated time
- Add logging if move group takes longer than 1.x calculated time

# Review requests

# Risk assessment

### DIFF
--- a/hardware/opentrons_hardware/hardware_control/move_group_runner.py
+++ b/hardware/opentrons_hardware/hardware_control/move_group_runner.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 import logging
 from typing import List, Set, Tuple, Iterator, Union, Optional
 import numpy as np
+import time
 
 from opentrons_hardware.firmware_bindings import ArbitrationId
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Excellent, thank you!

PR #12873, opened 2023-06-07
docs(api): Describe Flex-style deck slot names in the Python Protocol API

SyntaxColoring:
# Overview

Update https://docs.opentrons.com for the Flex's new style of deck slot names, like `"A1"`. Closes RTC-286.

# Changelog

* Update the reference documentation for any method that takes a deck slot as a parameter, or returns a deck slot.
* Add a new top-level "Deck Slots" page. This is a central place to introduce the two formats, and to describe their interchangeability. The reference documentation links to it.

And an unrelated thing that I'm squeezing into this PR:

* Make sure the `OFF_DECK` constant will get documented and cross-referenced properly in the future, when we publicize `move_labware()`. This takes a bit of Sphinx trickery.

# Review requests

Preview on http://sandbox.docs.opentrons.com/deck_label_docs/v2/.

* This is intended to be a full, complete, comprehensive update of every place in the Python Protocol API that deals with deck slots. I worked off of @jbleon95's [list](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3708256298/Audit+of+deck+slots+in+the+Python+Protocol+API+and+Protocol+Engine). Please let me know if it looks like we missed any place.
* Do we agree with adding the new "Deck Slots" page? Do we agree with its position in the table of contents?

# Risk assessment

Low.


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -777,19 +791,23 @@ def location_cache(self, loc: Optional[Location]) -> None:
     @property  # type: ignore
     @requires_version(2, 0)
     def deck(self) -> Deck:
```
SyntaxColoring, line 811: One thing I'm omitting here is that if you iterate over the keys, like `for deck_slot_name in protocol.deck`, the format will change depending on which robot the protocol is for. I couldn't figure out how to describe that concisely. Any suggestions?

SyntaxColoring: It sort of feels like we're getting to the point where the `Deck` class and its methods deserve their own reference documentation, instead of trying to stuff it all into the docstring of `ProtocolContext.deck`.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -777,19 +791,23 @@ def location_cache(self, loc: Optional[Location]) -> None:
...
 
         This object is useful for determining if a slot in the deck is free.
+
         Rather than filtering the objects in the deck map yourself,
         you can also use :py:attr:`loaded_labwares` to see a dict of labwares
         and :py:attr:`loaded_modules` to see a dict of modules.
 
-        For advanced control you can delete an item of labware from the deck
+        For advanced control, you can delete an item of labware from the deck
         with e.g. ``del protocol.deck['1']`` to free a slot for new labware.
```
sanni-t, line 829: While we're at it, I think we should mention that deck item deletion is not supported from 2.14 onwards. 

If I tested this correctly, on API v2.14+, deleting deck items results in TypeError saying that deck item deletion is not supported. I have not gotten to the bottom of why we get this typeerror, but even without the type error, deck item deletion doesn't work with engine backed API.

sanni-t (approved):
LGTM! Just the one comment about deck item deletion.
And I agree about making a separate deck documentation

api/docs/v2/deck_slots.rst
```diff
@@ -0,0 +1,74 @@
...
+
+.. _deck-slots:
+
+##########
+Deck Slots
+##########
+
+When you load an item onto the robot's deck, like with :py:obj:`ProtocolContext.load_labware()` or :py:obj:`ProtocolContext.load_module()`, you need to specify which slot to put it in.
+
+When doing that, you can choose between two formats:
```
ecormany:
```suggestion
Specify a slot in one of two formats:
```

api/docs/v2/deck_slots.rst
```diff
@@ -0,0 +1,74 @@
...
+
+* A letter-number pair like ``"A1"``. This matches how the deck is physically labeled on an Opentrons Flex.
+* A number like ``"10"`` or ``10``. This matches how the deck is physically labeled on an Opentrons OT-2.
```
ecormany: Maybe easier to parse as a table, like:
```suggestion
+----------------+-------------+----------------+
| Robot          | Slot format | Example        |
+================+=============+================+
| Opentrons Flex | Numeric     | `"A1"`         |
+----------------+-------------+----------------+
| Opentrons OT-2 | Coordinate  | `"10"` or `10` |
+----------------+-------------+----------------+
```

api/docs/v2/deck_slots.rst
```diff
@@ -0,0 +1,74 @@
...
+
+
+Opentrons Flex Deck Layout
+==========================
+
+..
+    TODO(mm, 2023-06-05): Embed a nice SVG instead of having these tables.
```
ecormany, line 26: Should have one of these very soon. Can then remove the large headers and put some transition text between the table and the image.

api/docs/v2/deck_slots.rst
```diff
@@ -0,0 +1,74 @@
...
+
+.. [#ft] Slot 12 has the fixed trash.
+
+
+Equivalent Slots
+================
+
+The Flex and OT-2 formats are interchangeable. You can use either format, regardless of which robot your protocol is for.
+
+For example, these are equivalent:
```
ecormany, line 66: We might want to be clear on whether you can mix and match within protocols, or whether you have to pick one system and stick to it (even if it doesn't match your requirements block). The fact that these are put in separate snippets is perhaps too subtle. If they were in the same protocol, would analysis raise an error on the second one, since the slot is already occupied?

api/docs/v2/index.rst
```diff
@@ -11,6 +11,7 @@ Welcome
     versioning
     new_labware
     new_modules
+    deck_slots
```
ecormany, line 14: I think this is fine for :shipit: purposes — these changes won't be deployed for a bit — but I'm still thinking about how the TOC and information architecture will look in a couple months… 🤔

api/docs/v2/new_protocol_api.rst
```diff
@@ -71,6 +74,10 @@ Useful Types and Definitions
 .. automodule:: opentrons.types
    :members: PipetteNotAttachedError, Point, Location, Mount
 
+..
+   TODO(mm, 2023-06-06): This should be uncommented when move_labware() is published. See above.
+   .. autodata:: opentrons.protocol_api.OFF_DECK
+      :no-value:
```
ecormany, line 82: 🎟️ RTC-288

api/src/opentrons/protocol_api/_types.py
```diff
@@ -8,3 +8,11 @@ class OffDeckType(enum.Enum):
 
 
 OFF_DECK: Final = OffDeckType.OFF_DECK
+
+# Set __doc__ manually as a workaround. When this docstring is written the normal way, right after
+# the constant definition, Sphinx has trouble picking it up.
+OFF_DECK.__doc__ = """\
+A special placeholder value for a labware's location, indicating that it's not currently on the robot's deck.
```
ecormany:
```suggestion
A special location value, indicating that a labware is not currently on the robot's deck.
```
"placeholder" has a connotation of being _default_, which this definitely isn't.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -342,11 +342,16 @@ def uri(self) -> str:
...
-        """The parent of this labware.
+        """The parent of this labware---where this labware is loaded.
 
-        If the labware is on the deck, a `str` deck slot name will be returned.
-        If on a module, the parent :py:class:`ModuleContext` will be returned.
-        If off deck, `None` will be returned.
+        Returns:
+            If the labware is directly on the robot's deck, the `str` name of the deck slot.
+            On a Flex, this will be like ``"D1"``, and on an OT-2, this will be like ``"1"``.
+            See :ref:`deck-slots`.
```
ecormany: With the addition of the new page, I'd be cool with:
```suggestion
            If the labware is directly on the robot's deck, the `str` name of the :ref:`deck slot <deck-slots>`.
```

And similarly below, link rather than spelling out in docstrings.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -500,15 +506,21 @@ def load_module(
...
-                         in Slot 7.
+
+        :param location: The location of the module.
+
+            This is usually the name or number of the slot on the deck where you
+            will be placing the module, like ``1``, ``"1"``, or ``"D1"``. See :ref:`deck-slots`.
+
+            Some modules, like the Thermocycler, are only valid in one deck location.
+            You do not have to specify a location when loading a Thermocycler---it will always
+            be in Slot 7.
```
ecormany: Is it always treated as B1 on Flex then? Maybe we can just eliminate the en dash and everything after and _not worry about it_. That's what we want users to do!

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -500,15 +506,21 @@ def load_module(
...
+
+            .. versionchanged:: 2.15
+                You can now specify a deck slot as a letter-number pair, like ``"D1"``.
```
ecormany:
```suggestion
                You can now specify a deck slot with coordinates, like ``"D1"``.
```
And do we need a similar notice on `load_labware`?

api/src/opentrons/protocol_api/protocol_context.py
ecormany: This would be great for making the new Deck Slots page less stubby. Can be another commit to this PR or as a follow-up.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -342,11 +342,16 @@ def uri(self) -> str:
...
+
+            If the labware is on a module, a :py:class:`ModuleContext`.
```
ecormany, line 352: NB: this makes nice formatting even though there's no referenceable thing that is a `ModuleContext` (each type of module has its own context class).

ecormany (approved):
This is a great start to documenting this set of features — thank you, Max! I left many comments, some of which we'll probably want to act upon before merging, but no big pressure on any of them. All of this will be rolled up into a much larger public deployment describing Flex features later this summer. So if you're feeling [bold](https://en.wikipedia.org/wiki/Wikipedia:Be_bold), go ahead and get this into `edge`.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: I think the best advice would be to direct users to `move_labware` off deck in higher API versions. (That also should work in 2.14 unless feature flags have changed recently…)

SyntaxColoring: Starting a discussion in `#proj-python-protocol-api`. It seems like we'd want to make `del deck[foo]` an alias of `protocol.moveLabware(foo, OFF_DECK)`?

api/src/opentrons/protocol_api/_types.py
SyntaxColoring: Good note, thanks.. Just changing "the labware" to "a labware."

api/src/opentrons/protocol_api/labware.py
SyntaxColoring: I think I'd prefer to keep these as-is because, at the time of writing, we're not consistent with *returned* deck slots, so it doesn't seem right to describe "general behavior" somewhere and then link to that description from everywhere.

`Protocol.loaded_modules` and `Protocol.loaded_labwares` return integers even on Flexen, but these methods return coordinate strings. Since it's currently dependent on the specific method or property you're calling, it seems appropriate to describe in the method reference.

api/docs/v2/deck_slots.rst
SyntaxColoring: Leaving this out because I couldn't quite figure out how to phrase it, but I think it's a good idea and we should do it.

api/docs/v2/deck_slots.rst
SyntaxColoring: Leaving this out for now because I liked explicitly spelling out "this matches how the deck is physically labeled on an Opentrons {Flex,OT-2}". I thought it would be helpful context if you've only ever seen one of these robots, and you're wondering WTF I'm talking about with these \~two styles of deck slot names\~.

Happy for this to be rewritten later if anyone disagrees, though.

api/src/opentrons/protocol_api/protocol_context.py
SyntaxColoring: Yes, thank you.

api/src/opentrons/protocol_api/protocol_context.py
SyntaxColoring: It is treated as B1 on the Flex, yeah. Updated to mention this.

SyntaxColoring:
Failing tests appear to be unrelated, from `tests/opentrons/hardware_control/integration/test_thermocycler.py` flakiness.

PR #12877, opened 2023-06-08
fix(robot-server): pulsing white status bar for stop requested engine state

b-cooper:
# Overview

While a protocol is being canceled, Design has updated the status bar behavior to transition to a pulsing white light until resolving a stop action. 

Also the protocol engine's Finishing state (performing final homing) will map to the pulsing white, instead of static blue as currently implemented.

A cancelled protocol will now have a static white status bar instead of static blue (which is the running state)


fsinapi (approved):
Yeah sure, LGTM if it's tested and works.

sfoster1 (approved):
yay

PR #12879, opened 2023-06-08
chore(robot-server): fix timing-out test

sfoster1:
The ot2-only tests were running an ot3 test and that simply won't do. The right fix for this is more in depth but this will fix it quickly.


vegano1 (approved):

SyntaxColoring (approved):
Thanks for getting to the bottom of this.

PR #12881, opened 2023-06-08
fix(api): revert firmware binaries dir change to use constant.

vegano1:
This [pr](https://github.com/Opentrons/opentrons/pull/12866) broke the unit tests, let's revert that change and just have the `ROBOT_FIRMWARE_DIR` constant get set to `/usr/lib/firmware` for both ot2 and Flex devices.

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Sorry for the bad advice on that PR! Good if the tests all pass.

PR #12886, opened 2023-06-09
feat(api): check if a tip is present on a pipette in the hardware controller

Laura-Danielle:
# Overview
We have tip presence ability on pipettes and so we should use it to inform the user of an error if a tip fails to pick up. This is the very beginning of tip presence handling and so it does **not**:

- continuously check to see if the tip state has changed while running a protocol
- re-try picking up or dropping a tip (although we can do that in a follow-up PR once we figure out a number here)
- Support tip sensing for the 96 channel

# Test Plan

Create a protocol that picks up and drops tip a few times. On the robot, you should have a tiprack that is missing some tips so that when a pick up tip fails to detect the tip and error raises. This will stop the protocol. You can do the same thing for drop tip by simply removing a tip after it has successfully picked up. I would suggest adding pauses between pick up/drop tip to make this easier to test.

- [ ] Check that pick up tip will fail in a protocol if it does not pick up a tip
- [ ] Check that drop tip will fail in a protocol if it does not drop an existing tip (we will have to fake this by removing the tip beforehand)
- [ ] Check that LPC can handle a failed pick up or drop gracefully
- [ ] Once we have done the above, run a regular (short) protocol to check the reliability
- [ ] Test all in-use DVT pipettes to see if they have incorrect photointerrupter (definitely on single channels, maybe on multi-channels)

# Changelog

- Add a command you can use in the hardware controller to query the tip state
- Raises an error if the state does not match the expected value when you're picking up or dropping tip.

# Review requests

Might've flipped the state. Please check me on this.

# Risk assessment

Medium/high. Could potentially be disruptive to ABR work.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -95,6 +95,8 @@ from .types import (
     UpdateStatus,
     StatusBarState,
     SubSystemState,
+    TipStateType,
+
 )
 from .errors import (
     MustHomeError,
@@ -1658,6 +1660,9 @@ class OT3API(
         for rel_point, speed in spec.shake_off_list:
             await self.move_rel(realmount, rel_point, speed=speed)
 
+        # TODO change to enum
+        await self._backend.get_tip_present(realmount, TipStateType.PICK_UP)
+
         _add_tip_to_instrs()
 
         if prep_after:
@@ -1735,6 +1740,9 @@ class OT3API(
                 for axis, current in spec.ending_current.items()
             }
         )
+
+        # TODO change to enum
+        await self._backend.get_tip_present(realmount, TipStateType.DROP)
         _remove()
 
     async def clean_up(self) -> None:
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1659,7 +1659,6 @@ class OT3API(
         for rel_point, speed in spec.shake_off_list:
             await self.move_rel(realmount, rel_point, speed=speed)
 
-        # TODO change to enum
         await self._backend.get_tip_present(realmount, TipStateType.PRESENT)
 
         _add_tip_to_instrs()
@@ -1740,7 +1739,6 @@ class OT3API(
             }
         )
 
-        # TODO change to enum
         await self._backend.get_tip_present(realmount, TipStateType.ABSENT)
         _remove()
 
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -96,7 +96,6 @@ from .types import (
     StatusBarState,
     SubSystemState,
     TipStateType,
-
 )
 from .errors import (
     MustHomeError,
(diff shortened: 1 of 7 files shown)


api/src/opentrons/hardware_control/types.py
```diff
@@ -636,6 +636,14 @@ def to_type(cls, gp: "GripperProbe") -> InstrumentProbeType:
             return InstrumentProbeType.PRIMARY
 
 
+class TipStateType(enum.Enum):
+    DROP = False
```
sfoster1: kind of a nit but the name is `State` and the elements are actions. Can we make these `PRESENT` and `ABSENT` or somethign

sfoster1 (approved):
Looks good though I'd love to see the enum entries change names

api/src/opentrons/hardware_control/types.py
Laura-Danielle: yeah actually agree. Will make that change.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1651,6 +1652,9 @@ async def pick_up_tip(
         for rel_point, speed in spec.shake_off_list:
             await self.move_rel(realmount, rel_point, speed=speed)
 
+        # TODO change to enum
```
sfoster1: not a todo anymore right?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1651,6 +1652,9 @@ async def pick_up_tip(
...
+        await self._backend.get_tip_present(realmount, TipStateType.PRESENT)
```
sfoster1, line 1662: this code seems to be written as if `self._backend.get_tip_present()` will raise if the tip state doesn't match, but it now returns a bool instead, so the call doesn't do anything

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1726,6 +1730,9 @@ async def drop_tip(
                 for axis, current in spec.ending_current.items()
             }
         )
+
+        # TODO change to enum
```
sfoster1: ditto both re: todo and re: exception/bool return

api/src/opentrons/hardware_control/types.py
```diff
@@ -618,6 +618,14 @@ def to_type(cls, gp: "GripperProbe") -> InstrumentProbeType:
             return InstrumentProbeType.PRIMARY
 
 
+class TipStateType(enum.Enum):
+    ABSENT = False
```
sfoster1: something about this makes me feel gross... maybe just use `enum.auto()`?

sfoster1 (changes requested):
Looks like the changes to make the tip state query return a bool don't match the way the function is called anymore so the checks don't do much

api/src/opentrons/hardware_control/ot3api.py
sfoster1: no wait it does raise

sfoster1:
- [x] Check that pick up tip will fail in a protocol if it does not pick up a tip
- [x] Check that drop tip will fail in a protocol if it does not drop an existing tip (we will have to fake this by removing the tip beforehand)
- [x] Check that LPC can handle a failed pick up or drop gracefully

sfoster1:
- [x]  Once we have done the above, run a regular (short) protocol to check the reliability

sfoster1 (approved):
Looks good to me!

sfoster1:
We'll rework the remaining pipettes with the misplaced photointerrupter flag.

sfoster1 (approved):
Looks good to me, I think we're all set!

PR #12887, opened 2023-06-09
feat(api, robot-server): add setting to disable the status bar

fsinapi:
# Overview

Adds a feature flag to the robot to disable the status bar. If the setting is enabled, the status bar is always going to be off. The status bar status cache is still updated so that, if the setting is disabled (to reenable the status bar), it will be set to the correct animation for the current robot context.

# Test Plan

Uploaded to an OT-3 and hit the `POST /settings` endpoint to turn the status bar off and on. It turned off and on, and the value persists when you restart the robot server.

# Changelog

- Add feature flag to disable the status bar
- Added API function to enable/disable the status bar
- Added handling in the Status Bar controller to logically handle being enabled/disabled
- In the `POST /settings` endpoint, always try to update the status bar enabled/disabled setting in case it changed

# Review requests

# Risk assessment

Pretty low.

### DIFF
--- a/api/src/opentrons/__init__.py
+++ b/api/src/opentrons/__init__.py
@@ -117,6 +117,7 @@ async def _create_thread_manager() -> ThreadManagedHardware:
             OT3API.build_hardware_controller,
             use_usb_bus=ff.rear_panel_integration(),
             threadmanager_nonblocking=True,
+            status_bar_enabled=ff.status_bar_enabled(),
         )
     else:
         thread_manager = ThreadManager(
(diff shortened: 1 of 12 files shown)


vegano1:
In which case would we need to disable the status bar?

fsinapi:
> In which case would we need to disable the status bar?

This was specifically requested by the design team. The app is going to need to be able to set this over HTTP.

sfoster1 (approved):
Looks good to me!

vegano1:
> > In which case would we need to disable the status bar?
> 
> This was specifically requested by the design team. The app is going to need to be able to set this over HTTP.

fair point, sometimes you want to disable led lights.

vegano1 (approved):
very nice, lgtm

shlokamin (approved):
🐎 thanks @fsinapi !

PR #12889, opened 2023-06-09
refactor(api,robot-server): change module calibration slot usage from Int to string.

vegano1:
# Overview

The module calibration feature was added before the new deck slot coordinate system was implemented, which meant that we were using the old int-style slots. To stay consistent, this PR switches the module calibration slot arg from an int to a string wherever it is used.

# Test Plan

- [x] Make sure we can load/save module calibrations
- [x] Make sure we can run the Module calibration test script using the new slot name
- [x] Make sure we report the correct slot name when reporting module offsets in the `/modules` endpoint

# Changelog

- Change Slot usage for module calibrations from int to a string

# Review requests

# Risk assessment
Low, flex only


SyntaxColoring (approved):
Sweet, thanks. Looks good to me if it works.

If you want, you can use `opentrons.types.DeckSlotName` instead of `str` for the `BaseModel`s, to be more self-documenting.

PR #12892, opened 2023-06-09
fix(shared-data, app): update the slot name for Flex

koji:
# Overview
update the slot name 7+10 to A1+B1

close RAUT-473 and RQA-697

Screenshot 2023-06-09 at 5 15 37 PM

# Test Plan

# Changelog
- update the const in shared-data  7+10 -> A1+B1

# Review requests

# Risk assessment
low


vegano1 (approved):
lgtm

PR #12894, opened 2023-06-12
refactor(robot-server): use the server.log file to store logs coming in from the robot server

Laura-Danielle:
## Overview

Looking at the api logs can get confusing when you have a bunch of requests from the robot server. Let's utilize the file that we already download from the app called 'server.log'

# Test Plan

- [ ] Throw this on a robot and run a protocol or two. Verify that the server.log file is populated with robot server requests.

# Review requests

Let me know if you think anything is missing.

# Risk assessment

Should be fairly low.

### DIFF
--- a/robot-server/robot_server/service/legacy/routers/logs.py
+++ b/robot-server/robot_server/service/legacy/routers/logs.py
@@ -1,16 +1,16 @@
 from fastapi import APIRouter, Query, Response
-from typing import Dict, List
+from typing import Dict
 
 from opentrons.system import log_control
 from robot_server.service.legacy.models.logs import LogIdentifier, LogFormat
 
 router = APIRouter()
 
-IDENTIFIER_TO_SYSLOG_ID: Dict[LogIdentifier, List[str]] = {
-    LogIdentifier.api: ["opentrons-api"],
-    LogIdentifier.serial: ["opentrons-api-serial"],
-    LogIdentifier.server: ["uvicorn"],
-    LogIdentifier.api_server: ["uvicorn", "opentrons-api"],
+IDENTIFIER_TO_SYSLOG_ID: Dict[LogIdentifier, str] = {
+    LogIdentifier.api: "opentrons-api",
+    LogIdentifier.serial: "opentrons-api-serial",
+    LogIdentifier.server: "uvicorn",
+    LogIdentifier.api_server: "opentrons-robot-server",
 }
 
 
--- a/robot-server/robot_server/service/legacy/routers/logs.py
+++ b/robot-server/robot_server/service/legacy/routers/logs.py
@@ -1,15 +1,16 @@
 from fastapi import APIRouter, Query, Response
-from typing import Dict
+from typing import Dict, List
 
 from opentrons.system import log_control
 from robot_server.service.legacy.models.logs import LogIdentifier, LogFormat
 
 router = APIRouter()
 
-IDENTIFIER_TO_SYSLOG_ID: Dict[LogIdentifier, str] = {
-    LogIdentifier.api: "opentrons-api",
-    LogIdentifier.serial: "opentrons-api-serial",
-    LogIdentifier.server: "uvicorn",
+IDENTIFIER_TO_SYSLOG_ID: Dict[LogIdentifier, List[str]] = {
+    LogIdentifier.api: ["opentrons-api"],
+    LogIdentifier.serial: ["opentrons-api-serial"],
+    LogIdentifier.server: ["uvicorn"],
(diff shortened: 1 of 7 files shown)


sfoster1 (changes requested):
This is a good idea but it's not how the logs work. `GET /logs/server.log` doesn't retrieve a file anymore; since the buildroot migration it [looks up the syslog identifier associated with the file and returns those logs](https://github.com/Opentrons/opentrons/blob/edge/robot-server/robot_server/service/legacy/routers/logs.py#L16). What I think will accomplish this is
1. add a new log config that sends the uvicorn access logs to a different syslog identifer (or just `uvicorn`
2. map that to the `server.log` (if you do `uvicorn` as the syslog identifier it'll Just Work (tm) since that's what's mapped already)
3. Get those logs out of the `opentrons-api` syslog identifier, BUT imo we should keep it in the `opentrons-robot-server` journald logs since sometimes it's actually quite nice to be able to see it all unified

sfoster1 (changes requested):
Let's take a step back and talk about how the logging is organized here, because this commit isn't quite what we want either.

So the top line goals of this PR are
1. Prevent uvicorn access logs from going to the `opentrons-api` syslog identifier, which `GET /logs/api.log` retrieves and are accessible locally via `journalctl -t opentrons-api`
2. Make uvicorn access logs go to the `uvicorn` syslog identifier, which `GET /logs/server.log` retrieves and are accessible locally via `journalctl -t uvicorn`
3. Keep everything going to the `opentrons-robot-server` unit logs, which no API retrieves but which are accessible locally via `journalctl -u opentrons-robot-server` (and are the moral equivalent of just `print`int stuff to stdout)

What the latest commit does is
- Keep everything involved going to `opentrons-robot-server` unit logs, because that happens with all `systemd.journal.JournalHandler` logs; good! This does (3)
- Adds a new syslog identifier `opentrons-robot-server` via the new `SYSLOG_IDENTIFIER` entries in the handler configs. This is unrelated to the `opentrons-robot-server` unit logs; this would not be gathered by an api and would be accessible locally via `journalctl -t opentrons-robot-server`. I don't think there's a good reason to do this, but it doesn't hurt anything really.
- Switches two handlers, `syslog_plus_unit` and `syslog_plus_unit_above_warn`, from logging to (a) unit logs and (b) `opentrons-api` syslog id to log to (i) unit logs and (ii) `opentrons-robot-server` syslog id. 

I think there's two problems with this. The first is that nothing is getting mapped to the `uvicorn` syslog id, which is what `GET /logs/server.log` returns; the `SYSLOG_IDENTIFIER` entries in the handlers need to be `uvicorn` for that, not `opentrons-robot-server`. Second, I'm not sure I like putting _all_ robot server logs only over there; our logs only have second-resolution timestamps, so this would make it pretty tough to line up api logs and server logs from a robot export. Putting the uvicorn access and error logs there makes sense, I suppose. Maybe we want to be able to get the full unit logs from some new logging api.

So I think let's
- Make a new error handler that's called `server_syslog_plus_unit` that maps to syslog identifier `uvicorn` and put the `uvicorn` logs there, and maybe the starlette logs too? I'm honestly not super sure where the dang access logs are coming from
- Maybe add that handler to the other server logs too but don't remove the normal `syslog_plus_unit`; this would make uvicorn errors and robot_server logs show up in _both_ `opentrons-api`/`GET /logs/api.log` _and_ `uvicorn`/`GET /logs/server.log`, which doesn't cut down on wordiness but does preserve the usefulness of the logging

Either way, definitely gotta get rid of the `opentrons-robot-server` syslog identifier.

robot-server/robot_server/service/legacy/routers/logs.py
```diff
@@ -28,7 +29,7 @@ async def get_logs(
     syslog_id = IDENTIFIER_TO_SYSLOG_ID[log_identifier]
     modes = {
         LogFormat.json: ("json", "application/json"),
-        LogFormat.text: ("short", "text/plain"),
+        LogFormat.text: ("short-precise", "text/plain"),
```
sfoster1, line 32: rather than doing this, let's add a way to get a unit log instead of a syslog identifier - so we can be doing `journalctl -u opentrons-robot-server` rather than `journalctl -t opentrons-api -t uvicorn`

sfoster1 (changes requested):
Slight change to the way we implement the combined log but looks good as a whole

sfoster1 (approved):
Code lgtm but will test

sfoster1 (approved):
Tested, looks good to me!

PR #12897, opened 2023-06-12
feat: Add script for provisioning Flex pipettes in emulation

DerekMaggio:
# Overview

Add script for provisioning Flex pipettes in emulation. This script is run inside the `ot3-firmware-builder` container from opentrons-emulation

Script Flow: 

1. Read from environment variables: `LEFT_OT3_PIPETTE_DEFINITION` and `RIGHT_OT3_PIPETTE_DEFINITION`
2. Convert env var content to `OT3PipetteEnvVar` dataclass object
3. Generate serial code by using `opentrons_hardware.instruments.pipettes.serials.serial_val_from_parts`
4. Create eeprom files for each pipette inside of that pipette's volume folder. Note that the volume folders are created inside the emulation scope. 

----

**Note:**
It is out of scope for this PR to actually ensure that the pipettes run correctly in emulation. They currently do not, and I cannot figure out why. Going to circle up with the wizards in embedded land to see if they can help me figure it out. A followup PR will address them actually working

----

# Test Plan

- [x] Add unit test to validate that exercising the above flow creates the file
- [x] Validate against emulation repo
  - [x] Using opentrons-emulation local dev, bind mount the latest version of this branch into the emulation system. 
  - [x] Validate that framework automatically creates correct env vars for script to parse
  - [x] Validate that script is automatically called with no required user inpur
  - [x] Validate that eeprom files are created on ot3-firmware-builder
  - [x] Validate that eeprom files are pushed to the pipette containers through the volume connection

# Changelog

- Add script
- Add test

# Review requests

Do you think it is okay to not validate that the content of the eeprom file is correct? 
I am not doing this for a few reasons: 

1. Figuring out what the content is actually supposed to be sounds like a pain
2. I am using a monorepo utility function to do the actual serial number provisioning. This functionality should be tested in the monorepo and not in the context of emulation
3. The new functionality is parsing the env vars and pushing to the file. Which testing was added for 

# Risk assessment

Low 

- Script is not used by this repo
- Tests do not affect this repo
- No dependencies or existing source code were modified

### DIFF
--- a/hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
+++ b/hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
@@ -1,0 +1,99 @@
+"""Script to write pipette serial numbers to eeprom files.
+
+This script is configured to run against the emulation container.
+High level overview of the script:
+
+- Read pipette definition from environment variables
+- Generate pipette serial code
+- Write pipette serial code to eeprom file
+- Profit
+"""
+
+import os
+import json
+from dataclasses import dataclass
+from typing import Optional
+from typing_extensions import Literal
+from opentrons_hardware.firmware_bindings.constants import PipetteName
+from opentrons_hardware.instruments.pipettes.serials import serial_val_from_parts
+
+LEFT_PIPETTE_ENV_VAR_NAME = "LEFT_OT3_PIPETTE_DEFINITION"
+RIGHT_PIPETTE_ENV_VAR_NAME = "RIGHT_OT3_PIPETTE_DEFINITION"
+
+LEFT_PIPETTE_EEPROM_DIR_PATH = "/volumes/left-pipette-eeprom"
+RIGHT_PIPETTE_EEPROM_DIR_PATH = "/volumes/right-pipette-eeprom"
+
+
+@dataclass
+class OT3PipetteEnvVar:
+    """OT3 Pipette Environment Variable."""
+
+    pipette_name: PipetteName
+    pipette_model: int
+    pipette_serial_code: bytes
+    eeprom_file_name: str
+    mount: Literal["left", "right"]
+
+    @classmethod
+    def from_json_string(
+        cls, env_var_string: str, mount: Literal["left", "right"]
+    ) -> "OT3PipetteEnvVar":
+        """Create OT3PipetteEnvVar from json string."""
+        env_var = json.loads(env_var_string)
+        return cls(
+            pipette_name=PipetteName[env_var["pipette_name"]],
+            pipette_model=env_var["pipette_model"],
+            pipette_serial_code=env_var["pipette_serial_code"].encode("utf-8"),
+            eeprom_file_name=env_var["eeprom_file_name"],
+            mount=mount,
+        )
+
+    def _eeprom_dir_path(self) -> str:
+        """Get eeprom directory path."""
+        return (
+            LEFT_PIPETTE_EEPROM_DIR_PATH
+            if self.mount == "left"
+            else RIGHT_PIPETTE_EEPROM_DIR_PATH
+        )
+
+    @property
+    def eeprom_file_path(self) -> str:
+        """Get eeprom file path."""
+        return "/".join([self._eeprom_dir_path(), self.eeprom_file_name])
+
+    def _generate_serial_code(self) -> bytes:
+        """Generate serial code."""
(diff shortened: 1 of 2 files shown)


hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
```diff
@@ -0,0 +1,99 @@
...
+from dataclasses import dataclass
+from typing import Optional
+from typing_extensions import Literal
+from opentrons_hardware.firmware_bindings.constants import PipetteName
+from opentrons_hardware.instruments.pipettes.serials import serial_val_from_parts
+
+LEFT_PIPETTE_ENV_VAR_NAME = "LEFT_OT3_PIPETTE_DEFINITION"
+RIGHT_PIPETTE_ENV_VAR_NAME = "RIGHT_OT3_PIPETTE_DEFINITION"
+
+LEFT_PIPETTE_EEPROM_DIR_PATH = "/volumes/left-pipette-eeprom"
```
fsinapi: Part of me thinks that these should be parameters instead of hard coded, but this is such a specialized script that I don't think it matters.

hardware/tests/test_scripts/test_emulation_pipette_provision.py
```diff
@@ -0,0 +1,67 @@
...
+def test_main(
+    set_env_vars: Generator[None, None, None],
+    left_eeprom_path: str,
+    right_eeprom_path: str,
+) -> None:
+    """Test main."""
+    assert not os.path.exists(left_eeprom_path)
+    assert not os.path.exists(right_eeprom_path)
+    emulation_pipette_provision.main()
+    assert os.path.exists(left_eeprom_path)
```
fsinapi, line 78: It might be worth it to verify that the contents of these files is correct by calling `serial_val_from_parts` and reading the files. Unfortunate that we can't deserialize the encoded serial values though!

fsinapi (approved):
Yeah this looks correct to me, couple of comments you can ignore 👍

hardware/tests/test_scripts/test_emulation_pipette_provision.py
DerekMaggio: Just to clarify, I am calling that function under the hood to generate the serial number. 
Is it redundant to call the same function to compare that value to?

hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
DerekMaggio: I can easily make these part of the parsed environment variable. 
Would that suffice?

hardware/tests/test_scripts/test_emulation_pipette_provision.py
fsinapi: I think what makes it worth testing is that you want to verify that the script, viewed as a black box, is generating the right contents in the right files. I agree that it isn't the most thorough way to test it, but it at least makes sure you're 1) getting all of the info from the environment variable 2) putting the left/right info in the correct respective files.

hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
fsinapi: 💯  sounds good to me

Laura-Danielle (approved):
This looks good to me.

hardware/tests/test_scripts/test_emulation_pipette_provision.py
DerekMaggio: Well good thing you suggested this. 
I, being an idiot, was only ever evaluating the left pipette and storing it to both left and right. 
Thanks @fsinapi. You saved me from myself

DerekMaggio: https://github.com/Opentrons/opentrons/pull/12897/commits/73ea94dbadec3c163085630d28d382b4462a4673

hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
DerekMaggio: https://github.com/Opentrons/opentrons/pull/12897/commits/73ea94dbadec3c163085630d28d382b4462a4673

fsinapi (approved):
LGTM !

PR #12898, opened 2023-06-12
feat(api): add errorCode field to ErrorOccurrence model

fsinapi:
# Overview

Adds a field `errorCode` to the `ErrorOccurrence` model as a part of the OT-3 specific error messaging. The error code originates from the exception class that was raised to indicate an error, and in future work the exceptions for protocol engine errors will be configured with their own hard coded error codes. For now, everything gets the catch-all code 4000.

We want this new field to be required in all responses, but we _also_ need to be able to deserialize saved run reports on the robot. If the field is defined without a default, the robot server cannot load old runs; if the field is defined _with_ a default, the schema will imply that it is optional, and the client would have to account for cases where it is missing! To solve this, we define a default value and manually append the `errorCode` field to the list of required fields.

# Test Plan

Run a protocol that results in an error (or cancel the run). Download the run log after the run is cancelled/fails, and verify that there is an error code 4000 in the `ErrorOccurrence`.

To test backwards compatibility, perform the above on a robot _without_ this patch. Upload this patch and view the protocol run record; there should be an error code 4000 even though it wasn't included in the saved run report.

I tested this with a dev server using a persistent server, switching from `edge` to this branch. I was able to download the run fine _as long as there was a default defined for errorCode_.

Tested pushing this patch on top of an OT-3 at 0.11.0 and it was able to load run records fine as well.

# Changelog

# Review requests

I defined the type of the field as a `str` rather than the actual Enum for flexibility, but is it better to be more specific and type the field off of the enum?

Also please confirm that I'm right about the risk assessment :) 

# Risk assessment

The biggest risk should be backwards compatibility. This _shouldn't_ be an issue because:
- When loading old run records, the default value for the `errorCode` will let it parse successfully
- If you downgrade from this to an older robot server version, the extra field should be ignored because that's how pydantic parsing works
- The app should be fine ignoring this new field for now

### DIFF
--- a/api/src/opentrons/protocol_engine/errors/exceptions.py
+++ b/api/src/opentrons/protocol_engine/errors/exceptions.py
@@ -1,11 +1,20 @@
 """Protocol engine exceptions."""
 
+from enum import IntEnum, unique
+
+
+@unique
+class ErrorCode(IntEnum):
+    """Enumerated error codes."""
+
+    UNKNOWN = 4000  # Catch-all code for any unclassified error
+
 
 class ProtocolEngineError(RuntimeError):
     """Base Protocol Engine error class."""
 
     # This default error code should be overridden in every child class.
-    ERROR_CODE = 4000
+    ERROR_CODE: int = ErrorCode.UNKNOWN
 
 
 class UnexpectedProtocolError(ProtocolEngineError):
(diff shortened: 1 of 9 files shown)


sfoster1 (approved):
This makes sense (haven't tested) and I support keeping the errors as strings, we'll be moving around and centralizing them and keeping them as a fundamental data type meshes well with that

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -1,9 +1,21 @@
 """Protocol engine exceptions."""
 
+from enum import Enum, unique
+
+
+@unique
```
SyntaxColoring, line 6: Neat, thanks for teaching me about `@unique`.

SyntaxColoring (approved):
Nice.

The code, risk analysis, and test plan all look great to me.

And I agree with leaving it as a `str` for now.

PR #12899, opened 2023-06-12
chore: small api log cleanup

ahiuchingau:
# Overview
This PR removes some duplicate logs and move some non-crucial info to the debug level. This will make debugging ABR issues a bit easier.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1063,6 +1063,7 @@ class OT3API(
         _, moves = self._move_manager.plan_motion(
             origin=origin, target_list=[move_target]
         )
+        self._log.info(f"built move: {moves[0].to_dict()}")
         return moves
 
     @ExecutionManagerProvider.wait_for_running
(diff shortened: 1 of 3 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -577,12 +577,12 @@ async def _update_position_estimation(
         """
         Function to update motor estimation for a set of axes
         """
-
         if axes:
             checked_axes = [OT3Axis.from_axis(ax) for ax in axes]
         else:
             checked_axes = [ax for ax in OT3Axis]
         await self._backend.update_motor_estimation(checked_axes)
+        self._log.info(f"Updated position estimation for {checked_axes}")
```
Laura-Danielle, line 585: can we make this debug

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1063,6 +1063,7 @@ def _build_moves(
         _, moves = self._move_manager.plan_motion(
             origin=origin, target_list=[move_target]
         )
+        self._log.info(f"built move: {moves[0].to_dict()}")
```
Laura-Danielle, line 1066: same with this

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1091,15 +1092,15 @@ async def _move(
         check_motion_bounds(to_check, target_position, bounds, check_bounds)
 
         origin = await self._backend.update_position()
+        self._log.info(
```
Laura-Danielle, line 1095: this also feels like a debug log

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1221,14 +1222,14 @@ async def home(
...
             checked_axes.append(OT3Axis.Q)
-        self._log.info(f"Homing {axes}")
 
         home_seq = [
             ax
             for ax in OT3Axis.home_order()
             if (ax in checked_axes and self._backend.axis_is_present(ax))
         ]
-        self._log.info(f"home was called with {axes} generating sequence {home_seq}")
+        self._log.info(f"Homing {home_seq}")
```
Laura-Danielle, line 1231: why this change?

Laura-Danielle (changes requested):
Most of these logs feel like debug status to me. I think by default all ABR robots should probably be on a debug log level. Maybe we can include that check for internal releases specifically?

api/src/opentrons/hardware_control/ot3api.py
ahiuchingau: See the deleted line in 1124. This just makes these two log info less repetitive.

Info: we're telling what axes we ended up homing
Debug: we're also saying what axes the func was called with

PR #12901, opened 2023-06-12
fix(hardware): move group timeout exception should be raised

ahiuchingau:
# Overview
This PR raises an exception and stops the move group runner when a move group has timed out. This means the active protocol will be stopped when a timeout occurs.


sfoster1 (approved):
This is good for now but we should make it a custom error soon.

fsinapi (approved):
This LGTM! I'll mention that this is going to cause a merge conflict with #12870 once the `internal-release_0.11.0` branch is ready to pull into edge, so we should just be ready to handle that :)

PR #12902, opened 2023-06-12
feat(app): add ErrorBoundary component

koji:
# Overview
add ErrorBoundary component to provide reload the app with users when the touchscreen app shows the white-screen
install `react-error-boundary` to use our own hooks since we have to create a class component for Error Boundary.
https://github.com/bvaughn/react-error-boundary

This is a temporary modal since we don't have the design for this function. 

Screenshot 2023-06-16 at 5 36 08 PM

FYI
Currently this PR doesn't pass most CI checks because of causing an error on the ODD app.
Once this gets approval, the error code will be removed.

# Test Plan
0. `make teardown-js && make setup-js` and push the app
1. Tap All protocols
2. Tap a Protocol
The modal shows up and tap the restart button

# Changelog
- add OnDeviceDisplayAppFallback component and test
- install react-error-boundary
- uninstall react-use since we are not using

# Review requests

# Risk assessment
low


koji:
- [ ] add the function to reload the app
- [ ] send error to logs

app/src/pages/OnDeviceDisplay/ProtocolDetails/index.tsx
```diff
@@ -68,6 +68,8 @@ const ProtocolHeader = (props: {
     displayedTitle = truncateString(title, 92, 69)
   }
 
+  fdfd
+
```
koji: This is to cause the error so once this PR gets approval, this line will be removed.
The CI gets errors for this.

app/package.json
```diff
@@ -42,14 +42,14 @@
...
     "react-dom": "17.0.1",
+    "react-error-boundary": "^4.0.10",
     "react-hot-loader": "4.13.0",
     "react-i18next": "^11.7.3",
     "react-intersection-observer": "^8.33.1",
     "react-redux": "7.2.1",
     "react-router-dom": "5.1.1",
     "react-select": "5.4.0",
     "react-simple-keyboard": "^3.4.187",
-    "react-use": "^17.4.0",
```
koji, line 53: Actually we don't use this.

app-shell-odd/src/ui.ts
```diff
@@ -69,7 +69,8 @@ export function createUi(dispatch: Dispatch): BrowserWindow {
   return mainWindow
 }
 
-function waitForRobotServerAndShowMainWIndow(dispatch: Dispatch): void {
+export function waitForRobotServerAndShowMainWIndow(dispatch: Dispatch): void {
+  console.log('waitForRobotServerAndShowMainWIndow')
```
shlokamin: leftover console log

app/package.json
```diff
@@ -42,14 +42,14 @@
     "path-to-regexp": "3.0.0",
     "react": "17.0.1",
     "react-dom": "17.0.1",
+    "react-error-boundary": "^4.0.10",
```
shlokamin, line 46: how come we need this package? error boundaries are supported out of the box with react: https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

we do this in LPC: https://github.com/Opentrons/opentrons/blob/8ae921cba2760f43efc60df8f70f616c1e05662b/app/src/organisms/LabwarePositionCheck/index.tsx#L47

app/src/App/OnDeviceDisplayAppFallback.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Flex
+        width="100%"
+        flexDirection={DIRECTION_COLUMN}
+        gridGap={SPACING.spacing32}
+        padding={SPACING.spacing32}
+        alignItems={ALIGN_CENTER}
+        justifyContent={JUSTIFY_CENTER}
+      >
+        <StyledText as="h4" fontWeight={TYPOGRAPHY.fontWeightBold}>
+          {'Something went wrong'}
```
shlokamin, line 47: can we also show what error occurred just like we do for the LPC error boundary? it will be helpful for users

app/package.json
koji: to use `useDispatch`

shlokamin: fair enough!

app/src/App/__tests__/OnDeviceDisplayAppFallback.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import { appRestart } from '../../redux/shell'
+import { OnDeviceDisplayAppFallback } from '../OnDeviceDisplayAppFallback'
+
+jest.mock('../../redux/shell')
+
+const mockAppRestart = appRestart as jest.MockedFunction<typeof appRestart>
+
+const render = () => {
+  return renderWithProviders(<OnDeviceDisplayAppFallback />)
```
brenthagen: need to fix this type error

brenthagen (approved):
this is a great addition - tested on dev kit with and without the protocol details error. may need more through testing, but i think this is good to merge once the fallback test is fixed.

PR #12903, opened 2023-06-12
fix(app): do not render maintenance modal when starting LPC from ODD

shlokamin:
# Overview

This PR properly sets the ODD maintenance run flag when starting LPC from the the ODD. It also deletes a component that was causing confusion and is no longer used.

Closes [RQA-951](https://opentrons.atlassian.net/browse/RQA-951)

# Test Plan

Verified on my dev kit that starting LPC from the ODD does not cause the takeover modal to show up
# Changelog

- Set maintenance run flag when starting LPC from the ODD

# Risk assessment

Low

[RQA-951]: https://opentrons.atlassian.net/browse/RQA-951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


b-cooper (approved):
☑️

PR #12905, opened 2023-06-13
feat(shared-data): add error code data

sfoster1:
This PR adds information in `shared-data` for consistent error codes across the stack. This will let us use and assign error codes consistently especially on the python side.

Some things to note:
- This doesn't add any actual used code; it just adds the data and accessors
- The typescript accessors are autogenerated from json, which is nice
- The python accessors are handcoded, which is because that way they can work more nicely with mypy. It's annoying, but we have consistency tests, so it's not that annoying

I can't decide whether to implement an exception hierarchy here or just do what we're doing now, which is a base exception that captures a code and lets individual modules derive their own hierarchies from it.

## Risks
- None, code isn't used

## Review requests
- Does this style of language binding work for you?
- What do you think about the shared-data exception hierarchy concept?


shared-data/js/__tests__/validateErrors.test.js
```diff
@@ -0,0 +1,34 @@
...
+        'errors.json does not follow its schema',
+        JSON.stringify(validationFailures, null, 4)
+      )
+    }
+    expect(valid).toBe(true)
+    expect(validationFailures).toBeNull()
+  })
+})
+
+describe('error code unschemad requirements should match', () =>
```
shlokamin, line 25: nit: this test name is confusing to me. youre just checking that each error has a category and that there is clean mapping from error code to error category code right?

shared-data/js/errors.ts
```diff
@@ -0,0 +1,35 @@
...
+  category: ErrorCategory
+  code: ErrorCodes
+}
+
+interface JSONErrorBody {
+  detail: string
+  category: ErrorCategories
+}
+
+export type ErrorDefinitions = Map<ErrorCodes, ErrorBody>
```
shlokamin: should this be a Map type? i think its just an object: 

```js
type ErrorDefinitions  = {
  [Code in ErrorCodes]: ErrorBody;
};
```

shlokamin (approved):
seems reasonable to me

shared-data/js/__tests__/validateErrors.test.js
sfoster1: yeah. which i really wish i could capture in the schema as like "this must be a reference to a real existing category" but can't as far as i'm aware. so these are unschema'd requirements

shared-data/js/errors.ts
sfoster1: Sure! I think i've noticed that in other parts of the ts code we prefer objects like this to maps but i can do a map pretty easily instead

sfoster1: I mean it _can_... i guess...

sfoster1: but really it's a homogenous map, there's one key type and one error type, and generalizes over data. to me that's a map

fsinapi (approved):
I like the Python bindings a lot, this LGTM.

I'm assuming that the method for associating error codes with exceptions will be to derive from `EnumeratedError` and just hard code the call to the super's initializer?

sfoster1:
> I like the Python bindings a lot, this LGTM.
> 
> I'm assuming that the method for associating error codes with exceptions will be to derive from `EnumeratedError` and just hard code the call to the super's initializer?

Pretty much, but what I'm trying to decide is if I should make an exception for each error code _now_ in `exceptions.py` or leave it for later. I'm kind of leaning towards now honestly

shared-data/js/errors.ts
SyntaxColoring: Yeah, as a general practice, isn't a `Map` data structure inherently safer than using an object as a map? To avoid modification of special object properties?

shared-data/errors/definitions/1/errors.json
```diff
@@ -0,0 +1,114 @@
...
+      "codePrefix": "3"
+    },
+    "generalError": {
+      "detail": "A software error occurred",
+      "codePrefix": "4"
+    }
+  },
+  "codes": {
+    "1000": {
+      "detail": "Communication Error",
```
SyntaxColoring, line 22: What's the intended use of these `detail` strings? Are they for consumption by the client, or the server?

@fsinapi and I were wondering how we'd represent dynamic information in errors. For example, the OT-2 currently tells you which particular axis had a hard limit error. @fsinapi thought the server would stuff that information in a `detail` field, which the client would display as-is.

I guess I'm worried about us getting confused when there are different `detail` sources competing with each other. But it seems like things would work out if:

1. The server bases its dynamic `detail`s on these static `detail`s, constructing strings like `"Communication Error: blah blah"`
2. The client always uses exactly what the server gives it, and never consults this list to do its own `details` lookup

Does that match what you have in mind?

sfoster1: @SyntaxColoring I think that this might be clearer now with the latest commit that adds exceptions, but I think there's a couple separate answers to the questions here.

> What's the intended use of these `detail` strings?

Basically nothing. I expect the app to do its own formatting and messaging; and I don't intend to send these `detail` from the server. These are here for the dual purpose of (1) making it possible to understand what the codes are for while looking at this file and (2) providing a fallback message for something the app forgot to implement or a default message for a naive client.

> wondering how we'd represent dynamic information in errors

I think the answer to this can't and shouldn't be contained in this file as long as the errors are simple dynamic fallbacks. Rather, that answer is properly expressed externally in the robot server API models for errors; and internally, in the data that the individual exception types hold. I think those should have a programmer written message that maybe is about what operation was happening, and a details dict about what axis something had, for instance. If we want to keep these consistent, we should have that detailed in the error models rather than here.

> getting confused when there are different detail sources competing with each other

definitely a possibility, but i think the rule is that "customized is better than general" if there's a conflict. messages in error returns or exceptions beat what's in the file; designed messages for specific errors beat the error returns. the server shouldn't use what's in this file either, just the message from the code.

PR #12909, opened 2023-06-13
feature(api): Allow load_labware to accept off-deck as a valid location

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/RLAB-344.
Allow `load_labware` to accept `off-deck` as a valid location. 

# Test Plan
Upload the following protocol with versions 2.15 and version 2.14.
Make sure you are able to upload the protocol with an off-deck location for 2.15.
Make sure this works of OT-2 and OT-3.
```
from typing import cast
from opentrons.protocol_api import OFF_DECK

metadata = {
    "protocolName": "Mad Block",
    "author": "Opentrons <engineering@opentrons.com>",
    "description": "A Magnetic Module Test",
    "source": "Opentrons Repository",
    "apiLevel": "2.15",
}

    tiprack_200_1       = protocol.load_labware('opentrons_ot3_96_tiprack_200ul', OFF_DECK)
    print(tiprack_200_1.parent)
```

Make sure you are not able to upload the protocol with an off-deck location for 2.14.
```
from typing import cast
from opentrons.protocol_api import OFF_DECK

metadata = {
    "protocolName": "Mad Block",
    "author": "Opentrons <engineering@opentrons.com>",
    "description": "A Magnetic Module Test",
    "source": "Opentrons Repository",
    "apiLevel": "2.14",
}

    tiprack_200_1       = protocol.load_labware('opentrons_ot3_96_tiprack_200ul', OFF_DECK)
    print(tiprack_200_1.parent)
```

uploading the following protocol should resolve in an error that you cannot access a off-deck item
```
from opentrons.protocol_api import OFF_DECK
metadata = {"apiLevel": "2.15"}


def run(ctx):
    ctx.home()

    tiprack1 = ctx.load_labware_by_name("opentrons_96_tiprack_300ul", "1")
    tiprack2 = ctx.load_labware_by_name("opentrons_96_tiprack_300ul", "2")
    tiprack3 = ctx.load_labware("opentrons_96_filtertiprack_20ul", "3")
    tiprack4 = ctx.load_labware("opentrons_96_tiprack_20ul", "4")

    plate1 = ctx.load_labware("corning_96_wellplate_360ul_flat", "7")
    plate2 = ctx.load_labware("corning_96_wellplate_360ul_flat", OFF_DECK)
    print(f"parent {plate2.parent}")
    pip = ctx.load_instrument(
        "p20_single_gen2", mount="left", tip_racks=[tiprack3, tiprack4]
    )
    pip2 = ctx.load_instrument(
        "p300_single_gen2", mount="right", tip_racks=[tiprack1, tiprack2]
    )
    pip.transfer(
        10,
        plate1.wells("A1"),
        [
            plate2[well].top()
            for well in ["A1", "B1", "C1", "D1", "E1", "F1", "G1", "H1"]
        ],
    )
    pip2.transfer(
        50,
        plate1.wells("A2"),
        [
            plate2[well].bottom()
            for well in ["A1", "B1", "C1", "D1", "E1", "F1", "G1", "H1"]
        ],
    )
    # mix behavior
    pip.transfer(
        50,
        plate1.wells("A3"),
        [
            plate2[well].top()
            for well in ["A1", "B1", "C1", "D1", "E1", "F1", "G1", "H1"]
        ],
    )
```

# Changelog

- Added logic for `load_labware` location arg to accept a OffDeckType gated to 2.15
- `get_labware_location` now defaulting to `off-deck` instead of None.
- ` Labware.parent` will return off-deck as one of the options.
- added OffDeckType to LabwareLike type options.
- added OffDeckType to `LocationLabware` type options.

# Review requests

Do you have any protocols I should test with to validate `LabwareLike` changes and `LocationLabware` changes? looking at the code there is no way this will resolve as `off-deck` from legacy bc we were not able to load a labware with an unknown location and move labware was not implemented before 2.15.
Does the raise exception make more sense?

# Risk assessment

Low. Should only effect versions 2.15 that involve move_labware and load_labware off-deck.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -133,17 +133,23 @@ class ProtocolCore(
     def load_labware(
         self,
         load_name: str,
-        location: Union[DeckSlotName, ModuleCore, NonConnectedModuleCore],
+        location: Union[DeckSlotName, ModuleCore, NonConnectedModuleCore, OffDeckType],
         label: Optional[str],
         namespace: Optional[str],
         version: Optional[int],
     ) -> LabwareCore:
         """Load a labware using its identifying parameters."""
-        module_location: Union[ModuleLocation, DeckSlotLocation]
+        load_location: Union[ModuleLocation, DeckSlotLocation, str]
         if isinstance(location, (ModuleCore, NonConnectedModuleCore)):
-            module_location = ModuleLocation(moduleId=location.module_id)
+            load_location = ModuleLocation(moduleId=location.module_id)
+        elif location == OffDeckType.OFF_DECK:
+            load_location = OFF_DECK_LOCATION
+        elif isinstance(location, DeckSlotName):
+            load_location = DeckSlotLocation(slotName=location)
         else:
-            module_location = DeckSlotLocation(slotName=location)
+            raise UnknownLocationError(
+                f"load_labware is not supported with the given location: {location}."
+            )
 
         custom_labware_params = (
             self._engine_client.state.labware.find_custom_labware_load_params()
@@ -154,7 +160,7 @@ class ProtocolCore(
 
         load_result = self._engine_client.load_labware(
             load_name=load_name,
-            location=module_location,
+            location=load_location,
             namespace=namespace,
             version=version,
             display_name=label,
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -29,6 +29,7 @@ from opentrons.protocol_engine import (
 from opentrons.protocol_engine.types import (
     ModuleModel as ProtocolEngineModuleModel,
     OFF_DECK_LOCATION,
+    LabwareLocation,
 )
 from opentrons.protocol_engine.clients import SyncClient as ProtocolEngineClient
 from opentrons.protocol_engine.errors import (
@@ -139,17 +140,7 @@ class ProtocolCore(
         version: Optional[int],
     ) -> LabwareCore:
         """Load a labware using its identifying parameters."""
-        load_location: Union[ModuleLocation, DeckSlotLocation, str]
-        if isinstance(location, (ModuleCore, NonConnectedModuleCore)):
-            load_location = ModuleLocation(moduleId=location.module_id)
-        elif location == OffDeckType.OFF_DECK:
-            load_location = OFF_DECK_LOCATION
-        elif isinstance(location, DeckSlotName):
-            load_location = DeckSlotLocation(slotName=location)
-        else:
-            raise UnknownLocationError(
-                f"load_labware is not supported with the given location: {location}."
-            )
+        load_location = self._convert_labware_location(location=location)
 
         custom_labware_params = (
             self._engine_client.state.labware.find_custom_labware_load_params()
@@ -212,17 +203,7 @@ class ProtocolCore(
         drop_offset: Optional[Tuple[float, float, float]],
     ) -> None:
         """Move the given labware to a new location."""
-        to_location: Union[ModuleLocation, DeckSlotLocation, str]
-        if isinstance(new_location, (ModuleCore, NonConnectedModuleCore)):
-            to_location = ModuleLocation(moduleId=new_location.module_id)
-        elif new_location == OffDeckType.OFF_DECK:
-            to_location = OFF_DECK_LOCATION
-        elif isinstance(new_location, DeckSlotName):
-            to_location = DeckSlotLocation(slotName=new_location)
-        else:
-            raise UnknownLocationError(
-                f"move_labware is not supported with the given location: {new_location}."
-            )
+        to_location = self._convert_labware_location(location=new_location)
 
         strategy = (
             LabwareMovementStrategy.USING_GRIPPER
@@ -518,8 +499,17 @@ class ProtocolCore(
                 labware_location.slotName, self._engine_client.state.config.robot_type
             )
         elif isinstance(labware_location, ModuleLocation):
-            # TODO (tz. 06-12-23) this will raise a key exception if module not found instead of returning None.
-            # This does not need to return OFF_DECK.
             return self._module_cores_by_id[labware_location.moduleId]
 
         return OffDeckType.OFF_DECK
+
+    @staticmethod
+    def _convert_labware_location(
+        location: Union[DeckSlotName, ModuleCore, NonConnectedModuleCore, OffDeckType]
+    ) -> LabwareLocation:
+        if isinstance(location, (ModuleCore, NonConnectedModuleCore)):
+            return ModuleLocation(moduleId=location.module_id)
+        elif location is OffDeckType.OFF_DECK:
+            return OFF_DECK_LOCATION
+        elif isinstance(location, DeckSlotName):
+            return DeckSlotLocation(slotName=location)
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -508,7 +508,7 @@ class ProtocolCore(
 
     def get_labware_location(
         self, labware_core: LabwareCore
-    ) -> Union[str, ModuleCore, NonConnectedModuleCore, None]:
+    ) -> Union[str, ModuleCore, NonConnectedModuleCore, OffDeckType]:
         """Get labware parent location."""
         labware_location = self._engine_client.state.labware.get_location(
             labware_core.labware_id
@@ -519,4 +519,5 @@ class ProtocolCore(
             )
         elif isinstance(labware_location, ModuleLocation):
             return self._module_cores_by_id.get(labware_location.moduleId)
-        return None
+
+        return OffDeckType.OFF_DECK
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -52,7 +52,7 @@ from .module_core import (
     NonConnectedModuleCore,
     MagneticBlockCore,
 )
-from .exceptions import InvalidModuleLocationError, UnknownLocationError
+from .exceptions import InvalidModuleLocationError
 from . import load_labware_params
 from . import deck_conflict
 
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -518,6 +518,8 @@ class ProtocolCore(
                 labware_location.slotName, self._engine_client.state.config.robot_type
             )
         elif isinstance(labware_location, ModuleLocation):
-            return self._module_cores_by_id.get(labware_location.moduleId)
+            # TODO (tz. 06-12-23) this will raise a key exception if module not found instead of returning None.
+            # This does not need to return OFF_DECK.
+            return self._module_cores_by_id[labware_location.moduleId]
 
         return OffDeckType.OFF_DECK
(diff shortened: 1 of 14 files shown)


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -297,12 +297,12 @@ def load_labware_from_definition(
...
         self,
         load_name: str,
-        location: DeckLocation,
+        location: Union[DeckLocation, OffDeckType],
         label: Optional[str] = None,
         namespace: Optional[str] = None,
         version: Optional[int] = None,
     ) -> Labware:
-        """Load a labware onto the deck given its name.
+        """Load a labware onto a location.
```
TamarZanzouri, line 305:
```suggestion
        """Load a labware into a location.
```
?

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -316,9 +316,9 @@ def load_labware(
             `Labware Library <https://labware.opentrons.com>`_.
 
         :param location: The slot into which to load the labware,
-            such as ``1`` or ``"1"``.
+            such as ``1`` or ``"1"`` or off deck using :py:obj:`OFF_DECK`.
```
TamarZanzouri:
```suggestion
            such as ``1`` or ``"1"`` or off-deck using :py:obj:`OFF_DECK`.
```

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -338,12 +338,20 @@ def load_labware(
         :param version: The version of the labware definition. You should normally
             leave this unspecified to let the implementation choose a good default.
         """
+        if isinstance(location, OffDeckType) and self._api_version < APIVersion(2, 15):
+            raise APIVersionError(
+                "Loading a labware off deck is supported at apiLevel 2.15 and higher."
```
TamarZanzouri:
```suggestion
                "Loading a labware off-deck is supported at apiLevel 2.15 and higher."
```

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -512,5 +518,8 @@ def get_labware_location(
                 labware_location.slotName, self._engine_client.state.config.robot_type
             )
         elif isinstance(labware_location, ModuleLocation):
-            return self._module_cores_by_id.get(labware_location.moduleId)
-        return None
+            # TODO (tz. 06-12-23) this will raise a key exception if module not found instead of returning None.
+            # This does not need to return OFF_DECK.
+            return self._module_cores_by_id[labware_location.moduleId]
```
SyntaxColoring: Is this actually something that we have to do anything about? Are we ever expecting the `labware_location.moduleId` key to not be present?

api/src/opentrons/protocol_api/core/legacy/legacy_protocol_core.py
```diff
@@ -147,12 +147,17 @@ def add_labware_definition(
...
-        location: Union[DeckSlotName, legacy_module_core.LegacyModuleCore],
+        location: Union[DeckSlotName, legacy_module_core.LegacyModuleCore, OffDeckType],
         label: Optional[str],
         namespace: Optional[str],
         version: Optional[int],
     ) -> LegacyLabwareCore:
         """Load a labware using its identifying parameters."""
+        if isinstance(location, OffDeckType):
+            raise APIVersionError(
+                "Loading a labware off deck is not supported in this API version"
```
SyntaxColoring:
```suggestion
                "Loading a labware off deck is only supported with apiLevel 2.15 and newer."
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -341,16 +342,18 @@ def uri(self) -> str:
...
     @property  # type: ignore[misc]
     @requires_version(2, 0)
-    def parent(self) -> Union[str, ModuleTypes, None]:
+    def parent(self) -> Union[str, ModuleTypes, OffDeckType]:
         """The parent of this labware.
 
         If the labware is on the deck, a `str` deck slot name will be returned.
         If on a module, the parent :py:class:`ModuleContext` will be returned.
-        If off deck, `None` will be returned.
+        If off deck, `off-deck` will be returned.
```
SyntaxColoring: To reference the actual value that users should use:

```suggestion
        If off deck, :py:obj:`OFF_DECK` will be returned.
```

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -338,12 +338,20 @@ def load_labware(
         :param version: The version of the labware definition. You should normally
             leave this unspecified to let the implementation choose a good default.
         """
+        if isinstance(location, OffDeckType) and self._api_version < APIVersion(2, 15):
+            raise APIVersionError(
+                "Loading a labware off-deck is supported at apiLevel 2.15 and higher."
```
SyntaxColoring: To be a little clearer about what's going wrong:

```suggestion
                "Loading a labware off-deck requires apiLevel 2.15 or higher."
```

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -106,8 +106,8 @@ def get_labware_parent_nominal_position(self, labware_id: str) -> Point:
         elif labware_data.location == OFF_DECK_LOCATION:
             # Labware is off-deck
             raise errors.LabwareNotOnDeckError(
-                f"Labware {labware_id} does not have a parent associated with it"
-                f" since it is no longer on the deck."
+                f"Cannot access labware {labware_id} since it is no longer on the deck. "
```
SyntaxColoring: Thanks for this better error message.

One clarification:

```suggestion
                f"Cannot access labware {labware_id} since it is not on the deck. "
```

Since the labware might not have been on the deck to begin with, if it was loaded off-deck.

api/src/opentrons/types.py
```diff
@@ -318,17 +345,17 @@ def __str__(self) -> str:
...
+    (DeckSlotName.SLOT_2, DeckSlotName.SLOT_D2),
+    (DeckSlotName.SLOT_3, DeckSlotName.SLOT_D3),
+    (DeckSlotName.SLOT_4, DeckSlotName.SLOT_C1),
+    (DeckSlotName.SLOT_5, DeckSlotName.SLOT_C2),
+    (DeckSlotName.SLOT_6, DeckSlotName.SLOT_C3),
+    (DeckSlotName.SLOT_7, DeckSlotName.SLOT_B1),
+    (DeckSlotName.SLOT_8, DeckSlotName.SLOT_B2),
+    (DeckSlotName.SLOT_9, DeckSlotName.SLOT_B3),
+    (DeckSlotName.SLOT_10, DeckSlotName.SLOT_A1),
+    (DeckSlotName.SLOT_11, DeckSlotName.SLOT_A2),
```
SyntaxColoring, line 352: We can remove the `# fmt: off` and `# fmt: on` comments if we're getting rid of the pretty column alignment.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -341,16 +342,18 @@ def uri(self) -> str:
...
 
         .. versionchanged:: 2.14
             Return type for module parent changed to :py:class:`ModuleContext`.
             Prior to this version, an internal geometry interface is returned.
+        .. versionchanged:: 2.15
+            Return type for off-deck changed from None to off-deck.
```
SyntaxColoring: As far as I can tell, there has never been a public release of the Python Protocol API where people would see `None` returned for off-deck labware.

But I just ran this quick experiment, which shows that they would have seen where the labware used to be:

```python
metadata = {
    "apiLevel": "2.13",
}

def run(protocol) -> None:
    l = protocol.load_labware("opentrons_96_tiprack_300ul", 1)
    print("Before del:", l.parent)  # prints "1"
    del protocol.deck[1]
    print("After del:", l.parent)  # prints "1" again
```

Weird! We may as well document it.

```suggestion
        .. versionchanged:: 2.15
            Will now return :py:obj:`OFF_DECK` if the labware is off-deck.
            Formerly, if the labware was removed by using ``del`` on :py:obj:`.deck`,
            this would return where it was before its removal.
```

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -133,17 +133,23 @@ def add_labware_definition(
...
+            load_location = ModuleLocation(moduleId=location.module_id)
+        elif location == OffDeckType.OFF_DECK:
+            load_location = OFF_DECK_LOCATION
+        elif isinstance(location, DeckSlotName):
+            load_location = DeckSlotLocation(slotName=location)
         else:
-            module_location = DeckSlotLocation(slotName=location)
+            raise UnknownLocationError(
+                f"load_labware is not supported with the given location: {location}."
+            )
```
SyntaxColoring: I see that this is copy-pasted from `move_labware()`, so it's not new, but I don't think it's appropriate to raise an exception here. This is an internal function that takes trusted input. If we reach the `else` statement, that's a bug in this `if`/`else` ladder, and we should indicate that with an `assert`, not a custom exception type.

Or, even better, we can turn the runtime error into a static type-checking error, by splitting this into a function:

```python
def _convert_labware_location(
    location: Union[
        DeckSlotName, ModuleCore, NonConnectedModuleCore, OffDeckType
    ]
) -> LabwareLocation:
    if isinstance(location, (ModuleCore, NonConnectedModuleCore)):
        return ModuleLocation(moduleId=location.module_id)
    elif location == OffDeckType.OFF_DECK:
        return OFF_DECK_LOCATION
    elif isinstance(location, DeckSlotName):
        return DeckSlotLocation(slotName=location)
```

And then updating `load_labware` and `move_labware` to use that.

Those are the only two uses of `UnknownLocationError`, so after doing that, we can then delete `UnknownLocationError`.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -316,9 +316,9 @@ def load_labware(
             `Labware Library <https://labware.opentrons.com>`_.
 
         :param location: The slot into which to load the labware,
-            such as ``1`` or ``"1"``.
+            such as ``1`` or ``"1"`` or off-deck using :py:obj:`OFF_DECK`.
 
-        :type location: int or str
+        :type location: int or str or :py:obj:`OFF_DECK
```
SyntaxColoring: Typo:

```suggestion
        :type location: int or str or :py:obj:`OFF_DECK`
```

api/src/opentrons/types.py
```diff
@@ -146,7 +166,14 @@ def move(self, point: Point) -> "Location":
...
         """
-        return Location(point=self.point + point, labware=self._labware.object)
+        from opentrons.protocol_api._types import OffDeckType
+
+        return Location(
+            point=self.point + point,
+            labware=self._labware.object
+            if not isinstance(self._labware.object, OffDeckType)
+            else None,
+        )
```
SyntaxColoring: It's impossible for `self._labware.object` to actually be an `OffDeckType`, right? Since that's not a possible value for the `labware` arg of `Location`'s `__init__()`?

If that's correct, let's not add conditionals here that we don't expect to actually run in production. I'm wary of adding more confusion around this class about what values are actually possible.

Let's either:

* `assert not isinstance(self._labware.object, OffDeckType)`. This should convince mypy that it's possible to construct the new `Location` with `self._labware.object`.
* Probably better: in `__init__()`, store both the original `labware` argument value *and* the `LabwareLike` that we've constructed from it. We should keep returning the `LabwareLike` from `.labware` to preserve existing behavior, but when we construct a new `Location` here in `move()`, we should construct it with the original `labware` argument value.

api/src/opentrons/protocol_api/core/engine/protocol.py
TamarZanzouri: nope! just wanted to be clear about remove the get. I can remove this comment

SyntaxColoring: Yeah. Nice catch and good to think about, but I think it's okay to just remove it.

SyntaxColoring (approved):
Code looks great to me once CI passes. Thanks!

PR #12911, opened 2023-06-14
refactor(app): zAxis error occurs during Mounting plate step

jerader:
closes RLIQ-407

# Overview

Previously, the zAxis error modal occurred during the `Carriage` step but we wanted to move it to the `MountingPlate` step since it doesn't fail until then.

# Test Plan

on a Flex, test the attach 96-channel flow and see that the zaxis modal error is being triggered properly. Observe the modal on both the ODD and the desktop app (make sure that the buttons are correct).

# Changelog

- remove error modal from Carriage component and fix test, also remove the `home` command.
- add the error modal + home command to the Mounting plate component
- fix the tests

# Review requests

see test plan

# Risk assessment

low


smb2268 (approved):
looks good and works as expected!

PR #12916, opened 2023-06-14
fix(api): increase pipette detection timeout

sfoster1:
We have a timeout for getting the pipette info response during tool attach, and it was too short, breaking the pipette attach process. By increasing the timeout we fix the issue.

Closes RQA-961


Laura-Danielle (approved):
makes sense

fsinapi (approved):
🕔

PR #12917, opened 2023-06-14
refactor(hardware-testing): Set photometric shaker RPM based on test volume

andySigler:
# Overview

This PR attempts to make is a bit simpler for the operator to correctly set the RPM of the shaker. The operator sets the test volume of the plate using `--volume`, then the script determines the correct RPM to use based on the volume:

- 1-200uL = `1500` rpm
- 200-250uL = `1100` rpm

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
lgtm

PR #12918, opened 2023-06-14
refactor(shared-data): move pipette name conversion functions to shared-data

Laura-Danielle:
## Overview

The protocol engine needs to utilize some of the pipette name conversions that were originally only used in the hardware controller. In order for the protocol engine to access this, I've moved these functions to shared-data.

# Test Plan

- Throw on a robot and make sure nothing breaks

# Changelog

- Moves all the helper functions from `ot3_pipette_config.py` to a new file in `shared_data/python` and associated tests.

# Review requests

Anything you'd like to change?

# Risk assessment

Ideally low...just moving around functions

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -25,7 +25,7 @@ from typing import (
     Union,
 )
 from opentrons.config.types import OT3Config, GantryLoad
-from opentrons.config import ot3_pipette_config, gripper_config, feature_flags as ff
+from opentrons.config import gripper_config, feature_flags as ff
 from .ot3utils import (
     axis_convert,
     create_move_group,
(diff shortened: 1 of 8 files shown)


DerekMaggio:
Would this at all affect provisioning of serial numbers for pipettes?

DerekMaggio (approved):
LGTM, how come you took out the 2 tests?

Laura-Danielle:
> LGTM, how come you took out the 2 tests?

I removed the helper load function and there are two tests in shared-data already that test the shared-data load function.

Laura-Danielle:
@DerekMaggio no this will not affect provisioning.

sfoster1 (approved):
Looks good, tested on a flex and it looks all right. Bit worried about those robot server ot2 only tests, let's make sure those pass.

PR #12920, opened 2023-06-15
feat(protocol-designer, step-generation, components, shared-data): add flex and move labware

b-cooper:
# Overview

Introduce a new file creation wizard within protocol designer that sequentially walks users through
initializing a protocol for either the OT2 or the Opentrons Flex. Plug these new values into the
rest of the application's state, and follow the path all the way to the analysis/protocol engine via
updates to step generation that allow it to handle moving labware and speccing a robot type inside of a JSON
protocol file.

 Use modern module visualization components for rendering deck map items now that the
deck definition is variable within protocol designer. Render timeline module state visually on the
module svg with a new generic module name label. Remove step-generation error handling for deck
collisions that don't apply to the Flex. Present the Magnetic Block on the File Info Modules
section.

# Test Plan

- create a Flex protocol in PD
- create an OT2 protocol in PD

# Risk assessment
medium, this has a large footprint


api/src/opentrons/protocol_runner/json_translator.py
```diff
@@ -136,6 +136,7 @@ def translate_commands(
         """Takes json protocol v6 and translates commands->protocol engine commands."""
         commands_list: List[pe_commands.CommandCreate] = []
         for command in protocol.commands:
+            print(f'\n\n\n\n\n\n{command}\n\n\n\n')
```
jerader, line 139: what's this? @b-cooper

shared-data/js/types.ts
```diff
@@ -324,11 +324,7 @@ export interface ModuleDefinition {
   twoDimensionalRendering: INode
 }
 
-export type AffineTransformMatrix = [
```
b-cooper, line 327: The matrix is actually 4x4

step-generation/src/commandCreators/atomic/replaceTip.ts
```diff
@@ -103,26 +106,28 @@ export const replaceTip: CommandCreator<ReplaceTipArgs> = (
...
-    pipetteAdjacentHeaterShakerWhileShaking(prevRobotState.modules, slotName)
-  ) {
-    return { errors: [errorCreators.heaterShakerNorthSouthEastWestShaking()] }
-  }
-  if (
-    getIsHeaterShakerEastWestWithLatchOpen(prevRobotState.modules, slotName)
-  ) {
-    return { errors: [errorCreators.heaterShakerEastWestWithLatchOpen()] }
-  }
+  if (!isFlexPipette) {
```
b-cooper, line 109: This logic is fine, but hopefully the errors will be more of a "OT2 only check", it would therefore be a little more future proof here to say `if (OT2)` compared to `if (!Flex)`

jerader: yeah, that makes sense. I was having a hard time finding how to get the robot type and pipettes was already there so figured it was fine to filter based on pipettes. But happy to change it to checking robot type in a follow up pr

jerader (approved):
🤩 woo!

PR #12923, opened 2023-06-15
fix(api): change Flex front USB port number to 9 to disambiguate

pmoegenburg:
# Overview

The Flex has 1 front USB port. It was previously given port number 1. This changes it to port number 9 to disambiguate it from the first port on the left USB port bank.

This PR also changes simulation and emulation port numbers from 0 so as not to be interpreted as null by layers higher in the stack.

# Test Plan

 - [x] verify front USB port displays as 9 in App network/module settings
 - [x] verify USBPort(port=1, hub=false, portGroup="main") for OT-2 simulation
 - [x] verify USBPort(port=1, hub=false, portGroup="left") for Flex simulation

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Nice, thank you!

PR #12928, opened 2023-06-16
feat(engine): load adapter in engine/load and move labware on new OnLabwareLocation

jbleon95:
# Overview

Closes RCORE-342.

This PR adds to the Protocol Engine the ability to load an adapter, load a labware on top of a compatible adapter, and updates all the geometry math for pipette and gripper movement to support this new location type.

The newly added command `loadAdapter` allows the loading of labware that has `adapter` in its defined `allowedRoles`. This is treated as a regular `LoadedLabware` object except for checks related to loading and moving labware around.

New to this PR is a new `LabwareLocation` where labware can be loaded onto and moved, `OnLabwareLocation`. This has one parameter `labwareId`, which refers to a labware on deck. This new location is handled recursively by the `GeometryView` class, so labware can sit on an adapter on a module and this will all be handled and calculated accurately.

# Test Plan

Loading labware onto adapters, on top of module and the deck, and testing pipette movement logic and gripper movement has been tested extensively via Postman on actual hardware.

# Changelog

- Added new engine command `loadAdapter`
- Added new location `OnLabwareLocation` for `loadLabware` and `moveLabware`
- Updated engine `GeometryView` position, slot name and module offset logic to handled `OnLabwareLocation`
- Added methods to `LabwareView` to get module offsets and raises for improper `OnLabwareLocation` loading
- Added restrictions for moving and loading labware onto adapters and other labware

# Review requests

- Do the shapes of the new adapter load, `OnLabwareLocation` class and other client facing objects make sense
- If any better names for methods or variables are had, please share, there are only so many synonyms for "beneath" I could think of.

# Risk assessment
Medium, this PR touches a lot of geometry logic that, if wrong, could result in pipette or gripper crashes. However, the existing location logic has not been functionally changed at all, and the new `OnLabwareLocation` logic has been tested to ensure that pipette movement is identical to prior joint adapter/labware definitions, and that gripper movement is accurate and does not crash.


api/src/opentrons/protocol_engine/commands/load_adapter.py
```diff
@@ -0,0 +1,126 @@
...
+    version: int = Field(
+        ...,
+        description="The labware definition version.",
+    )
+    adapterId: Optional[str] = Field(
+        None,
+        description="An optional ID to assign to this adapter. If None, an ID "
+        "will be generated.",
+    )
+    displayName: Optional[str] = Field(
```
jbleon95: consider removing this

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -95,43 +96,80 @@ def get_min_travel_z(
...
+            module_id = labware_location.moduleId
             deck_type = DeckType(self._labware.get_deck_definition()["otId"])
             module_offset = self._modules.get_nominal_module_offset(
                 module_id=module_id, deck_type=deck_type
             )
-            return Point(
-                slot_pos.x + module_offset.x,
-                slot_pos.y + module_offset.y,
-                slot_pos.z + module_offset.z,
+            module_model = self._modules.get_connected_model(module_id)
```
jbleon95, line 123: this won't work for magnetic block, find a different way to get the module model's name

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -335,38 +376,56 @@ def get_ancestor_slot_name(self, labware_id: str) -> DeckSlotName:
         return slot_name
 
     def ensure_location_not_occupied(
-        self, location: LabwareLocation
+        self, labware_id: str, location: LabwareLocation
```
jbleon95: look into splitting this up to check definition related checks in another function

api/src/opentrons/protocol_engine/state/geometry.py
jbleon95: correction: this will work for magnetic block, the method is just named inaccurately

b-cooper: Is it possible to at least add this "*NOTE despite the mag block not really being connected, this method will return all physically loaded modules, including mag block". to the docstring of `get_connected_model`. that way the next person will be less confused by the same confusing naming convention. Alternatively you could change the name to something more appropriate.

b-cooper (approved):
The logic makes sense to me. Looks good 👍

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -139,6 +140,10 @@ def _map_labware(
         # let _map_module() pick it up.
         return None
 
+    elif isinstance(location_from_engine, OnLabwareLocation):
+        # TODO(jbl 2023-06-08) check if we need to do any logic here or if this is correct
```
vegano1, line 144: still need this todo?

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -618,6 +671,37 @@ def raise_if_labware_in_location(
                     f"Labware {labware.loadName} is already present at {location}."
                 )
 
+    def raise_if_labware_cannot_be_stacked(
```
vegano1, line 674: There is a physical limit to how many things can be stacked on top of each other defined by the top of the deck to the critical point of the instrument - maybe I'm missing it or another part of the system is taking care of this but I don't see this being taken into account.

vegano1:
Nice, looks good but left a question.
Also, this affects the basic functions we use for module calibration so we should probably make sure that still works.
You can use [this](https://github.com/Opentrons/opentrons/blob/edge/hardware-testing/hardware_testing/scripts/module_calibration.py) HTTP script to perform a full module calibration as if you were the client.

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -618,6 +671,37 @@ def raise_if_labware_in_location(
...
+        if not labware_validation.validate_labware_can_be_stacked(
+            top_labware_definition=top_labware_definition,
+            below_labware_load_name=below_labware.loadName,
+        ):
+            raise errors.LabwareCannotBeStackedError(
+                f"Labware {top_labware_definition.parameters.loadName} cannot be loaded onto labware {below_labware.loadName}"
+            )
+        elif isinstance(below_labware.location, ModuleLocation):
+            below_definition = self.get_definition(labware_id=below_labware.id)
+            if not labware_validation.validate_definition_is_adapter(below_definition):
```
TamarZanzouri, line 688: why do we need to validate if a `ModuleLocation` is an adapter?

api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -609,9 +609,13 @@ def get_requested_model(self, module_id: str) -> Optional[ModuleModel]:
         except KeyError as e:
             raise errors.ModuleNotLoadedError(module_id=module_id) from e
 
+    # TODO(jbl 2023-06-20) rename this method to better reflect it's not just "connected" modules
     def get_connected_model(self, module_id: str) -> ModuleModel:
```
TamarZanzouri, line 613: maybe we should change this to 
```suggestion
    def get_loaded_model(self, module_id: str) -> ModuleModel:
```

api/src/opentrons/protocol_engine/types.py
```diff
@@ -63,12 +63,26 @@ class ModuleLocation(BaseModel):
     )
 
 
+class OnLabwareLocation(BaseModel):
+    """The location of something placed atop another labware."""
```
TamarZanzouri, line 67: is this something or a labware?

api/src/opentrons/protocol_engine/types.py
```diff
@@ -348,6 +362,22 @@ class LabwareOffsetVector(BaseModel):
     y: float
     z: float
 
+    def __add__(self, other: Any) -> LabwareOffsetVector:
+        """Adds two vectors together."""
+        if not isinstance(other, LabwareOffsetVector):
+            return NotImplemented
```
TamarZanzouri, line 368: can we add a description for this exception?

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -248,6 +259,17 @@ def get_id_by_module(self, module_id: str) -> str:
             "There is no labware loaded on this Module"
         )
 
+    def raise_if_labware_is_not_on_top(self, labware_id: str) -> None:
```
TamarZanzouri: can we change this method name? I find this very confusing. maybe we can do something like:
```suggestion
    def raise_if_labware_has_labware_on_top(self, labware_id: str) -> None:
```

really hard finding names for this PR ;-)

TamarZanzouri:
```suggestion
    def raise_if_labware_has_labware_above(self, labware_id: str) -> None:
```

!!!

api/src/opentrons/protocol_engine/commands/move_labware.py
```diff
@@ -94,12 +97,29 @@ async def execute(self, params: MoveLabwareParams) -> MoveLabwareResult:
         """Move a loaded labware to a new location."""
         # Allow propagation of LabwareNotLoadedError.
         current_labware = self._state_view.labware.get(labware_id=params.labwareId)
+        current_labware_definition = self._state_view.labware.get_definition(
+            labware_id=params.labwareId
+        )
         definition_uri = current_labware.definitionUri
 
         empty_new_location = self._state_view.geometry.ensure_location_not_occupied(
```
TamarZanzouri: sorry for the nip but it can also be a OnLabwareLocation correct? can we change the var name to `move_location` or something like that?
```suggestion
        move_to_location = self._state_view.geometry.ensure_location_not_occupied(
```

TamarZanzouri (changes requested):
didn't finish it all but its a good start. these changes are really really good! just a bunch of questions/suggestions.

api/src/opentrons/protocol_engine/state/labware.py
jbleon95: It's validating if the labware that's sitting on a module is an adapter, for the reason that the only two "labware" you can place in a module is a labware on top of an adapter. This logic prevents two labwares from being placed on a module. The below check will prevent two labware from sitting on an adapter

api/src/opentrons/protocol_engine/types.py
jbleon95: It's not an exception, it's returning a built-in that's recommend for dunder methods https://docs.python.org/3.7/library/constants.html#NotImplemented

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
jbleon95: this will be addressed in the PAPI implementation ticket/PR

api/src/opentrons/protocol_engine/state/labware.py
jbleon95: There's currently no designated hard limit, but the current shared definition data does not have any labware that can stack with itself, i.e. any arbitrarily tall stacking, only labware on top of specific adapters. A limit, whether it is a set number of labware and/or based on labware height, will be added when that feature is officially added.

api/src/opentrons/protocol_engine/state/labware.py
TamarZanzouri: This makes more sense! thank you for explaining that!

api/src/opentrons/protocol_engine/types.py
TamarZanzouri: good to know! thank you!

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -170,29 +175,35 @@ def handle_action(self, action: Action) -> None:
...
-            self._state.labware_by_id[labware_id] = LoadedLabware.construct(
-                id=labware_id,
+            self._state.definitions_by_uri[definition_uri] = command.result.definition
+
+            if isinstance(command.result, LoadLabwareResult):
+                deck_item_id = command.result.labwareId
+                display_name = command.params.displayName
+            else:
+                deck_item_id = command.result.adapterId
+                display_name = None
```
TamarZanzouri, line 196: is this meant to be None? can we not use the labware definition display name? I just wonder what this used for in the UI

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -344,6 +366,13 @@ def get_location(self, labware_id: str) -> LabwareLocation:
         """Get labware location by the labware's unique identifier."""
         return self.get(labware_id).location
 
+    def get_parent_location(self, labware_id: str) -> NonStackedLocation:
+        """Get labware's non-labware parent location."""
+        parent = self.get_location(labware_id)
+        if isinstance(parent, OnLabwareLocation):
```
TamarZanzouri, line 372: Nice!

api/src/opentrons/protocol_engine/state/modules.py
TamarZanzouri: not for this pr

api/tests/opentrons/protocol_engine/commands/test_move_labware.py
```diff
@@ -78,6 +91,70 @@ async def test_manual_move_labware_implementation(
...
+        equipment.find_applicable_labware_offset_id(
+            labware_definition_uri="opentrons-test/load-name/1",
+            labware_location=OnLabwareLocation(labwareId="my-even-cooler-labware-id"),
+        )
+    ).then_return("wowzers-a-new-offset-id")
+
+    result = await subject.execute(data)
+    decoy.verify(
+        state_view.labware.raise_if_labware_is_not_on_top("my-cool-labware-id"),
+        state_view.labware.raise_if_labware_is_not_on_top("my-even-cooler-labware-id"),
```
TamarZanzouri: hahaha

api/src/opentrons/protocol_engine/commands/load_adapter.py
```diff
@@ -0,0 +1,121 @@
...
+            location=params.location,
+            labware_id=params.adapterId,
+        )
+
+        if not labware_validation.validate_definition_is_adapter(
+            loaded_labware.definition
+        ):
+            raise LabwareDefinitionIsNotAdapterError(
+                f"{params.loadName} is not defined as an adapter"
+            )
```
sanni-t, line 99: Hmm, it sucks that we have to do this validation *after* the labware is loaded. Let's leave a comment to see if we can validate before loading

sanni-t (approved):
This looks really great! I'm good with merging this so that the API work is unblocked.
I have a few minor suggestions coming up (mostly naming & comments) but we can address them in a later PR.

TamarZanzouri (approved):
Beautiful! Thank you for your changes!

PR #12932, opened 2023-06-16
fix(app): Detach pipette refetch

jerader:
RQA-983

# Overview

at the end of detach pipette flow, the instruments dashboard on ODD refetches instruments

# Test Plan

start a detach pipette flow on desktop app, on completion, the ODD instruments will update correctly.

# Changelog

- add refetch interval to InstrumentsDashboard

# Review requests

see test plan

# Risk assessment

low


sfoster1 (approved):
Tested on a Flex by pushing this branch ODD to the machine and checking the error conditions in RQA-983; looks fixed and good to me!

PR #12938, opened 2023-06-20
fix(app): remove temporary instrument loading text on robot card

b-cooper:
# Overview

Remove unnecessary loading text for instruments on the robot card.  In a performance fix from a few weeks back, we lazily loaded the instruments and pipettes on each instance of the Robot Card on the device list page.  There were leftover placeholder loading states that got merged in that need to be removed.


brenthagen (approved):
thanks, i was wondering about this

PR #12941, opened 2023-06-20
fix(hardware): allow longer for pipettes to boot

sfoster1:
Pipettes sometimes don't boot fast enough on attach and miss the InstrumentInfoRequest that happens when the head tells us a new instrument is present. This didn't used to be an issue because we wouldn't send the request until the user was finished screwing things in, so we just didn't notice.

In addition to our previous timeout - or, at a certain level, instead of it - where we'd wait a long time for the pipette to respond, we'll also now retry sending the request a couple times too.

## Testing
On a flex,
- [x] test that when you attach a 96 channel you retry the instrument info request a couple times until the pipette responds; then it stops retrying
- [x] test that when you then remove it, it doesn't retry

Fixes RQA-998


ahiuchingau (approved):
⌛

PR #12942, opened 2023-06-21
fix(hardware-testing): Always call OT3API.cache_instruments() after building

andySigler:
# Overview

This PR adds a call to `OT3API.cache_instruments()` after each time the hcapi is instantiated for hardware-testing scripts, in order to get all attached devices to be in state.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (changes requested):
Hm. I think it's probably better to just use `cache_instruments` - `reset` is a function of an earlier time and machine that's probably not super reliable on the ot3 going forward.

sfoster1 (approved):
ty, ty!

PR #12944, opened 2023-06-21
feat(hardware): add acceleration to pick up tip for 96 channel

caila-marashaj:
This pr adds acceleration to pick up tip and drop tip moves for the gear motors on the 96 channel pipette. The `ot3api` signature's for `pick_up_tip` and `drop_tip` haven't changed, but the hardware controller's `tip_action` function will now optionally take in a list of moves. For the purpose of homing the gear motors, the backend `tip_action` function can also just take in a `distance` and `velocity`.

   When `pick_up_tip` or `drop_tip` is called, the api function will now:
  - check to see if the gear motor's position is known- if not, it will home the gear motors before starting to move
  - perform the downward part of the tip action movement as normal, with acceleration
  - move most of the way back up also with acceleration
  - move the last 5 mm at a constant speed while looking for the limit switch to be pressed

## Changelog

- changed default speed for Q axis to 10
- created separate homing functions in `ot3controller` and `ot3utils` for the gear motors
- made backoff optional for homing the gear motors
- added a `gear_motor_position` property in `ot3controller`
- tweaked `get_system_constraints` so we can grab default acceleration for the Q axis
- gave `Axis.Q` its own enum value, rather than having it be part of `OTHER`
- added parsing for gear motor positions in `move_group_runner`

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -638,13 +638,11 @@ class OT3Controller:
         tip_action: str = "home",
     ) -> None:
         move_group = []
-        if tip_action == "clamp":
-            assert moves is not None
+        if moves is not None:
             move_group = create_tip_action_group(
                 moves, [NodeId.pipette_left], tip_action
             )
-        elif tip_action == "home":
-            assert distance is not None and velocity is not None
+        elif distance is not None and velocity is not None:
             move_group = create_gear_motor_home_group(float(distance), float(velocity))
         runner = MoveGroupRunner(
             move_groups=[move_group],
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -628,6 +628,7 @@ class OT3Controller:
         distance: Optional[float] = None,
         velocity: Optional[float] = None,
         tip_action: str = "home",
+        back_off: Optional[bool] = False,
     ) -> None:
         move_group = []
         if moves is not None:
@@ -635,7 +636,9 @@ class OT3Controller:
                 moves, [NodeId.pipette_left], tip_action
             )
         elif distance is not None and velocity is not None:
-            move_group = create_gear_motor_home_group(float(distance), float(velocity))
+            move_group = create_gear_motor_home_group(
+                float(distance), float(velocity), back_off
+            )
         runner = MoveGroupRunner(
             move_groups=[move_group],
             ignore_stalls=True if not ff.stall_detection_enabled() else False,
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -368,7 +368,7 @@ class OT3Controller:
         if updated_successfully:
             self._gear_motor_position = {NodeId.pipette_left: motor_response[0]}
 
-    async def update_motor_estimation(self, axes: Sequence[OT3Axis]) -> None:
+    async def update_motor_estimation(self, axes: Sequence[Axis]) -> None:
         """Update motor position estimation for commanded nodes, and update cache of data."""
         nodes = set([axis_to_node(a) for a in axes])
         response = await update_motor_position_estimation(self._messenger, nodes)
@@ -605,14 +605,14 @@ class OT3Controller:
         ]
         async with self._monitor_overpressure(moving_pipettes):
             positions = await asyncio.gather(*coros)
-        if Axis.Q in checked_axes:
-            await self.tip_action(
-                [Axis.Q],
-                self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
-                self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
-                    Axis.to_kind(Axis.Q)
-                ],
-            )
+        # if Axis.Q in checked_axes:
+            # await self.tip_action(
+            #     [Axis.Q],
+            #     self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
+            #     self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
+            #         Axis.to_kind(Axis.Q)
+            #     ],
+            # )
         for position in positions:
             self._handle_motor_status_response(position)
         return axis_convert(self._position, 0.0)
@@ -631,7 +631,7 @@ class OT3Controller:
 
     async def tip_action(
         self,
-        moves: Optional[List[Move[OT3Axis]]] = None,
+        moves: Optional[List[Move[Axis]]] = None,
         distance: Optional[float] = None,
         velocity: Optional[float] = None,
         tip_action: str = "home",
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -610,21 +610,15 @@ class OT3Controller:
         axes: Sequence[Axis],
         distance: float,
         speed: float,
+        acceleration: float = 0,
         tip_action: str = "home",
     ) -> None:
         if tip_action == "home":
             speed = speed * -1
-            runner = MoveGroupRunner(
-                move_groups=create_tip_action_home_group(axes, distance, speed)
-            )
-        else:
-            runner = MoveGroupRunner(
-                move_groups=[
-                    create_tip_action_group(
-                        axes, distance, speed, cast(PipetteAction, tip_action)
-                    )
-                ]
-            )
+        move_group = create_tip_action_group(
+            axes, distance, speed, acceleration, cast(PipetteAction, tip_action)
+        )
+        runner = MoveGroupRunner(move_groups=[move_group])
         positions = await runner.run(can_messenger=self._messenger)
         for axis, point in positions.items():
             self._position.update({axis: point[0]})
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -605,14 +605,15 @@ class OT3Controller:
         ]
         async with self._monitor_overpressure(moving_pipettes):
             positions = await asyncio.gather(*coros)
-        # if Axis.Q in checked_axes:
-        # await self.tip_action(
-        #     [Axis.Q],
-        #     self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
-        #     self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
-        #         Axis.to_kind(Axis.Q)
-        #     ],
-        # )
+        # TODO(CM): default gear motor homing routine to have some acceleration
+        if Axis.Q in checked_axes:
+            await self.tip_action(
+                distance=self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
+                velocity=self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
+                    Axis.to_kind(Axis.Q)
+                ],
+                tip_action="home",
+            )
         for position in positions:
             self._handle_motor_status_response(position)
         return axis_convert(self._position, 0.0)
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -5,7 +5,6 @@ import asyncio
 from contextlib import asynccontextmanager
 from functools import wraps
 import logging
-import time
 from copy import deepcopy
 from typing import (
     Any,
@@ -67,7 +66,6 @@ from opentrons_hardware.hardware_control.move_group_runner import MoveGroupRunne
 from opentrons_hardware.hardware_control.motion_planning import (
     Move,
     Coordinates,
-    Block,
 )
 from opentrons_hardware.hardware_control.estop.detector import (
     EstopDetector,
@@ -612,24 +610,35 @@ class OT3Controller:
             self._handle_motor_status_response(position)
         return axis_convert(self._position, 0.0)
 
+    def _filter_move_group(self, move_group: MoveGroup) -> MoveGroup:
+        new_group: MoveGroup = []
+        for step in move_group:
+            new_group.append(
+                {
+                    node: axis_step
+                    for node, axis_step in step.items()
+                    if node in self._motor_nodes()
+                }
+            )
+        return new_group
+
     async def tip_action(
         self,
-        axes: Sequence[Axis],
-        distance: float,
-        speed: float,
-        acceleration: float = 0,
+        moves: List[Move],
         tip_action: str = "home",
+        accelerate_during_move: bool = True,
     ) -> None:
-        if tip_action == "home":
-            speed = speed * -1
-        move_group = create_tip_action_group(
-            axes, distance, speed, acceleration, cast(PipetteAction, tip_action)
+        group = create_tip_action_group(moves, [NodeId.pipette_left], tip_action, accelerate_during_move)
+        move_group = group
+        runner = MoveGroupRunner(
+            move_groups=[move_group],
+            ignore_stalls=True if not ff.stall_detection_enabled() else False,
         )
-        # print(f"about to create runner at {time.time()}")
-        runner = MoveGroupRunner(move_groups=[move_group])
         positions = await runner.run(can_messenger=self._messenger)
-        # print(f"positions returned {positions}, at {time.time()}")
-        # await update_gear_motor_position_estimation(self._messenger)
+        print(f"positions returned = {positions}")
+        if NodeId.pipette_left in positions:
+            self._gear_motor_position = positions[NodeId.pipette_left][0]
+        print(f"gear position is now {self._gear_motor_position}")
 
     @requires_update
     async def gripper_grip_jaw(
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -81,6 +81,7 @@ from opentrons_hardware.hardware_control.motor_enable_disable import (
 from opentrons_hardware.hardware_control.motor_position_status import (
     get_motor_position,
     update_motor_position_estimation,
+    update_gear_motor_position_estimation,
 )
 from opentrons_hardware.hardware_control.limit_switches import get_limit_switches
 from opentrons_hardware.hardware_control.tip_presence import get_tip_ejector_state
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -606,13 +606,13 @@ class OT3Controller:
         async with self._monitor_overpressure(moving_pipettes):
             positions = await asyncio.gather(*coros)
         # if Axis.Q in checked_axes:
-            # await self.tip_action(
-            #     [Axis.Q],
-            #     self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
-            #     self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
-            #         Axis.to_kind(Axis.Q)
-            #     ],
-            # )
+        # await self.tip_action(
+        #     [Axis.Q],
+        #     self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
+        #     self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
+        #         Axis.to_kind(Axis.Q)
+        #     ],
+        # )
         for position in positions:
             self._handle_motor_status_response(position)
         return axis_convert(self._position, 0.0)
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -631,6 +631,8 @@ class OT3Controller:
         back_off: Optional[bool] = False,
     ) -> None:
         move_group = []
+        # make sure either moves or distance and velocity is not None
+        assert any([moves, distance and velocity])
         if moves is not None:
             move_group = create_tip_action_group(
                 moves, [NodeId.pipette_left], tip_action
@@ -639,6 +641,7 @@ class OT3Controller:
             move_group = create_gear_motor_home_group(
                 float(distance), float(velocity), back_off
             )
+
         runner = MoveGroupRunner(
             move_groups=[move_group],
             ignore_stalls=True if not ff.stall_detection_enabled() else False,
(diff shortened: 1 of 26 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1615,17 +1615,28 @@ async def _motor_pick_up_tip(
...
-            )
+
+            gear_motor_origin = await self._backend.gear_motor_position_estimation()
+            gear_origin_dict = {OT3Axis.Q: gear_motor_origin[0]}
+            gear_motor_target = pipette_spec.pick_up_distance + pipette_spec.home_buffer
+            gear_target_dict = {OT3Axis.Q: gear_motor_target}
+            moves = self._build_moves(gear_origin_dict, gear_target_dict)
+            blocks = moves[0][0].blocks
+
+            for block in blocks:
```
Laura-Danielle: we should pass moves into the backend tip action

Laura-Danielle (changes requested):
We talked on slack about it, but you need to make sure you're essentially mimicking what `ot3_utils::create_move_group` does for the tip action steps. Otherwise you won't be moving the motors correctly.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -241,6 +243,7 @@ def __init__(
             FirmwareUpdate(),
         )
         self._position = self._get_home_position()
+        self._gear_motor_position: float
```
Laura-Danielle: why doesn't this default to anything?

hardware/opentrons_hardware/firmware_bindings/messages/payloads.py
```diff
@@ -223,6 +223,13 @@ class MoveCompletedPayload(MoveGroupResponsePayload):
     ack_id: utils.UInt8Field
 
 
+@dataclass(eq=False)
+class GearMotorPositionResponse(EmptyPayload):
+    """Read Gear Motor Position Estimation."""
+
+    current_position: utils.UInt32Field
```
Laura-Danielle: We can still have position flags here imo, it would just be related to boot-up only.

hardware/opentrons_hardware/firmware_bindings/messages/payloads.py
```diff
@@ -556,6 +564,12 @@ class TipActionResponsePayload(MoveCompletedPayload):
     success: utils.UInt8Field
     gear_motor_id: GearMotorIdField
 
+    seq_id: utils.UInt8Field
```
Laura-Danielle: This should all be included in the `MoveCompletedPayload` so we don't need these values.

hardware/opentrons_hardware/hardware_control/motion.py
```diff
@@ -172,23 +172,24 @@ def create_home_step(
...
     present_nodes: Iterable[NodeId],
     action: PipetteTipActionType,
 ) -> MoveGroupStep:
     """Creates a step for tip handling actions that require motor movement."""
+    if action == PipetteTipActionType.home:
+        stop_condition = MoveStopCondition.limit_switch
+    else:
+        stop_condition = MoveStopCondition.none
     step: MoveGroupStep = {}
-    stop_condition = (
```
Laura-Danielle, line 194: This change isn't really necessary.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -437,10 +450,12 @@ def __call__(
             self._handle_move_completed(message, arbitration_id)
         elif isinstance(message, TipActionResponse):
             gear_id = GearMotorId(message.payload.gear_motor_id.value)
-            self._expected_tip_action_motors.remove(gear_id)
-            if len(self._expected_tip_action_motors) == 0:
+            seq_id = message.payload.seq_id.value
+            self._expected_tip_action_motors[seq_id].remove(gear_id)
+            print(f"expected tip motors = {self._expected_tip_action_motors}")
```
Laura-Danielle: print statement here.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -370,16 +378,21 @@ def _remove_move_group(
         group_id = message.payload.group_id.value - self._start_at_index
         node_id = arbitration_id.parts.originating_node_id
         try:
+
             in_group = (node_id, seq_id) in self._moves[group_id]
             self._moves[group_id].remove((node_id, seq_id))
+            print(f"removing {node_id}, {seq_id}")
```
Laura-Danielle: print statement here

hardware/opentrons_hardware/hardware_control/motor_position_status.py
```diff
@@ -143,3 +146,69 @@ def _listener_filter(arbitration_id: ArbitrationId) -> bool:
...
+    """
+
+    def _listener_filter(arbitration_id: ArbitrationId) -> bool:
+        return (
+            NodeId(arbitration_id.parts.originating_node_id) in {NodeId.pipette_left}
+        ) and (
+            MessageId(arbitration_id.parts.message_id)
+            in [
+                UpdateGearMotorPositionEstimationResponse.message_id,
+                # TipActionResponse.message_id,
```
Laura-Danielle: accidental add here?

hardware/opentrons_hardware/hardware_control/motor_position_status.py
```diff
@@ -93,7 +96,7 @@ async def _parser_update_motor_position_response(
         node = NodeId(arb_id.parts.originating_node_id)
         if node == expected:
             return (
-                float(response.payload.current_position.value / 1000.0),
+                float(response.payload.current_position.value),
```
Laura-Danielle: accidental change?

hardware/opentrons_hardware/hardware_control/motor_position_status.py
```diff
@@ -143,3 +146,69 @@ def _listener_filter(arbitration_id: ArbitrationId) -> bool:
...
+            MessageId(arbitration_id.parts.message_id)
+            in [
+                UpdateGearMotorPositionEstimationResponse.message_id,
+                # TipActionResponse.message_id,
+            ]
+        )
+
+    data = []
+
+    with WaitableCallback(can_messenger, _listener_filter) as reader:
```
Laura-Danielle: You can use `MultipleMessagesWaitableCallback`. See `get_motor_position` above. It will make the code much cleaner.

hardware/opentrons_hardware/hardware_control/motor_position_status.py
```diff
@@ -143,3 +146,69 @@ def _listener_filter(arbitration_id: ArbitrationId) -> bool:
...
+                        reader, NodeId.pipette_left
+                    ),
+                    timeout,
+                )
+                data.append(response)
+            # make sure response from both gear motors is the same
+            assert data[0] == data[1]
+            if not data[0][1] or not data[1][1]:
+                # If the stepper_ok flag isn't set, that means the node didn't update position.
+                raise RuntimeError(
```
Laura-Danielle: I think we should only be checking for a timeout error in this function. We should ensure that the motor position flag is set to `not OK` if the two gear motor positions don't match though.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -169,8 +169,6 @@ def _accumulate_move_completions(
             List[Tuple[Tuple[int, int], float, float, bool, bool]]
         ] = defaultdict(list)
         for arbid, completion in completions:
-            if isinstance(completion, TipActionResponse):
```
Laura-Danielle, line 177: We need this there otherwise it will mess up the plunger position. We will have to keep track of the tip action positions separately.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -346,10 +355,9 @@ def __init__(self, move_groups: MoveGroups, start_at_index: int = 0) -> None:
                 move_set.update(set((k.value, seq_id) for k in move.keys()))
                 duration += float(movesteps[0].duration_sec)
                 if any(isinstance(g, MoveGroupTipActionStep) for g in movesteps):
-                    self._expected_tip_action_motors = [
-                        GearMotorId.left,
-                        GearMotorId.right,
-                    ]
+                    self._expected_tip_action_motors.append(
```
Laura-Danielle: this change is also not necessary.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -317,6 +324,17 @@ async def update_motor_status(self) -> None:
         response = await get_motor_position(self._messenger, motor_nodes)
         self._handle_motor_status_response(response)
 
+    async def gear_motor_position_estimation(self) -> Tuple[float, bool]:
+        motor_response = await update_gear_motor_position_estimation(self._messenger)
+        updated_successfully = motor_response[1]
+        if updated_successfully:
```
Laura-Danielle: This function shouldn't return anything, only update `_gear_motor_position`

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -563,14 +581,6 @@ async def home(
             if runner
         ]
         positions = await asyncio.gather(*coros)
-        if OT3Axis.Q in checked_axes:
```
Laura-Danielle: We should be updating the position here and also why did we remove the tip action from the home command? We need to handle the use-case where a home command is called without any parameters.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -587,23 +597,33 @@ def _filter_move_group(self, move_group: MoveGroup) -> MoveGroup:
...
         return new_group
 
-    async def tip_action(
+    async def home_gear_motors(
         self,
-        axes: Sequence[OT3Axis],
         distance: float,
-        speed: float,
+        velocity: float,
+    ) -> None:
```
Laura-Danielle: imo this should still just be one function. Let the utility function handle any difference between passing acceleration or not.

Laura-Danielle (changes requested):
Initial comments.

hardware/tests/opentrons_hardware/hardware_control/test_move_group_runner.py
```diff
@@ -770,10 +771,10 @@ async def test_home_timeout(
         await subject.run(can_messenger=mock_can_messenger)
 
 
-async def test_tip_action_move_runner_receives_two_responses(
+async def test_tip_action_move_runner_receives_response(
```
Laura-Danielle: We are still receiving multiple responses from the tip action request. We should have a test that captures that.

In fact, we should expand on this test to mimic all the varying responses you can get from a tip action (i.e. if it's an acceleration move, then 3 responses for each motor) or if it's a regular home then just two responses.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -478,10 +491,11 @@ def __call__(
             self._handle_move_completed(message, arbitration_id)
         elif isinstance(message, TipActionResponse):
             gear_id = GearMotorId(message.payload.gear_motor_id.value)
-            self._expected_tip_action_motors.remove(gear_id)
-            if len(self._expected_tip_action_motors) == 0:
+            seq_id = message.payload.seq_id.value
+            self._expected_tip_action_motors[seq_id].remove(gear_id)
```
Laura-Danielle: Didn't you need to modify this to handle the varying responses you'd get back from the gear motors?

hardware/opentrons_hardware/hardware_control/motor_position_status.py
Laura-Danielle: Bump on this comment.

hardware/opentrons_hardware/hardware_control/motor_position_status.py
Laura-Danielle: Bump on this comment.

api/tests/opentrons/hardware_control/test_ot3_api.py
```diff
@@ -1662,6 +1714,8 @@ async def test_tip_presence_disabled_ninety_six_channel(
         instr_data = OT3AttachedPipette(config=pipette_config, id="fakepip")
         await ot3_hardware.cache_pipette(OT3Mount.LEFT, instr_data, None)
         await ot3_hardware._configure_instruments()
+        await ot3_hardware.set_gantry_load(GantryLoad.HIGH_THROUGHPUT)
+        # await ot3_hardware.cache_instruments()
```
Laura-Danielle: is this comment meant to be here, or did this somehow leak in from a rebase maybe?

Laura-Danielle: This is what the test looks like in edge:

```
    with patch.object(
        ot3_hardware.managed_obj._backend,
        "get_tip_present",
        AsyncMock(spec=ot3_hardware.managed_obj._backend.get_tip_present),
    ) as tip_present:
        pipette_config = load_pipette_data.load_definition(
            PipetteModelType("p1000"),
            PipetteChannelType(96),
            PipetteVersionType(3, 3),
        )
        instr_data = AttachedPipette(config=pipette_config, id="fakepip")
        await ot3_hardware.cache_pipette(OT3Mount.LEFT, instr_data, None)
        await ot3_hardware._configure_instruments()
        await ot3_hardware.pick_up_tip(OT3Mount.LEFT, 60)

        tip_present.assert_not_called()
 ```

File: api/tests/opentrons/hardware_control/test_ot3_api.py [hunk 20]
```diff
@@ -468,6 +469,8 @@ async def test_blow_out_position(
     blowout_volume: float,
 ) -> None:
     for mount, configs in load_configs.items():
+        if configs["channels"] == 96:
```
Laura-Danielle, line 473: why do we need this change?

api/tests/opentrons/hardware_control/test_ot3_api.py
```diff
@@ -513,6 +515,8 @@ async def test_blow_out_error(
...
```
Laura-Danielle, line 519: same here

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1730,18 +1760,17 @@ async def drop_tip(
...
-                await self._backend.tip_action(
-                    [Axis.of_main_tool_actuator(mount)],
-                    move.target_position + move.home_buffer,
-                    move.speed,
-                    "home",
+                await self._backend.update_gear_motor_position()
+                gear_start_position = axis_convert(
+                    self._backend.gear_motor_position, 0.0
                 )
+                drop_target_dict = {Axis.Q: move.target_position}
```
Laura-Danielle: nit: this is in pick up tip so should be called `pickup_target` etc

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1730,18 +1760,17 @@ async def drop_tip(
...
+                drop_moves = self._build_moves(gear_start_position, drop_target_dict)
+
+                await self._backend.tip_action(moves=drop_moves[0], tip_action="clamp")
+
+                await self.home_gear_motors()
```
Laura-Danielle: Let's not use this function. We should be able to use tip action just like we previously did in this function just with kwargs now instead.

For example, 

```
  await self._backend.tip_action(
      [Axis.of_main_tool_actuator(mount)],
      distance=move.target_position + move.home_buffer,
      velocity=move.speed,
      tip_action="home")
```

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -698,6 +700,36 @@ async def home_z(
             axes = list(Axis.ot3_mount_axes())
         await self.home(axes)
 
+    async def home_gear_motors(self) -> None:
```
Laura-Danielle: We should remove this function completely.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1646,19 +1678,17 @@ async def _motor_pick_up_tip(
...
+            gear_origin_float = axis_convert(self._backend.gear_motor_position, 0.0)[
+                Axis.P_L
+            ]
+            gear_origin_dict = {Axis.Q: gear_origin_float}
+            clamp_move_target = pipette_spec.pick_up_distance
+            gear_target_dict = {Axis.Q: clamp_move_target}
+            clamp_moves = self._build_moves(gear_origin_dict, gear_target_dict)
+            await self._backend.tip_action(moves=clamp_moves[0], tip_action="clamp")
+
+            await self.home_gear_motors()
```
Laura-Danielle: Same feedback applies here as in `pick_up_tip`.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1646,19 +1678,17 @@ async def _motor_pick_up_tip(
...
-            # back clamps off the adapter posts
-            await self._backend.tip_action(
-                [Axis.of_main_tool_actuator(mount)],
-                pipette_spec.pick_up_distance + pipette_spec.home_buffer,
-                pipette_spec.speed,
-                "home",
-            )
+            await self._backend.update_gear_motor_position()
+            gear_origin_float = axis_convert(self._backend.gear_motor_position, 0.0)[
+                Axis.P_L
```
Laura-Danielle: You should change Axis.P_L to -> `[Axis.of_main_tool_actuator(mount)]`. Gives us wiggle room in the off chance hardware decides to change what mount this can attach to.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -563,6 +563,8 @@ async def cache_instruments(
         async with self.instrument_cache_lock():
             await self._cache_instruments(require)
             await self._configure_instruments()
+        if self._gantry_load == GantryLoad.HIGH_THROUGHPUT:
```
Laura-Danielle: This check should only be in `def refresh_positions`.

api/src/opentrons/hardware_control/backends/ot3simulator.py
```diff
@@ -244,6 +247,18 @@ async def update_motor_status(self) -> None:
                 (node, MotorStatus(True, True)) for node in self._present_nodes
             )
 
+    @ensure_yield
+    async def update_gear_motor_position(self) -> None:
+        return None
+
+    @ensure_yield
+    async def home_gear_motors(
```
Laura-Danielle: We shouldn't need this function.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -626,21 +640,29 @@ def _filter_move_group(self, move_group: MoveGroup) -> MoveGroup:
...
+        velocity: Optional[float] = None,
         tip_action: str = "home",
     ) -> None:
-        if tip_action == "home":
-            speed = speed * -1
-        move_group = create_tip_action_group(
-            axes, distance, speed, cast(PipetteAction, tip_action)
+        move_group = []
+        if tip_action == "clamp":
+            assert moves is not None
```
Laura-Danielle: You should almost never do asserts in runtime code. It's very confusing when it raises as an error.

Here we should instead check that you have either moves or distance/velocity (with if checks). If it's distance/velocity then you need to modify the speed to be negative one.

Laura-Danielle (commented):
Some more comments.

Mainly, it looks like you aren't handling the multiple types of responses you can now get from the 96 channel and there is some general code structure cleanup requests.

You should also make this PR not a draft and add a PR description please once you've finished those requests.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -593,42 +597,45 @@ async def home(
...
         for position in positions:
             self._handle_motor_status_response(position)
         return axis_convert(self._position, 0.0)
 
     async def tip_action(
         self,
-        axes: Sequence[Axis],
-        distance: float,
-        speed: float,
+        moves: Optional[List[Move[Axis]]] = None,
```
ahiuchingau, line 627: What happens when this function is called without moves nor distance/velocity? 

Instead of making these three args optional, we should probably separate them into two functions, one requiring moves and the other both distance and velocity

ahiuchingau (changes requested):

hardware/opentrons_hardware/firmware_bindings/messages/message_definitions.py
```diff
@@ -272,6 +272,24 @@ class UpdateMotorPositionEstimationResponse(BaseMessage):  # noqa: D101
     ] = MessageId.update_motor_position_estimation_response
 
 
+@dataclass
+class UpdateGearMotorPositionEstimationRequest(EmptyPayloadMessage):  # noqa: D101
```
fsinapi: I really don't think we should have these message definitions. The point of the `UpdateMotorPositionEstimation` message for the other axes is to set the open loop stepper accumulator based on the encoder position. Since there's no encoder on the gear motors, there's no benefit to having a dedicated message for this.

If the goal is just to be able to read the motor position, we should rename this to something like `GearMotorPositionRequest`

api/src/opentrons/hardware_control/backends/ot3controller.py
Laura-Danielle: We could also just raise an error.

ahiuchingau: That's fine by me

sfoster1: Let's just raise an error for now but fix it Soon

fsinapi (changes requested):
We should definitely rename the `UpdateGearMotorPositionEstimation` command and make sure the implementation of it just pulls the current position.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -593,42 +597,60 @@ async def home(
...
+            new_group.append(
+                {
+                    node: axis_step
+                    for node, axis_step in step.items()
+                    if node in self._motor_nodes()
+                }
+            )
+        return new_group
+
     async def tip_action(
```
sfoster1, line 625: add a TODO to split this up pls

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -593,42 +597,60 @@ async def home(
...
         tip_action: str = "home",
+        back_off: Optional[bool] = False,
     ) -> None:
-        if tip_action == "home":
-            speed = speed * -1
-            runner = MoveGroupRunner(
-                move_groups=create_tip_action_home_group(axes, distance, speed)
+        move_group = []
+        # make sure either moves or distance and velocity is not None
+        assert any([moves, distance and velocity])
```
sfoster1: this allows all three to be set, which we don't want. you really want `bool(moves) xor (bool(distance) and bool(velocity))`

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -593,42 +597,60 @@ async def home(
...
+
+        runner = MoveGroupRunner(
+            move_groups=[move_group],
+            ignore_stalls=True if not ff.stall_detection_enabled() else False,
+        )
         positions = await runner.run(can_messenger=self._messenger)
-        for axis, point in positions.items():
-            self._position.update({axis: point[0]})
-            self._encoder_position.update({axis: point[1]})
+        if NodeId.pipette_left in positions:
```
sfoster1, line 651: if it isn't in positions then something's gone wrong, right? we should log that

sfoster1 (approved):
A couple nitpicks but really solid!

PR #12945, opened 2023-06-21
fix(protocol-designer): repair move labware step form

b-cooper:
# Overview

Fix bug in move labware step form via the LabwareLocationField


jerader (approved):
works as expected!

PR #12946, opened 2023-06-21
refactor(api): Update flex acc speeds based on EE recs

ahiuchingau:
# Overview
Fresh accelerations and speeds for FLEX gantry axes after some rigorous EE testing.
Note: gripper z axis will be updated in a later PR, as testing is still underway

On top of changing the values for the default accelerations and max speeds, EE also recommended that we use 30 mm/s as the max speed for each axis during calibration. So i added `set_system_constraints_for_calibration()` in `OT3Api` to update the system constraints in the move manager to lower the max speed, and a restore method `restore_system_constrants()` to revert it back once we're done calibrating. 

Motion setting changes requested by the EE team can be viewed [here](https://docs.google.com/presentation/d/1OlfuGUn_MCNfsENFJpg-UaeiAGT1ODlYUIoUoTEKhA8/edit#slide=id.g249d8e7cffe_0_0)


sfoster1 (approved):
Okay. I don't really love the specific method for calibration but we can check it later.

PR #12947, opened 2023-06-21
fix(shared-data): remove deprecated 'ot3' tip rack defs 

b-cooper:
# Overview

In favor of the new flex tip rack definitions, remove the old ot3 definitions

BREAKING CHANGE: all usage of "opentrons_ot3_96_tiprack_*" labware loadnames should be replaced by
"opentrons_flex_96_tiprack_*"


SyntaxColoring (approved):
✂️

Laura-Danielle (dismissed):
@b-cooper , to make this less of a headache we should keep a look-up table around for the old def names.

Happy to do this for you quickly if you're busy :)

b-cooper:
@Laura-Danielle That would be great if you could add that in. Thank you!

vegano1 (approved):
Nice

TamarZanzouri (approved):
tested on an OT-3. old names and new names work as expected.

PR #12948, opened 2023-06-21
fix(api): fix thread manager definition

pmoegenburg:
# Overview

Attempts to address the following issue:

`Hi, we are having the following issue when trying to run calibrate_belts() from ot3repl on DVT-10 robot with ot3@0.12.0-alpha.3:`
`>>> calibrate_belts(OT3Mount.LEFT, "P50SV3420230621A01")
  File "<console>", line 1, in <module>
  File "/opt/opentrons-robot-server/opentrons/hardware_control/scripts/repl.py", line 102, in synchronizer
    return asyncio.get_event_loop().run_until_complete(
  File "/usr/lib/python3.8/asyncio/events.py", line 639, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'MainThread'.`

# Test Plan

 - [x] verify multiple calls of `calibrate_belts` method call produce accurate results without above error

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
If this fixes things and passes tests, go for it!

PR #12949, opened 2023-06-21
feat(build): update the VERSION.json file whenever we manually push.

vegano1:
# Overview

It can get confusing knowing what version of each python package the flex has whenever we are developing, to keep things in sync this pr will update the `/etc/VERSION.json` file whenever a package is pushed to the robot.

# Test Plan

- [x] Make sure top-level and individual target push works on MacOS
- [x] Make sure top-level and individual target push works on Linux
- [x] Make sure top-level and individual target push works on Windows
- [x] Make sure local openembedded builds are not affected by this change
- [x] Make sure CI openembedded builds are not affected by this change
- [x] Make sure local buildroot builds are not affected by this change
- [x] Make sure CI buildroot builds are not affected by this change
- [x] Make sure we can still push to the OT2 without issues

# Changelog

- Added `sync-version-file` function to push.mk to update the VERSION.json file with a new entry
- Updated all packages to use new `sync-version-file` function
- Updated `scripts/python_build_utils.py` so it now gets the branch name if available

# Review requests

- Does this make sense?
- Anything that stands out?
- Do we want to enable this for the OT2 as well?

# Risk assessment
Med, 
- The `scripts/python_build_utils.py` script is used by both openembedded and buildroot build systems and needs to be tested
- The push.mk script is used as part of the development workflow and needs to be tested across all supported platforms


api/somefile.test
```diff
@@ -0,0 +1 @@
+{robot_type: OT-3 Standard, build_type: develop, openembedded_version: v0.2.4-15-gcb31ffd, openembedded_sha: cb31ffdb18882dd9db6642630a983a2a2f2b43d9, openembedded_branch: store-logs-better-changes, robot_server_version: 0.11.0, robot_server_sha: cb7629b87b7b1adbd04811de5142469644ad746a, robot_server_branch: edge, update_server_version: 0.11.0, update_server_sha: cb7629b87b7b1adbd04811de5142469644ad746a, update_server_branch: edge, system_server_version: 0.11.0, system_server_sha: cb7629b87b7b1adbd04811de5142469644ad746a, system_server_branch: edge, opentrons_api_version: 0.11.0.devopentrons, opentrons_api_sha: 9201f5743f08332b4103f578ad712f6f23c39013, opentrons_api_branch: edge, usb_bridge_version: 0.11.0, usb_bridge_sha: cb7629b87b7b1adbd04811de5142469644ad746a, usb_bridge_branch: edge, firmware_version: v18-2-g663266bb, firmware_sha: 663266bbf60e80c45b94c009b6a4ef170b5df975, firmware_branch: main}
```
sfoster1: accidental checkins?

notify-server/Makefile
```diff
@@ -133,7 +135,12 @@ push: wheel
 
 .PHONY: push-no-restart-ot3
 push-no-restart-ot3: sdist
-	$(call push-python-sdist,$(host),,$(ssh_opts),$(sdist_file),"/opt/opentrons-robot-server","notify_server")
+	$(call push-python-sdist,$(host),,$(ssh_opts),$(sdist_file),"/opt/opentrons-robot-server","notify_server",,,$(version_file))
+
+
+.PHONY: push-ot3
+push-ot3: push-no-restart-ot3
+	$(call restart-server,$(host),,$(ssh_opts),"opentrons-robot-server")
```
sfoster1, line 143: this wants the version file change too right?

scripts/python_build_utils.py
```diff
@@ -89,6 +89,9 @@ def _latest_version_for_project(project, git_dir):
     return prefix.join(tag.split(prefix)[1:])
 
 def _ref_from_sha(sha):
+    branch_name = subprocess.check_output(
```
sfoster1, line 95: see the comment below - this code is used in, among other places, the buildroot build process for the OT-2. that process runs inside aws codebuild, where we're in a detached-HEAD state and there is no branch information. you can simulate this by doing a `git checkout` of a commit hash. we should make sure that this function behaves sanely in this case

sfoster1 (changes requested):
Looks good overall but we have those extra files checked in and i'm a little nervous about the `python_build_utils` change - we sould make sure it doesn't break buildroot.

scripts/python_build_utils.py
vegano1: yep, that's why we are now checking for this case explicitly
if we are in a detached head then we won't have a branch name and then nothing changes
if we are not in a attached head branch_name will be populated, then we try and match it on line 115

notify-server/Makefile
vegano1: yeah this happens on `push-no-restart-ot3` (line 138)

sfoster1 (approved):
Looks great to me!

PR #12955, opened 2023-06-22
feat(protocol-designer, components, step-generation): update modules section to accommodate gripper

jerader:
closes RLIQ-416, RAUT-504, RAUT-506, RAUT-510

# Overview

For the Flex protocols, the modules section is updated to allow for adding/removing gripper and the deckmap and edit modules is updated.

Screen Shot 2023-06-22 at 9 29 40 AM

# Test Plan

Test creating a Flex protocol, modify the modules and gripper and ensure that it works properly. When exporting with a magnetic block, you shouldn't get an "unused module" modal. When exporting with an unused gripper, you sohuld get an "unused gripper" modal. 

Test creating an Ot-2 protocol, should work as normal.

# Changelog

- add `additionalEquipmentInvariantProperties` key to root state in redux
- create a `IS_GRIPPER_REQUIRED` type redux action (`toggleIsGripperRequired`), reducer and selector (`getAdditionalEquipment`)
- add the action to `CreateFileWizard` and `EditModulesCard` for adding/removing a gripper
- create a `gripperRow` to the Modules section and rename the section to `AdditionalItems` to encompass modules and gripper
- update the slotmap view for Flex and make sure it highlights the correct slots where the modules sit
- update the `FileSideBar` container to take in robot type and additional equipment for wiring up logic for unused gripper and unused module modal
- update tests

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/components/src/slotmap/SlotMap.tsx
+++ b/components/src/slotmap/SlotMap.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import cx from 'classnames'
-import { RobotType } from '@opentrons/shared-data'
+import { FLEX_ROBOT_TYPE, RobotType } from '@opentrons/shared-data'
 import { Icon } from '../icons'
 import styles from './styles.css'
 
@@ -40,7 +40,7 @@ const numCols = 3
 export function SlotMap(props: SlotMapProps): JSX.Element {
   const { collisionSlots, occupiedSlots, isError, robotType } = props
   const slots =
-    robotType === 'OT-3 Standard' ? FLEX_SLOT_MAP_SLOTS : OT2_SLOT_MAP_SLOTS
+    robotType === FLEX_ROBOT_TYPE ? FLEX_SLOT_MAP_SLOTS : OT2_SLOT_MAP_SLOTS
 
   return (
     <svg
--- a/components/src/slotmap/SlotMap.tsx
+++ b/components/src/slotmap/SlotMap.tsx
@@ -13,13 +13,21 @@ export interface SlotMapProps {
   collisionSlots?: string[]
   /** Optional error styling */
   isError?: boolean
+  isOt3?: boolean
 }
 
-const SLOT_MAP_SLOTS = [
+const OT2_SLOT_MAP_SLOTS = [
   ['10', '11'],
   ['7', '8', '9'],
   ['4', '5', '6'],
   ['1', '2', '3'],
+]
+
+const FLEX_SLOT_MAP_SLOTS = [
+  ['A1', 'A2', 'A3'],
+  ['B1', 'B2', 'B3'],
+  ['C1', 'C2', 'C3'],
+  ['D1', 'D2', 'D3'],
 ]
 
 const slotWidth = 33
@@ -29,12 +37,14 @@ const numRows = 4
 const numCols = 3
 
 export function SlotMap(props: SlotMapProps): JSX.Element {
-  const { collisionSlots, occupiedSlots, isError } = props
+  const { collisionSlots, occupiedSlots, isError, isOt3 } = props
+  const slots = isOt3 ? FLEX_SLOT_MAP_SLOTS : OT2_SLOT_MAP_SLOTS
+
   return (
     <svg
       viewBox={`-1,-1,${slotWidth * numCols + 2}, ${slotHeight * numRows + 2}`}
     >
-      {SLOT_MAP_SLOTS.flatMap((row, rowIndex) =>
+      {slots.flatMap((row, rowIndex) =>
         row.map((slot, colIndex) => {
           const isCollisionSlot =
             collisionSlots && collisionSlots.includes(slot)
--- a/components/src/slotmap/SlotMap.tsx
+++ b/components/src/slotmap/SlotMap.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import cx from 'classnames'
+import { RobotType } from '@opentrons/shared-data'
 import { Icon } from '../icons'
 import styles from './styles.css'
 
@@ -13,7 +14,7 @@ export interface SlotMapProps {
   collisionSlots?: string[]
   /** Optional error styling */
   isError?: boolean
-  isOt3?: boolean
+  robotType?: RobotType
 }
 
 const OT2_SLOT_MAP_SLOTS = [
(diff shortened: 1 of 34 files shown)


components/src/slotmap/SlotMap.tsx
```diff
@@ -13,28 +13,38 @@ export interface SlotMapProps {
   collisionSlots?: string[]
   /** Optional error styling */
   isError?: boolean
+  isOt3?: boolean
```
b-cooper: I'd prefer this branch on `RobotType`, but at the very least can we not refer to this as OT3 (use Flex instead)

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -432,7 +432,7 @@ export const DeckSetup = (): JSX.Element => {
   })
 
   return (
-    <React.Fragment>
+    <>
```
b-cooper: looks like this fragment can be deleted completely?

protocol-designer/src/components/FileSidebar/FileSidebar.tsx
```diff
@@ -174,27 +206,40 @@ export function FileSidebar(props: Props): JSX.Element {
       command => !LOAD_COMMANDS.includes(command.commandType)
     ) ?? []
 
+  const hasMoveLabware =
```
b-cooper: We also need to check that the move labware is using the gripper here. As written this will trigger for manual moves as well

protocol-designer/src/components/FileSidebar/FileSidebar.tsx
```diff
@@ -174,27 +206,40 @@ export function FileSidebar(props: Props): JSX.Element {
...
   )
   const modulesWithoutStep = getUnusedEntities(
     modulesOnDeck,
     savedStepForms,
-    'moduleId'
+    'moduleId',
+    robotType
   )
 
+  //  TODO(jr, 6/22/23): need to refactor when the gripper toggle is wired up
```
b-cooper: isn't the gripper toggle wired up?

protocol-designer/src/components/modals/EditModulesModal/ConnectedSlotMap.tsx
```diff
@@ -5,18 +5,20 @@ import styles from './EditModules.css'
 
 interface ConnectedSlotMapProps {
   fieldName: string
+  isOt3: boolean
```
b-cooper: same here, can we rename this prop, or make it `RobotType`?

protocol-designer/src/components/modals/EditModulesModal/index.tsx
```diff
@@ -278,15 +285,18 @@ const EditModulesModalComponent = (
...
                       disabled={!enableSlotSelection}
                       tabIndex={1}
                     />
                   </FormGroup>
                 </div>
 
-                <ConnectedSlotMap fieldName={'selectedSlot'} />
+                <ConnectedSlotMap
+                  fieldName="selectedSlot"
+                  isOt3={robotType === 'OT-3 Standard'}
```
b-cooper: import the `FLEX_ROBOT_TYPE` constant to compare against here

protocol-designer/src/components/modules/EditModulesCard.tsx
```diff
@@ -67,18 +75,22 @@ export function EditModulesCard(props: Props): JSX.Element {
     ].some(pipetteSpecs => pipetteSpecs?.channels !== 1)
 
   const warningsEnabled = !moduleRestrictionsDisabled
+  const isOt3 = robotType === 'OT-3 Standard'
```
b-cooper: use `FLEX_ROBOT_TYPE` constant

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -299,7 +299,9 @@ export const createFile: Selector<ProtocolFile> = createSelector(
     } => {
       const loadedPipettes = Object.values(pipettes)
       const pipetteGEN = loadedPipettes.some(
-        pipette => getPipetteNameSpecs(pipette.name)?.displayCategory === GEN3
+        pipette =>
```
b-cooper, line 304: This entire method should be replaced with the contents of `robotType` key on the protocol file instead of inferring it from pipette types. If you don't want to cram that into this PR, add a TODO

protocol-designer/src/step-forms/actions/additionalItems.ts
```diff
@@ -0,0 +1,7 @@
+export interface SetGripperAction {
+  type: 'IS_GRIPPER_REQUIRED'
```
b-cooper: The naming conventions here are a little wonky.  I know it seems redundant, but the action type/function/interface should all reflect the action verb itself.  E.g. here I would name them accordingly

```
interface ToggleIsGripperRequiredAction {
    type: 'TOGGLE_IS_GRIPPER_REQUIRED'
}

export const toggleIsGripperRequired ...
```

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -47,7 +47,10 @@ import { getLabwareOnModule } from '../../ui/modules/utils'
 import { nestedCombineReducers } from './nestedCombineReducers'
 import { PROFILE_CYCLE, PROFILE_STEP } from '../../form-types'
 import { Reducer } from 'redux'
-import { NormalizedPipetteById } from '@opentrons/step-generation'
+import {
+  NoramlizedAdditionalEquipmentById,
```
b-cooper: `NoramlizedAdditionalEquipmentById` should be `NormalizedAdditionalEquipmentById`

protocol-designer/src/components/FileSidebar/FileSidebar.tsx
jerader: oops ya idk why i thought it wasn't wired up 🤦

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1259,6 +1265,67 @@ export const pipetteInvariantProperties: Reducer<
...
+      const gripper = file.commands.filter(
+        command =>
+          // @ts-expect-error (jr, 6/22/23): moveLabware doesn't exist in schemav6
+          command.commandType === 'moveLabware' &&
+          // @ts-expect-error (jr, 6/22/23): moveLabware doesn't exist in schemav6
+          command.params.strategy === 'usingGripper'
+      )
+      const hasGripper = gripper.length > 0
+      // @ts-expect-error  (jr, 6/22/23): OT-3 Standard doesn't exist on schemav6
+      const isOt3 = file.robot.model === 'OT-3 Standard'
```
b-cooper: This const should be renamed, and we should be comparing against `FLEX_ROBOT_TYPE` instead of the bare string

b-cooper (approved):
🏁 🏎️

PR #12957, opened 2023-06-22
feat(api-client, app): add subsystem types and update flows

smb2268:
# Overview

Enable updating instrument firmware when attaching an instrument, handle possible subsystem response on instruments endpoint

# Test Plan
Attach an out of date instrument through an attachment flow, see that the firmware is updated before the successful attach screen is shown 

# Changelog
1. Add `BadPipette` and `BadGripper` types to shared-data, add `subsystem` and `ok` fields to Instruments type
2. Adapt code to ensure we're handling a not ok response for instruments
3. Add subsystem endpoint for initiating and viewing status of an update
4. Add update step to instrument attach flows after mount before results screen

# Review requests
Make sure code looks sound

# Risk assessment
Low


api-client/src/instruments/types.ts
```diff
@@ -13,6 +12,8 @@ export interface GripperData {
   instrumentType: 'gripper'
   mount: string
   serialNumber: string
+  subsystem: 'gripper'
+  ok: true
```
jerader, line 16: this key is not very descriptive 🙃

app/src/organisms/Devices/InstrumentsAndModules.tsx
```diff
@@ -142,8 +145,12 @@ export function InstrumentsAndModules({
                 }
                 isPipetteCalibrated={
                   isOT3
-                    ? attachedInstruments?.data?.find(i => i.mount === 'left')
-                        ?.data?.calibratedOffset != null
+                    ? attachedInstruments?.data?.find(
+                        (i): i is PipetteData =>
+                          i.instrumentType === 'pipette' &&
+                          i.ok &&
+                          i.mount === 'left'
```
jerader: it would read a bit clearer if you pulled this out into it's own `attachedPipette` const like you did with the `attachedGripper` - not necessary though, just a rec!

app/src/organisms/Devices/InstrumentsAndModules.tsx
```diff
@@ -188,7 +191,10 @@ export function InstrumentsAndModules({
                   isPipetteCalibrated={
                     isOT3
                       ? attachedInstruments?.data?.find(
-                          i => i.mount === 'right'
+                          (i): i is PipetteData =>
```
jerader: same here

react-api-client/src/subsystems/useSubsystemUpdateQuery.ts
```diff
@@ -0,0 +1,30 @@
...
+  const query = useQuery<SubsystemUpdateProgressData, TError>(
+    [host, 'subsystems/updates/all/', updateId],
+    () =>
+      getSubsystemUpdate(host as HostConfig, updateId as string).then(
+        response => response.data
+      ),
+    {
+      ...options,
+      enabled: updateId != null,
+      refetchInterval: 2000,
```
jerader, line 25: this is a really fast refetch interval?

jerader (approved):
not able to test this but it logic looks sound to me!

react-api-client/src/subsystems/useSubsystemUpdateQuery.ts
smb2268: yeah I talked to Brian about how quickly we should set this - we settled on 2sec bc it's linked to a progress bar!

api-client/src/instruments/types.ts
smb2268: 😅

PR #12960, opened 2023-06-22
feat(api): drop tips in predetermined locations

sanni-t:
Closes RQA-1007

# Overview

Changes the drop tip randomization to drop tips in predetermined locations that it will alternate between. I've selected these two locations based on coverage area and the ability of pipettes to reach that location:
1. location along x-axis from well top-center, at a distance closer to left edge of labware well
2. location along x-axis from well top-center, at a distance closer to right edge of labware well

See comments and docstrings in the PR for detailed descriptions of these locations

# Test Plan

Make sure the following tests succeed:
- [ ] A single/8- channel pipette on **LEFT** mount is able to go through the predetermined locations without hitting the trash or walls when:
    - [ ] trash is in column 1
    - [ ] trash is in column 3
- [ ] A single/8- channel pipette on **RIGHT** mount is able to go through the predetermined locations without hitting the trash or walls when:
    - [ ] trash is in column 1
    - [ ] trash is in column 3
- [ ] A 96-channel pipette is able to go through the predetermined locations without hitting the trash or walls when:
    - [ ] trash is in column 1
    - [ ] trash is in column 3

- [ ] Make sure that return_tip() behavior is not affected by this

Sample test protocol:
```
requirements = {
	"robotType": "Flex",
	"apiLevel": "2.15"
}

def run(protocol_context):
	# For Flex:
	tiprack1 = protocol_context.load_labware("opentrons_ot3_96_tiprack_1000ul", "D2")
	pipette = protocol_context.load_instrument("p1000_96", 'left', tip_racks=[tiprack1])
	# pipette = protocol_context.load_instrument("p50_single_gen3", "left", tip_racks=[tiprack1])

	# For OT-2
	# tiprack1 = protocol_context.load_labware("opentrons_96_tiprack_300ul", 2)
	# pipette = protocol_context.load_instrument("p300_single", "right", tip_racks=[tiprack1])

	for _ in range(10):
		pipette.pick_up_tip()
		pipette.drop_tip()
		
		# resets the tiprack so that we can use the same labware for testing w/ 96 channel
		# not needed if testing with other pipettes
		tiprack1.reset()	
		protocol_context.pause(msg="Reload the tips in the tiprack")

	pipette.pick_up_tip()
	pipette.drop_tip(pipette.trash_container['A1'].top())	# Should only go to trash center. `alternateDropLocation` should be false

```

For QA & ABR testing, we only need to test 1, 8 & multi-channel pipettes with the OT3 trash bin in slot A3. Placing the trash in the left column will currently not calculate labware position on deck correctly so we cannot use that for testing.

Should also be tested for OT2 but is not a blocker for internal QA/ science testing for flex.

# Changelog

- removed total randomized drop tip
- added next tip drop location getter to geometry view
- other changes to facilitate fetching the correct locations

# Review requests

- Code review
- Make sure the above points in test plan are tested and they cover all scenarios 
- I have tested all pipette combinations on OT3 for the fixed trash using engine commands and verified that the gantry or pipettes don't crash into anything anymore. We should also test running a python protocol that has a few tip drops.

# Risk assessment

- Less risky than dropping tip at total randomized location, which this replaces. There is still a decent risk that this would cause the pipette to go to an inaccessible location that could result in pipette/ tips bumping against the labware, but hopefully the check for fixed trash suffices for now.

### DIFF
--- a/api/src/opentrons/protocol_api/core/legacy_simulator/legacy_instrument_core.py
+++ b/api/src/opentrons/protocol_api/core/legacy_simulator/legacy_instrument_core.py
@@ -194,11 +194,11 @@ class LegacyInstrumentCoreSimulator(AbstractInstrument[LegacyWellCore]):
         location: Optional[types.Location],
         well_core: LegacyWellCore,
         home_after: Optional[bool],
-        randomize_drop_location: Optional[bool] = False,
+        alternate_drop_location: Optional[bool] = False,
     ) -> None:
-        if randomize_drop_location:
+        if alternate_drop_location:
             raise APIVersionError(
-                "Tip drop randomization is not supported in this API version."
+                "Tip drop alternation is not supported in this API version."
             )
         labware_core = well_core.geometry.parent
 
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -45,8 +45,8 @@ _PREP_AFTER_ADDED_IN = APIVersion(2, 13)
 """The version after which the pick-up tip procedure should also prepare the plunger."""
 _PRESSES_INCREMENT_REMOVED_IN = APIVersion(2, 14)
 """The version after which the pick-up tip procedure deprecates presses and increment arguments."""
-_DROP_TIP_LOCATION_RANDOMIZED_IN = APIVersion(2, 15)
-"""The version after which a drop-tip-into-trash procedure drops tips in a random location within the trash well."""
+_DROP_TIP_LOCATION_ALTERNATING_ADDED_IN = APIVersion(2, 15)
+"""The version after which a drop-tip-into-trash procedure drops tips in different alternating locations within the trash well."""
 
 
 class InstrumentContext(publisher.CommandPublisher):
@@ -837,8 +837,8 @@ class InstrumentContext(publisher.CommandPublisher):
 
         If no location is passed, the Pipette will drop the tip into its
         :py:attr:`trash_container`, which if not specified defaults to
-        the fixed trash in slot 12.  From API version 2.15 on, the exact position
-        where the pipette drops the tip(s) within the trash will be randomized
+        the fixed trash in slot 12.  From API version 2.15 on, the API will default to
+        alternating between two different drop tip locations within the trash container
         in order to prevent tips from piling up in a single location in the trash.
 
         The location in which to drop the tip can be manually specified with
@@ -902,11 +902,11 @@ class InstrumentContext(publisher.CommandPublisher):
 
         :returns: This instance
         """
-        randomize_drop_location: bool = False
+        alternate_drop_location: bool = False
         if location is None:
             well = self.trash_container.wells()[0]
-            if self.api_version >= _DROP_TIP_LOCATION_RANDOMIZED_IN:
-                randomize_drop_location = True
+            if self.api_version >= _DROP_TIP_LOCATION_ALTERNATING_ADDED_IN:
+                alternate_drop_location = True
 
         elif isinstance(location, labware.Well):
             well = location
@@ -942,7 +942,7 @@ class InstrumentContext(publisher.CommandPublisher):
                 location=location,
                 well_core=well._core,
                 home_after=home_after,
-                randomize_drop_location=randomize_drop_location,
+                alternate_drop_location=alternate_drop_location,
             )
 
         self._last_tip_picked_up_from = None
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -1,7 +1,9 @@
 """Geometry state getters."""
-from typing import Optional, List, Set, Tuple, Union
+import enum
+from typing import Optional, List, Set, Tuple, Union, cast
 
-from opentrons.types import Point, DeckSlotName
+from opentrons.types import Point, DeckSlotName, MountType
+from opentrons_shared_data.labware.constants import WELL_NAME_PATTERN
 
 from .. import errors
 from ..types import (
@@ -28,6 +30,19 @@ from .labware import LabwareView
 from .modules import ModuleView
 from .pipettes import PipetteView
 
+from opentrons_shared_data.pipette import PIPETTE_X_SPAN
+from opentrons_shared_data.pipette.dev_types import ChannelCount
+
+
+SLOT_WIDTH = 128
+
+
+class _TipDropSection(enum.Enum):
+    """Well sections to drop tips in."""
+
+    LEFT = "left"
+    RIGHT = "right"
+
 
 # TODO(mc, 2021-06-03): continue evaluation of which selectors should go here
 # vs which selectors should be in LabwareView
@@ -46,6 +61,7 @@ class GeometryView:
         self._labware = labware_view
         self._modules = module_view
         self._pipettes = pipette_view
+        self._last_drop_tip_location_spot: Optional[_TipDropSection] = None
 
     def get_labware_highest_z(self, labware_id: str) -> float:
         """Get the highest Z-point of a labware."""
@@ -325,13 +341,17 @@ class GeometryView:
             volume=int(well_def.totalLiquidVolume),
         )
 
-    def get_tip_drop_location(
+    def get_checked_tip_drop_location(
         self,
         pipette_id: str,
         labware_id: str,
         well_location: DropTipWellLocation,
     ) -> WellLocation:
-        """Get tip drop location given labware and hardware pipette."""
+        """Get tip drop location given labware and hardware pipette.
+
+        This makes sure that the well location has an appropriate origin & offset
+        if one is not already set previously.
+        """
         if well_location.origin != DropTipWellOrigin.DEFAULT:
             return WellLocation(
                 origin=WellOrigin(well_location.origin.value),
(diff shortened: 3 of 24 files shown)


api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -225,6 +221,7 @@ def __init__(self, state: LabwareState) -> None:
             state: Labware state dataclass used for all calculations.
         """
         self._state = state
+        self._last_drop_tip_location_index = 0
```
sanni-t:
```suggestion
```

shared-data/python/opentrons_shared_data/pipette/__init__.py
```diff
@@ -17,9 +17,24 @@
...
+The span of pipettes in X-direction based on number of channels.
+This is needed in order to determine safe tip drop location within a labware. 
+"""
+PIPETTE_X_SPAN: Dict[ChannelCount, float] = {
+    1: 75,
+    8: 75,
+    96: 161,
+}
+
+# PIPETTE_WELL_ACCESS_MARGIN:
```
sanni-t:
```suggestion
```

shared-data/python/opentrons_shared_data/pipette/__init__.py
```diff
@@ -17,9 +17,22 @@
         PipetteName,
         PipetteModel,
         PipetteFusedSpec,
+        ChannelCount,
     )
 
 
+# TODO (spp, 2023-06-22: should probably move this to definition)
```
sfoster1, line 24: the v2 pipette definitions either do currently or will soon (ish) have a labware-def-style list out of all the channels that should help

sfoster1 (commented):
Does this prevent tip stackup? It kind of seems like it would make the tip stackup just happen in two places.

shared-data/command/schemas/7.json
```diff
@@ -718,9 +718,9 @@
           "description": "Whether to home this pipette's plunger after dropping the tip. You should normally leave this unspecified to let the robot choose a safe default depending on its hardware.",
           "type": "boolean"
         },
-        "randomizeDropLocation": {
-          "title": "Randomizedroplocation",
-          "description": "Whether to randomize the location where tip is dropped within the labware. If True, this command will ignore the wellLocation provided and drop tip at a random location within a set area of the specified labware well. If False, the tip will be dropped at the top center of the well.",
+        "alternateDropLocation": {
+          "title": "Alternatedroplocation",
+          "description": "Whether to alternate location where tip is dropped within the labware. If True, this command will ignore the wellLocation provided and alternate between dropping tips at two predetermined locations inside the specified labware well. If False, the tip will be dropped at the top center of the well.",
```
SyntaxColoring, line 729: The fact that we alternate between two locations feels like an implementation detail that we should not document as part of the public interface. I think we want to leave room for changing this back to be random, or changing it to have more points, if we determine that it's necessary to stop tips from piling up.

api/src/opentrons/protocol_api/core/legacy_simulator/legacy_instrument_core.py
```diff
@@ -194,9 +194,9 @@ def drop_tip(
         location: Optional[types.Location],
         well_core: LegacyWellCore,
         home_after: Optional[bool],
-        randomize_drop_location: Optional[bool] = False,
+        alternate_drop_location: Optional[bool] = False,
     ) -> None:
-        if randomize_drop_location:
+        if alternate_drop_location:
             raise APIVersionError(
                 "Tip drop randomization is not supported in this API version."
```
SyntaxColoring:
```suggestion
                "Tip drop alternation is not supported in this API version."
```

I guess?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -837,8 +837,8 @@ def drop_tip(
 
         If no location is passed, the Pipette will drop the tip into its
         :py:attr:`trash_container`, which if not specified defaults to
-        the fixed trash in slot 12.  From API version 2.15 on, the exact position
-        where the pipette drops the tip(s) within the trash will be randomized
+        the fixed trash in slot 12.  From API version 2.15 on, the API will default to
+        alternating between two different drop tip locations within the trash container
```
SyntaxColoring, line 841: See other comment about not guaranteeing exactly two alternating drop locations.

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -27,6 +29,19 @@
 from .modules import ModuleView
 from .pipettes import PipetteView
 
+from opentrons_shared_data.pipette import PIPETTE_X_SPAN
+from opentrons_shared_data.pipette.dev_types import ChannelCount
+
+
+SLOT_WIDTH = 128
```
SyntaxColoring, line 37: Should we take this from the deck definition instead of hard-coding it?

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -406,3 +426,136 @@ def get_slot_item(
...
+    def _get_drop_tip_well_x_offset(
+        tip_drop_section: _TipDropSection,
+        well_x_dim: float,
+        pipette_channels: int,
+        pipette_mount: MountType,
+        labware_slot_column: int,
+    ) -> float:
+        """Get the well x offset for DropTipWellLocation."""
+        drop_location_margin_from_labware_edge = (
+            PIPETTE_X_SPAN[cast(ChannelCount, pipette_channels)] / 2
```
SyntaxColoring, line 584: Can you explain why `PIPETTE_X_SPAN` factors into these calculations?

If I understand correctly, `PIPETTE_X_SPAN` is the x-dimension of the bounding box around the entire pipette.

But, for example, the leftmost addressable x-coordinate of the pipette doesn't depend on that, does it? It depends on:

* How far -x the gantry can move
* The x-offset of the pipette's mount within the gantry
* The x-offsets of the pipette's nozzles from the pipette's mount

SyntaxColoring (commented):
Mostly looks good to me, but one question about the calculations.

api/src/opentrons/protocol_engine/state/geometry.py
SyntaxColoring: Actually, why do we need the slot width at all? Isn't this totally dependent on whether the trash well partially lies outside of the safely addressable X/Y rectangle?

I remember a problem a while back on the OT-2 where a single-channel pipette on the left mount couldn't touch-tip with wells in the rightmost column. So this isn't just specific to labware that are bigger than deck slots, either.

shared-data/command/schemas/7.json
sanni-t: Good point! Any suggestions for the param names? 'random' is no longer relevant since we actually don't want to use random locations. I had 'cycleDropLocations' at one point but wasn't sure if that made sense.

api/src/opentrons/protocol_engine/state/geometry.py
sanni-t: Oh!! I was assuming that all pipettes are able to access the full area of all deck slots and hence the different calculations based on whether a labware is bigger than the slot. I did not know about the issue on OT2. But I would still stick with this calculation because it does add a margin to the slot width so that the tip gets dropped slightly inside the labware's or the slot's outer edge. So this should still be fine unless the unreachable area for the OT2 left pipette is more than this margin (37.5mm) from the outer edge of the slot.

api/src/opentrons/protocol_engine/state/geometry.py
sanni-t: It is actually not the x-dimension of the bounding box of the pipette. That's why I avoided calling it the pipette mount width or anything similar to misinform what it is (although pipette x span is also not great, evidently lol). This value actually includes the bounding box + a margin if a pipette needs more than width/2 space from the edge of the slot to drop tips into the labware reliably. This is actually needed for even the ot3 to make sure that the mount assembly doesn't hit the robot walls even a tiny bit, and even if the pipette can actually drop tips to the edge of slot.

sanni-t:
@sfoster1 , true that it will probably create two piles of tips instead of one and will only mitigate the tip piling issue to a certain extent, but it will be an improvement. The main reason for deciding on two locations was that the 96 channel's tip piling would actually be best resolved when there's only two drop locations since the trash bin is slightly less than twice the width of the 96 channel. If we go for 3 locations, the fallen tips will overlap more number of times than if there were only 2 locations. And I was told the tip piling is worst for 96-ch. 
We could still have 2 locations only for 96 channel and more for the others but even with the other two pipettes, there's a dead zone in the trash that they can't reach (like the right half of the ot3 bin for left pipette). So the area for dropping tips reliably is much smaller than the big bin and we might be able to add maybe one or two more locations max.

So all the above things considered, I concluded we could start with a blanket 2 locations for all pipettes and then add more for pipettes and situations that make more sense. And @SyntaxColoring has a good point above considering that plan, that calling them 'alternate' or 'alternating' locations is not a good idea. So I'm open to suggestions for naming this feature.

jbleon95 (commented):
This was tested on Friday, June 23rd on Flex DVT hardware, with both single/multi channel pipettes on both mounts, as well as the 96 channel pipette. This was only tested with trash in slot 12/A3/right column.

sfoster1 (approved):
This looks great in structure. I would love to continue to bikeshed names ("cycling drop tip"? "spread drop tip"?) but I think alternate is fine as long as we're pronouncing it like the verb.

That said, there's still a bunch of comments and stuff that say randomize, or two, or whatever. I think we can clean those up later if you want though.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -45,7 +45,7 @@
 """The version after which the pick-up tip procedure should also prepare the plunger."""
 _PRESSES_INCREMENT_REMOVED_IN = APIVersion(2, 14)
 """The version after which the pick-up tip procedure deprecates presses and increment arguments."""
-_DROP_TIP_LOCATION_RANDOMIZED_IN = APIVersion(2, 15)
+_DROP_TIP_LOCATION_ALTERNATING_ADDED_IN = APIVersion(2, 15)
 """The version after which a drop-tip-into-trash procedure drops tips in a random location within the trash well."""
```
sanni-t:
```suggestion
"""The version after which a drop-tip-into-trash procedure drops tips in different alternating locations within the trash well."""
```

api/tests/opentrons/protocol_engine/commands/test_drop_tip.py
```diff
@@ -113,7 +113,7 @@ async def test_drop_tip_implementation(
     )
 
 
-async def test_drop_tip_with_randomization(
+async def test_drop_tip_with_location_cycling(
```
sanni-t:
```suggestion
async def test_drop_tip_with_alternating_locations(
```

PR #12961, opened 2023-06-22
feat: Add support for empty mounts in provisioning script

DerekMaggio:
# Overview

Add support for empty pipette mounts.

# Test Plan

- [x] Run configuration that has only a left pipette. Make sure that right pipette eeprom size is 0. Make sure left pipette eeprom size is not.

# Changelog

- Add logic to parse empty mount env var
- Add tests

# Review requests

None

# Risk assessment

Low


hardware/opentrons_hardware/scripts/emulation_pipette_provision.py
```diff
@@ -20,36 +19,81 @@
 RIGHT_PIPETTE_ENV_VAR_NAME = "RIGHT_OT3_PIPETTE_DEFINITION"
 
 
-@dataclass
 class OT3PipetteEnvVar:
```
Laura-Danielle, line 28: Couldn't you just keep this a data class with defaults?

Laura-Danielle: or keep it a dataclass and use a `build` classmethod

DerekMaggio: Stop making so much sense. I have no idea why I did it that way. I will go back and fix it. 

If I remember the reason I did it that way, and the reason makes sense, I will drop a comment.

DerekMaggio: I remembered my reasoning for doing it that way. 
My reasoning was not good. 

Here is the fix
262d21e41748787d14a2b013f6c4d5401c7a6267

fsinapi (approved):
LGTM!

PR #12962, opened 2023-06-22
fix(hardware): don't stop reading from CAN after most exceptions

fsinapi:
# Overview

The CAN messenger reader task will exit if it gets any exceptions. We actually want to keep reading unless the task is cancelled. There are a few exceptions that are normal to receive and we don't want to completely stop reading.

# Test Plan

# Changelog

# Review requests

# Risk assessment


hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
```diff
@@ -349,17 +349,17 @@ def remove_listener(self, listener: MessageListenerCallback) -> None:
...
+            try:
+                await self._read_task()
+            except asyncio.CancelledError:
+                return
+            except (CANCommunicationError, AsyncHardwareError, CanError) as e:
+                log.exception(f"Nonfatal error in CAN read task: {e}")
+                continue
+            except Exception as exc:
+                log.exception("Exception in read")
+                raise CANCommunicationError(exc=exc)
```
sfoster1: we probably don't want to raise this either, right?

sfoster1 (changes requested):
wait do we really want to raise the catchall? i think we'd be good with just doing the cancellation

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
fsinapi: yea true

PR #12963, opened 2023-06-22
 fix(hardware): don't stop reading from CAN after most exceptions

fsinapi:
# Overview

The CAN messenger reader task will exit if it gets any exceptions. We actually want to keep reading unless the task is cancelled. There are a few exceptions that are normal to receive and we don't want to completely stop reading.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Looks much better to me!

PR #12965, opened 2023-06-22
docs(protocol-engine): Document basic instance lifetime

SyntaxColoring:
# Overview

`ProtocolEngine` is unfortunately sensitive to how it's set up with threads and event loops. And this is largely tribal knowledge on the RSS team. We should document it.

Also fix type-checking failures from a merge error that I introduced in #13060.

# Review requests

* Is this more understandable than before?
* Is this surprising or incorrect?
* Is there anything else we want to add?

# Risk assessment

No risk.


SyntaxColoring:
Re-drafting while I work on this, from discussion with the team:

* Explain `SyncClient` and transports
* Maybe delete the abstract transport because it's not providing a useful abstraction for us today and it just makes the documentation noisier

jbleon95 (approved):
Comments make sense, looks good!

PR #12966, opened 2023-06-23
feat(shared-data): Support quirks, mount and mutable configs in v2 pipettes

Laura-Danielle:
# Overview

Support mutable configs and smoothie configs with the new v2 pipette schema.

# Test Plan

Throw this on an OT-2 robot and mess around with the pipette config settings to make sure everything still renders correctly.

- [ ] Remove config changes
- [ ] Add a new config change
- [ ] click and un-click quirks.

# Changelog

- Add quirks to schema
- Add mutable config helpers to shared data
- Utilize the new mutable configs in the robot server endpoint

The next step after this PR is to port the OT-2 to the new configs _and_ support loading the new configs in the protocol engine.

# Review requests

Please check it over and let me know what thoughts you have.

# Risk assessment

Medium. Touches core functionality on the OT-2

### DIFF
--- a/robot-server/robot_server/service/legacy/routers/settings.py
+++ b/robot-server/robot_server/service/legacy/routers/settings.py
@@ -315,7 +315,7 @@ async def patch_pipette_setting(
             raise LegacyErrorResponse(
                 message=str(e), errorCode=ErrorCodes.GENERAL_ERROR.value.code
             ).as_error(status.HTTP_412_PRECONDITION_FAILED)
-    r = _pipette_settings_from_config(pipette_config, pipette_id)
+    r = _pipette_settings_from_config(pipette_id)
     return r
 
 
--- a/robot-server/robot_server/service/legacy/routers/settings.py
+++ b/robot-server/robot_server/service/legacy/routers/settings.py
@@ -8,8 +8,8 @@ from fastapi import APIRouter, Depends
 from opentrons_shared_data.errors import ErrorCodes
 from opentrons.hardware_control import HardwareControlAPI
 from opentrons.system import log_control
+from opentrons_shared_data.pipette import mutable_configurations
 from opentrons.config import (
-    pipette_config,
     reset as reset_util,
     robot_configs,
     advanced_settings,
@@ -261,10 +261,10 @@ async def get_robot_settings(
 )
 async def get_pipette_settings() -> MultiPipetteSettings:
     res = {}
-    for pipette_id in pipette_config.known_pipettes():
+    for pipette_id in mutable_configurations.known_pipettes():
         # Have to convert to dict using by_alias due to bug in fastapi
         res[pipette_id] = _pipette_settings_from_config(
-            pipette_config,
+            mutable_configurations,
             pipette_id,
         )
     return res
@@ -281,7 +281,7 @@ async def get_pipette_settings() -> MultiPipetteSettings:
     },
 )
 async def get_pipette_setting(pipette_id: str) -> PipetteSettings:
-    if pipette_id not in pipette_config.known_pipettes():
+    if pipette_id not in mutable_configurations.known_pipettes():
         raise LegacyErrorResponse(
             message=f"{pipette_id} is not a valid pipette id",
             errorCode=ErrorCodes.PIPETTE_NOT_PRESENT.value.code,
@@ -309,7 +309,7 @@ async def patch_pipette_setting(
     field_values = {k: None if v is None else v.value for k, v in fields.items()}
     if field_values:
         try:
-            pipette_config.override(fields=field_values, pipette_id=pipette_id)
+            mutable_configurations.save_overrides(fields=field_values, pipette_id=pipette_id)
         except ValueError as e:
             raise LegacyErrorResponse(
                 message=str(e), errorCode=ErrorCodes.GENERAL_ERROR.value.code
@@ -328,9 +328,9 @@ def _pipette_settings_from_config(pc, pipette_id: str) -> PipetteSettings:
     """
     mutuble_configs = pc.list_mutable_configs(pipette_id=pipette_id)
     fields = PipetteSettingsFields(**{k: v for k, v in mutuble_configs.items()})
-    c, m = pc.load_config_dict(pipette_id)
 
     # TODO(mc, 2020-09-17): s/fields/setting_fields (?)
+    # need model and name?
     return PipetteSettings(  # type: ignore[call-arg]
         info=PipetteSettingsInfo(name=c.get("name"), model=m), fields=fields
     )
--- a/robot-server/robot_server/service/legacy/routers/settings.py
+++ b/robot-server/robot_server/service/legacy/routers/settings.py
@@ -1,6 +1,6 @@
 from dataclasses import asdict
 import logging
-from typing import Dict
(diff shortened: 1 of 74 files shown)


sfoster1 (changes requested):
I don't think I agree with putting the quirks in code like this. It means if we ever want to add some new model we have to add it in conceptually separate places. If we can drop it in configs then we're back to the same system and I think we're fine.

shared-data/python/opentrons_shared_data/pipette/file_operation_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+    When this module is imported, this function is called automatically
+    and the result stored in :py:attr:`APP_DATA_DIR`.
+
+    This directory may not exist when the module is imported. Even if it
+    does exist, it may not contain data, or may require data to be moved
+    to it.
+
+    :return pathlib.Path: The path to the desired root settings dir.
+    """
+    IS_ROBOT = bool(
```
sfoster1: I don't think it's a good idea to duplicate this logic here - it's already complicated and messy and I don't want to have to remember an extra place to put it. I think that this function shouldn't exist and the things that call it should instead take a base dir as an argument. Yes, that makes the calls messier, but if we can provide a consistent place to get that information - i.e., `opentrons.config` - then it'll be okay enough that it's worth not duplicating.

robot-server/robot_server/service/legacy/routers/settings.py
```diff
@@ -309,28 +308,45 @@ async def patch_pipette_setting(
...
+        if isinstance(v, str):
+            converted_dict[k] = v
+        elif isinstance(v, pip_types.MutableConfig):
+            converted_dict[k] = v.dict_for_encode()
+        elif k == "quirks":
+            converted_dict[k] = {q: b.dict_for_encode() for q, b in v.items()}
+    fields = PipetteSettingsFields(**converted_dict)  # type: ignore
 
     # TODO(mc, 2020-09-17): s/fields/setting_fields (?)
+    # need model and name?
```
sfoster1, line 356: won't this break without it?

shared-data/python/opentrons_shared_data/pipette/load_data.py
```diff
@@ -65,6 +68,34 @@ def _physical(
     return _get_configuration_dictionary("general", channels, max_volume, version)
 
 
+@lru_cache(maxsize=None)
+def load_serial_lookup_table() -> Dict[str, str]:
```
sfoster1, line 72: Oof, but what are you gonna do

shared-data/python/opentrons_shared_data/pipette/mutable_configurations.py
```diff
@@ -0,0 +1,309 @@
...
+    return base_configurations
+
+
+def _add_new_overrides_to_existing(
+    base_configs_dict: Dict[str, Any],
+    existing_overrides: OverrideType,
+    overrides: TypeOverrides,
+) -> OverrideType:
+    """Helper function to add new overrides to the existing ones"""
+    # FIXME remove the validation here for the file save and rely
```
sfoster1, line 272: i guess we can do this by passing down a full pydantic model at this point?

sfoster1 (changes requested):
This overall looks, well, like something we have to do, but I really want to avoid duplicating that `IS_ROBOT` bool - I think that will eternally bite us if we duplicate it.

robot-server/robot_server/service/legacy/routers/settings.py
Laura-Danielle: Not sure which is why I want to test on a robot. I don't know why it _would_ break because the app only deals with the model for this feature.

shared-data/python/opentrons_shared_data/pipette/load_data.py
Laura-Danielle: big oof.

shared-data/python/opentrons_shared_data/pipette/file_operation_helpers.py
Laura-Danielle: big oof.

robot-server/robot_server/service/legacy/routers/settings.py
sfoster1: Can we add it anyway? Regardless of whether it's _needed_, this is in theory a controlled API and this would represent a change.

robot-server/tests/integration/test_settings_pipettes.tavern.yaml
```diff
@@ -1,38 +1,43 @@
...
-            top: !anydict
-          info:
-            model: p300_multi_v1
-            name: p300_multi
-      strict: false 
----
+# Temporarily remove this test. I realized that tavern is not
+# actually using the temporary directory path provided by the pytest
+# fixtures because it's technically not running in the same environment
+# I tried to timebox this and couldn't come up with a good solution.
```
sfoster1: What we have done in other cases is make a separate fixture used by tavern. I think we really want this test in before we merge.

robot-server/tests/integration/test_settings_pipettes.tavern.yaml
```diff
@@ -41,7 +46,7 @@ marks:
 stages:
```
sfoster1, line 41: Why did all these values change? The same issue with directories not being respected?

sfoster1 (changes requested):
Okaym, I really like the changes to get the base directory passed in and decided elsewhere, the lack of duplication is great. I think the last thing I'd really want to see is fixes to the HTTP API and the tavern tests. Those tests are for detecting regressions, and I think that this PR introduces changes both of value and of kind (not having names) that we really need to investigate and probably fix.

robot-server/tests/integration/test_settings_pipettes.tavern.yaml
Laura-Danielle: So that fixture actually doesn't work across tests. It just so happens to be OK for the other tavern tests even though they are accessing the file system locally.

For this test specifically, the folder it's checking for already has a lot of junk (including pipette serial numbers that would never exist in the folder). I could raise an error in this case, rather than a KeyError, but the test would still fail all the same.

So essentially all of these tavern tests relying on a tmp dir aren't actually doing that. I tried figuring out a way to manipulate the env variable for tavern and couldn't successfully do that. The ultimate problem is that tavern does not see the modifications to the environment variables from pytest fixtures so even though we're adding OT_CONFIG_DIR in the pytest fixture, it doesn't exist at the time of the `GET` request.

robot-server/tests/integration/test_settings_pipettes.tavern.yaml
Laura-Danielle: I changed the ID to be a realistic ID since we have to look up the configs by valid pipette names rather than the random test names we've been using.

sfoster1 (approved):
Awesome! Looks really good

sanni-t (approved):
Tested on an OT-2 and confirmed that pipette settings & quirks changes from app are propagated correctly to the settings files on the robot and the app reads file contents correctly. Also verified that all changes are correctly reset to default with 'Reset All'.

PR #12968, opened 2023-06-23
feat(protocol-designer, step-generation, components): pipette edit, moveLabware, ui changes

jerader:
closes RAUT-512 RAUT-519 RAUT-505 RAUT-509 RAUT-523

# Overview

Various things outlined in Changelog below. Basically fixing up the UI for the Create File Wizard, modifying tiprack and pipette selections, adding back custom tiprack support and adding some timeline errors to `moveLabware` when moving a labware out of a TC or H-S when the lid/latch is closed/is shaking. Additionally, I added an experimental setting to allow users the option to select any tiprack for their pipette.

# Test Plan

Should smoke test this. Focus on these sections:
1. create file wizard - look at the tiprack section, add a custom tiprack, etc. 
2. edit pipette section - make sure the pipettes are correct for the flex
3. `moveLabware` - make sure errors show up when expected and the dropdowns work/display correct info and `offDeck` slot option is appearing as expected
4. toggle on the "Allow all tiprack" setting and make sure you can select which ever tiprack you want

# Changelog

- fixes devTools warnings throughout `CreateFileWizard` by adding some `React.useEffects` to update formik
- allow the Flex module edit section to only allow users to select GEN2 TC and GEN2 Temp, rather than allowing the GEN1 options as well
- update the Pipette edit section to show correct pipettes for each robot and show only default tipracks.
- update the `CreateFileWizard` tiprack sections to have custom tiprack support
- various UI changes to the `CreateFileWizard` to make it look nicer with consistent spacing/font size
- adjust the OT-2 deck map view to display the map + module tags clearer
- add the LED lighting to the temperature and H-s modules upon heating and cooling
- changes to the `moveLabwareForm`'s location dropdown so the OT-2 TC says "slot 7, 8, 10, 11" instead of the `slot_7_8_10_11` string and allow an `offDeck` location only when it is a manual movement
- create a `errorCreator`: `heaterShakeLatchClosed` and add logic to `moveLabware` so you get a timeline error if you try to move a labware out of the H-S or TC when the latch/lid is closed or the H-S is shaking
- added an advanced setting under `Experimental Settings` for allowing all tiprack selection. When toggled on, and on an OT-2, users can select any OT-2 tip + 3rd party tip in our labware library (like TipOne). For a Flex, users can select any of the flex tipracks

# Review requests

see test plan

# Risk assessment

low


components/src/forms/Select.tsx
```diff
@@ -113,13 +117,18 @@ function DropdownIndicator(
...
-        })}
+      <Box
+        position={POSITION_ABSOLUTE}
+        top="0.55rem"
+        right={SPACING.spacing8}
+        width={SPACING.spacing20}
       >
-        <Icon name="menu-down" className={cx(styles.dropdown_indicator_icon)} />
-      </div>
+        {Boolean(props.selectProps.menuIsOpen) ? (
```
b-cooper: awesome, it looks like we could move this ternary inside of the transform e.g.
```
transform={`rotate(${props.selectProps.menuIsOpen === true ? '180' : '0' })`} 
```

components/src/hardware-sim/Module/index.tsx
```diff
@@ -158,6 +164,9 @@ export const Module = (props: Props): JSX.Element => {
       ? nestedLabwareOffsetY
       : 0
   // transform to be applied to children which render within the labware interfacing surface of the module
+  //  TODO(jr, 6/26/23): labware looks off in Flex PD when on top of the thermocycler: maybe need
+  //  {moduleType == 'thermocyclerModuleType' && robotType === FLEX_ROBOT_TYPE ? clampedLabwareOffsetY + 17
```
b-cooper: I've noticed that offset, let's investigate it at the definition level instead of arbitrarily adding 17mm. Some aspect of the geometry system is not being accurately represented with these compounded offsets.

components/src/hardware-sim/Module/index.tsx
```diff
@@ -182,6 +191,46 @@ export const Module = (props: Props): JSX.Element => {
   }
 
   const magneticBlockLayerBlockList = ['Module_Title', 'Well_Labels', 'Wells']
+  let ledLightOverlay: JSX.Element | null = null
```
b-cooper: This approach looks great.  It does diverge a little bit from the existing pattern that we use for the Thermocycler. In that case, we do the specialized branching at a lower level via the `innerProps` prop.

components/src/instrument/InstrumentInfo.tsx
```diff
@@ -30,13 +33,21 @@ export interface InstrumentInfoProps {
...
-  const className = cx(
+  const { robotType = 'OT-2 Standard' } = props
+  const baseClassname = cx(
     styles.pipette,
-    styles[props.mount],
     { [styles.disabled]: props.isDisabled },
     props.className
   )
 
+  const mountStyle = props.mount === LEFT ? styles.flex_left : styles.right
```
b-cooper: It looks like the only reason we're pulling in robot type here is to add a bit of margin for one of the pipettes. This feels like roundabout way to get to bit of custom layout. This looks like a component that could be updated to extend StyleProps instead of dealing with CSS modules.  That way the layout concerns like "margin" can be established clearly in the instance of this component.

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -84,9 +84,9 @@ type ContentsProps = RobotWorkSpaceRenderProps & {
 }
 
 export const VIEWBOX_MIN_X = -64
-export const VIEWBOX_MIN_Y = -10
+export const VIEWBOX_MIN_Y = -30
```
b-cooper: Why did these values get updated? Was there a problem with the OT2 deck map size?

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -261,6 +261,15 @@ export const DeckSetupContents = (props: ContentsProps): JSX.Element => {
           },
           compatibleModules: [THERMOCYCLER_MODULE_TYPE],
         }
+        let targetTemp: number | null = null
```
b-cooper: These additions should probably go inside the `getModuleInnerProps` function above

protocol-designer/src/components/SettingsPage/FeatureFlagCard/FeatureFlagCard.tsx
```diff
@@ -33,6 +33,11 @@ export const FeatureFlagCard = (props: Props): JSX.Element => {
 
   const getDescription = (flag: FlagTypes): JSX.Element => {
     const RICH_DESCRIPTIONS: Partial<Record<FlagTypes, JSX.Element>> = {
+      ALLOW_ALL_TIPRACKS: (
```
b-cooper: Why did you choose for this to be a feature flag?

jerader: I added it to the list of `Experimental Settings` which aren't feature flags but for some reason are lumped in with the feature flags.

jerader: Shlok had some explanation why `Experimental Settings` is under feature flags but can't remember what he said

protocol-designer/src/components/DeckSetup/index.tsx
jerader: The module tag gets cut off for me for modules in slot 1 or 3 otherwise. There are more issues with the module tags though because if the module is in slot 4, 6, 7, 9, and there is labware in the slot below it, the labware covers up the module tag. Perhaps we can change where the module tag is for ot-2

Screen Shot 2023-06-26 at 3 31 40 PM

components/src/hardware-sim/Module/index.tsx
jerader: cool, i'll make a separate ticket to look into this and i'll delete this comment!

protocol-designer/src/components/DeckSetup/index.tsx
b-cooper: Oh yeah, good call. Maybe we can bump the tags up a few mm on the OT2? So they mostly only cover the slot ridge?

jerader: I can delete this change in this pr: https://github.com/Opentrons/opentrons/pull/12993

components/src/hardware-sim/Module/ModuleFromDef.tsx
```diff
@@ -29,12 +29,14 @@ export function ModuleFromDef(props: ModuleFromDataProps): JSX.Element {
...
-    </g>
+    <>
+      <g>
+        {parseHtml(
+          stringify(standaloneSVG ? filteredSVGWrapper : groupNodeWrapper, {
+            selfClose: false,
+          })
+        )}
+      </g>
+    </>
```
b-cooper:
```suggestion
    <g>
      {parseHtml(
        stringify(standaloneSVG ? filteredSVGWrapper : groupNodeWrapper, {
          selfClose: false,
        })
      )}
    </g>
```

components/src/hardware-sim/Module/index.tsx
```diff
@@ -183,25 +186,41 @@ export const Module = (props: Props): JSX.Element => {
 
   const magneticBlockLayerBlockList = ['Module_Title', 'Well_Labels', 'Wells']
 
+  let moduleViz: JSX.Element = (
```
b-cooper, line 191: Let's update the type of the `innerProps` union above in this file to include the props of these two new components

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -84,9 +84,9 @@ type ContentsProps = RobotWorkSpaceRenderProps & {
 }
 
 export const VIEWBOX_MIN_X = -64
-export const VIEWBOX_MIN_Y = -10
+export const VIEWBOX_MIN_Y = -30
 export const VIEWBOX_WIDTH = 520
-export const VIEWBOX_HEIGHT = 414
+export const VIEWBOX_HEIGHT = 480
```
b-cooper: does this height still need to be changed now that we're addressing the module tag rendering on a per-platform basis?

jerader: nah, i'll change it back!

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
```diff
@@ -205,7 +205,7 @@ export function CreateFileWizard(): JSX.Element | null {
     <WizardHeader
       title={t('modal.create_file_wizard.create_new_protocol')}
       currentStep={currentStepIndex}
-      totalSteps={WIZARD_STEPS.length}
+      totalSteps={WIZARD_STEPS.length - 1}
```
b-cooper, line 208: 🥇

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
```diff
@@ -384,10 +384,14 @@ function CreateFileForm(props: CreateFileFormProps): JSX.Element {
           formikProps.setFieldTouched(name, true)
         }
 
-        return contentsByWizardStep[currentWizardStep]({
-          ...formikProps,
-          handleChange,
-        })
+        return currentWizardStep === 'metadata'
```
b-cooper, line 387: why is this ternary necessary?

jerader: the `handleChange` logic is different for the `metadata` step only (to allow you to hit the Next cta without having to click out of the input field). That logic doesn't work for selecting/deselecting modules on the OT-2 form though, which is why i added this ternary

protocol-designer/src/components/modals/EditModulesModal/index.tsx
```diff
@@ -254,6 +258,23 @@ const EditModulesModalComponent = (
     placement: 'top',
   })
 
+  //  TODO(jr, 6/26/23): should probably move this into a util component
+  function getFlexModuleOptions(
```
b-cooper: this function is actually more like `getModuleOptionsForRobotType` right?

protocol-designer/src/components/modals/FilePipettesModal/PipetteFields.tsx
```diff
@@ -82,32 +85,19 @@ export function PipetteFields(props: Props): JSX.Element {
...
-  )
-
   const initialTabIndex = props.initialTabIndex || 1
 
   const renderPipetteSelect = (props: PipetteSelectProps): JSX.Element => {
     const { tabIndex, mount } = props
     const pipetteName = values[mount].pipetteName
+
+    //  adding 96-channel to the nameBlockstlist for the Flex for now
+    OT2_PIPETTES.push('p1000_96')
```
b-cooper: Let's try to avoid mutating arrays in place.

instead 
```
const nameBlockList = [...OT2_PIPETTES, 'p1000_96']
```

protocol-designer/src/components/modals/utils.ts
```diff
@@ -0,0 +1,58 @@
+import {
+  getLabwareDefURI,
+  getLabwareDisplayName,
+  getPipetteNameSpecs,
+  PipetteName,
+} from '@opentrons/shared-data'
+import { LabwareDefByDefURI } from '../../labware-defs'
+
+interface tiprackOption {
```
b-cooper: `PascalCase` for interfaces

protocol-designer/src/feature-flags/reducers.ts
```diff
@@ -20,6 +20,7 @@ const initialFlags: Flags = {
   OT_PD_DISABLE_MODULE_RESTRICTIONS:
     process.env.OT_PD_DISABLE_MODULE_RESTRICTIONS === '1' || false,
   OT_PD_ENABLE_OT_3: process.env.OT_PD_ENABLE_OT_3 === '1' || false,
+  ALLOW_ALL_TIPRACKS: process.env.ALLOW_ALL_TIPRACKS === '1' || false,
```
b-cooper: We should probably stick with the `OT_PD_{NAME_OF_SETTING}`  convention unless we have a good reason to break it?

b-cooper (approved):
Sprinkled a few comments in there, but this is a huge step forward 👏  way to go!

PR #12936, opened 2023-06-16
feat(api,shared-data): error codes in PE

sfoster1:
## Draft alert!
This is a draft because I definitely need to get some fixes and further tests in the HTTP API layer and test on a robot and so on and so forth, but I wanted people to be able to see it since this does some medium-dubious jsonschema stuff and pokes into some core PE behavior.

## Overview
This PR adds our new enumerated error codes to ProtocolEngine. This is the first place we're using them. We add the error codes to the protocol engine internal custom error hierarchies and add logic to capture python exceptions that are not `EnumeratedError`s at the three places that we catch exceptions - the core engine action handler, the legacy executor, and the main executor.

The biggest new functionality is error wrapping. The `EnumeratedError`s can now catch, wrap, and store sequences of `raise from` exception contexts that could previously only really be represented as tracebacks. Now the structure of exception raise sequences is preserved all the way up to the `ErrorOccurrence` model and thus to the client to format as they see fit. While some detail isn't really schema controlled (you'll have a `Dict[str, str]` and you'll like it!) the structure actually is, as it's just a tree and pydantic and jsonschema do a pretty good job of supporting type recursion in models.

I'm interested to see early reactions from @sanni-t @jbleon95 @TamarZanzouri and @SyntaxColoring ; I think I'm happy with how this turned out, but further testing will tell.

## Scope and Followups

This PR is intended to add the enumerated error code handling and transformations to PD and correctly send them over HTTP. There's a followup PR needed to scrape through PE and make sure that everything has a correct error code, and that specific hardware calls that might fail in specific ways and already have specific `try`/`except` clauses correctly wrap things in the right error codes and so on. There's also one needed to do that same thing in the other parts of the `api` subdirectory, particularly `hardware_control`, and hardware; and finally, something similar's going to need to happen in the `robot-server`, along with some kind of top-level exception catching there.

## Risks
Well, this modifies serialized data, so a lot, but hopefully in as safe a way as we can do it. It shouldn't introduce new errors, just transform how they're handled.

## Testing
- [x] Put it on a robot and cause a bunch of protocol errors!

PR #7783, opened 2021-05-10
refactor(robot-server): consolidate and simplify error response logic

mcous:
## Overview

This PR is a bit of an RFC, and comes out of a rabbit hole I fell into while trying to build some new endpoints.

In a lot of discussions we've had about error responses from our API, there's been a desire to have a concrete list of all errors that may be returned by the HTTP API. Ideally, this is something we can bake right into our OpenAPI Spec, but there were a couple things in the way of that effort:

- Our error models did not require the `id` field
- It was difficult to spec a specific error return from a given endpoint
- Returning an error from an endpoint was a little hard to read (my opinion)

This PR reorganizes and simplifies our models to allow for this sort of usage:

```py
from fastapi import APIRouter, status
from typing_extensions import Literal
from robot_server.errors import ErrorResponse, ErrorDetails


class SpecificProblem(ErrorDetails):
    id: Literal["SpecificProblem"] = "SpecificProblem"
    title: str = "Specific Problem"


class SomeModel(BaseModel):
    id: str
    name: str


router = APIRouter()


@router.get(
    path="/some-model",
    response_model=SomeModel,
    responses={
        status.HTTP_400_BAD_REQUEST: {"model": ErrorResponse[SpecificProblem]},
    },
)
def get_some_model():
    """Get some model."""
    # ...
    raise SpecificProblem(detail="A specific problem occurred").as_error(
        status.HTTP_400_BAD_REQUEST
    )
```

If we successfully set `responses` for all of our endpoints in this way, we will have an exhaustive list of every specific error response returned baked right into the OpenAPI Spec, without (necessarily) needing to maintain some giant, central `ErrorId` enum or similar.

## Changelog

Some top-of-line changes:

- Moved as much as possible related to error responses to `robot_server.errors`
- Removed usage of `V1HandlerError` in favor of `robot_server.errors.LegacyErrorResponse`
- Simplified the handler interface for returning an error response to flow directly from the response models themselves
- Genericized `ErrorResponse` so JSON schema knows what specific error is coming back
    - Removed redundant `status` field from `ErrorDetails`
    - Moved `links` field from `ErrorDetails` to `ErrorResponse` for parity with our regular responses
        - Good response: `{ "data": { ... }, "links": { ... } }`, error response: `{ "errors": [ ... ], "links": { ... } }`
- Fixed up a few things about our exception handlers
    - Legacy endpoints now return legacy errors during unhandled exceptions
[...]

PR #12969, opened 2023-06-23
feat(robot-server): Present error codes in error responses

sfoster1:
This PR does the equivalent of #12936 for the non-protocol-engine robot server error responses - both the `LegacyErrorResponse` (which now has an `errorCode` entry as well as a `message`, but otherwise not a whole lot else) and the `ErrorResponse` and `MultiErrorResponse`, which now have a top level `errorCode` and also properly intern the exception chain in their multiple-error-body responses.

We also go through and add a couple new useful errors, which is a process that will continue.

## Review Requests
- Does the way we treat `MultipleErrorResponse` and `ErrorResponse` make sense? What's the point of the multi error response anyway, given that error response can hold multiple errors?

## Testing
- [x] Make sure that in practice, errors from both legacy and non-legacy routes that don't have anything to do with protocol engine send back errors properly

## Next steps
Similarly to #12936 , now that the catchalls and structure is in place we need to go through and make sure that the exceptions that we raise to start any given error response off have error codes and tuned messages.

### DIFF
--- a/robot-server/robot_server/errors/error_responses.py
+++ b/robot-server/robot_server/errors/error_responses.py
@@ -113,11 +113,7 @@ class ErrorDetails(BaseErrorBody):
 
     @classmethod
     def from_exc(
-        cls: Type["ErrorDetailsT"],
-        exc: BaseException,
-        *,
-        override_defaults: bool = False,
-        **supplemental_kwargs: Any
+        cls: Type["ErrorDetailsT"], exc: BaseException, **supplemental_kwargs: Any
     ) -> "ErrorDetailsT":
         """Build an ErrorDetails model from an exception.
 
@@ -143,10 +139,6 @@ class ErrorDetails(BaseErrorBody):
             }
 
         values["meta"] = _exc_to_meta(checked_exc)
-        if not override_defaults:
-            for fieldname, fieldval in cls.__fields__.items():
-                if not fieldval.required and fieldval.default is not None:
-                    values.pop(fieldname, None)
         return cls(**values)
 
     def as_error(self, status_code: int) -> ApiError:
--- a/robot-server/robot_server/errors/error_responses.py
+++ b/robot-server/robot_server/errors/error_responses.py
@@ -4,9 +4,10 @@ from pydantic.generics import GenericModel
 from typing import Any, Dict, Generic, Optional, Sequence, Tuple, TypeVar
 
 from robot_server.service.json_api import BaseResponseBody, ResourceLinks
+from opentrons_shared_data.errors import ErrorCodes, EnumeratedError
 
 
-class ApiError(Exception):
+class ApiError(EnumeratedError):
     """An exception to throw when an endpoint should respond with an error."""
 
     def __init__(self, status_code: int, content: Dict[str, Any]) -> None:
@@ -105,6 +106,10 @@ class ErrorDetails(BaseErrorBody):
             "occurrence of the error"
         ),
     )
+    errorCode: str = Field(
+        ...,
+        description=("The Opentrons error code associated with the error"),
+    )
 
     def as_error(self, status_code: int) -> ApiError:
         """Serial this ErrorDetails as an ApiError from an ErrorResponse."""
@@ -120,6 +125,9 @@ class LegacyErrorResponse(BaseErrorBody):
     message: str = Field(
         ...,
         description="A human-readable message describing the error.",
+    )
+    errorCode: str = Field(
+        ..., description="The Opentrons error code associated with the error"
     )
 
 
--- a/robot-server/robot_server/errors/error_responses.py
+++ b/robot-server/robot_server/errors/error_responses.py
@@ -1,13 +1,13 @@
 """JSON API errors and response models."""
 from pydantic import BaseModel, Field
 from pydantic.generics import GenericModel
-from typing import Any, Dict, Generic, Optional, Sequence, Tuple, TypeVar
+from typing import Any, Dict, Generic, Optional, Sequence, TypeVar, Type
 
 from robot_server.service.json_api import BaseResponseBody, ResourceLinks
-from opentrons_shared_data.errors import ErrorCodes, EnumeratedError
+from opentrons_shared_data.errors import EnumeratedError, PythonException
 
 
-class ApiError(EnumeratedError):
+class ApiError(Exception):
     """An exception to throw when an endpoint should respond with an error."""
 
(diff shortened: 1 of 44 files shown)


shared-data/python/opentrons_shared_data/errors/exceptions.py
```diff
@@ -202,6 +202,19 @@ def _descend_exc_ctx(exc: BaseException) -> List[PythonException]:
...
+class RobotInUseError(CommunicationError):
+    """An error indicating that an action cannot proceed because another is in progress."""
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        detail: Optional[Dict[str, Any]] = None,
+        wrapping: Optional[Sequence[EnumeratedError]] = None,
+    ) -> None:
+        """Build a CanbusCommunicationError."""
```
fsinapi, line 214: lil typo

sfoster1:
Openapi docs and spec now show the error code in error response schemas with a default of 4000!

shared-data/python/opentrons_shared_data/errors/exceptions.py
```diff
@@ -455,3 +481,59 @@ def __init__(
...
+class InvalidActuator(RoboticsInteractionError):
+    """An error indicating that a specified actuator is not valid."""
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        detail: Optional[Dict[str, Any]] = None,
+        wrapping: Optional[Sequence[EnumeratedError]] = None,
+    ) -> None:
+        """Build an GripperNotPresentError."""
```
fsinapi, line 495: typo

robot-server/robot_server/errors/error_responses.py
```diff
@@ -105,6 +106,48 @@ def get_some_model():
             "occurrence of the error"
         ),
     )
+    errorCode: str = Field(
```
fsinapi, line 109: Should the type be `Optional[str]` since there's a default of `None`?

fsinapi (commented):
Had one question on pydantic model typing and a couple little doc typos. LGTM but I think I lack context on the impact on model changes so I'll hold off approving.

robot-server/robot_server/errors/error_responses.py
sfoster1: That's weird that it allows that. i should change that to default to 4000 anyway

robot-server/robot_server/errors/error_responses.py
```diff
@@ -121,12 +156,29 @@ class LegacyErrorResponse(BaseErrorBody):
...
+        cls: Type["LegacyErrorResponse"], exc: BaseException
+    ) -> "LegacyErrorResponse":
+        """Build a response from an exception, preserving some detail."""
+        if not isinstance(exc, EnumeratedError):
+            checked_exc: EnumeratedError = PythonException(exc)
+        else:
+            checked_exc = exc
+
+        return cls(
+            message=checked_exc.message.strip(), errorCode=checked_exc.code.value.code
```
SyntaxColoring, line 174: Why do we need to `.strip()` the message?

robot-server/robot_server/errors/global_errors.py
```diff
@@ -1,39 +1,61 @@
...
+        **supplemental_kwargs: Any
+    ) -> "FirmwareUpdateRequired":
+        """Build a FirmwareUpdateRequired from a specific exception. Preserves metadata."""
+        parent_inst = ErrorDetails.from_exc(exc, **supplemental_kwargs)
+        inst = FirmwareUpdateRequired(**parent_inst.dict())
+        if not inst.meta:
+            inst.meta = {"update_url": "/subsystems/update"}
+        else:
+            inst.meta["update_url"] = "/subsystems/update"
+        return inst
```
SyntaxColoring, line 61: Let me make sure I understand what's going on here:

`meta` normally includes attributes automatically taken from the Python exception. But when we're returning a `FirmwareUpdateRequired` model, we also want to include a helpful `update_url` pointer. That won't be present in the original Python exception, so we need to add it here.

Is that right?

robot-server/robot_server/runs/router/actions_router.py
```diff
@@ -116,12 +116,10 @@ async def create_run_action(
...
 
     except RunActionNotAllowedError as e:
-        raise RunActionNotAllowed(detail=str(e)).as_error(
-            status.HTTP_409_CONFLICT
-        ) from e
+        raise RunActionNotAllowed.from_exc(e).as_error(status.HTTP_409_CONFLICT) from e
 
     except RunNotFoundError as e:
-        raise RunNotFound(detail=str(e)).as_error(status.HTTP_404_NOT_FOUND) from e
+        raise RunNotFound.from_exc(e).as_error(status.HTTP_404_NOT_FOUND) from e
```
SyntaxColoring, line 122: Very very very happy that we're making this broken `detail=str(e)` pattern less pervasive.

This `raise SomeModel.from_exc(..).as_error(...)` pattern is the confusing back-and-forth thing between models and exceptions that you were talking about before, right?

SyntaxColoring (approved):
> Does the way we treat `MultipleErrorResponse` and `ErrorResponse` make sense? What's the point of the multi error response anyway, given that error response can hold multiple errors?

Hmm. It looks like this was introduced in #7783. I don't know if this was the intent, but theoretically, it lets us provide a stricter model when we know we're only returning a single error. (Which is almost always.) Like, we can override the OpenAPI spec to include `"minItems": 1, "maxItems": 1` or something.

robot-server/robot_server/errors/error_responses.py
sfoster1: for a reason I can't be bothered with it comes with a trailing newline

robot-server/robot_server/errors/global_errors.py
sfoster1: Yeah pretty much. It's pretty equally metadata, so it doesn't really do any harm to put it in there next to all the other stuff

robot-server/robot_server/runs/router/actions_router.py
sfoster1: Yeah, and it's worse in some other places like the exception handlers, where we 
1. catch an exception
2. build a model from it
3. create an exception from the model with `as_error`
4. pass that somewhere else
5. ultimately put the exception instance's `.content` in the response

What I'd really prefer is that exceptions get turned into models and models get sent, without going back into ApiErrors.

PR #12567, opened 2023-04-26
test(api): Configure ctx fixture as PAPIv2.14 when it's configured as an OT-3

SyntaxColoring:
# Overview

The `api` tests have this top-level `ctx` pytest fixture:

https://github.com/Opentrons/opentrons/blob/a89d9657155937816793020b07c64d92194c6738/api/tests/opentrons/conftest.py#L258-L260

Right now, any tests that use this `ctx` fixture will try to run under two configurations:

| | OT-2 hardware simulator and deck definition | OT-3 hardware simulator and deck definition |
| :- | :- | :- |
| **PAPI `version="2.13"`** | ✅  | ✅  |

With this PR, that changes to this:

| | OT-2 hardware simulator and deck definition | OT-3 hardware simulator and deck definition |
| :- | :- | :- |
| **PAPI `version="2.13"`** | ✅  |  🚫   |
| **PAPI `version="2.14"`** | 🚫 | ✅ |

This work is one part of RLAB-250, and it will also help with RCORE-535.

# Test plan

* Make sure the `api` tests keep passing, both locally and in CI.
* Make sure the `api` tests don't have any hangs.
* Make sure the `api` tests don't have any new flaky failures.

# Why

PAPIv2.14 runs through Protocol Engine, and PAPIv2.13 runs through the legacy PAPI back-end.

The legacy PAPI back-end will not support the OT-3. So this fixture was "wrong," in the sense that it was causing tests to run in a configuration that we don't expect to work.

In practice, it's happened to work so far. At least, it's worked well enough. However, in my work on RLAB-250, it stopped working well enough. I wanted to change the OT-3's deck definition to have slot names like `"D1"` instead of `"1"`, but PAPIv2.13's `Deck` class isn't built to support that, so I got a bunch of errors from the PAPIv2.13 `Deck` class:

```
ERROR tests/opentrons/protocol_api_old/test_labware_load.py::test_load_to_slot[OT-3 Standard] - ValueError: invalid literal for int() with base 10: 'D1'
ERROR tests/opentrons/protocol_api_old/test_labware_load.py::test_loaded[OT-3 Standard] - ValueError: invalid literal for int() with base 10: 'D1'
ERROR tests/opentrons/protocol_api_old/test_labware_load.py::test_load_label[OT-3 Standard] - ValueError: invalid literal for int() with base 10: 'D1'
[...]

PR #12970, opened 2023-06-23
feat(api): Limited support for programmatically running PAPIv≥2.14

SyntaxColoring:
# Overview /  Changelog

These interfaces now support executing PAPIv≥2.14 protocols:

* The `opentrons_execute` CLI
* `opentrons.execute.execute()`
* `opentrons.execute.get_protocol_api()`

This works towards, but does not close, RSS-268.

Known limitations:

* *Simulating* PAPIv≥2.14 protocols will still raise a `NotImplementedError`.
* Executing or simulating *JSONv6+* protocols will still raise a `NotImplementedError`.
* It will not print the run log. (Ticketed as RSS-283.)

# Test Plan

## Regression testing

* [x] `opentrons_execute` with a PAPIv≤2.13 file
    * [x] Only standard labware
    * [x] Custom labware with `-L`
    * [x] Custom labware in the current working directory
* [x] `opentrons.execute.execute()` with a PAPIv≤2.13 file
    * [x] Only standard labware
    * [x] Custom labware with `custom_labware_paths=...`
    * [x] Custom labware in the Jupyter Notebook directory
* [x] `opentrons.execute.get_protocol_api()` with PAPIv≤2.13
    * [x] Only standard labware
    * [x] Custom labware with `extra_labware=...`
    * [x] Custom labware in the Jupyter Notebook directory

## New feature testing

* [x] `opentrons_execute` with a PAPIv≥2.14 file
    * [x] Only standard labware
    * [x] Custom labware with `-L`
    * [x] Custom labware in the current working directory
* [x] `opentrons.execute.execute()` with a PAPIv≥2.14 file
    * [x] Only standard labware
    * [x] Custom labware with `custom_labware_paths=...`
    * [x] Custom labware in the Jupyter Notebook directory
* [x] `opentrons.execute.get_protocol_api()` with PAPIv≥2.14
    * [x] Only standard labware
    * [x] Custom labware with `extra_labware=...`
    * [x] Custom labware in the Jupyter Notebook directory
* [x] Do at least one of these on an OT-3 

# Review requests

I hate several things about this.

* The `_LIVE_PROTOCOL_ENGINE_CONTEXTS` global.
* `_adapt_protocol_source()`.
* `create_protocol_engine_in_thread()`. This is technically not new—it was added in #12567—but now it's being used in production code for the first time.
* Letting `opentrons.execute` drift further apart from `opentrons.simulate`.

# Risk assessment

Medium. This is messy, highly-duplicated, difficult to automatically test, and public-facing.

Much of this code is sensitive to differences between `None` and `{}` for `bundled_labware`, `extra_labware`, and `bundled_data`.

# Next steps

We should try executing some Flex hardware testing scripts. From @ryanthecoder:

> `python3 -m hardware_testing.gravimetric --photometric --pipette 1000 --channels 96 --tip 200 --trials 5` this is how we call our scripts. the scripts themselves load a python protocol to fetch which type and location of labware but then just executes python directly talking to the protocol context

After that, we should continue working on RSS-268 until these interfaces are good enough to launch the Flex with.


SyntaxColoring:
Note to self: Try to add tests for PAPIv2.14 in `test_execute.py`. Something is better than nothing.

api/src/opentrons/execute.py
```diff
@@ -5,60 +5,101 @@
...
+# completes. However, when someone gets a live ProtocolContext through get_protocol_api(), we have
+# no way of knowing when they're done with it. So, as a hack, we keep these resources open
+# indefinitely, letting them leak.
+#
+# We keep this at module scope so that the contained context managers aren't garbage-collected.
+# If they're garbage collected, they can close their resources prematurely.
+# https://stackoverflow.com/a/69155026/497934
+_LIVE_PROTOCOL_ENGINE_CONTEXTS = contextlib.ExitStack()
+
+
```
SyntaxColoring, line 106: Reminder to self: Add an `atexit` handler for this, for consistency with `_THREAD_MANAGED_HW` and to eliminate unnecessary task leak warnings.

api/src/opentrons/execute.py
```diff
@@ -5,60 +5,101 @@
...
 # See Jira RCORE-535.
-_PYTHON_TOO_NEW_MESSAGE = (
-    "Python protocols with apiLevels higher than 2.13"
-    " cannot currently be executed with"
-    " the opentrons_execute command-line tool,"
-    " the opentrons.execute.execute() function,"
-    " or the opentrons.execute.get_protocol_api() function."
-    " Use a lower apiLevel"
-    " or use the Opentrons App instead."
-)
```
sanni-t, line 61: Maybe we can modify this message to talk about the limitations of the current implementation and log it as a warning instead. Just a thought

api/src/opentrons/execute.py
```diff
@@ -124,34 +171,39 @@ def get_protocol_api(
...
             deck_type=deck_type,
             hardware_api=hardware_controller,
             bundled_labware=bundled_labware,
             bundled_data=bundled_data,
             extra_labware=extra_labware,
         )
-    except protocol_api.ProtocolEngineCoreRequiredError as e:
-        raise NotImplementedError(_PYTHON_TOO_NEW_MESSAGE) from e  # See Jira RCORE-535.
+    else:
+        if bundled_labware is not None:
```
sanni-t, line 189: I don't know the context of bundled labware. Is it there for future implementation/completion or was it a part of some old version that we don't plan on supporting for newer versions? If it's the latter, we should update the error message to indicate that it's not supported anymore.

SyntaxColoring: Yeah, I also don't know. Let's ask around and see if we can remove it.

api/src/opentrons/execute.py
SyntaxColoring: Discussed in-person, I think:

We'll raise fine-grained errors when, and only when, you try to do the specific things that the implementation doesn't yet support. I think we should keep them `NotImplementedError`s for now, instead of warnings.

api/src/opentrons/execute.py
```diff
@@ -401,24 +476,260 @@ def main() -> int:
...
+        # First, do we print the first element, or the last, or all of them?
+        #
+        # Second, do we print the .detail? .errorCode? .errorInfo? .wrappedErrors?
+        # By contract, .detail seems like it would be insufficient, but experimentally,
+        # it includes a lot, like:
+        #
+        #     ProtocolEngineError [line 3]: Error 4000 GENERAL_ERROR (ProtocolEngineError):
+        #     UnexpectedProtocolError: Labware "fixture_12_trough" not found with version 1
+        #     in namespace "fixture".
+        return self._error_occurrences[0].detail
```
SyntaxColoring, line 533: @TamarZanzouri @sfoster1 Is there a better thing for me to do here, now that stuff like #13008 is merged?

Here's what I'm trying to solve:

When we execute a protocol through Protocol Engine, we get any kind of fatal error—bugs, hardware errors, Protocol Engine validation errors, non-protocol-related Python mistakes, everything—through Protocol Engine's `state_summary.errors`. This is a list of `ErrorOccurrence`s.

From that, we need to decide:

1. How do we propagate that to callers of `opentrons.execute.execute()`? It needs to be an exception, but what do we put in it?

   I currently have the list of `ErrorOccurrence`s as its only member. In stack traces, Python will stringify that as the exception message. This seems...okay, I guess? It's structured and lossless, at least. But it's nonstandard. I guess I'd really prefer if it were a normal Python exception chain all the way down. 

2. What do we print to stderr for callers of the `opentrons_execute` CLI? Stringifying the `ErrorOccurrence`, or parts of it, seems plausible, but I don't know the details of how to do that properly.

api/src/opentrons/execute.py
```diff
@@ -401,24 +476,260 @@ def main() -> int:
...
+
+        protocol_runner = create_protocol_runner(
+            protocol_config=protocol_source.config,
+            protocol_engine=protocol_engine,
+            hardware_api=hardware_api,
+        )
+
+        # TODO(mm, 2023-06-30): This will home and drop tips at the end, which is not how
+        # things have historically behaved with PAPIv2.13 and older or JSONv5 and older.
+        result = await protocol_runner.run(protocol_source)
```
sanni-t, line 660: Oof, another problem with default/ forced tip dropping.

api/src/opentrons/execute.py
```diff
@@ -401,24 +476,260 @@ def main() -> int:
...
 
 
+def _get_protocol_engine_config() -> Config:
+    """Return a Protocol Engine config to execute protocols on this device."""
+    return Config(
+        robot_type=_get_robot_type(),
+        deck_type=DeckType(guess_deck_type_from_global_config()),
+        # We deliberately omit ignore_pause=True because, in the current implementation of
+        # opentrons.protocol_api.core.engine, that would incorrectly make
+        # ProtocolContext.is_simulating() return True.
```
sanni-t, line 685: Oo! I did not realize that's how we are checking for simulating engine. Looking at that code, maybe there's scope to reduce redundancy in some configs (virtual pipettes/ modules/ pause ignore) and pooling them all together into an `is_simulating` check (not for this PR). But I might be forgetting a reason why it was split into multiple flags.

api/src/opentrons/execute.py
```diff
@@ -401,24 +476,260 @@ def main() -> int:
...
+) -> ProtocolContext:
+    """Return a live ProtocolContext that controls the robot through ProtocolEngine."""
+    assert api_version >= ENGINE_CORE_API_VERSION
+
+    global _LIVE_PROTOCOL_ENGINE_CONTEXTS
+    pe, loop = _LIVE_PROTOCOL_ENGINE_CONTEXTS.enter_context(
+        create_protocol_engine_in_thread(
+            hardware_api=hardware_api.wrapped(),
+            config=_get_protocol_engine_config(),
+            drop_tips_and_home_after=False,
```
sanni-t, line 575: So, am I understanding this correctly that live protocols like those over Jupyter will not drop tips and home at the end of the 'protocol session', but python/json protocol execution will?

sanni-t (approved):
🙌 Awesome! I know it has a few hacks and still needs to be implemented fully, but this is an amazing start. This ticket is quite a complex puzzle and I think you've approached it quite carefully, covering a lot of edge cases. I might have missed a couple of things while reviewing too (re: it's a complex change), but unit tests and on-robot testing has been successful so I think this is merge worthy!

api/src/opentrons/execute.py
SyntaxColoring: Yep, exactly.

api/src/opentrons/execute.py
SyntaxColoring: Yeah. the situation is actually not quite as bad as I thought it was, though. I thought we didn't have a way to clean up an engine without dropping tips. We do: we just have do do `ProtocolEngine.finish(drop_tips_and_home=False)`. The trouble is just at the `ProtocolRunner` level, which doesn't propagate that functionality up.

api/src/opentrons/execute.py
SyntaxColoring: Yeah, it seems like we should either split it *even further* and have a dedicated flag for something like `tell_python_protocol_it_is_simulating`, or merge them all. We're in a weird in-between right now.

SyntaxColoring:
~~This PR is causing `pipenv run pytest tests/opentrons/test_execute.py` to hang on exit when run on Python 3.10. Looking into it.~~

Here's why it was hanging:

A Python thread can be configured to be "daemon" or "non-daemon." The difference is that when the process is shutting down, the Python interpreter won't wait for daemon threads to be joined.

More precisely, the Python interpreter will only start to call our `@atexit` handlers *after all non-daemon threads have joined.* This creates a clear deadlock. A non-daemon thread will only exit after our `@atexit` handler tells it to close, but our `@atexit` handler will never get a chance to run because there's still a pending thread, so the process will hang on shutdown.

We're seeing this on Python 3.10 and not 3.7 because 3.9 changed `ThreadPoolExecutor`'s threads from daemon to non-daemon. (https://github.com/python/cpython/issues/83993).

My workaround is to ditch `ThreadPoolExecutor` in favor of a manually-managed `Thread` so we can set `daemon=True`. [Here's another workaround](https://stackoverflow.com/a/63075281/497934) that's less deeply intrusive, but is more horrifying to me in other ways.

I think the real solution is to let `get_protocol_api()` users call some kind of `.close()` method. Then, we don't need those `@atexit` handlers, and we don't need `daemon=True`. I agree with [this Stack Overflow answer](https://stackoverflow.com/a/20598791/497934) in that the interpreter shutdown phase is not a happy place to be coding. Per [prior discussion](https://opentrons.slack.com/archives/C02B2P5J708/p1687470261491379), I'm not adding that `.close()` method in this line of PRs. But I view all of this as more reason why we need to do that at some point.

SyntaxColoring:
I have a verbal 👍 from @TamarZanzouri on the above change to fix the shutdown hanging.

PR #12971, opened 2023-06-23
fix(shared-data): exclude module calibration adapters from being rendered.

vegano1:
# Overview

Module calibration adapters are labware but meant for internal uses so we want to exclude them so they aren't rendered on the app.

Closes: [RQA-811](https://opentrons.atlassian.net/browse/RQA-811)

# Test Plan

- [x] Make sure the labware are excluded

# Changelog

- add module calibration adapter load names to `LABWAREV2_DO_NOT_LIST` in shared-data so they are excluded when rendering.

# Review requests

Does this make sense?

# Risk assessment
low

[RQA-811]: https://opentrons.atlassian.net/browse/RQA-811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared-data/js/getLabware.ts
```diff
@@ -37,6 +37,10 @@ export const LABWAREV2_DO_NOT_LIST = [
   'eppendorf_96_tiprack_10ul_eptips',
   'opentrons_calibrationblock_short_side_left',
   'opentrons_calibrationblock_short_side_right',
+  'opentrons_calibration_adapter_heatershaker_module',
+  'opentrons_calibration_adapter_magnetic_module',
+  'opentrons_calibration_adapter_temperature_module',
+  'opentrons_calibration_adapter_thermocycler_module',
```
SyntaxColoring: Nit: Can we move these below the `TODO(lc 8-24-2022)` comment below? Right now, the comment above makes it look like these are for "back-compat with legacy v1 defs."

Also, do we need these in `PD_DO_NOT_LIST` too?

vegano1: will do, 
and I'm not sure,
@b-cooper do we also want to add to `PD_DO_NOT_LIST `?

jerader (approved):
this makes sense to me. trying to figure out if we want them in the PD_DO_NOT_LIST as well but easy enough to add them later.

ewagoner (approved):

PR #12972, opened 2023-06-25
docs(api): add tabs to Python API docs

ecormany:
# Overview

This PR adds a Sphinx extension for creating tabbed sets of content on documentation pages. We plan to use tabs for quick comparison between parallel content for Flex and OT-2 (mostly code, but also related descriptions).

There's only one tab set included here, on the home page of the documentation. Edits to the surrounding text and tabs on other pages will come in subsequent PRs.

# Test Plan

Check out the tab behavior in [the sandbox](http://sandbox.docs.opentrons.com/docs-sphinx-tabs/v2/).

# Changelog

- Adds `sphinx-tabs` to `/api/Pipfile`
- Loads `sphinx-tabs` and sets options in the `conf.py` file.
- Adds the first tab set on the Welcome page.

# Review requests

Looking for all feedback on organizing content this way and the Sphinx implementation.

# Risk assessment

Low. Adds a Sphinx extension that runs in CI.


api/Pipfile
```diff
@@ -24,6 +24,7 @@ wheel = "==0.30.0"
 typeguard = "==2.12.1"
 sphinx-substitution-extensions = "==2020.9.30.0"
 sphinxext-opengraph = "==0.8.1"
+sphinx-tabs = "~=3.4.1"
```
SyntaxColoring: This is [PEP 440's fault for specifying confusing behavior](https://peps.python.org/pep-0440/#compatible-release), but `~=3.4.1` means `>=3.4.1,<3.5`, which is probably not what we want if sphinx-tabs obeys semantic versioning. I think we should use `>=3.4.1,<4` instead.

SyntaxColoring (approved):
Neat!

jwwojak:
Tabs are an ideal feature to use.

- Saves us from maintaining separate `.rst` files for each robot.
- When OT-2 is deprecated, you can easily revert back to a single page by removing obsolete tab. 

As a style issue, I suggest always putting the Flex tab first followed by the OT-2 tab.

SyntaxColoring:
> As a style issue, I suggest always putting the Flex tab first followed by the OT-2 tab.

Agreed. I've been trying to mention Flex before OT-2 in prose, and it would be nice to be consistent.

jwwojak (approved):
Tabs: Works for me! In fact, this is an ideal way to handle small documentation changes that account for both robots. Tabs, tabs, tabs!

PR #12983, opened 2023-06-26
fix(api): always home right mount first (for FLEX)

ahiuchingau:
# Overview
We want to make sure the right mount never gets in the way during a home move when (1) 96 channel pipette is attached, or (2) if the 96-chan mounting plate is on. We don't actually know whether or not the plate is mounted (no sensor), so we just have to always home the right side first.

### DIFF
--- a/api/src/opentrons/hardware_control/types.py
+++ b/api/src/opentrons/hardware_control/types.py
@@ -190,7 +190,7 @@ class OT3Axis(enum.Enum):
     @classmethod
     def mount_axes(cls) -> Tuple["OT3Axis", "OT3Axis", "OT3Axis"]:
         """The axes which are used for moving instruments up and down."""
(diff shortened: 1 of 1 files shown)


Laura-Danielle:
@ahiuchingau , do we not want this to be an app concern? Are there other scenarios the right mount gets moved down during 96 channel actions?

sfoster1 (approved):
Looks good to me but let's get this in 0.13.0

ryanthecoder (approved):
looks good

ahiuchingau:
> @ahiuchingau , do we not want this to be an app concern? Are there other scenarios the right mount gets moved down during 96 channel actions?

During the pipette detach flow, the right mount would come down so user could remove the mounting plate. If the robot gets turned off at this stage, the head axes would not be able to home the next time the server starts up again.

PR #12986, opened 2023-06-26
fix(hardware): clear move group from node if move condition is met

ahiuchingau:
# Overview
The expected-stall move group times out whenever a stall is detected before the final move block, even though this is an acceptable behavior. This fixes it by clearing and finishing the move group whenever a move block is completed with a stall.

### DIFF
--- a/hardware/opentrons_hardware/hardware_control/move_group_runner.py
+++ b/hardware/opentrons_hardware/hardware_control/move_group_runner.py
@@ -530,6 +530,7 @@ class MoveScheduler:
                 log.warning(
                     f"Expected nodes in group {str(group_id)}: {str(self._get_nodes_in_move_group(group_id))}"
                 )
+                raise
             except (RuntimeError, MoveConditionNotMet) as e:
                 log.error("canceling move group scheduler")
                 raise e
--- a/hardware/opentrons_hardware/hardware_control/move_group_runner.py
+++ b/hardware/opentrons_hardware/hardware_control/move_group_runner.py
@@ -206,10 +206,12 @@ class MoveGroupRunner:
         Args:
             can_messenger: a can messenger
         """
-        await can_messenger.send(
+        error = await can_messenger.ensure_send(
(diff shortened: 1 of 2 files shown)


Laura-Danielle (changes requested):
This check looks good, but is there a way we could add a quick test for this scenario?

fsinapi (changes requested):
I really think this should just clear out the one single _node_ from the move group, and still wait until every node has reported that it's done. The way this is written, other axes could still be moving after the one axis reports its stall, and the robot could move on and start a new move before they stop moving.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -422,6 +422,17 @@ def _handle_move_completed(
                 )
                 self._should_stop = True
                 self._event.set()
+            if (
+                stop_cond == MoveStopCondition.stall
+                and ack_id == MoveAckId.stopped_by_condition
```
sfoster1: I think we might want this to just be `ack_id == MoveAckId.stopped_by_condition` whether or not the stop condition is stall; I think you're going to get similar behavior from home commands, including the home backoffs, and the tip action motors, and definitely the stop-on-sync commands from calibration

sfoster1 (changes requested):
The more I think about it, I think we don't want to do this _only_ if the stop condition is stall - I think this should happen any time we're `stopped_by_condition` no matter what the condition is

fsinapi (approved):
Nice, LGTM!

hardware/opentrons_hardware/hardware_control/move_group_runner.py
ahiuchingau: You're right!

sfoster1 (approved):
Nice, thank you!

Laura-Danielle (approved):
ty looks good to me!

PR #12991, opened 2023-06-27
fix(app): look for ok gripper

smb2268:
# Overview
I am accidentally looking for a !ok gripper instead of an ok gripper so it wasn't showing up in the app when attached


TamarZanzouri (approved):
LGTM! thank you!

PR #12993, opened 2023-06-27
refactor(protocol-designer): module tags for OT-2

jerader:
closes RAUT-524

# Overview

Update module tag for Ot-2 deckmap by creating `FlexModuleTag` and `Ot2ModuleTag` branching logic. Delete legacy `ModuleTag`

# Test Plan

- add all modules to Ot-2 and flex protocols. Make sure the module tag is visible and in correct position

# Changelog

- delete `ModuleTag` and css styles and test
- create new module tags for robot types
- add logic to `DeckSetup`

# Review requests

see test plan

# Risk assessment

low


b-cooper (approved):
The code looks good to me. Out of curiosity, why did we need to special-case the TC on the OT2?

jerader:
> The code looks good to me. Out of curiosity, why did we need to special-case the TC on the OT2?

It doesn't match up to the side of the module very well like the other modules so figured I could just use the same positioning as the Flex modules for the TC

PR #12994, opened 2023-06-27
fix(hardware): stop ot3 controller tests hanging

fsinapi:
# Overview

The `_read_task_shield` for the can messenger was never ending, which kept any tests using it from ending. On the robot the iterator will never end, but in the tests we should raise an exception to let the shield know when to let itself terminate.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
yup yup yup

PR #12996, opened 2023-06-27
fix(hardware-testing): OT3API.cache_instruments needs a delay before being called

andySigler:
# Overview

This PR adds a small delay in `hardware_testing.opentrons_api.helpers_ot3.build_async_ot3_hardware_api()` before caching instruments. It was found that trying to cache instruments immediately after instantiating the api does not detect attached instruments. At least a delay of 100ms is needed before caching instruments, so this PR sets the delay to 500ms.

I don't know why this is happening, and also don't know why we need to now run cache_instruments() when we didn't need to a couple weeks back, but this PR's goal is unblocking testing scripts.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
I don't know why that's happening either! But if it works it works

PR #12997, opened 2023-06-27
feat(step-generation, protocol-designer): add timeline error for off deck labware

jerader:
closes RAUT-525

# Overview

If you try to interact with a labware that you moved off deck previously in the protocol, you will get a timeline error

Also fix up some weirdness in the `Select` component causing some dropdowns to have too wide a height and width.

# Test Plan

- move the labware off deck and then add an aspirate, dispense, blow out, moveToWell with the labware that is off deck as a source. When you create the command, you should see an error
- try to move the labware back onto the deck after it is currently off deck with the gripper, you should see the same error

# Changelog

- add i18n strings back for `HEATER_SHAKER_LATCH_CLOSED` - i think i had a merge conflict and accidentally deleted the previous i18n strings 😢 
- create errorCreator `labwareOffDeck` and add to the appropriate atomic commands
- add the error to i18n
- change the width = 100% to be in a parent component rather than tweaking the Select component that more components use.

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/components/src/forms/Select.tsx
+++ b/components/src/forms/Select.tsx
@@ -15,7 +15,8 @@ import type {
   StylesConfig,
   CSSObjectWithLabel,
 } from 'react-select'
-import { Box, SPACING } from '..'
+import { Box } from '../primitives'
+import { SPACING } from '../ui-style-constants'
 
 export { reactSelectComponents }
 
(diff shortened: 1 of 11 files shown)


protocol-designer/src/components/modals/FilePipettesModal/FilePipettesModal.css
```diff
@@ -133,3 +133,7 @@
 .module_model {
   height: 3rem;
 }
+
+.pipette_select {
```
jerader, line 137: `classNames` was already being prop drilled so it made more sense to just add on to that rather than convert all the `PipetteSelect` components to using styled-components

protocol-designer/src/components/modals/FilePipettesModal/PipetteFields.tsx
```diff
@@ -99,24 +100,27 @@ export function PipetteFields(props: Props): JSX.Element {
     const nameBlockList = [...OT2_PIPETTES, 'p1000_96']
```
b-cooper: nit: if this is going to be called `nameBlockList` as a const, it would make more sense to have the ternary on robot type in this line. Right now, it is only actually the nameBlockList for Flexes.  
e.g.
```
const nameBlockList = robotType === OT2_ROBOT_TYPE ? OT3_PIPETTES : [...OT2_PIPETTES, 'p1000_96']
```

step-generation/src/commandCreators/atomic/moveToWell.ts
b-cooper, line 1: should we also throw an error in `moveLabware` if the labware is source or dest is 'offDeck', and we're using the gripper?

protocol-designer/src/components/modals/FilePipettesModal/PipetteFields.tsx
jerader: oops ya, good call

step-generation/src/commandCreators/atomic/moveToWell.ts
jerader: yeah! that makes sense to me

jerader: i think the same error creator should work for this?

step-generation/src/commandCreators/atomic/moveLabware.ts
```diff
@@ -28,6 +28,8 @@ export const moveLabware: CommandCreator<MoveLabwareArgs> = (
         labware,
       })
     )
+  } else if (prevRobotState.labware[labware].slot === 'offDeck' && useGripper) {
```
b-cooper, line 31: this is only checking if the source location is off deck, we should also check if the destination is off deck here

jerader: okay just added it, however, the `offDeck` option is being filtered out in the labware location dropdown if a gripper is being used for it so hopefully users never run into this.

b-cooper: Cool thanks for updating it. Since step-generation and protocol-designer are two separate modules, we shouldn't rely on form or field level protections in PD to prevent accessing non-sensical parameters in step-generation. Also theoretically, step-generation could be used by more than just PD. Its job is to validate step args against all known errors and turn them into protocol engine commands.

protocol-designer/src/localization/en/alert.json
```diff
@@ -139,6 +139,14 @@
...
         "title": "8-Channel pipette cannot access labware",
         "body": "8-Channel pipettes cannot access labware in front of or behind a Heater-Shaker. They can access Opentrons Tip Racks in this slot. Move labware to a different slot."
+      },
+      "LABWARE_OFF_DECK": {
+        "title": "Robot interacting with an off deck labware",
+        "body": "Labware must be on the deck for the robot to interact with it. To resolve this error, please make sure the labware is on deck."
+      },
+      "HEATER_SHAKER_LATCH_CLOSED": {
+        "title": "Heater-Shaker labware latch is closed",
+        "body": "Before the robot can interact with labware on the Heater-Shaker module, the labware latch must be open. To resolve this error, please add a Heater-Shaker step ahead of the current step, and set the labware latch status to \"open\"."
```
jerader, line 149: Note: going to have Ed review and rewrite these and include the change in a different PR.

Also, the previous i18n strings for `HEATER_SHAKER_LATCH_CLOSED` accidentally got deleted when i was rebasing something at some point, that is why it is randomly added here.

components/src/forms/Select.tsx
```diff
@@ -15,7 +15,8 @@ import type {
   StylesConfig,
   CSSObjectWithLabel,
 } from 'react-select'
-import { Box, SPACING } from '..'
+import { Box } from '../primitives'
+import { SPACING } from '../ui-style-constants'
```
shlokamin: these should go before type imports 😛

shlokamin (approved):
lgtm!

components/src/forms/Select.tsx
jerader: 😬

PR #12999, opened 2023-06-27
fix(hardware): use ensure send when clearing move group

ahiuchingau:
# Overview
We should use ensure_send whenever we can to make sure we're receiving ack messages before sending new moves to the nodes.


sfoster1 (approved):
Oh that's a good idea

fsinapi (approved):
I think some tests need fixing but this is Really Nice

PR #13000, opened 2023-06-27
fix(app): ODD white screen when instrument detaches

smb2268:
fix RQA-992, RQA-1001

# Overview
Adding null protection to the InstrumentInfo card where it was missing -- this was triggering the ODD error boundary when an instrument is detached during the detach flow


jerader (approved):
lgtm!

PR #13001, opened 2023-06-27
feat(api): add retract_axis to OT-2 and FLEX hardware-control api

ahiuchingau:
Addresses hardware-control side of RQA-997
Protocol engine & app changes will follow in #13006 

# Overview

Adds per-axis retraction feature to hardware controller API for OT2 & Flex

# Test Plan

Test on OT2 & Flex and make sure the axes retract correctly. Also make sure mount retraction remains unchanged.

# Changelog

- added `retract_axis` HC api methods and updated mount retraction to delegate to the new method

# Review requests

# Risk assessment

Low


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1291,8 +1291,33 @@ async def retract(
 
         Works regardless of critical point or home status.
         """
-        machine_ax = OT3Axis.by_mount(mount)
-        await self._home((machine_ax,))
+        await self.restract_axis(OT3Axis.by_mount(mount))
```
```suggestion
        await self.retract_axis(OT3Axis.by_mount(mount))
```

sfoster1 (approved):
Looks good to me, typo tho

sanni-t (commented):
Tested this along with #13006 (without app changes) on an OT3 and confirmed that axis retraction works. The only issue observed was that motion planning doesn't seem to know that the axis is already at its highest and doesn't need to move to a different point to clear all labware on deck. Pretty sure that's on the engine so will confirm and get back to you so we can merge this PR in.

sanni-t:
> The only issue observed was that motion planning doesn't seem to know that the axis is already at its highest and doesn't need to move to a different point to clear all labware on deck. Pretty sure that's on the engine so will confirm and get back to you so we can merge this PR in.

Alright, verified that fixing this issue required only a protocol engine change.
So @andySigler or @ahiuchingau you can undraft this PR and merge in edge. I'll rebase my branch on top of this one.

sanni-t (approved):

PR #13003, opened 2023-06-28
fix(hardware-testing): Actually use increment protocols so correct labware is loaded

andySigler:
# Overview

We were accidentally not loading in the "increment" labware types/locations when executing, so the vial was being assumed instead of the trough and liquid-levels were wrong.

Protocols file are currently used to define labware types and locations per each test scenario. These are then used to run LPC before a test is executed.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
Looks good

PR #13004, opened 2023-06-28
refactor(api): Refactors ahead of ProtocolEngine-based opentrons.execute

SyntaxColoring:
# Overview

Assorted refactors to prepare `opentrons.execute` to run PAPIv>=2.14 and JSONv>=6 protocols. This work goes towards, but does not close, RSS-267 and RSS-268.

# Test Plan

* [ ] Be careful to make sure all CI workflows continue to pass, especially ones outside `api`, like `hardware-testing` and `g-code-testing`.

# Changelog

See my review comments.

# Review requests

See my review comments.

# Risk assessment

Low. The only behavioral change so far is removing the `machine` argument. It's not public-facing, but we need to make sure nothing internal is depending on it.


api/tests/opentrons/util/test_entrypoint_utils.py
```diff
@@ -1,48 +1,56 @@
...
-            "fixture/fixture_irregular_example_1/1": lw2,
-        }
-    )
+    assert res == {
+        "fixture/fixture_96_plate/1": FoundLabware(path_1 / "labware1.json", lw1),
+        "fixture/fixture_24_tuberack/1": FoundLabware(path_1 / "labware2.json", lw2),
+        "fixture/fixture_irregular_example_1/1": FoundLabware(
+            path_2 / "labware3.json", lw3
+        ),
+    }
```
SyntaxColoring, line 50: Fun test bug: the `sorted()` calls here were making the test totally ignore the values in these dictionaries. `sorted({"k1": "v1", "k2": "v2"})` is `["k1", "k2"`]. This caused the test to keep passing despite a typo where `lw2` was given twice, instead of `lw3`.

api/src/opentrons/util/entrypoint_util.py
```diff
@@ -16,10 +17,23 @@
...
+    path: pathlib.Path
+    definition: "LabwareDefinition"
+
+
 def labware_from_paths(
     paths: Sequence[Union[str, pathlib.Path]]
-) -> Dict[str, "LabwareDefinition"]:
-    labware_defs: Dict[str, "LabwareDefinition"] = {}
+) -> Dict[str, FoundLabware]:
+    """Search paths for labware definitions.
```
SyntaxColoring, line 31: We're changing `labware_from_paths()` to return the file paths where it found the labware, in addition to information from the parsed definitions.

This addition is unused in this PR, but it will be helpful in the next one:

`ProtocolEngine` needs to be run through `ProtocolRunner`. `ProtocolRunner` requires static protocol information from `ProtocolReader`. `ProtocolReader` requires you to provide the protocol source as an opaque list of *file paths*; it doesn't let you just inject labware definitions that you've already parsed.

There are a lot of ways to improve that situation, such as, perhaps, resolving some of the duplication between `opentrons.protocols.parse` and `ProtocolReader`. Until we do that, this seemed like the least bad option.

api/src/opentrons/protocol_api/core/engine/__init__.py
```diff
@@ -1,18 +1,12 @@
...
-from opentrons.protocols.api_support.types import APIVersion
 
 from .protocol import ProtocolCore
 from .instrument import InstrumentCore
 from .labware import LabwareCore
 from .module_core import ModuleCore
 from .well import WellCore
 
-ENGINE_CORE_API_VERSION: Final = APIVersion(2, 14)
-
```
SyntaxColoring: We're moving `ENGINE_CORE_API_VERSION` from `opentrons.protocol_api.core.engine` to `opentrons.protocols.api_support.constants`. The execution code will need to access it to decide whether a protocol should be executed with the new machinery, or the old machinery. Importing it from `opentrons.protocol_api.core.engine` felt too private to me.

I'm not totally sure if this is worthwhile. What do you think? I can easily undo this.

api/src/opentrons/execute.py
```diff
@@ -61,7 +64,6 @@ def get_protocol_api(
     bundled_labware: Optional[Dict[str, "LabwareDefinition"]] = None,
     bundled_data: Optional[Dict[str, bytes]] = None,
     extra_labware: Optional[Dict[str, "LabwareDefinition"]] = None,
-    machine: Optional[MachineType] = None,
```
SyntaxColoring, line 64: Removing the `machine` argument goes towards RSS-267. It doesn't totally close it because there's work to do in `opentrons.simulate`, too, unfortunately.

api/src/opentrons/protocols/types.py
```diff
@@ -31,7 +33,11 @@ class StaticPythonInfo:
 class _ProtocolCommon:
     text: str
     filename: Optional[str]
+    # TODO(mm, 2023-06-22): Move api_level out of _ProtocolCommon and into PythonProtocol.
+    # JSON protocols do not have an intrinsic api_level, especially since JSONv6,
+    # where they are no longer executed via the Python Protocol API.
     api_level: "APIVersion"
+    robot_type: RobotType
```
SyntaxColoring, line 40: Executing or simulating a Flex protocol on an OT-2, or vice-versa, is unsupported. We want `opentrons.execute` and `opentrons.simulate` to raise a helpful error if you try to do that. To do that, `opentrons.parse` needs to report the protocol's declared `robotType`, which it wasn't doing before.

api/src/opentrons/protocol_api/core/engine/__init__.py
sfoster1: I think it should stay somewhere that more directly shows it's owned by PE, but I think most people trying to find it will do so by seeing its usage and going-to-definition so it's fine

sfoster1 (approved):
Big fan of this!

api/src/opentrons/protocol_api/core/engine/__init__.py
SyntaxColoring: > I think it should stay somewhere that more directly shows it's owned by PE

Mmmyeah. And thinking about this a little more, I'm a little worried this will make us more prone to circular import loops. I'll revert this change.

api/src/opentrons/execute.py
```diff
@@ -221,18 +236,13 @@ def get_arguments(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
 
 
 def execute(
-    protocol_file: TextIO,
+    protocol_file: Union[BinaryIO, TextIO],
```
SyntaxColoring, line 239: Even before this PR, `protocol_file` could actually be a `BinaryIO` at runtime. Notice the `argparse.FileType("rb")` above. Fortunately, the lower layers are carefully defined to work on `str`s *or* `bytes`s, so this mismatch hasn't caused problems in practice.

Why am I expanding this interface to match the `argparse` argument, instead of the other way around? Because technically, I think it's actually more appropriate to open a Python source file in binary mode than text mode.

Opening the source file in text mode forces us to make an up-front choice about its encoding, which is impossible in the general case. If we open it in binary mode instead, we defer that choice to the Python lexer, which can [detect the encoding more intelligently](https://docs.python.org/3.7/reference/lexical_analysis.html#encoding-declarations). This is related to, but does not close, RSS-159.

jbleon95 (approved):
Nice, this all looks good

PR #13006, opened 2023-06-28
 feat(api, app): add engine command for retracting axes and retract axes during LPC

sanni-t:
Closes RSS-278

# Overview

Adds a new protocol engine command- `retractAxis` to retract a specified robot axis.
Currently the execution of this command just no-ops. Will update the implementation once axis retraction is available in hardware control. This PR serves to unblock client side use of this command.

# Test Plan

Unit tests should suffice for now. Will test command calling with client-side changes.

# Changelog

- adds the new command and test
- updates schema v7

# Review requests

- Usual code review
- Main question @ahiuchingau and I had when deciding on the retraction feature is if it's okay to have the retraction be slightly different for OT2 vs Flex: on the OT-2, it will do a fast move to a safe distance from the axis' home position, and then probe the limit switch to reach the home position, while on the flex, it will move directly to the last known home position of the axis. The cautious approach for the OT2 is based on the original `retract_mount` method on the hardware control API which was created to home a mount after a tip pick-up in order to improve positional accuracy since a pick up action could make the mount motor skip steps.

So essentially `retractAxis` for OT-2 behaves the same as `home` on Flex. 
That said, the final outcome on both the robots is that the axis lands at its home position, faster than homing the axis on that robot.

On one hand, the behavior observed for retraction on the two robots is different and could confuse some users, while on the other hand, if we make the Flex's axis retraction behave the same way, we will be defining a more accurate robot's behavior based on a less accurate robot's constraints and have a slightly slower retraction on the Flex.

# Risk assessment

None. New command that's not wired up anywhere yet.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/retract_axis.py
+++ b/api/src/opentrons/protocol_engine/commands/retract_axis.py
@@ -1,7 +1,7 @@
 """Retract Axis command payload, result, and implementation models."""
 from __future__ import annotations
 from pydantic import BaseModel, Field
-from typing import TYPE_CHECKING, Optional, List, Type
+from typing import TYPE_CHECKING, Optional, Type
 from typing_extensions import Literal
 
 from ..types import MotorAxis
@@ -16,18 +16,17 @@ RetractAxisCommandType = Literal["retractAxis"]
 class RetractAxisParams(BaseModel):
     """Payload required for a Retract Axis command."""
 
-    axes: Optional[List[MotorAxis]] = Field(
-        None,
+    axis: MotorAxis = Field(
+        ...,
         description=(
-            "Axes to retract to their home positions."
-            " If ommitted, will retract all motors."
+            "Axis to retract as close to its home positions as safely possible."
             " The difference between retracting an axis and homing an axis using the"
             " home command is that a home will move the axis (slowly) until it finds"
             " the home limit switch, while retraction attempts to move the axis"
             " either to the previously recorded home position in case of the Flex,"
             " or a safe distance away from the previously recorded home position"
             " in the case of the OT-2."
-        )
+        ),
     )
 
 
@@ -35,19 +34,22 @@ class RetractAxisResult(BaseModel):
     """Result data from the execution of a Rectract Axis command."""
 
 
-class RetractAxisImplementation(AbstractCommandImpl[RetractAxisParams, RetractAxisResult]):
+class RetractAxisImplementation(
+    AbstractCommandImpl[RetractAxisParams, RetractAxisResult]
+):
     """Retract Axis command implementation."""
 
     def __init__(self, movement: MovementHandler, **kwargs: object) -> None:
         self._movement = movement
 
     async def execute(self, params: RetractAxisParams) -> RetractAxisResult:
-        """Retract some or all axes."""
-        await self._movement.retract_axis(axes=params.axes)
+        """Retract the specified axis."""
+        await self._movement.retract_axis(axis=params.axis)
         return RetractAxisResult()
 
+
 class RetractAxis(BaseCommand[RetractAxisParams, RetractAxisResult]):
-    """Command to retract some or all axes near their home positions."""
+    """Command to retract the specified axis near its home position."""
 
     commandType = RetractAxisCommandType = "retractAxis"
     params = RetractAxisParams
--- a/api/src/opentrons/protocol_engine/commands/retract_axis.py
+++ b/api/src/opentrons/protocol_engine/commands/retract_axis.py
@@ -51,9 +51,9 @@ class RetractAxisImplementation(
 class RetractAxis(BaseCommand[RetractAxisParams, RetractAxisResult]):
     """Command to retract the specified axis near its home position."""
 
-    commandType = RetractAxisCommandType = "retractAxis"
-    params = RetractAxisParams
-    result = Optional[RetractAxisResult]
+    commandType: RetractAxisCommandType = "retractAxis"
+    params: RetractAxisParams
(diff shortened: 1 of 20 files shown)


api/src/opentrons/protocol_engine/commands/retract_axis.py
```diff
@@ -0,0 +1,69 @@
...
+RetractAxisCommandType = Literal["retractAxis"]
+
+
+class RetractAxisParams(BaseModel):
+    """Payload required for a Retract Axis command."""
+
+    axis: MotorAxis = Field(
+        ...,
+        description=(
+            "Axis to retract to its home position as safely as possible."
```
SyntaxColoring: Do you mean "as quickly as possible" instead of "as safely as possible"? To me, "as safely as possible" would mean doing a slow home all the way over.

SyntaxColoring: I guess there are two possible notions of safety:

1. Doing our best to clear any potential obstacles on the deck. This means, for example, retracting Z before X and Y.
2. Accounting for the possibility that some of the axes might have been knocked out of place, so we can't assume anything about their current position.

I think we want homing to do (1) and (2), but we want retracting to only do (1), right?

b-cooper (approved):
The interface looks good to me 👍 !

api/src/opentrons/protocol_engine/commands/retract_axis.py
```diff
@@ -0,0 +1,69 @@
...
+            " The difference between retracting an axis and homing an axis using the"
+            " home command is that a home will always probe the limit switch and"
+            " will work as the first motion command a robot will need to execute;"
+            " On the other hand, retraction will rely on this previously determined "
+            " home position to move to it as fast as safely possible."
+            " So on the Flex, it will move (fast) the axis to the previously recorded home position"
+            " and on the OT2, it will move (fast) the axis a safe distance from the previously"
+            " recorded home position, and then slowly approach the limit switch."
```
SyntaxColoring, line 30: > Main question @ahiuchingau and I had when deciding on the retraction feature is if it's okay to have the retraction be slightly different for OT2 vs Flex

In general, if a command defines some loose-ish behavioral contract, I think it's okay for the exact implementation of that contract to depend on the hardware that the command is running on.

However:

> on the OT-2, it will do a fast move to a safe distance from the axis' home position, and then probe the limit switch to reach the home position,

> and on the OT2, it will move (fast) the axis a safe distance from the previously recorded home position, and then slowly approach the limit switch.

Why couldn't the OT-2 do what the OT-3 does?

This might just be a terminology thing. The "retract" that the Opentrons App wants for Labware Position Check is not the same thing as the "retract" that the hardware API needs to do after picking up a tip. When the Opentrons App wants to "retract", it knows that nothing has happened that could cause the motors to skip.

api/src/opentrons/protocol_engine/commands/retract_axis.py
sanni-t: Oh yes! I meant to write `as quickly as safely possible`.

> I think we want homing to do (1) and (2), but we want retracting to only do (1), right?

Yep!
Except, just making sure- you are using (1) as an example for level of safety and not actual implementation of the command, ya? Because the command will retract only the axis that is specified. It would be upto the consumer of the command to make sure it is used in proper context.

SyntaxColoring: > you are using (1) as an example for level of safety and not actual implementation of the command, ya? Because the command will retract only the axis that is specified. It would be upto the consumer of the command to make sure it is used in proper context.

Oh, I missed that it only takes a single axis.

That works for me if it works for the client team.

api/src/opentrons/protocol_engine/commands/retract_axis.py
sanni-t: > The "retract" that the Opentrons App wants for Labware Position Check is not the same thing as the "retract" that the hardware API needs to do after picking up a tip. When the Opentrons App wants to "retract", it knows that nothing has happened that could cause the motors to skip.

That is probably true but I don't know for sure so tagging @ahiuchingau for clarification. It would be confusing to call both of these behaviors as 'retract' though.

app/src/organisms/LabwarePositionCheck/CheckItem.tsx
```diff
@@ -203,6 +210,12 @@ export const CheckItem = (props: CheckItemProps): JSX.Element | null => {
 
   const handleConfirmPosition = (): void => {
     let confirmPositionCommands: CreateCommand[] = [
+      {
+        commandType: 'retractAxis' as const,
+        params: {
+          axis: [pipetteZMotorAxis],
```
sanni-t: This should be a single value instead of a list

shlokamin: my b, fixed!

sanni-t:
Tested on a RSS Flex and works as expected! However, after seeing it in action, I am starting to think it might be better to do `retract Z -> retract X -> retract Y` instead of `retract Z -> Move to trash -> retract Z`. The latter sequence can look quite confusing: the pipette moves up -> goes above trash -> moves down to trash well -> moves back up. 
What do you think @shlokamin @andySigler ?

sanni-t:
Talked with @andySigler in person and decided to merge this in for now and follow-up with an improved movement. Which could be either (a) retracting all axes one after other or (b) having the client add a Z offset to the 'move to trash' command. For (b) the sequence of commands would look like the following:
1. (after a position check) retract pipette axis
2. get current position of pipette
3. send a 'move to trash' with z offset set to the z height found in previous step

PR #13008, opened 2023-06-28
fix(api): raise Estop exception when Estop is pressed

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/RQA-601.
raise an e-stop activated from the hardware layer and catch it in PE. 

# Test Plan

- Tested with python protocols and json protocols on an OT-3. works as expected. 
- Need to raise a runtime exception somehow on an OT-2 and make sure the error handling has not changed.

# Changelog

- Raise an `PE_EStopActivatedError` from the `command_executer` when we catch an exception while running a command.
- Wrap the child thread exception with a `ProtocolCommandFailedError` and add logic to extract the Exception from it in the `CommandsView`. 
- Added `original_error` prop to `ProtocolCommandFailedError` to preserve the `ErrorOccurrence` that was raises from the `FailedCommandAction`

# Review requests

Do we feel good about moving`ProtocolCommandFailedError` into error_occurrence.py to avoid circular dependency?
Changes make sense?
Add more tests? 
Are you concerned from the errors raised now. 
@sfoster1 make sure I merged conflicts properly 

# Risk assessment

Medium. changing the way we bubble up python/json protocol exceptions. need to smoke test on OT-2 and OT-3.


api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -117,7 +119,7 @@ class FailCommandAction:
     command_id: str
     error_id: str
     failed_at: datetime
-    error: ProtocolEngineError
+    error: Union[ProtocolEngineError, EnumeratedError]
```
TamarZanzouri: use EnumeratedError

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -102,13 +109,14 @@ async def execute(self, command_id: str) -> None:
...
-
             # TODO(mc, 2022-11-14): mark command as stopped rather than failed
             # https://opentrons.atlassian.net/browse/RCORE-390
             if isinstance(error, asyncio.CancelledError):
                 error = RunStoppedError("Run was cancelled")
-            elif not isinstance(error, ProtocolEngineError):
-                error = UnexpectedProtocolError(message=str(error), wrapping=[error])
+            elif isinstance(error, EStopActivatedError):
+                error = PE_EStopActivatedError(message=str(error), wrapping=[error])
+            elif not isinstance(error, EnumeratedError):
```
TamarZanzouri, line 118: strip the pythonExcption message

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -253,10 +258,16 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
                 self._state.running_command_id = None
 
         elif isinstance(action, FailCommandAction):
+            log.info("in failed acation")
+            log.info(f"failed action error: {action.error} type {type(action.error)}")
             error_occurrence = ErrorOccurrence.from_failed(
-                id=action.error_id, createdAt=action.failed_at, error=action.error
+                id=action.error_id,
+                createdAt=action.failed_at,
+                error=ProtocolCommandFailedError(
```
TamarZanzouri: do not wrap in the error in ProtocolCommandFailedError

api/src/opentrons/protocol_engine/errors/error_occurrence.py
```diff
@@ -67,4 +71,19 @@ def schema_extra(schema: Dict[str, Any], model: object) -> None:
             schema["required"].extend(["errorCode", "wrappedErrors", "errorInfo"])
 
 
+class ProtocolCommandFailedError(ProtocolEngineError):
```
TamarZanzouri, line 75: I hate that this has to move bc of circular dependencies. should we just add a new error?

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -1,5 +1,6 @@
 """Protocol engine commands sub-state."""
 from __future__ import annotations
+
```
TamarZanzouri:
```suggestion
```

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -333,21 +334,31 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
...
-                            message=str(action.error_details.error),
-                            wrapping=[action.error_details.error],
+                    if (
+                        isinstance(
+                            action.error_details.error, ProtocolCommandFailedError
                         )
+                        and action.error_details.error.original_error is not None
+                    ):
+                        error_occurrence = action.error_details.error.original_error
+                        self._state.errors_by_id[error_id] = error_occurrence
```
TamarZanzouri:
```suggestion
                        self._state.errors_by_id[error_id] = action.error_details.error.original_error
```

api/src/opentrons/protocol_engine/errors/error_occurrence.py
sfoster1: As we talked about in offline review - I think we should refactor this in another PR so `ProtocolCommandFailedError` doesn't need to carry the ErrorOccurrence with it, which will fix the circular error

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -789,3 +776,16 @@ def __init__(
...
+class EStopActivatedError(ProtocolEngineError):
+    """Raised when an operation's required pipette tip is not attached."""
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        details: Optional[Dict[str, Any]] = None,
+        wrapping: Optional[Sequence[EnumeratedError]] = None,
+    ) -> None:
+        """Build a PIpetteNotAttachedError."""
```
sfoster1:
```suggestion
        """Build an EStopActivatedError."""
```

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -246,6 +249,16 @@ async def finish(
                 If `False`, will set status to `stopped`.
         """
         if error:
+            if (
+                isinstance(error, ProtocolCommandFailedError)
+                and error.original_error is not None
+                and any(
+                    ErrorCodes.E_STOP_ACTIVATED.value.code == wrapped_error.errorCode
```
sfoster1: this seems like a good thing to factor into a free function in `opentrons_shared_data.errors` - like a `def code_in_exception_stack(exc: EnumeratedError)`, I could see us using it in more places (and having a typescript equivalent, eventually)

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -556,7 +566,10 @@ def get_all_commands_final(self) -> bool:
             for command_id in self._state.all_command_ids:
                 command = self._state.commands_by_id[command_id].command
                 if command.error and command.intent != CommandIntent.SETUP:
-                    raise ProtocolCommandFailedError(command.error.detail)
+                    # TODO(tz, 7-11-23): avoid raising an error and return the status instead
```
sfoster1, line 571: 💯

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -492,6 +498,18 @@ def _get_nodes_in_move_group(self, group_id: int) -> List[NodeId]:
                 nodes.append(NodeId(node_id))
         return nodes
 
+    def _raise_exception_from_error(
```
sfoster1: This shouldn't be necessary I think. It's handled by the call to [`raise_from_error_message`](https://github.com/Opentrons/opentrons/blob/edge/hardware/opentrons_hardware/errors.py#L77)  [here](https://github.com/Opentrons/opentrons/blob/5394203ff9e6e225dcb59bc5b608b4b30e1ec157/hardware/opentrons_hardware/hardware_control/move_group_runner.py#L414)

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -509,7 +527,7 @@ async def _send_stop_if_necessary(
                         "Motion failed with multiple errors", wrapping=self._errors
                     )
                 else:
-                    raise self._errors[0]
+                    self._raise_exception_from_error(group_id=group_id, error=self._errors[0])
```
sfoster1: ditto here - should be handled by `raise_from_error_message` putting an exception instance into `self._errors`, and then getting raised from here anyway.

shared-data/python/opentrons_shared_data/errors/exceptions.py
```diff
@@ -196,7 +196,7 @@ def _descend_exc_ctx(exc: BaseException) -> List[PythonException]:
         base_details["class"] = type(exc).__name__
 
         super().__init__(
-            message="\n".join(format_exception_only(type(exc), exc)),
+            message="\n".join(format_exception_only(type(exc), exc)).strip(),
```
sfoster1, line 199: yessss

sfoster1 (changes requested):
Some small things to change but overall looks great.

I think the followup we should do is to get rid of this thing where we have to wrap `ErrorOccurence` instances in exceptions. THat's the cause of a lot of annoying wrapping and breaking of the structured exception metaphor, and it feels wasteful anyway because after those exceptions do their immediate jobs (i.e. stopping python protocol execution) they get... turned right back into error occurrences, except now with a lot of missing detail.

Let's (as you have commented) make `get_all_commands_final()` return some tristate enum of `RUNNING`, `STOPPED_OK`, `STOPPED_ERROR` or even `None` (running), `True` (done ok), and `False` (done with error) and have callers check for errors on the stored commands afterwards. This isn't a TOCTOU or race because it only returns these statuses once we're in `stopped` and we can't go back. It means we don't have to make this exception anymore, and we don't need to smush it in there.

Let's also do something similar at the layer above the python protocol execution, so we can take the contextual information out of the exception that we raise from the transport - including in the command line runners. The exception needs to exist to interrupt execution, sure, but as long as it's a stable type we can catch it special and go and investigate the state store for the actual error from there.

And that will rule! And fix the last little weird things in here. But again, overall looks good.

hardware/opentrons_hardware/hardware_control/move_group_runner.py
TamarZanzouri: nice! I believe this was introduced in your pr :-)

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -257,6 +263,16 @@ async def finish(
                 If `False`, will set status to `stopped`.
         """
         if error:
+            print(error)
```
sfoster1: stray diagnostic print

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -381,3 +397,17 @@ async def use_attached_modules(
 
         for a in actions:
             self._action_dispatcher.dispatch(a)
+
+    # TODO(tz, 7-12-23): move this to shared data when we dont relay on ErrorOccurrence
```
sfoster1, line 400: maybe make a note on the definition of `ProtocolCommandFailedError` to move this when we change that exception class?

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -333,21 +334,32 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
...
+                        self._state.errors_by_id[
+                            error_id
+                        ] = action.error_details.error.original_error
                     else:
-                        error = action.error_details.error
-
-                    self._state.errors_by_id[error_id] = ErrorOccurrence.from_failed(
-                        id=error_id, createdAt=created_at, error=error
-                    )
+                        if not isinstance(
```
sfoster1: nit: could flip these cases and then lose the explicit annotation and have only positive conditions

sfoster1 (changes requested):
Looks good though we need to get rid of that `print` hence the change request.

api/src/opentrons/protocol_engine/protocol_engine.py
TamarZanzouri:
```suggestion
```

TamarZanzouri: thank you!

api/src/opentrons/protocol_engine/protocol_engine.py
TamarZanzouri: yessssss good call!

sfoster1 (approved):
Looks good! Nice work!

PR #13009, opened 2023-06-28
feat(hardware): Hardware error codes

sfoster1:
Now that we can express error codes in all our client interfaces, let's start raising them.

This PR makes the `opentrons_hardware` module use the new exceptions that have the new error codes. There's a couple new error codes, but mostly all this does is raise different exceptions in the same places.

The two exception is in the can messenger and the movegrouprunner.

### CAN Messenger
The CAN messenger has this asyncio task that reads stuff asynchronously from the bus and passes it to registered listeners. In addition, if a message is an error message it would raise an exception. But, it's a task, so that exception doesn't go anywhere. #12963 made this exception just logged and swallowed but, like... why does it exist? So we removed it.

### MoveGroupRunner
A similar issue happens in MoveGroupRunner, where a background task accumulates responses and then pops them on over to the caller. This works the same now, but we can also move the move stop condition checking inside that loop. We can also now properly handle having multiple errors in a move group!

**Note**: You need to run `make setup-py` again since now hardware depends on shared-data.

## Review Requests
- Give a skim over most stuff
- Look more in-depth at MoveGroupRunner and at CANMessenger

## Testing
- [x] Cause some errors and maek sure it doesn't break in new ways.


hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -174,7 +176,14 @@ async def capacitive_probe(
         messenger,
     )
     if not threshold:
-        raise RuntimeError("Could not set threshold for probe")
+        raise CanbusCommunicationError(
```
ahiuchingau, line 183: Just a general question: We are raising the `CommandTimedOutError` in the send threshold listener if the sensor scheduler didn't get a response back. Are we expecting to see both the `CommandTimedOutError` and `CanbusCommunicationError` if the listener does time out?

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -461,15 +476,20 @@ async def _send_stop_if_necessary(
...
+                        "Motion failed with multiple errors", wrapping=self._errors
+                    )
+                else:
+                    raise self._errors[0]
             else:
                 # This happens when the move completed without stop condition
-                raise MoveConditionNotMet
-        elif self._error is not None:
-            log.warning(f"Recoverable firmware error during {group_id}: {self._error}")
+                raise MoveConditionNotMetError(detail={"group-id": str(group_id)})
```
ahiuchingau, line 505: I don't think we'll need to raise `MoveConditionNotMetError` there as we have already appended it to`self._errors` in `_handle_move_completed()`

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -528,9 +548,6 @@ async def run(self, can_messenger: CanMessenger) -> _Completions:
                 log.warning(
                     f"Expected nodes in group {str(group_id)}: {str(self._get_nodes_in_move_group(group_id))}"
                 )
```
ahiuchingau, line 548: This TimeoutError should be raised now (I got it in internal-release_0.13.0 but not edge)

hardware/opentrons_hardware/hardware_control/move_group_runner.py
```diff
@@ -461,15 +476,20 @@ async def _send_stop_if_necessary(
...
             if err != ErrorCode.stop_requested:
                 log.warning("Stop request failed")
-            if self._error:
-                raise RuntimeError(
-                    f"Unrecoverable firmware error during move group {group_id}: {self._error}"
-                )
+            if self._errors:
+                if len(self._errors) > 1:
+                    raise MotionFailedError(
+                        "Motion failed with multiple errors", wrapping=self._errors
```
ahiuchingau, line 499: AMAZING

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
```diff
@@ -375,37 +371,27 @@ async def _read_task(self) -> None:
...
                         if filter and not filter(message.arbitration_id):
                             log.debug("message ignored by filter")
                             continue
                         listener(message_definition(payload=build), message.arbitration_id)  # type: ignore[arg-type]
-                    if (
-                        message.arbitration_id.parts.message_id
-                        == MessageId.error_message
-                    ):
-                        await self._handle_error(build)
+                        handled = True
```
ahiuchingau, line 377: Is it possible that we could get in a state where a listener accepts an error message but for whatever reason, it doesn't have an error handling method and hence not raise it? 

In that case, would this error message just be thrown away since the can messenger thinks it was handled by a listener?

ahiuchingau (changes requested):
Some general questions and i think we need to raise an extra TimeourError in the MoveScheduler

hardware/opentrons_hardware/hardware_control/tool_sensors.py
sfoster1: No, you'd only see the timeout error since we're not catching and ignoring it. I think the only time you'd see this getting raised is if the microcontroller responded with `threshold=0`

hardware/opentrons_hardware/hardware_control/move_group_runner.py
sfoster1: Then this code won't run, right? because `if self._errors` will run? I think this clause might not necessarily have to exist or at least be a weird generic catchall though, maybe it should just be a general error

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
sfoster1: Yup. All I can really say is, like, "don't do that"

hardware/opentrons_hardware/hardware_control/move_group_runner.py
sfoster1: I think that I don't want to _add_ that in this PR - I want your commit in its entirety. So I think that either we can merge this now as is and re-add it after yours merges (and, we're going to have a lot of fixups for errors added since 0.13.0 diverged from edge, so it won't be unique) or wait on this PR until 0.13.0 gets merged back.

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
```diff
@@ -349,18 +346,17 @@ def remove_listener(self, listener: MessageListenerCallback) -> None:
...
-            except (asyncio.CancelledError, StopAsyncIteration):
-                return
-            except (CANCommunicationError, AsyncHardwareError, CanError) as e:
-                log.exception(f"Nonfatal error in CAN read task: {e}")
-                continue
-            except Exception as e:
-                # Log this separately if it's some unknown error
-                log.exception(f"Unexpected error in CAN read task: {e}")
-                continue
+        try:
```
vegano1: What restarts this task now if we have an exception?

hardware/opentrons_hardware/firmware_update/run.py
```diff
@@ -60,7 +68,15 @@ async def find_dfu_device(pid: str, expected_device_count: int) -> str:
         if stdout is None and stderr is None:
             continue
         if stderr:
-            raise RuntimeError(f"Error finding dfu device: {stderr.decode()}")
+            raise BootloaderNotReady(
+                USBTarget.rear_panel,
```
vegano1, line 72: is this guaranteed to always be the rear panel?
maybe so, since its the only USB update target for now

vegano1 (approved):
This is great!
Is there a follow-up pr on the robot-server side to handle any of these errors getting propagated?

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
sfoster1: Huh good catch I guess this PR was old enough that it missed those changes

hardware/opentrons_hardware/firmware_update/run.py
sfoster1: Right now yeah... I should really rework it to not do this though

sfoster1:
> This is great! Is there a follow-up pr on the robot-server side to handle any of these errors getting propagated?

Nope, that one was already merged. All the structures to forward these errors should already be in place in robot server and protocol engine.

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
sfoster1: Should be fixed @vegano1

ahiuchingau (approved):

PR #13011, opened 2023-06-28
feat(app): add a new type for red frame

koji:
# Overview
A modal with red frame is used by e-stop modal when a run is active which is the same as touchscreen app.
close RCORE-800

# Test Plan
- run StoryBook and check LegacyModal and select `outlinedError`

# Changelog
- update LegacyModal components
- add test for LegacyModalHeader

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/molecules/LegacyModal/index.tsx
+++ b/app/src/molecules/LegacyModal/index.tsx
@@ -3,7 +3,6 @@ import { SPACING, COLORS, Box } from '@opentrons/components'
 import { LegacyModalHeader } from './LegacyModalHeader'
 import { LegacyModalShell } from './LegacyModalShell'
 import type { StyleProps } from '@opentrons/components'
-import { isError } from 'lodash'
 
 type ModalType = 'info' | 'warning' | 'error'
 export * from './LegacyModalShell'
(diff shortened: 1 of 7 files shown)


app/src/molecules/LegacyModal/index.tsx
```diff
@@ -16,6 +16,7 @@ export interface LegacyModalProps extends StyleProps {
   fullPage?: boolean
   childrenPadding?: string | number
   children?: React.ReactNode
+  isError?: boolean
```
ewagoner: Really not a fan of having both a `type` of `error` and an `isError` boolean, especially since it can be an `error` type but have an `isError` of false.

Can you add a fourth type, maybe, of `outlinedError` or change the boolean to something like `isOutlined`? The former would be cleaner (and I think better) but the latter is less work.

app/src/molecules/LegacyModal/index.tsx
```diff
@@ -29,21 +29,67 @@ export const LegacyModal = (props: LegacyModalProps): JSX.Element => {
...
+  }
+
+  const modalIcon: IconProps = {
+    name: 'ot-alert',
+    color: iconColor(type),
+    size: '1.25rem',
+    marginRight: SPACING.spacing8,
+  }
+
+  // const MODAL_PROPS_BY_TYPE: Record<
```
ewagoner: Detritus?

ewagoner (approved):
Other than the left-over commented out code, looks good. So much cleaner this way!

app/src/molecules/LegacyModal/index.tsx
koji: thanks I'll remove that line

PR #13012, opened 2023-06-29
fix(hardware-testing): All tipracks names to use flex instead of ot3

andySigler:
# Overview

Use the new tiprack "flex" names everywhere, to avoid confusion.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
LGTM

PR #13014, opened 2023-06-29
fix(hardware-testing): Update FW after building hcapi

andySigler:
# Overview

FW updates used to be automatic when building the hcapi, but now they are not, so this PR forces an update each time a test script builds the hcapi.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
yea seems like a good guard.

PR #13015, opened 2023-06-29
fix(hardware-testing): Bug in checking reservoir volume and tip-rack names

andySigler:
# Overview

This PR fixes a bug where the volume in the reservoir was not being calculated correctly to determine if enough liquid is present for the following aspiration.

Also renames tip-racks to "flex" tipracks

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
good catch

PR #13017, opened 2023-06-29
feat(app): send fatal ODD browser logs to electron layer

shlokamin:
# Overview

This PR sends the electron layer a message when a fatal error occurs in the app so it gets picked up in logs. This will help tremendously when debugging ODD browser side errors.

# Risk assessment

Low

### DIFF
--- a/app-shell-odd/src/restart.ts
+++ b/app-shell-odd/src/restart.ts
@@ -7,9 +7,9 @@ export function registerAppRestart(): (action: Action) => unknown {
   return function handleAction(action: Action) {
     switch (action.type) {
       case APP_RESTART:
-        systemd.sendStatus('restart app')
+        systemd.sendStatus(`restarting app: ${action.payload.message}`)
+        console.log(`restarting app: ${action.payload.message}`)
         systemd.restartApp()
-
         break
     }
   }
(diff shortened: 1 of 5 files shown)


app-shell-odd/src/restart.ts
```diff
@@ -7,9 +7,9 @@ export function registerAppRestart(): (action: Action) => unknown {
   return function handleAction(action: Action) {
     switch (action.type) {
       case APP_RESTART:
-        systemd.sendStatus('restart app')
+        systemd.sendStatus(`restarting app: ${action.payload.message}`)
+        console.log(`restarting app: ${action.payload.message}`)
```
shlokamin, line 11: this console log is intentional, it will show up inside of the opentrons-robot-app journal logs

app-shell-odd/src/robot-logs.ts
```diff
@@ -12,25 +13,37 @@ export function registerRobotLogs(
...
+        break
+      case 'shell:SEND_LOG':
+        systemd
+          .sendStatus(action.payload.message)
+          .catch((e: Error) =>
+            console.error(`error sending status to systemd ${e.message}`)
+          )
+        console.log(
+          `shell message from browser layer: ${action.payload.message}`
+        )
```
shlokamin, line 45: same thing here, this console log is intentional

sfoster1 (approved):
Looks good to me - unless we want to pull in some journald integration this is the right way to go, and all the other random deps spewing to console are already caught like this anyway

PR #13021, opened 2023-06-30
feat(api): support overpressure monitoring for multiple mounts

Laura-Danielle:
# Overview

This is a follow-up PR to the comments on #12864 regarding a few edge cases that were missed. Specifically from @andySigler:

> There are two scenarios where pressure should be monitored but it is not yet as currently implemented:

> prepare for aspirate
> api.home()
> Might be easier to manage all these cases if the pressure monitoring was instead happening inside OT3Controller.home() and OT3Controller.move(), and it's called whenever a plunger axis is passed in as an argument, but we could save that for a followup PR.

# TODO

- Add support for different sensor ids

# Test Plan

- [ ] Execute a protocol with a few pipetting functions (i.e. aspirate/dispense/blowout/drop tip) and verify via the serial log or CAN logs that a bind request for max threshold is sent. To ease testing, you can also tell the request to log so you can see the pressure values being print to screen.
- [ ] Using monitor sensors, determine a sufficient occlusion material (maybe a piece of cloth?) to get it to exceed 3.9kpa. This should cause the firmware to throw and error and stop all movement.

# Changelog

- Move the monitor sensor check to the `home` command and the `move` command. We should make sure that aspirate/dispense/blowout/drop tip still throw errors in addition to prepare for aspirate and `home()`. 

# Review requests

Anything seem particularly weird and/or gross?

# Risk assessment

Medium. Modifying the behavior of monitor over pressure.


hardware/opentrons_hardware/sensors/scheduler.py
```diff
@@ -473,70 +473,80 @@ def _filter(arbitration_id: ArbitrationId) -> bool:
...
+        error_response_queue: "asyncio.Queue[Tuple[NodeId, ErrorCode]]" = (
+            asyncio.Queue()
+        )
 
         def _async_error_listener(
             message: MessageDefinition, arb_id: ArbitrationId
         ) -> None:
             # FIXME could possibly make this error more generalized for other
             # sensors in the future
+            # FIXME we don't quite have a way to access the sensor id
```
sfoster1, line 489: don't we have the lexically-captured `target_sensors` list?

sfoster1 (changes requested):
I think that at this point `monitor_overpressure` doesn't have to be a public method anymore; and if it's not a public method, it doesn't have to speak `Mount`s. Instead, we can make it take a list of nodes and figure stuff out itself.

Also am I missing/forgetting about from a previous PR/misreading things or where are we doing the monitoring while moving the _plunger_?

Laura-Danielle:
@sfoster1 , @andySigler wants us to just always be monitoring the pressure sensors whenever the plunger motors are moving (regardless of tip status).

Unless I am misunderstanding your question?

hardware/opentrons_hardware/sensors/scheduler.py
Laura-Danielle: Error messages do not contain sensor ids. If an error message is coming from the same node id (which it potentially could for the 8/96 channel) then we have no way of knowing whether the error is coming from say the back nozzle vs the front nozzle.

Make sense?

andySigler (approved):
tested on grav bot running protocol with normal tip and sealed tip. Sealed tip triggered over-pressure at expected plunger position.

hardware/opentrons_hardware/sensors/scheduler.py
sfoster1: yup! thank you!

sfoster1 (approved):
Looks good to me!

PR #13026, opened 2023-06-30
refactor(hardware): move scheduler

ahiuchingau:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (changes requested):
This is looking good, but I think for readability wins it has to be taken a bit farther - get the runner all the way out of the business of iterating through things based on node+seq; maybe make that the responsibility of something new replacing `self._moves` or something. Ideally we should start with the concept of factoring out some chunk of responsibility and making it a separately-testable thing.

Also, we should wait until 0.13.0 is in to merge thi; there's a lot of changes to this code in there.

ahiuchingau:
Bit off too much by trying to put everything in this giant refactor - closing this and splitting it up into smaller PRs instead

PR #13028, opened 2023-06-30
refactor(api): Change ungrip to use positional move instead of force move

andySigler:
# Overview

This PR changes `OT3API.ungrip()` to use encoder positioning (as opposed to constant force). The advantage here is less energy used (less heat), faster motion (less time), and better jaw positional accuracy.

Will likely followup with a few more small gripper motion-control updates similar to this.

# Test Plan

- [ ] Use ot3repl to grip() and ungrip() on a labware, making sure it still works as expected
- [ ] Home the gripper jaw, and measure the distance between the paddles. Then grip() and ungrip(), and measure distance between paddles again, checking to make sure they are the same.

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -707,7 +707,7 @@ class OT3API(
             dc = self._gripper_handler.get_duty_cycle_by_grip_force(
                 gripper.default_home_force
             )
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Looks good to me as long as we're still homing, like, sometimes

PR #13029, opened 2023-06-30
feat(app, components): add a new const to spacing constants

koji:
# Overview
add a new const to spacing constants since 6px is used in some components and the design team uses it.

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


components/src/atoms/CheckboxField/index.tsx
```diff
@@ -130,7 +130,7 @@ export function CheckboxField(props: CheckboxFieldProps): JSX.Element {
           size={SIZE_1}
         >
           <Box
-            height="1.5px"
+            height="0.09375rem"
```
ewagoner: That seems impossibly small.

ewagoner (approved):

PR #13030, opened 2023-06-30
fix(app): toast and snackbar tweaks [RAUT-379]

ewagoner:
# Overview

This PR adds a new tap-to-close behavior for snackbars, tweaks icon spacing for toasts, and adds closing animation for both.

Closes RAUT-379

# Test Plan

Manually tested on desktop, ODD, and storybook

# Changelog

- Closes active snackbar when tapped
- Tweaks padding between icon and text on toasts
- Adds closing animations to toasts and snackbars

# Review requests

Trigger a snackbar and tap on it to watch it disappear. Let toasts and snackbars close on their own and watch as the slide out or fade from view gracefully.

# Risk assessment

None

### DIFF
--- a/app/src/atoms/Toast/__tests__/Toast.test.tsx
+++ b/app/src/atoms/Toast/__tests__/Toast.test.tsx
@@ -1,10 +1,13 @@
 import * as React from 'react'
 import { act, fireEvent } from '@testing-library/react'
 import { renderWithProviders } from '@opentrons/components'
+import { i18n } from '../../../i18n'
 import { Toast } from '..'
 
 const render = (props: React.ComponentProps<typeof Toast>) => {
-  return renderWithProviders(<Toast {...props} displayType="desktop" />)[0]
+  return renderWithProviders(<Toast {...props} displayType="desktop" />, {
+    i18nInstance: i18n,
+  })[0]
 }
 
 describe('Toast', () => {
--- a/app/src/atoms/Toast/__tests__/Toast.test.tsx
+++ b/app/src/atoms/Toast/__tests__/Toast.test.tsx
@@ -119,7 +119,7 @@ describe('Toast', () => {
     })
     expect(props.onClose).not.toHaveBeenCalled()
     act(() => {
-      jest.advanceTimersByTime(8000)
+      jest.advanceTimersByTime(9000)
     })
     expect(props.onClose).toHaveBeenCalled()
   })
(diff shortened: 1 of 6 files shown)


ewagoner:
https://github.com/Opentrons/opentrons/assets/2960/30959f3e-973a-4361-bf04-f3069cd4bfbd

app/src/atoms/Toast/__tests__/Toast.test.tsx
koji, line 1: we will need to update `render` like below.

todo @koji 
```ts
const render = (props: React.ComponentProps<typeof Toast>) => {
  return renderWithProviders(<Toast {...props} displayType="desktop" />, {
    i18nInstance: i18n,
  })[0]
}
```

ewagoner: I think you only need to do that if you're using any language translations in your tests. The toasts do not.

koji: Hmm, but actually when I run the test, warning shows up.

ewagoner: Oh, I see that now. You're right -- the test doesn't use translations but the component does, so the tests pass but it throws a warning. I'll make that change.

ewagoner: Done!

koji: thanks!

app/src/organisms/ToasterOven/ToasterOven.tsx
```diff
@@ -114,6 +114,9 @@ export function ToasterOven({ children }: ToasterOvenProps): JSX.Element {
           position="absolute"
           bottom={SPACING.spacing40}
           zIndex={1000}
+          onClick={() => {
+            eatSnackbar()
```
koji, line 118: Unfortunately this change doesn't fit Mel's request, `the functionality of tapping anywhere on the screen to make the snackbar disappear? `

ewagoner: This is true -- on the ticket I said I had to compromise by only enabling tapping on the snackbar itself (and to the left and right of it) because the ToasterOven does not take over the entire screen like modals do -- just a strip across the bottom. I think this solution will satisfy the users' need, who expect to be able to dismiss the notification by tapping on it.

koji: Okay, got it. Thanks, @ewagoner

koji (approved):

PR #13031, opened 2023-06-30
fix(hardware-testing): Retract farther away from scale

andySigler:
# Overview

This PR moves the pipette farther away from the scale during readings to avoid measurement errors from static buildup in the tips and/or labware.

This PR also cleans up the custom labware, so that all custom labware files are located in ./hardware-testing/labware and no where else, and running `make push-ot3-gravimetric` now pushes all labware to the `/data/labware` folder on the robot.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
looks good

PR #13032, opened 2023-06-30
chore(hardware): can motion log analyzer

sfoster1:
This fun little script can parse through serial log captures from an ot3 and print out some stuff about the motion, like a per-axis dump of the motor and encoder position tracking.

It should be usable on a robot as long as you're not doing any of the plot operations, it can print stuff out just fine.

Some usage notes:
- In the plot, if you click inside a data axis you will get some nice black vertical lines that let you visually line up data


Laura-Danielle (approved):
love this!! sad I can't use it irl

vegano1 (approved):

PR #13033, opened 2023-07-01
feat(app): update device reset for gripper calibration

koji:
# Overview
update device reset component to activate gripper calibration reset and update buttons' styles.
In terms of `Factory reset`, it will be implemented by a following PR and it has been ticketed already.

design
https://www.figma.com/file/AoTLAYuWawlaWItB1umOjr/Release%3A-Opentrons-Flex-Touchscreen?type=design&node-id=7948-118690&mode=design&t=AwB5o6W0Bsxv2Gmy-0

Screenshot 2023-07-05 at 3 11 45 PM

There are a couple of things I need to ask about design to the design team.
- Desktop app uses `Factory Reset` but Touch screen uses `Factory reset`
- In Touchscreen design uses `Pipette Offset Calibrations` and `Pipette Offset Calibration(s)`

# Test Plan

# Changelog
- update DeviceReset components and tests

# Review requests

# Risk assessment
low


app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
```diff
@@ -78,15 +80,23 @@ export function DeviceReset({
     }
   }
 
+  let subText: string
   const renderText = (optionId: string): string => {
```
shlokamin: rather than having the `renderText` function mutate a global variable (`subText`), can we rename + update this function to return text AND subtext? then we can use it when we render JSX.

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
```diff
@@ -78,15 +80,23 @@ export function DeviceReset({
...
         return t('clear_option_pipette_calibrations')
-      // ToDo (kj:02/07/2023) gripperCalibration same as the above
-      case 'gripperCalibration':
+      case 'gripperOffsetCalibrations':
         return t('clear_option_gripper_calibration')
       case 'runsHistory':
+        subText = t('clear_option_runs_history_subtext')
         return t('clear_option_runs_history')
+      // ToDo (kj:07/01/2023) we need to add factory reset case
+      // Asking a question about text Desktop Factory Reset/Touchscreen Factory rest
```
shlokamin: what's the question? lmk how i can help unblock you!

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
```diff
@@ -128,33 +138,37 @@ export function DeviceReset({
...
+                  </StyledText>
+                  {option.id === 'runsHistory' ? (
+                    <StyledText
+                      as="p"
+                      color={
+                        resetOptions[option.id]
+                          ? COLORS.white
+                          : COLORS.darkBlack70
+                      }
+                    >
```
shlokamin: i don't follow this block — why is this necessary?

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
koji: Actually this part was solved on Monday.
I will remove the comment

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
koji: A couple of options have subtext and it is needed to change the text color when a option is selected or deselected

shlokamin: but this is only for runs history, why is that one special? 

```js
{option.id === 'runsHistory' ? (
```

koji: Because when I opened this PR, runHistory was the only one item that has subText.
But I just updated this part.

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
```diff
@@ -40,8 +40,10 @@ const OptionLabel = styled.label<LabelProps>`
     ${({ isSelected }) =>
       isSelected === true ? COLORS.blueEnabled : COLORS.light2};
```
brenthagen: i don't see any border on this button in figma

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
```diff
@@ -111,50 +134,60 @@ export function DeviceReset({
...
         </Flex>
-        <AlertPrimaryButton
-          paddingY={SPACING.spacing24}
+        <MediumButton
+          data-testid="DeviceReset_clear_data_button"
+          buttonText={t('clear_data_and_restart_robot')}
+          buttonType="alert"
           disabled={
             Object.keys(resetOptions).length === 0 ||
             availableOptions.every(option => resetOptions[option.id] === false)
```
brenthagen: this button remains enabled after selecting and "unselecting" an option - we need to check for whether the option exists in `resetOptions` in addition to its value:
```
            availableOptions.every(
              option =>
                resetOptions[option.id] === false ||
                resetOptions[option.id] === undefined
            )
```

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
koji: oh good catch, I forgot to remove when the design was update.

jerader: Looks like Brent commented on there not needing a border but i still see a border? 
```suggestion
      isSelected === true ? COLORS.blueEnabled : undefined};
```
or
```suggestion
      isSelected === true ? COLORS.blueEnabled : COLORS.mediumBlueEnabled};
```

jerader (approved):
lgtm but left a question about border

brenthagen (approved):

app/src/organisms/RobotSettingsDashboard/DeviceReset.tsx
jerader: on nevermind, i must have been looking at an older commit!

jerader: looks good now 😄

PR #13034, opened 2023-07-03
chore(hardware): plot z and gripper

sfoster1:
Didn't handle the special nodes for the motor axes properly, oops.


Laura-Danielle (approved):
oops

vegano1 (approved):

PR #13038, opened 2023-07-03
feat(hardware-testing): Configurable plunger acceleration per liquid-class

andySigler:
# Overview

This PR adds plunger acceleration (ul/sec/sec) as one of the configurations that are applied during pipetting motions. This allows the system default accelerations to be used during homing and positioning, and then these faster accelerations can be used during only aspirate/dispense/blow-out movements.

# Test Plan

Run 2x trials on the following (make sure they are wet-runs so liquid is actually moved):

- [x] 1ch P50
- [x] 8ch P50
- [x] 1ch P1000 (any tip)
- [x] 8ch P1000 (any tip)
- [ ] 96ch gravimetric (any tip)

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
looks good to me

ryanthecoder (approved):
looks good

PR #13041, opened 2023-07-05
fix(hardware-testing): 96ch photometric confirm volume at beginning

andySigler:
# Overview

If the 96ch is tilted then the volume in the reservoir may need to be greater than specified, so this PR adds a step where the operator can add more liquid at the beginning of the test-run if needed.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
Ahh yes this will save a lot of time.

PR #13044, opened 2023-07-05
feat(hardware-testing): 96ch tip-rack adapters share LPC offsets

andySigler:
# Overview

This PR speeds up setup/calibration for 96ch volumetric testing, by allowing all tip-rack adapters to share the same labware offsets regardless of which tip-volume they currently hold. Offsets are only shared within each slot, so if you run LPC on a 1000uL rack in slot D2, then running the volumetric test with a 200uL rack in slot D2 will use the same LPC offsets as the 1000uL rack.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
Yea this makes sense

PR #13049, opened 2023-07-05
fix(api): Position plunger to bottom before picking up a tip

andySigler:
# Overview

This PR moves the plunger to its bottom position immediately before picking up a new tip. This is to avoid the scenario where a protocol (or a test script!) has just started with all axes homed, and the then positions to bottom after picking up the first tip. This is bad for two reasons:
1. We need to wait for it, for sometimes several seconds, and it always kind of feels like something wrong is happening
2. Volumetric tests show that tips perform differently depending on how long they are held, and so we want all tips to be held for the same amount of time during volumetric tests

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1668,6 +1668,7 @@ class OT3API(
             realmount, tip_length, presses, increment
         )
 
+        await self._move_to_plunger_bottom(realmount, rate=1.0)
         if spec.pick_up_motor_actions:
             await self._motor_pick_up_tip(realmount, spec.pick_up_motor_actions)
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
That looks good to me and the rationale makes sense!

Laura-Danielle (approved):
makes sense. Can we remove the `_move_to_plunger_bottom()` command from `prepare_for_aspirate` now?

andySigler:
> makes sense. Can we remove the `_move_to_plunger_bottom()` command from `prepare_for_aspirate` now?

we still need to include it in the scenario when the pipette needs to aspirate following a blow-out

PR #13050, opened 2023-07-06
fix(app): prevent longpress to analysis failed protocol

koji:
# Overview
According to Mel, any interaction on that card should lead to that pop up asking them to delete it.

# Test Plan
1. All Protocols
2-a. tap a failed protocol -> delete modal shows up
2-b. long press a failed protocol -> delete modal shows up

# Changelog
- update ProtocolCard component and test

# Review requests

# Risk assessment
low


jerader (approved):
lgtm!

PR #13051, opened 2023-07-06
refactor(shared-data): allow all flex tipracks for PD

jerader:
# Overview

Just noticed a small bug for flex tiprack selection in PD and found the culprit was from the `PD_DO_NOT_LIST`. I think something changed in the past week regarding flex tiprack definitions (like there were duplicates or something?) so thats why i noticed this bug. But anyways, it is fixed now.

# Test Plan

just verify that that a PD flex pipette can select both unfiltered and filtered version of the tip
sandbox can be found here when the build is done: https://sandbox.designer.opentrons.com/shared_data-pd-flex-tips

# Changelog

- change const

# Review requests

see test plan

# Risk assessment

low


koji (approved):
They show up in the wizard screens.

Screenshot 2023-07-06 at 12 24 53 PM
Screenshot 2023-07-06 at 12 25 05 PM

PR #13053, opened 2023-07-06
feat(hardware-testing): Option to isolate specific channels on 8ch pipette during grav tests

andySigler:
# Overview

This PR helps to speed up debugging issues in gravimetric tests for 8ch pipettes, by allowing the operator to specify which channels to isolate and test only, skipping all other channels. The use case for this is right now P50 multis need their 1uL volumes re-tested but only on 2x or 3x of their channels. Without being able to isolate just those channels, the test would take ~6 hours instead of ~2 hours.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
Looks good

PR #13054, opened 2023-07-06
feat(hardware): added a script write and interact with the Flex eeprom.

vegano1:
# Overview

We need to store the serial information for the Flex on the eeprom, this script will be used to write and verify that value along with any other data we might want to write to the eeprom.

Closes: [RET-1347](https://opentrons.atlassian.net/browse/RET-1347) & [RET-1349](https://opentrons.atlassian.net/browse/RET-1349)

# Test Plan

- [x] Write the serial number a Flex and verify that the value is stored after reboot

# Changelog

- added eeprom_writter.py script to read/write/print eeprom data on the Flex

# Review requests

- usage should be straightforward but let me know if there's anything we might need

# Risk assessment

Low

[RET-1347]: https://opentrons.atlassian.net/browse/RET-1347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RET-1349]: https://opentrons.atlassian.net/browse/RET-1349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hardware/opentrons_hardware/scripts/eeprom_writter.py
```diff
@@ -0,0 +1,293 @@
+"""This script is able to read/write to/from the Flex EEPROM.
```
sfoster1: nit: all of the detail about what arguments do should be in the argument help text instead of here so we don't have to write them twice

hardware/opentrons_hardware/scripts/eeprom_writter.py
```diff
@@ -0,0 +1,293 @@
...
+def write_eeprom(
+    eeprom_api: EEPROMDriver, properties: Dict[PropId, Any]
+) -> Tuple[bool, str]:
+    """Write properties to the eeprom, this will clear any data already stored."""
+    if not properties:
+        print(
+            "Need at least one valid property and value to write.\n"
+            "Pass in the '--help' arg for an example."
+        )
+        exit(1)
```
sfoster1: nit: in a script i generally prefer to get most things being exceptions rather than `sys.exit` so the functions are usable elsewhere

sfoster1 (changes requested):
`eeprom_writer` not `eeprom_writter` and some nits, ut otherwise looks good!

hardware/opentrons_hardware/scripts/eeprom_writter.py
vegano1: We print out this module doc string using the __doc__ variable as part of the argParser description whenever we pass in --help.

hardware/opentrons_hardware/scripts/eeprom_writer.py
```diff
@@ -0,0 +1,303 @@
...
+    address = 0
+    pages = eeprom_api.size // DEFAULT_READ_SIZE
+
+    # set the write bit low
+    eeprom_api._gpio.activate_eeprom_wp()
+
+    # write
+    try:
+        for idx in range(pages):
+            data = b"\xff" * DEFAULT_READ_SIZE
```
laviera: this "data= " statement should be outside of the for loop

hardware/opentrons_hardware/scripts/eeprom_writer.py
```diff
@@ -0,0 +1,303 @@
...
+        msg = (
+            "Need at least one valid property and value to write.\n"
+            "Pass in the '--help' arg for an example."
+        )
+        raise RuntimeError(msg)
+
+    # clear the eeprom first
+    success, msg = clear_eeprom(eeprom_api)
+    if not success:
+        msg = f"Error writting to eeprom: Could not clear eeprom - {msg}"
```
laviera: nit pick - spelling writting == writing

laviera:
@vegano1 - Should there be a "print raw" - allowing us to just dump the raw eeprom

hardware/opentrons_hardware/scripts/eeprom_writter.py
sfoster1: Yeah but `--help` will also print out the helpstring for each argument on its own, so itll be duplicated

sfoster1 (approved):
Looks good overall!

PR #13058, opened 2023-07-07
fix(hardware): stop waiting for moves after move-to-stall

sfoster1:
When we command move groups with a stall stop condition, if the stop condition occurs in the first or second move the microcontroller will cancel the remaining moves in the group. The move group runner needs to not wait for those moves since they'll never occur.

Closes RQA-1043

## Testing
- [ ] put on a robot and run some pick up tips


Laura-Danielle (approved):
makes sense to me

sfoster1:
Ran the problematic protocol that discovered the bug and all good!

PR #13060, opened 2023-07-07
refactor(protocol-engine): Delete AbstractSyncTransport

SyntaxColoring:
# Overview

This is a trivial refactor based on prior discussion with @TamarZanzouri, @sanni-t, and @jbleon95.

`AbstractSyncTransport` and `ChildThreadTransport` are internal classes that are involved in executing PAPIv≥2.14 protocols. `AbstractSyncTransport` is the abstract interface, and `ChildThreadTransport` is the concrete implementation.

These days, we feel that `AbstractSyncTransport` is a layer of indirection and abstraction that doesn't have much of a benefit. Since these classes were originally written ~2 years ago, no concrete implementations besides `ChildThreadTransport` have materialized. So let's just delete it.

This helps out with PR #12965 by reducing the number of things that we need to explain.

# Test Plan

None needed.

# Review requests

None in particular.

# Risk assessment

Very low.

### DIFF
--- a/api/src/opentrons/protocol_engine/clients/__init__.py
+++ b/api/src/opentrons/protocol_engine/clients/__init__.py
@@ -1,5 +1,5 @@
 """ProtocolEngine clients."""
 from .sync_client import SyncClient
-from .transports import AbstractSyncTransport, ChildThreadTransport
+from .transports import ChildThreadTransport
 
-__all__ = ["SyncClient", "AbstractSyncTransport", "ChildThreadTransport"]
+__all__ = ["SyncClient", "ChildThreadTransport"]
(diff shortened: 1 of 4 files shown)


sfoster1:
@SyntaxColoring We will, I swear, get these dang things out of process eventually and have a `ChildProcessTransport` someday but I guess at that point that would be the only one?

SyntaxColoring:
> @SyntaxColoring We will, I swear, get these dang things out of process eventually and have a `ChildProcessTransport` someday

Hmm, that's a good point that I hadn't considered.

> but I guess at that point that would be the only one?

I think so, yeah.

TamarZanzouri (approved):
Thank you for making this change!

PR #13063, opened 2023-07-10
fix(app, labware library): prevent white screen when opening adapter definitions

shlokamin:
# Overview

This PR fixes a bug in both LL + the opentrons app where clicking on an adapter would cause a white screen. It also moves the measurement guide assets into the components library since they are identical between LL and the app.

Not sure what to do about the images for an adapter's spacing (probably need new assets) but for now I think this is fine.

# Review requests

Go to an adapter's detail page on both the app and LL, there should no longer be a white screen.

# Risk assessment

Low


jerader (approved):
lgtm! unrelated but we probably should add an adapter option to the category filter options.

Screen Shot 2023-07-11 at 9 10 56 AM

PR #13064, opened 2023-07-10
fix(hardware-testing): Remove tips using them for finding liquid height

andySigler:
# Overview

Using the same tips to find liquid height as well as do Trial 1's aspiration is causing error on that first trial. This PR makes it so that the tips used for finding the liquid height are entirely different tips than what is used for Trial 1.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/hardware-testing/.flake8
+++ b/hardware-testing/.flake8
@@ -5,8 +5,8 @@
 max-line-length = 100
 
 # max cyclomatic complexity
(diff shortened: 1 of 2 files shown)


ryanthecoder:
This will break doing 1000ul tip tests. Which I now we aren't going to do anymore but just a heads up

ryanthecoder (approved):
This looks good to me. You're fine to add the ignore too complex tag for now so it passes lint. My big refactor of this is coming together and it drastically reduces complexity anyway

PR #13067, opened 2023-07-10
feat(robot-server): add e-stop state query & acknowledge-disengage endpoints

fsinapi:
# Overview

Adds placeholder endpoints and models for the Estop State Query and Estop Acknowledge-Disengage endpoints.

- The state query endpoint returns the overall estop status for the robot, as well as the physical status of each of the estop mount points
- The acknowledge-disengage endpoint is intended to move the status from `logically_engaged` over to `not_present` or `disengaged`, depending on the attachment status of the estops.
  - This endpoint also returns the same information as the query endpoint to eliminate the need for another query after PUTting

Right now, these don't do anything - they will just return placeholder information until the estop is integrated into the hardware controller. I'm putting up this PR before the actual backend work so that, once the estop state machine is set up, there will be a way to actually clear it by hooking up the acknowledge-disengage endpoint.

# Test Plan

Made sure the models look right in a local dev server, but mostly tested through the simple integration tests.

# Changelog

- Added GET estop state endpoint
- Added PUT estop status acknowledge-disengage endpoint

# Review requests

# Risk assessment

Quite low, two new endpoints that aren't hooked up to anything yet.


sfoster1 (approved):
Looks good to me, I think this is the right way to go about this work too

koji (commented):
Nice 🚀 
lgtm

Laura-Danielle (approved):
As a Pydantic User this lgtm

robot-server/robot_server/robot/control/models.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import List
+
+
+class EstopState(enum.Enum):
+    """Current status of the estop on this robot."""
+
+    not_present = "not_present"
+    physically_engaged = "physically_engaged"
+    logically_engaged = "logically_engaged"
+    disengaged = "disengaged"
```
SyntaxColoring: Nitpick: Can we capitalize these enums like `NOT_PRESENT = "notPresent"`? (UPPERCASE for Python constants and camelCase for HTTP constants.)

robot-server/robot_server/robot/control/models.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class EstopStatusModel(BaseModel):
+    """Model for the current estop status."""
+
+    status: EstopState = Field(
+        ..., description="The current status of the estop on this robot."
+    )
+
+    physical_status: List[EstopPhysicalStatusModel] = Field(
```
SyntaxColoring: Yeah, I softly prefer we make this a `left`/`right` pair instead of a list, unless someone else knows of a reason to do otherwise.

robot-server/robot_server/robot/control/router.py
```diff
@@ -0,0 +1,81 @@
...
+    },
+)
+async def get_estop_status(
+    thread_manager: ThreadManagedHardware = Depends(get_thread_manager),
+) -> PydanticResponse[SimpleBody[EstopStatusModel]]:
+    """Return the current status of the estop."""
+    return await _get_estop_status(thread_manager)
+
+
+@control_router.post(
```
SyntaxColoring: I think this should probably be an HTTP `PUT`, not `POST`.

We're "replacing" (acknowledging) an existing "resource" (`acknowledgeEstopDisengage`), not appending a new sub-resource under `acknowledgeEstopDisengage`.

SyntaxColoring (commented):
A few nitpicks, but this otherwise looks good to me.

Note that this is missing unit tests for the new endpoints. I think that's fine for now—they wouldn't add anything over the integration tests—but we should be careful about it, as this code grows.

Thanks for remembering to do `.construct()`!

robot-server/robot_server/robot/control/models.py
fsinapi: My main thinking was that a list with named entries is much more adaptable for any theoretical alternative hardware configurations using the same interface. But we can cross that bridge when we get to it 🤷

robot-server/robot_server/robot/control/router.py
fsinapi: I think we're stretching the intentions of these methods no matter how you slice it, but I agree that PUT makes more sense here than POST

fsinapi: 42bf40cbcadd2270a18b94d6aa5df6a9f1a243de

robot-server/robot_server/robot/control/models.py
fsinapi: e82c770af7fc690451e71dc44c18d0b34f6cc717

robot-server/robot_server/robot/control/models.py
fsinapi: 84304cd5a4948e65ca526ef0bbab3ad0b7b6b107

PR #13071, opened 2023-07-10
chore(refactor): Refactor and cleanup volumetric scripts

ryanthecoder:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


hardware-testing/hardware_testing/gravimetric/__main__.py
```diff
@@ -235,8 +229,9 @@ def run_photometric(
         deck_version="2",
         extra_labware=custom_defs,
     )
+    union_cfg: Union[PhotometricConfig, GravimetricConfig]
```
sfoster1, line 191: nit: we could factor this into like
```
def config_from_args(args: argparse.Namespace, context: Whatever_Ctx_is) -> Union[PhotometricConfig, GravimetricConfig]`
```
which means we wouldn't have to do the temporarily-unbound variable thing and could test it easily if we ever added tests

hardware-testing/hardware_testing/gravimetric/__main__.py
```diff
@@ -266,3 +261,35 @@ def run_photometric(
...
+        run_id=run_id,
+        start_time=start_time,
+        operator_name=helpers._get_operator_name(_ctx.is_simulating()),
+        robot_serial=helpers._get_robot_serial(_ctx.is_simulating()),
+        tip_batch=helpers._get_tip_batch(_ctx.is_simulating()),
+        git_description=get_git_description(),
+        tips=get_tips(_ctx, pipette, all_channels=all_channels_same_time),
+    )
+
+    if args.photometric:
```
sfoster1, line 258: If we added an element to each config called `kind: Literal['photometric']` and `kind: Literal['gravimetric']` then you'd be able to have this be a discriminated union which means you could do `if cfg.kind == 'photometric'` and mypy would figure it out. You could also bind the run function into the config object.

hardware-testing/hardware_testing/gravimetric/execute.py
```diff
@@ -613,64 +321,137 @@ def run(ctx: ProtocolContext, cfg: config.GravimetricConfig) -> None:
...
+def _calculate_evaporation(
+    cfg: config.GravimetricConfig,
+    resources: TestResources,
+    recorder: GravimetricRecorder,
+    liquid_tracker: LiquidTracker,
+    test_report: report.CSVReport,
+    labware_on_scale: Labware,
+    measure_height: float,
+) -> Tuple[float, float]:
+    if not cfg.blank or cfg.inspect:
```
sfoster1: recommend factoring the body of each `if`/`else` into its own function to save an indent level

hardware-testing/hardware_testing/gravimetric/execute.py
```diff
@@ -679,69 +460,37 @@ def run(ctx: ProtocolContext, cfg: config.GravimetricConfig) -> None:
...
+            resources.test_volumes,
+            channels_to_test,
+            recorder,
+            test_report,
+            liquid_tracker,
+            False,
+            measure_height=measure_height,
+        )
+        for volume in trials.keys():
             actual_asp_list_all = []
```
sfoster1, line 515: Might be worth getting these for loops to each be their own function

hardware-testing/hardware_testing/gravimetric/helpers.py
```diff
@@ -104,3 +115,273 @@ def get_pipette_unique_name(pipette: protocol_api.InstrumentContext) -> str:
...
+    location: Location,
+) -> None:
+    print(
+        f"picking tip {location.labware.as_well().well_name} "
+        f"from slot #{location.labware.parent.parent}"
+    )
+    pipette.pick_up_tip(location)
+    # NOTE: the accuracy-adjust function gets set on the Pipette
+    #       each time we pick-up a new tip.
+    if getattr(cfg, "increment", False):
```
sfoster1: does it need to be a `getattr`?

hardware-testing/hardware_testing/gravimetric/helpers.py
```diff
@@ -104,3 +115,273 @@ def get_pipette_unique_name(pipette: protocol_api.InstrumentContext) -> str:
...
+    assert pipette.max_volume == cfg.pipette_volume, (
+        f"expected {cfg.pipette_volume} uL pipette, "
+        f"but got a {pipette.max_volume} uL pipette"
+    )
+    if hasattr(cfg, "gantry_speed"):
+        pipette.default_speed = getattr(cfg, "gantry_speed")
+
+    # NOTE: 8ch QC testing means testing 1 channel at a time,
+    #       so we need to decrease the pick-up current to work with 1 tip.
+    if pipette.channels == 8 and not getattr(cfg, "increment", False):
```
sfoster1: does it need to be a `getattr`?

sfoster1 (approved):
This is looking really solid. I think we could make it a little better by splitting some massive triple for loops into functions, but really my most specific rec is: we should get replace `print` and `input` with a call to that `ui` module. It'll really help if we ever want to add fanciness there.

hardware-testing/hardware_testing/gravimetric/helpers.py
```diff
@@ -104,3 +114,247 @@ def get_pipette_unique_name(pipette: protocol_api.InstrumentContext) -> str:
...
+    pipette: InstrumentContext, return_tip: bool, minimum_z_height: int = 0
+) -> None:
+    if return_tip:
+        pipette.return_tip(home_after=False)
+    else:
+        pipette.drop_tip(home_after=False)
+    if minimum_z_height > 0:
+        cur_location = pipette._get_last_location_by_api_version()
+        if cur_location is not None:
+            pipette.move_to(cur_location.move(Point(0, 0, minimum_z_height)))
```
andySigler, line 266: where does this happen?

hardware-testing/hardware_testing/gravimetric/helpers.py
```diff
@@ -104,3 +114,247 @@ def get_pipette_unique_name(pipette: protocol_api.InstrumentContext) -> str:
...
+    if hasattr(cfg, "gantry_speed"):
+        pipette.default_speed = getattr(cfg, "gantry_speed")
+
+    # NOTE: 8ch QC testing means testing 1 channel at a time,
+    #       so we need to decrease the pick-up current to work with 1 tip.
+    if pipette.channels == 8 and not cfg.increment:
+        hwapi = get_sync_hw_api(ctx)
+        mnt = OT3Mount.LEFT if cfg.pipette_mount == "left" else OT3Mount.RIGHT
+        hwpipette: Pipette = hwapi.hardware_pipettes[mnt.to_mount()]
+        hwpipette.pick_up_configurations.current = 0.2
```
andySigler, line 321: we should do something to call this out more, maybe pull its definition up to somewhere more visible in a followup PR

hardware-testing/hardware_testing/gravimetric/trial.py
```diff
@@ -0,0 +1,287 @@
...
+                    channel=0,
+                    channel_count=num_channels_per_transfer,
+                    trial=trial,
+                    recorder=recorder,
+                    test_report=test_report,
+                    liquid_tracker=liquid_tracker,
+                    blank=blank,
+                    inspect=cfg.inspect,
+                    mix=cfg.mix,
+                    stable=False,
```
andySigler: should be `True` I think

hardware-testing/hardware_testing/gravimetric/execute.py
```diff
@@ -615,130 +330,172 @@ def run(ctx: ProtocolContext, cfg: config.GravimetricConfig) -> None:
...
-        print("moving to scale")
+        ui.print_info("moving to scale")
         well = labware_on_scale["A1"]
-        pipette.move_to(well.top(0))
-        _liquid_height = _jog_to_find_liquid_height(ctx, pipette, well)
+        resources.pipette.move_to(well.top(0), minimum_z_height=_minimum_z_height(cfg))
+        _liquid_height = _jog_to_find_liquid_height(
+            resources.ctx, resources.pipette, well
+        )
+        resources.pipette.move_to(well.top().move(Point(0, 0, _minimum_z_height(cfg))))
```
andySigler, line 469: I don't think this is necessary b/c the following movements will move the tip away from the via. Also I think the `Location.move()` method takes a relative xyz point, not an absolute value

hardware-testing/hardware_testing/gravimetric/execute.py
```diff
@@ -615,130 +330,172 @@ def run(ctx: ProtocolContext, cfg: config.GravimetricConfig) -> None:
...
+            cfg,
+            labware_on_scale["A1"],
+            resources.test_volumes,
+            channels_to_test,
+            recorder,
+            test_report,
+            liquid_tracker,
+            False,
+        )
+        for volume in trials.keys():
```
andySigler, line 514: is this guaranteed to be in the configured order (lowest -> highest)?

hardware-testing/hardware_testing/gravimetric/execute.py
```diff
@@ -749,53 +506,37 @@ def run(ctx: ProtocolContext, cfg: config.GravimetricConfig) -> None:
             trial_disp_dict: Dict[int, List[float]] = {
                 trial: [] for trial in range(cfg.trials)
             }
-            for channel in parameters.channels_to_test:
-                if cfg.isolate_channels and (channel + 1) not in cfg.isolate_channels:
-                    print(f"skipping channel {channel + 1}")
-                    continue
+            for channel in trials[volume].keys():
```
andySigler, line 525: same here

andySigler (approved):
looks good other than the `stable=True` part 👍

PR #13072, opened 2023-07-10
fix(hardware): specify nodes when clearing move groups

fsinapi:
# Overview

The call to `ensure_send` when clearing move groups didn't specify what nodes to listen for and it ended up printing out a ton of errors. This should eliminate those errors.

# Test Plan

Run ot3repl on an OT-3 with this change, make sure the warning messages don't pop up.

# Changelog

- Specify expected CAN nodes when clearing move groups.

# Review requests

# Risk assessment


ryanthecoder (approved):
lgtm

ahiuchingau (approved):
Good call

sfoster1 (approved):
Looks good to me!

PR #13073, opened 2023-07-10
refactor(shared-data): group slot clips in deck def rendering

shlokamin:
# Overview

This PR groups slot clips by slot so we can easily hide them as a unit, as well as renames the CSS style classes embedded in the deck def to be semantic.

# Test Plan

Verified deckmap renderings still look like they should

# Review requests
- Look at deck map renderings and make sure they look like they should
- Inspect the deck map svg and make sure the slot clip layers show up
# Risk assessment
Low


brenthagen (approved):
🖇️

jerader (approved):
🤩 nice!

PR #4574, opened 2019-12-09
feat(api): Handle extra labware in script entrypoints

sfoster1:
Previously, extra_labware (the mechanism for v2 custom labware) wasn't really
handled in the script entrypoints opentrons_execute and opentrons_simulate. This
adds that handling. Specifically,

- opentrons_execute learned -L, which lets you specify paths to search for
labware definitions (previously only known by opentrons_simulate for building
bundles). This defaults to '.', as in opentrons_simulate
- opentrons.{simulate,execute}.get_protocol_api learned the extra_labware
argument, which should be a dict mapping labware names to labware definitions.
If this is not specified (as in, not as an empty dict), if this is a robot the
software will look in /var/lib/jupyter/notebooks/labware for custom labware.
- Add the jupyter lookup thing to docs. This is the canonical "right place" to
put custom labware defs you want to use in jupyter.
- Add some tests.

## Testing
Use the simulate and execute entrypoints:
- `opentrons_simulate` as a script from a notebook terminal should look in the current directory for custom labware files. You should be able to specify more with `-L`. If there are json files that aren't labware, nothing should break. This should all be true both on a local computer and a robot, including a robot jupyter notebook.
- `opentrons_execute` as a script from a terminal should do the same thing
- if you import `opentrons.simulate.get_protocol_api` or `opentrons.execute.get_protocol_api` in a jupyter terminal or normal python prompt on a robot, it should look in `/var/lib/jupyter/notebooks/labware` for custom labware defs. Nothing should break if there are json files that aren't labware in there. Nothing should break if the directory doesn't exist.

PR #13074, opened 2023-07-10
fix(api): Fix opentrons_execute ignoring args for custom labware and data

### LINKED ISSUE #9256: Unable to load custom labware with opentrons_execute
A user notes that he is unable to load custom labware with opentrons_execute in ssh.

Original report:

A user was "try and run a protocol from the command line using a custom labware file, and was using a simple protocol to test out the functionality (see attached protocol and custom labware file).  The command used to execute the protocol and use the custom labware file: opentrons_execute /data/user_storage/Customlabwareexecution.py -L /data/user_storage (the location of abgene_96_wellplate_330ul.json).  Running this command throws an Unable to find a labware definition for "abgene_96_wellplate_330ul", etc. error.  Interestingly, when running opentrons_simulate /data/user_storage/Customlabwareexecution.py -L /data/user_storage, it simulates with no issues.  We have looked over the formatting of the json, tried a number of different custom labware directories, etc., but no luck with opentrons_execute."

This is similar to #7134, except this one shows execute not working and simulate working, where as 7134 shows simulate not working

SyntaxColoring:
# Changelog

* Fix a bug where the `--custom-labware-path` (`-L`), `--custom-data-path` (`-D`), and `--custom-data-file`) (`-d`) arguments to `opentrons_execute` had no effect.
    * This closes #9256. :)
    * This bug appears to have been present since #4574.
* In the `--help` text, clarify how `--custom-labware-path` (`-L)` affects the default behavior of searching the current working directory.
* Refactor: Remove some unnecessary `getattr()` calls to read attributes that should always be present.

# Test Plan

* [x] Run `opentrons_execute -L <dir>` and make sure it searches the directory you provide, in addition to the Jupyter Notebook directory.
* [x] Run `opentrons_execute` without `-L` and make sure it searches just the Jupyter Notebook directory.
* [x] Test `--custom-data-file` and `--custom-data-path` with `opentrons_execute` and `opentrons_simulate` to ensure it still works.

# Review requests

Does anyone understand what this is doing in the specification for  `--custom-data-path`?

```python
    parser.add_argument(
        "-D",
        "--custom-data-path",
        action="append",
        nargs="?",
        const=".",
        default=[],
        ...
```

What are `const` and `nargs` doing there? Why aren't they also present on `--custom-labware-path` or `--custom-data-file`?

# Risk assessment

Medium? Testing this exhaustively has to be done manually.

### DIFF
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -413,6 +413,8 @@ def main() -> int:
     execute(
         protocol_file=args.protocol,
         protocol_name=args.protocol.name,
+        custom_labware_paths=args.custom_labware_path,
+        custom_data_paths=(args.custom_data_path + args.custom_data_file),
         log_level=log_level,
         emit_runlog=printer,
     )
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -188,8 +188,9 @@ def get_arguments(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
         "Only directories specified directly by "
         "this argument are searched, not their children. JSON files that "
         "do not define labware will be ignored with a message. "
-        "By default, the current directory (the one in which you are "
-        "invoking this program) will be searched for labware.",
+        "The current directory (the one from which you are "
+        "invoking this program) will always be included implicitly, "
+        "in addition to any directories that you specify.",
     )
     parser.add_argument(
         "-D",
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -411,8 +411,8 @@ def main() -> int:
         log_level = "warning"
     # Try to migrate containers from database to v2 format
     execute(
-        args.protocol,
-        args.protocol.name,
+        protocol_file=args.protocol,
+        protocol_name=args.protocol.name,
         log_level=log_level,
         emit_runlog=printer,
     )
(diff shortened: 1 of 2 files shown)


sfoster1:
> Does anyone understand what this is doing in the specification for `--custom-data-path`?
> 
> ```python
>     parser.add_argument(
>         "-D",
>         "--custom-data-path",
>         action="append",
>         nargs="?",
>         const=".",
>         default=[],
>         ...
> ```
> What are `const` and `nargs` doing there? Why aren't they also present on `--custom-labware-path` or `--custom-data-file`?

`nargs` on an option string and `const` in combination have some funny behavior documented [here](https://docs.python.org/3/library/argparse.html#nargs):
> '?'. One argument will be consumed from the command line if possible, and produced as a single item. If no command-line argument is present, the value from [default](https://docs.python.org/3/library/argparse.html#default) will be produced. Note that for optional arguments, there is an additional case - the option string is present but not followed by a command-line argument. In this case the value from [const](https://docs.python.org/3/library/argparse.html#const) will be produced.

So it lets you specify `-D` on its own without an argument (in which case the value is `'.'`); or with an argument as `-D some/path` (in which case the value is `'some/path'`); or not at all in which case the value is `default` aka in this case `[]`. So you can change behavior based on the presence or absence of the argument while still providing a default.

sfoster1:
And the `default` is `[]` because if `nargs` is specified then the argument value is always stored as a list, and `const` is `'.'` because that becomes an element of the list

sfoster1 (approved):
Looks good to me! Love deep argparse wizardry

PR #13076, opened 2023-07-11
fix(api): Switch plunger backlash to pre-load for aspirations

andySigler:
# Overview

This PR changes backlash to ensure that all aspirations are pre-loaded.

Test data is showing that aspiration volumes are varying more between pipettes with the currently implemented backlash compensation. This is likely due to the variations between each pipette's true backlash distance causing both differences in aspiration distances. This is especially true at low volumes, where we are seeing large variations in volumes between pipettes.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Okay, talked through it in person and aspirate-preload makes sense and the implementation looks right. Nice!

PR #13077, opened 2023-07-11
fix(update-server): persist the machine-id after writing the rootfs to the unused partition.

vegano1:
# Overview

The `/etc/machine-id` on the Flex is not persisting through updates since the file gets overwritten during an update, this pr fixes that by writing the machine-id to the unused partition after the rootfs has been written to it.

This needs [or-core-pr-87](https://github.com/Opentrons/oe-core/pull/87) openemdedded change to work 

# Test Plan

- [x] Make sure that the UUID in `/etc/machine-id` persists after an update

# Changelog

- After the new rootfs has been written to the unused partition, read the old `/etc/machine-id` and write it to the new rootfs.

# Review requests

# Risk assessment
Low, this changes a small part of the update mechanism.

### DIFF
--- a/update-server/otupdate/openembedded/update_actions.py
+++ b/update-server/otupdate/openembedded/update_actions.py
@@ -266,7 +266,10 @@ class OT3UpdateActions(UpdateActionsInterface):
 
     def write_machine_id(self, current_root: str, new_root: str) -> None:
         """Copy the machine id over to the new partition"""
-        pass
+        mid = open(os.path.join(current_root, "etc", "machine-id")).read()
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Ooh interesting. This looks correct to me and should get merged but I might also pop it over to a new alpha if we do one

vegano1:
@sfoster1 
Sounds good, once I test this ill merge it in.

vegano1:
@sfoster1 

[This](https://github.com/Opentrons/oe-core/pull/87) openembedded change is required to persist the machine-id, it bind-mounts `/mnt` -> `userfs/var/mnt` so the update-server change above can write the machine-id to the unused partition.

PR #13082, opened 2023-07-11
feat(api, app, shared-data, pd): use new flex pipette names in backend & clients

sanni-t:
# Overview

#12966 updated the Flex's pipette names from `.._gen3` to `.._flex` which created a conflict with pipette names expected by the engine & PAPI. This PR updates the engine to switch to using the new names, while adding a *temporary* backwards compatible shim to python API to convert the old names to new ones.
 
Ticket to remove the backwards compatibility shim before launch: RSS-284

Update after PR reviews: 
- also added new PAPI-only pipette names from [this comment.](https://opentrons.atlassian.net/browse/RSS-261)
- replaced `gen3` references in v1 pipette definitions to `flex`
- added an integration test on robot server to test a minimal flex protocol

# Test Plan

- Test on a flex bot and make sure:
- [x] you can calibrate pipettes and previously calibrated pipettes show the correct calibration
- [x] you can analyze & run python protocols that load pipettes with both the old and the new names
- [x] protocol setup is able to detect & match the correct pipette attached
- [x] LPC runs correctly

# Review requests

This approach would cause existing analyses & run logs on the robot to raise errors. It would require users to reset their database. Is it okay to have to do that?

# Risk assessment

This is basically a fix for the breaking change introduced in the above linked PR so that way it reduces risk, but it also fixes it by switching the engine over to use the new names so we should be careful with merging it in.


SyntaxColoring (approved):
Code changes make sense to me. Thanks for your quick response on this.

> This approach would cause existing analyses & run logs on the robot to raise errors. It would require users to reset their database. Is it okay to have to do that?

👍 Yeah, this is the same sort of thing as with #12571. If we have to have a breaking API change like this, we may as well rip off the band-aid while it's still internal.

shlokamin (commented):
when running analysis on a protocol that uses a pipette load name of `p1000_single_flex` i get the following analysis error: 

```
ProtocolEngineError [line 28]: Error 4000 GENERAL_ERROR (ProtocolEngineError): UnexpectedProtocolError: 'p1000_single_flex_v1'
```

Laura-Danielle (changes requested):
Since you already have this PR open, can you go ahead and do the requirements listed in [this ticket](https://opentrons.atlassian.net/browse/RSS-261)?

Make sure the "gen3" version of the pipettes get passed into the hardware controller.

shared-data/python/opentrons_shared_data/pipette/dev_types.py
```diff
@@ -21,16 +21,16 @@
     "p20_multi_gen2",
     "p50_single",
     "p50_multi",
-    "p50_single_gen3",
-    "p50_multi_gen3",
+    "p50_single_flex",
```
Laura-Danielle, line 24: Let's keep the `gen3` in these dev types for now and also test on a robot w/ app to see if we need to additionally add the `_flex` models to schema v1 for pipettes (hopefully we do not).

Laura-Danielle (approved):
Approving for now, but see comment I added.

sanni-t (commented):
Successfully tested pipette attach, calibration, LPC & running python protocol with `flex_8channel_1000` & `p50_single_gen3` pipettes without any issues!

PR #13084, opened 2023-07-12
fix(api): Do not change pipette load names in the backend

Laura-Danielle:
# Overview

#butterflyeffect #mybad

# Test Plan

- [ ] Throw on a robot. Make sure the load names are matching.

### DIFF
--- a/shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
+++ b/shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
@@ -59,6 +59,7 @@ class PipetteModelType(Enum):
 class PipetteGenerationType(Enum):
     GEN1 = "GEN1"
     GEN2 = "GEN2"
+    GEN3 = "GEN3"  # equivalent to FLEX
     FLEX = "FLEX"
 
 
--- a/shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
+++ b/shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
@@ -33,6 +33,8 @@ class PipetteNameType:
             return base_name
         elif self.pipette_channels == PipetteChannelType.NINETY_SIX_CHANNEL:
             return base_name
+        elif self.pipette_generation == PipetteGenerationType.FLEX:
+            return f"{base_name}_{PipetteGenerationType.GEN3.name.lower()}"
         else:
             return f"{base_name}_{self.pipette_generation.name.lower()}"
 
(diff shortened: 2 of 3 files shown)


shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -59,6 +59,7 @@ class PipetteModelType(Enum):
 class PipetteGenerationType(Enum):
     GEN1 = "GEN1"
     GEN2 = "GEN2"
+    GEN3 = "GEN3"  # equivalent to FLEX
```
SyntaxColoring, line 62: This seems like a red flag to me. This is an internal implementation-detail enum, isn't it? Why do we need to have both values? What internal implementation-detail code is using `GEN3` that can't be changed to use `FLEX`?

Laura-Danielle: This enum is currently used for two things:

1. To support `PipetteName` like shapes (i.e. `p300_single_gen3`) which we'll keep around for now until we can deal with simulating pipette heights better.
2. The generation itself from a schema perspective should be `FLEX`, but we'll want to return `p1000_single_gen3` to things like the Protocol Engine so it should map FLEX <-> GEN3.

We honestly don't have time for a refactor required to match the the new user-facing load names to `FLEX` which is why we have both here.

sanni-t: I agree with @SyntaxColoring. It seems to me like the only place we need to keep the 'gen3' backwards compatibility for internal testing is in the python API, which we can easily do by mapping the old name to the new name in that one place only. Then everything from the API implementations to engine and onwards deals with the new names. #13082 already updates those things for API & engine. 

> To support PipetteName like shapes (i.e. p300_single_gen3) which we'll keep around for now until we can deal with simulating pipette heights better.

Interesting. So sounds like we use hardcoded pipette names for simulating the heights? Wouldn't that also be easier to just change the names?

> The generation itself from a schema perspective should be FLEX, but we'll want to return p1000_single_gen3 to things like the Protocol Engine so it should map FLEX <-> GEN3.

I don't think we need to do this. If we make the engine expect the new names, we don't have to give it old names.

Laura-Danielle: As in I don't want to make changes required for the hardware controller to handle multiple name styles. The protocol engine should just map new names to `XXX_gen3` and we'll handle the rest later.

PR #13085, opened 2023-07-12
refactor(api): configure instrs only if needed

sfoster1:
We've had this consistent problem with race conditions around cache_instruments, which gets called when something hits /instruments. It can reconfigure a bunch of internal data caches and also do things like change the active currents on the axes. We used to do this pretty often because it was the only time we actually checked what instruments were connected. We configured everything unconditionally.

However, the app is basically always polling /instruments. That means that if a process is running - a protocol, a calibration - some app is going to hit /instruments and cause a current reconfiguration and anything that requires a specific current will break, for instance. This is bad!

The fix is to add in a concept from the OT2, of only running a system reconfiguration if something actually changed since the last cache_pipette. It's really easy to make this comparison, since we cache stuff anyway - that's the point of cache pipette - and a lot of the worker functions we call during the process actually return a bool to let us know we can skip config anyway.

## Review requests
- Does this make sense?

## Testing
This is a change in a really sensitive area and so requires a fair amount of testing. It's intended to fix an issue where starting recalibration for a 96-channel pipette would cause the z to drop immediately because a call to `reset()` would try and reconfigure but hit the cache lock and set a current that was too low, so we should test that, but we should also test a bunch of other swap-pipette operations.

- [ ] Run a pipette recalibration with a 96-channel attached, while something is polling `GET /instruments`
- [ ] Swap between multiple pipette combinations - 96, gripper, one or two other pipettes
- [ ] Run some other calibration flows while stuff is polling

Closes RQA-1050

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/gripper.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/gripper.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 """ Classes and functions for gripper state tracking
 """
 import logging
-from typing import Any, Optional, Set
+from typing import Any, Optional, Set, Tuple
 
 from opentrons.types import Point
 from opentrons.config import gripper_config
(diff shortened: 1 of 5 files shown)


fsinapi (approved):
I like this change a lot. LGTM

vegano1 (approved):
Oh yeah, this is nice

Laura-Danielle (approved):
OK this LGTM

PR #13087, opened 2023-07-12
fix(update-server): clean up the downloaded update files after an update.

vegano1:
# Overview

The downloaded updated files are kept on the robot indefinitely, even after an update has been completed and the files are no longer needed. This leads to us always holding on to ~500MB+ of stale data on the user partition. This pr addresses this issue by cleaning up the download dir after the rootfs has been successfully written to the unused partition, the update has been committed, and the system is ready to restart.

# Test Plan

- [x] Make sure we can still update the Flex with this change
- [x] Make sure we can still update the OT2 with this change
- [x] Make sure that the download dir `/var/lib/otupdate/downloads` on the FLex is empty after an update is complete
- [x] Make sure that the download dir `/var/lib/otupdate/downloads` on the OT2 is empty after an update is complete

# Changelog

- Added a clean_up update action that removes all files from the otupdate download dir after an update.

# Review requests

- Make sure there arent any obvious issues.

# Risk assessment

Medium, this touches on update code for both the Flex and OT2 and needs to be tested on both.

### DIFF
--- a/update-server/otupdate/buildroot/update_actions.py
+++ b/update-server/otupdate/buildroot/update_actions.py
@@ -179,6 +179,13 @@ class OT2UpdateActions(UpdateActionsInterface):
             new_mid.write(mid)
         LOG.info(f"Wrote machine_id {mid.strip()} to {new_root}/etc/machine-id")
 
+    def clean_up(self, download_dir: str) -> None:
+        """Deletes the update contents in the download dir."""
+        LOG.info(f"Cleaning up download dir {download_dir}.")
+        for file in os.listdir(download_dir):
+            LOG.debug(f"Deleting {file}")
+            os.remove(file)
+
 
 def _find_unused_partition() -> RootPartitions:
     """Find the currently-unused root partition to write to"""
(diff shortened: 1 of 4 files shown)


update-server/otupdate/buildroot/update_actions.py
```diff
@@ -179,6 +179,13 @@ def write_machine_id(self, current_root: str, new_root: str) -> None:
             new_mid.write(mid)
         LOG.info(f"Wrote machine_id {mid.strip()} to {new_root}/etc/machine-id")
 
+    def clean_up(self, download_dir: str) -> None:
+        """Deletes the update contents in the download dir."""
+        LOG.info(f"Cleaning up download dir {download_dir}.")
+        for file in os.listdir(download_dir):
+            LOG.debug(f"Deleting {file}")
+            os.remove(file)
```
sfoster1: this probably wants a universal logging try-except, since a failure would break the update but it's non-essential

update-server/otupdate/openembedded/update_actions.py
```diff
@@ -302,3 +302,10 @@ def decomp_and_write(
         )
         if not success:
             raise RuntimeError(msg)
+
+    def clean_up(self, download_dir: str) -> None:
+        """Deletes the update contents in the download dir."""
+        LOG.info(f"Cleaning up download dir {download_dir}.")
+        for file in os.listdir(download_dir):
+            LOG.debug(f"Deleting {file}")
+            os.remove(file)
```
sfoster1: same here re: try-except that logs

sfoster1 (changes requested):
Let's get some `try`/`except` wrappers around these; they're non critical functionality and if they throw as-is they'd break the update. Otherwise looks great!

sfoster1 (approved):
Looks good to me!

PR #13088, opened 2023-07-12
fix(hardware): bump minimum move timeouts

sfoster1:
Closes RQA-1043


fsinapi (approved):
Yeah I think that'll do it

PR #13089, opened 2023-07-12
fix(app): send custom labware files to flex with protocol

b-cooper:
# Overview

Just as the create run from protocol function spreads in all known custom labware files when creating a protocol record on a robot, the send to robot feature should also include custom labware.

props to @skowalski08 for uncovering this bug 

# Review requests

- upload a new protocol that includes custom labware to a flex via the "send to robot" menu item on the protocol page
- the robot side analysis should succeed on the flex

# Risk assessment
low


skowalski08 (approved):
Tested and verified on 7/12

PR #13091, opened 2023-07-12
test: Flex protocol analysis e2e

DerekMaggio:
# Overview

- Add Flex ABR protocols to Protocol Analysis E2E.
- Add some super minor changes

# Test Plan

- [x] Make sure all Flex protocol tests pass
  - Currently all these tests are not passing for the following reasons:
    - Getting a `Cannot aspirate more than pipette max volume` exception in `ABR5/6 Quick_Zymo_RNA_Bacteria_OT3 (Zach edit).py`
    - `opentrons_ot3_96_tiprack_200ul_temporary_adapter` needs to be updated to rss tip rack for `ABR5/6 Illumina DNA Prep 96x Head PART III.py`

# Changelog

- Add a ton of Flex protocols and labware. All of them should pass without error
- In README, clarify where to find package.json to get electron version
- Specify bash shell for linux get chromedriver script
- Print error text if test has unexpected analysis error

# Review requests

- @y3rsh, any issue with the following?
  - My usage of the protocol naming scheme
  - Clarification in README
  - Binding linux chromedriver script to bash shell
  - Modification to AssertionError if unexpected analysis erorr is hit
- @skowalski08, Can you please comment on this PR when issues with protcols are resolved?
- @nusrat813
  - Do you want any other protocols added at this point? 
  - Do you want these tests added to a nightly CI test? 

# Risk assessment

Low. 

- Tests are not currently being ran in CI
- Mainly just additions with this PR
- Changes to framework are extremely minimal


app-testing/tests/protocol_analyze_test.py
```diff
@@ -40,6 +40,7 @@ def test_analyses(
     # Upload labware if any
     if protocol.custom_labware:
         for labware in protocol.labware_paths:
+            print(labware)
```
y3rsh: So there are cleaner ways to get Rich (the package) output but this is how it is set up right now.
```suggestion
            console.print(labware)
```

y3rsh (changes requested):
Awesome!  So glad this was usable.  Nice work.

- run `make print-protocols` and see what it outputs.  I think we are missing updating teh example.env and it audits stuff for you.
- And `make format-readme` should work too if you have global yarn

y3rsh (approved):
Nice.

PR #13092, opened 2023-07-12
feat(app, api-client): usbPort type to match what robot server returns

jerader:
closes RAUT-544

# Overview

USB port info should show up correctly for module cards and module map view in setup

# Test Plan

verify that the module cards/module setup view looks correct with the correct USBi info

# Changelog

- fix usb port type to match the backend by adding `portGroup` and changing the hub and port types
- fix all affected components and fixtures

# Review requests

see test plan

# Risk assessment

low


api-client/src/modules/api-types.ts
```diff
@@ -16,7 +18,7 @@ export interface ApiBaseModule {
   moduleType: ModuleType
   firmwareVersion: string
   hasAvailableUpdate: boolean
-  usbPort: PhysicalPort
+  usbPort: PhysicalPort | null
```
shlokamin: will this ever be null?

app/src/organisms/Devices/ModuleInfo.tsx
```diff
@@ -42,17 +42,14 @@ export const ModuleInfo = (props: ModuleInfoProps): JSX.Element => {
...
   }
-  if (usbPort === null && hubPort === null && isAttached) {
+  if (physicalPort === null && isAttached) {
     connectionStatus = t('usb_connected_no_port_info')
-  } else if (hubPort === null && usbPort !== null && isAttached) {
-    connectionStatus = t('usb_port_connected', { port: usbPort })
-  } else if (hubPort !== null && isAttached) {
-    connectionStatus = t('hub_port_connected', { port: hubPort })
+  } else if (physicalPort != null && isAttached) {
+    connectionStatus = t('usb_port_connected', { port: physicalPort?.port })
```
shlokamin: since you're already checking if `physicalPort != null` do you need to do the optional chain?

app/src/redux/modules/api-types.ts
```diff
@@ -27,7 +29,7 @@ export interface ApiBaseModule {
   moduleType: ModuleType
   firmwareVersion: string
   hasAvailableUpdate: boolean
-  usbPort: PhysicalPort
+  usbPort: PhysicalPort | null
```
shlokamin: same question above, is this ever gonna be null? also is this the same type as what's in api client? if so should we import that type rather than redeclaring?

shlokamin (approved):
left some comments but overall lgtm. for posterity, we're removing hub info as per @ecormany's guidance. if we decide we want to add hub info back it's an easy addition we can do

api-client/src/modules/api-types.ts
jerader: i was following the api and thought it could be null for some reason, but you're right it doesn't make sense for it to be null. will fix it!

PR #13099, opened 2023-07-13
fix(hardware): additional changes to the eeprom writer script

vegano1:
# Overview

A few small follow-up changes to the eeprom writer script, It now also writes the serial number to `/var/serial` on the filesystem. We also renamed the file to `provision_robot.py` so it matches the rest of the provision scripts since it's doing something similar.

# Test Plan

- [x] Use the script to set the serial number on the Flex and make sure the `/var/serial` is populated.
- [x] Start the robot/update servers and make sure that the serial number shows up in `Advanced -> Robot Serial Number`

# Changelog

- Renamed eeprom_writer.py -> provision_robot.py
- Write the serial number to the `/var/serial` file on the filesystem

# Review requests

# Risk assessment
Low

### DIFF
--- a/hardware/opentrons_hardware/scripts/provision_robot.py
+++ b/hardware/opentrons_hardware/scripts/provision_robot.py
@@ -242,7 +242,7 @@ def _format_properties(properties: List[List[Any]]) -> Dict[PropId, Any]:
     return formated_properties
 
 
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me!

PR #13102, opened 2023-07-13
feat(protocol-designer): add flex trash

jerader:
closes RAUT-543

# Overview

Add flex trash to the deck map and disallow adding labware onto the A3 slot

Screen Shot 2023-07-13 at 5 15 24 PM

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_add-flex-trash/

just test in PD that the flex deck map shows the trash slot in A3 and that you can't add a labware on top of that slot.

# Changelog

- modify `deckSetup` component in PD

# Review requests

see test plan

# Risk assessment

low


koji (approved):

PR #13104, opened 2023-07-14
feat(api): Support V2 pipette configs in the OT-2 and Protocol Engine

Laura-Danielle:
# Overview

Well..uh I am sorry in advance. This PR touches _a lot_ of code. The most important things to check out are the changes to `pipette_data_provider.py` and `ot2/pipette.py`. There are few small changes in the `shared-data/python` code that could also use some eyes. Luckily some of our older tests caught some mistakes in the pipette configs which was great (and hence why there are a lot of shared-data data changes).

I think I'll still need to do one more follow-up PR in Protocol Engine to support dynamically changing fallback tip length during run-time but I decided not to do that for this PR.

# To-Do
There is one more failing test related to supporting "old" and "new aspirate functions on super old pipettes. I haven't yet decided how to handle this, but it would not block folks from beginning to test the PR so I decided to just put this up for review.

# Test Plan

This PR needs to be _thoroughly_ tested on both an OT-2 and Flex. It touches a lot of core functionality including:

- pipette quirks
- pipette backwards compatibility
- tip length lookup
- return tip height
- Aspiration functions
- Current changing

At a minimum we need to test on both an OT2/Flex:

- [x] Attaching pipettes
- [x] Loading pipettes in a protocol (with both old and new Flex names)
- [x] Analyzing a protocol
- [x] Running a protocol on the robot
- [x] Simulating a protocol in the App

Specific to OT-2
- [x] another quick check of mutable configs on the OT2

# Changelog
- Lots of small shared-data data changes (sorry about that)
- Update OT-2 configs to use new shared-data pipette configs
- Update Protocol Engine to grab data from the V2 configs
- Additional helper functions in shared-data to support the two points above
- Small fix to mutable configurations

# Review requests

Please look closely at changes to `ot2/pipette` and `pipette_data_provision.py`.  Make sure I did not miss any critical components in Protocol Engine for loading pipette configs.

# Risk assessment

High. This PR affects a critical part of the stack. Please see the test plan above.


api/src/opentrons/hardware_control/backends/simulator.py
```diff
@@ -119,25 +123,31 @@ def __init__(
...
+                    }
             raise KeyError(
                 "If you specify attached_instruments, the model "
                 "should be pipette names or pipette models, but "
                 f'{passed_ai["model"]} is not'
             )
 
         self._attached_instruments = {
-            m: _sanitize_attached_instrument(attached_instruments.get(m))
+            m: _sanitize_attached_instrument(attached_instruments.get(m))  # type: ignore
```
sfoster1, line 150: what's this suppressing?

api/src/opentrons/hardware_control/instruments/instrument_abc.py
```diff
@@ -27,11 +27,39 @@ def reload_configurations(self) -> None:
...
         ...
 
     @abstractmethod
     def critical_point(self, cp_override: Optional[CriticalPoint] = None) -> Point:
         """Computate critical point of an instrument."""
         ...
+
+
+# TODO (lc 07-05-2023) Move this back to the combined Pipette object file.
+def piecewise_volume_conversion(
```
sfoster1: this really should not be here. we should avoid touching it if we possibly can and if we can't we should just move it once

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
```diff
@@ -59,97 +73,189 @@ class Pipette(AbstractInstrument[pipette_config.PipetteConfig]):
...
-        self._config_as_dict = asdict(config)
-
-    def act_as(self, name: PipetteName) -> None:
+        # TODO Clean this lookup up!
+        self._active_tip_settings = self._config.supported_tips.get(
+            PipetteTipType(self._working_volume),
+            self._config.supported_tips[list(self._config.supported_tips.keys())[0]],
+        )
+        self._fallback_tip_length = self._active_tip_settings.default_tip_length
+        self._aspirate_flow_rates_lookup = (
```
sfoster1, line 125: i'm suspicious of these parentheses I think there might be a syntax problem here

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
```diff
@@ -302,6 +428,10 @@ def working_volume(self) -> float:
     def working_volume(self, tip_volume: float) -> None:
         """The working volume is the current tip max volume"""
         self._working_volume = min(self.config.max_volume, tip_volume)
+        self._active_tip_settings = self._config.supported_tips[
+            PipetteTipType(int(self._working_volume))
```
sfoster1: do we definitely have configurations for all the pipettes for all the working volumes they use, including the ot2 filter tips?

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
```diff
@@ -31,24 +34,31 @@ def get_virtual_pipette_static_config(
...
-            default_aspirate=config.default_aspirate_flow_rates,
-            default_dispense=config.default_dispense_flow_rates,
+            default_blow_out=tip_configuration.default_blowout_flowrate.values_by_api_level,
+            default_aspirate=tip_configuration.default_aspirate_flowrate.values_by_api_level,
+            default_dispense=tip_configuration.default_dispense_flowrate.values_by_api_level,
         ),
-        return_tip_scale=config.return_tip_height,
-        nominal_tip_overlap=config.tip_overlap,
+        return_tip_scale=tip_configuration.default_return_tip_height,
+        nominal_tip_overlap=config.tip_overlap_dictionary,
```
sfoster1, line 61: don't you have to modify `get_pipette_static_config` too?

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -228,7 +228,9 @@ def _set_current_state(self, to_state: State):
     @property
     def critical_point_override(self) -> Optional[CriticalPoint]:
         return (
-            CriticalPoint.FRONT_NOZZLE if self.hw_pipette.config.channels == 8 else None
+            CriticalPoint.FRONT_NOZZLE
+            if self.hw_pipette.config.channels.as_int == 8
```
sfoster1: change can go away if `PipetteChannelType` is `class PipetteChannelType(int, Enum)`

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -284,7 +286,7 @@ def _select_starting_pipette(self) -> Tuple[PipetteInfo, List[PipetteInfo]]:
             for mount, pip in pips.items():
                 pip_calibration = self._pipette_calibrations[mount]
                 info = PipetteInfo(
-                    channels=pip.config.channels,
+                    channels=pip.config.channels.as_int,
```
sfoster1: change can go away if `PipetteInfo` can be build from a `PipetteChannelType` (or in fact is automatically so built if it descends from `int`)

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -323,7 +325,7 @@ def _select_starting_pipette(self) -> Tuple[PipetteInfo, List[PipetteInfo]]:
         )
         if (
             left_pip.config.max_volume > right_pip.config.max_volume
-            or right_pip.config.channels > left_pip.config.channels
+            or right_pip.config.channels.as_int > left_pip.config.channels.as_int
```
sfoster1: change can go away with a custom ordering operator or indeed just for free if this descends from `int`

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -558,7 +560,7 @@ def get_active_pipette(self) -> CheckAttachedPipette:
         return CheckAttachedPipette(
             model=self.hw_pipette.model,
             name=self.hw_pipette.name,
-            tipLength=self.hw_pipette.config.tip_length,
+            tipLength=self.hw_pipette.active_tip_settings.default_tip_length,
```
sfoster1, line 561: why's this different from line 538 above?

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -840,8 +842,10 @@ def _get_tip_length(self) -> float:
                 self.active_tiprack._core.get_definition(),
             ).tipLength
         except cal_types.TipLengthCalNotFound:
-            tip_overlap = self.hw_pipette.config.tip_overlap.get(
-                self.active_tiprack.uri, self.hw_pipette.config.tip_overlap["default"]
+            tip_overlap = (
+                self.hw_pipette.tip_overlap.get(
```
sfoster1: wonder if we might benefit from a helper on `hw_pipette` called like `tip_overlap_for()` that does the defaulting for you

robot-server/robot_server/robot/calibration/deck/user_flow.py
```diff
@@ -356,7 +356,11 @@ def _get_tip_length(self) -> float:
                 self._tip_rack._core.get_definition(),
             ).tipLength
         except cal_types.TipLengthCalNotFound:
-            tip_overlap = self._hw_pipette.config.tip_overlap.get(self._tip_rack.uri, 0)
+            tip_overlap = (
+                self._hw_pipette.tip_overlap.get(
```
sfoster1: i guess not if we want to default to 0

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -533,7 +535,7 @@ def get_instruments(self) -> List[CheckAttachedPipette]:
             CheckAttachedPipette(
                 model=hw_pip.model,
                 name=hw_pip.name,
-                tipLength=hw_pip.config.tip_length,
+                tipLength=hw_pip.current_tip_length,
```
sfoster1: is this right? this changes based on the last call to `pick_up_tip`, right? it didn't used to

shared-data/pipette/definitions/2/general/eight_channel/p10/1_0.json
```diff
@@ -17,7 +17,7 @@
     "increment": 0.0,
```
sfoster1, line 17: Let's drop a lot of these json changes into another pr. certainly anything that doesn't change structure, maybe all of it. it'll be a lot easier to read.

shared-data/python/opentrons_shared_data/pipette/load_data.py
```diff
@@ -105,6 +106,69 @@ def load_serial_lookup_table() -> Dict[str, str]:
     return _lookup_table
 
 
+def load_liquid_model(
+    max_volume: PipetteModelType,
```
sfoster1: i don't love this match of argname and argtype. The point of this pr is to continue to tease apart the overloaded concepts that were smushed into pipette names and models into structured data; this conflates two of them again. let's call it `model` or something and get the max volume from the model def.

shared-data/python/opentrons_shared_data/pipette/load_data.py
```diff
@@ -105,6 +106,69 @@ def load_serial_lookup_table() -> Dict[str, str]:
...
+    channels: PipetteChannelType,
+    version: PipetteVersionType,
+) -> PipetteLiquidPropertiesDefinition:
+    liquid_dict = _liquid(channels, max_volume, version)
+    return PipetteLiquidPropertiesDefinition.parse_obj(liquid_dict)
+
+
+def _change_to_camel_case(c: str) -> str:
```
sfoster1, line 120: oh no

shared-data/python/opentrons_shared_data/pipette/load_data.py
```diff
@@ -105,6 +106,69 @@ def load_serial_lookup_table() -> Dict[str, str]:
...
+    # Tiny helper function to convert to camelCase.
```
sfoster1, line 121: i would almost infinitely prefer an explicit map of old to new keys

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -88,20 +93,65 @@ def as_tuple(
         return (self.major, self.minor)
 
 
+# TODO (lc 12-5-2022) Ideally we can deprecate this
```
sfoster1, line 12: honestly i think this sort of class is entirely reasonable

api/src/opentrons/hardware_control/api.py
```diff
@@ -555,14 +559,16 @@ async def _do_plunger_home(
                 await stack.enter_async_context(self._motion_lock)
             with self._backend.save_current():
                 self._backend.set_active_current(
-                    {checked_axis: instr.config.plunger_current}
+                    {checked_axis: instr.plunger_motor_current.run}
```
sfoster1, line 564: honestly wondering if, just for the sake of this pr, we mapped at least some of the new values to the old names. most of the time we wouldn't use them, but it would prevent us from needing a million changes that have no semantic meaning but just need different letters in sequence

sfoster1 (changes requested):
I think this is gonna be rough but important to do, but I think there's some things we can do to make it a bit less of a bitter pill to swallow.

Couple inline questions and some larger stuff:

### subtle but probably important distinctions
I don't quite get the difference between `pipette.active_tip_settings.default_return_height` and 
### small changes, lots and lots of small consequences
 There's a lot of changes that are sometimes almost the only changes in the file that are the consequence of a little type change that we can avoid by changing the new types or changing the upstreams. For instance, if you define `opentrons_shared_data.pipette.pipette_definitions.PipetteChannelType` as `PipetteChannelType(int, Enum)` then you get int comparisons for free:
```python
Python 3.7.17 (default, Jun 27 2023, 10:52:02)
[Clang 14.0.3 (clang-1403.0.22.14.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from enum import Enum
>>> class PipetteChannelType(int, Enum):
...  SINGLE_CHANNEL = 1
...  EIGHT_CHANNEL = 8
...  NINETY_SIX_CHANNEL = 96
...
>>> PipetteChannelType.SINGLE_CHANNEL
<PipetteChannelType.SINGLE_CHANNEL: 1>
>>> PipetteChannelType.SINGLE_CHANNEL ==1
True
>>> PipetteChannelType.SINGLE_CHANNEL == 8
False
>>> PipetteChannelType.EIGHT_CHANNEL == 8
True
>>> PipetteChannelType.EIGHT_CHANNEL == 1
False
```
Combine that, maybe a custom comparison operator (if it's even needed, may get that for free too) to sort them, and some handling for the `PipetteInfo` model in the old calibration stuff to build its `channels` element from the enum, and a lot of the changes in `robot_server.robot.calibration` fall away.

This seems like not so big a win, but one of the reasons prs like this are tough to deal with is that git(hub) has no way of emphasizing or separating parts of the diff, and so stuff that's complete boilerplate like adding `as_int` everywhere is presented on the same level as stuff that's more important. Let's cut down on some of those boilerplate changes.

api/src/opentrons/hardware_control/backends/simulator.py
Laura-Danielle: Just moved the #type: ignore from the internal function to have that be slightly better typed. It's suppressing a `TypeDict` type.

api/src/opentrons/hardware_control/instruments/instrument_abc.py
Laura-Danielle: okay. I'll leave them both in their respective `pipette.py` files.

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
Laura-Danielle: I think this is what the formatter did but I'll double check.

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
Laura-Danielle: No because I kept the old names in the pipette dict as keys.

robot-server/robot_server/robot/calibration/check/user_flow.py
Laura-Danielle: Do you mean `hw_pip.current_tip_length`? That's the tip length of the pipette currently attached. I only replaced things that were mapped to their respective pipette config. It's possible there are bugs in this code.

robot-server/robot_server/robot/calibration/check/user_flow.py
Laura-Danielle: Possibly, but not sure we should do that for this PR since we weren't doing that previously.

robot-server/robot_server/robot/calibration/check/user_flow.py
Laura-Danielle: Good catch.

shared-data/python/opentrons_shared_data/pipette/load_data.py
Laura-Danielle: 😆

shared-data/python/opentrons_shared_data/pipette/load_data.py
Laura-Danielle: *sigh* Okayyyy

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
Laura-Danielle: I suppose it could be, but ideally we move away from this being an external concept entirely. You truly should be able to request a certain pipette size/channels without having to worry about generation of the pipette.

api/src/opentrons/hardware_control/api.py
Laura-Danielle: in the config itself? Otherwise if we mapped it in the `Pipette` class there would still be a million changes. Plus mapping in a Pydantic model will get confusing since it's supposed to server as a schema.

shared-data/pipette/definitions/2/general/eight_channel/p10/1_0.json
Laura-Danielle: There will be test failures if we do this. Unless I rebase the this branch onto the json PR change. Plus some of those changes also modified the schema structure.

Laura-Danielle: and some of it changed the structure of the schema

robot-server/robot_server/robot/calibration/check/user_flow.py
Laura-Danielle: good catch

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
Laura-Danielle: Will double check

shared-data/python/opentrons_shared_data/pipette/load_data.py
Laura-Danielle: Not sure what you're asking for here. Do you just want me to call the `max_volume`, `model` instead? I guess it ultimately doesn't matter to me. I would also make the change in the `load_data` function in this file since it's using the same nomenclature.

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
sfoster1: Oh okay, I just see that a lot when there's a missing comma or somethign so the formatter groups statements wrong

shared-data/python/opentrons_shared_data/pipette/load_data.py
sfoster1: Yeah that makes sense to me

sfoster1 (approved):
Okay, this structure looks good to me. We obviously need to fix up tests and such, and run some protocols on a machine with this, but in particular we should pay close attention to anything going on with hardware testing - they poke into internals a lot and I'm a bit worried.

shared-data/pipette/schemas/2/pipetteLiquidPropertiesSchema.json
```diff
@@ -24,14 +24,13 @@
...
       "properties": {
-        "value": {
+        "default": {
           "$ref": "#/definitions/positiveNumber",
-          "$comment": "This key is deprecated in favor of valuesByApiLevel"
+          "$comment": "The default value of the flowrate"
         },
-        "min": { "$ref": "#/definitions/positiveNumber" },
-        "max": { "$ref": "#/definitions/positiveNumber" }
+        "valuesByApiLevel": { "type": "object", "required": [] }
```
SyntaxColoring, line 34: Can we flesh out this schema documentation for `default` and `valuesByApiLevel`? I'm confused about what's correct when, for example, we have something like this:

```
"default": 6,
"valuesByApiLevel": {
    "2.14": 7
}
```

If we're using `apiLevel` *2.15,* is the correct flow rate 6, or 7?

shared-data/pipette/definitions/2/liquid/single_channel/p50/1_4.json
```diff
@@ -2,12 +2,19 @@
   "$otSharedSchema": "#/pipette/schemas/2/pipetteLiquidPropertiesSchema.json",
   "supportedTips": {
     "t200": {
-      "defaultAspirateFlowRate": 25.0,
-      "defaultDispenseFlowRate": 50.0,
-      "defaultBlowOutFlowRate": 1000.0,
+      "defaultAspirateFlowRate": {
+        "default": 25,
+        "valuesByApiLevel": { "2.0": 25 }
```
SyntaxColoring, line 7: Question:

It looks like for all (?) the pipette definitions so far,`valuesByApiLevel` always has one key that's equal to the `default`. Is it functionally necessary to specify both, or is it just a hint for future extension?

shared-data/pipette/schemas/2/pipetteLiquidPropertiesSchema.json
```diff
@@ -97,6 +94,16 @@
...
+    "defaultTipOverlapDictionary": {
+      "type": "object",
+      "description": "Map of tip overlap values with defaults",
+      "required": ["default"],
+      "$comment": "Other keys in here should be labware URIs",
+      "properties": {
+        "default": { "type": "number" }
+      },
+      "additionalProperties": { "type": "number" }
+    },
```
SyntaxColoring, line 107: Hmm, what makes tip overlap different from flow rate, tip length, and tip height? Tip overlap is specified here as a dict that's keyed on labware URIs, but those other things are specified as a dict that's keyed on a "tip type" like "t300."

shared-data/python/opentrons_shared_data/pipette/load_data.py
```diff
@@ -105,6 +106,69 @@ def load_serial_lookup_table() -> Dict[str, str]:
...
+    config_name = c.split("_")
+    if len(config_name) == 1:
+        return config_name[0]
+    return f"{config_name[0]}" + "".join(s.capitalize() for s in config_name[1::])
+
+
+def update_pipette_configuration(
```
SyntaxColoring, line 128: You might just not have gotten to this yet, but this function seems like a high-value target for unit test coverage, since it has a type-unsafe `Dict[str, Any]` parameter and it's doing legacy name mapping and stuff.

shared-data/python/opentrons_shared_data/pipette/load_data.py
```diff
@@ -105,6 +106,69 @@ def load_serial_lookup_table() -> Dict[str, str]:
...
+                        dict_of_base_model[top_name][k][nested_name] = v
+                else:
+                    # isinstances are needed for type checking.
+                    dict_of_base_model[top_name][nested_name] = v
+    dict_of_base_model["quirks"] = list(
+        set(dict_of_base_model["quirks"]) - set(quirks_list)
+    )
+
+    # re-serialization is not great for this nested enum so we need
+    # to perform this workaround.
```
SyntaxColoring, line 168: Out of curiosity, what was the problem here?

shared-data/python/opentrons_shared_data/pipette/model_constants.py
```diff
@@ -3,9 +3,19 @@
 from .types import Quirks, RobotMountConfigs, AvailableUnits
 
 MOUNT_CONFIG_LOOKUP_TABLE = {
-    "GEN1": RobotMountConfigs(768, 220, 30),
-    "GEN2": RobotMountConfigs(3200, 155.75, 60),
-    "FLEX": RobotMountConfigs(2133.33, 230.15, 80),
+    "GEN1": {
+        "single": RobotMountConfigs(768, 220, 30),
```
SyntaxColoring: Can we make these keys `PipetteChannelType`s instead of plain `"single"`/`"multi"` strings?

shared-data/python/opentrons_shared_data/pipette/model_constants.py
```diff
@@ -3,9 +3,19 @@
 from .types import Quirks, RobotMountConfigs, AvailableUnits
 
 MOUNT_CONFIG_LOOKUP_TABLE = {
-    "GEN1": RobotMountConfigs(768, 220, 30),
-    "GEN2": RobotMountConfigs(3200, 155.75, 60),
-    "FLEX": RobotMountConfigs(2133.33, 230.15, 80),
+    "GEN1": {
```
SyntaxColoring: Can we make these keys `PipetteGenerationType`s instead of plain `"GEN1"`/`"GEN2"`/`"FLEX"` strings?

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -30,15 +30,11 @@ class PipetteTipType(Enum):
     t1000 = 1000
 
 
-class PipetteChannelType(Enum):
+class PipetteChannelType(int, Enum):
```
SyntaxColoring: As a general practice, I don't think we should subclass `Enum`s from other types unless we really need to. And I don't think we need to, here?

https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3479896162/Don+t+subclass+Python+enums+from+str+proposed

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -88,20 +89,65 @@ def as_tuple(
...
+        if self.pipette_generation == PipetteGenerationType.FLEX:
+            return PipetteVersionType(3, 0)
+        elif self.pipette_generation == PipetteGenerationType.GEN2:
+            return PipetteVersionType(2, 0)
+        else:
+            return PipetteVersionType(1, 0)
+
+
+@dataclass
+class PipetteModelVersionType:
```
SyntaxColoring, line 39: Question:

What does the `Type` suffix signify? What kind of Python type *isn't* a `Type`? I've seen this in other places in our codebase, too.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
SyntaxColoring: > I suppose it could be, but ideally we move away from this being an external concept entirely. You truly should be able to request a certain pipette size/channels without having to worry about generation of the pipette.

Hmm. Even in a world where protocol authors request pipettes like that, it seems like this concept would still exist in HTTP. Something would still need to tell the app what pipettes the user needs to attach in order to run a protocol, and the app would still need to determine whether the pipettes that are already attached are a match.

Or is your vision that those would get reported as a fine-grained list of models? Like: "You can attach a `p1000_single_1_0`, or `p1000_single_1_3`, or `p1000_single_1_4`, or ..." etc.

SyntaxColoring (commented):
Makin' my way downtown. I still have a way left to go, but here are some questions so far, and a couple requests. 🕺

shared-data/pipette/schemas/2/pipetteLiquidPropertiesSchema.json
Laura-Danielle: The flowrate would be 6 and yes I can add more details.

shared-data/pipette/definitions/2/liquid/single_channel/p50/1_4.json
Laura-Danielle: hint for future extension. There are some models that do indeed have multiple values.

shared-data/pipette/schemas/2/pipetteLiquidPropertiesSchema.json
Laura-Danielle: It was was previously used. I made it match the V1 schema for sake of not touching a bunch more code. Protocol Engine and other users of this data still pass in LabwareUri to get this data.

The tip length imo should also eventually be removed from the schema all together because it doesn't make sense to be there. It is not a property of a pipette but rather a property of a tip rack.

You can think of the top level `t...` keys as being notation for representing _all_ tips under the 300ul volume which may infact have different properties. That is why right now we're keying the aspirate/dispense by labware uri as well. Unfortunately don't have a fully fleshed out future yet for Liquid classes but you can imagine this type of stuff expanding when we get there.

shared-data/python/opentrons_shared_data/pipette/load_data.py
Laura-Danielle: when you convert the model (with the nested values) to a dict, it's still using the enum as keys but the pydantic model is expecting strings to be converted to keys.

Pydantic doesn't really want you patching data in this way unfortunately which is why things can get ugly.

Laura-Danielle: And we shouldn't be updating static configs dynamically as we are! But that is the legacy code we struggle with today. My ideal case is completely destroying the pipette dict to replace it with a stateful pipette dict/class/whatever and a static pipette dict/class/whatever.

shared-data/python/opentrons_shared_data/pipette/model_constants.py
Laura-Danielle: Sure

shared-data/python/opentrons_shared_data/pipette/model_constants.py
Laura-Danielle: Sure

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
Laura-Danielle: We need to for direct int comparison (which happens a lot with this class). See Seth's initial comment in this PR.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
Laura-Danielle: In this instance, I did it to try and distinguish as much as possible from the `PipetteModel`/`PipetteName` dev_types we already have in our code base.

Normally in the past, it's been used when we have type defines at the top of the file. IMO I think it just sounds nicer. For example in my most recent PR, I had a type called something to the effect of `OverrideType`.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
Laura-Danielle: In my ideal world, the app shouldn't have to determine if pipettes match at all or not. The robot server/protocol engine should just say 'hey this doesn't match what's currently on the robot'.

For simulation, we should have something that sends back a list of compatible _generations_ (like we kinda talked about in the partial tip meeting). We haven't -- and probably won't in the future -- cared about the functionality between models.

shared-data/python/opentrons_shared_data/pipette/load_data.py
SyntaxColoring: Makes sense. Thanks for the explanation!

shared-data/python/opentrons_shared_data/pipette/load_data.py
Laura-Danielle: yep, thanks for the reminder though :D

jbleon95:
Successfully smoke tested, for Flex and OT-2, attaching single and multi-channel pipettes, loading pipettes in protocols (with both GEN-3 and user facing names for Flex), analyzing on both app and robot and running protocols on robot. Additionally confirmed that mutable configs were functioning on OT-2.

jbleon95 (approved):

PR #13105, opened 2023-07-14
chore(hardware): more opts for motion_analyzer

sfoster1:
Fix acceleration parsing, add an option to print out move commands as well as move results


ahiuchingau (approved):
Makes sense to me

vegano1 (approved):

PR #13106, opened 2023-07-14
refactor(components): add higher and lower-level deck abstraction for Flex configuration

b-cooper:
# Overview

Breakdown the RobotWorkSpace component to be more granular and make a new higher abstraction above it called `BaseDeck` that handles the positioning and dynamic styling of all deck layers, modules, and labware.

Create a RobotCoordinateSpace component that manages only scaling/reflection for the coordinate
system anchored in the bottom left. Create DeckSlotLocation component to rener a specific deck slot
location and bake path information into the component itself instead of deserialzing svgson. Create
a BaseDeck asbstraction that handles the basic rendering of a deck with labware, modules, and deck
labels

# Review requests

- Check storybook story for BaseDeck 

# Risk assessment
low, this component is not being used yet


shlokamin (approved):
woooo!

PR #13107, opened 2023-07-15
docs(api): Clean up custom/extra labware in opentrons.simulate and opentrons.execute

SyntaxColoring:
# Overview

The user-facing functions in `opentrons.simulate` and `opentrons.execute` are sometimes confusing in their handling of custom (or "extra") labware. The docs don't always match the behavior, and the behavior is sometimes subtly different between functions. This cleans some of that up.

This goes towards RSS-268, reducing the risk of #12970.

# Test Plan

Just make sure CI keeps passing in all environments.

# Changelog

* Docs correction for `opentrons.simulate.get_protocol_api()` and `opentrons.execute.get_protocol_api()`: The Jupyter directory will be searched when `extra_labware` is *`None`*, *not* when it's an empty dict.

  This has been the behavior since [PR #4574](https://github.com/Opentrons/opentrons/pull/4574/files). https://github.com/Opentrons/opentrons/blob/f9509e70f776f3313b13814f1b65d3af822f6915/api/src/opentrons/execute.py#L108

* Docs addition for `opentrons.simulate.simulate()` and `opentrons.execute.execute()`: The Jupyter directory will be searched when `custom_labware_paths` is `None`.

  This part is a bit nuanced. These functions *have* been searching Jupyter for labware since [PR #4574](https://github.com/Opentrons/opentrons/pull/4574/files), so we should certainly document that.

  However, the actual current behavior is to check for `None` *or `[]`*. The exact mechanism of this is a little subtle. https://github.com/Opentrons/opentrons/blob/f9509e70f776f3313b13814f1b65d3af822f6915/api/src/opentrons/simulate.py#L281-L284
https://github.com/Opentrons/opentrons/blob/f9509e70f776f3313b13814f1b65d3af822f6915/api/src/opentrons/simulate.py#L307-L309

  I propose we *don't* document the `[]` behavior, and that we remove that behavior in a follow-up PR. (This PR just adds the docs and an `xfail` test.)

  This would be a ⚠️ breaking change ⚠️ for any users who are passing `custom_labware_paths=[]` and expecting it to search Jupyter. I think that's okay. This behavior has been undocumented so far; I doubt anybody is relying on it, and if they are, it's easy to identify and fix; and resolving this inconsistency would be really make things less confusing.

* Test refactors:

  Split some of `test_execute.py` and `test_simulate.py` from a few big functions to many small functions. This gives us more granular test results and lets us `xfail` things more granularly, which is helpful while working on RSS-268. The downside is more boilerplate.

# Review requests

* Do we agree with the ⚠️ breaking change ⚠️ described above?
* See my GitHub comments below.

# Risk assessment

Low.


api/tests/opentrons/test_execute.py
```diff
@@ -20,8 +25,24 @@
 HERE = Path(__file__).parent
 
 
+@pytest.fixture(
+    params=[
+        APIVersion(2, 0),
+        # TODO(mm, 2023-07-14): Enable this for https://opentrons.atlassian.net/browse/RSS-268.
+        # ENGINE_CORE_API_VERSION,
```
SyntaxColoring, line 32: What I'd really like to do is run all the tests with `APIVersion(2, 0)` and `APIVersion(2, 14)`, but automatically mark all the `APIVersion(2, 14)` ones as `xfail(strict=True)`. I couldn't figure out how to do that.

api/tests/opentrons/test_execute.py
```diff
@@ -20,8 +25,24 @@
...
+
+    Newer API versions execute through Protocol Engine, and older API versions don't.
+    The two codepaths are very different, so we need to test them both.
+    """
+    return request.param  # type: ignore[attr-defined,no-any-return]
+
+
 @pytest.fixture
-def mock_get_attached_instr(
+def mock_get_attached_instr(  # noqa: D103
```
SyntaxColoring, line 45: I wasn't sure how to document this because I'm not totally sure what it's doing. I think I understand the `get_attached_instruments` patch (we need the hardware API to return the pipettes that the test protocols try to load), but I don't know what the `delay` patch is doing here.

api/tests/opentrons/data/test_simulate.py
SyntaxColoring, line 1: The tests for `opentrons.execute` were using a Testosaur file, and but the tests for opentrons.simulate were using this one. I don't see any reason for them to be different, so I'm deleting this one in favor of the Testosaur.

api/tests/opentrons/test_simulate.py
SyntaxColoring, line 1: Duplication between `test_simulate.py` and `test_execute.py` is just as bad as between `simulate.py` and `execute.py`, and I feel bad about making it worse here. But I think adding these tests improves things more than it harms things, given the work that we're doing on the production code.

api/src/opentrons/execute.py
```diff
@@ -107,14 +103,11 @@ def get_protocol_api(
...
-    :param extra_labware: If specified, a mapping from labware names to
-                          labware definitions for labware to consider in the
-                          protocol in addition to those stored on the robot.
-                          If this is an empty dict, and this function is called
-                          on a robot, it will look in the 'labware'
-                          subdirectory of the Jupyter data directory for
-                          custom labware.
+                         :py:obj:`opentrons.protocol_api.ProtocolContext.bundled_data`.
+    :param extra_labware: A mapping from labware load names to custom labware definitions.
+                          If this is `None` (the default), and this function is called on a robot,
```
ecormany: musing about rst formatting…my inclination would be to put both ``None`` (Python) and ``labware`` (directory name) in double backticks to set them in monospace.

api/src/opentrons/execute.py
```diff
@@ -279,11 +272,11 @@ def execute(
...
                         single argument (the name doesn't matter) which will
                         be a dictionary (see below). Default: ``None``
-    :param custom_labware_paths: A list of directories to search for custom
-                                 labware, or None. Ignored if the apiv2 feature
-                                 flag is not set. Loads valid labware from
-                                 these paths and makes them available to the
-                                 protocol context.
+    :param custom_labware_paths: A list of directories to search for custom labware.
+                                 Loads valid labware from these paths and makes them available
+                                 to the protocol context. If this is `None` (the default), and this
```
ecormany: worth linking ProtocolContext? or overkill?

api/src/opentrons/simulate.py
ecormany, line 1: match whatever docstring decisions you make for `execute.py` here.

ecormany (commented):
A few tiny docstring comments/musings. Jupyter behavior being accurately documented anywhere is awesome, thanks for including those changes in this PR. 🙌

api/src/opentrons/execute.py
SyntaxColoring: If anything, I think we should link to the conceptual/overview docs on custom labware, (https://docs.opentrons.com/v2/new_labware.html#custom-labware), right?

TamarZanzouri (approved):
LGTM!

api/tests/opentrons/test_execute.py
sanni-t: Yea, it might have been a leftover patch from some old code where delays were called during protocol execution or pipetting commands. Could test by removing the patch.

api/tests/opentrons/test_execute.py
```diff
@@ -203,68 +238,191 @@ def emit_runlog(entry: Any) -> None:
...
-    monkeypatch: pytest.MonkeyPatch,
-    virtual_smoothie_env: None,
-    mock_get_attached_instr: mock.AsyncMock,
-) -> None:
-    fixturedir = (
-        HERE / ".." / ".." / ".." / "shared-data" / "labware" / "fixtures" / "2"
-    )
-    entries = []
+class TestExecutePythonLabware:
+    """Tests for making sure execute() handles custom labware correctly for Python files."""
```
sanni-t, line 242: Interesting tests structure!

SyntaxColoring: Yeah, I've been trying this out for modules that have multiple independent free functions in them. It's nice because it keeps the fixtures closely scoped to the tests that actually need them. It also lets you get away with more concise names.

PR #13109, opened 2023-07-17
feat(protocol-designer): allow tip racks but not trash selection in moveLabware form

jerader:
closes RAUT-561 and RQA-1076 and RQA-1114

# Overview

When you move a labware, tip racks on the deck are now selectable but the trash is no longer selectable (for now since only a fixed trash is supported)

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_labware-dropdown-options/

add a move labware step in the timeline, doesn't matter which robot type it is. The tipracks on the deck should be selectable but the trash should not be selectable in the drop down.

add a transfer or mix to the timeline. The labware selection should not show tip racks but should show the trash.

# Changelog

- modify `getLabwareOptions` selector and test to have branching logic for tipracks and trash when the presaved step is `moveLabware`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/ui/labware/__tests__/selectors.test.ts
+++ b/protocol-designer/src/ui/labware/__tests__/selectors.test.ts
@@ -151,6 +151,36 @@ describe('labware selectors', () => {
       ])
     })
 
+    it('should return labware options for move labware with tips and no trash', () => {
+      const labwareEntities = {
+        ...tipracks,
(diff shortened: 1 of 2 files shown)


jerader:
i ended up making this change in this pr: https://github.com/Opentrons/opentrons/pull/13237 so gonna close this

PR #13110, opened 2023-07-17
feat(api): add limit switch backoff move to 96-channel pipette tip motor home sequence

ahiuchingau:
# Overview
We want to add limit switch backoff to the tip motors during the home sequence so they behave the same way as the other stepper motors in the system.


sfoster1 (approved):
structure looks good!

fsinapi (approved):
👍

sfoster1 (approved):
Works when homing a 96 and the rest of the system on a machine!

PR #13111, opened 2023-07-17
refactor(protocol-designer): properly disable correct StepList buttons

jerader:
closes RQA-1081 and RQA-1075 and RQA-1074

# Overview

🔈  Gonna hold from merging this until post webinar on 7/19/23! 

Some bugs that QA captured. This fixes the zIndex of the `Add step` button so it doesn't show up when the hint modal shows up. And it hides the disabled buttons instead of wiring them up in the step creation button component to address several of the bug tickets mentioned

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_step-creation-button

- go to Design tab and add a new step, the module buttons that don't have the correct module in the protocol should not show up. 

# Changelog

- hide disabled buttons and fix test
- change z index

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepCreationButton.tsx
+++ b/protocol-designer/src/components/StepCreationButton.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react'
 import { useDispatch, useSelector } from 'react-redux'
-import cx from 'classnames'
 import {
   Tooltip,
   DeprecatedPrimaryButton,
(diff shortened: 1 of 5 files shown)


koji (approved):
when the splash is displayed, buttons aren't clickable
when added a mag block, the button for mag block is disabled and tooltip works

PR #13112, opened 2023-07-17
refactor(app): retract all axes in LPC instead of move to fixed trash

b-cooper:
# Overview

Instead of using fixed trash's A1 well as a proxy for "a safe location for the gantry to be out of the way", use the new purpose built retract axis command to move the gantry out of the user's way during LPC.

Closes RSS-254

# Review requests

- LPC should function as expected on OT-2 and Flex. When moving gantry out of the way, it will retract axes in order of z, x, y

# Risk assessment
med, this is a small change, but it's replacing old stable functionality with new more stable behavior.

### DIFF
--- a/app/src/organisms/LabwarePositionCheck/CheckItem.tsx
+++ b/app/src/organisms/LabwarePositionCheck/CheckItem.tsx
@@ -8,7 +8,6 @@ import { PrepareSpace } from './PrepareSpace'
 import { JogToWell } from './JogToWell'
 import {
   CreateCommand,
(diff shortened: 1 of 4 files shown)


sanni-t (approved):
LGTM!

There's one thing we should test though- we should confirm that when doing LPC on a labware in a thermocycler on the Flex, the gripper arm doesn't hit the thermocycler's lid (more specifically the latch pins of the lid) when retracting the X axis.

shlokamin (approved):
tested on RSS bot with a TC, seems to work just fine!

PR #13115, opened 2023-07-18
fix(hardware): Remove OT3GPIO from can-control script

andySigler:
# Overview

The OT3GPIO instance in this script was blocking it from being ran at the same time as the robot-server, which would be helpful while debugging behaviors seen during App usage. Test this and was able to run can_control while the server was still running.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):

sfoster1 (approved):
This should be fine now, we only ever needed this for EVT main boards

PR #13117, opened 2023-07-18
feat(shared-data): add the PVT models for all pipettes

Laura-Danielle:
## Overview

All pipettes will use version v3.5. For now, we can keep the same configurations as v3.4.


sfoster1 (approved):
looks good to me once we fix lint!

PR #13119, opened 2023-07-18
refactor(hardware-testing): Use new force-gauge fixture for PVT Grippers

andySigler:
# Overview

This PR modifies the Gripper QC script to use the new force-gauge fixture hardware. This PR also updates the measurement method so that all forces are recorded from the start of a grip, so that our averages can include both initial and later forces that occur regardless of variance.

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
lgtm

PR #13016, opened 2023-06-29
feat(api): load adapter and load/move labware onto adapters in PAPI

jbleon95:
# Overview

Closes RLAB-343.

This PR adds PAPI support for the labware adapter split work. This is implemented in three main ways in the public API:
- The ability to load adapters (which are any labware that has `"adapter"` in `allowedRoles` in its Labware Definition) via `ProtocolContext` and `ModuleContext` `load_adapter` and `load_adapter_from_definition`.
- The ability to load labware on top of compatible labware objects (this includes adapters) via a `Labware.load_labware` and `Labware.load_labware_from_definition`. Labware that exists on top of a `Labware` object can be accessed via `Labware.child` as a corollary to `Labware.parent`. Loading an adapter underneath a compatabile labware can also be done via the added `adapter=` keyword arg for protocol context and module `load_labware`.
- The ability to move labware on top of another labware via `move_labware`, both manually and with the gripper. This is done by giving the `Labware` object as the `new_location`

A usable protocol demonstrating the new api:
```
metadata = {
    'protocolName': 'PAPI Labware Split Demonstration',
}

requirements = {
    "robotType": "OT-3",
    "apiLevel": "2.15"
}


def run(context):
    # Loading pipette and tiprack and module
    tiprack = context.load_labware('opentrons_flex_96_tiprack_50ul', location="D3")
    pipette = context.load_instrument("p50_single_gen3", mount="right", tip_racks=[tiprack])
    heater_shaker = context.load_module("heaterShakerModuleV1", location="D1")
    heater_shaker.open_labware_latch()

    # Adapters can be loaded directly onto the deck or a module
    flat_adapter = context.load_adapter("opentrons_universal_flat_adapter", location="C3")
    pcr_adapter = heater_shaker.load_adapter("opentrons_96_pcr_adapter")

    # Labware can then be loaded onto the adapter 
    nest_plate = pcr_adapter.load_labware("nest_96_wellplate_100ul_pcr_full_skirt")

    # Alternatively, the adapter and labware can be loaded in one line, this is also possible on a module.
    # Note that via this method, if you specify a namespace it will use that for both the labware and adapter,
    # but this is not true with version, to allow the app to pick the most recent adapter definition if necessary
    armadillo_plate = context.load_labware(
        load_name="armadillo_96_wellplate_200ul_pcr_full_skirt",
        location="C1",
        adapter="opentrons_96_pcr_adapter"
    )
    # The adapter labware object can then be accessed via .parent
[...]

PR #13120, opened 2023-07-18
feat(engine): support loading offsets for labware with OnLabwareLocation

jbleon95:
# Overview

Closes RLAB-359.

This PR is a follow-up to #12928 and #13016, and adds logic to Protocol Engine's `find_applicable_labware_offset_id` to support labware that has the `OnLabwareLocation` i.e. is loaded on top of a labware/adapter. It uses the `LabwareOffsetLocation` `definitionUri` field and, if the adapter is loaded onto a module, the `LabwareOffsetLocation` `moduleModel` field to determine the correct offset ID when a labware is loaded.

# Test Plan

This was tested using Postman to ensure that an added labware offset with the `definitionUri` field, with and without the `moduleModel` field, had its offset ID correctly found when loading the labware.

# Changelog

- Updated `EquipmentHandler.find_applicable_labware_offset_id` to correctly handle finding the offset for labware with an `OnLabwareLocation`
- Updated doc string for `LabwareOffsetLocation` definitionUri to reflect that it can be used with `moduleModel` 

# Review requests

Does the logic for determining offsets make sense and can any issues be foreseen that would make this untenable when more than one labware can be stacked on top of each other.

# Risk assessment

Low, this only affects labware with the `OnLabwareLocation` and adds behavior that wasn't correctly supported before.

### DIFF
--- a/api/src/opentrons/protocol_engine/execution/equipment.py
+++ b/api/src/opentrons/protocol_engine/execution/equipment.py
@@ -388,16 +388,26 @@ class EquipmentHandler:
             The ID of the labware offset that will apply,
             or None if no labware offset will apply.
         """
-        if isinstance(labware_location, DeckSlotLocation):
-            offset = self._state_store.labware.find_applicable_labware_offset(
-                definition_uri=labware_definition_uri,
-                location=LabwareOffsetLocation(
-                    slotName=labware_location.slotName,
-                    moduleModel=None,
-                ),
-            )
-            return self._get_id_from_offset(offset)
+        labware_offset_location = (
+            self._get_labware_offset_location_from_labware_location(labware_location)
+        )
+        if labware_offset_location is None:
+            # No offset for off-deck location.
+            # Returning None instead of raising an exception allows loading a labware
+            # with 'offDeck' as valid location.
+            # Also allows using `moveLabware` with 'offDeck' location.
+            return None
+        offset = self._state_store.labware.find_applicable_labware_offset(
+            definition_uri=labware_definition_uri,
+            location=labware_offset_location,
+        )
+        return self._get_id_from_offset(offset)
 
+    def _get_labware_offset_location_from_labware_location(
+        self, labware_location: LabwareLocation
+    ) -> Optional[LabwareOffsetLocation]:
+        if isinstance(labware_location, DeckSlotLocation):
+            return LabwareOffsetLocation(slotName=labware_location.slotName)
         elif isinstance(labware_location, ModuleLocation):
             module_id = labware_location.moduleId
             # Allow ModuleNotLoadedError to propagate.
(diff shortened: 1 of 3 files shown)


api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -439,6 +440,46 @@ def find_applicable_labware_offset_id(
...
+            parent_labware_id = labware_location.labwareId
+            parent_labware_uri = self._state_store.labware.get_definition_uri(
+                parent_labware_id
+            )
+
+            base_location = self._state_store.labware.get_parent_location(
+                parent_labware_id
+            )
+
+            if isinstance(base_location, ModuleLocation):
```
SyntaxColoring: Hmm, it feels like there should be a recursive helper somewhere so this method doesn't have to duplicate itself so much.

For example, this would be easier if we had something that reduced a `LabwareLocation` linked-list into a `(slot, moduleModel, adapterURI)` tuple. And then we could just pass that tuple to `find_applicable_labware_offset()`.

There's a sort of fundamental problem with that, which is that it's a lossy conversion if there are multiple adapters in a stack. But as we discussed in-person, that's always going to be a problem as long as our HTTP input is flat instead of hierarchical. So it wouldn't matter if the implementation kept the top of the stack or the bottom of the stack.

SyntaxColoring (approved):
Thanks, this looks like it'll do the trick. One suggestion for reducing the code complexity, if you're up for it, but no pressure if you want to move on to something else.

sanni-t (approved):
LGTM!

> can any issues be foreseen that would make this untenable when more than one labware can be stacked on top of each other.

It doesn't look like there would be any issues code-wise. I think you have covered all labware placement cases pretty well. We do need to hash out how to handle offsets and LPC for stacked labware; I think it's a slightly bigger discussion than is needed for this PR, and will probably need some testing. Until then, we can leave a comment somewhere appropriate that currently all stacked labware will share a single offset (except the bottom-most labware directly on the deck).

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -430,20 +441,28 @@ def find_applicable_labware_offset_id(
...
+            base_labware_offset_location = (
+                self._get_labware_offset_location_from_labware_location(base_location)
+            )
+            if base_labware_offset_location is None:
+                # No offset for labware sitting on labware off-deck
+                return None
+            return LabwareOffsetLocation(
+                slotName=base_labware_offset_location.slotName,
+                moduleModel=base_labware_offset_location.moduleModel,
+                definitionUri=parent_labware_uri,
```
SyntaxColoring, line 467: Agreed with Sanniti's comment about calling out labware stacking. Here would be one place we could call that out, since we're throwing away `base_labware_offset_location.definitionUri`.

SyntaxColoring (approved):
Changes since my last review look good, thanks!

PR #13121, opened 2023-07-18
fix(api): dont emit 0-duration moves

sfoster1:
They really stress the microcontrollers and they do absolutely nothing, so let's not put them on the bus.

This should fix the last known-issue gantry stall. 

Tested by running normalize (a lot) on a machine and running some calibrations.


fsinapi (approved):
love it

Laura-Danielle (approved):
nioce

vegano1 (commented):
nice!

vegano1 (approved):

sfoster1:
I'm having a devil of a time getting all the way through the 3.5 hour protocol because I keep forgetting tips but here's a plot of the axis and you can see it avoid the drift.
Screenshot 2023-07-18 at 1 02 21 PM

PR #13124, opened 2023-07-18
fix(api): tip action test should check for both home and backoff step

ahiuchingau:
# Overview
Fix an api test

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
lgtm!

PR #13125, opened 2023-07-18
fix(app): redirect to run summary screen based on run endpoint

shlokamin:
# Overview
This PR changes the app's current run route logic to reroute based on info coming from the `/runs/runId` endpoint rather than the `/runs` endpoint. This is because the run summary screen reads data from the `/runs/runId` endpoint and we want run state information to be coming from the same place.

This fixes a bug where a run would succeed/fail but we wouldn't render the run succeeded/failed spash screen inside the run summary screen. The reason this bug was happening was because the app would redirect the run summary screen when the `/runs` endpoint reflected that the current run is in a completed/failed/stop requested state. BUT the run summary screen only renders the splash when the `/runs/runId` endpoint reflects that the run is in a completed/failed state. Because these are two different endpoints with two different react queries with two different refetch intervals, the data got out of sync.

closes RQA-1098

# Risk assessment

Low

### DIFF
--- a/app/src/App/hooks.ts
+++ b/app/src/App/hooks.ts
@@ -8,6 +8,7 @@ import {
   useAllProtocolIdsQuery,
   useAllRunsQuery,
   useHost,
+  useRunQuery,
 } from '@opentrons/react-api-client'
 import {
   getProtocol,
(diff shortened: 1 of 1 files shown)


koji (approved):
The changes look good to me.

PR #13126, opened 2023-07-18
feat(shared-data): alphabetizes orderedSlots in Flex deck def

jerader:
closes RQA-1077

# Overview

In PD, it is hard to tell which slot you want to add a module to since the Flex deck slots were ordered the way the Ot-2 slots are ordered. But since the flex slots are prefaced with a letter, that way of ordering is confusing. So this PR alphabetizes the ordered Slots.

# Test Plan

Do some smoke testing in app/odd and PD to make sure everything works as expected

# Changelog

- change the flex deck definition `orderedSlots` to be in alphabetical order 

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/shared-data/deck/definitions/3/ot3_standard.json
+++ b/shared-data/deck/definitions/3/ot3_standard.json
@@ -13,15 +13,15 @@
   "locations": {
     "orderedSlots": [
       {
-        "id": "D1",
-        "position": [0.0, 0.0, 0.0],
+        "id": "A1",
+        "position": [0.0, 321.0, 0.0],
         "matingSurfaceUnitVector": [-1, 1, -1],
         "boundingBox": {
           "xDimension": 128.0,
           "yDimension": 86.0,
           "zDimension": 0
         },
-        "displayName": "Slot D1",
+        "displayName": "Slot A1",
         "compatibleModuleTypes": [
           "magneticModuleType",
           "temperatureModuleType",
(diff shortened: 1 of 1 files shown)


SyntaxColoring (changes requested):
I don't think we can make this change without some additional cleanup work, and even then it's risky. Some things scattered about are assuming that you can map slots between an OT-2 and Flex by their index in this list. Let me see if I can dig up some specifics.

SyntaxColoring:
Okay, here's what I was thinking of:

https://github.com/Opentrons/opentrons/blob/0c52b1e96d275855bfe761384e5731765c5e791b/shared-data/python/opentrons_shared_data/deck/__init__.py#L69-L73

https://github.com/Opentrons/opentrons/blob/0c52b1e96d275855bfe761384e5731765c5e791b/hardware-testing/hardware_testing/opentrons_api/helpers_ot3.py#L859-L863

There might be more, not sure.

I think these are fixable by changing the things that call them to [not use integers anymore](https://opentrons.atlassian.net/browse/RLAB-345), or, at least, change the integer-based array lookup into an [explicit map](https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/types.py#L341-L354). The former requires some case-by-case thinking, and the latter might run into [frustrating database hazards](https://opentrons.atlassian.net/browse/RSS-98). Definitely possible, but for your purposes of fixing the Protocol Designer sorting, you'll probably have a better time if you do the sort locally in Protocol Designer?

jerader:
> I don't think we can make this change without some additional cleanup work, and even then it's risky. Some things scattered about are assuming that you can map slots between an OT-2 and Flex by their index in this list. Let me see if I can dig up some specifics.

thanks @SyntaxColoring , lmk if you dig up any specifics. On the frontend, we map based off of the `id` and i haven't seen anything weird in my smoke testing in app/pd/odd

jerader:
> Definitely possible, but for your purposes of fixing the Protocol Designer sorting, you'll probably have a better time if you do the sort locally in Protocol Designer?

Oops posted my last comment before i saw this @SyntaxColoring, thanks for digging this up! Yeah, sorting locally might be better. Will look into it!

jerader:
As @SyntaxColoring pointed out, there are a few instances that rely on the slot indexes to map the same way as the OT-2. For now, the best approach is to alphabetize locally in PD

PR #13130, opened 2023-07-19
edits to error type copy

ecormany:
# Overview

just a shell PR for now so we can diff / discuss the contents of this JSON file line-by-line, following on from RUXD-810.

# Test Plan

# Changelog

- a few value changes
- _no test changes yet_

# Review requests

# Risk assessment

low, if we even merge this

### DIFF
--- a/shared-data/errors/definitions/1/errors.json
+++ b/shared-data/errors/definitions/1/errors.json
@@ -71,7 +71,7 @@
       "category": "roboticsControlError"
     },
     "2004": {
-      "detail": "Motion Planning Failure",
+      "detail": "Motion Planning Failed",
       "category": "roboticsControlError"
     },
     "2005": {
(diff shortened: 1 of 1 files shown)


ecormany:
other lines i might consider editing…

https://github.com/Opentrons/opentrons/blob/e265a057fc3c7ba68f3ef360175446d184ad0856/shared-data/errors/definitions/1/errors.json#L22
`General Communication Error` or similar, since there are other more specific categories below?

--- 

https://github.com/Opentrons/opentrons/blob/e265a057fc3c7ba68f3ef360175446d184ad0856/shared-data/errors/definitions/1/errors.json#L54 
less redundant way of saying this?

ecormany:
@sfoster1 Any thoughts on what we should do with this error copy? Keep this draft open or close it for now and revisit the topic later?

PR #13131, opened 2023-07-19
fix(app): make update flow copy robot agnostic

b-cooper:
# Overview

Instead of instructing users to update their OT-2, instruct them to update their robot.

# Risk assessment
low


smb2268 (approved):
lgtm! fixed the linting error

PR #13132, opened 2023-07-19
fix(api): add yocto to architectures that observe the disabledLogAggregation setting

b-cooper:
# Overview

Add the advanced setting for disabled log aggregation for yocto system architecture.

Closes RSS-289

# Risk assessment
low

### DIFF
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -193,7 +193,7 @@ settings = [
     ),
 ]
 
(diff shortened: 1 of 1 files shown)


SyntaxColoring (approved):
Makes sense to me. Thanks for pinpointing this. 💉

sanni-t (approved):
Verified that the Log aggregation setting becomes available on the flex with this branch.
You probably already know this but just wanted to note this somewhere that it doesn't actually update the log settings on the robot yet because syslog doesn't seem to set up for sending the logs upstream yet.

PR #13135, opened 2023-07-19
feat(app,components): add intervention modal to ODD

brenthagen:
# Overview

this PR makes necessary changes to the intervention modal and base modal components to display different layout/styling on ODD and desktop. adds the intervention modal to the RunningProtocol ODD page.

re RLAB-323, closes RLAB-321, RLAB-324, RLAB-325

Screen Shot 2023-07-19 at 3 13 31 PM
Screen Shot 2023-07-19 at 3 13 39 PM
Screen Shot 2023-07-19 at 3 15 34 PM
Screen Shot 2023-07-19 at 3 15 43 PM

# Test Plan

 - update unit tests, intervention modal story

# Changelog

 - Adds intervention modal to ODD
 - Updates Modal and ModalHeader to allow custom styling

# Review requests

 - test with move labware commands, desktop/ODD, ideally on a robot/dev kit

# Risk assessment

low


app/src/molecules/Modal/ModalHeader.tsx
```diff
@@ -25,7 +31,7 @@ export function ModalHeader(props: ModalHeaderBaseProps): JSX.Element {
       flexDirection={DIRECTION_ROW}
       justifyContent={JUSTIFY_SPACE_BETWEEN}
       alignItems={ALIGN_CENTER}
-      borderRadius={`${BORDERS.borderRadiusSize3} ${BORDERS.borderRadiusSize3} 0px 0px`}
```
koji, line 28: I think we still need this for other normal modals.

Screenshot 2023-07-20 at 2 41 29 PM

app/src/assets/localization/en/protocol_command_text.json
koji, line 1: Could you order all items by VSCode Sort Lines asc?
Actually VSCode sorting result is a little bit weird but right now all jsons are re-ordered by it.

https://opentrons.slack.com/archives/CSCLVUW3C/p1683831830189419

app/src/organisms/InterventionModal/PauseInterventionContent.tsx
```diff
@@ -26,15 +38,43 @@ export function PauseInterventionContent({
...
+  border-radius: ${BORDERS.radiusSoftCorners};
+  grid-gap: ${SPACING.spacing6};
+  padding: ${SPACING.spacing16};
+  @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
+    align-self: ${ALIGN_CENTER};
+    background-color: ${COLORS.light1};
+    border-radius: ${BORDERS.borderRadiusSize3};
+    grid-gap: ${SPACING.spacing32};
+    padding: ${SPACING.spacing24};
+    min-width: 36.5rem;
```
koji, line 60: maybe fix width instead of min-width to avoid changing the size?


https://github.com/Opentrons/opentrons/assets/474225/831226cf-f5e2-4f6e-9214-96e49991574b

app/src/molecules/Modal/ModalHeader.tsx
brenthagen: yeah, i see this happening occasionally in the app. the way we have the modal containers laid out is a bit confusing, with border radius set on both outer and inner containers, but I can move the rest `styleProps` from the outer modal container to the children container to get the border. probably when modal unification happens this JSX structure should change.

app/src/organisms/InterventionModal/PauseInterventionContent.tsx
brenthagen: i'm not sure how common a 100-hour plus pause is - min-width at that value is a compromise to get close to the desired styling in figma without resizing for typical `--:--:--` pause times. maybe a wider min-width is fine to allow 100-hour plus pauses. how does fixed width help instead of min-width? fixed width will overflow for large numbers - is overflow better than resizing in that case?

app/src/molecules/Modal/ModalHeader.tsx
```diff
@@ -38,13 +46,13 @@ export function ModalHeader(props: ModalHeaderBaseProps): JSX.Element {
             marginRight={SPACING.spacing16}
           />
         ) : null}
-        <StyledText
+        <Flex
```
b-cooper: Is this one-off text styling? I'm curious why we're changing to using a `<Flex>` here

b-cooper (approved):
Changes look good to me 👍

app/src/molecules/Modal/ModalHeader.tsx
brenthagen: initially, the PR changed the type of the title prop from `string` to `ReactNode` to pass an icon to the header, so this had to change to a div to make the html semantic (no `<p>` within a `<p>`, etc). but after finalizing the styling, i don't think the title change to a node is necessary, so i'll revert it.

PR #13138, opened 2023-07-20
feat(api, hardware-testing): Support liquid-probe in pipette QC for PVT build

andySigler:
# Overview

This PR adds liquid probing to the pipette QC script (single-channels tested only for now). Accuracy and precision are recorded to that CSV.

This PR also makes a few modification to `OT3API.liquid_probe()` so that simulation was possible, and to better match desired physical behavior during the QC test.

Other things added to the script:

- Support new revision of "pressure-fixture"
- Measure tip-presence flag assembly tolerances
- Fix current/speed script to work with new stall-detection features
- Run deck-probe during pipette QC
- Increase plunger acceleration to `1500` mm/sec^2 during current/speed test (to match gravimetric accelerations)

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
lgtm

PR #13140, opened 2023-07-20
feat(hardware-testing): Added hardware testing usb package

vegano1:
# Overview

The hardware-testing package is used for testing various parts of the Flex, however in order to use it you have to have a computer that has all of our packages, and dependencies and has been set up to run our software. This poses some friction as it requires a lot to get up and running. This pr solves this problem by creating a hardware-testing package, that can be run from a USB drive plugged directly into the Flex.

# Test Plan

- [x] Use `make -C hardware-testing setup-usb-module` and make sure it generates a **/dist/hardware_testing_usb-*.tar.gz** file
- [x] Use `make -C hardware-testing setup-usb-module usb_dir="<usb dir>"` and make sure a **hardware_testing_usb** dir is extracted to the usb dir specified.
- [x] Make sure you can run hardware-testing scripts after running `./setup.sh` from the Flex
- [x] Make sure you can run hardware-testing scripts even after reboots
- [x] Make sure you cant run hardware-testing scripts if the USB thumb drive has been removed.
- [x] Make sure you cant run hardware-testing scripts after running `./setup teardown` and re-login.
- [x] Make sure the script does an auto teardown if the USB is not found.

# Changelog

- Added `setup-usb-module` to hardware-testing Makefile, which creates a USB package
- Added hardware-testing entry to scripts/python_build_utils.py so we can get the package version
- Added correct versioning based on git tags when producing sdist/wheel files

# Review requests

- If we want to have hardware-testing logging go to the thumb drive that should be a separate pr.
- Anything I might be missing?

# Risk assessment

Low, internal tooling

### DIFF
--- a/hardware-testing/Makefile
+++ b/hardware-testing/Makefile
@@ -9,11 +9,13 @@ ot_project := $(OPENTRONS_PROJECT)
 project_rs_default = $(if $(ot_project),$(ot_project),robot-stack)
 project_ot3_default = $(if $(ot_project),$(ot_project),ot3)
 
-sdist_file = dist/hardware_testing-0.0.1.tar.gz
-wheel_file = dist/hardware_testing-0.0.1-py3-none-any.whl
+package_name = hardware_testing
+package_version = $(call python_package_version,hardware-testing,$(project_ot3_default))
+wheel_file = dist/$(call python_get_wheelname,hardware-testing,$(project_rs_default),$(package_name),$(BUILD_NUMBER))
+sdist_file = dist/$(call python_get_sdistname,hardware-testing,$(project_ot3_default),$(package_name))
+usb_file =   dist/$(package_name)-usb-$(package_version).tar.gz
 # Find the branch, sha, version that will be used to update the VERSION.json file
-usb_fileversion_file = $(call python_get_git_version,hardware-testing,$(project_ot3_default),hardware-testing)
-usb_file = dist/hardware_testing_usb-0.0.1.tar.gz
+version_file = $(call python_get_git_version,hardware-testing,$(project_ot3_default),hardware-testing)
 
 # These variables can be overriden when make is invoked to customize the
 # behavior of pytest. For instance,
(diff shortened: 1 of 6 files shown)


hardware-testing/Makefile
```diff
@@ -5,9 +5,15 @@ include ../scripts/python.mk
 
 SHX := npx shx
 
+ot_project := $(OPENTRONS_PROJECT)
+project_rs_default = $(if $(ot_project),$(ot_project),robot-stack)
+project_ot3_default = $(if $(ot_project),$(ot_project),ot3)
+
 sdist_file = dist/hardware_testing-0.0.1.tar.gz
```
sfoster1: this probably wants to pattern capture the version

hardware-testing/Makefile
```diff
@@ -5,9 +5,15 @@ include ../scripts/python.mk
...
 wheel_file = dist/hardware_testing-0.0.1-py3-none-any.whl
-
+# Find the branch, sha, version that will be used to update the VERSION.json file
+usb_fileversion_file = $(call python_get_git_version,hardware-testing,$(project_ot3_default),hardware-testing)
+usb_file = dist/hardware_testing_usb-0.0.1.tar.gz
```
sfoster1: as does this

hardware-testing/hardware_testing/tools/usb-package/setup.sh
```diff
@@ -0,0 +1,150 @@
...
+	fi
+
+	echo "Writing usb-package env profile - $ENV_PROFILE"
+
+cat <<EOF > $ENV_PROFILE
+#!/usr/bin/env sh
+
+# check that the hardware-tesing dir in the usb device exists
+if [ ! -d $USB_DIR ]; then
+	echo "################## WARNING ##################"
```
sfoster1: I think we should consider having this case auto-teardown. The reason is that this will be used on the line, and if the line forgets to run `teardown` on a robot then it'll ship with it installed and printing this on every new ssh login which customers will see. If instead we disable it, all that has to happen is they have to run `setup` again.

sfoster1 (changes requested):
This will print the error message in the console _at login_ if the usb drive isn't plugged in, right? We shouldn't do that because it will happen every time if we forget to remove it before shipping a robot. In that case it should either silently fail or silently teardown.

hardware-testing/hardware_testing/tools/usb-package/setup.sh
vegano1: good call!

sfoster1 (approved):
Nice, looks great!

andySigler (approved):

PR #13142, opened 2023-07-20
chore(shared-data): add v1 configs for pvt pipette

sfoster1:
We added these to the configs for the hardware control but did not do it for the app, so add them here too.


ryanthecoder:
Working now, We're going to need to change the display catagory to from GEN3 to FLEX and add the quirks section when we merge back to edge

Laura-Danielle (approved):
lgtm

PR #13144, opened 2023-07-20
feat(api): remove use of LPC offsets during labware movement

sanni-t:
# Overview

Closes RLAB-237

From hardware testing it has been found that adding LPC offsets during labware movement with gripper makes labware pick ups and drops unreliable. But having the pickUp & drop offsets is useful for fine tuning the default positions of labware pickUp & drop. So this PR removes the use of just LPC offsets from API and engine's labware movement commands.

Helps reduce change diff for RLAB-295.

# Test Plan

Just unit tests and linting should suffice since this is removing a functionality that neither science/beta testers not hardware testing is using. Still, a quick test of labware movement on the flex would be helpful in making sure everything still works before making the larger changes of RLAB-295.

# Changelog

- removed `use_pick_up_location_lpc_offset` and `use_drop_location_lpc_offset` from `protocol_context.move_labware` args
- removed `usePickUpLocationLpcOffset` and `useDropLocationLpcOffset` from `moveLabware` engine command
- renamed any args/ classes that were called "experimental"
- added a missing sync_client test

# Review requests

- It's a very straightforward removal of args & params so should be an easy review

# Risk assessment

Low. Just a risk about 'whether the lpc offsets could ever be useful?'. But we can always put them back in if that happens.


api/src/opentrons/protocol_engine/types.py
```diff
@@ -94,11 +94,9 @@ class LabwareMovementStrategy(str, Enum):
 
 # TODO (spp, 2022-12-14): https://opentrons.atlassian.net/browse/RLAB-237
```
SyntaxColoring: Looks like you can get rid of this `# TODO`.

Also, is this intended to be both a `@dataclass` and a `BaseModel`?

SyntaxColoring (approved):
Code LGTM, thanks! Nice catch on the missing `SyncClient` test.

api/src/opentrons/protocol_engine/types.py
sanni-t: Very good catch! Made it a regular dataclass

PR #13145, opened 2023-07-20
chore(docs): update styling of admonition blocks

ecormany:
# Overview

This PR improves the styling of admonitions (e.g. notes and warnings) in the Python API docs. Previously, admonitions were introduced with `Note:` or `Warning:` in very large text — comparable to an H2 on the page — which broke up the flow of the page too much.

These CSS changes reduce the admonition title font size and use background colors from the Opentrons color palette. It also makes text in nested list items the same color as all other body text. Those items used to be in grey, which only reduced their readability, especially when they appeared inside admonitions.

# Test Plan

Check out the styles in [the sandbox](http://sandbox.docs.opentrons.com/docs-admonition-css/v2/).

# Changelog

- reduced `admonition-title` font size
- Opentrons official colors for `background-color` of notes and warnings
- adjusted padding in admonition blocks

# Review requests

New style is better, yes/yes?

old
image

new
image

# Risk assessment

v low, docs style only


api/docs/static/override_sphinx.css
```diff
@@ -197,3 +196,23 @@ table.docutils.field-list {
...
     padding-left: 30px;
 }
+
+/* Adjustments to Note and Warning admonitions */
+div.admonition {
+    padding-top: 20px;
+}
+
+div.admonition p.admonition-title {
+    font-size: 18px;
```
SyntaxColoring, line 206: Should we use `130%` instead of `18px` to slot into [the theme's default heading hierarchy](https://github.com/sphinx-doc/alabaster/blob/7a9df17b6366c96baba6bf5e14f9b6ff6bf7defa/alabaster/static/alabaster.css_t#L261-L266)? (130% computes to 18.2px for me.) I'm not sure why [they don't do that themselves](https://github.com/sphinx-doc/alabaster/blob/7a9df17b6366c96baba6bf5e14f9b6ff6bf7defa/alabaster/static/alabaster.css_t#L298).

On the other hand, the last time I knew anything about good CSS was like 2014, so ignore me if this seems wrong to you at all.

api/docs/static/override_sphinx.css
```diff
@@ -197,3 +196,23 @@ table.docutils.field-list {
...
+    margin-bottom: 0;
```
SyntaxColoring: Is this `margin-bottom: 0` doing anything for you? I don't notice any change when I toggle it on and off in Chrome devtools.

SyntaxColoring (approved):
👨‍🎨

api/docs/static/override_sphinx.css
ecormany: ah i see, it contradicts a line in `alabaster.css` but it's redundant with another line in `override_sphinx.css`.

api/docs/static/override_sphinx.css
ecormany: oof, it's a real hodgepodge of `em`, `rem`, `px`, and `%` in there. this is directly overriding `24px` in `alabaster.css`, so i used the same units.

PR #13146, opened 2023-07-21
feat(api, robot-server): add estop state machine

fsinapi:
# Overview

Adds a state machine to handle the Estop status on the hardware controller, and makes it available to the robot server.

The state machine looks like this:
image

This PR **does not** block out any movements when the Estop is in the "logically-engaged" state, because the only way to clear that is through the HTTP endpoint and if the app doesn't give that option we'd be blocking all movements until the robot server restarts.

# Test Plan

Threw it out a robot and walked through all of the state transitions in the diagram. Checked the state & acknowledged the `logically-engaged` state with the endpoints under `/robot/control`. Everything worked correctly, physical changes all seemed to get detected by the rear panel.

# Changelog

- Added message definition to request estop state (this already existed in the rear panel firmware?)
- Added Estop Detector to listen for binary messages, parse them, and forward them
- Added Estop State machine to listen for the messages from the Detector and transition through the required state machine
- Hooked up `/robot/control` endpoints to hardware controller

# Review requests

# Risk assessment

Pretty low until we hook this up to block movements in a later PR!


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1162,3 +1169,18 @@ def _door_listener(msg: BinaryMessageDefinition) -> None:
...
+        if self._drivers.usb_messenger is None:
+            return False
+        self._estop_detector = await EstopDetector.build(
+            usb_messenger=self._drivers.usb_messenger
+        )
+        self._estop_state_machine = EstopStateMachine(self._estop_detector)
+        return True
+
+    @property
+    def estop_state_machine(self) -> Optional[EstopStateMachine]:
```
sfoster1: I don't love the difference being exposed in this way (with it returning an optional) - I'd rather have the controller and the simulator return a guaranteed object that just stays locked in "disengaged" if the backing hardware isn't present. It'll save us a bunch of little tests in the hardware controller.

api/src/opentrons/hardware_control/estop_state.py
```diff
@@ -0,0 +1,135 @@
...
+    def remove_listener(self, listener: HardwareEventHandler) -> None:
+        """Remove an existing hardware event listener for estop detector changes."""
+        if listener in self._listeners:
+            self._listeners.remove(listener)
+
+    def detector_listener(self, summary: EstopSummary) -> None:
+        """Callback from the detector."""
+        self._handle_state_transition(new_summary=summary)
+
+    def _transition_from_physically_engaged(self) -> None:
```
sfoster1: For testability can we make these take summaries as arguments and return new states?

sfoster1 (changes requested):
Looks good overall, but I think as we add more stuff to the hardware controller that depends on estop stuff we'll really thank ourselves if we make the upward interface of the backend non-`Optional`.

api/src/opentrons/hardware_control/backends/ot3controller.py
fsinapi: 👌

api/src/opentrons/hardware_control/estop_state.py
fsinapi: 38e51f0f9ff4f6c4d6c02bfa4fca564c0bb804d1

api/src/opentrons/hardware_control/backends/ot3controller.py
fsinapi: f826a1be698004040621d02cdcc9e3b833104efc

sfoster1 (approved):
Excellent, looks great!

PR #13147, opened 2023-07-21
feat(api): block robot movement when Estop is engaged or missing

fsinapi:
# Overview

If the OT-3 Estop is engaged, or if there is no Estop plugged in, the robot should not be allowed to move. This includes the "logically engaged" state, where an Estop has been pressed but not acknowledged.

We block these movements with a simple function wrapper that checks the current estop status. If it is anything other than "disengaged," the OT3Controller will refuse to move and it will raise one of our nice exceptions about it.

# Test Plan

On an OT-3, I tested moving the Estop state machine through all 4 of its states and sending a POST request to home the robot. In any state other than Disengaged, the robot responded with an error with the expected error message (with the correct error code!)

# Changelog

* Added a function wrapper around OT3Controller movement functions to block movement if the estop is enabled.
* Added an option `estopNotRequired` that will allow the hardware controller to function with no estop connected.

# Review requests

Do we need a way to disable this? I'm mostly thinking of any hardware-testing stuff that might not have an estop connected.

# Risk assessment

If there _are_ situations where we need to run OT3API functions without an Estop, this is pretty annoying.

Also, we should NOT merge this until the App supports acknowledging and clearing the Estop state. I will wait on @koji to decide when it is ready :)


sfoster1 (approved):
Looks good but yes we should have a way to disable it. A feature flag that we do not expose via HTTP would work.

fsinapi:
> Looks good but yes we should have a way to disable it. A feature flag that we do not expose via HTTP would work.

Added the feature flag and tested that 1) it lets movements run without an estop, and 2) it does not show up on the HTTP endpoints

fsinapi:
With latest commits, tested that the `estopNotRequired` option only shows up in HTTP _after_ it's been set to True, and tested that if the option has been set then the Estop State endpoints never return the "not present" state

PR #13148, opened 2023-07-21
chore: add ot-2 build starting

sfoster1:
And make ot-3 builds explicitly internal release

### DIFF
--- a/.github/workflows/start-internal-release-ot2-build.yaml
+++ b/.github/workflows/start-internal-release-ot2-build.yaml
@@ -27,7 +27,7 @@ jobs:
           owner: opentrons
           repo: buildroot
           workflow-id: build.yml
-          ref: main
+          ref: opentrons-develop
           inputs: |
             {
               "buildroot-ref": "-",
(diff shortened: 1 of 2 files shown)


sfoster1:
Well will ya look at that https://github.com/Opentrons/opentrons/actions/runs/5625917004 https://github.com/Opentrons/buildroot/actions/runs/5625917954 (actual build might fail because some workflow stuff isn't merged)

.github/workflows/start-internal-release-ot2-build.yaml
```diff
@@ -0,0 +1,59 @@
+name: 'Start Internal-Release OT-2 build'
+on:
+  push:
+    branches:
+      - edge
+      - '*internal-release*'
+    tags:
+      - ot3@*
```
vegano1, line 8: just for testing?

Laura-Danielle (approved):

.github/workflows/start-internal-release-ot2-build.yaml
sfoster1: No this is intentional. Remember, this tag means "internal release" now not just "ot3 build". There will be duplicates of this workflow and the ot3 one for normal releases that both use `v*` later.

PR #13150, opened 2023-07-24
chore(docs): single-page Python docs option

ecormany:
# Overview

We'd like to have the option to render the entire docs on a single page (as a replacement for having a PDF of the docs). The way to do this is to include every page in a separate, single RST file, which isn't listed in the docs table of contents.

RST is wild and lets you use basically whatever characters you want as underlines to indicate different header levels. This is fine when rendering separate pages, but if there's variation it breaks the single page. This PR implements a single hierarchy of symbols across all the doc pages:

```
# with overline, for parts = Title of entire docs site
* with overline, for chapters = H1 / title of page
= for sections = H2
- for subsections = H3
^ for subsubsections = H4
```

# Test Plan

Compare the TOC and headers in this branch to the live docs (or `edge`) and make sure the hierarchy is the same.

# Changelog

- changed header underline/overline characters
- adds `single_page.rst`

# Review requests

# Risk assessment

v low, docs formatting only


ecormany:
Oof, too much has changed since I opened this draft PR. Better to close for now than try to sort out the conflicts.

PR #13151, opened 2023-07-24
feat(hardware-testing): Belt calibration script defaults to just calibrate and not test

andySigler:
# Overview

This PR makes the belt-calibration script in hardware-testing more useful to the QC team, so they can run belt-calibration with one command.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
I like it if you do!

PR #13154, opened 2023-07-24
feat(api, shared-data): Increase plunger-acceleration only during aspirate/dispense/blow-out

andySigler:
# Overview

This PR defines a plunger's acceleration in shared-data, so that it can be used by the hardware controller during calls to aspirate/dispense/blowout.

The reason we can't just have 1x acceleration for all moves is:
- Default accelerations (during homing and positional moves) must be slower, to overcome static friction buildup
- Aspirate/dispense/blow-out accelerations must be faster, so it's possible to transfer 1uL (plus improved %CV)

We are confident in the performance given these motion parameters because they have been used in testing for some time now. The gravimetric script has overcome this in the past in two ways:
1. Initially we used a patch file to overwrite the default acceleration, but this applied to 100% of plunger movements
2. Later we changed it so the acceleration was updated as intended, during just the aspirate/dispense/blow-out motions, by directly accessing the hardware api during the script run

# Test Plan

- [x] Confirm visually that accelerations are changing
- [x] Confirm with logs that accelerations are converting to the correct mm/sec/sec value

# Changelog

# Review requests

I wasn't sure what to do with the old pipettes in shared-data. The OT2 code will never use these values so I gave them dummy ones.

# Risk assessment

The change in motion at the core of this PR is low risk, because they have already been tested extensively.


sfoster1 (approved):
Code looks good to me! Nice.

Laura-Danielle (approved):
Added test, code looks good to me.

sfoster1 (approved):
Looks good to me!

PR #13156, opened 2023-07-24
fix(api): pick_up_tip in 2.14+ protocols does not pick up tips before starting tip

jbleon95:
# Overview

Closes RESC-145.

In api version 2.14 and above (which uses the Protocol Engine), a regression was introduced that made it that if you chose a starting tip and used `pick_up_tip` without any arguments, the wells after the starting tip would be `A1`, `B1`, `C1`, etc instead of returning the wells that followed the starting one. This has been addressed so that the pick up tip behavior is identical to versions 2.13 and below.

# Test Plan

Tested the following protocol with both API levels 2.13 and 2.14 on an OT-2 to verify that they picked up tips in `B2`, `C2`, `D2`...`A3`

```
metadata = {
    'protocolName': 'Tip Regression Behavior',
}

requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.13"
}

def run(context):
    tip_rack = context.load_labware("opentrons_96_tiprack_20ul", location=5)
    pipette = context.load_instrument("p20_single_gen2", mount="left", tip_racks=[tip_rack])

    pipette.starting_tip = tip_rack["B2"]

    for _ in range(8):
        pipette.pick_up_tip()
        pipette.drop_tip()
```

# Changelog

- fixed `TipView.get_next_tip` to drop potential wells before the starting tip well

# Review requests

# Risk assessment

Low, behavior only changes if starting tip is used, and that has been tested to ensure it matches with previous behavior.

### DIFF
--- a/api/docs/v2/versioning.rst
+++ b/api/docs/v2/versioning.rst
@@ -160,9 +160,9 @@ This version introduces support for the Opentrons Flex robot, instruments, modul
   
   - By default, repeated calls to :py:meth:`.drop_tip` cycle through multiple locations above the trash bin to prevent tips from stacking up.
   
-- Known limitations
+- Bug fixes
 
-  - :py:attr:`.InstrumentContext.starting_tip` is not respected on the second and subsequent calls to :py:meth:`.InstrumentContext.pick_up_tip` with no argument.
+  - :py:attr:`.InstrumentContext.starting_tip` is now respected on the second and subsequent calls to :py:meth:`.InstrumentContext.pick_up_tip` with no argument.
 
   
 
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -101,7 +101,13 @@ class InstrumentContext(publisher.CommandPublisher):
     @property  # type: ignore
     @requires_version(2, 0)
     def starting_tip(self) -> Union[labware.Well, None]:
-        """The starting tip from which the pipette pick up"""
+        """
+        Which well of a tip rack the pipette should start at when tracking tips.
+
+        The first call to :py:meth:`.pick_up_tip` with no argument will pick up from
+        that well. Subsequent calls will follow the ordering specified by the labware
+        definition and :py:meth:`.Labware.wells`.
+        """
         return self._starting_tip
 
     @starting_tip.setter
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -102,11 +102,16 @@ class InstrumentContext(publisher.CommandPublisher):
     @requires_version(2, 0)
     def starting_tip(self) -> Union[labware.Well, None]:
         """
-        Which well of a tip rack the pipette should start at when tracking tips.
+        Which well of a tip rack the pipette should start at when automatically choosing tips to pick up.
 
-        The first call to :py:meth:`.pick_up_tip` with no argument will pick up from
-        that well. Subsequent calls will follow the ordering specified by the labware
-        definition and :py:meth:`.Labware.wells`.
+        See :py:meth:`.pick_up_tip()`.
+
+        .. note::
+
+            In robot software versions 6.3.0 and 6.3.1, protocols specifying API level 2.14 would
+            not respect ``starting_tip`` on the second and subsequent calls to
+            :py:meth:`.InstrumentContext.pick_up_tip` with no argument. This is fixed for all API
+            levels as of robot software version 7.0.0.
         """
         return self._starting_tip
 
@@ -666,6 +671,9 @@ class InstrumentContext(publisher.CommandPublisher):
 
         If no location is passed, the Pipette will pick up the next available
         tip in its :py:attr:`InstrumentContext.tip_racks` list.
+        Within each tip rack, tips will be picked up in the order specified by
+        the labware definition and :py:meth:`.Labware.wells`.
+        To adjust where the sequence starts, see :py:obj:`.starting_tip`.
 
         The tip to pick up can be manually specified with the `location`
         argument. The `location` argument can be specified in several ways:
--- a/api/src/opentrons/protocol_engine/state/tips.py
+++ b/api/src/opentrons/protocol_engine/state/tips.py
@@ -167,14 +167,11 @@ class TipView(HasState[TipState]):
                 return next(iter(wells))
 
         else:
-            for well_name, tip_state in wells.items():
-                seen_start = (
-                    starting_tip_name is None
-                    or well_name == starting_tip_name
-                    or wells[starting_tip_name] == TipRackWellState.USED
-                )
+            if starting_tip_name is not None:
+                wells = _drop_wells_before_starting_tip(wells, starting_tip_name)
 
-                if seen_start and tip_state == TipRackWellState.CLEAN:
+            for well_name, tip_state in wells.items():
+                if tip_state == TipRackWellState.CLEAN:
                     return well_name
 
         return None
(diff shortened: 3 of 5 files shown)


y3rsh:
@DerekMaggio would you please look at the failing G-code test?  I think this is correctly identifying a change in the G-Code capture due to a software change.

DerekMaggio:
@y3rsh, I agree it is. 
I will update the file and push it to this branch

DerekMaggio:
....So they passed when I reran them

ecormany:
Can this PR also fix up the docstring for `InstrumentContext.starting_tip`? 🫣

https://github.com/Opentrons/opentrons/blob/6ae02e467f59cd9e567caf5dd65ae210f3da2057/api/src/opentrons/protocol_api/instrument_context.py#L104

Something like 
```
Which well of a tip rack the pipette should start at when tracking tips. The first call to 
:py:meth:`.pick_up_tip` with no argument will pick up from that well. Subsequent calls 
will follow the ordering specified by the labware definition and :py:meth:`.Labware.wells`.
```

ecormany:
Now in `edge`, this Known Limitation warning should be changed to a Bug Fix when this is implemented.
https://github.com/Opentrons/opentrons/blob/a639bae50dc1f5974c4070af32295dc81ae64a20/api/docs/v2/versioning.rst?plain=1#L163-L165

api/docs/v2/versioning.rst
```diff
@@ -160,9 +160,9 @@ This version introduces support for the Opentrons Flex robot, instruments, modul
   
   - By default, repeated calls to :py:meth:`.drop_tip` cycle through multiple locations above the trash bin to prevent tips from stacking up.
   
-- Known limitations
+- Bug fixes
 
-  - :py:attr:`.InstrumentContext.starting_tip` is not respected on the second and subsequent calls to :py:meth:`.InstrumentContext.pick_up_tip` with no argument.
+  - :py:attr:`.InstrumentContext.starting_tip` is now respected on the second and subsequent calls to :py:meth:`.InstrumentContext.pick_up_tip` with no argument.
```
SyntaxColoring, line 165: nice diff.

However, since [we're fixing this retroactively for PAPIv2.14](https://github.com/Opentrons/opentrons/pull/13156#issuecomment-1688619946), I think it's misleading for this note to appear here under the PAPIv2.15 release notes. I think we should either:

* Remove this note, and instead mention this in the robot software changelog (`api/release-notes.md`).
* Remove this note, and instead mention this as a historical quirk in the `starting_tip` reference.
* Both of the above.

But I'll defer to @ecormany.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -101,7 +101,13 @@ def api_version(self) -> APIVersion:
...
     @requires_version(2, 0)
     def starting_tip(self) -> Union[labware.Well, None]:
-        """The starting tip from which the pipette pick up"""
+        """
+        Which well of a tip rack the pipette should start at when tracking tips.
+
+        The first call to :py:meth:`.pick_up_tip` with no argument will pick up from
+        that well. Subsequent calls will follow the ordering specified by the labware
+        definition and :py:meth:`.Labware.wells`.
+        """
```
SyntaxColoring: I would consider making `pick_up_tip()` the single source of truth for how automatic tip tracking works, since it has the `location=None` arg. Then `starting_tip` and `pick_up_tip()` can link to each other.

I'll again defer to @ecormany though.

api/src/opentrons/protocol_engine/state/tips.py
```diff
@@ -167,14 +167,11 @@ def get_next_tip(  # noqa: C901
                 return next(iter(wells))
 
         else:
-            for well_name, tip_state in wells.items():
-                seen_start = (
-                    starting_tip_name is None
-                    or well_name == starting_tip_name
-                    or wells[starting_tip_name] == TipRackWellState.USED
```
SyntaxColoring, line 174: I'm trying to figure out what this `or wells[starting_tip_name] == TipRackWellState.USED` line was trying to do before. Apparently this was the bug, but do you know what its original intent was?

api/src/opentrons/protocol_engine/state/tips.py
SyntaxColoring, line 1: Since this changes the analysis of preexisting public protocols, let's bump this constant from `"initial"` to `"2"`, and leave a comment that the prior value was `"initial"`.

https://github.com/Opentrons/opentrons/blob/e896522014031558660d41b42e12ba023daf780b/robot-server/robot_server/protocols/analysis_store.py#L55

This will help us automatically trigger reanalysis in the future, if we deem that necessary.

api/tests/opentrons/protocol_engine/state/test_tip_state.py
```diff
@@ -212,6 +212,80 @@ def test_get_next_tip_skips_picked_up_tip(
     assert result == result_well_name
 
 
+def test_get_next_tip_with_starting_tip(
```
SyntaxColoring, line 215: I would consider also adding tests for 8-Channels, unless you think they're already covered sufficiently.

SyntaxColoring (commented):
Nice, looks good to me overall. A few small comments.

SyntaxColoring:
There was a question of whether we should:

* Just fix the bug. This would fix it retroactively for PAPIv2.14. So certain Python files would behave differently on robot software v7.0 than on v6.3. 
* Or, preserve the buggy behavior in PAPIv2.14 but fix it in PAPIv2.15, by checking the protocol's `apiLevel`.

@sanni-t, @jbleon95, @laviera and I discussed this. We're going with just fixing the bug.

There are certain cases where, given our current architecture, it's impractical to preserve bug-for-bug compatibility. This is one of those. @sanni-t brought up a prior case related to transfer planning. I brought up the whole Magnetic Module GEN1 engage height thing (RSS-109 etc.), where our attempts to preserve bug-compatibility have backfired and caused lasting confusion and *additional* bugs years later.

There are architectural things that we can do to improve this, which I have been meaning to write up, but they would be a long-term post-Flex thing. So, for now, we think just fixing the bug is the best way to move forward.

api/docs/v2/versioning.rst
ecormany: I like doing both.
- For release notes, I've given myself a reminder to do this in RTC-307
- For the reference, how about:
```
.. note::
    In robot software versions 6.3.0 and 6.3.1, protocols specifying API level 2.14 would 
    not respect ``starting_tip`` on the second and subsequent calls to
    :py:meth:`.InstrumentContext.pick_up_tip` with no argument. This is fixed for all API
    levels as of robot software version 7.0.0.
```

api/src/opentrons/protocol_api/instrument_context.py
ecormany: Is the suggestion here to strike ~~and :py:meth:`.Labware.wells`~~ ?
I'm good with that.

SyntaxColoring: That and more. My suggestion is for this to say:

```
Which well of a tip rack the pipette should start at when automatically choosing tips to pick up.

See :py:meth:`.pick_up_tip()`.
```

And then, where `pick_up_tip()` currently says:

```
If no location is passed, the Pipette will pick up the next available
tip in its :py:attr:`InstrumentContext.tip_racks` list.
```

It would change to say:

```
If no location is passed, the Pipette will pick up the next available
tip in its :py:attr:`InstrumentContext.tip_racks` list.
Within each tip rack, tips will be picked up in the order specified by
the labware definition and :py:meth:`.Labware.wells`.
To adjust where the sequence starts, see :py:obj:`.starting_tip`.
```

api/src/opentrons/protocol_engine/state/tips.py
jbleon95: I believe the intention of this line was to make sure the `starting_tip was skipped if it was already used, but considering the following line is checking if its clean, in practice I don't think it did anything

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -666,6 +671,9 @@ def pick_up_tip(
 
         If no location is passed, the Pipette will pick up the next available
         tip in its :py:attr:`InstrumentContext.tip_racks` list.
+        Within each tip rack, tips will be picked up in the order specified by
+        the labware definition and :py:meth:`.Labware.wells`.
+        To adjust where the sequence starts, see :py:obj:`.starting_tip`.
```
ecormany, line 676: would maybe go so far as to say 
```
use :py:obj:`.starting_tip`
```
…but that's me being a fairly picky editor 😁 the information here is great! :shipit:

SyntaxColoring (approved):

PR #13158, opened 2023-07-24
feat(app): ODD add error codes to protocol failed modal

shlokamin:
# Overview
This PR adds an error code to the ODD's protocol failed modal. It does so by recursively looking for the "highest priority" error, which is the error with the lowest status code (preferring non generic status codes). 

i.e. `2000` is higher priortity than `3000`, but `2001` is higher priority than `2000`

closes RCORE-799

# Test Plan

- Tested on ODD using the protocol attached

Screenshot 2023-07-24 at 5 46 06 PM

# Changelog

- show error code in run failed modal on ODD

# Risk assessment

Low


app/src/organisms/OnDeviceDisplay/RunningProtocol/RunFailedModal.tsx
```diff
@@ -134,3 +131,63 @@ const SCROLL_BAR_STYLE = css`
     border-radius: 11px;
   }
 `
+
+const _getHighestPriorityError = (error: RunError): RunError => {
```
shlokamin, line 136: gonna move all of these utils into a shared place in a follow up PR since ill need to use the public function for the desktop app

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunFailedModal.tsx
```diff
@@ -20,13 +21,7 @@ import { SmallButton } from '../../../atoms/buttons'
...
 
 import type { ModalHeaderBaseProps } from '../../../molecules/Modal/types'
-
-interface RunError {
-  id: string
-  errorType: string
-  createdAt: string
-  detail: string
-}
+import type { RunError } from '@opentrons/api-client'
```
koji, line 24: We can put this line before `ModalHeaderBaseProps`

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunFailedModal.tsx
```diff
@@ -44,11 +39,13 @@ export function RunFailedModal({
   const { stopRun } = useStopRunMutation()
   const [isCanceling, setIsCanceling] = React.useState(false)
 
-  if (errors == null) return null
+  if (errors == null || errors.length === 0) return null
```
koji, line 42: Do we need to keep `errors == null` ?
errors type is RunError[] which errors would not get null or undefined?

koji (approved):
Tested on a DVT and it worked as expected.

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunFailedModal.tsx
shlokamin: yeah sorry i think it was after a weird merge, should be fixed now

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunFailedModal.tsx
shlokamin: right now errors are optional: 
```
interface RunFailedModalProps {
  runId: string
  setShowRunFailedModal: (showRunFailedModal: boolean) => void
  errors?: RunError[]
}
```
because of what we get back from react query, so this is just to make type checking happy. if we want to get rid of this check then inside of `RunSummary` before we render `RunFailedModal` we have to make sure `runRecord?.data.error` is not null

PR #13159, opened 2023-07-24
fix(app) modify play button disabled condition in protocol setup

koji:
# Overview
Currently users can tap the play button in ProtocolSetup screen since `disabled` isn't true when the button looks `disabled`.
Updated `disabled` condition on play button 

### before
Screenshot 2023-08-02 at 5 08 18 PM

### after
Screenshot 2023-08-02 at 5 11 57 PM

close RAUT-508 and RAUT-606

# Test Plan
case1: a protocol that is runnable 
1. tap the protocol
2. tap setup
3. tap play button in ProtocolSetup

case2: a protocol that isn't runnable
1. tap the protocol
2. tap setup
3. tap play button
no reactions on tapping (no background color change)

# Changelog
- update play button condition and update tests

# Review requests

# Risk assessment


koji:
- [ ] test this branch on DVT

PR #13160, opened 2023-07-25
feat(robot-server): Expose tip status in instruments

Laura-Danielle:
# Overview
We are now exposing the tip presence status (which is only valid on the Flex) through the `/instruments` endpoint. Closes [RSS-291](https://opentrons.atlassian.net/jira/software/c/projects/RSS/boards/155?modal=detail&selectedIssue=RSS-291&assignee=60343207783a4600685f4759)

# Test Plan

- Test on Flex, make sure the instruments endpoint returns updated data for the tip presence. You can test this on a single/multi by pushed the ejector flag up. (This feature is NOT yet implemented on the 96 channel and so the data there cannot be trusted).

# Changelog

- Add a `PipetteStateDict` in the hardware controller. This is currently only used in the ot3api, but we can (and should) definitely expand on this state in the future
- Add a separate `get_instrument_state` function in the `ot3api.py` which gets the current state of the tip ejector. Unfortunately we currently have to ask for the state each time this function is called because we currently don't have a state store for the hardware controller as it relates to specific hardware components (such as sensors).

# Review requests

Does it look OK for now? Anything you want me to add and/or take away?

# Risk assessment

Low. New feature add.

[RSS-291]: https://opentrons.atlassian.net/browse/RSS-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/instruments/instrument_models.py
+++ b/robot-server/robot_server/instruments/instrument_models.py
@@ -81,6 +81,15 @@ class PipetteData(BaseModel):
     #  add calibration data as decided by https://opentrons.atlassian.net/browse/RSS-167
 
 
+class PipetteState(BaseModel):
+    """State from an attached pipette."""
+
+    tip_detected: bool = Field(
+        None,
+        description="Physical state of the tip photointerrupter on the Flex. Null for OT-2",
+    )
+
+
 class Pipette(_GenericInstrument[PipetteModel, PipetteData]):
     """Attached pipette info & configuration."""
 
@@ -88,6 +97,7 @@ class Pipette(_GenericInstrument[PipetteModel, PipetteData]):
     instrumentName: PipetteName
     instrumentModel: PipetteModel
     data: PipetteData
+    state: Optional[PipetteState]
 
 
 class Gripper(_GenericInstrument[GripperModelStr, GripperData]):
--- a/robot-server/robot_server/instruments/instrument_models.py
+++ b/robot-server/robot_server/instruments/instrument_models.py
@@ -84,9 +84,10 @@ class PipetteData(BaseModel):
 class PipetteState(BaseModel):
     """State from an attached pipette."""
 
-    tip_detected: bool = Field(
+    tipDetected: bool = Field(
         None,
         description="Physical state of the tip photointerrupter on the Flex. Null for OT-2",
+        alias="tip_detected",
     )
 
 
--- a/robot-server/tests/instruments/test_router.py
+++ b/robot-server/tests/instruments/test_router.py
@@ -35,6 +35,7 @@ from robot_server.instruments.instrument_models import (
     InstrumentCalibrationData,
     BadPipette,
     BadGripper,
+    PipetteState,
 )
 from robot_server.instruments.router import (
     get_attached_instruments,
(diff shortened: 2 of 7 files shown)


sfoster1 (approved):
Looks good to me, and tested on a Flex!

robot-server/robot_server/instruments/instrument_models.py
```diff
@@ -81,13 +81,23 @@ class PipetteData(BaseModel):
     #  add calibration data as decided by https://opentrons.atlassian.net/browse/RSS-167
 
 
+class PipetteState(BaseModel):
+    """State from an attached pipette."""
+
+    tip_detected: bool = Field(
```
SyntaxColoring: 🐫-case for HTTP-facing field names.

```suggestion
    tipDetected: bool = Field(
```

robot-server/tests/instruments/test_router.py
```diff
@@ -217,6 +226,7 @@ def rehearse_instrument_retrievals() -> None:
                     last_modified=None,
                 ),
             ),
+            state=PipetteState(tip_detected=True),
```
SyntaxColoring, line 229:
```suggestion
            state=PipetteState(tipDetected=True),
```

robot-server/tests/instruments/test_router.py
```diff
@@ -237,6 +247,7 @@ def rehearse_instrument_retrievals() -> None:
                     last_modified=None,
                 ),
             ),
+            state=PipetteState(tip_detected=False),
```
SyntaxColoring, line 250:
```suggestion
            state=PipetteState(tipDetected=False),
```

robot-server/robot_server/instruments/instrument_models.py
```diff
@@ -81,13 +81,23 @@ class PipetteData(BaseModel):
...
+
+
 class Pipette(_GenericInstrument[PipetteModel, PipetteData]):
     """Attached pipette info & configuration."""
 
     instrumentType: Literal["pipette"] = "pipette"
     instrumentName: PipetteName
     instrumentModel: PipetteModel
     data: PipetteData
+    state: Optional[PipetteState]
```
SyntaxColoring: Confirmed in-person with @sanni-t:

Instead of what's written now, where `tipDetected` is inside `data.state`:

```jsonc
{
  "data": {
    "instrumentType": "pipette",
    "data": {
      "calibratedOffset": { /* ... */ }
      // ..
    }
    "state": {
        "tipDetected": true
    }
    // ...
  },
  "meta": { /* ... */ }
}
```

We'd like `data.state` to not exist, and `tipDetected` to be directly inside the existing `data.data`, like this:

```jsonc
{
  "data": {
    "instrumentType": "pipette",
    "data": {
      "calibratedOffset": { /* ... */ }
      "tipDetected": true
      // ...
    }
    // ...
  },
  "meta": { /* ... */ }
}
```

This is consistent with a gripper's `jawState`, for example.

vegano1 (approved):

robot-server/robot_server/instruments/instrument_models.py
Laura-Danielle: Just because it's consistent doesn't mean we should stick with what we've previously done. IMO `jawState` should also go into a similarly named `state` component of the body returned from `/instruments` -- though obviously let's not address that till post launch. I've already got sign off on this from JS devs.

`tipDetected` is in fact active state about the pipette whereas things listed in data can be considered "stored" state (albeit it can change from time to time but no where near as frequently as data).

PR #13161, opened 2023-07-25
feat(shared-data): validate all models in the v2 pipette folder

Laura-Danielle:
## Overview

We should actually have a test that passes all of the files into the pydantic validator to ensure that the configurations can be properly loaded. Found a bug from the test! Let me know if you want me to test anything else in particular.


sfoster1 (approved):
Looks good to me! Thank you!

PR #13163, opened 2023-07-25
chore: parse tags in utf8 explicitly in python

sfoster1:
When we parse tags, we didn't explicitly set a codec for parsing the bytes. This is fine in most cases because the locale is set for utf-8, but in the docker container that builds buildroot the locale is C and that means ascii and that means that any tags that have utf-8 characters cause builds to fail.


Laura-Danielle (approved):
lgtm

vegano1 (approved):

PR #13164, opened 2023-07-25
chore(hardware): starting points for moves in plot

sfoster1:
When we plot the can logs, we need to plot points for execute commands - the time of the execute on the x, the previous position on the y - to avoid a problem where movecompletes that are far apart in time plot a line straight from the endpoint of the last MoveComplete to the endpoint of the next MoveComplete which ends up being a weird shallow line.

Using the execute commands isn't perfect but what are you gonna do.


Laura-Danielle (approved):

fsinapi (approved):
👍

vegano1 (approved):

PR #13165, opened 2023-07-25
fix(api): do not open jaw after gripper calibration

ahiuchingau:
# Overview
We don't need to ungrip the jaw whenever we finishes calibration, because holding the jaw closed is the ideal gripper state, we can just leave it as that.


sfoster1 (approved):
Looks grand to me

PR #13168, opened 2023-07-26
fix(robot-server): do not save calibration offsets if an error occurs

Laura-Danielle:
# Overview

A fix for the behavior where pipette offsets get saved even if an error occurs during the calibration flow.

# Test Plan

Follow steps in [this ticket](https://opentrons.atlassian.net/browse/RQA-1066) to reproduce.

# Changelog

- Save the offset inside the robot server function rather than the pipette function
- Ensure that we're raising on edge detection failed errors as well

# Review requests

Please test on a robot, let me know if anything seems off.

# Risk assessment

Low. Should be fixing a small bug.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3_calibration.py
+++ b/api/src/opentrons/hardware_control/ot3_calibration.py
@@ -583,6 +583,7 @@ async def _calibrate_mount(
             InaccurateNonContactSweepError,
             EarlyCapacitiveSenseTrigger,
             CalibrationStructureNotFoundError,
+            EdgeNotFoundError,
         ):
             LOG.info(
                 "Error occurred during calibration. Resetting to current saved calibration value."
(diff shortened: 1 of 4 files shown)


sfoster1 (changes requested):
I like the idea but it's sort of a weird middle ground between a refactor and a new feature.

I think the simplest way to do this is to change `ot3_calibration.calibrate_pipette` to put the `save_instrument_offset_call` after the `finally`? (although we definitely want to restore the old one). But if we want a whole new different function like this, then we should have `calibrate_pipette` call the new function instead of just copy-pasting it; and also we should probably remove `calibrate_pipette` since I don't think anything else would use it now.

Let's do one of those two (either alter `calibrate_pipette`, or implement it via `find_pipette_offset` and/or delete it).

ahiuchingau:
Do you think we'll need to fix the gripper offset saving mechanism as well? I don't see why this behavior would only be happening to the pipettes and not the gripper as well? What am i missing?

Laura-Danielle:
> Do you think we'll need to fix the gripper offset saving mechanism as well? I don't see why this behavior would only be happening to the pipettes and not the gripper as well? What am i missing?

@ahiuchingau the gripper already does the mechanism above. Previously, the pipette was saving calibration inside the try/except of the `ot3_calibration.py` function.

sfoster1 (approved):
Thank you, looks good to me!

PR #13169, opened 2023-07-26
fix(app): close maintenance run modal if run gets deleted

smb2268:
Fix RQA-1061

# Overview

When a maintenance procedure is canceled from the "Remote Activity" takeover modal on ODD, the corresponding modal on the desktop app should close.

# Test Plan

1. Start a pipette or gripper flow from a desktop app
2. Cancel remote activity from the ODD takeover modal
3. See the modal on desktop close within 5 seconds

# Changelog
1. Refactor chainMaintenanceRunCommands to take the runId in the callback instead of on instantiation
2. Update Gripper flows, Pipette flows, and LPC to pass the runId in the callback
3. Add a `useRef` to keep track of the previous maintenance run id - if this was previously a string and becomes null, we close the gripper and pipette flows since this means the run has been deleted

# Review requests
Run through the test plan, look over approach

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/GripperWizardFlows/MovePin.tsx
+++ b/app/src/organisms/GripperWizardFlows/MovePin.tsx
@@ -46,9 +46,9 @@ export const MovePin = (props: MovePinProps): JSX.Element | null => {
   const { t } = useTranslation(['gripper_wizard_flows', 'shared'])
 
   const handleOnClick = (): void => {
-    if (movement === REMOVE_PIN_FROM_REAR_JAW || maintenanceRunId == null) {
+    if (movement === REMOVE_PIN_FROM_REAR_JAW) {
       proceed()
-    } else {
+    } else if (maintenanceRunId != null) {
       const jaw = movement === MOVE_PIN_TO_FRONT_JAW ? 'front' : 'rear'
       createRunCommand({
         maintenanceRunId,
--- a/app/src/organisms/GripperWizardFlows/MovePin.tsx
+++ b/app/src/organisms/GripperWizardFlows/MovePin.tsx
@@ -19,14 +19,14 @@ import calibratingFrontJaw from '../../assets/videos/gripper-wizards/CALIBRATING
 import calibratingRearJaw from '../../assets/videos/gripper-wizards/CALIBRATING_REAR_JAW.webm'
 
 import type { Coordinates } from '@opentrons/shared-data'
-import type { CreateMaintenaceCommand } from '../../resources/runs/hooks'
+import type { CreateMaintenanceCommand } from '../../resources/runs/hooks'
 import type { GripperWizardStepProps, MovePinStep } from './types'
 
 interface MovePinProps extends GripperWizardStepProps, MovePinStep {
   setFrontJawOffset: (offset: Coordinates) => void
   frontJawOffset: Coordinates | null
-  createRunCommand: CreateMaintenaceCommand
   isExiting: boolean
+  createRunCommand: CreateMaintenanceCommand
 }
 
 export const MovePin = (props: MovePinProps): JSX.Element | null => {
--- a/app/src/organisms/GripperWizardFlows/MovePin.tsx
+++ b/app/src/organisms/GripperWizardFlows/MovePin.tsx
@@ -36,6 +36,7 @@ export const MovePin = (props: MovePinProps): JSX.Element | null => {
     goBack,
     movement,
     setFrontJawOffset,
+    maintenanceRunId,
     frontJawOffset,
     createRunCommand,
     errorMessage,
@@ -45,11 +46,12 @@ export const MovePin = (props: MovePinProps): JSX.Element | null => {
   const { t } = useTranslation(['gripper_wizard_flows', 'shared'])
 
   const handleOnClick = (): void => {
-    if (movement === REMOVE_PIN_FROM_REAR_JAW) {
+    if (movement === REMOVE_PIN_FROM_REAR_JAW || maintenanceRunId == null) {
       proceed()
     } else {
       const jaw = movement === MOVE_PIN_TO_FRONT_JAW ? 'front' : 'rear'
       createRunCommand({
+        maintenanceRunId,
         command: {
           commandType: 'home' as const,
           params: {
@@ -63,6 +65,7 @@ export const MovePin = (props: MovePinProps): JSX.Element | null => {
             setErrorMessage(data.error?.detail ?? null)
           }
           createRunCommand({
+            maintenanceRunId,
             command: {
               commandType: 'calibration/calibrateGripper' as const,
               params:
@@ -80,6 +83,7 @@ export const MovePin = (props: MovePinProps): JSX.Element | null => {
                 setFrontJawOffset(data.result.jawOffset)
               }
               createRunCommand({
+                maintenanceRunId,
                 command: {
                   commandType: 'calibration/moveToMaintenancePosition' as const,
                   params: {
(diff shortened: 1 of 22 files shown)


app/src/organisms/GripperWizardFlows/index.tsx
```diff
@@ -66,6 +67,18 @@ export function GripperWizardFlows(
       setMaintenanceRunId(response.data.id)
     },
   })
+  const { data: maintenanceRunData } = useCurrentMaintenanceRun({
+    refetchInterval: 5000,
```
koji: can we have this as const?

app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -116,6 +117,17 @@ export const PipetteWizardFlows = (
     },
     host
   )
+  const { data: maintenanceRunData } = useCurrentMaintenanceRun({
+    refetchInterval: 5000,
```
koji: same

smb2268:
I added a timeout to this and believe it now works as expected. This makes it so that we wait 10 seconds from when a maintenance run is created before we expect that same id to be returned from `useCurrentMaintenanceRun`. After that point, if the id's don't match, or there is no current run, it is safe to assume this procedure was terminated via the ODD and we can close the modal.

koji (commented):
The current implementation closes the wizard without users' action.
I think closeFlow() should be kicked when users click `Terminate activity`

https://github.com/Opentrons/opentrons/assets/474225/42dba12c-b5bb-48a6-b96d-692e1d08ee23

b-cooper:
The buggy behavior that Koji showed in the video points to a weakness in this set timeout approach.  We talked this through in person. @koji, @shlokamin, and I think it may be worth shifting away from using React state to track the created maintenance run id, and instead use the current maintenance run poll (as written in this PR) to act as the source of truth for the active maintenance run linked to the wizard. Instead of a set timeout in the use effect, we now want to monitor the change from a current maintenance run "existing" to "not existing". When this occurs, the wizard should close.

app/src/organisms/GripperWizardFlows/MovePin.tsx
```diff
@@ -43,11 +44,12 @@ export const MovePin = (props: MovePinProps): JSX.Element | null => {
   const { t } = useTranslation(['gripper_wizard_flows', 'shared'])
 
   const handleOnClick = (): void => {
-    if (movement === REMOVE_PIN_FROM_REAR_JAW) {
+    if (movement === REMOVE_PIN_FROM_REAR_JAW || maintenanceRunId == null) {
```
b-cooper: won't the proceed button be disabled in this case?  what is the purpose of adding the `||` to this condition ?

smb2268: yeah it will be disabled, I added this for null protection since maintenanceRunId can't be null when we call `createRunCommand` but I should move this to the `else if` - I'll push that up now

app/src/organisms/GripperWizardFlows/MovePin.tsx
koji, line 1: This is unrelated to your PR, but if you fix typo, that will be awesome. Missing `n`

L28 createRunCommand: CreateMaintenaceCommand --> createRunCommand: CreateMaintenanceCommand

app/src/organisms/GripperWizardFlows/index.tsx
```diff
@@ -49,24 +53,34 @@ export function GripperWizardFlows(
...
+  } = useCreateMaintenanceCommandMutation()
 
   const {
     createMaintenanceRun,
     isLoading: isCreateLoading,
-  } = useCreateMaintenanceRunMutation({
-    onSuccess: response => {
-      setMaintenanceRunId(response.data.id)
-    },
+  } = useCreateMaintenanceRunMutation({})
```
koji: need to pass an empty object?

app/src/organisms/GripperWizardFlows/index.tsx
```diff
@@ -77,24 +91,31 @@ export function GripperWizardFlows(
...
-      })
-      .catch(error => {
-        console.error(error.message)
-        deleteMaintenanceRun(maintenanceRunId)
-        setIsExiting(false)
-        props.onComplete?.()
-      })
+    if (maintenanceRunData?.data.id == null) closeFlow()
+    else {
+      chainRunCommands(
```
koji: Probably using {} would make code more readable.

```suggestion
   if (maintenanceRunData?.data.id == null) {
      closeFlow()
    } else {
```

app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -39,6 +41,9 @@ import { UnskippableModal } from './UnskippableModal'
 
 import type { PipetteMount } from '@opentrons/shared-data'
 import type { PipetteWizardFlow, SelectablePipettes } from './types'
+import type { CommandData } from '@opentrons/api-client'
```
koji, line 43: can you move up this line?

app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -100,22 +104,33 @@ export const PipetteWizardFlows = (
...
+  const { data: maintenanceRunData } = useCurrentMaintenanceRun({
+    refetchInterval: RUN_REFETCH_INTERVAL,
+  })
+  const prevMaintenanceRunId = React.useRef<string | undefined>(
+    maintenanceRunData?.data.id
+  )
+
+  React.useEffect(() => {
+    prevMaintenanceRunId.current = maintenanceRunData?.data.id
+  }, [maintenanceRunData?.data.id])
```
koji, line 116: maybe 

```suggestion
  }, [maintenanceRunData])
```

smb2268: I think keeping this as  `maintenanceRunData?.data.id` makes sense because we don't need this to execute if a different portion of the `maintenanceRunData` such as the `status` changes.

koji: thanks, that makes sense.

app/src/organisms/GripperWizardFlows/index.tsx
```diff
@@ -49,24 +53,34 @@ export function GripperWizardFlows(
...
+  const prevMaintenanceRunId = React.useRef<string | undefined>(
+    maintenanceRunData?.data.id
+  )
+  // this will close the modal in case the run was deleted by the terminate
+  // activity modal on the ODD
+  React.useEffect(() => {
+    if (maintenanceRunData?.data.id == null && prevMaintenanceRunId != null) {
+      closeFlow()
+    }
+  }, [maintenanceRunData, closeFlow])
```
koji: maybe

```suggestion
  }, [maintenanceRunData?.data.id, closeFlow])
```

app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -171,11 +190,25 @@ export const PipetteWizardFlows = (
...
+    chainMaintenanceRunCommands = (
+      commands: CreateCommand[],
+      continuePastCommandFailure: boolean
+    ): Promise<CommandData[]> =>
+      chainRunCommands(
+        maintenanceRunData.data.id,
+        commands,
+        continuePastCommandFailure
+      )
+  }
```
jerader, line 205: this seems the same or similar to the logic in `app/src/organisms/ModuleWizardFlows/index.tsx`, maybe it can be refactored into a util in a followup PR

jerader (approved):
works as expected in my dev environment 🦒

PR #13170, opened 2023-07-26
fix(hardware-testing): Updates gripper assembly QC script to work for PVT build

andySigler:
# Overview

Assembly QC script needed some updates because of recent software changes and updated production spec.
- New force fixture is used to measure force
- Deck probing is added as part of the capacitive sensor testing
- Stall detection is handled during the Z current/speed test
- Gripper width error specs are updated based on our internal testing

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -736,6 +736,7 @@ async def calibrate_gripper_jaw(
         LOG.info(f"Gripper {probe.name} probe offset: {offset}")
         return offset
     finally:
+        await hcapi.retract(OT3Mount.GRIPPER)
         hcapi.remove_gripper_probe()
         await hcapi.ungrip()
```
ahiuchingau: I've added a change in this [PR](https://github.com/Opentrons/opentrons/pull/13165) that's merged into 0.14.0 
that removes the ungrip() call since it's really not necessary anymore.

If you adopt that change in this PR then we don't have to call retract twice (once here and another time in your test script)

ahiuchingau (changes requested):
You could remove the ungrip() function in `calibrate_gripper_jaw()` like this PR here now: https://github.com/Opentrons/opentrons/pull/13165

Let me know if it doesn't work for you

ahiuchingau (approved):
YAY

PR #13171, opened 2023-07-26
fix(app): only show run meter ticks for move labware command is manual with pause

b-cooper:
# Overview

Don't show moveLabware commands as ticks on the progress meter if their strategy parameter is anything but manualMoveWithPause

Closes RQA-1002

# Review requests

- upload a protocol with gripper move labware commands, confirm they don't appear as ticks on the run progress meter.

# Risk assessment
low


brenthagen (approved):
LGTM !

PR #13173, opened 2023-07-26
feat(robot-server): red light when estop is engaged

fsinapi:
# Overview

When the estop is engaged, the status bar should turn red. This PR makes that happen.

Before this PR, the light control task was only started once the Engine Store was initialized (which could be for a number of endpoints). This poses an issue if the estop is pressed before any of those endpoints are called. As a part of this PR, we start the light control task as soon as the hardware is initialized, and supply the engine store later. There are some nasty circular dependencies if you try to include the `runs` module from the `hardware` module so we don't do that and instead register a callback for when the hardware is done initializing. I don't like this but I think we'll need a pretty big refactor to make it feasible.

# Test Plan

On a Flex, tried pressing the estop and releasing it while the robot was sitting idle right after restarting the robot server. The light goes red & white.

During a run, if you press the Estop then the light turns solid red. Once you release the Estop it transitions into the hardware error mode (blinking red).

# Changelog

- Create light control task on startup
- Light control task can be created without an EngineStore and have it supplied later
- Light control task checks estop state
- Hardware API provides an accessor for estop state

# Review requests

As mentioned above, I couldn't put the Light Control Task initializer directly into `robot_server.hardware` because it needs to depend on `robot_server.runs.engine_store` and `robot_server.runs` ends up depending on `robot_server.hardware`. Even if you pull the engine store out of `runs`, the `robot_server.protocols` module ends up getting pulled in and it also depends on `robot_server.hardware`. If I'm missing an easy way to shoehorn this in without having a circular dependency then I'm all ears.

# Risk assessment

Medium level that this could affect status bar workflows but that should be it, tested on a Flex that it works as expected in basic situations.


api/src/opentrons/hardware_control/protocols/chassis_accessory_manager.py
```diff
@@ -60,3 +60,10 @@ def get_status_bar_state(self) -> StatusBarState:
 
         :returns: The current status bar state enumeration."""
         ...
+
+    def get_estop_state(self) -> EstopState:
+        """Get the current Estop state.
+        
+        If the Estop is not supported on this robot, this will always return Disengaged.
+
+        :returns: The current Estop state."""
```
sfoster1:
```suggestion
        :returns: The current Estop state."""
        ...
```
not strictly speaking necessary but we do it for the other methods

sfoster1 (approved):
Kinda gross but not sure what else to do really.

api/src/opentrons/hardware_control/protocols/chassis_accessory_manager.py
fsinapi: d4f6c8017e0957d5b151bebc8e422f4f8014bdcb

PR #13174, opened 2023-07-26
refactor(api): improve labware movement code

sanni-t:
# Overview

Last refactoring (ish) PR before I add the contextual gripper offsets. This makes updating the code much easier!

# Test Plan

- Test on a robot- even better, mimic labware movements from the science protocols on the robot to make sure everything still works correctly.

# Changelog

- compartmentalizes the labware movement code to make it easier to test
- the only functional change is removal of the gripper Z-home commands before each `move_to` to the gripper Z home position. See RLAB-215 for more info. With RLAB-211 fixed, there doesn't seem to be a need to issue a home command before moving to gripper Z each time

# Review requests

- Mainly just- does the refactor make sense? any other improvements I should make?
- I have tried to preserve tests as is so that I can catch any unintended changes. Caught one outdated value in the tests that we were just not using so that's good and is reassuring that tests work
- @ahiuchingau do you think removing any of the homes before the move to gripper Z home position cause any issues?

# Risk assessment

Low since tests


api/src/opentrons/motion_planning/types.py
```diff
@@ -29,3 +29,12 @@ class MoveType(Enum):
...
     DIRECT = auto_enum_value()
+
+
+@dataclass(frozen=True)
+@final
+class GripperMovementWaypointsWithJawStatus:
+    """Gripper motion waypoint with expected jaw status while moving to the waypoint."""
+
+    position: Point
+    jawOpen: bool
```
SyntaxColoring: No need for `camelCase` if this is just an internal thing.

ahiuchingau (approved):
Makes sense to me!

I think removing the home commands should be fine

SyntaxColoring (commented):
Code looks good to me. Discussed in-person: we'll QA the `home()` vs. `move_to()` thing.

sanni-t:
Tested moving labware between all modules and deck on a QA bot and it worked exactly the same as before the change.

PR #13177, opened 2023-07-27
fix(app): fetch all commands when downloading run log

brenthagen:
# Overview

when downloading run log, fetches the total length of commands initially in order to fetch all commands.

closes RAUT-532

# Test Plan

 - manually verified that protocols of greater than 30 commands download all commands

# Changelog

 - Fetches all commands when downloading run log

# Review requests

confirm extra getCommands approach
ideally, test with a long protocol

# Risk assessment

low


b-cooper (approved):
Code looks good, I still need to test on a robot

PR #13178, opened 2023-07-27
feat(api-client, app, react-api-client) hook up E-stop modals to desktop app and touchscreen app

koji:
# Overview
hook up estop modals to desktop app and touchscreen app

# Test Plan
Test needs a Flex since we don't have E-stop that is capable of connecting to a dev kit
[Desktop app]
pressed modal
1. Open the app and go to Device Details
2. Press E-stop button
Check: the E-stop pressed modal shows up
3. Twist E-stop button
Check: Resume operation button in the E-stop pressed modal is activated (This will take some time) fetching interval is 10sec
4. Click close icon
Check: Device Details page shows E-stop Banner and access to /Devices/:robotName/something (RobotSettings, Protocol Run page) and check those page don't show the E-stop pressed modal
5. Click `Resume operation` button in E-stop Banner
Check: E-stop pressed modal shows up
6. Click `Resume operation` button in E-stop pressed modal

missing modal
1. Open the app and go to Device Details
2. Detatch E-stop from a Flex
Check: the E-stop missing modal shows up
3. Click close icon
Check: Device Details page shows E-stop Banner and access to /Devices/:robotName/something (RobotSettings, Protocol Run page) and check those page don't show the E-stop pressed modal
5. Click text button in the Banner
Check: E-stop missing modal shows up
6. Connect E-stop button to the Flex
Check: E-stop missing modal is dismissed without any actions

[ODD app]
pressed modal
1. Press E-stop button
Check: the E-stop pressed modal shows up
2. Twist E-stop button
Check: Resume operation button in the E-stop pressed modal is activated (This will take some time) fetching interval is 10sec

missing modal
1. Detatch E-stop from a Flex
Check: E-stop missing modal shows up
2. Connect E-stop to the Flex
Check: E-stop missing modal is dismissed

# Changelog
- Add takeover component and react context component 
- Update E-stop modals
- Update each top component to hook up modals and banners to each app
- Update DeviceDetails component and ProtocolRunHeader component

# Review requests
Please see Test plan

# Risk assessment
low-mid


koji:
- [ ] update E-stop missing modal
- [ ] fix E-stop modals' rendering priority
- [ ] add tests

koji:
- [x] banner for running protocol page

app/src/App/DesktopApp.tsx
```diff
@@ -91,30 +97,70 @@ export const DesktopApp = (): JSX.Element => {
...
+          </Box>
+        </EmergencyStopContext.Provider>
       </ToasterOven>
     </>
   )
 }
+
+// interface EmergencyStopTakeoverProps {
+//   children: React.ReactNode
+// }
```
shlokamin: leftover comment

app/src/App/__tests__/DesktopApp.test.tsx
```diff
@@ -15,10 +15,13 @@ import { RobotSettings } from '../../pages/Devices/RobotSettings'
 import { GeneralSettings } from '../../pages/AppSettings/GeneralSettings'
 import { Alerts } from '../../organisms/Alerts'
 import { useSoftwareUpdatePoll } from '../hooks'
+// import { useRobot } from '../../organisms/Devices/hooks'
+// import { mockConnectedRobot } from '../../redux/discovery/__fixtures__'
```
shlokamin: leftover imports

app/src/App/__tests__/DesktopApp.test.tsx
```diff
@@ -77,6 +81,7 @@ describe('DesktopApp', () => {
     mockAlerts.mockReturnValue(<div>Mock Alerts</div>)
     mockAppSettings.mockReturnValue(<div>Mock AppSettings</div>)
     mockBreadcrumbs.mockReturnValue(<div>Mock Breadcrumbs</div>)
+    // mockUseRobot.mockReturnValue(mockConnectedRobot)
```
shlokamin: leftover comment

app/src/organisms/EmergencyStop/EstopMissingModal.tsx
```diff
@@ -24,28 +25,38 @@ import type { LegacyModalProps } from '../../molecules/LegacyModal'
 interface EstopMissingModalProps {
   robotName: string
   closeModal: () => void
+  isDismissedModal?: boolean
+  setIsDismissedModal?: (isDismissedModal: boolean) => void
```
shlokamin: how come these are both optional? it looks like they are always provided. if they are required, the logic in this component below can be simplified

app/src/organisms/EmergencyStop/EstopTakeover.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  React.useEffect(() => {
+    setIsDismissedModal(isDismissedModal)
+    if (!isDismissedModal) {
+      setShowEstopModal(true)
+    }
+  }, [isDismissedModal, setIsDismissedModal, showEstopModal])
+
```
shlokamin: this `useEffect` seems a bit dangerous because we are unconditionally setting state. how come we need to do `setIsDismissedModal(isDismissedModal)`  on line 62? `setIsDismissedModal` and `isDismissedModal` both come from the same provider, so this shouldn't be necessary right?

i'm also not following why this is necessary either: 

```js
   if (!isDismissedModal) {
      setShowEstopModal(true)
    }
```

why do we need two boolean variables to determine whether to show the modal? we have both `isDismissedModal` and `showEstopModal`. it seems like we should always show the an estop modal if the estop is not disengaged _unless_  `isDismissedModal` is true. not sure why we need the other boolean

app/src/organisms/EmergencyStop/EmergencyStopContext.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react'
+
+export interface EmergencyStopContextType {
+  isDismissedModal: boolean
+  setIsDismissedModal: (isDismissedModal: boolean) => void
```
shlokamin: can we name these variables something more explicit to indicate this is the stop modal and not just any modal? maybe `isEmergencyStopModalDismissed` and `setIsEmergencyStopModalDismissed` or something?

app/src/organisms/EmergencyStop/EstopTakeover.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [showEstopModal, setShowEstopModal] = React.useState<boolean>(false)
+  const { isDismissedModal, setIsDismissedModal } = useEstopContext()
+  const isUnboxingFlowOngoing = useIsUnboxingFlowOngoing()
+  const closeModal = (): void => {
+    if (estopStatus?.data.status === DISENGAGED) {
+      setShowEstopModal(false)
+    }
+  }
+
+  const targetEstopModal = (): JSX.Element | null => {
```
shlokamin: this is a fn that returns `JSX.Element | null`, why not just make it a react component?

```js
 const TargetEstopModal = (): JSX.Element | null => {}

// and then use it
<TargetEstopModal />
```

shlokamin (commented):
lookin good! left some comments. still not able to get the desktop app rendering the e stop modals but im looking into why

shlokamin:
ah! the reason i cant see the modal on desktop is because `useIsUnboxingFlowOngoing` is returning `true` because on desktop `unfinishedUnboxingFlowRoute` is ` '/welcome'`

koji:
> ah! the reason i cant see the modal on desktop is because `useIsUnboxingFlowOngoing` is returning `true` because on desktop `unfinishedUnboxingFlowRoute` is ` '/welcome'`

oh I see.

app/src/organisms/EmergencyStop/EstopTakeover.tsx
koji: `showEstopModal` to display the modal or not.
`isDismissedModal` to handle banner display and modal display.

shlokamin: > showEstopModal to display the modal or not.

hmm but we just forward along `isEmergencyStopModalDismissed` to `showEstopModal `
```js
if (!isEmergencyStopModalDismissed) {
      setShowEstopModal(true)
    }
```
so if `isEmergencyStopModalDismissed` changes, so does `showEstopModal` right?

koji: you are right, we could remove setShowEstopModal.
Tested the code and it worked as expected.

koji:
- [x] check the modal with Windows app

app/src/App/OnDeviceDisplayApp.tsx
```diff
@@ -214,6 +217,8 @@ export const OnDeviceDisplayApp = (): JSX.Element => {
   const isIdle = useIdle(sleepTime, options)
   const scrollRef = React.useRef(null)
   const isScrolling = useScrolling(scrollRef)
+  const localRobot = useSelector(getLocalRobot)
+  const robotName = localRobot?.name ?? 'no name'
```
shlokamin: if there is no `robotName` we shouldn't try to render the `EstopTakeover` at all since it needs it. 

maybe instead `EstopTakeover` doesnt need to have `robotName` passed in via props and instead it can use the `getLocalRobotSelector`

app/src/pages/Devices/DeviceDetails/index.tsx
```diff
@@ -37,31 +41,54 @@ export function DeviceDetails(): JSX.Element | null {
...
   ) : (
     <Redirect to="/devices" />
   )
 }
+
+interface RenderDeviceDetailsProps {
+  robotName: string
+}
+
+function RenderDeviceDetails({
```
shlokamin: i think we should either call this `DeviceDetailsComponent` or inline this new logic in the body of `DeviceDetails`

app/src/organisms/EmergencyStop/EstopPressedModal.tsx
```diff
@@ -104,22 +129,46 @@ function TouchscreenModal({
...
   }
 
-  const handleClick = (): void => {
-    // ToDo (kk:07/11/2023) this will be implemented by a following pr
-    console.log('resume robot operations')
+  const handleClick: React.MouseEventHandler<HTMLButtonElement> = (e): void => {
+    e.preventDefault()
+    e.stopPropagation()
+    setIsResuming(true)
+    setEstopPhysicalStatus(null, {
```
shlokamin: how come `setEstopPhysicalStatus` needs to take in `null` as an argument? looking at `useSetEstopPhysicalStatusMutation`: 

```js
 const mutation = useMutation<EstopStatus, AxiosError, null>(
    [host, 'robot/control/acknowledgeEstopDisengage'],
    (newStatus: null) =>
      setEstopPhysicalStatus(host as HostConfig, newStatus)
        .then(response => response.data)
        .catch(e => {
          throw e
        }),
    options
  )
```

`newStatus` is always `null`, so there doesnt seem to be a need to have there be an input to the function at all. 

drilling down further into `setEstopPhysicalStatus`:

```js
export function setEstopPhysicalStatus(
  config: HostConfig,
  data: null
): ResponsePromise<EstopStatus> {
  return request<EstopStatus, null>(
    PUT,
    '/robot/control/acknowledgeEstopDisengage',
    data,
    config
  )
}

```
this endpoint doesnt set the status to anything, it only sets the status to disengaged. maybe we can rename this to `acknowledgeEstopDisengage`?

app/src/App/__tests__/DesktopApp.test.tsx
```diff
@@ -54,6 +55,7 @@ const mockBreadcrumbs = Breadcrumbs as jest.MockedFunction<typeof Breadcrumbs>
 const mockUseSoftwareUpdatePoll = useSoftwareUpdatePoll as jest.MockedFunction<
   typeof useSoftwareUpdatePoll
 >
+// const mockUseRobot = useRobot as jest.MockedFunction<typeof useRobot>
```
shlokamin: leftover comment

shlokamin (changes requested):
left some comments but nothing code wise that blocks merging. 

what does block merging, unfortunately, is that we either need to get more estops or implement some kind of override which might not even be in compliance. so we gotta sit tight till this gets resolved: https://opentrons.slack.com/archives/C049WNGU22H/p1691508733660019

app/src/pages/Devices/DeviceDetails/index.tsx
koji: I will export this as a component since this needs to be wrapped by `ApiHostProvider` so inline might not work or make codes messy.

app/src/organisms/EmergencyStop/EstopPressedModal.tsx
```diff
@@ -104,22 +129,49 @@ function TouchscreenModal({
...
     childrenPadding: SPACING.spacing24,
     width: '47rem',
   }
 
-  const handleClick = (): void => {
-    // ToDo (kk:07/11/2023) this will be implemented by a following pr
-    console.log('resume robot operations')
+  const handleClick: React.MouseEventHandler<HTMLButtonElement> = (e): void => {
+    e.preventDefault()
+    e.stopPropagation()
```
shlokamin: do we need these?

shlokamin (approved):

app/src/organisms/EmergencyStop/EstopPressedModal.tsx
koji: removed

PR #13180, opened 2023-07-27
fix(api): Decrease 96ch Z holding current from 0.8 to 0.5 amps

andySigler:
# Overview

The Z motor(s) are getting pretty warm when the 96ch is not moving. Testing shows that we can safely reduce the holding current from `0.8` amps down to `0.5` amps without performance issues.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
sweet

PR #13181, opened 2023-07-27
feat(api): reduce unused mount currents with 96

sfoster1:
This will make the hold and run currents for the unused right mount be much lower when the 96 is attached.


Laura-Danielle (approved):
lgtm

PR #13182, opened 2023-07-27
fix(app): filter for duplicate defUri after filtering for tipracks usedx

smb2268:
Fix RQA-1119

# Overview
In LPC we were comparing a list of unique labware definitionUri's and locations present in a protocol with a list of labware that was visited by pipettes. If there were multiples of the same loaded labware, some of which were visited and some that wasn't used by a pipette, it was possible for labware to be missing in LPC. 

# Test Plan
Upload the protocol linked in the above ticket and ensure the tiprack that was missing is now present in LPC.

# Changelog
1. Move filtering for duplicate definitionUri/location combos from `appendLocationComboIfUniq` to later in the matching process in `getCheckSteps`
2. Update tests

# Review requests
See test plan

# Risk assessment
Low


brenthagen (approved):
verified fix with DSM Test Protocol ✔️

PR #13184, opened 2023-07-27
fix(app): show recent run skeleton if still loading hardware 

b-cooper:
# Overview

In order to avoid a flash of orange for run records that are actually ready to run, show the
skeleton view for a recent run item, if the modules and instruments are still loading initially.

# Changelog

- add `isLoading` to getMissingProtocolHardware and internals, show skeleton for recent run card if still loading
- fix Xpadding discrepancy for navigation across ODD pages.

# Risk assessment
low


koji (approved):
The changes look good to me.
Tested this branch on App&UIBot and worked as expected.
Also the navigation shrinking issue was solved, thanks!

For lint-js
need to format the file, `app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx`

PR #13185, opened 2023-07-27
fix(api): separate disengage and stop

ahiuchingau:
# Overview
When we issue a StopRequest to the Flex axes, the following happens on the firmware side:
1. the motor gets disengaged to stop any movement
2. any queued up motor tasks are cleared

This has worked pretty well for immediately halting robot motion. But, this also means sometimes we're unexpectedly disengaging the motors when not necessary and leaving the axes in a non-movable state. 

We can solve this by not disengaging the motors in the StopRequest task and let the DisableMotorRequest handles that. You'll need the changes in https://github.com/Opentrons/ot3-firmware/pull/710 in order for this to work properly.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -692,14 +692,25 @@ class OT3API(
 
         asyncio.run_coroutine_threadsafe(_chained_calls(), self._loop)
 
-    async def halt(self) -> None:
-        """Immediately stop motion."""
+    async def stop_motors(self) -> None:
+        """Immediately stop motors."""
         await self._backend.halt()
+
+    def stop_modules(self) -> None:
+        """Immediately stop modules."""
         asyncio.run_coroutine_threadsafe(self._execution_manager.cancel(), self._loop)
+
+    async def halt(self) -> None:
+        """Immediately disengage all present motors and clear motor and module tasks."""
+        await self.disengage_axes(
+            [ax for ax in Axis if self._backend.axis_is_present(ax)]
+        )
+        await self.stop_motors()
+        self.stop_modules()
 
     async def stop(self, home_after: bool = True) -> None:
         """Stop motion as soon as possible, reset, and optionally home."""
-        await self._backend.halt()
+        await self.stop_motors()
         self._log.info("Recovering from halt")
         await self.reset()
 


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -692,14 +692,25 @@ async def _chained_calls() -> None:
 
         asyncio.run_coroutine_threadsafe(_chained_calls(), self._loop)
 
-    async def halt(self) -> None:
-        """Immediately stop motion."""
+    async def stop_motors(self) -> None:
+        """Immediately stop motors."""
         await self._backend.halt()
+
+    def stop_modules(self) -> None:
```
Laura-Danielle, line 699: What is this guy meant to do?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -692,14 +692,25 @@ async def _chained_calls() -> None:
...
+        """Immediately stop modules."""
         asyncio.run_coroutine_threadsafe(self._execution_manager.cancel(), self._loop)
 
+    async def halt(self) -> None:
+        """Immediately disengage all present motors and clear motor and module tasks."""
+        await self.disengage_axes(
+            [ax for ax in Axis if self._backend.axis_is_present(ax)]
+        )
+        await self.stop_motors()
```
Laura-Danielle, line 708: Should the order be
```
await self.stop_motors()
await self.disengage_axes()
```
?

Laura-Danielle (commented):
Leaving as a comment. Don't want my q's to be blocking.

api/src/opentrons/hardware_control/ot3api.py
ahiuchingau: The disengage request is used to immediately stop the motor while it's still in motion. This allows the halt request to behave similarly to an estop.

api/src/opentrons/hardware_control/ot3api.py
ahiuchingau: This clears any future module tasks scheduled by the protocol

Laura-Danielle: Ah yes I see it now. The diff on github was being weird for me and I thought it was a free floating func

Laura-Danielle (approved):
lgtm

PR #13186, opened 2023-07-27
feat(api, shared-data): add location-based gripper offsets to final labware movement offsets

sanni-t:
Closes RLAB-295 

# Overview

Takes the location-based offsets we have been using in protocols and puts them in definitions so that they are included by default while calculating total gripper offsets.

# Test Plan

- Test the well-tested science protocols that do labware movement with gripper and verify that they work the same without any offsets provided in the protocol.

# Changelog

- adds a new field of `gripperOffsets` to deck, labware & module schemas
- updates all the relevant definition models
- adds offset getters to engine module & labware state views
- updates the final gripper offset calculations to include these offsets

# Review requests

- make sure the calculations are being done correctly
- does the shape of the new fields look okay?
- Updated definition models look fine or can they be improved?

# Risk assessment

Medium. The offset calculations are unit tested & will be robot tested well so that's not a risk. But we should make sure the actual offsets being added are correct.


shared-data/deck/schemas/3.json
```diff
@@ -250,7 +279,32 @@
...
+              "$ref": "#/definitions/xyzArray",
+              "description": "Offset added to calculate drop coordinates of a labware placed on this deck."
+            }
+          },
+          "required": [
+            "pickUpOffset",
+            "dropOffset"
+          ]
+        }
+      }
```
sanni-t: add default as required

shared-data/python/opentrons_shared_data/module/dev_types.py
```diff
@@ -94,7 +105,9 @@
         "slotTransforms": Dict[str, Dict[str, Dict[str, List[List[float]]]]],
         "compatibleWith": List[ModuleModel],
         "twoDimensionalRendering": Dict[str, Any],
+        "gripperOffsets": Dict[str, GripperOffsets]
     },
+    total=False,
```
sanni-t, line 111: Change this assigned TypedDict to class-based and see if anything breaks. If not, switch to that and split it into required vs optional typed dicts

sanni-t: punting this to later

shared-data/labware/definitions/2/opentrons_universal_flat_adapter/1.json
sanni-t, line 1: Tagging @andySigler @ChrisYarka for verification

shared-data/module/definitions/3/heaterShakerModuleV1.json
sanni-t, line 1: Tagging @andySigler @ChrisYarka for verification

shared-data/module/definitions/3/magneticBlockV1.json
sanni-t, line 1: Tagging @andySigler @ChrisYarka for verification

shared-data/module/definitions/3/thermocyclerModuleV2.json
sanni-t, line 1: Tagging @andySigler @ChrisYarka for verification

b-cooper (approved):
Changes look good to me 👍

SyntaxColoring (commented):
Here are my results from testing the PR as-is, and from testing it with [adjusted offsets](https://github.com/Opentrons/opentrons/commit/163b1bd47786c554c1e773a8187a6c510f1b6702) that @andySigler gave me in [this spreadsheet](https://docs.google.com/spreadsheets/d/1-BXwaqTIB8GiaptIZ0nNZhSW5iGPk5jrO5O9SnV3-fY/edit?usp=sharing).

| Pickup/drop location                                                                            | Sanniti's original offsets                                                                         | Andy's offsets |
|-------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------|----------------|
| `armadillo_96_wellplate_200ul_pcr_full_skirt` on bare deck slot                                   | ✅                                                                                                  | ✅              |
| `armadillo_96_wellplate_200ul_pcr_full_skirt` on `thermocyclerModuleV2`                             | ❌ In the jaw-open immediately after the drop, the jaws collide with the TC. Pickup not attempted.  | ❌ Same.        |
| `armadillo_96_wellplate_200ul_pcr_full_skirt` on `temperatureModuleV2`                              | ❌ During both pickup and drop, it collides with the ~2mm lip around the edge of the module's slot. | ❌ Same.        |
| `armadillo_96_wellplate_200ul_pcr_full_skirt` on `magneticBlockV1`                                  | 🤷‍♂️ It seems to think the labware sits ~2 mm higher than it actually does.                          | 🤷‍♂️ Same.       |
| `corning_384_wellplate_112ul_flat` on `opentrons_universal_flat_adapter` on `heaterShakerModuleV1`    | ✅                                                                                                  | ✅              |
| `armadillo_96_wellplate_200ul_pcr_full_skirt` on `opentrons_96_pcr_adapter` on `heaterShakerModuleV1` | ✅                                                                                                  | ✅              |

[Here are the Python files that I used.](https://gist.github.com/SyntaxColoring/e44637224ca89884c0b94c0c27121385)

SyntaxColoring:
Recap from in-person work with @sanni-t, @andySigler, and @jbleon95:

* `thermocyclerModuleV2`: A pickup *and drop* offset of z=4.6 avoids the collision, and happens to align with the official value for how far up the Thermocycler pops the labware. We've added that to this PR.
* `magneticBlockV1`: This problem will remain unresolved for this PR.
* `temperatureModuleV2`: This problem will remain unresolved for this PR. This might be lower-priority because realistic protocols would load the PCR plate atop an *adapter* atop the module, which would happen to avoid this problem. We couldn't test with that setup because that adapter hasn't yet been split into its own definition.

Thinking continues on where we should take this work in the medium/long-term.

* If you pick up closer to the top of a labware, theoretically, you should also drop it higher to compensate.
* Defining these offsets like we are now is probably not appropriate for labware other than `armadillo_96_wellplate_200ul_pcr_full_skirt`, with different heights.

PR #13187, opened 2023-07-27
fix(api): update estimation before homing if encoder pos is good, regardless of stepper status

ahiuchingau:
# Overview
When stall detection is disabled, homing an axis might be an impossible task because the hardware controller is tricked into thinking that the motor position is valid. If the motor is actually much further away from the limit switch than the value the firmware reports, the home command will time out and the move will fail. If the motor is much closer to the limit switch, the home command would most definitely cause the axis to crash into the limit switch. 

Since there are safe guards already on the firmware side (we _only_ update the stepper position if the encoder position is good and the motor is not current moving and/or does not have any queued up moves), we should be able to just update the position based on the encoder value whenever we are about to **home** because the stepper & encoder positions will be subsequently reset anyway.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1209,17 +1209,12 @@ class OT3API(
         # G, Q should be handled in the backend through `self._home()`
         assert axis not in [Axis.G, Axis.Q]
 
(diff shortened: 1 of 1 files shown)


Laura-Danielle (approved):
lgtm

sfoster1 (approved):
Makes sense to me!

PR #13189, opened 2023-07-28
fix(api): enable deck conflict checker for the flex

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/RSS-294 -> enable `deck_conflict` checker for the flex..
closes https://opentrons.atlassian.net/browse/RSS-273 -> fixes bug when loading a thermocycler in slots that are not allowed. 
closes https://opentrons.atlassian.net/browse/RQA-993 -> does not allow loading mag deck's on an OT-3. just mag blocks. 

# Test Plan

- Allow an OT-3 protocol with modules as described [here](https://opentrons.atlassian.net/browse/RSS-294). make sure no exception is being raise.  
```

metadata = {
    "apiLevel": "2.15",
}

requirements = {
    "robotType": "OT-3"
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    tip_rack_50 = ctx.load_labware("opentrons_ot3_96_tiprack_50ul", 4)
    tip_rack_200 = ctx.load_labware("opentrons_flex_96_tiprack_200ul", 5)
    tip_rack_1000 = ctx.load_labware("opentrons_ot3_96_tiprack_1000ul", 6)
    heatershaker = ctx.load_module('heaterShakerModuleV1',1)
    heatershaker = ctx.load_module('thermocyclerModuleV2')
    mag_block = ctx.load_module('magneticBlockV1',2)
    temp = ctx.load_module('temperatureModuleV2', 3)
```

- Try loading a module on the flex in a middle slot. make sure you are getting an exception. 
```

metadata = {
    "apiLevel": "2.15",
}

requirements = {
    "robotType": "OT-3"
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    # this should fail bc h/s is not allowed in the middle.
    mag_block = ctx.load_module('heaterShakerModuleV1',2)
```
- Try loading 2 modules on the flex in same slot slot. make sure you are getting an exception. 
```

metadata = {
    "apiLevel": "2.15",
}

requirements = {
    "robotType": "OT-3"
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    tip_rack_50 = ctx.load_labware("opentrons_ot3_96_tiprack_50ul", 4)
    tip_rack_200 = ctx.load_labware("opentrons_flex_96_tiprack_200ul", 5)
    tip_rack_1000 = ctx.load_labware("opentrons_ot3_96_tiprack_1000ul", 2)
    heatershaker = ctx.load_module('heaterShakerModuleV1',1)
    # this should fail bc h/s is already loaded in slot 1
    mag_block = ctx.load_module('magneticBlockV1',1)
```

- Should fail bc mag decks are not allowed on the flex.
```

metadata = {
    "apiLevel": "2.15",
}

requirements = {
    "robotType": "OT-3"
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    # this should fail bc h/s is not allowed in the middle.
    mag_deck = ctx.load_module('magneticModuleV2',1)
```

- Should fail bc thermocyclers are not allowed in slot 1.
```

metadata = {
    "apiLevel": "2.15",
}

requirements = {
    "robotType": "OT-3"
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    # this should fail bc h/s is not allowed in the middle.
    thermo = ctx.load_module('thermocyclerModuleV2',1)
```

- Smoke test OT-2 protocols and make sure it all works as expected. 
- I have also created an integration test for that
- We should test this with the app and a mag block in a protocol to make sure that it appears properly in the app

# Changelog

- Change `deck_conflict.check` to get DeckSlotName for the new location and and existing location instead of int. 
- Added `robot_type` arg inorder to know what restrictions to load. 
- Changed all access with method to DeckSlotName instead of int.
- Added integration test for a failed protocol
- Updated all tests to use DeckSlotName instead of int/str
- Make sure you are able to load the thermocycler in allowed slots. 
- Added slot validation for load_module in protocol core 

# Review requests

Changes make sense? Did I forget anything? 

# Risk assessment

Medium. need to smoke test OT-2 protocols both json on python for all version (We care about) to make sure loading labware/modules on deck has been affected. need to test OT-3 protocols and to make sure `load_module` is raising an exception for the restrictions in the ticket.


api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -184,10 +191,13 @@ def check(
         DeckConflictError: Adding this item should not be allowed.
     """
     restrictions: List[_DeckRestriction] = [_FixedTrashOnly()]
-
+    if robot_type == "OT-2 Standard":
```
TamarZanzouri: hate this

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -272,6 +284,59 @@ def _create_restrictions(item: DeckItem, location: int) -> List[_DeckRestriction
...
+        for covered_location in _flex_slots_covered_by_thermocycler():
+            restrictions.append(
+                _NothingAllowed(
+                    location=covered_location,
+                    source_item=item,
+                    source_location=location,
+                )
+            )
+
+    elif isinstance(item, (HeaterShakerModule, TemperatureModule)):
```
Laura-Danielle: Are these restrictions still valid? The spacing between slots is very different on the flex.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -169,9 +175,10 @@ class DeckConflictError(ValueError):
 # things that don't fit into a single deck slot, like the Thermocycler.
 # Refactor this interface to take a more symbolic location.
 def check(
-    existing_items: Mapping[int, DeckItem],
+    existing_items: Mapping[Union[int, str], DeckItem],
     new_item: DeckItem,
-    new_location: int,
+    new_location: Union[int, str],
+    robot_type: str = "OT-2 Standard",
```
Laura-Danielle: I think this is totally reasonable to do, but if you really hate it you could create a enum and/or bitfield in which you can specify what items you want to check for collisions and push up the robot type check higher.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -272,6 +284,59 @@ def _create_restrictions(item: DeckItem, location: int) -> List[_DeckRestriction
     return restrictions
 
 
+def _create_flex_restrictions(item: DeckItem, location: str) -> List[_DeckRestriction]:
+    restrictions: List[_DeckRestriction] = []
+
+    if isinstance(item, ThermocyclerModule):
+        for covered_location in _flex_slots_covered_by_thermocycler():
```
Laura-Danielle, line 305: I thought the TC could be in multiple slots, is this not the case?

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -272,6 +284,59 @@ def _create_restrictions(item: DeckItem, location: int) -> List[_DeckRestriction
...
+
+def _create_restrictions(
+    item: DeckItem, location: Union[int, str], robot_type: str
+) -> List[_DeckRestriction]:
+
+    if robot_type == "OT-2 Standard":
+        try:
+            new_location = int(location)
+        except ValueError:
+            raise DeckConflictError("OT-2 deck location is expected to be a number")
```
Laura-Danielle: Is this true? I thought Max had put in a mapping between the two

Laura-Danielle (commented):
Initial comments, but overall looks pretty good to me. Agree w/ Max that you should not be using Unions here though.

api/src/opentrons/motion_planning/deck_conflict.py
TamarZanzouri: Confirmed with Chris that this is still mandatory. also PD still has restrictions.

api/src/opentrons/motion_planning/deck_conflict.py
TamarZanzouri: its no slot 7,10. I am setting both slots as occupied if we are in the TC case

api/src/opentrons/motion_planning/deck_conflict.py
TamarZanzouri: removed this. I will use DeckSlotName instead

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -314,14 +314,21 @@ def load_module(
...
+        module_type = ModuleType.from_model(model)
         # TODO(mc, 2022-10-20): move to public ProtocolContext
         # once `Deck` and `ProtocolEngine` play nicely together
         if deck_slot is None:
-            if ModuleType.from_model(model) == ModuleType.THERMOCYCLER:
-                deck_slot = DeckSlotName.SLOT_7
+            if module_type == ModuleType.THERMOCYCLER:
+                deck_slot = (
+                    DeckSlotName.SLOT_7
+                    if self._engine_client.state.config.robot_type == "OT-2 Standard"
```
TamarZanzouri: can also use deck robot property but this feels the same. comparing a string.

shared-data/deck/schemas/3.json
```diff
@@ -167,7 +167,8 @@
                     "magneticModuleType",
                     "temperatureModuleType",
                     "thermocyclerModuleType",
-                    "heaterShakerModuleType"
+                    "heaterShakerModuleType",
+                    "magneticBlockType"
```
TamarZanzouri, line 171: what do we think about this?

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -249,19 +256,19 @@ def _create_restrictions(item: DeckItem, location: int) -> List[_DeckRestriction
...
                 _NoModule(
-                    location=covered_location,
+                    location=DeckSlotName.from_primitive(hs_covered_location),
                     source_item=item,
                     source_location=location,
                 )
             )
 
-        for covered_location in get_east_west_slots(location):
+        for hs_covered_location in get_east_west_slots(location.as_int()):
```
Laura-Danielle, line 276: Should these functions also take in `DeckSlotNames` and perform the check on that type?

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -314,14 +314,21 @@ def load_module(
...
         assert configuration is None, "Module `configuration` is deprecated"
 
+        module_type = ModuleType.from_model(model)
         # TODO(mc, 2022-10-20): move to public ProtocolContext
         # once `Deck` and `ProtocolEngine` play nicely together
         if deck_slot is None:
-            if ModuleType.from_model(model) == ModuleType.THERMOCYCLER:
-                deck_slot = DeckSlotName.SLOT_7
+            if module_type == ModuleType.THERMOCYCLER:
+                deck_slot = (
```
Laura-Danielle: @SyntaxColoring can answer this better, but can we not just return the "new" deck slot name and let the lower level functions handle the conversion if needed?

Otherwise, could we make this a small helper func somewhere? Kinda ugly to look at.

shared-data/deck/schemas/3.json
Laura-Danielle: I'm OK with it. We treat it as a module essentially anyways.

Laura-Danielle (approved):
Looks great. Have some small comments that you can ignore.

api/src/opentrons/motion_planning/deck_conflict.py
TamarZanzouri: I tired doing it but its tied to much to the legacy implementation and we are already parsing this as int in other places so might as well leave it

api/src/opentrons/protocol_api/core/engine/protocol.py
TamarZanzouri: I can use to_ot3_equivalent/to_ot2_equivalent but we still need to check the robot type

robot-server/tests/integration/protocols/deck_coordinate_load_fail.py
```diff
@@ -0,0 +1,23 @@
...
+    thermocycler = context.load_module("thermocyclerModuleV2")
+    assert thermocycler.parent == "B1"
+
+    module = context.load_module("temperatureModuleV2", "A1")
+    assert module.parent == "A1"
+
+    pipette.move_to(labware.wells()[0].top())
+    module_labware = module.load_labware("armadillo_96_wellplate_200ul_pcr_full_skirt")
+
+    pipette.move_to(module_labware.wells()[0].top())
```
SyntaxColoring: Since we're expecting this to fail on this `context.load_module()`, can we trim out the parts of this file that won't run?

```suggestion
    # This should fail because the Thermocycler is in the way.
    module = context.load_module("temperatureModuleV2", "A1")
```

robot-server/tests/integration/http_api/protocols/test_deck_coordinate_load_fail.tavern.yaml
```diff
@@ -0,0 +1,51 @@
...
+            - name: deck_coordinate_load_fail.py
+              role: main
+          createdAt: !re_search "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+\\+\\d{2}:\\d{2}$"
+          robotType: OT-3 Standard
+          metadata:
+            protocolName: Deck Coordinate PAPIv2 Test
+          analyses: []
+          analysisSummaries:
+            - id: !anystr
+              status: pending
```
SyntaxColoring: I think we could omit this. This integration test doesn't need to test the `POST /protocols` response because other tests are already doing that. We really just care about the analysis result.

```suggestion
```

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -169,9 +170,10 @@ class DeckConflictError(ValueError):
 # things that don't fit into a single deck slot, like the Thermocycler.
 # Refactor this interface to take a more symbolic location.
 def check(
-    existing_items: Mapping[int, DeckItem],
+    existing_items: Mapping[DeckSlotName, DeckItem],
     new_item: DeckItem,
-    new_location: int,
+    new_location: DeckSlotName,
+    robot_type: str = "OT-2 Standard",
```
SyntaxColoring: This is a little nitpicky, but if possible, I would prefer not to default `robot_type` to `"OT-2 Standard"`. Instead, I'd require the caller to specify it.

My thinking is that in general, there is no safe default robot type. The caller always needs to think about the right value to provide in their context.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -184,10 +186,11 @@ def check(
         DeckConflictError: Adding this item should not be allowed.
     """
     restrictions: List[_DeckRestriction] = [_FixedTrashOnly()]
```
SyntaxColoring: I think this will only work for OT-2s because of the way `_FixedTrashOnly` hard-codes `_FIXED_TRASH_SLOT`, which is the OT-2-specific `DeckSlotName.FIXED_TRASH` and does not include the Flex-specific `DeckSlotName.SLOT_A3`. Is that intentional?

api/src/opentrons/protocol_api/core/engine/protocol.py
SyntaxColoring: I would write this something like this:

```python
if deck_slot is None:
    module_type = ModuleType.from_model(model)
    if module_type == ModuleType.THERMOCYCLER:
        robot_type = self._engine_client.state.config.robot_type
        deck_slot = DeckSlotName.SLOT_7.to_equivalent_for_robot_type(robot_type)
```

To answer @Laura-Danielle's question, I think: The lower-level Protocol Engine `loadModule` command *will* convert the slot name for us, but separately from that, this class is doing its own inspection of the deck definition to make sure that the location is "compatible." So this class still needs to do match the right kind of slot name to the deck definition.

SyntaxColoring (commented):
This looks great! One question about the fixed trash, and a few other minor suggestions.

api/src/opentrons/motion_planning/deck_conflict.py
TamarZanzouri: it was intentional when we thought of the movable trash. should probably incorporate it now that it is fixed. good catch

api/src/opentrons/protocol_api/core/engine/protocol.py
TamarZanzouri: also, when checking against the deck definition we are using robot specific deck slots and it is done in this level.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -183,11 +188,18 @@ def check(
     Raises:
         DeckConflictError: Adding this item should not be allowed.
     """
-    restrictions: List[_DeckRestriction] = [_FixedTrashOnly()]
-
+    restrictions: List[_DeckRestriction] = [
+        _FixedTrashOnly(
+            location=DeckSlotName.FIXED_TRASH.to_equivalent_for_robot_type(
+                robot_type  # type: ignore[arg-type]
```
SyntaxColoring: Oh, let's make the `robot_type` argument an `opentrons_shared_data.robot.dev_types.RobotType` instead of a `str` to fix this `# type: ignore`.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -176,23 +176,22 @@ def check(
...
-    robot_type: str,
+    robot_type: RobotType,
 ) -> None:
     """Check a deck layout for conflicts.
 
     Args:
         existing_items: Existing items on the deck, assumed to be valid.
         new_item: New item to add to the deck.
         new_location: Location where the new item will be added.
+        robot_type: Robot deck definition.
```
SyntaxColoring: This is a nitpick and I think we should merge this now anyway, but for future reference, the `RobotType` is not exactly the deck definition.

The OT-2 has one `RobotType` (`"OT-2 Standard"`), but two possible deck definitions (`ot2_standard` and `ot2_short_trash`). The names are very confusing.

SyntaxColoring (approved):
Code looks good to me. Thanks!

TamarZanzouri:
Tested ot OT-3:
loading modules where they are allowed
loading modules where they are not allowed and getting the proper error.
loading mag decks will raise an error.

Smoke tested OT-2 as well. Works as expected!

SyntaxColoring (approved):
Ah, yep. Fix makes sense to me. Nice catch.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -302,11 +354,22 @@ def _create_deck_conflict_error_message(
     return message
 
 
-def _slots_covered_by_thermocycler(thermocycler: ThermocyclerModule) -> Set[int]:
+def _slots_covered_by_thermocycler(
```
sanni-t: Just a nitpick- this would be less ambiguous if called as `_ot2_slots_covered_by_thermocycler`

sanni-t (approved):
LGTM! 🙌

PR #13190, opened 2023-07-28
fix(shared-data): update flex tip rack definitions with DVT/PVT measurements

jbleon95:
# Overview

While working on RLAB-358 on the 96 channel tip rack adapter split, it was discovered that the existing flex tip rack adapters had slight discerpancies in the geometry of their definition when compared to the schematics in the DVT Robot Extents spreadsheet. In order from most to least important:

- the `y` property of the wells was set to to `11.1` offset from the well edge instead of `11.38`. The spacing was proper so this did not compound but was still off by `.28`.
- the x and y dimensions of the tip racks themselves were set to `127.76` and `85.48` instead of `127.75` and `85.75`
- the well depth of each tip rack was set to its tip length, with the `z` property different for each so that the two added together to `99`, the z height of the tip rack. In practice this did not cause any issue since the pipette only moves to the tops of the wells for tip racks, and as long as these added to the correct height there was no noticeable change in behavior. In an effort to be more accurate though, these values were changed to the accurate uniform well `depth` of `97.5` and well `z` of `1.5`.

With these changes there should be a small reduction in the y-dimension LPC offsets for the tip racks.

# Test Plan

LPC using the same robot and pipette and tip rack location was used with both the old and the newly updated definitions and confirmed to have less offset needed for the new definitions. Overall correctness in tip pick up behavior for single and multi channel pipettes with all 3 tip lengths was tested using this protocol.

```
metadata = {
    'protocolName': 'Tip Rack Definition Change Test',
}

requirements = {
    "robotType": "OT-3",
    "apiLevel": "2.15"
}

def run(context):
    pipette_single = context.load_instrument("p50_single_flex", mount="left")
    pipette_multi = context.load_instrument("p1000_multi_flex", mount="right")

    tip_rack_50 = context.load_labware('opentrons_flex_96_tiprack_50ul', location="C1")
    tip_rack_200 = context.load_labware('opentrons_flex_96_tiprack_200ul', location="C2")
    tip_rack_1000 = context.load_labware('opentrons_flex_96_tiprack_1000ul', location="C3")

    labware = context.load_labware('armadillo_96_wellplate_200ul_pcr_full_skirt', location="D2")
    labware_well = labware.wells()[0].top()

    # Test picking up and dropping off in column 1
    for well in tip_rack_50.columns_by_name()['1']:
        pipette_single.pick_up_tip(well)
        pipette_single.move_to(labware_well)
        pipette_single.drop_tip(well)


    # Test picking up and dropping off in row A
    for well in tip_rack_50.rows_by_name()['A']:
        pipette_single.pick_up_tip(well)
        pipette_single.move_to(labware_well)
        pipette_single.drop_tip(well)


    # Test picking up four corners
    wells_by_name_200 = tip_rack_200.wells_by_name()

    for well_name in ['A1', 'H1', 'A12', 'H12']:
        well = wells_by_name_200[well_name]
        pipette_single.pick_up_tip(well)
        pipette_single.move_to(labware_well)
        pipette_single.drop_tip(well)


    # Test multi pipette pick up
    wells_by_name_1000 = tip_rack_1000.wells_by_name()
    for well_name in ['A1', 'A6', 'A7', 'A12']:
        well = wells_by_name_1000[well_name]
        pipette_multi.pick_up_tip(well)
        pipette_multi.move_to(labware_well)
        pipette_multi.drop_tip(well)
```

# Changelog

- Updated the definitions for the six tip flex tip racks (50uL, 200uL, and 1000uL for filter and non-filter tips)

# Review requests

Double check that the measurements compare properly to the DVT/PVT schematics. Also, in keeping with the amount of precision in other definitions, the decimal points of `.375` were rounded up to `.38`.

# Risk assessment

Medium-low, the definitions have been changed but by small amounts. There will still be a need for LPC but this PR should be reducing that amount on a properly calibrated robot.

Importantly:
**NEW LPC OFFSETS WILL NEED TO BE CALCULATED. USING OLD OFFSETS MIGHT CAUSE THE PIPETTE TO BE MISALIGNED AND CRASH INTO THE TIPS!**

### DIFF
--- a/protocol-designer/fixtures/protocol/7/doItAllV7.json
+++ b/protocol-designer/fixtures/protocol/7/doItAllV7.json
@@ -370,874 +370,874 @@
         "tags": []
       },
       "dimensions": {
-        "xDimension": 127.76,
-        "yDimension": 85.48,
+        "xDimension": 127.75,
+        "yDimension": 85.75,
         "zDimension": 99
       },
       "wells": {
         "A1": {
-          "depth": 57.9,
-          "totalLiquidVolume": 50,
+          "depth": 97.5,
           "shape": "circular",
           "diameter": 5.58,
+          "totalLiquidVolume": 50,
           "x": 14.38,
-          "y": 74.1,
-          "z": 41.1
+          "y": 74.38,
+          "z": 1.5
         },
         "B1": {
-          "depth": 57.9,
-          "totalLiquidVolume": 50,
+          "depth": 97.5,
           "shape": "circular",
           "diameter": 5.58,
+          "totalLiquidVolume": 50,
           "x": 14.38,
-          "y": 65.1,
-          "z": 41.1
+          "y": 65.38,
+          "z": 1.5
         },
         "C1": {
-          "depth": 57.9,
-          "totalLiquidVolume": 50,
+          "depth": 97.5,
           "shape": "circular",
           "diameter": 5.58,
+          "totalLiquidVolume": 50,
           "x": 14.38,
-          "y": 56.1,
-          "z": 41.1
+          "y": 56.38,
+          "z": 1.5
         },
         "D1": {
-          "depth": 57.9,
-          "totalLiquidVolume": 50,
+          "depth": 97.5,
           "shape": "circular",
           "diameter": 5.58,
+          "totalLiquidVolume": 50,
           "x": 14.38,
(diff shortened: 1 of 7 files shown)


sfoster1 (approved):
Looks good to me! Great catch!

PR #9715, opened 2022-03-18
refactor(protocol-engine): move command-specific errors into command model

mcous:
## Overview

While working on #9617, I realized that we had no good way to expose stateless command errors to the API user.

After chatting this through with @SyntaxColoring, this PR moves **command-specifc** commands into the command resources themselves, replacing the `errorId` field of a command with an `error` field, instead.

The top-level `errors` field of the ProtocolEngine state / run still exists, and still represents fatal errors for the run itself.

Unblocks #9617, and as a happy side effect, fixes #9436

## Changelog

- Do not place command-specific errors into `run.errors`
    - This does not affect run failure, since the errors still bubble up to the `ProtocolRunner`
    - The `ProtocolRunner` is then in charge of placing the `ExceptionInProtocol` error into run state
- Place command-specific errors into `run.commands[].error`

## Review requests

- [x] Smoke test a protocol run that fails due to command failure
    - e.g. two pickup tips in a row
- [x] Smoke test a protocol run that fails due to programmer error
    - e.g. `IndexError`
- [x] Client-side code does not need changes
    - I could not find any references to the old `errorId` field in client code, so I think we're ok here

## Risk assessment

Low, provided we are able to confirm that this doesn't mess with client expectations

## Response examples

```py
from opentrons import types

metadata = {
    "protocolName": "Tipception",
    "apiLevel": "2.12",
}


def run(ctx):
    tr = ctx.load_labware("opentrons_96_tiprack_300ul", 1)
    right = ctx.load_instrument("p300_single", types.Mount.RIGHT, [tr])

    right.pick_up_tip()
    right.pick_up_tip()
```

### Before (`edge`)

```js
// GET /runs/:run_id
{
    "data": {
        "id": "fae0ab16-9d83-4817-a52e-617d0d85188f",
        "createdAt": "2022-03-18T18:52:37.038251+00:00",
        "status": "failed",
        "current": true,
        "errors": [
            {
                "id": "fc530a13-bf1e-4786-9602-cf0c5149c168",
                "errorType": "LegacyContextCommandError",
                "createdAt": "2022-03-18T18:52:39.112172+00:00",
                "detail": "Cannot pick up tip with a tip attached"
            },
            {
                "id": "724d19d8-a668-4a8a-98e3-277c621cccbb",
                "errorType": "ExceptionInProtocolError",
                "createdAt": "2022-03-18T18:52:39.113973+00:00",
                "detail": "TipAttachedError [line 14]: Cannot pick up tip with a tip attached"
            }
[...]

Issue #9436, opened 2022-02-11
bug(engine): errors during PAPIv2 commands are duplicated in the /runs response


mcous:
## Overview

When a Python Protocol API v2 protocol encounters an error during a command handling, it ends up in the ProtocolEngine's (and therefore, the HTTP `/runs` response's) errors list twice, once for the command, and once again for the `raise` from the call to `run(ctx)`.

## Steps to reproduce

Run a PAPIv2 protocol that will fail trivially, such as two pick up tips in a row with the same pipette, on a v5.x robot or emulator. Then fetch `GET /runs/:run_id` and observe the two nearly identical errors.

## Expected behavior

The engine state only tracks the single, root error. Ideally, we figure out how to preserve important information like the stacktrace, that is only present in one of the two current "duplicate" errors.

PR #13191, opened 2023-07-28
docs(api,robot-server): Document errors list as having no more than 1 element

SyntaxColoring:
# Overview

In the public (HTTP-facing) models for runs and analyses, you can access the errors of individual commands. In addition, there's an array of top-level errors. That top-level array is used to indicate fatal errors, including those that happen in between commands.

This documents, officially, that the top-level array will only ever have zero elements or one element. It will never contain multiple errors. The fact that it's an array, instead of a nullable object, is just a historical quirk.

This goes towards RSS-146 by clarifying how we should expose different kinds of run failures.

# Rationale

* There is no use for `errors` to have multiple elements today.

  This array is left over from a time before individual commands held their own errors. Today, commands have a self-contained `error` field, but in þe olden tymes, they had an `errorId` reference that pointed into this array. This was problematic for a few reasons, such as [inherent duplication](https://github.com/Opentrons/opentrons/issues/9436).

  [PR #9715](https://github.com/Opentrons/opentrons/pull/9715) fixed this. Since then, `errors` has only had up to 1 element.

* Having multiple elements is confusing, and muddies what the field is supposed to be doing, conceptually. One internal comment describes it as "a list of fatal errors." But there can only be one fatal error, by definition.

* For the case where multiple things contributed to a run's failure, and we want to convey them all, we now have @sfoster1's `EnumeratedError` architecture. We should commit to that tree-based API instead of keeping around a competing array-based API.

* According to @shlokamin, the app just does, and has always just done, something like `errors[0]`.

* If we really need to have multiple elements in this list later on, we can just revert this documentation change. We can do that revert without breaking any clients.

# Test Plan

None needed.

# Changelog

Update the public HTTP documentation in run and analysis models, and internal comments.

# Review requests

Do we agree with this?

# Risk assessment

No risk.

### DIFF
--- a/robot-server/robot_server/maintenance_runs/maintenance_run_models.py
+++ b/robot-server/robot_server/maintenance_runs/maintenance_run_models.py
@@ -76,7 +76,11 @@ class MaintenanceRun(ResourceModel):
     )
     errors: List[ErrorOccurrence] = Field(
         ...,
-        description="Any errors that have occurred during the run.",
+        description=(
+            "The protocol's fatal error, if there was one."
+            " For historical reasons, this is an array,"
+            " but it will only have up to one element."
+        ),
     )
     pipettes: List[LoadedPipette] = Field(
         ...,
--- a/robot-server/robot_server/maintenance_runs/maintenance_run_models.py
+++ b/robot-server/robot_server/maintenance_runs/maintenance_run_models.py
@@ -77,7 +77,7 @@ class MaintenanceRun(ResourceModel):
     errors: List[ErrorOccurrence] = Field(
         ...,
         description=(
-            "The protocol's fatal error, if there was one."
+            "The run's fatal error, if there was one."
             " For historical reasons, this is an array,"
             " but it won't have more than one element."
         ),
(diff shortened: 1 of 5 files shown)


robot-server/robot_server/maintenance_runs/maintenance_run_models.py
```diff
@@ -76,7 +76,11 @@ class MaintenanceRun(ResourceModel):
     )
     errors: List[ErrorOccurrence] = Field(
         ...,
-        description="Any errors that have occurred during the run.",
+        description=(
+            "The protocol's fatal error, if there was one."
+            " For historical reasons, this is an array,"
+            " but it will only have up to one element."
```
ecormany:
```suggestion
            " but it will never have more than one element."
```
sounds a little more natural, but nbd as-is.

sfoster1 (approved):
Looks good to me!

robot-server/robot_server/maintenance_runs/maintenance_run_models.py
```diff
@@ -76,7 +76,11 @@ class MaintenanceRun(ResourceModel):
     )
     errors: List[ErrorOccurrence] = Field(
         ...,
-        description="Any errors that have occurred during the run.",
+        description=(
+            "The protocol's fatal error, if there was one."
```
SyntaxColoring:
```suggestion
            "The run's fatal error, if there was one."
```

SyntaxColoring:
Got an in-person approval from @sanni-t.

PR #13193, opened 2023-07-28
fix(app-shell-odd): fix the brightness with sleep touchscreen

koji:
# Overview
This pr avoids updating brightness config value when the app sleeps
Add condition before updating the config value. When path is brightness and value is 7, app-shell-odd skips storing the value.

The current implementation isn't good and it should be temporary since changing the brightness needs to edge the file that we should be via endpoint.

I could not reproduce the issue that when tapping the display, display stays off on my end.

RAUT-592

# Test Plan
0. Check config.json brightness
1. Go to Settings
2. Tap Touchscreen sleep and set sleep time 3min
3. After 3 min, display is off and config.json brightness is the same as step 0
4. /sbin/reboot or reboot the robot via desktop app
5. Check display is on

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app-shell-odd/src/config/index.ts
+++ b/app-shell-odd/src/config/index.ts
@@ -106,8 +106,10 @@ export function registerConfig(dispatch: Dispatch): (action: Action) => void {
         }
 
         log().debug('Updating config', { path, nextValue })
-        store().set(path, nextValue)
-        dispatch(Cfg.configValueUpdated(path, nextValue))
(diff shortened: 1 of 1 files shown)


shlokamin (changes requested):
i think what we wanna do here rather than updating the config value when sleeping is write directly to the operating system's brightness file. 

the config is meant to be a user setting, so when the screen turns back on we can reset the brightness to whatever the config says.

so, we'll need to expose a method via the app shell odd to write to the OS's brightness file, and then have the app call the function. does that make sense?

PR #13195, opened 2023-07-28
fix(app): show applied offset count in ODD protocol setup step

b-cooper:
# Overview

Expose the number of applied offsets on a given run within the Labware Position Check setup step
Screen Shot 2023-07-28 at 5 54 56 PM
Screen Shot 2023-07-28 at 5 54 47 PM

# Review requests

Run LPC on ODD, confirm that the number of offsets present is visible on the Labware Position Check setup step

# Risk assessment
low


koji (approved):
The changes look good to me.
Thank you for fixing Estop modals' storybook titles.

PR #13196, opened 2023-07-28
feat(app): add gripper calibration to device settings

smb2268:
fix RQA-821

# Overview
This PR wires up the gripper calibration section of Device Settings/Calibration

# Test Plan
Look at the device settings page of a Flex, see that there is a gripper cal section if the gripper is attached. The last calibrated date should be reflected and overflow menu will have either "Calibrate gripper" or "Recalibrate gripper" CTA that will launch the calibration flow.

# Changelog
Wire up gripper calibration data
Add test coverage

# Review requests
Go through test plan

# Risk assessment
Low


app/src/organisms/RobotSettingsCalibration/index.tsx
```diff
@@ -120,6 +122,11 @@ export function RobotSettingsCalibration({
 
   const pipetteOffsetCalibrations = useAllPipetteOffsetCalibrationsQuery().data
     ?.data
+  const { data: attachedInstruments } = useInstrumentsQuery()
```
koji: Gripper Calibration section won't fetch the data after calibration.
But the thing is that useInstrumentsQuery includes pipette data.
Maybe maybe we can add if statement with `isOnDevice` to line 250-252  to execute fetch when a robot is an OT-2 🤔 

```suggestion
  const { data: attachedInstruments } = useInstrumentsQuery({refetchInterval: CALS_FETCH_MS})
```

b-cooper:
@koji I added a commit that polls the instruments endpoint as per you suggestion, but also consolidates the queries now that we can safely hit the GET /instruments endpoint for OT2's

koji (approved):
lgtm
Checked gripper cal on APPandUIBot and the last calibrated was updated properly.

PR #13202, opened 2023-08-01
fix(hardware-testing): only delay for scale stabilization if not simulating

andySigler:
# Overview

Fixes bug where `sleep()` wasn't being skipped during simulation runs, causing integration tests to have a 30-seconds delay per test run.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
nice

PR #13203, opened 2023-08-01
feat(api): Adds alignment check to ot3 belt-calibration

andySigler:
# Overview

This PR adds an alignment check immediately following the belt-calibration sequence, to check and see how much XYZ axes are shifting relative to the deck when the gantry moves from one side of the machine to the other.

The Flex 96ch must be aligned with the deck so that both it's A1 and H12 channels can enter a 384 well plate without collision. This means the deck-to-gantry mechanical alignment must not exceed some amount of rotation along XY or tilt along Z.

The `ot3_calibration.calibrate_belts()` method compensates for rotation around XY, but if a machine if found to exceed that maximum tolerable rotation while running the belt-calibration sequence, then we should not allow that calibration to take affect.

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -1010,3 +1011,100 @@ class OT3Transforms(RobotCalibration):
     left_mount_offset: Point
     right_mount_offset: Point
     gripper_mount_offset: Point
+
+
+class AlignmentError(RuntimeError):
```
sfoster1: Make this inherit from something in `shared-data/python/errors/exceptions.py` (make something new and a new error code in the `2xxx` range if you need to)

api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -1010,3 +1011,100 @@ class OT3Transforms(RobotCalibration):
...
+        super().__init__(
+            f"this machine is misaligned and requires maintenance: "
+            f"x-along-y: {x}, y-along-x: {y}, z-along-x: {zx}, z-along-y: {zy}"
+        )
+
+
+@dataclass
+class CalibrationSlot:
+    slot: int
+    nominal: Optional[Point]
```
sfoster1: don't make these optional - as you can see it leads to annoying type-ignores later. Instead, have this:
```
def _find_slot(slot: int) -> CalibrationSlot:
    actual, nominal = await find_slot_binary_..._(hcapi, mount, slot)
    await hcapi.retract()
    return CalibrationSlot(slot, actual, nominal)
```

Then build the `BeltCalibrationData` after the fact.

This ties together "data is present" and "object exists" for `BeltCalibrationData` which means you don't need `_has_cal_data()` anymore (if you didn't, the object wouldn't be there). It also means you don't need any of the `type: ignore`s.

api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -1010,3 +1011,100 @@ class OT3Transforms(RobotCalibration):
     left_mount_offset: Point
     right_mount_offset: Point
     gripper_mount_offset: Point
+
+
+class AlignmentError(RuntimeError):
+    def __init__(self, x: float, y: float, zx: float, zy: float) -> None:
```
sfoster1: Let's have this system indicate which data is correct and which is not. When we use an `EnumeratedError` from `shared-data/python/opetnrons_shared_data/errors/exceptions.py`, we can put each point, each deviation, and each error threshold in the `details` dict, which will help; but let's make the message include which axes are off.

sfoster1 (changes requested):
Definitely a good thing to do but there's a couple things I'd like to see change about the code.

sfoster1 (approved):
Awesome, thank you for doing the new errors!

PR #13205, opened 2023-08-01
fix(app): do not call direction control handlers if disabled

shlokamin:
# Overview

This PR fixes an issue where horizontal direction controls in the app here getting called twice via keypress.

# Test Plan

- Ran LPC, verified jog commands via key press work as expected

# Changelog

- Do not call direction control handlers if disabled

# Review requests
Run through LPC and verify that key presses work as expected when issuing jog commands

# Risk assessment

Low


b-cooper (approved):
✅

koji (approved):

PR #13206, opened 2023-08-01
fix(app): show robot network connection in app if able to communicate via that connection

brenthagen:
# Overview

adds a check to ensure a connected network status ip reported by a robot has a healthy response from the desktop app discovery-client. for example, robots connected to a different wifi network than the client app should not show a wifi connection in the app, because the app cannot communicate with that robot via wifi. updated robot settings networking tab, and robot status header rendered on devices and device details pages.

closes RQA-1150

example robot settings networking tab when robot connected to a different wifi network:

Screen Shot 2023-07-31 at 5 15 50 PM

# Test Plan

 - manually verified app behavior for robot connected to a different wifi network
 - updated unit tests

# Changelog

 -  Show robot network connection in app if able to communicate via that connection

# Review requests

confirm on robot connected to different wifi network than app

# Risk assessment

low

### DIFF
--- a/app/src/organisms/Devices/RobotSettings/RobotSettingsNetworking.tsx
+++ b/app/src/organisms/Devices/RobotSettings/RobotSettingsNetworking.tsx
@@ -63,19 +63,30 @@ export function RobotSettingsNetworking({
 
   const canDisconnect = useCanDisconnect(robotName)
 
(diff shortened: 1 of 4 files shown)


koji (approved):
The changes look good to me

PR #13207, opened 2023-08-01
fix(protocol-engine): Propagate errors during cleanup phase

SyntaxColoring:
# Overview

When a run ends (either by completing or by erroring out), the robot homes and drops tips in the trash.

This fixes RSS-146, which is a bug where, if something went wrong during those post-run cleanup steps, the run would get stuck permanently in the "finishing" state.

# Changelog

If an error happens during the cleanup phase, treat it basically as if the error happened during the main part of the run. That means returning the error over HTTP in the top-level `errors` field, with our existing `ErrorOccurrence` model. And mark the run's final status as `failed`.

The spicy case is when the main part of the run fails, *and* the post-run cleanup phase fails. We want to propagate *both* errors. From discussions with @sfoster1, we should do this by returning a root `RunAndFinishFailed` error, with both of the real errors as its children:

```
                 RunAndFinishFailed
                /                  \
       Run error                    Finish error
           |                             |
 ...underlying errors...       ...underlying errors...
```

(See #13191 for why we want there to be a single node at the root.)

Here's a real example of what that looks like over HTTP:

<details>

```json
{
  "id": "59482a84-9c17-4bc2-bc3d-e53d73182603",
  "errorType": "RunAndFinishFailed",
  "createdAt": "2023-08-01T13:51:48.997575+00:00",
  "detail": "The run had a fatal error, and another error happened while doing post-run cleanup.",
  "errorCode": "4000",
  "errorInfo": {},
  "wrappedErrors": [
    {
      "id": "62f6b19f-21ae-496c-b1e0-103566969ad5",
      "errorType": "ExceptionInProtocolError",
      "createdAt": "2023-08-01T13:51:30.811210+00:00",
      "detail": "ZeroDivisionError [line 14]: Jesus Christ",
      "errorCode": "4000",
      "errorInfo": {},
      "wrappedErrors": [
        {
          "id": "62f6b19f-21ae-496c-b1e0-103566969ad5",
          "errorType": "PythonException",
          "createdAt": "2023-08-01T13:51:30.811210+00:00",
          "detail": "ZeroDivisionError: Jesus Christ",
          "errorCode": "4000",
          "errorInfo": {
            "args": "('Jesus Christ',)",
            "traceback": "  File \"/opt/opentrons-robot-server/opentrons/protocols/execution/execute_python.py\", line 61, in run_python\n    exec(\"run(__context)\", new_globs)\n\n  File \"<string>\", line 1, in <module>\n\n  File \"raise_after_picking_up.py\", line 14, in run\n",
            "class": "ZeroDivisionError"
          },
          "wrappedErrors": []
        }
      ]
    },
    {
      "id": "59482a84-9c17-4bc2-bc3d-e53d73182603",
      "errorType": "PythonException",
      "createdAt": "2023-08-01T13:51:48.997575+00:00",
      "detail": "opentrons.hardware_control.types.FailedTipStateCheck: Failed to correctly determine tip state for tip ABSENT received True but expected False",
      "errorCode": "4000",
      "errorInfo": {
        "args": "('Failed to correctly determine tip state for tip ABSENT received True but expected False',)",
        "traceback": "  File \"/opt/opentrons-robot-server/opentrons/protocol_engine/protocol_engine.py\", line 298, in finish\n    await exit_stack.aclose()\n\n  File \"/usr/lib/python3.8/contextlib.py\", line 621, in aclose\n    await self.__aexit__(None, None, None)\n\n  File \"/usr/lib/python3.8/contextlib.py\", line 679, in __aexit__\n    raise exc_details[1]\n\n  File \"/usr/lib/python3.8/contextlib.py\", line 662, in __aexit__\n    cb_suppress = await cb(*exc_details)\n\n  File \"/usr/lib/python3.8/contextlib.py\", line 557, in _exit_wrapper\n    await callback(*args, **kwds)\n\n  File \"/opt/opentrons-robot-server/opentrons/protocol_engine/execution/hardware_stopper.py\", line 101, in do_stop_and_recover\n    await self._drop_tip()\n\n  File \"/opt/opentrons-robot-server/opentrons/protocol_engine/execution/hardware_stopper.py\", line 80, in _drop_tip\n    await self._tip_handler.drop_tip(\n\n  File \"/opt/opentrons-robot-server/opentrons/protocol_engine/execution/tip_handler.py\", line 101, in drop_tip\n    await self._hardware_api.drop_tip(\n\n  File \"/opt/opentrons-robot-server/opentrons/hardware_control/thread_manager.py\", line 151, in wrapper\n    return await call_coroutine_threadsafe(loop, attr, *args, **kwargs)\n\n  File \"/opt/opentrons-robot-server/opentrons/hardware_control/thread_manager.py\", line 54, in call_coroutine_threadsafe\n    return await wrapped\n\n  File \"/opt/opentrons-robot-server/opentrons/hardware_control/ot3api.py\", line 1795, in drop_tip\n    await self._backend.get_tip_present(realmount, TipStateType.ABSENT)\n\n  File \"/opt/opentrons-robot-server/opentrons/hardware_control/backends/ot3controller.py\", line 770, in get_tip_present\n    raise FailedTipStateCheck(tip_state, res)\n",
        "class": "FailedTipStateCheck"
      },
      "wrappedErrors": []
    }
  ]
}
```

</details>

And a bonus: Fix an unrelated old concurrency hazard with `DoorWatcher`.

# Test Plan

* [x] Regression test: Induce an error during the main part of the run, and make sure it's still propagated. The run should show up as failed in the app, with some error details. Further error details should be available over HTTP.

  One way is to run a Python protocol that does `raise Exception()`.

* [x] New code test: Induce an error during the post-run finish steps (homing and dropping tips), and make sure it's propagated now just like errors in the main part of the run.

  One way to induce an error like this is to use your finger to press the pipette ejector up into the pipette body immediately after it homes and drops the tip in the trash. The hardware controller will raise an exception because it thinks it failed to eject the tip.

* [x] New code test: Induce an error during the main part of the run *and* the post-run finish steps, and make sure they're *both* propagated.

  I think it's a little up in the air how we want this to show up in the app. But over HTTP, at least, the details from both errors should be present.

  Here's how it currently shows up on the on-device display:

  IMG_6516

# Review requests

* I would love opinions on naming. I inconsistently see this called the "finishing steps," "recovery steps," "cleanup steps," etc.
* Do the fixes to `DoorWatcher` make sense?

# Risk assessment

Low. The way we calculate the run state ("failed," "succeeded", etc.) is kind of complicated, but I'm satisfied with the testing.

We could cut the `DoorWatcher` fix out of this if we want to minimize the diff.

### DIFF
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -1,4 +1,5 @@
 """ProtocolEngine class definition."""
+from contextlib import AsyncExitStack, ExitStack
 from typing import Dict, Optional
 
 from opentrons.protocols.models import LabwareDefinition
@@ -284,22 +285,32 @@ class ProtocolEngine:
             FinishAction(error_details=error_details, set_run_status=set_run_status)
         )
 
+        exit_stack = AsyncExitStack()
+        exit_stack.push_async_callback(self._queue_worker.join)
+        exit_stack.callback(self._door_watcher.stop)
+        exit_stack.push_async_callback(
+            self._hardware_stopper.do_stop_and_recover,
+            drop_tips_and_home=drop_tips_and_home,
+        )
+        exit_stack.push_async_callback(self._plugin_starter.stop)
+
         try:
-            await self._queue_worker.join()
-
-        # todo(mm, 2022-01-31): We should use something like contextlib.AsyncExitStack
-        # to robustly clean up all these resources
-        # instead of try/finally, which can't scale without making indentation silly.
-        finally:
-            self._door_watcher.stop()
-
-            await self._hardware_stopper.do_stop_and_recover(drop_tips_and_home)
-
-            completed_at = self._model_utils.get_timestamp()
-            self._action_dispatcher.dispatch(
-                HardwareStoppedAction(completed_at=completed_at)
+            await exit_stack.aclose()
+        except Exception as hardware_stopped_exception:
+            finish_error_details: Optional[FinishErrorDetails] = FinishErrorDetails(
+                error_id=self._model_utils.generate_id(),
+                created_at=self._model_utils.get_timestamp(),
+                error=hardware_stopped_exception,
             )
-            await self._plugin_starter.stop()
+        else:
+            finish_error_details = None
+
+        self._action_dispatcher.dispatch(
+            HardwareStoppedAction(
+                completed_at=self._model_utils.get_timestamp(),
+                finish_error_details=finish_error_details,
+            )
+        )
 
     def add_labware_offset(self, request: LabwareOffsetCreate) -> LabwareOffset:
         """Add a new labware offset and return it.
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -291,10 +291,7 @@ class ProtocolEngine:
         # to robustly clean up all these resources
         # instead of try/finally, which can't scale without making indentation silly.
         finally:
-            # Note: After we stop listening, straggling events might be processed
-            # concurrently to the below lines in this .finish() call,
-            # or even after this .finish() call completes.
-            self._door_watcher.stop_soon()
+            self._door_watcher.stop()
 
             await self._hardware_stopper.do_stop_and_recover(drop_tips_and_home)
 
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -253,9 +253,9 @@ class ProtocolEngine:
         and then shut down. After an engine has been `finished`'ed, it cannot
         be restarted.
 
-        This method should not raise, but if any exceptions happen during
-        execution that are not properly caught by the CommandExecutor, they
-        will be raised here.
+        This method should not raise. If any exceptions happened during execution that were not
+        properly caught by the CommandExecutor, or if any exceptions happen during this
+        `finish()` call, they should be saved as `.state_view.get_summary().errors`.
 
         Arguments:
             error: An error that caused the stop, if applicable.
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -1,5 +1,5 @@
 """ProtocolEngine class definition."""
-from contextlib import AsyncExitStack, ExitStack
+from contextlib import AsyncExitStack
 from typing import Dict, Optional
 
 from opentrons.protocols.models import LabwareDefinition
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -1,5 +1,6 @@
 """ProtocolEngine class definition."""
 from contextlib import AsyncExitStack
+from logging import getLogger
 from typing import Dict, Optional
 
 from opentrons.protocols.models import LabwareDefinition
@@ -48,6 +49,9 @@ from .actions import (
     ResetTipsAction,
     SetPipetteMovementSpeedAction,
 )
+
+
+_log = getLogger(__name__)
 
 
 class ProtocolEngine:
@@ -302,6 +306,7 @@ class ProtocolEngine:
             # If any teardown steps failed, this will raise something.
             await exit_stack.aclose()
         except Exception as hardware_stopped_exception:
+            _log.exception("Exception during post-run finish steps.")
             finish_error_details: Optional[FinishErrorDetails] = FinishErrorDetails(
                 error_id=self._model_utils.generate_id(),
                 created_at=self._model_utils.get_timestamp(),
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -285,16 +285,21 @@ class ProtocolEngine:
             FinishAction(error_details=error_details, set_run_status=set_run_status)
         )
 
+        # We have a lot of independent things to tear down. If any teardown fails, we want
+        # to continue with the rest, to avoid leaking resources or leaving the engine with a broken
+        # state. We use an AsyncExitStack to avoid a gigantic try/finally tree. Note that execution
+        # order will be backwards because the stack is first-in-last-out.
         exit_stack = AsyncExitStack()
-        exit_stack.push_async_callback(self._queue_worker.join)
-        exit_stack.callback(self._door_watcher.stop)
+        exit_stack.push_async_callback(self._plugin_starter.stop)  # Last step.
         exit_stack.push_async_callback(
             self._hardware_stopper.do_stop_and_recover,
             drop_tips_and_home=drop_tips_and_home,
         )
-        exit_stack.push_async_callback(self._plugin_starter.stop)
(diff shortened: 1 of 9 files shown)


api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -284,25 +285,32 @@ async def finish(
...
-            # or even after this .finish() call completes.
-            self._door_watcher.stop_soon()
-
-            await self._hardware_stopper.do_stop_and_recover(drop_tips_and_home)
-
-            completed_at = self._model_utils.get_timestamp()
-            self._action_dispatcher.dispatch(
-                HardwareStoppedAction(completed_at=completed_at)
+            await exit_stack.aclose()
+        except Exception as hardware_stopped_exception:
```
SyntaxColoring, line 308: To do: Log the traceback here in case nobody saves the HTTP response.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -284,25 +285,32 @@ async def finish(
             FinishAction(error_details=error_details, set_run_status=set_run_status)
         )
 
+        exit_stack = AsyncExitStack()
```
SyntaxColoring, line 296: ~~To do: Comments for `exit_stack`.~~ Edit: Done.

api/tests/opentrons/protocol_engine/state/test_command_store.py
```diff
@@ -876,20 +889,22 @@ def __init__(self, message: str) -> None:
...
+        ),
+        finish_error=None,
         run_started_at=None,
         latest_command_hash=None,
     )
 
 
+# TODO: Add finish error tests.
+
+
```
SyntaxColoring: Reminder to self to do this before merge.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -445,9 +475,31 @@ def get_slice(
...
+            combined_error = ErrorOccurrence.construct(
+                id=finish_error.id,
+                createdAt=finish_error.createdAt,
+                errorType="RunAndFinishFailed",
+                detail=(
+                    "The run had a fatal error,"
+                    " and another error happened while doing post-run cleanup."
+                ),
+                # TODO(mm, 2023-07-31): Consider adding a low-priority error code so clients can
+                # deemphasize this root node, in favor of its children in wrappedErrors.
```
sfoster1, line 493: this does actually exist on the client side I believe

sfoster1 (approved):
This looks really solid to me and I agree with the error structuring and the stopping flow.

api/src/opentrons/protocol_engine/protocol_engine.py
TamarZanzouri: You might have reviewed this in person but would love ab explanation about these following changes

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -377,6 +369,44 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
                 elif action.door_state == DoorState.CLOSED:
                     self._state.is_door_blocking = False
```
TamarZanzouri, line 371: Do these 2 mapping methods need to be separated? Seems like it has the same signature and we can know by the exception type what to raise? Maybe Im missing something

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -445,9 +475,31 @@ def get_slice(
...
+                createdAt=finish_error.createdAt,
+                errorType="RunAndFinishFailed",
+                detail=(
+                    "The run had a fatal error,"
+                    " and another error happened while doing post-run cleanup."
+                ),
+                # TODO(mm, 2023-07-31): Consider adding a low-priority error code so clients can
+                # deemphasize this root node, in favor of its children in wrappedErrors.
+                errorCode=ErrorCodes.GENERAL_ERROR.value.code,
+                wrappedErrors=[
```
TamarZanzouri, line 495: 👍

api/src/opentrons/protocol_engine/protocol_engine.py
SyntaxColoring: The old code was basically doing this:

```python
clean_up_resource_1()
clean_up_resource_2()
clean_up_resource_3()
clean_up_resource_4()
...
```

In general, any time you have code like that, it's probably a bug: if an exception raises from `clean_up_resource_1()`, your code won't reach any of the subsequent lines, and those resources will leak. Every resource should at least be attempted to be cleaned up, no matter what.

You could fix it like this:

```python
try:
    clean_up_resource_1()
finally:
    try:
        clean_up_resource_2()
    finally:
        try:
            clean_up_resource_3()
        finally:
            try:
                clean_up_resource_4()
            finally:
                ...
```

But that's pretty unscalable and horrible to read.

You can use an [`ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) or [`AsyncExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack) to flatten it out. The code below is equivalent to the `try`/`finally` tree:

```python
exit_stack = contextlib.ExitStack()
...
exit_stack.callback(clean_up_resource_4)
exit_stack.callback(clean_up_resource_3)
exit_stack.callback(clean_up_resource_2)
exit_stack.callback(clean_up_resource_1)
exit_stack.close()
```

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -284,25 +289,38 @@ async def finish(
...
+            # If any teardown steps failed, this will raise something.
+            await exit_stack.aclose()
+        except Exception as hardware_stopped_exception:
+            _log.exception("Exception during post-run finish steps.")
+            finish_error_details: Optional[FinishErrorDetails] = FinishErrorDetails(
+                error_id=self._model_utils.generate_id(),
+                created_at=self._model_utils.get_timestamp(),
+                error=hardware_stopped_exception,
             )
-            await self._plugin_starter.stop()
```
sanni-t, line 305: Not fully sure of what this line does but it's now going to be handled before the `HardwareStoppedAction` is dispatched. I'm guessing the order doesn't matter in this case?

api/src/opentrons/protocol_engine/protocol_engine.py
TamarZanzouri: amazing explanation! thank you!

api/src/opentrons/protocol_engine/state/commands.py
TamarZanzouri: I see. for the run we are raising the `UnexpectedProtocolError` and for the finish we are raising `PythonException`

TamarZanzouri (approved):
Looks great and great documentation :-)

api/src/opentrons/protocol_engine/protocol_engine.py
SyntaxColoring: Yeah. As far as I can tell, order doesn't matter in this case.

I made this change because if a `ProtocolEngine` is now catching and storing its own cleanup errors, `self._plugin_starter.stop()` needs to be moved earlier. If it's the last step like it was originally, it's too late to catch any errors that come from it.

A different way of handling this could have been for `ProtocolEngine` to only catch and store *certain kinds* of its own cleanup errors, but that seems like too confusing and subtle of an API to me.

PR #13208, opened 2023-08-01
fix(api, shared-data): raise error when using gripper in ot2 protocols

sanni-t:
# Overview

Closes RSS-297

# Test Plan

Check if an OT2 protocol fails analysis when there's a `move_labware` step using gripper.

# Changelog

- added new `NotSupportedOnRobotType` error
- added the robot type check

# Review requests

- usual

# Risk assessment

None.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/move_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/move_labware.py
@@ -122,7 +122,8 @@ class MoveLabwareImplementation(
         if params.strategy == LabwareMovementStrategy.USING_GRIPPER:
             if self._state_view.config.robot_type == "OT-2 Standard":
                 raise NotSupportedOnRobotType(
-                    "Labware movement using a gripper is not supported on the OT2"
+                    message="Labware movement using a gripper is not supported on the OT2",
+                    details={"strategy": params.strategy},
                 )
             if labware_validation.validate_definition_is_adapter(
                 current_labware_definition
--- a/api/src/opentrons/protocol_engine/commands/move_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/move_labware.py
@@ -122,7 +122,7 @@ class MoveLabwareImplementation(
         if params.strategy == LabwareMovementStrategy.USING_GRIPPER:
             if self._state_view.config.robot_type == "OT-2 Standard":
                 raise NotSupportedOnRobotType(
-                    message="Labware movement using a gripper is not supported on the OT2",
+                    message="Labware movement using a gripper is not supported on the OT-2",
                     details={"strategy": params.strategy},
                 )
             if labware_validation.validate_definition_is_adapter(
--- a/api/src/opentrons/protocol_engine/commands/move_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/move_labware.py
@@ -12,7 +12,7 @@ from ..types import (
     LabwareOffsetVector,
     LabwareMovementOffsetData,
 )
-from ..errors import LabwareMovementNotAllowedError
+from ..errors import LabwareMovementNotAllowedError, NotSupportedOnRobotType
 from ..resources import labware_validation
 from .command import AbstractCommandImpl, BaseCommand, BaseCommandCreate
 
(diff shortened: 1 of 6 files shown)


api/src/opentrons/protocol_engine/commands/move_labware.py
```diff
@@ -120,6 +120,10 @@ async def execute(self, params: MoveLabwareParams) -> MoveLabwareResult:
         )
 
         if params.strategy == LabwareMovementStrategy.USING_GRIPPER:
+            if self._state_view.config.robot_type == "OT-2 Standard":
+                raise NotSupportedOnRobotType(
```
sfoster1, line 124: maybe add like `"action": "gripper-move-labware"` or something to details?

sfoster1 (approved):
Looks good to me!

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -789,3 +789,16 @@ def __init__(
...
+    """Raised when attempting to perform an action that is not supported for the given RobotType."""
+
+    def __init__(
+        self,
+        message: Optional[str] = None,
+        details: Optional[Dict[str, Any]] = None,
+        wrapping: Optional[Sequence[EnumeratedError]] = None,
+    ) -> None:
+        """Build a NotSupportedOnRobotType exception."""
+        super().__init__(ErrorCodes.E_STOP_ACTIVATED, message, details, wrapping)
```
sfoster1: Wait no this is not the right code for this. Let's add a 4000 series "not supported on hardware" error code in `shared-data/errors/definitions/1.json`, a python binding for the code, and maybe an exception in shared-data (don't necessarily need it if you're building a protocol engine exception instead) for it.

sfoster1 (changes requested):
Can't use the estop error code for this, let's make a new one for hardware not supported

api/src/opentrons/protocol_engine/errors/exceptions.py
sanni-t: 🤦🏻‍♀️ you can tell where I copy pasted that from.
Will make the changes

api/src/opentrons/protocol_engine/commands/move_labware.py
SyntaxColoring: What's the difference between that and putting `"Labware movement using a gripper is not supported on the OT2"` in `message`?

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -789,3 +791,18 @@ def __init__(
     ) -> None:
         """Build an EStopActivatedError."""
         super().__init__(ErrorCodes.E_STOP_ACTIVATED, message, details, wrapping)
+
+
+class NotSupportedOnRobotType(ProtocolEngineError):
+    """Raised when attempting to perform an action that is not supported for the given robot type."""
```
SyntaxColoring, line 797: What's the difference between `HardwareNotSupportedError` and `NotSupportedOnRobotType`?

api/src/opentrons/protocol_engine/commands/move_labware.py
```diff
@@ -120,6 +120,11 @@ async def execute(self, params: MoveLabwareParams) -> MoveLabwareResult:
...
+                    message="Labware movement using a gripper is not supported on the OT2",
```
ecormany:
```suggestion
                    message="Labware movement using a gripper is not supported on the OT-2",
```
😇

api/src/opentrons/protocol_engine/errors/exceptions.py
sanni-t: `HardwareNotSupportedError` is raised after checking the specific hardware API type and will only be able to be checked when running on an actual robot (or emulator), while `NotSupportedOnRobotType` checks only the robot type specified for the particular protocol/ engine. Subtle difference but an analysis saying 'not supported on *hardware*' sounds kinda misleading.

api/src/opentrons/protocol_engine/commands/move_labware.py
sfoster1: nothing :) it's just nice to have both

sfoster1 (approved):
Great! Thank you!

PR #13209, opened 2023-08-01
feat(shared-data): aluminum block split adapter definition

jbleon95:
# Overview

This PR adds the `Opentrons 96 Well Aluminum Block` adapter definition to shared data, in addition to adding stacking overlaps for the `Bio-Rad 96 Well Plate 200 µL PCR` and the `NEST 96 Well Plate 100 µL PCR Full Skirt`, allowing both of those to be loaded on top of the aluminum block.

# Test Plan

Loading the compatible plates onto the aluminum block was tested with Postman, to check the command responses and ensure that when a pipette was moved to a well, the resulting `Point` it moved to was identical to moving to the joint definitions loaded onto the same deck slot with the same pipette.

# Changelog

- Added `Opentrons 96 Well Aluminum Block` adapter definition
- Added adapter to `stackingOffsetWithLabware` for `Bio-Rad 96 Well Plate 200 µL PCR` and the `NEST 96 Well Plate 100 µL PCR Full Skirt`
- Updated load labware logic so `biorad_96_wellplate_200ul_pcr` and `opentrons_96_wellplate_200ul_pcr_full_skirt` automatically load version 2 of the labware if no version is explicitly provided.

# Review requests

Just a general check that the name is good and that theirs nothing out of place in the new definition.

# Risk assessment

Low, just adding a new definition. The only API changes are automatically loading `2.json` for the new biorad definition, and fixing not loading the version 2 of `opentrons_96_wellplate_200ul_pcr_full_skirt`. In the future a more holistic way of loading these will be added, hence the TODO statement also included.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/load_labware_params.py
+++ b/api/src/opentrons/protocol_api/core/engine/load_labware_params.py
@@ -7,7 +7,8 @@ from opentrons.protocol_engine.state.labware import LabwareLoadParams
 # Default versions of Opentrons standard labware definitions in Python Protocol API
 # v2.14 and above. Labware not explicitly listed here default to 1.
 #
-# This will need to be extended t
+# TODO(jbl 2023-08-01) this needs to be done more holistically, both to find the version and make sure that
+#   it corresponds to the API level is was released with
 _APILEVEL_2_14_OT_DEFAULT_VERSIONS: Dict[str, int] = {
     # v1 of many labware definitions have wrong `zDimension`s. (Jira RSS-202.)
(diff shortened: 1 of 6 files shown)


jerader:
what about `Opentrons 24 Well Aluminum Block`?

jbleon95:
> what about `Opentrons 24 Well Aluminum Block`?

the 24 Well Aluminum block is not being made a separated definition since the only labware that goes on there is tube racks, which aren't being split.

sfoster1 (approved):
Looks good to me!

sanni-t (approved):
LGTM!

PR #13210, opened 2023-08-01
feat(hardware-testing): Adds first line to CSV reports helping operators copy data

andySigler:
# Overview

As requested by SZ team, the first line in all of our CSV reports should include a single row with the word `copy` populating each column of the CSV where data is present. This will help operators know how many columns they should be copy/pasting into our templates.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
👍

PR #13211, opened 2023-08-01
Production qc ot3 overnight test

djgomez-opentrons:
# Overview

Adds a stress test to the production QC testing suite.

# Test Plan

Tested on DVT robot in BK


andySigler (approved):
LGTM, we ran it together and tested out different scenarios of it stalling and failing. If we are good with the number of cycles we default to, then it's good to go

PR #13213, opened 2023-08-01
fix(app): add a condition to firmware takeover for unboxing flow

koji:
# Overview
Currently if a user does factory reset and does the unboxing flow again when the user  attaches instruments, Firmware update modal may show up during the unboxing flow.

close RAUT-611

# Test Plan

# Changelog
- rename the custom hook since as Shlok mentioned the name isn't clear
- add unbox flow check in FirmwareUpdateTakeover

# Review requests

# Risk assessment
low


koji:
This will be closed when E-stop PR is merged into edge

PR #13214, opened 2023-08-01
feat(gripper):  specify whether or not a gripper move should stay engaged after an error

ahiuchingau:
# Overview
If the gripper is actively gripping a labware when a protocol gets terminated prematurely, we want to make sure the gripper doesn't just release the labware wherever as part of the protocol cleanup process, instead, we should allow user to recover the labware safely. The first step to achieving this is to tell firmware whether the current move is a labware-grip or an idling grip by using the `stay engaged` param in the GripperGripRequest.

Gripper firmware will use this information to keep the motor engaged if it receives an error while holding a labware. 

Next step is to retrieve the real-time gripper jaw state from the firmware. We can then use it to determine the proper cleanup process for the gripper after a protocol ends.

Changes in this PR need to be used together with https://github.com/Opentrons/ot3-firmware/pull/711

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Looks good to me!

PR #13220, opened 2023-08-02
fix(app): fix app white screen issue when open Device Details

koji:
# Overview
The previous implementation passed setState as a function instead of a callback function
close RQA-1153

Screenshot 2023-08-02 at 3 24 58 PM

# Test Plan

# Changelog
- call setState as a callback function

# Review requests

# Risk assessment
low


sfoster1 (approved):
great!

PR #13221, opened 2023-08-02
refactor(app): do not show firmware update modal if maintenance run in progress

shlokamin:
# Overview

The useEffect that triggers showing the update modal on the ODD didn't take in the current maintenance run query as a dependency. This PR just adds the maintenance run query's data to the dependency list

# Risk assessment

Low


koji (approved):
lgtm

PR #13222, opened 2023-08-02
feat(shared-data): add grip force and grip height to labware definitions

sanni-t:
# Overview

Closes RLAB-357

# Test Plan

Just unit- and integration tests should suffice for this since it's only updating definitions

# Changelog

- added `gripForce` and `gripHeightFromLabwareBottom` optional fields to labware schema and the supported, tested labware's definitions
- updated labwareDefinition model & type dict signature
- updated definition/schema verification tests

# Review requests

- make sure the grip force & height added are correct for the specific labware (tested values are in the sheet linked in the jira ticket above)
- 

# Risk assessment


shared-data/js/__tests__/labwareDefSchemaV2.test.ts
```diff
@@ -82,6 +82,7 @@ const expectedWellsNotMatchingZDimension: Record<string, Set<string>> = {
   // These height mismatches need to be investigated. See Jira RSS-202.
   // Each one should either be explained here or marked as a known bug.
   'nest_1_reservoir_195ml/1.json': new Set(['A1']),
+  'nest_1_reservoir_195ml/2.json': new Set(['A1']),
```
sanni-t, line 85: @SyntaxColoring is it ok to keep the reservoir v2 definition as an exception to height check as well or should we correct it in the new version? I couldn't find any solid conclusion to this labware's anomaly in your RSS-202 sheet so don't know if there's any new info on that.

shared-data/labware/schemas/2.json
```diff
@@ -360,6 +360,14 @@
...
       },
       "required": ["default"]
+    },
+    "gripForce": {
+      "type": "number",
+      "description": "Force, in Newtons, with which the gripper should grip the labware."
+    },
+    "gripHeightFromLabwareBottom": {
+      "type": "number",
+      "description": "Z-height, from the labware bottom, where the gripper should grip the labware."
```
SyntaxColoring: Can we specify whether this is the center of the metal paddles, the center of the cushy black pads, or something else?

shared-data/js/__tests__/labwareDefSchemaV2.test.ts
SyntaxColoring: Thanks for checking. I don't have any new info, so let's leave it as an exception that still needs to be investigated.

shared-data/labware/schemas/2.json
SyntaxColoring: I see the spreadsheet column name is "Bottom of plate to middle of pad (mm)", which I guess refers to the cushy black pad. But there's a different column that says "Pad z height working distance"—do you know what that means?

SyntaxColoring (approved):
Spot-checked a few of the labware definitions, and they look good to me. Thanks!

One documentation request.

shared-data/labware/schemas/2.json
sanni-t: Ya I don't know what the "Pad z height working distance" means either but @andySigler confirmed that we should be using the 'Bottom of plate to middle of pad (mm)' column.

But good point about adding that detail in the description. I'll update that.

TamarZanzouri (approved):
LGTM!

PR #13224, opened 2023-08-02
feat(api): turn off modules when estop is pressed and forward estop messages to api listeners

fsinapi:
# Overview

When the estop is pressed, we need to deactivate the following module behaviors:
* Any thermal control
* Any shaking

This PR adds that automatic deactivation to the OT3Api class. In addition, this PR also hooks up the OT3Api to forward any EstopStateNotification events that occur as a result of the Estop being pressed, connected/disconnected, or acknowledged.

Note that, per the PRD, modules are not prevented from performing any thermal control once the Estop is engaged; they just need to be turned off when the estop is first pressed.

# Test Plan

On a Flex with the Thermocycler, Heater-Shaker, and Tempdeck I did the following:

* Set all of the modules to control temperature, and set the heater shaker to shake
* Hit the estop and observe all of the modules deactivating
* Repeat the first step and then start a protocol
* **Pause** the protocol and then hit the estop; observe all of the modules deactivating 

# Changelog
* Add ability to force module deactivation even if the execution manager is not running (protocol paused, etc)
* Add listener in OT3Api that is hooked up to the estop state machine to listen for (and forward) any estop events
* When an estop event signals that the estop was just pressed, deactivate all modules

# Review requests

I don't like the `asyncio.sleep` in the new test at all. The estop listener _needs_ to start the deactivation all at once and it _shouldn't_ block and wait for all of the deactivation tasks to complete though. 

# Risk assessment

Pretty low


caila-marashaj (approved):
Nice!

api/tests/opentrons/hardware_control/test_ot3_api.py
```diff
@@ -1671,3 +1683,51 @@ async def test_tip_presence_disabled_ninety_six_channel(
...
+
+    decoy.when(api._backend.module_controls.available_modules).then_return(
+        [tc, hs, md, td]
+    )
+
+    estop_event = EstopStateNotification(old_state=old_state, new_state=new_state)
+
+    api._update_estop_state(estop_event)
+
+    await asyncio.sleep(0.1)
```
ahiuchingau: I think you can do this instead if you don't want to use `asyncio.sleep`? 
```suggestion
    await asyncio.get_running_loop().run_in_executor(None, api._update_estop_state, estop_event)
```

ahiuchingau (approved):
LGTM

api/tests/opentrons/hardware_control/test_ot3_api.py
ahiuchingau: Wait actually this passes only ~50% of the time... I don't get it

fsinapi: Yeah I tried using `asyncio.all_tasks` to check for when the shutoff tasks all completed but that didn't seem to actually pick up any of the tasks and it would fail most of the time. The actual tasks finish pretty much instantly so this really just has to release the event loop, but I'm not sure why they won't show up 🤷 

I could set up the callback to return all of the futures and just ignore that return value when it gets called by the estop state machine.........

PR #13227, opened 2023-08-03
feat(app-shell-odd,robot-server): Allow resetting on-device display configuration

SyntaxColoring:
# Overview

This lets you reset the configuration of a Flex's on-device display. This lets you restart from the beginning of the setup flow. Closes RSS-298.

# Test Plan

We can't really cover this with automated tests because it's an interaction between three different projects (`robot-server`, `app`, and `app-shell-odd`). We need to test this manually.

* [ ] Use the desktop app to perform the reset. It should reboot the robot, and after the reboot, the Flex's on-device display should show the first-time setup instructions.
* [ ] Use the on-device-display to perform the reset. It should reboot the robot, and after the reboot, the Flex's on-device display should show the first-time setup instructions.

Per @shlokamin [in Slack](https://opentrons.slack.com/archives/C049WNGU22H/p1691420814855919):

> hey there, this PR makes it so that users can reset their ODD settings (brightness, sleep duration, ODD first time set up state) through both the desktop app’s factory reset slideout and the ODD’s device reset tab IF all other settings are selected. so this is a thing that happens implicitly if all other settings are selected, there is no specific checkbox or setting to select to clear the ODD’s settings.

We should include these tests in the release QA checklist, going forward.

# Changelog

We're handling this like we handle resetting robot-server's persistence directory.

The reset is deferred until the next boot. When the server handles the reset, it creates a special marker file on the robot's filesystem. The next time the `opentrons-robot-app` process launches, it looks for that file, and if it's there, it clears its configuration as the first thing. This avoids race conditions where `robot-server` tries to clear the file while `opentrons-robot-app` is actively reading and writing it.

# Review requests

Any ideas for adding automated tests to this?

# Risk assessment

Medium. This architecture is prone to reset loops and filesystem permission surprises.


app-shell-odd/src/main.ts
```diff
@@ -41,6 +50,12 @@ app.once('window-all-closed', () => {
 
 function startUp(): void {
   log.info('Starting App')
+  const storeNeedsReset = fse.existsSync(
+    path.join(ODD_DIR, `_CONFIG_TO_BE_DELETED_ON_REBOOT`)
+  )
+  if (storeNeedsReset) {
+    clearStore()
```
SyntaxColoring: Reminder to also delete the `_CONFIG_TO_BE_DELETED_ON_REBOOT` marker, so we don't get stuck in a reset loop.

robot-server/robot_server/service/legacy/reset_odd.py
```diff
@@ -0,0 +1,22 @@
...
+from anyio import Path as AsyncPath
+
+
+# The on-device display process knows to look for this file here.
+_RESET_MARKER_PATH: Final = AsyncPath("/data/ODD/_CONFIG_TO_BE_DELETED_ON_REBOOT")
+
+_RESET_MARKER_FILE_CONTENTS: Final = """\
+This file was placed here by robot-server.
+It tells the on-device display process to clear its configuration on the next boot.
+"""
```
sanni-t: Might be good to make it super clear that the file will also be deleted when ODD clears its configuration on the next boot

sanni-t (approved):
Python changes make sense to me!

robot-server/robot_server/service/legacy/reset_odd.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+async def mark_odd_for_reset_next_boot() -> None:
+    """Mark the configuration of the Flex's on-device display so it gets reset on the next boot.
+
+    This assumes you're running on a real Flex, as opposed to an OT-2 or a local dev machine.
+    """
+    await _RESET_MARKER_PATH.write_text(
```
SyntaxColoring: We should probably no-op if the parent directory doesn't exist, instead of raising an exception. Someone might factory-reset multiple things at once, and we don't want this to fail and prevent the other things from running.

robot-server/robot_server/service/legacy/reset_odd.py
SyntaxColoring, line 1: Maybe leave a comment like the one in the persistence resetter for why we need to do this file handshake thing.

sfoster1 (approved):
I think this is a good way to do it. Automated testing can really only happen unit-wise for this until and unless we get a full system emulator going and that is a long way off.

TamarZanzouri (approved):
Very nice! thank you for doing this so quick!

api/src/opentrons/config/reset.py
```diff
@@ -37,6 +37,7 @@ class ResetOptionId(str, Enum):
     gripper_offset = "gripperOffsetCalibrations"
     tip_length_calibrations = "tipLengthCalibrations"
     runs_history = "runsHistory"
+    on_device_display = "onDeviceDisplay"
```
koji, line 40: Maybe the name is factory_reset or all_data something since this reset option will be used by Desktop app and Touchscreen app?

PR #13228, opened 2023-08-03
fix(app,app-shell-odd): fix touchscreen brightness issue for reboot

koji:
# Overview
This PR changes the way to update the screen brightness and stops updating the config value when the touchscreen app goes sleep.

close RAUT-592

# Test Plan
- set Sleep config 3 min
- after 3 min, the screen turn off but the config value isn't updated to `7`
- touch the screen and check the turn on the brightness
- wait 3 min until the screen goes sleep. then ssh in a dev kit or a flex and run `/sbin/reboot`
Then after rebooting, the touchscreen app isn't in the sleep mode 

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app-shell-odd/src/systemd.ts
+++ b/app-shell-odd/src/systemd.ts
@@ -25,6 +25,7 @@ interface SystemD {
   setRemoteDevToolsEnabled: (enabled: boolean) => Promise<string>
   getisRobotServerReady: () => Promise<boolean>
   restartApp: () => Promise<string>
+  updateBrightness: (text: string) => Promise<string>
 }
 
 const provideExports = (): SystemD => {
@@ -46,6 +47,10 @@ const provideExports = (): SystemD => {
         ).then(state => state.trim() === 'active'),
       restartApp: () =>
         promisifyProcess(`/bin/systemctl restart opentrons-robot-app`),
+      updateBrightness: text =>
+        promisifyProcess(
+          `echo "${text}" | cat > /sys/class/backlight/backlight/device/backlight/backlight/brightness`
+        ),
     }
   } else {
     return {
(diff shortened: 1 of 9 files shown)


app-shell-odd/src/systemd.ts
```diff
@@ -46,6 +47,10 @@ const provideExports = (): SystemD => {
         ).then(state => state.trim() === 'active'),
       restartApp: () =>
         promisifyProcess(`/bin/systemctl restart opentrons-robot-app`),
+      updateBrightness: text =>
+        promisifyProcess(
+          `echo "${text}" | cat > /sys/class/backlight/backlight/device/backlight/backlight/brightness`
```
vegano1: can also do
`echo "${text}" > /sys/class/backlight/backlight/device/backlight/backlight/brightness`

app/src/App/OnDeviceDisplayApp.tsx
```diff
@@ -199,17 +200,19 @@ const onDeviceDisplayEvents: Array<keyof DocumentEventMap> = [
...
+    getOnDeviceDisplaySettings
+  )
 
   const sleepTime = sleepMs != null ? sleepMs : SLEEP_NEVER_MS
   const options = {
     events: onDeviceDisplayEvents,
     initialState: false,
   }
-  const [usersBrightness, setUsersBrightness] = React.useState(brightness)
+  // const [usersBrightness, setUsersBrightness] = React.useState(brightness)
```
vegano1: needed?

koji: no, I will remove this.
thanks!

app-shell-odd/src/systemd.ts
koji: Thanks, I will switch from the current to your suggestion.

vegano1 (approved):
Looking good!

Tested on the ODD
1. Set Screen timeout to 3m
2. Wait for the screen to turn off
3. verify that the `brightness` key in the `/data/ODD/config.json` file Is unchanged. 
4. once off, reboot the device
5. once on, verify that the screen comes on once the robot-app starts up
6. verify that the `brightness` key in the `/data/ODD/config.json` file Is the same as before.
7. verify that the brightness file `/sys/class/backlight/backlight/device/backlight/backlight/brightness` is set to the value of the `brightness` key in the `/data/ODD/config.json` file.
8. verify that the screen stills times out after 3 minutes

shlokamin (approved):
lgtm, thanks koji!

PR #13230, opened 2023-08-04
feat(shared-data): Add the nozzle map lookup nozzle position for partial tip support

Laura-Danielle:
# Overview

Now we can directly look up the exact location of a nozzle (relative to a mount) for every pipette model. All of these maps were generated using a script, yay!

# Risk assessment

None. Not being used anywhere and the json is still valid.


sfoster1 (approved):
Looks good to me, nice!

PR #13231, opened 2023-08-04
feat(shared-data): add a current scale for partial tip configurations

Laura-Danielle:
## Overview

Once we have concrete data for a current scale on different pipette models, we can add those values to the configuration file. For now, we can load everything in with a default value.

RLIQ-423

# Risk assessment

None. Un-used parameter right now.

### DIFF
--- a/shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
+++ b/shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
@@ -174,6 +174,11 @@ class PartialTipDefinition(BaseModel):
         description="A list of the types of partial tip configurations supported, listed by channel ints",
         alias="availableConfigurations",
     )
+    current_scale: float = Field(
+        default=0.5,
+        description="A current scale for pick up tip in a parital tip configuration",
+        alias="currentScale",
+    )
 
 
 class PipettePhysicalPropertiesDefinition(BaseModel):
--- a/shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
+++ b/shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
@@ -174,10 +174,10 @@ class PartialTipDefinition(BaseModel):
         description="A list of the types of partial tip configurations supported, listed by channel ints",
         alias="availableConfigurations",
     )
-    current_scale: float = Field(
+    per_tip_pickup_current: float = Field(
         default=0.5,
-        description="A current scale for pick up tip in a parital tip configuration",
-        alias="currentScale",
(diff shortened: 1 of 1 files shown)


shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -174,6 +174,11 @@ class PartialTipDefinition(BaseModel):
         description="A list of the types of partial tip configurations supported, listed by channel ints",
         alias="availableConfigurations",
     )
+    current_scale: float = Field(
+        default=0.5,
+        description="A current scale for pick up tip in a parital tip configuration",
```
sfoster1:
```suggestion
        description="A current scale for pick up tip in a partial tip configuration",
```

sfoster1 (changes requested):
What's the intended use of this? That a partial tip pickup of N tips on an M channel pipette would set current `N/M * current_scale * tip_pickup_current`? Or something else? If that's what it is, 0.5 seems like a weird default.

We should also at least establish these values in the schema and definitions right, even if we have to change them later?

Laura-Danielle:
> What's the intended use of this? That a partial tip pickup of N tips on an M channel pipette would set current `N/M * current_scale * tip_pickup_current`? Or something else? If that's what it is, 0.5 seems like a weird default.
> 
> We should also at least establish these values in the schema and definitions right, even if we have to change them later?

@sfoster1, no it's intended to be => 0.5 * (N % 8) where N is number of tips. And yeah I guess we can add it to the definition. I was mostly trying to avoid merge conflicts between this and the other branch (+ we don't have exact numbers yet).

I could go either way adding them to the definition now or not. I just wanted to make sure this value was available while working on the tip manager class.

sfoster1:
> > What's the intended use of this? That a partial tip pickup of N tips on an M channel pipette would set current `N/M * current_scale * tip_pickup_current`? Or something else? If that's what it is, 0.5 seems like a weird default.
> > We should also at least establish these values in the schema and definitions right, even if we have to change them later?
> 
> @sfoster1, no it's intended to be => 0.5 * (N % 8) where N is number of tips. And yeah I guess we can add it to the definition. I was mostly trying to avoid merge conflicts between this and the other branch (+ we don't have exact numbers yet).
> 
> I could go either way adding them to the definition now or not. I just wanted to make sure this value was available while working on the tip manager class.

Oh can we call it something else then? like `perTipPickupCurrent` or something?

I'd really prefer if we added the value to the configs before merging this.

sfoster1 (approved):
cool!

PR #13232, opened 2023-08-04
fix(api, hardware): Only monitor overpressure if plunger axis is actually moving

fsinapi:
# Overview

The OT3Controller is configured monitor overpressure on a pipette if it's in a move group's set of nodes. Unfortunately, because the move group generator creates zero-velocity move groups for present nodes that are not moving in a group, the controller would monitor overpressure on every call to `move` when pipettes are attached to the system.

This posed a problem in some internal testing cases where the pipettes would mash tips against a plate and throw an overpressure error. Once the gantry tries to home the Z axes, it fails yet again due to an overpressure even though it should be ignored.

This PR adds filtering to be able to discern whether a node in a move group is actually moving, and we only monitor overpressure if the plunger is actually moving.

The OT3Controller home function didn't need any changes because it already filters based on the axes passed in directly.

# Test Plan

Compared homing with two pipettes attached, with the gantry in the middle of the deck and encoders valid so the API would call `move` to get the axes near the home switches.

Before this change, there were `bind_sensor_output_request` messages when the X and Y axes were commanded to move to the "safe-home" position. After this PR, the `bind_sensor_output_request` messages only occur when homing the plunger axes.

# Changelog

- Add member function to check if a move group step is nonmoving
- Add function to get all moving nodes in a move group
- When setting up overpressure monitoring, only monitor based on _moving_ nodes

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -44,6 +44,7 @@ from .ot3utils import (
     motor_nodes,
     LIMIT_SWITCH_OVERTRAVEL_DISTANCE,
     map_pipette_type_to_sensor_id,
+    moving_axes_in_move_group,
 )
 
 try:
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Okay, looks good to me, thank you!

PR #13233, opened 2023-08-04
feat(engine): use grip force and height from definitions in pe

sanni-t:
Closes RSS-279

# Overview

Updates the labware movement code in PE to use the grip height & force specified by the labware definition, or fallback to default if those are not specified.

# Test Plan

(Currently testing with @andySigler )
- [ ] test moving to/from deck all specified labware with & without liquid on a robot
- [ ] test moving a non-specified labware and verify it falls back to default force & height values
- [ ] test moving to/from modules all specified labware supported on the modules
- [ ] make sure the force specified in definitions is actually being set (use can monitor to check the duty cycle being set for gripper jaw)

# Changelog

- updated the gripper grip point getter to use `gripHeightFromLabwareBottom` property if present in definition, otherwise use default we've been using so far
- updated labware movement code to use `gripForce` specified in definition if present, otherwise fallback to the default force we've been using so far.

# Review requests

- make sure the code changes and calculations are correct (esp movement of labware on labware/adapter)

# Risk assessment

Medium. Code is pretty isolated and well-tested so it should not cause any issues as long as tests are passing.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/labware.py
+++ b/api/src/opentrons/protocol_engine/state/labware.py
@@ -767,9 +767,16 @@ class LabwareView(HasState[LabwareState]):
 
     def get_grip_force(self, labware_id: str) -> float:
         """Get the recommended grip force for gripping labware using gripper."""
-        return self.get_definition(labware_id).gripForce or LABWARE_GRIP_FORCE
+        recommended_force = self.get_definition(labware_id).gripForce
+        return (
+            recommended_force if recommended_force is not None else LABWARE_GRIP_FORCE
+        )
 
     def get_grip_height_from_labware_bottom(self, labware_id: str) -> float:
         """Get the recommended grip height from labware bottom, if present."""
         recommended_height = self.get_definition(labware_id).gripHeightFromLabwareBottom
-        return recommended_height or self.get_dimensions(labware_id).z / 2
+        return (
+            recommended_height
+            if recommended_height is not None
+            else self.get_dimensions(labware_id).z / 2
+        )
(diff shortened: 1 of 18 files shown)


api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -763,3 +764,12 @@ def get_labware_gripper_offsets(
...
         )
+
+    def get_grip_force(self, labware_id: str) -> float:
+        """Get the recommended grip force for gripping labware using gripper."""
+        return self.get_definition(labware_id).gripForce or LABWARE_GRIP_FORCE
+
+    def get_grip_height_from_labware_bottom(self, labware_id: str) -> float:
+        """Get the recommended grip height from labware bottom, if present."""
+        recommended_height = self.get_definition(labware_id).gripHeightFromLabwareBottom
+        return recommended_height or self.get_dimensions(labware_id).z / 2
```
SyntaxColoring: I think this is probably fine in practice because nobody will do this, but beware that this will do the wrong thing if someone does `"gripForce": 0` or `"gripHeightFromLabwareBottom": 0` for some reason.

SyntaxColoring (approved):
Code looks good to me! To answer your review question, though, I'm not sure where the stacking offset from an adapter between the deck and target labware is being applied, in the code.

api/src/opentrons/protocol_engine/state/labware.py
sanni-t: Good point. Although it really doesn't make sense to have a force or height of 0. But it makes sense to correct this.

PR #13234, opened 2023-08-04
refactor(app,app-shell,app-shell-odd): support updating both ot2 and flex

sfoster1:
The current app stack supports only one kind of update file and path. In internal-releases, we reuse that code for flex by pointing it at flex update media and flex robots... which just means the internal-release app can't update OT-2s with the automatically-distributed releases.

This PR changes that by
- Downloading both ot-2 and flex update files, separately
- Keeping them around separately (in the app data dir)
- Keeping track of them in redux separately
   - Including changing selectors to require specifying a robot name so we can get the right kind of update
- Pulling what kind of robot is getting updated from state when you start to update, and using the code paths previously only taken by user file robot updates (with the browse button) for everything

**Note**: This PR changes the download location for robot updates to be targeted on new infrastructure for flex launch. Specifically,
- internal-release robot software at `ot3-development.builds.opentrons.com` (sorry about the url, it's old). Flex builds can be seen here: https://ot3-development.builds.opentrons.com/ot3-oe/releases.json and OT-2 builds can be seen here: https://ot3-development.builds.opentrons.com/ot2-br/releases.json
- release robot software at `builds.opentrons.com`. This infra isn't up yet, but we can still merge this if we want

Also, I renamed all the redux and app-shell stuff from buildroot to robot-update. Sorry. It's contained in a single commit that you can skip when reviewing. The view layer is all the same still.

## Review Requests
- I'm not a very frequent typescript dev so please point out stuff that I'm doing wrong
- Pay particularly close attention to the epics in `app/src/redux/robot-update` and see if I'm doing things write; in general, as with a lot of our old react/redux code, the behavior of the actual update (as separate from displaying the right screens and such) is all driven via redux so this is pretty important
- Like I mentioned above, one thing I did here is rename all the redux and app-shell stuff from `buildroot` to `robot-update`, which touches Like A Billion files. However, I did keep it in one commit, so browse the pr commit by commit and you'll be able to skip it.

## Testing
This has to be tested... a lot. And I haven't, which is why this just a draft. The testing looks like this:
- [x] Make sure we automatically download both ot-2 and flex updates from the internet on desktop internal-release from internal-release
- [x] Ditto ditto ditto release
- [x] Make sure we automatically download flex updates on ODD internal release
- [x] Ditto ditto ditto release
- [x] Make sure we select the right files in redux on the desktop for flex
- [x] Ditto ditto ditto ot-2
- ~[ ] Ditto ditto ditto balena (?) We can maybe drop this~
- [x] Once all that stuff is good go through the upgrade/downgrade/reinstall logic

## Other notes


app-shell-odd/src/config/migrate.ts
```diff
@@ -62,8 +59,8 @@ export const DEFAULTS_V12: ConfigV12 = {
   protocols: { sendAllProtocolsToOT3: false, protocolsStoredSortKey: null },
   robotSystemUpdate: {
     manifestUrls: {
-      OT2: OT2_MANIFEST_URL,
-      OT3: OT3_MANIFEST_URL,
+      OT2: 'dummy-ot2-manifest',
+      OT3: 'dummy-ot3-manifest',
```
shlokamin, line 63: i think its fine youre changing this beacuse we havent publicly released the app shell odd but flagging that weird things can happen when you make changes to prev config defaults outside of a migration

shlokamin: but also is this even necessary since `robotSystemUpdate` gets scrapped in `ConfigV18`?

app-shell-odd/src/system-update/index.ts
```diff
@@ -26,6 +27,24 @@ const log = createLogger('systemUpdate/index')
...
+      payload: {
+        systemFile: fileInfo.systemFile,
+        version: fileInfo.versionInfo.opentrons_api_version,
+      },
+    }))
+    .catch((error: Error) => ({
+      type: 'robotUpdate:UNEXPECTED_ERROR' as const,
+      payload: { message: error.message },
+    }))
+    .then(dispatch)
```
shlokamin, line 46: lawl sick implicit fn call

app-shell-odd/src/system-update/update.ts
```diff
@@ -8,7 +8,7 @@ const OT2_FILENAME = 'ot2-system.zip'
 const SYSTEM_FILENAME = 'system-update.zip'
 
 const getSystemFileName = (robotModel: RobotModel): string => {
-  if (robotModel === 'OT-2 Standard') {
+  if (robotModel === 'OT-2 Standard' || robotModel === null) {
```
shlokamin, line 11: uh can `robotModel` ever be `null`?

app-shell-odd/src/update.ts
```diff
@@ -13,6 +13,10 @@ import type { ReleaseSetUrls } from './system-update/types'
 
 const log = createLogger('update')
 
+export const FLEX_MANIFEST_URL = _OPENTRONS_PROJECT_.includes('robot-stack')
+  ? 'https://builds.opentrons.com/ot3-oe/releases.json'
+  : 'https://ot3-development.builds.opentrons.com/ot3-oe/releases.json'
+
```
shlokamin, line 19: i know u already called out that https://builds.opentrons.com/ot3-oe/releases.json isn't up yet but the app-shell-odd's make file has `OPENTRONS_PROJECT ?= robot-stack` so won't this put the odd's update flow in a broken state?

shlokamin (commented):
thanks for doing this! lgtm, gonna test this out on a bot now

app-shell-odd/src/config/migrate.ts
sfoster1: It's necessary because those migrations run in order so you need to have values there, but i want to remove the const values. I could just write in the values of the consts there I guess?

app-shell-odd/src/system-update/update.ts
sfoster1: it isn't typed that way now that i look at it but it for sure can be, really old ot-2's don't have that in their update server responses

app-shell-odd/src/update.ts
sfoster1: Only for pushed odd's - the oe core build workflow [specifies](https://github.com/Opentrons/oe-core/blob/b376966e3840bf9b1520e98ad7f2f9dc204faeb2/.github/workflows/build-ot3-actions.yml#L109) ot-3

shlokamin:
just synced irl: 

for the ot-3 project it looks like the correct artifacts get downloaded, but for the robot-stack things don't work because the release infra isn't up yet. we're gonna wait to merge till the release infra for the robot stack is up

shlokamin (approved):
i didn't get to test file uploads on the flex but i think i went through everything else and it... works? wow.

sfoster1:
> i didn't get to test file uploads on the flex but i think i went through everything else and it... works? wow.

Ye of little faith!

PR #13236, opened 2023-08-04
fix(app): generify update robot message

shlokamin:
# Overview
When updating a flex via the desktop app we still show "OT-2" as the robot name. This PR uses "robot" instead to make it generic.
# Risk assessment
Low


sfoster1 (approved):
looks good to me

PR #13237, opened 2023-08-04
feat(protocol-designer,shared-data,components,app): adapter split support

jerader:
closes RAUT-609 and RAUT-612
closes [RAUT-561](https://opentrons.atlassian.net/browse/RAUT-561) and [RQA-1076](https://opentrons.atlassian.net/browse/RQA-1076) and [RQA-1114](https://opentrons.atlassian.net/browse/RQA-1114)

# Overview

Adapter Split work in PD and App

Note: I will add more test coverage in a follow up pr

# Test Plan

For Pd:
sandbox: https://sandbox.designer.opentrons.com/pd_adapter-split/

1. import an older version PD protocol that has an adapter (one with H-s would work) and make sure that it migrates and exports successfully with the correct `loadLabware` commands
2. make a new protocol and add a H-S or Temperature module, make sure you can load the adapter + labware on top
3. mess around with the `moveLabware` form, move labware on/off the adapters and modules and make sure it properly errors if you try to move a labware onto an adapter/module that doesn't work with that labware. Also make sure the `movelabware` form allows you to select tipracks and not select the trash.

For the app:
1. upload a protocol that has an adapter, make sure the deck thumbnail loads the adapter + liquid information correctly. Check Protocol details in the labware and liquids tab. 
2. go to protocol setup and make sure labware setup, liquid setup list and map view show up correctly
3. go through LPC, should work as expected
4. run a protocol, look at the command text, it should display the adapter and labware on top of adapters text correctly
5. look through protocol and labware setup on the odd, map and list should display correctly

# Changelog

PD:
- add to `7_0_0` migration and the `fileCreator` to accommodate load adapter
- separate out the adapters in the deck setup to their own AdapterControls layer and change labware restrictions for H-S and Temperature module to allow for acceptable adapters
- add some more restrictions to moveLabware

Shared-data:
- added type `NonStackedLabware` and extended the `LabwareLocation` type to include `labwareId`.

App:
- added a bunch of logic to accommodate for if the labwareLocation has `labwareId` which included all the deck map renders in the app and odd (deckThumbnail, protocol setup, odd protocol setup) and the command text.

# Review requests

see test plan

# Risk assessment

low

[RAUT-561]: https://opentrons.atlassian.net/browse/RAUT-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RQA-1076]: https://opentrons.atlassian.net/browse/RQA-1076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RQA-1114]: https://opentrons.atlassian.net/browse/RQA-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared-data/js/getLabware.ts
```diff
@@ -62,6 +62,8 @@ export const LABWAREV2_DO_NOT_LIST = [
 export const PD_DO_NOT_LIST = [
   'opentrons_calibrationblock_short_side_left',
   'opentrons_calibrationblock_short_side_right',
+  'opentrons_96_aluminumblock_biorad_wellplate_200ul',
+  'opentrons_96_aluminumblock_nest_wellplate_100ul',
```
jerader, line 66: Maybe i should filter out these old aluminum block defs more locally and not here? We basically don't want them selectable at all from the `aluminumBlock` category

jerader:
Components that need test coverage from this. It will likely be completed in a follow up PR:

`getLabwareLocationCombos.test.ts`
`LabwareListItem.test.tsx`
Create tests for` parseInitialLoadedLabwareByAdapter`, `parseInitialLoadAdapterBySlot`, `adapterControls` and `getAdapterAndLabwareSplitInfo.ts`
`getLabwareOffsetLocation.test.tsx`
`getLabwareRenderInfo.test.ts`
`getProtocolModulesInfo.test.ts`
`getSlotLabwareName.test.ts`
`CheckItem.test.tsx`
`ProtocolSetupLabware.test.tsx`
`LabwareSelectionModal.test.tsx`
`createSnippet.test.tsx`

ticket for the follow up: https://opentrons.atlassian.net/browse/RAUT-624

protocol-designer/src/components/StepEditForm/fields/LabwareLocationField/index.tsx
```diff
@@ -16,7 +17,19 @@ export function LabwareLocationField(
...
+  const bothFieldsSelected = props.labware != null && props.value != null
+
   return (
-    <StepFormDropdown {...props} options={unoccupiedLabwareLocationsOptions} />
+    <StepFormDropdown
+      {...props}
+      errorToShow={
+        !props.canSave && bothFieldsSelected
+          ? i18n.t(
+              'form.step_edit_form.labwareLabel.errors.labwareSlotIncompatible'
```
jerader, line 28: I will specify the labware and the new location in this string in a follow up ticket as well as specifying the adapter name in the dropdown.

https://opentrons.atlassian.net/browse/RAUT-621

app/src/assets/localization/en/labware_position_check.json
```diff
@@ -1,4 +1,6 @@
 {
+  "adapter_in_mod_in_slot": "adapter in {{module}} in {slot}}",
+  "adapter_in_slot": "adapter in {{slot}}",
```
b-cooper: Should the "adapter" here be an interpolated display name?

app/src/molecules/PythonLabwareOffsetSnippet/createSnippet.ts
```diff
@@ -53,7 +54,11 @@ export function createSnippet(
         loadStatement = `labware_${labwareCount} = ${moduleVariable}.load_labware("${String(
           loadName
         )}")`
+        //  load labware on adapter
+      } else if ('labwareId' in command.params.location) {
+        loadStatement = `labware_${labwareCount} = protocol.load_labware("${loadName}}, location="${command.params.location.labwareId}")`
       }
+
```
b-cooper: did you test this addition in a python interpreter?  this doesn't look like valid syntax to me. Looks like an extra closing brace where a double quote should be

jerader: oops good catch, will fix this!

app/src/assets/localization/en/labware_position_check.json
jerader: yeah probably, i think i was being lazy and didn't add it 😅 , i can add that though!

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
```diff
@@ -69,61 +72,88 @@ export function SetupLabwareMap({
...
+                    const labwareIdToUse =
+                      labwareInAdapterInMod?.result?.labwareId ??
+                      nestedLabwareId
+                    const displayNameToUse =
+                      labwareInAdapterInMod?.result?.definition.metadata
+                        .displayName ?? nestedLabwareDisplayName
+
+                    return (
+                      <Module
+                        key={`LabwareSetup_Module_${String(
```
b-cooper: this key should just be the module's `id`

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
```diff
@@ -69,61 +72,88 @@ export function SetupLabwareMap({
...
+                          moduleDef.model === THERMOCYCLER_MODULE_V1
+                            ? { lidMotorState: 'open' }
+                            : {}
+                        }
+                      >
+                        {definition != null &&
+                        displayNameToUse != null &&
+                        labwareIdToUse != null ? (
+                          <React.Fragment
+                            key={`LabwareSetup_Labware_${displayNameToUse}_${x}${y}`}
```
b-cooper: this key could just be the labware `id`

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
```diff
@@ -69,61 +72,88 @@ export function SetupLabwareMap({
...
-                        </React.Fragment>
-                      ) : null}
-                    </Module>
-                  )
+                  }) => {
+                    const labwareInAdapterInMod =
+                      nestedLabwareId != null
+                        ? initialLoadedLabwareByAdapter[nestedLabwareId]
+                        : null
+                    const definition =
```
b-cooper: It might be helpful to note here in a comment that we're not rendering the adapter on the module at all, if there happens to be a labware on top of it.  This isn't really an issue now, because the adapter will always be occluded anyway, but in the future it could be confusing if we're expecting to see some of the adapter peeking out the sides.

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
```diff
@@ -69,61 +72,88 @@ export function SetupLabwareMap({
...
+                      labwareInAdapter?.result?.definition ?? labwareDef
+                    const labwareIdToUse =
+                      labwareInAdapter?.result?.labwareId ?? labwareId
+                    const displayNameToUse =
+                      labwareInAdapter?.result?.definition.metadata
+                        .displayName ?? displayName
+
                     return (
                       <React.Fragment
                         key={`LabwareSetup_Labware_${String(
```
b-cooper: labwareIdTo would be a fine key here. Also the same comment goes for this labware. We're not rendering the adapters at all if they're under a labware.  Fine for now, but may be confusing in the future.

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
```diff
@@ -132,6 +162,7 @@ export function SetupLabwareMap({
         <OffDeckLabwareList
           labwareItems={offDeckItems}
           isOt3={robotType === 'OT-3 Standard'}
```
b-cooper: We've got a constant for this now `FLEX_ROBOT_TYPE` should be used

app/src/organisms/Devices/ProtocolRun/SetupLiquids/SetupLiquidsMap.tsx
```diff
@@ -101,74 +105,131 @@ export function SetupLiquidsMap(props: SetupLiquidsMapProps): JSX.Element {
...
+                    ? initialLoadedLabwareByAdapter[nestedLabwareId]
+                    : null
+                const definition =
+                  labwareInAdapterInMod?.result?.definition ?? nestedLabwareDef
+                const labwareIdToUse =
+                  labwareInAdapterInMod?.result?.labwareId ?? nestedLabwareId
+                const displayNameToUse =
+                  labwareInAdapterInMod?.result?.definition.metadata
+                    .displayName ?? nestedLabwareDisplayName
+                let id: string = ''
```
b-cooper: nit: `id` is a very vague variable name given the number of id'ed entities we're working with here. maybe `labwareId`. Also how is this id different than the `labwareIdToUse` above?

app/src/organisms/Devices/ProtocolRun/SetupLiquids/SetupLiquidsMap.tsx
```diff
@@ -101,74 +105,131 @@ export function SetupLiquidsMap(props: SetupLiquidsMapProps): JSX.Element {
...
+                      moduleDef.model === THERMOCYCLER_MODULE_V1
+                        ? { lidMotorState: 'open' }
+                        : {}
+                    }
+                  >
+                    {definition != null &&
+                    displayNameToUse != null &&
+                    labwareIdToUse != null ? (
+                      <React.Fragment
+                        key={`LabwareSetup_Labware_${displayNameToUse}_${x}${y}`}
```
b-cooper: key can just be the `id`

app/src/organisms/Devices/ProtocolRun/SetupLiquids/SetupLiquidsMap.tsx
```diff
@@ -101,74 +105,131 @@ export function SetupLiquidsMap(props: SetupLiquidsMapProps): JSX.Element {
...
+                const labwareInAdapter =
+                  initialLoadedLabwareByAdapter[labwareId]
+                const definition =
+                  labwareInAdapter?.result?.definition ?? labwareDef
+                const labwareIdToUse =
+                  labwareInAdapter?.result?.labwareId ?? labwareId
+                const displayNameToUse =
+                  labwareInAdapter?.result?.definition.metadata.displayName ??
+                  displayName
+                let id: string = ''
```
b-cooper: Same comment, it seems like `id` and `labwareIdToUse` are doing the same thing and only one should exist

app/src/organisms/Devices/ProtocolRun/SetupLiquids/SetupLiquidsMap.tsx
```diff
@@ -101,74 +105,131 @@ export function SetupLiquidsMap(props: SetupLiquidsMapProps): JSX.Element {
...
             {map(
               labwareRenderInfoById,
               ({ x, y, labwareDef, displayName }, labwareId) => {
+                const labwareInAdapter =
+                  initialLoadedLabwareByAdapter[labwareId]
+                const definition =
+                  labwareInAdapter?.result?.definition ?? labwareDef
+                const labwareIdToUse =
+                  labwareInAdapter?.result?.labwareId ?? labwareId
+                const displayNameToUse =
```
b-cooper: This `fooBarToUse` naming convention is a vague. It's the main reason I suggested adding a comment in the DeckThumbnail component explaining that we don't render any adapter that has a labware on top of it.  This will be especially useful when we stack labware. The thing we're actually calculating in each of these instances is the "top-most labware" (adapter or otherwise) and then only rendering that.  Maybe that variable name would make this fall-back clearer without an explanatory comment in each component?

app/src/organisms/Devices/ProtocolRun/SetupLiquids/SetupLiquidsMap.tsx
jerader: oops good catch, it is the same as `labwareIdToUse`

app/src/molecules/PythonLabwareOffsetSnippet/createSnippet.ts
```diff
@@ -53,7 +53,11 @@ export function createSnippet(
         loadStatement = `labware_${labwareCount} = ${moduleVariable}.load_labware("${String(
           loadName
         )}")`
+        //  load labware on adapter
+      } else if ('labwareId' in command.params.location) {
+        loadStatement = `labware_${labwareCount} = protocol.load_labware("${loadName}}", location="${command.params.location.labwareId}")`
```
b-cooper: same question I asked about in this file before.  `protocol.load_labware("${loadName}}"` doesn't look like it will resolve to a valid python string. It looks like an extra curly bracket. Try running the output in a python protocol through analysis to confirm.

api-client/src/runs/types.ts
```diff
@@ -99,6 +99,7 @@ export interface CreateRunActionData {
 export interface LabwareOffsetLocation {
   slotName: string
   moduleModel?: ModuleModel
+  labwareId?: string
```
b-cooper: This is not correct. (`labwareId` -> `definitionUri`) The matching python model can be seen here opentrons/api/src/opentrons/protocol_engine/types.py 

This type update will require a number of changes to the staged LPC changes here. Let me know if you want to talk through any of it.

app/src/molecules/PythonLabwareOffsetSnippet/createSnippet.ts
jerader: sorry i def fixed it but then didn't check carefully when i removed the `loadAdapters` 🤦 i think it is correct now

app/src/molecules/PythonLabwareOffsetSnippet/createSnippet.ts
```diff
@@ -53,7 +53,11 @@ export function createSnippet(
         loadStatement = `labware_${labwareCount} = ${moduleVariable}.load_labware("${String(
           loadName
         )}")`
+        //  load labware on adapter
+      } else if ('labwareId' in command.params.location) {
+        loadStatement = `labware_${labwareCount} = protocol.load_labware("${loadName}", location="${command.params.location.labwareId}")`
```
b-cooper: As written, this won't work.  This shouldn't be the `labwareId`. It needs to be the python variable that we're storing the LabwareContext of the loaded adapter onto. We're naming those variables deterministically based on the labware count right now.

Just checked with Sanniti, and there is some backend work we're going to need to do anyway to fix up this implementation. If the fix isn't clear after a quick spike, please just add a large TODO to the top of the file outlining that this function is currently broken and will need to be fixed as soon as RSS has a story for setting offsets for Jupyter and CLI.

app/src/organisms/ApplyHistoricOffsets/hooks/getLabwareLocationCombos.ts
```diff
@@ -38,6 +38,24 @@ export function getLabwareLocationCombos(
               labwareId: command.result.labwareId,
               moduleId,
             })
+      } else if ('labwareId' in command.params.location) {
+        const { labwareId } = command.params.location
+        const adapterLocation = resolveAdapterLocation(
```
b-cooper: This `else if` will not function properly as written. It's converting specific `labwareId` and `moduleId` instances into a `LabwareOffsetLocation` (which is defined generically e.g. defURI, moduleModel, slotName), and then converting back to a specific module in line 48 via the `Object.values(modules).find`. If there are multiple modules of the same type the adapter offset will not be applied to the correct module location.

jerader: i see what you mean, ya this doesn't work for MoaM

protocol-designer/src/utils/labwareModuleCompatibility.ts
```diff
@@ -78,9 +76,79 @@ export const getLabwareIsCompatible = (
...
   labwareOnDeck: LabwareOnDeck
 ): boolean => {
   return labwareOnDeck.labwareDefURI in customLabwares
 }
+
+export const getAdapterLabwareIsAMatch = (
+  labwareId: string,
+  draggedLabwareLoadname: string
+): boolean => {
+  const labwareDefUri = labwareId.split(':')[1]
```
b-cooper: Parsing the `labwareId` or any `entityId` directly is not a safe thing to do. We shouldn't be making any assumptions about the contents of any Protocol Engine `id`'s other than the fact that they are unique.  There ~6 instances of this across the PR.  I'll drop a comment on them to make sure they're covered.  To get the labwareDefURI, you should dereference the `labwareId` using the `labware` top-level entities array.

app/src/organisms/ApplyHistoricOffsets/hooks/getLabwareLocationCombos.ts
```diff
@@ -110,3 +141,52 @@ function resolveModuleLocation(
...
+    const moduleId = labwareEntity.location.moduleId
+    const resolvedModuleLocation: LabwareOffsetLocation | null = resolveModuleLocation(
+      modules,
+      moduleId
+    )
+    moduleIdUnderAdapter = moduleId
+    adapterOffsetLocation =
+      resolvedModuleLocation != null
+        ? {
+            definitionUri: labwareId.split(':')[1],
```
b-cooper: Let's update this so we're not doing string parsing on an Id

app/src/organisms/ApplyHistoricOffsets/hooks/getLabwareLocationCombos.ts
```diff
@@ -110,3 +141,52 @@ function resolveModuleLocation(
...
+            ...resolvedModuleLocation,
+          }
+        : null
+  } else {
+    adapterOffsetLocation = {
+      definitionUri: labwareId.split(':')[1],
```
b-cooper: Let's update this so we're not doing string parsing on an Id

protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -50,10 +74,91 @@ export const migrateFile = (
...
+        getIsAdapter(command.params.labwareId)
+    )
+    .flatMap(command => {
+      const {
+        adapterUri,
+        labwareUri,
+        adapterDisplayName,
+        labwareDisplayName,
+      } = getAdapterAndLabwareSplitInfo(command.params.labwareId)
+      const previousLabwareIdUuid = command.params.labwareId.split(':')[0]
```
b-cooper, line 103: Let's update this so we're not doing string parsing on an Id

protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -80,18 +186,96 @@ export const migrateFile = (
...
+  }, {})
+
+  const getNewLabwareIngreds = (
+    ingredLocations?: DesignerApplicationData['ingredLocations']
+  ): DesignerApplicationData['ingredLocations'] => {
+    const updatedIngredLocations: DesignerApplicationData['ingredLocations'] = {}
+    if (ingredLocations == null) return {}
+    for (const [labwareId, wellData] of Object.entries(ingredLocations)) {
+      if (getIsAdapter(labwareId)) {
+        const labwareIdUuid = labwareId.split(':')[0]
```
b-cooper, line 240: Let's update this so we're not doing string parsing on an Id

protocol-designer/src/steplist/formLevel/profileErrors.ts
```diff
@@ -63,3 +71,56 @@ export const getProfileFormErrors = (
...
+    const modValueDefUri = newLocation.moduleId.split(':')[1] as ModuleType
+    const modAllowList =
+      COMPATIBLE_LABWARE_ALLOWLIST_BY_MODULE_TYPE[modValueDefUri]
+    errorString = !modAllowList.includes(loadName)
+      ? i18n.t(
+          'form.step_edit_form.labwareLabel.errors.labwareIncompatibleWithMod'
+        )
+      : null
+  } else if ('labwareId' in newLocation) {
+    const adapterValueDefUri = newLocation.labwareId.split(':')[1]
```
b-cooper: Let's update this so we're not doing string parsing on an Id

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -202,6 +202,44 @@ export const createFile: Selector<ProtocolFile> = createSelector(
...
+        acc,
+        labware: typeof initialRobotState.labware[keyof typeof initialRobotState.labware],
+        labwareId: string
+      ): LoadLabwareCreateCommand[] => {
+        const { labwareDefURI, def } = labwareEntities[labwareId]
+        const isAdapter = def.metadata.displayCategory === 'adapter'
+        if (!isAdapter) return [...acc]
+        const isOnTopOfModule = labware.slot in initialRobotState.modules
+        const namespace = def.namespace
+        const loadName = labwareDefURI.split('/')[1].replace(/\/1$/, '')
```
b-cooper: This is another instance of unsafe string parsing. Shouldn't we just be getting the loadName from the def?

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -213,13 +251,18 @@ export const createFile: Selector<ProtocolFile> = createSelector(
...
         const { labwareDefURI, def } = labwareEntities[labwareId]
+        const isAdapter = def.metadata.displayCategory === 'adapter'
+        if (labwareId === FIXED_TRASH_ID || isAdapter) return [...acc]
+        const isOnTopOfModule = labware.slot in initialRobotState.modules
+        const isOnAdapter =
+          loadAdapterCommands.find(
+            command => command.params.labwareId === labware.slot
+          ) != null
         const namespace = def.namespace
         const loadName = labwareDefURI.split('/')[1].replace(/\/1$/, '')
```
b-cooper: same comment as above, it seems like we should be getting the loadName from the def here.

protocol-designer/src/steplist/formLevel/profileErrors.ts
```diff
@@ -63,3 +71,56 @@ export const getProfileFormErrors = (
...
+  labware: LabwareEntity,
+  newLocation: LabwareLocation
+): string | null => {
+  let errorString: string | null = null
+  if (labware == null || newLocation == null || newLocation === 'offDeck')
+    return null
+  const selectedLabwareDefUri = labware?.labwareDefURI
+  if ('moduleId' in newLocation) {
+    const loadName = selectedLabwareDefUri.split('/')[1].split('/')[0]
+    const modValueDefUri = newLocation.moduleId.split(':')[1] as ModuleType
```
b-cooper: both of these lines include unsafe string parsing. The values should be dereferenced properly.

protocol-designer/src/utils/labwareModuleCompatibility.ts
jerader: ya way less brittle to do it that way, my bad. fixing the instances now! thanks for leaving comments on all of them.

protocol-designer/src/load-file/migration/7_0_0.ts
jerader: I am not sure how to update this one, my logic is to keep the previous labwareId's uuid and then update the defintiionUri and this is so that when the labware/adapter combo migrates to be labware and adapter separately, the liquids on the labware/adapter combo migrate to be liquids on the labware. Otherwise, how do we know which locations the liquids go?

b-cooper: Oh gotcha, that makes sense. Cool, let's keep it as is here then

shared-data/js/getLabware.ts
b-cooper: This is probably ok for now. As long as there's no change legacy protocols will need these definitions (assuming they get swapped out in migration)

app/src/organisms/Devices/ProtocolRun/SetupLabwarePositionCheck/index.tsx
```diff
@@ -77,7 +77,7 @@ export function SetupLabwarePositionCheck(
           }}
           id="LabwareSetup_checkLabwarePositionsButton"
           {...targetProps}
-          disabled={lpcDisabledReason !== null}
+          // disabled={lpcDisabledReason !== null}
```
b-cooper:
```suggestion
          disabled={lpcDisabledReason !== null}
```

app/src/organisms/Devices/ProtocolRun/utils/__tests__/getSlotLabwareName.test.ts
```diff
@@ -74,4 +74,5 @@ describe('getSlotLabwareName', () => {
       expected
     )
   })
+  //  TODO(jr, 8/16/23): add test cases for adapter and module name
```
b-cooper: you can use `it.todo(...)` to allow this message to be visible when the test suite is run

app/src/organisms/Devices/ProtocolRun/utils/getLabwareOffsetLocation.ts
```diff
@@ -26,6 +31,27 @@ export const getLabwareOffsetLocation = (
...
+      location = null
+    } else if (adapter.location === 'offDeck') {
+      location = null
+    } else if ('slotName' in adapter.location) {
+      location = {
+        slotName: adapter.location.slotName,
+        definitionUri: adapter.definitionUri,
+      }
+    } else if ('moduleId' in adapter.location) {
+      const adapterModLocation = adapter.location.moduleId
```
b-cooper: this variable is actually an `id` not a `location` right?

app/src/organisms/Devices/ProtocolRun/utils/getSlotLabwareName.ts
```diff
@@ -1,43 +1,113 @@
...
+} from '@opentrons/shared-data'
 
+export interface SlotInfo {
+  slotName: string
+  labwareName: string
+  adapterName?: string
+  moduleName?: string
+}
+
+//  TODO: (jr, 8/16/23): probably need to call this a new name and should slotName and labwareName be optional?
```
b-cooper, line 16: yeah maybe `getLocationInfo` is more accurate now

app/src/organisms/LabwarePositionCheck/CheckItem.tsx
```diff
@@ -68,6 +70,16 @@ export const CheckItem = (props: CheckItemProps): JSX.Element | null => {
   const pipette = protocolData.pipettes.find(
     pipette => pipette.id === pipetteId
   )
+  //  find labware on adapter's location
+  const adapterId =
+    Object.values(protocolData.labware).find(
```
b-cooper: `Object.values` isn't needed here, `protocolData.labware` is an array

b-cooper: Also, we shouldn't be finding an `adapterId` at all here. We're trying to get the display name of the adapter (which lives on the definition), and we have the defURI.  No reason to go back to a specific `adapterId`.  What you want here is only part of the LPC `getLabwareDef` util:

```
 const adapterDisplayName = getLabwareDefinitionsFromCommands(protocolData.commands).find(def => getLabwareDefURI(def) === labwareDefUri)?.metadata.displayName
```

app/src/organisms/Devices/ProtocolRun/utils/getLabwareOffsetLocation.ts
jerader: but the `id` is a location? i'm a bit confused

app/src/organisms/LabwarePositionCheck/CheckItem.tsx
jerader: i thought i need the `adapterId` to move it on/off the deck? and also to move the labware on/off it.

app/src/organisms/Devices/ProtocolRun/utils/getSlotLabwareName.ts
jerader: I think renaming this util and plugging the adapter + module names in to the components that use this util can be a separate PR - is it ok if i keep the util name the way it is?

jerader: here is a ticket i made to address this in a followup: https://opentrons.atlassian.net/browse/RAUT-628

jerader:
Follow up work post merge:
1. add test coverage https://opentrons.atlassian.net/browse/RAUT-624
2. refactor `offsetTable` to not include `protocolData` in `ResultSummary` https://opentrons.atlassian.net/browse/RAUT-624
3. refactor `ProtocolSetupLabware` - needs to be reorganized and split up into several components. There is copied logic that can be its own util. https://opentrons.atlassian.net/browse/RAUT-631
4. refactor `getSlotLabwareName` https://opentrons.atlassian.net/browse/RAUT-628

app/src/organisms/LabwarePositionCheck/utils/getDisplayLocation.ts
```diff
@@ -1,17 +1,51 @@
...
 import type { TFunction } from 'i18next'
 
 export function getDisplayLocation(
   location: LabwareOffsetLocation,
-  t: TFunction
+  labwareDefinitions: LabwareDefinition2[],
+  t: TFunction,
+  //  TODO: (jr, 8/18/23): perhaps find a new route here and refactor this,
+  //  only some instances of slot need to be capitalized depending on the parent component
+  capitalizeSlot?: boolean
```
b-cooper, line 18: this function already returns a string, can we just wrap the output of the function in a format call `i18n.format(getDisplayLocation(location, defs, t), 'capitalize')` instead of adding an optional param?

b-cooper: Also, I just checked with Ed, and he confirmed that having the word "Slot" be capitalized any time we render "Slot 3" or "Slot A3" is totally acceptable, so maybe the final return should just be wrapped with an i18n.format call?

jerader: for your 1st comment: we can't do that because the display location can include the adapter + module display names and capitalizing the string will only capitalize the 1st word.

2nd comment: perfect, i can either pass down i18n as a prop or keep the capitalize function from lodash? I think i prefer to keep the capitalize function 🤔

app/src/organisms/LabwarePositionCheck/ResultsSummary.tsx
```diff
@@ -222,10 +224,11 @@ const Header = styled.h1`
 interface OffsetTableProps {
   offsets: LabwareOffsetCreateData[]
   labwareDefinitions: LabwareDefinition2[]
+  protocolData: CompletedProtocolAnalysis
 }
 
 const OffsetTable = (props: OffsetTableProps): JSX.Element => {
-  const { offsets, labwareDefinitions } = props
+  const { offsets, labwareDefinitions, protocolData } = props
```
b-cooper, line 231: this `protocolData` prop looks unnecessary right? We already have the definitions to pass directly to `getDisplayLocation`. Should be able to remove this prop.

jerader: ya i marked that as a follow up action item but can clean it up here too.

protocol-designer/src/components/steplist/MoveLabwareHeader.tsx
```diff
@@ -15,6 +24,8 @@ interface MoveLabwareHeaderProps {
 //  TODO(jr, 7/31/23): add text to i18n
 export function MoveLabwareHeader(props: MoveLabwareHeaderProps): JSX.Element {
   const { sourceLabwareNickname, destinationSlot, useGripper } = props
+  const moduleEntites = useSelector(getModuleEntities)
```
b-cooper: `moduleEntites` and `labwareEntites` should be => `moduleEntities` and `labwareEntities`

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -202,6 +202,44 @@ export const createFile: Selector<ProtocolFile> = createSelector(
...
+      LoadLabwareCreateCommand[]
+    >(
+      initialRobotState.labware,
+      (
+        acc,
+        labware: typeof initialRobotState.labware[keyof typeof initialRobotState.labware],
+        labwareId: string
+      ): LoadLabwareCreateCommand[] => {
+        const { def } = labwareEntities[labwareId]
+        const isAdapter = def.metadata.displayCategory === 'adapter'
```
b-cooper: should use `allowedRoles` instead of `displayCategory` here to filter for adapter types

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -213,13 +251,18 @@ export const createFile: Selector<ProtocolFile> = createSelector(
         labware: typeof initialRobotState.labware[keyof typeof initialRobotState.labware],
         labwareId: string
       ): LoadLabwareCreateCommand[] => {
-        if (labwareId === FIXED_TRASH_ID) return [...acc]
-        const isLabwareOnTopOfModule = labware.slot in initialRobotState.modules
-        const { labwareDefURI, def } = labwareEntities[labwareId]
+        const { def } = labwareEntities[labwareId]
+        const isAdapter = def.metadata.displayCategory === 'adapter'
```
b-cooper: use `allowedRoles` here too

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -213,13 +251,18 @@ export const createFile: Selector<ProtocolFile> = createSelector(
...
+        if (labwareId === FIXED_TRASH_ID || isAdapter) return [...acc]
```
b-cooper:
```suggestion
        if (labwareId === FIXED_TRASH_ID || isAdapter) return acc
```

protocol-designer/src/file-data/selectors/fileCreator.ts
```diff
@@ -202,6 +202,44 @@ export const createFile: Selector<ProtocolFile> = createSelector(
...
+    >(
+      initialRobotState.labware,
+      (
+        acc,
+        labware: typeof initialRobotState.labware[keyof typeof initialRobotState.labware],
+        labwareId: string
+      ): LoadLabwareCreateCommand[] => {
+        const { def } = labwareEntities[labwareId]
+        const isAdapter = def.metadata.displayCategory === 'adapter'
+        if (!isAdapter) return [...acc]
```
b-cooper:
```suggestion
        if (!isAdapter) return acc
```

protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -14,15 +21,32 @@ import type {
...
-// params to the load commands.
+// params to the load commands. Also, this migrates previous combined
+//  adapter + labware commands to all labware commands and definitions to their commands/definitions split up
 const PD_VERSION = '7.0.0'
 const SCHEMA_VERSION = 7
+interface LabwareLocationUpdate {
+  [id: string]: string
+}
+
+//  might need better way to filter this???
```
b-cooper: yeah it would be better to get this out of the definition's load name instead of parsing it from the id

protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -50,10 +74,91 @@ export const migrateFile = (
...
+        adapterLocation = { moduleId: labwareLocation.moduleId }
+      } else if ('slotName' in labwareLocation) {
+        adapterLocation = { slotName: labwareLocation.slotName }
+      }
+      const defUris = Object.keys(allLatestDefs)
+      const adapterDefUri = defUris.find(defUri => defUri === adapterUri) ?? ''
+      const labwareDefUri = defUris.find(defUri => defUri === labwareUri) ?? ''
+      const adapterLoadname = allLatestDefs[adapterDefUri].parameters.loadName
+      const labwareLoadname = allLatestDefs[labwareDefUri].parameters.loadName
+      const adapterId = `${uuid()}:${adapterUri}`
```
b-cooper, line 118: what's the difference between `adapterUri` and `adapterDefUri` here? same for `labwareUri` and `labwareDefUri`

protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -80,18 +186,96 @@ export const migrateFile = (
...
+            locationString = location.slotName
+          } else if ('labwareId' in location) {
+            locationString = location.labwareId
+          }
+          adapterAndLabwareAcc[labId] = locationString
+          return adapterAndLabwareAcc
+        },
+        {}
+      )
+      Object.assign(acc, adapterAndLabwareLocationUpdate)
```
b-cooper: spreading here would be more consistent with reduces elsewhere in the codebase e.g. `[...acc, ...adapterAndLabwareLocationUpdate]`

protocol-designer/src/step-forms/selectors/index.ts
```diff
@@ -524,6 +525,12 @@ const _dynamicFieldFormErrors = (
   return getProfileFormErrors(hydratedForm)
 }
 
+const _dynamicMoveLabwareFieldFormErrors = (
+  hydratedForm: FormData,
+  invariantContext: InvariantContext
+): ProfileFormError[] => {
```
b-cooper, line 531: what does `Profile` have to do with this error?  Isn't that only relevant to TC profile steps?

protocol-designer/src/steplist/fieldLevel/index.ts
```diff
@@ -53,6 +53,9 @@ const getLabwareEntity = (
   return state.labwareEntities[id] || null
 }
 
+const getIsAdapter = (id: string): boolean =>
```
b-cooper: this function needs to be reading from `allowedRoles` instead of parsing id's

protocol-designer/src/steplist/formLevel/profileErrors.ts
```diff
@@ -63,3 +71,63 @@ export const getProfileFormErrors = (
   // NOTE: since errors stacking doesn't seem to serve a purpose, remove repeats
   return uniqBy(errors, error => error.title)
 }
+
+const getMoveLabwareError = (
```
b-cooper: is this function in the wrong file? profile is a TC concept

protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -42,14 +42,21 @@ export const getLabwareOptions: Selector<Options> = createSelector(
...
       (
         acc: Options,
         labwareEntity: LabwareEntity,
         labwareId: string
       ): Options => {
+        const isAdapter =
+          labwareEntity.def.metadata.displayCategory === 'adapter'
+        const isAdapterOrAluminumBlock =
+          isAdapter ||
+          labwareEntity.def.metadata.displayCategory === 'aluminumBlock'
```
b-cooper: allowedRoles ?

protocol-designer/src/utils/index.ts
```diff
@@ -104,3 +104,8 @@ export const makeTimerText = (
     : `${targetMinutes}  ${i18n.t(
         'application.units.minutes'
       )} ${targetSeconds}  ${i18n.t('application.units.seconds')} timer`
+
+export const getIsAdapter = (id: string): boolean =>
```
b-cooper: same comment, we shouldn't be parsing the `id` at all, and we should be reading the `allowedRoles` value of the labware definition

protocol-designer/src/utils/labwareModuleCompatibility.ts
```diff
@@ -78,9 +76,82 @@ export const getLabwareIsCompatible = (
...
+
+export const getAdapterLabwareIsAMatch = (
+  labwareId: string,
+  allLabware: LabwareOnDeck[],
+  draggedLabwareLoadname: string
+): boolean => {
+  const labwareDefUri = Object.values(allLabware).find(
+    lab => lab.id === labwareId
+  )?.labwareDefURI
+
```
b-cooper, line 127: can't the rest of this function just be a look up from the mapping ALLOWLIST above?

b-cooper (approved):
🌐

protocol-designer/src/step-forms/selectors/index.ts
jerader: oops i did not occur to me that `ProfileErrors` is for TC profile errors. I'll separate it into a new file. thanks!

protocol-designer/src/load-file/migration/7_0_0.ts
jerader: oops good call, they are the same

PR #13238, opened 2023-08-04
fix(app): show network ip info if it exists

brenthagen:
# Overview

shows the robot-reported wifi and ethernet ip info even when the client app is unable to communicate via that ip

reverts part of https://github.com/Opentrons/opentrons/pull/13206

Screen Shot 2023-08-04 at 2 51 44 PM

# Test Plan

 - manually verified with non-communicative OT-2 (screenshot above)

# Changelog

 - Shows network ip info if it exists

# Review requests

# Risk assessment

low


sfoster1 (approved):
excellent, thank you!

PR #13239, opened 2023-08-04
feat(api): Alias `del deck[...]` to `moveLabware(..., OFF_DECK)`

SyntaxColoring:
# Overview

Closes RSS-269.

# Changelog

* `del protocol.deck[...]` is now equivalent to `protocol.move_labware(..., OFF_DECK)`. So, for labware, `del protocol.deck[...]` should now work the way it did before PAPIv2.14.
* Deleting *modules* in PAPIv>=2.14 remains unsupported. This is a Protocol Engine limitation. This now raises a descriptive error.
* Trying to use `del` in PAPIv2.14 will now raise an actionable `APIVersionError` error instead of `TypeError: 'Deck' object doesn't support item deletion`.

# Test Plan

Unit tests are probably sufficient. But we can try running this on a robot and make sure it behaves sensibly. It should *not* pause the run or prompt the user to do anything.

```python
requirements = {"apiLevel": "2.15", "robotType": "Flex"}


def run(protocol):
    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "D1")
    )

    labware_1 = protocol.load_labware("agilent_1_reservoir_290ml", "D2")

    pipette.pick_up_tip()

    pipette.aspirate(200, labware_1["A1"].top())

    del protocol.deck["D2"]

    labware_2 = protocol.load_labware(
        "armadillo_96_wellplate_200ul_pcr_full_skirt", "D2"
    )
    pipette.dispense(200, labware_2["A1"].top())

    pipette.return_tip()

```

# Review requests

There's now a weird API asymmetry where `protocol.move_labware()` can only do manual moves *with* a pause, whereas `del protocol.deck[...]` can only do manual moves *without* a pause. Does `protocol.move_labware()` need to support manual moves *without* a pause?

Does [the `Deck` documentation](https://docs.opentrons.com/v2/new_protocol_api.html?highlight=deck#opentrons.protocol_api.ProtocolContext.deck) need to be updated to explain all these nuances of what's supported in what versions? @ecormany @jwwojak 

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/deck.py
+++ b/api/src/opentrons/protocol_api/deck.py
@@ -6,17 +6,24 @@ from opentrons_shared_data.deck.dev_types import SlotDefV3
 
 from opentrons.motion_planning import adjacent_slots_getters
 from opentrons.protocols.api_support.types import APIVersion
+from opentrons.protocols.api_support.util import APIVersionError
 from opentrons.types import DeckLocation, DeckSlotName, Location, Point
 from opentrons_shared_data.robot.dev_types import RobotType
 
+
 from .core.common import ProtocolCore
 from .core.core_map import LoadedCoreMap
+from .core.module import AbstractModuleCore
 from .labware import Labware
 from .module_contexts import ModuleContext
+from ._types import OFF_DECK
 from . import validation
 
 
 DeckItem = Union[Labware, ModuleContext]
+
+
+_DELITEM_VERSION_GATE = APIVersion(2, 15)
 
 
 @dataclass(frozen=True)
@@ -90,6 +97,34 @@ class Deck(Mapping[DeckLocation, Optional[DeckItem]]):
         item = self._core_map.get(item_core)
 
         return item
+
+    def __delitem__(self, key: DeckLocation) -> None:
+        if self._api_version < _DELITEM_VERSION_GATE:
+            raise APIVersionError(
+                f"Deleting deck elements is not supported with apiLevel {self._api_version}."
+                f" Try increasing your apiLevel to {_DELITEM_VERSION_GATE}."
+            )
+
+        slot_name = _get_slot_name(
+            key, self._api_version, self._protocol_core.robot_type
+        )
+        item_core = self._protocol_core.get_slot_item(slot_name)
+
+        if item_core is None:
+            raise TypeError(f"Slot {repr(key)} doesn't contain anything to delete.")
+        elif isinstance(item_core, AbstractModuleCore):
+            raise TypeError(
+                f"Slot {repr(key)} contains a module, {item_core.get_display_name()}."
+                f" You can only delete labware, not modules."
+            )
+        else:
+            self._protocol_core.move_labware(
+                item_core,
+                new_location=OFF_DECK,
+                use_gripper=False,
+                pick_up_offset=None,
+                drop_offset=None,
+            )
 
     def __iter__(self) -> Iterator[str]:
         """Iterate through all deck slots."""
--- a/api/src/opentrons/protocol_api/deck.py
+++ b/api/src/opentrons/protocol_api/deck.py
@@ -23,9 +23,6 @@ from . import validation
 DeckItem = Union[Labware, ModuleContext]
 
 
-_DELITEM_VERSION_GATE = APIVersion(2, 15)
-
-
 @dataclass(frozen=True)
 class CalibrationPosition:
     """A calibration point on the deck of the robot.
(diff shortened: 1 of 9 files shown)


ecormany:
> "[The test protocol] should not pause the run or prompt the user to do anything."

Check, `dev-odd` runs through this protocol with no pauses.

But I can't see why we would want that. The test protocol is _dangerous_, because the robot blithely assumes that `labware_1` was replaced with `labware_2` in the blink of an eye. If 1 is taller than 2, the robot will quite happily smash into it. _Forcing_ a pause and user intervention eliminates that risk. 

If users have already coded delete-pause-load flows in ≤2.14, treating deletion as ordinary move off-deck will at worst put two consecutive pauses in their protocol. If they're truly migrating forward to 2.15, they can delete the additional pause. If they remove the pause and then lower their `apiLevel` they could get into trouble…but that's an edge case. 

And I don't really want to even add a note about that edge case to docs, because we don't mention `del` anywhere except the API Reference currently. Seems like a real [[[WP:BEANS]]](https://en.wikipedia.org/wiki/Wikipedia:Don%27t_stuff_beans_up_your_nose) scenario.

jwwojak:
I am missing some context or background to really understand what's at stake here. However, as this is the question:

> Does [the Deck documentation](https://docs.opentrons.com/v2/new_protocol_api.html?highlight=deck#opentrons.protocol_api.ProtocolContext.deck) need to be updated to explain all these nuances of what's supported in what versions?

Short answer: Yes

Longer answer: My preference is to err on the side of always providing information to users. It's better to surface this somewhere rather than have it be functionality that users understand only after a (what might be) frustrating trial and error experience. 

Additionally, if an explanation is too long for what's typical in a property description, then is there somewhere else in the docs we can put it?

api/src/opentrons/protocol_api/deck.py
```diff
@@ -37,7 +40,7 @@ def _get_slot_name(slot_key: DeckLocation, api_version: APIVersion) -> DeckSlotN
     try:
         return validation.ensure_deck_slot(slot_key, api_version)
     except (TypeError, ValueError) as error:
-        raise KeyError(str(error)) from error
+        raise KeyError(slot_key) from error
```
sanni-t: 🙈

api/src/opentrons/protocol_api/deck.py
```diff
@@ -82,6 +85,42 @@ def __getitem__(self, key: DeckLocation) -> Optional[DeckItem]:
...
+            raise TypeError(
+                f"Slot {repr(key)} contains a module, {item_core.get_display_name()}."
+                f" You can only delete labware, not modules."
+            )
+        else:
+            self._protocol_core.move_labware(
+                item_core,
+                new_location=OFF_DECK,
+                use_gripper=False,
+                pause_for_manual_move=False,
```
sanni-t, line 131: Hmm.. @ecormany has a good point that if we use a pause in here then it won't be too bad, rather, it would be a more cautious approach. Maybe we can quickly run this by the science/ apps engg team if we're too concerned.

Also, how does this no-pause command look in the run preview/ run log text? It might need an app-side update if we want to keep the no-pause approach.

sanni-t (commented):
Mostly looks good to me but I think we should do at least a quick check with science/ apps eng teams to make sure it isn't more confusing to have the deletion be without pause.

api/src/opentrons/protocol_api/deck.py
SyntaxColoring: > Also, how does this no-pause command look in the run preview/ run log text? It might need an app-side update if we want to keep the no-pause approach.

Like this, which I think is fine:

Screenshot 2023-08-10 at 4 24 42 PM

(notwithstanding the "off deck to off deck" part, which is a known separate problem that applies to any `moveLabware` command)

Also see #13171, which is a recent fix.

SyntaxColoring:
~~I think I have a deck slot normalization bug somewhere where if you swap the test protocol to an OT-2, it complains that slot 2 is still occupied when you try to load the second labware. Re-drafting while I look into that.~~ Turns out that's a bug in all of the methods of the `Deck` class. Fixed in #13320.

api/src/opentrons/protocol_api/deck.py
sanni-t: Nice! I think we should add something like `(will not pause)` to the text. You can do it here if you want to tinker with some js code or we can talk to someone on app/ui.

SyntaxColoring: > You can do it here if you want to tinker with some js code or we can talk to someone on app/ui.

Leaving that out of this PR.

api/src/opentrons/protocol_api/deck.py
```diff
@@ -91,6 +95,44 @@ def __getitem__(self, key: DeckLocation) -> Optional[DeckItem]:
...
         return item
 
+    def __delitem__(self, key: DeckLocation) -> None:
+        if self._api_version == APIVersion(2, 14):
+            # __delitem__() support history:
+            #
+            # * PAPIv<=2.13 (non Protocol Engine): Yes, but that goes through a different Deck class
+            # * PAPIv2.14 (Protocol Engine): No
+            # * PAPIv2.15 (Protocol Engine): Yes
+            raise APIVersionError(
```
SyntaxColoring, line 105: Instead of enshrining this as an official `apiLevel` thing, do we want to just treat it as a bug, and have v7.0.0 retroactively add support for this to PAPIv2.14?

The advantage to enshrining it as an official `apiLevel` thing is that new software will prevent you from writing a protocol that will raise a potentially confusing error on v6.3.x. The confusing error says something like "object is not subscriptable."

SyntaxColoring: I think we should just treat it as a bug and have v7.0 retroactively add support to apiLevel 2.14.

The downside of the confusing error message is temporary. 

The upside, that we get to avoid a confusing back-and-forth thing in the API, is permanent.

SyntaxColoring: Discussed with @sanni-t: we'll enshrine this as an official `apiLevel` thing.

I question whether enshrining things like this is a good tradeoff, but it has been our policy, so far, to at least attempt to do it. It's easy to do here, so we don't have a good reason to break that precedent in this PR.

sanni-t (approved):
🙌

PR #13240, opened 2023-08-04
fix(app): change desktop network icon priority

brenthagen:
# Overview

changes the desktop network icon priority to ethernet > wifi > usb

closes RQA-1204

# Test Plan

 - updated unit tests
 - manually verified desktop/ODD icon match

# Changelog

 - Changes desktop network icon priority

# Review requests

# Risk assessment

low


koji (approved):
Tested with a dev kit and worked as expected.

There is be one thing that we may need to change.
In RobotCard, we don't refetch networkInterface info until the app renders device listing page so the icon in RobotCard isn't the latest status.
I guess this isn't critical since most users won't stay device landing page for a long time.

b-cooper (approved):
📶

PR #13241, opened 2023-08-04
feat(app): add probe-based LPC behind feature flag, add ODD feature flag toggles

b-cooper:
# Overview

Add `lpcWithProbe` dev internal feature flag.  When set to true, LPC prompts to attach probe, then visits all tipracks and labware with the probe, then prompts to detach probe. Allow feature flags
to be toggled on and off in the ODD

# Changelog

Add the ability to turn on and off app side feature flags through the ODD settings page when the dev tools are enabled (ala desktop)

# Risk assessment
low, all changes are behind FF


app/src/organisms/Devices/ProtocolRun/SetupLabwarePositionCheck/index.tsx
b-cooper, line 1: This was a small bug that was leading to the disabled tooltip being rendered on two different buttons, instead of just the "start LPC" button

app/src/pages/OnDeviceDisplay/RobotSettingsDashboard/RobotSettingsList.tsx
```diff
@@ -18,8 +36,9 @@ import { onDeviceDisplayRoutes } from '../../../App/OnDeviceDisplayApp'
 import { useNetworkConnection } from '../hooks'
 import { RobotSettingButton } from './RobotSettingButton'
 
-import type { State } from '../../../redux/types'
+import type { Dispatch, State } from '../../../redux/types'
 import type { SetSettingOption } from './'
+import { StyledText } from '../../../atoms/text'
```
shlokamin, line 41: should be before type imports

shlokamin (approved):
wooo, this is dope. thanks for doing this!

PR #13246, opened 2023-08-07
refactor(hardware-testing): Add ability for optional air-gap during low-volume aspirations

andySigler:
# Overview

Test results are showing that we may want to add an extra air-gap for low-volume (<5.0 uL) aspirations, to help make sure that everything is later fully dispensed.

This PR does a couple of refactors to allow this:

- rename what we previously called `leading_air_gap` and renaming it to `blow_out_submerged` because that is a more descriptive name of what it is doing
- use the previous parameter `leading_air_gap` to describe an actual "leading" air-gap done before aspirations
- default all `leading_air_gap` to `0.0` for now, so it has no affect

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
lgtm

PR #13247, opened 2023-08-07
refactor(api): Remove some opentrons.protocols type-checking exceptions

SyntaxColoring:
# Overview

While I'm in the neighborhood for RSS-265, remove some mypy overrides in `opentrons.protocols` so we type-check that whole part of the module tree a bit more strictly.

# Changelog

Attain strict type-checking for:
* `tests/opentrons/protocols/*`
* Any new modules created in `src/opentrons/protocols/*`
* `bundle.py`

# Test Plan

Make sure CI keeps passing.

# Review requests

None in particular.

# Risk assessment

No risk.

### DIFF
--- a/api/mypy.ini
+++ b/api/mypy.ini
@@ -14,31 +14,14 @@ warn_untyped_fields = True
 
 # TODO(mc, 2021-09-08): fix and remove any / all of the
 # overrides below whenever able
-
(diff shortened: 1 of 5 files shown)


Laura-Danielle (approved):
nice

PR #13251, opened 2023-08-08
refactor(api): allow tip presence to be turned off

Laura-Danielle:
## Overview

Sometimes, you may not want the tip presence check on. We should allow users to turn this feature off.


sfoster1 (approved):
Looks good to me!

PR #13254, opened 2023-08-08
feat(api): gate pre-protocol-home behind check for door status

sanni-t:
Closes RSS-135

# Overview

Removes the call to `hardware.home` at the beginning of a protocol run and instead, adds a home command to the command queue during protocol load, before any protocol commands are queued. This makes the initial homing follow the same rules as any protocol command and will pause when the robot door is opened.

# Test Plan

On an OT2/ Flex, test with a python and a json protocol. Make sure the home command shows up as the first command in the run command view. With the 'Pause on door open' setting enabled for the robot, verify that the homing pauses when door is opened and is able to be resumed after closing the door and clicking resume. Check that the run statuses are updated correctly.

# Review requests

- I've implemented this as the ticket says i.e, by adding an engine command instead of using hardware home. It's been a while since that ticket was written so want to make sure that it's still the recommended approach.
- A side-effect of this approach is that the run commands list will now always have homing as the first command in the protocol. Pros of this effect- the homing action will be documented and visible in the run log so one does not have to assume that the robot homed (or did not home) before starting a protocol. Cons- having an implicit home command as the first command of a protocol might confuse users. We could make the app/ ODD omit this first command from view but that might create additional problems.
- Should we be extending this pause-on-door-open behavior to the post-protocol cleanup procedure (homing and maybe dropping tips) as well? If yes, how do we implement it? And what would 'pausing' and 'resuming' mean for commands after the run has been told to stop? 

# Risk assessment

- Medium. Has a risk of protocols not running if implemented wrongly. But it's a simple change and easy to test.

### DIFF
--- a/api/src/opentrons/protocol_runner/protocol_runner.py
+++ b/api/src/opentrons/protocol_runner/protocol_runner.py
@@ -15,7 +15,12 @@ from opentrons.protocol_reader import (
     JsonProtocolConfig,
     PythonProtocolConfig,
 )
-from opentrons.protocol_engine import ProtocolEngine, StateSummary, Command
+from opentrons.protocol_engine import (
+    ProtocolEngine,
+    StateSummary,
+    Command,
+    commands as pe_commands,
+)
 
 from .task_queue import TaskQueue
 from .json_file_reader import JsonFileReader
@@ -140,6 +145,10 @@ class PythonAndLegacyRunner(AbstractRunner):
             broker=broker,
             equipment_broker=equipment_broker,
         )
+        initial_home_command = pe_commands.HomeCreate(
+            params=pe_commands.HomeParams(axes=None)
+        )
+        self._protocol_engine.add_command(request=initial_home_command)
 
         self._task_queue.set_run_func(
             func=self._legacy_executor.execute,
@@ -156,7 +165,6 @@ class PythonAndLegacyRunner(AbstractRunner):
         if protocol_source:
             await self.load(protocol_source=protocol_source)
 
-        await self._hardware_api.home()
         self.play()
         self._task_queue.start()
         await self._task_queue.join()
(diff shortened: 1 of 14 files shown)


sanni-t:
Notes from talking with the RSS team:
> - I've implemented this as the ticket says i.e, by adding an engine command instead of using hardware home. It's been a while since that ticket was written so want to make sure that it's still the recommended approach.
> - A side-effect of this approach is that the run commands list will now always have homing as the first command in the protocol. Pros of this effect- the homing action will be documented and visible in the run log so one does not have to assume that the robot homed (or did not home) before starting a protocol. Cons- having an implicit home command as the first command of a protocol might confuse users. We could make the app/ ODD omit this first command from view but that might create additional problems.

Discussed with RSS team and confirmed that this is the behavior we want and has been an anticipated change.

> - Should we be extending this pause-on-door-open behavior to the post-protocol cleanup procedure (homing and maybe dropping tips) as well? If yes, how do we implement it? And what would 'pausing' and 'resuming' mean for commands after the run has been told to stop? 

We don't yet have a great answer to this. It's not currently possible to add and execute commands after a run is completed. We can make homing a client-initiated action after the protocol ends, but that leaves holes for cases where opentrons client interaction is not guaranteed. We use homing after a protocol as a way to 'reset' the OT2 & Flex motor drivers after a hardware halt action. Without this home, the next gantry movement command will result in a MustHomeError unless the robot is sent a home command. So we stick with always homing after the halt in the mostly foolproof way of calling it implicitly as part of run cleanup. 
Maybe we could have different behaviors for OT-2 vs Flex since the driver resetting is mostly an OT2 concern. But that is a separate topic.

sanni-t:
@SyntaxColoring I tested re-running an older protocol that was already analyzed on a previous release version and it ran without running into issues. It showed the run preview with analyzed protocol steps but ran the actual new steps. The only weirdness noticeable was that the 'current step' text differed from the run preview.

Screenshot 2023-08-09 at 4 54 23 PM

jbleon95 (approved):
Simple and straightforward change, LGTM

PR #13255, opened 2023-08-08
fix(app): change pipette calibration error handling to match gripper

smb2268:
# Overview

Adding a formal remove probe step after a calibration failure is causing errors for a few error types (including E-Stop errors). To mitigate this, we'll remove that screen and instead add a line of copy on the error screen telling the user to remove the probe before exiting.

# Test Plan
Trigger an E-Stop error during a calibration flow and confirm you are able to exit after viewing the error screen.

# Changelog
- Remove calibration error screen that continues to a remove probe screen
- Add a line of copy on the normal error screen telling user to remove the probe

# Review requests
See test plan

# Risk assessment
Low


jerader (approved):
nice!

PR #13256, opened 2023-08-08
fix(api): Centralize Python parsing logic, and fix syntax error reporting

SyntaxColoring:
# Overview

`FileIdentifier` and `opentrons.protocols.parse` have had a duplication problem for a while now. This resolves a lot of that duplication for Python protocols, leaving JSON protocols alone for now.

Fixes RSS-161 (unactionable error messages for certain kinds of malformed protocols). [Before](https://github.com/Opentrons/opentrons/assets/3236864/6ba5273b-b446-453f-902d-395f41f8f298) vs. [after](https://github.com/Opentrons/opentrons/assets/3236864/a5c19353-f126-4d3a-9b32-81627b99cc6e).

Helps with RSS-268 (supporting Protocol Engine through `opentrons.execute` etc.).

Helps with RSS-264 and RSS-265 (adding some new validation for Python files).

# Test Plan

I think this is decently covered by automated tests.

One quick manual test that could be helpful is pushing this branch to some random robots in the office and confirming that the robot server still boots up. We've had problems in the past where schema changes make the server think its stored protocols are corrupted, which can block general usability. I don't think that will happen here, but it's good to check.

# Changelog

* Instead of letting `FileIdentifier` do its own `ast.parse()` crap, have it delegate to `opentrons.protocols.parse.parse()`.
* Resolve a lot of the overlap between the two modules' test suites. The `FileIdentifier` tests now mock out `opentrons.protocols.parse.parse()`. 
* Simplify the the public interface of `opentrons.protocols.parse`. There were several public functions that were veering into weedy implementation-detail territory. Now that `FileIdentifier` doesn't rely on those, we can make them private. Their tests have been ported to go through the top-level `parse()` function instead, so we should still be testing the same codepaths.
* When `opentrons.protocols.parse` identifies a user error in a Python file, ensure it raises it with a distinct error type, `MalformedPythonError`. Also ensure it includes a human-readable message that the higher layers can relay to the Opentrons App.

# Review requests

None in particular.

# Risk assessment

I think low, but see my thoughts in the test plan.

### DIFF
--- a/api/src/opentrons/protocol_reader/file_identifier.py
+++ b/api/src/opentrons/protocol_reader/file_identifier.py
@@ -204,9 +204,13 @@ def _analyze_python_protocol(
 ) -> IdentifiedPythonMain:
     try:
         parsed = parse.parse(protocol_file=py_file.contents, filename=py_file.name)
-        assert isinstance(parsed, PythonProtocol)
     except MalformedPythonProtocolError as e:
         raise FileIdentificationError(e.short_message) from e
+
+    # We know this should never be a JsonProtocol. Help out the type-checker.
+    assert isinstance(
+        parsed, PythonProtocol
+    ), "Parsing a Python file returned something other than a Python protocol."
 
     if parsed.api_level > MAX_SUPPORTED_VERSION:
         raise FileIdentificationError(
--- a/api/src/opentrons/protocol_reader/file_identifier.py
+++ b/api/src/opentrons/protocol_reader/file_identifier.py
@@ -10,7 +10,7 @@ from opentrons_shared_data.robot.dev_types import RobotType
 
 from opentrons.protocols.api_support.definitions import MAX_SUPPORTED_VERSION
 from opentrons.protocols.api_support.types import APIVersion
-from opentrons.protocols.parse import parse
+from opentrons.protocols import parse
 from opentrons.protocols.types import MalformedPythonError, PythonProtocol
 
 from .file_reader_writer import BufferedFile
@@ -203,7 +203,7 @@ def _analyze_python_protocol(
     py_file: BufferedFile,
 ) -> IdentifiedPythonMain:
     try:
-        parsed = parse(protocol_file=py_file.contents, filename=py_file.name)
+        parsed = parse.parse(protocol_file=py_file.contents, filename=py_file.name)
         assert isinstance(parsed, PythonProtocol)
     except MalformedPythonError as e:
         raise FileIdentificationError(e.short_message) from e
--- a/api/src/opentrons/protocol_reader/file_identifier.py
+++ b/api/src/opentrons/protocol_reader/file_identifier.py
@@ -1,6 +1,5 @@
 """File identifier interface."""
 
-import ast
 import json
 from dataclasses import dataclass
 from typing import Any, Dict, Sequence, Union
@@ -10,13 +9,9 @@ import anyio
 from opentrons_shared_data.robot.dev_types import RobotType
 
 from opentrons.protocols.api_support.definitions import MAX_SUPPORTED_VERSION
-from opentrons.protocols.parse import (
-    extract_static_python_info,
-    robot_type_from_static_python_info,
-    version_from_static_python_info,
-)
 from opentrons.protocols.api_support.types import APIVersion
-from opentrons.protocols.types import MalformedPythonError, StaticPythonInfo
+from opentrons.protocols.parse import parse
+from opentrons.protocols.types import MalformedPythonError, PythonProtocol
 
 from .file_reader_writer import BufferedFile
 from .protocol_files_invalid_error import ProtocolFilesInvalidError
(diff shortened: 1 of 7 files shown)


api/tests/opentrons/protocols/test_parse.py
```diff
@@ -470,28 +517,128 @@ def test_parse_extra_contents(
...
-        """
-        metadata = {"apiLevel": "2.0"}
+        # Bad Python syntax:
+        (
+            """
+            metadata = {"apiLevel": "2.0"}
+            def run(ctx)  # Missing ":".
+                pass
+            """,
+            '(invalid syntax)|(Missing ":")',  # This error message depends on the Python version.
```
SyntaxColoring, line 529: Maybe add an integration test for this because it was broken before in an integrationy way.

api/src/opentrons/protocol_reader/file_identifier.py
```diff
@@ -204,64 +199,25 @@ def _analyze_json_protocol(
...
-        # todo(mm, 2021-09-13): Investigate whether it's really appropriate to leave
-        # the Python compilation flags at their defaults. For example, we probably
-        # don't truly want the protocol to inherit our own __future__ features.
-        module_ast = ast.parse(source=py_file.contents, filename=py_file.name)
-    except (SyntaxError, ValueError) as e:
-        # ast.parse() raises SyntaxError for most errors,
-        # but ValueError if the source contains null bytes.
-        raise FileIdentificationError(f"Unable to parse {py_file.name}.") from e
+        parsed = parse.parse(protocol_file=py_file.contents, filename=py_file.name)
+        assert isinstance(parsed, PythonProtocol)
```
Laura-Danielle: Why do you need this assert? Wouldn't the parser take care of determining if it's a `MalformedPythonError`? And if it doesn't, shouldn't it do that?

api/src/opentrons/protocols/parse.py
```diff
@@ -87,9 +95,11 @@ def version_from_string(vstr: str) -> APIVersion:
     """
     matches = API_VERSION_RE.match(vstr)
     if not matches:
-        raise ValueError(
-            f"apiLevel {vstr} is incorrectly formatted. It should "
-            "major.minor, where both major and minor are numbers."
+        raise MalformedPythonError(
```
Laura-Danielle: I am not sure I agree with this error change. Technically the python code is fine, they just need to change their API version. `ValueError` or `RuntimeError` still makes sense to me here.

Laura-Danielle (approved):
I had two comments, but otherwise this looks really good to me

api/src/opentrons/protocol_reader/file_identifier.py
SyntaxColoring: `parse()` can either return a `PythonProtocol` or a `JsonProtocol`. We know that in practice, this one will always be a `PythonProtocol` because JSON files currently go down a different code path. But the type-checker needs to be told that.

The `assert isinstance(parsed, PythonProtocol)` is effectively acting as `cast(parsed, PythonProtocol)`. The advantage over `cast()` is that it will raise a slightly better runtime error if there's a bug that causes our assumption to be wrong.

api/src/opentrons/protocols/parse.py
TamarZanzouri: I might agree. maybe a ValueError is more appropriate in this case

SyntaxColoring: By "Technically the python code is fine," are you talking about how it's valid as vanilla Python code, and it's only invalid because of our additional Opentrons rules on top of that?

That's true. But I *am* intending `MalformedPythonError` to cover both kinds of problem. For example, it's also used if you define multiple `run()` functions, which is also perfectly valid Python.

The reason these need to be a custom type instead of a `ValueError` or `RuntimeError` is that if an exception raises out of `parse()`, the calling code needs some way of identifying:

1. Whether it's the user's fault for giving us bad input, or a bug in our code.
2. If it's the user's fault, what message we should show to the user.

Both of which are difficult to do robustly without a custom type.

We could split `MalformedPythonError` into two types (`PythonProtocolStructureError` vs. `PythonProtocolPythonError`?), but the caller would treat them equivalently, so I'm not sure if that's worthwhile.

Would a name other than `MalformedPythonError` improve things? I don't want to do `MalformedProtocolError` because it's not currently used for JSON protocols.

api/src/opentrons/protocols/types.py
```diff
@@ -98,16 +96,38 @@ def run(ctx):
 """
 
 
-class MalformedProtocolError(Exception):
-    def __init__(self, message: str) -> None:
-        self.message = message
-        super().__init__(message)
+# TODO(mm, 2023-08-07): Align with the app team on how to surface errors like this,
```
TamarZanzouri, line 99: :+1:

TamarZanzouri (approved):
I think this is a great change! thank you!

api/tests/opentrons/protocols/test_parse.py
SyntaxColoring: Leaving this alone for now.

I thought the original bug was that the message wasn't being propagated through robot-server HTTP responses, but it's actually that it wasn't being propagated through the `python -m opentrons.cli analyze` CLI. We don't have a pattern for integration-testing that today, and I'd rather not start one now because our Python protocol integration tests are already kind of too spread out and confusing.

api/src/opentrons/protocols/parse.py
Laura-Danielle: @SyntaxColoring I would prefer maybe `MalformedOpentronsPythonAPIError` maybe or something equivalent to that

api/src/opentrons/protocol_reader/file_identifier.py
Laura-Danielle: Ok. I'm generally not a huge fan of asserts because 1. the error is confusing and 2. it's a bit confusing for me code-wise.

Typically for these cases I prefer to do `isinstance` checks and raise a custom error if the need arises. This is not a change that urgently needs to happen for the PR to get merged in, but just my 2 cents.

api/src/opentrons/protocols/parse.py
SyntaxColoring: 👍 I'll try `MalformedPythonProtocolError`.

api/src/opentrons/protocol_reader/file_identifier.py
SyntaxColoring: Okay, I disagree. I think this is precisely the kind of internal consistency check that `assert`s are meant for. And that it would be *more* confusing to have a custom exception type, because that implies it's a condition that we actually expect to happen and that we care about being prepared for.

I'll certainly add a comment and custom message to clarify this one's intent and make its error less confusing, though.

PR #13257, opened 2023-08-08
feat(api, robot-server): stop protocol engine if estop is pressed

fsinapi:
# Overview

It is important that runs are invalidated when the estop is pressed. This was already provided implicitly if the estop was pressed during a run _while the gantry is moving_, but if a protocol is paused and then the estop is pressed then the run is actually still valid.

This PR adds callbacks to the Robot Server that ensure all protocol engines are invalidated if the estop is pressed.

* For a normal protocol run, every pending command is marked as failed with an Estop error. The Protocol Runner will then implicitly clean up the run with the `finish` function.
* For a maintenance run, the run is stopped and a `HardwareStoppedAction` ensures that the protocol has a stopped_at point. This ensures that the app cannot send any more commands to that protocol, and that the maintenance run endpoints will be able to clean up the engine.

# Test Plan

Run a bunch of protocols and hit the estop at various points! Tested hitting the estop:
* While the gantry is moving
* After pausing the protocol between movements
* During LPC
* During attach flow
* During pipette/gripper calibration

In all of the above cases, the run is marked as failed and the robot does _not_ try to home the gantry. There is an issue during pipette calibration that has an open fix PR so it is not a blocker.

Additionally, if you hit the estop during the initial home before the protocol technically starts, it does _not_ trash the run (because there is not a running command yet). This seems appropriate because the gantry is just preparing to be able to start the run, and it will not move again until the user hits the Play button and the gantry tries to home again.

# Changelog

# Review requests

The mechanism here is possibly a little kludgy but I don't think it seems risky - it will reliably clear the status on the protocol engine such that it can be cleared by the robot server. Is there any reason I'm wrong about that?

# Risk assessment

Medium - this could result in annoying user experience if the implementation is bugged, but also it will only happen when someone hits the estop.


api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -217,6 +219,41 @@ async def add_and_execute_command(
...
+        elif maintenance_run:
+            stop_action = self._state_store.commands.validate_action_allowed(
+                StopAction(from_estop=True)
+            )
+            self._action_dispatcher.dispatch(stop_action)
+            hardware_stop_action = HardwareStoppedAction(
+                completed_at=self._model_utils.get_timestamp()
+            )
+            self._action_dispatcher.dispatch(hardware_stop_action)
+            self._queue_worker.cancel()
```
sfoster1, line 280: I think we need an `else` here to prevent an (unlikely) race condition because we're dealing with threads, so we could be in a state where we don't have an active command but do have an active engine. I'm not sure what to put in the `else`, though.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -400,13 +435,22 @@ async def use_attached_modules(
...
         if (
             isinstance(error, ProtocolCommandFailedError)
             and error.original_error is not None
         ):
+            if error.original_error.errorCode == code.value.code:
+                return True
             return any(
-                code.value.code == wrapped_error.errorCode
-                for wrapped_error in error.original_error.wrappedErrors
+                wrapped.errorCode == code.value.code
```
sfoster1: I think we want this to be recursive, right?

sfoster1 (changes requested):
I think we're missing a state case in the estop handling but other than that looks good to me.

api/src/opentrons/protocol_engine/protocol_engine.py
fsinapi: Yeah very good point. I think what we need to do is check if the run was started. From there, my thinking is:

* If it has not been started:
  * We shouldn't have to do anything here, the run is either waiting to be `play`'d or was in the midst of the homing step before it first starts
  * In the latter case, what happens now is the Protocol Runner task exits with an exception and it's like the protocol never started. The user can hit the Play button (once the estop is released) and it should continue homing from where it was and start the run like normal.
* If it _has_ been started:
  * The `StopAction` should be reconfigured such that, if the stop is from an Estop event, every queued command is marked failed and removed from the queue (just like with the `FailCommandAction` except we don't need to provide an active command).
  * If we do that, the Protocol Runner should pick up that it failed and then handle the cleanup.

fsinapi: okay I was overthinking it but fix is in e032f1cc5c9dbdfc25df8b027b2bc6b9f04fe3bb

api/src/opentrons/protocol_engine/protocol_engine.py
fsinapi: indeed cd63dfeb070da67b0a80f16075ec014edf5a4a2a

sfoster1:
FYI https://github.com/Opentrons/opentrons/pull/13254 will make estop during the pre-protocol home trash the run I think

sfoster1:
Which is good

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -222,6 +224,61 @@ async def add_and_execute_command(
...
+                completed_at=self._model_utils.get_timestamp(),
+                finish_error_details=FinishErrorDetails(
+                    error=EStopActivatedError(message="Estop Activated"),
+                    error_id=self._model_utils.generate_id(),
+                    created_at=self._model_utils.get_timestamp(),
+                ),
+            )
+            self._action_dispatcher.dispatch(hardware_stop_action)
+            self._queue_worker.cancel()
+
```
sfoster1, line 283: Okay, that makes sense. Can we just drop a log here in the (not currently present) straight-up `else` case just in case?

sfoster1 (approved):
This makes sense to me, nice. I think we should have a log in that `else` case still but the precondition argument carries me.

api/src/opentrons/protocol_engine/protocol_engine.py
fsinapi: yeeep 2464f3b4c8712fbbb4b0fb1250bfb37dd539245f

PR #13258, opened 2023-08-08
feat(api): Add more validation to metadata and requirements dicts

SyntaxColoring:
# Overview

Closes RSS-264 and RSS-265.

# Changelog

Add additional validation when parsing Python protocol files:

* Forbid putting `apiLevel` in `metadata` and `requirements` simultaneously. Each protocol should choose one or the other. The historical behavior is that `requirements` overrides `metadata`.
* Forbid unrecognized keys in `requirements`.
* Require `"apiLevel": "2.15"` or newer if  `"robotType"` is `"Flex"` or `"OT-3"`.

# ⚠️Compatibility

These are breaking changes, but thankfully, they're only breaking for our internal testing protocols. For example, [according to](https://opentrons.slack.com/archives/C049WNGU22H/p1691613304413849) @nusrat813, some QA and ABR protocols *do* specify `apiLevel` in both `metadata` and `requirements`.

We want to add this validation for launch, but we don't want to cause database problems on all the existing robots and force our internal testers to do factory-resets.

To accomplish that, this PR introduces a temporarily-configurable "strict mode" for Python parsing.

**We *disable* strict mode for protocols that are already on robots.** That means when:

* robot-server parses already-stored protocols to support GET /protocols.
* robot-server runs an already-stored protocol.

**And we *enable* strict mode for protocols that are not.** That means when:

* robot-server receives new protocols over HTTP.
* The local analysis engine analyzes a protocol from the Opentrons App.
* Someone uses opentrons_simulate et. al.

In the future, some time after the Flex launch, we'll always use strict mode. 

# Test Plan

Automated unit tests should be sufficient. We added integration tests in #13295.

# Review requests

Do we agree with the approach in the ⚠️ compatibility section above?

# Risk assessment

See the ⚠️ compatibility section above.

### DIFF
--- a/api/src/opentrons/protocol_reader/protocol_reader.py
+++ b/api/src/opentrons/protocol_reader/protocol_reader.py
@@ -71,7 +71,9 @@ class ProtocolReader:
             ProtocolFilesInvalidError: Input file list given to the reader
                 could not be validated as a protocol.
         """
-        identified_files = await self._file_identifier.identify(files)
+        identified_files = await self._file_identifier.identify(
+            files, flex_dev_compat=False
+        )
         role_analysis = self._role_analyzer.analyze(identified_files)
         await self._file_format_validator.validate(role_analysis.all_files)
 
@@ -101,6 +103,7 @@ class ProtocolReader:
         files: Sequence[Path],
         directory: Optional[Path],
         files_are_prevalidated: bool = False,
+        flex_dev_compat: bool = False,
     ) -> ProtocolSource:
         """Compute a `ProtocolSource` from protocol source files on the filesystem.
 
@@ -116,6 +119,8 @@ class ProtocolReader:
                 stuff for the `ProtocolSource`. This can be 10-100x faster, but you
                 should only do it with protocols that have already been validated
                 by this module.
+            flex_dev_compat: See the parameter of the same name in
+                `opentrons.protocols.parse.parse()`.
 
         Returns:
             A `ProtocolSource` describing the validated protocol.
@@ -125,7 +130,10 @@ class ProtocolReader:
                 could not be validated as a protocol.
         """
         buffered_files = await self._file_reader_writer.read(files)
-        identified_files = await self._file_identifier.identify(buffered_files)
+        identified_files = await self._file_identifier.identify(
+            files=buffered_files,
+            flex_dev_compat=flex_dev_compat,
+        )
         role_analysis = self._role_analyzer.analyze(identified_files)
         if not files_are_prevalidated:
             await self._file_format_validator.validate(role_analysis.all_files)
--- a/api/src/opentrons/protocol_reader/protocol_reader.py
+++ b/api/src/opentrons/protocol_reader/protocol_reader.py
@@ -2,6 +2,8 @@
 from pathlib import Path
 from typing import Optional, Sequence
 
+from opentrons.protocols.parse import PythonParseMode
+
 from .file_identifier import (
     FileIdentifier,
     IdentifiedFile,
@@ -72,7 +74,7 @@ class ProtocolReader:
                 could not be validated as a protocol.
         """
         identified_files = await self._file_identifier.identify(
-            files, flex_dev_compat=False
+            files, python_parse_mode=PythonParseMode.NORMAL
         )
         role_analysis = self._role_analyzer.analyze(identified_files)
         await self._file_format_validator.validate(role_analysis.all_files)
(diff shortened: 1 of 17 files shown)


api/src/opentrons/protocols/parse.py
```diff
@@ -48,6 +50,10 @@
 API_VERSION_FOR_JSON_V5_AND_BELOW = APIVersion(2, 8)
 
 
+# Matches the minimum version for the Protocol Engine PAPI core.
+_MIN_API_VERSION_FOR_FLEX = APIVersion(2, 14)
```
sanni-t: Is this to allow some older Flex test protocols to work? Because 2.15 is the official Flex support version.
(Update: verified that all our ABR, QA & science test protocols are on 2.15)

robot-server/tests/integration/http_api/persistence/test_compatibility.py
```diff
@@ -147,23 +152,22 @@ async def test_protocols_analyses_and_runs_available_from_older_persistence_dir(
...
     snapshot = flex_dev_compat_snapshot
-    with DevServer(persistence_directory=snapshot.get_copy()) as server:
-        server.start()
-        async with RobotClient.make(
-            base_url=f"http://localhost:{server.port}", version="*"
-        ) as client:
+    async with RobotClient.make(
+        base_url=f"http://localhost:{_PORT}", version="*"
+    ) as client:
+        assert await client.wait_until_dead(), "Dev Robot is running and must not be."
```
sanni-t:
```suggestion
        assert await client.wait_until_dead(), "Dev Robot is running but it should not be."
```
maybe?

sanni-t (approved):
LGTM except the question about API level for Flex.
Also, just a tiny nitpick that we should document somewhere the 'dev compat' behavior that the apiLevel specified in requirements dict get preference if it's present in both requirements and metadata. Even if it's just in the PR description/ ticket/ an internal doc.

api/src/opentrons/protocols/parse.py
SyntaxColoring: Yep, that was the idea.

Thanks for verifying. I'll change this to 2.15.

api/src/opentrons/protocols/parse.py
```diff
@@ -86,6 +91,50 @@ def _validate_v2_ast(protocol_ast: ast.Module) -> None:
...
+    if api_level_in_metadata and api_level_in_requirements:
+        # If a user does this, it's almost certainly a mistake. Forbid it to avoid complexity in
+        # which dict takes precedence, and in what happens when you upload to an old software
+        # version that only knows about the metadata dict, not the requirements dict.
+        raise MalformedPythonProtocolError(
+            "You may only put apiLevel in the metadata dict or the requirements dict, not both."
+        )
+
+
+def _validate_robot_type_at_version(robot_type: RobotType, version: APIVersion) -> None:
```
TamarZanzouri, line 169: :white_check_mark:

api/src/opentrons/protocol_reader/protocol_reader.py
```diff
@@ -71,7 +71,9 @@ async def save(
             ProtocolFilesInvalidError: Input file list given to the reader
                 could not be validated as a protocol.
         """
-        identified_files = await self._file_identifier.identify(files)
+        identified_files = await self._file_identifier.identify(
+            files, flex_dev_compat=False
```
TamarZanzouri: why is this set to False?

SyntaxColoring: This is in `ProtocolReader.save()`, which is exclusively used by [`POST /protocols`](https://github.com/Opentrons/opentrons/blob/b63c78e49245deedf58d2261d2a95c0790ec8f1f/robot-server/robot_server/protocols/router.py#L213). We want to disallow uploading any additional weird protocol files.

TamarZanzouri: I see. I think the name really confuses me :-(

TamarZanzouri (approved):
I agree with the changes. I think `flex_dev_compat`is a confusing name but I cant think of a better name? I thought strict_mode is better but its also a midtime solution? if it wont stay in the code im ok with leaving it as is. thank you for fixing this!

api/src/opentrons/protocol_reader/protocol_reader.py
SyntaxColoring: Discussed in a call. I've reworked this as an enum and given it a more verbose name. I'm hoping this is both more self-documenting, and more IDE-friendly:

Screenshot 2023-08-15 at 6 49 16 PM

Screenshot 2023-08-15 at 6 38 07 PM

TamarZanzouri (approved):
Love the change you did with the flex compatibility! Double approve now ☺️ thank you for making this change

PR #13259, opened 2023-08-09
fix(shared-data): Updates Gripper force-to-duty-cycle polynomial for PVT build

andySigler:
# Overview

Force measurements were taken of the PVT grippers built, and this polynomial was created.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
yay

PR #13260, opened 2023-08-09
refactor(app): button positioning in pipette flow modals

jerader:
closes RQA-1165 and RQA-1166 and RQA-1233 and RAUT-623

# Overview

Button positioning for the on device display modals in the pipette flows

Also, make sure early exiting out of the flows goes to the correct page on ODD

# Test Plan

Verify that the screenshots match the tickets. Make sure both the odd and the desktop version look correct

on odd:
Screen Shot 2023-08-10 at 4 02 09 PM
Screen Shot 2023-08-09 at 10 31 03 AM

on desktop:
Screen Shot 2023-08-10 at 1 22 29 PM

Also, make sure the loading screen has no scroll and if you early exit out of a flow in the ODD make sure it goes to the correct instruments page.

Screen Shot 2023-08-10 at 4 08 12 PM

# Changelog

- added stylings to the different buttons and made sure it only affected the onDeviceDisplay modal
- tweak onClose logic for early exiting out of flows

# Review requests

see test plan

# Risk assessment
low

### DIFF
--- a/app/src/molecules/SimpleWizardBody/index.tsx
+++ b/app/src/molecules/SimpleWizardBody/index.tsx
@@ -27,7 +27,14 @@ interface Props extends StyleProps {
   children?: React.ReactNode
   subHeader?: string | JSX.Element
   isPending?: boolean
+  /**
+   *  change justifyContent of OnDeviceDisplay buttons
+   *  TODO(jr, 8/9/23): this should really be refactored though so the
+   *  buttons' justifyContent is specified at the parent level
+   */
+  justifyContentForOddButton?: string
 }
+
 const BACKGROUND_SIZE = '47rem'
 
 const HEADER_STYLE = css`
@@ -90,7 +97,8 @@ export function SimpleWizardBody(props: Props): JSX.Element {
     padding-bottom: ${SPACING.spacing32};
 
     @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
-      justify-content: ${props.justifyContent ?? JUSTIFY_SPACE_BETWEEN};
+      justify-content: ${props.justifyContentForOddButton ??
+      JUSTIFY_SPACE_BETWEEN};
       padding-bottom: ${SPACING.spacing32};
       padding-left: ${SPACING.spacing32};
     }
--- a/app/src/molecules/SimpleWizardBody/index.tsx
+++ b/app/src/molecules/SimpleWizardBody/index.tsx
@@ -55,18 +55,6 @@ const SUBHEADER_STYLE = css`
     margin-right: 4.5rem;
   }
 `
-const BUTTON_STYLE = css`
-  width: 100%;
-  justify-content: ${JUSTIFY_FLEX_END};
-  padding-right: ${SPACING.spacing32};
-  padding-bottom: ${SPACING.spacing32};
-
-  @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
-    justify-content: ${JUSTIFY_SPACE_BETWEEN};
-    padding-bottom: ${SPACING.spacing32};
-    padding-left: ${SPACING.spacing32};
-  }
-`
 const WIZARD_CONTAINER_STYLE = css`
   min-height: 394px;
   flex-direction: ${DIRECTION_COLUMN};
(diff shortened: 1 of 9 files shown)


koji:
Screenshot 2023-08-09 at 10 50 52 AM

Do you think you can fix the scrollbar in this PR?
I think we don't need to display it in a wizard.

koji:
Screenshot 2023-08-09 at 10 53 58 AM

From the design, the button should be in the right side.

app/src/molecules/SimpleWizardBody/index.tsx
```diff
@@ -95,6 +83,19 @@ export function SimpleWizardBody(props: Props): JSX.Element {
...
   const isOnDevice = useSelector(getIsOnDevice)
 
+  const BUTTON_STYLE = css`
+    width: 100%;
+    justify-content: ${JUSTIFY_FLEX_END};
+    padding-right: ${SPACING.spacing32};
+    padding-bottom: ${SPACING.spacing32};
+
+    @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
+      justify-content: ${props.justifyContent ?? JUSTIFY_SPACE_BETWEEN};
```
koji: interesting approach 👍

jerader: i can make a new prop specifically for this instead of using a style prop! wasn't sure the best way

koji:
> Screenshot 2023-08-09 at 10 53 58 AM
> 
> From the design, the button should be in the right side.


https://github.com/Opentrons/opentrons/assets/474225/13464275-2c1d-46d7-97c0-0d6d76c7855e

I guess there would be two approaches for this.

1. move Exit to the right col space and flex-end
2. change the layout for touchscreen (1 col for exit button)

app/src/molecules/SimpleWizardBody/index.tsx
koji: I saw one component is passing justify-content so I think we don't really need to create a new prop.

@b-cooper 
Any thoughts on this?

jerader:
> 1. move Exit to the right col space and flex-end
> 2. change the layout for touchscreen (1 col for exit button)

thanks @koji , i'll probably move it to `flex-end` like i did with the other modals. Thanks for noticing that this button is misaligned, I missed it!

also I'll make sure that other modal doesn't have a scroll.

koji:
> > 1. move Exit to the right col space and flex-end
> > 2. change the layout for touchscreen (1 col for exit button)
> 
> thanks @koji , i'll probably move it to `flex-end` like i did with the other modals. Thanks for noticing that this button is misaligned, I missed it!
> 
> also I'll make sure that other modal doesn't have a scroll.

feel free to ping me to check your update since a dev kit may not work well.

app/src/molecules/SimpleWizardBody/index.tsx
jerader: I think the best route is to remove hard coded `justifyContent`s in `BUTTON_STYLE` and specify it at the parent level. But there are a bunch of components to go through and change that for.

jerader: Posted about this in slack and here is a follow up ticket to refactor this component later: https://opentrons.atlassian.net/browse/RAUT-622

app/src/organisms/InstrumentMountItem/AttachedInstrumentMountItem.tsx
```diff
@@ -98,7 +98,7 @@ export function AttachedInstrumentMountItem(
                 setShowChoosePipetteModal(false)
               },
               onComplete: () => {
-                history.push(`/instruments/${mount}`)
+                history.push(`/instruments`)
```
smb2268: We still want this to redirect to `instruments/mount` if the instrument is attached properly. Can we change these `onComplete` functions to check for the presence of an instrument on the given mount and redirect based on that?

app/src/organisms/InstrumentMountItem/AttachedInstrumentMountItem.tsx
```diff
@@ -55,7 +55,7 @@ export function AttachedInstrumentMountItem(
         flowType: GRIPPER_FLOW_TYPES.ATTACH,
         attachedGripper: attachedInstrument,
         onComplete: () => {
-          history.push(`/instruments/${mount}`)
+          history.push(`/instruments`)
```
koji: If you could add a test for this, that would be great.

koji (approved):
I think once you address Sarah's feedback, this PR will be good to go.

app/src/organisms/InstrumentMountItem/AttachedInstrumentMountItem.tsx
jerader: I think the logic is in there already in the `if/else if` statement searching for if `attachedInstruments` is null or not but i can add extra logic just to make sure

PR #13261, opened 2023-08-09
feat(hardware-testing): add upgrades for improved 8 channel QC tests

ryanthecoder:
# Overview
In order to improve throughput in the QC test at the factory we have re-worked the tests so that they can load all of the tips in at once for the gravimetric tests. There is a new deck layout for the 8 channel QC test in order to get all the tips onto the deck at once. 
In addition to the new layout we also need to load the racks slightly differently, the pattern is the same but the starting row has shifted to the center of the rack by one row. Instead of loading B,E,G,H and A,B,D,F for the rear channels, and front channels respectively, they will now need to be loaded A, D, F,G and B,C,E,H. The last change to the tips is that there can't be any extra unused tips loaded into the rack before the run. The must be loaded with just the tips that the test will use to avoid collisions. While this is not technically true for every single rack I believe that it's just good practice to do this anyway.

The output CVS now contains all of the tip-volume combos in one report. with the normal QC volumes always running first, then the "extra" QC volumes running after on the subset of pipettes that will run that test.

This PR also removes the need to replace tips mid run, it will just load and use extra tips for the first channel water level probe and the operator won't need to open and close the door at all.

https://opentrons.atlassian.net/browse/RH-11
https://opentrons.atlassian.net/browse/RH-19
https://opentrons.atlassian.net/browse/RH-18
https://opentrons.atlassian.net/browse/RH-17

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2194,6 +2194,7 @@ async def liquid_probe(
 
         pos = await self.gantry_position(mount, refresh=True)
         probe_start_pos = pos._replace(z=probe_settings.starting_mount_height)
+        print(f"start pos {probe_start_pos}")
```
Laura-Danielle: Do we need this print statement and or can we make it a log statement?

hardware-testing/hardware_testing/gravimetric/workarounds.py
```diff
@@ -44,7 +43,7 @@ def http_get_all_labware_offsets() -> List[dict]:
     runs_response = urlopen(req)
     runs_response_data = runs_response.read()
     stop_server_ot3()
-    atexit_register(start_server_ot3)
+    # atexit_register(start_server_ot3)
```
Laura-Danielle: Do you just want to delete this?

Laura-Danielle (approved):
Quibble about print statement in ot3api.py and one commented out line. Otherwise lgtm.

PR #13264, opened 2023-08-09
feature(robot-server): GET ODD logs from the /health and /logs

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/RSS-288.
adds ODD logs to the `/health` if this is an OT-3. GET logs from `/logs`.

# Test Plan

- Tested on the OT-3 via POSTMAN. GET /health returns logs/odd.log in the `links` and in `logs`.
- Verified with integration tests that it's not returning ODD logs for the OT-2. 
- Verified that `/logs/odd.log` return the actual odd logs on an OT-3. 

# Changelog

- append `/logs/odd.log` to `links` and `logs` if this is an OT-3. 
- changed `Health` to inherit from `BaseResponseBody` in order to exclude none in the response. 
- added logic to `log_control.py` to use -u when getting the odd logs. 

# Risk assessment

low. added logic for OT-3 only.


robot-server/robot_server/health/router.py
```diff
@@ -122,6 +126,23 @@ async def get_health(
...
+    )
+    health_links = HealthLinks(
+        apiLog="/logs/api.log",
+        serialLog="/logs/serial.log",
+        serverLog="/logs/server.log",
+        apiSpec="/openapi.json",
+        systemTime="/system/time",
+    )
+    if robot_model == "OT-3 Standard":
+        if OOD_LOG_PATH not in LOG_PATHS:
```
TamarZanzouri: should we use set instead?

robot-server/robot_server/health/router.py
```diff
@@ -122,6 +126,23 @@ async def get_health(
     and ready to operate. A 200 OK response means the server is running.
     The response includes information about the software and system.
     """
+    robot_model: RobotModel = (
+        "OT-3 Standard"
+        if config.feature_flags.enable_ot3_hardware_controller()
```
SyntaxColoring: Let's do a `Depends` on `robot_server.hardware.get_robot_type` for this instead, if we can. Like this one: 
https://github.com/Opentrons/opentrons/blob/f9d4a8b07e17b847a3e4b90e5c52a04a41c78d25/robot-server/robot_server/service/legacy/routers/settings.py#L193.

robot-server/robot_server/health/router.py
```diff
@@ -13,12 +13,16 @@
...
+
+from opentrons_shared_data.deck.dev_types import RobotModel
+
 from .models import Health, HealthLinks
 
 _log = logging.getLogger(__name__)
 
 LOG_PATHS = ["/logs/serial.log", "/logs/api.log", "/logs/server.log"]
 VERSION_PATH = "/etc/VERSION.json"
+OOD_LOG_PATH = "/logs/odd.log"
```
SyntaxColoring: Typo: `OOD` instead of `ODD`

robot-server/robot_server/health/router.py
```diff
@@ -122,6 +126,23 @@ async def get_health(
...
+    health_links = HealthLinks(
+        apiLog="/logs/api.log",
+        serialLog="/logs/serial.log",
+        serverLog="/logs/server.log",
+        apiSpec="/openapi.json",
+        systemTime="/system/time",
+    )
+    if robot_model == "OT-3 Standard":
+        if OOD_LOG_PATH not in LOG_PATHS:
+            LOG_PATHS.append(OOD_LOG_PATH)
```
SyntaxColoring: Since `LOG_PATHS` is a module-level constant, I think there's a strong expectation of immutability. Let's not append to it from this router function.

I guess we could just have two constants, `OT2_LOG_PATHS` plus `FLEX_LOG_PATHS`.

SyntaxColoring (commented):
Nice. 🪵

robot-server/robot_server/health/router.py
koji: I'm not sure that `odd.log` is a proper name since I guess Ed and Emily said that ODD is an internal terminology.
https://opentrons.slack.com/archives/CSCLVUW3C/p1685721715836799?thread_ts=1681324376.043909&cid=CSCLVUW3C

TamarZanzouri: do we want to call it `touchscreen.log`? isn't this internal?

koji: @ecormany, do you have any suggestion on this?

ecormany: Would we expect users to interact with this file? For example, is it included in downloading logs that they would send to support? If so, `touchscreen.log` is probably better. But if other parts of the software expect a file named `odd.log`, it's probably not worth changing it right now.

TamarZanzouri: its a small change. I can do it anyway!

robot-server/tests/service/legacy/routers/test_logs.py
```diff
@@ -177,8 +177,9 @@ async def mock_get_records_dumb(identifier, records, format_type):
 
     with patch("opentrons.system.log_control.get_records_dumb") as m:
         m.side_effect = mock_get_records_dumb
-        response = api_client.get("/logs/app.log")
+        response = api_client.get("/logs/touchscreen.log")
         body = response.text
+        print(body)
```
SyntaxColoring:
```suggestion
```

SyntaxColoring (approved):
Looks good to me. Thanks!

PR #13269, opened 2023-08-09
fix(app): Remove home, change to extension mount for gripper cal

smb2268:
fix [RQA-1256](https://opentrons.atlassian.net/browse/RQA-1256)

# Overview
We need to use the `extension` mount param when calling `moveToMaintenancePosition` for the gripper. We can also update the `home` command we send directly before calibration to use the `extensionZ` and `extensionJaw` commands instead of homing the whole gantry.

# Test Plan
Run through gripper calibration, ensure left pipette doesn't drop and gripper doesn't lower

# Changelog
1. Change mount from `left` to `extension`
2. Update `home` params before `calibrateGripper`
3. Update tests

# Review requests
See test plan

# Risk assessment
Low

[RQA-1256]: https://opentrons.atlassian.net/browse/RQA-1256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/GripperWizardFlows/BeforeBeginning.tsx
+++ b/app/src/organisms/GripperWizardFlows/BeforeBeginning.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 import { UseMutateFunction } from 'react-query'
 import { Trans, useTranslation } from 'react-i18next'
 import { COLORS } from '@opentrons/components'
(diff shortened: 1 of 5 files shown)


TamarZanzouri (approved):
LGTM! thank you for making this change :-)

PR #13274, opened 2023-08-10
fix(hardware-testing): Pipette QC, forgot to include deck-capacitance in overal cap pass/fail

andySigler:
# Overview

The test where it measures the capacitance when touching the deck wasn't being factored into the overall pass/fail for that pipette, resulting in some pipettes failing that test but saying overall they passed.

# Test Plan

# Changelog

# Review requests

# Risk assessment


djgomez-opentrons (approved):
Approved

PR #13276, opened 2023-08-10
fix(protocol-designer): error handling in top-selector for switching …

jerader:
…between robot types

closes RQA-1206

# Overview

When you create a protocol with 1 robot and then go back and make a new protocol with the other robot, when you go to the `design` tab, PD will white screen. But this PR fixes that by ensuring that the labwareId isn't null. 

# Test Plan

follow the steps in the ticket to replicate the white screen (the ticket identifies a timeline error but i wasn't able to replicate that) - make sure PD doesn't white screen though!

# Changelog

return null if `labwareId` is not in the `labwareEntities`

# Review requests

see test plan

# Risk assessment

low


koji (approved):
The changes look good to me.
Followed the steps in the ticket and the app didn't go white.

PR #13277, opened 2023-08-10
add sensor id param to tip presence query

caila-marashaj:
Accompanying Python changes for [this firmware branch](https://github.com/Opentrons/ot3-firmware/pull/713)- just adds a `sensor_id` parameter to the `TipStatusQueryRequestPayload`.

### DIFF
--- a/hardware/opentrons_hardware/firmware_bindings/messages/payloads.py
+++ b/hardware/opentrons_hardware/firmware_bindings/messages/payloads.py
@@ -542,6 +542,7 @@ class PushTipPresenceNotificationPayload(EmptyPayload):
     """A notification that the ejector flag status has changed."""
 
     ejector_flag_status: utils.UInt8Field
+    sensor_id: SensorIdField
 
 
 @dataclass(eq=False)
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Looks good but I think either
- we should allow and document that you can send this _without_ the new parameter, in which case the pipette will respond with data for all the sensors it has
or
- we should add an ALL entry to the enum
or
- we should just keep it payload free

hardware/opentrons_hardware/firmware_bindings/messages/payloads.py
```diff
@@ -542,6 +542,7 @@ class PushTipPresenceNotificationPayload(EmptyPayload):
     """A notification that the ejector flag status has changed."""
 
     ejector_flag_status: utils.UInt8Field
+    sensor_id: SensorIdField
```
sfoster1, line 575: we probably want to handle this being not present, or at least test with an old firmware, because when you attach a pipette with an out of date firmware it will boot and might send a notification without this early.

sfoster1 (changes requested):
Looks good (although I think we're going to remove the `TipStatusQueryRequest` payload change?) but we should probably make this parse ok if the sensor id field is missing just in case.

hardware/opentrons_hardware/firmware_bindings/messages/payloads.py
caila-marashaj: the newest commit now works with old firmware 👍

sfoster1 (approved):
Looks good to me, thanks!

PR #13279, opened 2023-08-10
feat(api): leave home-after-drop-tip to hw

sfoster1:
The OT-2 and the Flex have different default behaviors in hardware for how they should handle homing after drop tip. That was getting overridden by behavior in the protocol engine itself. Leave the defaults to the hardware. This will let the flex avoid homing after dropping a tip, which should certainly be possible and will speed up protocol execution a lot.

# Risk: Medium
This is pretty central to correct protocol execution but also that means it's pretty easy to test! And we have stall detection! So we should put this on a robot and run it through a protocol and see if it works.


Laura-Danielle (approved):
Definitely agree with this change.

TamarZanzouri (approved):
LGTM!

PR #13280, opened 2023-08-10
fix(app): wire up gripper flow to odd run setup

smb2268:
fix RQA-1266

# Overview
Gripper wizard flows were never wired up to ODD run setup step so pipette flows were being launched for the gripper. This PR fixes that and also deletes maintenance runs at the end of gripper flows so the "Remote activity in progress" modal closes when we end a flow on desktop.

# Test Plan
1. Launch gripper attach and calibrate flow from run setup and see that the correct wizard is rendered on ODD
2. Run any gripper wizard flow from desktop, see that remote activity modal on ODD closes shortly after the flow is closed on desktop

# Changelog
1. Launch gripper wizard flow for extension mount instrument in `ProtocolInstrumentMountItem`
2.  Delete maintenance run in `handleCleanUpAndClose` gripper wizard function
3.  Update tests accordingly

# Review requests
See test plan

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/GripperWizardFlows/index.tsx
+++ b/app/src/organisms/GripperWizardFlows/index.tsx
@@ -13,6 +13,7 @@ import {
 import {
   useCreateMaintenanceCommandMutation,
   useCreateMaintenanceRunMutation,
+  useDeleteMaintenanceRunMutation,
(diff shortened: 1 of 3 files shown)


jerader (approved):
lgtm!

PR #13281, opened 2023-08-10
fix(robot-server): move update animation controls to light task

fsinapi:
# Overview

The original implementation of the status bar animations during firmware updates was in the Firmware Update Manager because the light control task didn't start up until the engine store was created by an HTTP request. Now that the light control task is kicked off on startup, we can move all of the status bar animations to it.

This fixes an issue where pressing an Estop during an update would override the Firmware Update Manager animation; once the estop was released, the status bar would go back to IDLE instead of UPDATING.

# Test Plan

Restarted the protocol engine after pushing new firmware. While the firmware was updating I pressed & released the estop a few times. The status bar changed animations as expected, turning red while the estop was activate and going back to pulsing once it was released. At the end of the firmware updates, the startup animation (quick flash of blue) played as expected and the status bar changed to IDLE.

Also tested running a firmware update after attaching a gripper, same thing worked fine.

# Changelog

- Added the ability to differentiate post-init callbacks as either before or after the `hardware.py`-post-init task, in order to tell the light control task that the hardware is fully done initializing.
- Added logic in the Light Control task to indicate firmware updates on the status bar
- Removed all status bar functionality from the Firmware Update Manager
- Added a stub implementation of `attached_subsystems` for the OT2 API, since it's included in the base Hardware Controller api spec
- Added tests for new light controller functionality

# Review requests

# Risk assessment

Medium but it's very easy to test that this is working correctly by looking at the status bar.

### DIFF
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -51,6 +51,8 @@ from .types import (
     PauseType,
     StatusBarState,
     EstopState,
+    SubSystem,
+    SubSystemState,
 )
 from .errors import (
     MustHomeError,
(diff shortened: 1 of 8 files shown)


sfoster1 (approved):
This makes sense to me!!

PR #13284, opened 2023-08-11
feat(robot-server): Sort robot settings by robot type

Laura-Danielle:
# Overview

We should sort feature flags by robot type instead of constantly displaying the list.

# Test Plan
Throw this PR on both a FLEX and OT2. See that the feature flags for each respective robot show. Ensure that restart for certain feature flags still works (like disabling home on boot)

# Changelog

- Add a robot type to the SettingsDefinition
- Filter robot type by endpoint

# Review requests

Test on a robot. Check that implementation is OK.

# Risk assessment

Low. Only dealing with advanced feature flags.

### DIFF
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -259,6 +259,17 @@ def get_adv_setting(setting: str, robot_type: RobotTypeEnum) -> Optional[Setting
     return s.get(setting, None)
 
 
+def _filtered_key(
+    definition: SettingDefinition,
+    robot_type: RobotTypeEnum,
+    include_internal: bool,
+) -> bool:
+    if include_internal:
+        return robot_type in definition.robot_type
+    else:
+        return robot_type in definition.robot_type and not definition.internal_only
+
+
 @lru_cache(maxsize=1)
 def get_all_adv_settings(
     robot_type: RobotTypeEnum, include_internal: bool = False
@@ -271,9 +282,7 @@ def get_all_adv_settings(
     return {
         key: Setting(value=value, definition=settings_by_id[key])
         for key, value in values.items()
-        if key in settings_by_id
-        and robot_type in settings_by_id[key].robot_type
-        and not (include_internal and settings_by_id[key].internal_only)
+        if key in settings_by_id and _filtered_key(settings_by_id[key], robot_type, include_internal)
     }
 
 
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -77,7 +77,10 @@ class SettingDefinition:
         """
         if not self.show_if:
             return True
-        return get_setting_with_env_overload(self.show_if[0], self.robot_type[0]) == self.show_if[1]
+        return (
+            get_setting_with_env_overload(self.show_if[0], self.robot_type[0])
+            == self.show_if[1]
+        )
 
     async def on_change(self, value: Optional[bool]) -> None:
         """
@@ -96,6 +99,7 @@ class DisableLogIntegrationSettingDefinition(SettingDefinition):
             description="Prevent the robot from sending its logs to Opentrons"
             " for analysis. Opentrons uses these logs to"
             " troubleshoot robot issues and spot error trends.",
+            robot_type=[RobotTypeEnum.OT2, RobotTypeEnum.FLEX],
         )
 
     async def on_change(self, value: Optional[bool]) -> None:
@@ -129,7 +133,7 @@ settings = [
         old_id="short-fixed-trash",
         title="Short (55mm) fixed trash",
         description="Trash box is 55mm tall (rather than the 77mm default)",
-        robot_type=[RobotTypeEnum.OT2]
+        robot_type=[RobotTypeEnum.OT2],
     ),
     SettingDefinition(
         _id="deckCalibrationDots",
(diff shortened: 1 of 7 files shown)


api/src/opentrons/config/advanced_settings.py
```diff
@@ -154,6 +166,7 @@ class Setting(NamedTuple):
         "Opening the robot door during a run will "
         "pause your robot only after it has completed its "
         "current motion.",
+        robot_type=[RobotTypeEnum.OT2, RobotTypeEnum.FLEX],
```
sfoster1: this one is OT-2 only also - because of the 96, we're not going to let users change it (and need it to default to true)

api/src/opentrons/config/advanced_settings.py
```diff
@@ -175,21 +189,25 @@ class Setting(NamedTuple):
             "new hardware."
         ),
         restart_required=True,
+        robot_type=[RobotTypeEnum.FLEX],
```
sfoster1, line 196: Oof I forgot we haven't removed this, gotta get there

api/src/opentrons/config/advanced_settings.py
```diff
@@ -175,21 +189,25 @@ class Setting(NamedTuple):
...
     ),
     SettingDefinition(
         _id="rearPanelIntegration",
         title="Enable robots with the new usb connected rear-panel board.",
         description="This is an Opentrons-internal setting to test new rear-panel.",
+        robot_type=[RobotTypeEnum.FLEX],
```
sfoster1, line 203: Ditto

api/src/opentrons/config/advanced_settings.py
```diff
@@ -199,6 +217,19 @@ class Setting(NamedTuple):
             "estopNotRequired",
             True,
         ),  # Configured so this only shows if it has been set by a user
+        robot_type=[RobotTypeEnum.FLEX],
```
sfoster1, line 226: Definitely don't show this either, it's internal only

sfoster1 (changes requested):
I think we want to separate the concepts of (1) this setting should be _exposed to the http api_ on only certain platforms and (2) this setting should only _exist_ on certain platforms. For instance, estop settings only exist on flex, but shouldn't be shown there either; home on boot exists on both, but shouldn't be shown on flex; things like that.

api/src/opentrons/config/advanced_settings.py
```diff
@@ -175,21 +193,27 @@ class Setting(NamedTuple):
...
         restart_required=True,
+        robot_type=[RobotTypeEnum.FLEX],
+        interal_only=True,
     ),
     SettingDefinition(
         _id="rearPanelIntegration",
         title="Enable robots with the new usb connected rear-panel board.",
         description="This is an Opentrons-internal setting to test new rear-panel.",
+        robot_type=[RobotTypeEnum.FLEX],
+        interal_only=True,
```
sfoster1:
```suggestion
        internal_only=True,
```

api/src/opentrons/config/advanced_settings.py
```diff
@@ -44,9 +46,11 @@ def __init__(
         _id: str,
         title: str,
         description: str,
+        robot_type: List[RobotTypeEnum],
         old_id: Optional[str] = None,
         restart_required: bool = False,
         show_if: Optional[Tuple[str, bool]] = None,
+        interal_only: bool = False,
```
sfoster1:
```suggestion
        internal_only: bool = False,
```

api/src/opentrons/config/advanced_settings.py
```diff
@@ -62,6 +66,10 @@ def __init__(
         self.show_if = show_if
         #: A tuple of (other setting id, setting value) that must match reality
         #: to show this setting in http endpoints
+        self.robot_type = robot_type
+        #: A list of RobotTypeEnums that are compatible with this feature flag.
+        self.internal_only = interal_only
```
sfoster1:
```suggestion
        self.internal_only = internal_only
```

api/src/opentrons/config/advanced_settings.py
```diff
@@ -175,21 +193,27 @@ class Setting(NamedTuple):
             "new hardware."
         ),
         restart_required=True,
+        robot_type=[RobotTypeEnum.FLEX],
+        interal_only=True,
```
sfoster1:
```suggestion
        internal_only=True,
```

api/src/opentrons/config/advanced_settings.py
```diff
@@ -199,6 +223,20 @@ class Setting(NamedTuple):
             "estopNotRequired",
             True,
         ),  # Configured so this only shows if it has been set by a user
+        robot_type=[RobotTypeEnum.FLEX],
+        interal_only=True,
```
sfoster1:
```suggestion
        internal_only=True,
```

api/src/opentrons/config/advanced_settings.py
```diff
@@ -232,6 +270,8 @@ def get_all_adv_settings() -> Dict[str, Setting]:
         key: Setting(value=value, definition=settings_by_id[key])
         for key, value in values.items()
         if key in settings_by_id
+        and robot_type in settings_by_id[key].robot_type
+        and not settings_by_id[key].internal_only
```
sfoster1: this means the settings won't be readable internally either since the functions used to get feature flags call `get_adv_setting_with_env_overload` which calls `get_adv_setting` which calls this to filter the results

sfoster1 (changes requested):
Typo, plus a structural problem with `internal_only`

TamarZanzouri (approved):
Thank god for RobotTypeEnum :-) I like this approach! LGTM once you fix Seth's comment about filtering the internal only.

sfoster1 (approved):
Looks good to me!

PR #13285, opened 2023-08-11
feat(hardware-testing): adds Z-Stage production QC script

andySigler:
# Overview

Script where a Z-Stage sub-assembly presses down into a force-guage, to check that all our Z-Stages are pressing with similar forces. This happens before Z-Stage is attached to gantry.

# Test Plan

# Changelog

# Review requests

# Risk assessment


djgomez-opentrons (approved):
Tested on Z-stage test fixture

PR #13286, opened 2023-08-11
chore: deploy app to new infra

sfoster1:
This PR changes the app to deploy to builds.opentrons.com instead of the legacy bucket.

It also refactors the app build workflow in some ways that should make it a little more pleasant to deal with:
- Separate backend tests from builds
- Reduce duplication between internal release and release builds
- Make the logic about what kind of build we have clearer

# Risk
Medium - we have time to work this out.

# Testing
The testing is entirely about deploying app builds to various places - both variants, develop and release. All this should now be safe to do because this workflow no longer contains the public release bucket location. You can trigger builds on specific refs via manual dispatch.
- [x] Deploy some internal release develop app builds and check that they go to ot3-development.builds.opentrons.com and get slack notifications
- [x] Deploy some internal release release app builds (this might be tough, may have to debug at a later date unfortuntely) and check that they go to ot3-development.builds.opentrons.com and get slack notifications and get the yml payloads
- [x] Deploy some release develop app builds and check that they go to builds.opentrons.com and get slack notifications
- [x] Deploy some release release app builds and check that they go to builds.opentrons.com and get slack notifications and yml payloads

### DIFF
--- a/.github/workflows/app-test-build-deploy.yaml
+++ b/.github/workflows/app-test-build-deploy.yaml
@@ -289,7 +289,7 @@ jobs:
           APPLE_APP_SPECIFIC_PASSWORD: ${{ secrets.OT_APP_APPLE_ID_PASSWORD }}
           APPLE_TEAM_ID: ${{ secrets.OT_APP_APPLE_TEAM_ID }}
           HOST_PYTHON: python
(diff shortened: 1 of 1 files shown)


sfoster1:
This action: https://github.com/Opentrons/opentrons/actions/runs/5836723414/job/15832369784
Produced non-release builds, deployed them to the right place, and sent notifications!

shlokamin (approved):
lgtm

PR #13287, opened 2023-08-11
docs(api): Revisions to Positions Within the OT-2

jwwojak:
# Overview

This is a revision to the API doc, [Positions Within the OT-2](https://docs.opentrons.com/v2/robot_position.html) (`robot_position.rst`).

We're revising this doc to include Flex and make it more generic to cover both robot types.

# Test Plan

- Review text for grammar, style, accuracy.
- Check code samples for accuracy.
- Sandbox: http://sandbox.docs.opentrons.com/position-within-robot-revisions/v2/

# Changelog

- Title change to "Labware and Deck Positions" (suggested) from "Positions Within the OT-2", which is too robot-specific.
- Check and fix all section header styles.
- Text and code changes throughout.
- Removing mentions of specific models in favor of more general terminology  like "robot," "the robot," "your robot" etc.

# Review requests

No particular requests. Appreciate any comments and corrections as always.

# Risk assessment

Low. Text changes only.


api/docs/v2/robot_position.rst
```diff
@@ -2,33 +2,35 @@
...
-########################
-Position Within the OT-2
-########################
-
-Most of the time when executing a protocol, the Python Protocol API's methods take care of determining where the robot needs to go to perform the commands you've given. But sometimes you need to modify how the robot moves in order to achieve the purposes of your protocol. This document will cover the two main ways to define positions — relative to labware, or relative to the entire deck — as well as how to alter the robot's speed or trajectory as it moves to those positions.
+##########################
+Labware and Deck Positions
+##########################
 
+The API automatically determines how the robot needs to move when working with the instruments and labware in your protocol. But sometimes you need direct control over these activities. The API lets you do just that. Specifically, you can control movements relative to labware and deck locations. You can also manage the gantry’s speed and trajectory as it traverses the enclosure. This document explains how to use API commands to take direct control of the robot and position it exactly where you need it.
```
ecormany: "working area" instead of "enclosure"?

api/docs/v2/robot_position.rst
```diff
@@ -2,33 +2,35 @@
...
 .. _robot-position:
 
-########################
-Position Within the OT-2
-########################
-
-Most of the time when executing a protocol, the Python Protocol API's methods take care of determining where the robot needs to go to perform the commands you've given. But sometimes you need to modify how the robot moves in order to achieve the purposes of your protocol. This document will cover the two main ways to define positions — relative to labware, or relative to the entire deck — as well as how to alter the robot's speed or trajectory as it moves to those positions.
+##########################
+Labware and Deck Positions
+##########################
```
ecormany: asterisks for H1

api/docs/v2/robot_position.rst
```diff
@@ -2,33 +2,35 @@
...
 .. _position-relative-labware:
 
-****************************
+
 Position Relative to Labware
-****************************
+============================
 
-When you instruct the robot to move to a position on a piece of labware, the exact point in space it moves to is calculated based on the labware definition, the type of action the robot will perform there, and labware offsets for the specific deck slot on your robot. This section describes how each of these components of a position are calculated and methods for modifying them.
+When the robot positions itself relative to a piece of labware, the labware definition determines movement, distance, height, offsets, and the actions it can perform. This section describes how these positional components are calculated and how to change them.
```
ecormany: This over-collapses these concepts. The labware definition determines only a portion of the behavior.
- LPC determines labware offsets
- "type of action the robot will perform" determines whether the tip goes to the top or bottom of the well — it differs for aspirate and dispense.

api/docs/v2/robot_position.rst
```diff
@@ -2,33 +2,35 @@
...
 
 Top, Bottom, and Center
 =======================
 
-Every well on every piece of labware you load has three addressable positions — top, bottom, and center — that are determined by the labware definition and whether the labware is on a module or directly on the deck. You can use these positions as-is or calculate other positions relative to them.
+Every well on every piece of labware has three addressable positions: top, bottom, and center. The position is determined by the labware definition and whether the labware is on a module or in a deck slot. You can use these positions as-is or calculate other positions relative to them.
```
ecormany:
```suggestion
Every well on every piece of labware has three addressable positions: top, bottom, and center. The position is determined by the labware definition and what the labware is loaded on top of. You can use these positions as-is or calculate other positions relative to them.
```
Something like this? Generalizes now that we have the module/adapter split and `load_adapter()`.

api/docs/v2/robot_position.rst
```diff
@@ -2,33 +2,35 @@
...
 
-:py:meth:`.Well.top` returns a position level with the top of the well, centered in both horizontal directions. 
+Top
+---
 
-.. code-block:: python
+As an example, let's look at the :py:meth:`.Well.top` method. It returns a position level with the top of the well, centered in both horizontal directions.
```
ecormany: I don't think the "As an example" text does much besides making this longer to read.

api/docs/v2/robot_position.rst
```diff
@@ -37,29 +39,33 @@ This is a good position to use for :ref:`new-blow-out` or any other operation wh
...
+
+As an example, lets look at the :py:meth:`.Well.bottom` method. It returns a position level with the bottom of the well, centered in both horizontal directions. 
 
 .. code-block:: python
 
-   plate['A1'].bottom()     # the bottom center of the well
+   plate['A1'].bottom()  # the bottom center of the well
 
-This is a good position to start for aspiration or any other operation where you want the tip to contact the liquid. The same as with :py:meth:`.Well.top`, you can adjust the height of this position with the optional argument ``z``, which is measured in mm. Positive ``z`` numbers move the position up, and negative ``z``` numbers move it down:
+This is a good position to for :ref:`aspirating liquid <new-aspirate>` or an activity where you want the tip to contact the liquid. Similar to the ``Well.top()`` method, you can adjust the height of this position with the optional argument ``z``, which is measured in mm. Positive ``z`` numbers move the position up, negative ``z``` numbers move it down.
```
ecormany:
```suggestion
This is a good position for :ref:`aspirating liquid <new-aspirate>` or an activity where you want the tip to contact the liquid. Similar to the ``Well.top()`` method, you can adjust the height of this position with the optional argument ``z``, which is measured in mm. Positive ``z`` numbers move the position up, negative ``z``` numbers move it down.
```

api/docs/v2/robot_position.rst
```diff
@@ -73,133 +79,102 @@ This is a good position to start for aspiration or any other operation where you
...
 .. versionadded:: 2.0
 
-
 .. _using_lpc:
 
 Using Labware Position Check
 ============================
 
-All positions relative to labware are automatically adjusted based on the labware's offset, an x, y, z vector. The best way to calculate and apply these offsets is by using Labware Position Check when you run your protocol in the Opentrons App. As of version 6.0 of the app, you can apply previously calculated offsets — even across different protocols — as long as they are for the same type of labware in the same deck slot on the same robot.
+All positions relative to labware are adjusted automatically based on the labware's x, y, z axis offset. The best way to calculate and apply these offsets is by using Labware Position Check when you run your protocol in the Opentrons App. With v6.0 of the app, you can apply previously calculated offsets on the same robot for the same labware type and deck slot, even with different protocols.
```
ecormany: Let's make this cover Flex and the touchscreen as well.
```suggestion
All positions relative to labware are adjusted automatically based on labware offset data. Calculate labware offsets by running Labware Position Check during protocol setup, either in the Opentrons App or on the Flex touchscreen. Version 6.0.0 and later of the robot software can apply previously calculated offsets on the same robot for the same labware type and deck slot, even across different protocols.
```

api/docs/v2/robot_position.rst
```diff
@@ -73,133 +79,102 @@ This is a good position to start for aspiration or any other operation where you
...
 
-You shouldn't adjust labware offsets in your Python code if you plan to run your protocol in the app. However, if you are running your protocol in Jupyter notebook or with ``opentrons_execute``, Labware Position Check is not directly available. For these applications, you can calculate and apply labware offsets by:
+You shouldn't adjust labware offsets in your Python code if you plan to run your protocol in the app. However, if you are running your protocol in Jupyter notebook or with ``opentrons_execute``, the Labware Position Check is not directly available. For these applications, you can calculate and apply labware offsets by:
```
ecormany:
```suggestion
You should only adjust labware offsets in your Python code if you plan to run your protocol in Jupyter notebook or with ``opentrons_execute``. This is because you can't run Labware Position Check from Jupyter or the command line. For these applications, calculate and apply labware offsets by:
```

touchscreen makes this confusing. could add something about it but then it gets wordy. cutting to the chase instead.

api/docs/v2/robot_position.rst
```diff
@@ -73,133 +79,102 @@ This is a good position to start for aspiration or any other operation where you
...
-        p300 = protocol.load_instrument('p300_single_gen2', 'left', tip_racks=[tiprack])
-        p300.pick_up_tip()
-        p300.return_tip()
-		
-After importing this protocol to the Opentrons App, run Labware Position Check to get the x, y, and z offsets for the tip rack and labware. When complete, you can click **Get Labware Offset Data** to view automatically generated code that uses :py:meth:`.set_offset` to apply the offsets to each piece of labware:
+To prepare code written for Jupyter notebook and run it in the app, you need to specify an API level (see the :ref:`tutorial`) and put your commands in a ``run()`` function. To enable the Labware Position Check, you need to add a :py:meth:`.pick_up_tip` action for each pipette the protocol uses. For example, say you have a protocol that uses a Flex pipette, a tip rack, a 15 mL reservoir, and a 96-well plate. After importing this protocol to the Opentrons App, run Labware Position Check to get the x, y, and z offsets for the tip rack and labware. When complete, you can click **Get Labware Offset Data** to view automatically generated code that uses :py:meth:`.set_offset` to apply the offsets to each piece of labware.
```
ecormany: could also link to "Specifying Versions". would require adding an anchor to versioning.rst

ecormany: we basically never use the definite article "the" with "Labware Position Check" 📱

api/docs/v2/robot_position.rst
```diff
@@ -73,133 +79,102 @@ This is a good position to start for aspiration or any other operation where you
...
-	3. Running Labware Position Check
-	4. Adding the offsets to your protocol
+	1. Creating a "dummy" protocol that loads your labware and has each used pipette pick up a tip from a tip rack.
+	2. Importing the dummy protocol to the Opentrons App.
+	3. Running Labware Position Check.
+	4. Adding the offsets to your protocol.
 	
-To prepare code written for Jupyter notebook so it can be run in the app, you need to include a metadata block and a ``run()`` function. And to enable Labware Position Check, you need to add a :py:meth:`.pick_up_tip` action for each pipette the protocol uses. For example, a dummy protocol using a P300 Single-Channel pipette, a reservoir, and a well plate would look like this:
-
-.. code-block:: python
```
ecormany, line 131: We definitely need to keep this code block. This is the requisite "dummy protocol", without which you can't run LPC. 

I'm going to have a think about the broader purpose and structure of this section. I may put together an alternative in a gist or something.

api/docs/v2/robot_position.rst
```diff
@@ -73,133 +79,102 @@ This is a good position to start for aspiration or any other operation where you
...
 .. _protocol-api-deck-coords:
 
-*****************************
 Position Relative to the Deck
-*****************************
+=============================
 
-
-The OT-2’s base coordinate system is known as *deck coordinates*. Many API functions use this coordinate system, and you can also reference it directly. It is a right-handed coordinate system always specified in mm, with the origin ``(0, 0, 0)`` at the front left of the robot. The positive ``x`` direction is to the right, the positive ``y`` direction is to the back, and the positive ``z`` direction is up. 
+The robot's base coordinate system is known as *deck coordinates*. Many API functions use this coordinate system, and you can also reference it directly. It is a right-handed coordinate system always specified in mm, with the origin ``(0, 0, 0)`` at the front left of the robot. The positive ``x`` direction is to the right, the positive ``y`` direction is to the back, and the positive ``z`` direction is up.
```
ecormany, line 118: I'm double-checking that this system is the same for Flex and OT-2.

api/docs/v2/robot_position.rst
```diff
@@ -73,133 +79,102 @@ This is a good position to start for aspiration or any other operation where you
...
 
 
 Move To
-=======
+-------
 
-You can use the :py:meth:`.InstrumentContext.move_to` method to move a pipette to any reachable location on the deck. If the pipette has picked up a tip, it will move the end of the tip to that position; if it hasn't, it will move the pipette nozzle to that position. As with all movement in a protocol, the OT-2 calculates where to move in physical space by using its `pipette offset and tip length calibration <https://support.opentrons.com/s/article/Get-started-Calibrate-tip-length-and-pipette-offset>`_ data.
+The :py:meth:`.InstrumentContext.move_to` method moves a pipette to any reachable location on the deck. If the pipette has picked up a tip, it will move the end of the tip to that position; if it hasn't, it will move the pipette nozzle to that position. As with all movement in a protocol, the robot calculates where to move in physical space by using its `pipette offset and tip length calibration <https://support.opentrons.com/s/article/Get-started-Calibrate-tip-length-and-pipette-offset>`_ data.
```
ecormany: Let's strike the last sentence here. It's different for Flex and I don't think this level of detail is particularly illuminating. The prior sentence explains the behavior completely.

api/docs/v2/robot_position.rst
```diff
@@ -280,32 +244,26 @@ This distinction is important for the :py:meth:`.Location.move` method, which op
...
         
+	
+	pipette.move_to(plate['A1'].top())  # move to the first well at default speed
+	pipette.default_speed = 100         # reduce pipette speed
+	pipette.move_to(plate['D6'].top())  # move to the last well at the slower speed
+
 .. warning::
 
-	The default of 400 mm/s was chosen because it is the maximum speed Opentrons knows will work with the gantry. Your specific robot may be able to move faster, but you shouldn't increase this value above 400 unless instructed by Opentrons Support.
+	These default speeds were chosen because they're the maximum speed Opentrons knows will work with the gantry. Your robot may be able to move faster, but you shouldn't increase this value unless instructed by Opentrons Support.
```
ecormany:
```suggestion
	These default speeds were chosen because they're the maximum speeds that Opentrons knows will work with the gantry. Your robot may be able to move faster, but you shouldn't increase this value unless instructed by Opentrons Support.
```
agreement + overt clause boundary

api/docs/v2/robot_position.rst
```diff
@@ -314,17 +272,17 @@ The OT-2's gantry usually moves as fast as it can given its construction: 400 mm
...
 In addition to controlling the overall gantry speed, you can set speed limits for each of the individual axes: ``x`` (gantry left/right motion), ``y`` (gantry forward/back motion), ``z`` (left pipette up/down motion), and ``a`` (right pipette up/down motion). Unlike ``default_speed``, which is a pipette property, axis speed limits are stored in a protocol property :py:obj:`.ProtocolContext.max_speeds`; therefore the ``x`` and ``y`` values affect all movements by both pipettes. This property works like a dictionary, where the keys are axes, assigning a value to a key sets a max speed, and deleting a key or setting it to ``None`` resets that axis's limit to the default:
 
 .. code-block:: python
     :substitutions:
 
-	protocol.max_speeds['x'] = 50       # limit x-axis to 50 mm/s
-	del protocol.max_speeds['x']        # reset x-axis limit
-	protocol.max_speeds['a'] = 10       # limit a-axis to 10 mm/s
-	protocol.max_speeds['a'] = None     # reset a-axis limit
+	protocol.max_speeds['x'] = 50    # limit x-axis to 50 mm/s
```
ecormany, line 243: yes, this spacing is better!

ecormany (changes requested):
This generally integrates Flex behavior into what was a very "OT-2" article before. There's a couple factual points to clean up, and I'll have a separate go at the Jupyter section.

api/docs/v2/robot_position.rst
ecormany: Confirmed with Seth, they're the same at the level of PAPI.

ecormany:
I made a gist with a possible rewrite of the Jupyter LPC instructions here. I think this is much more understandable. _Howeva_, I think this probably belongs better on Advanced Control, with a link from here. If we decide to do that, it can/should be part of this PR.

https://gist.github.com/ecormany/cdb8e512e5dc2cab561554d056c344a7

api/docs/v2/robot_position.rst
jwwojak: Yes! Great suggestion. Adding that.

api/docs/v2/robot_position.rst
jwwojak: Ah, rereading this, I understand now. Will revise, something akin to the original:

> When the robot positions itself relative to a piece of labware, where it moves is determined by the labware definition, the actions you want it to perform, and the labware offsets for a specific deck slot. This section describes how these positional components are calculated and how to change them.

api/docs/v2/robot_position.rst
```diff
@@ -71,135 +77,104 @@ This is a good position to start for aspiration or any other operation where you
...
+All positions relative to labware are adjusted automatically based on labware offset data. Calculate labware offsets by running Labware Position Check during protocol setup, either in the Opentrons App or on the Flex touchscreen. Version 6.0.0 and later of the robot software can apply previously calculated offsets on the same robot for the same labware type and deck slot, even across different protocols.
 
-You shouldn't adjust labware offsets in your Python code if you plan to run your protocol in the app. However, if you are running your protocol in Jupyter notebook or with ``opentrons_execute``, Labware Position Check is not directly available. For these applications, you can calculate and apply labware offsets by:
+You should only adjust labware offsets in your Python code if you plan to run your protocol in Jupyter notebook or with ``opentrons_execute``. This is because you can't run Labware Position Check from Jupyter or the command line. For these applications, calculate and apply labware offsets by:
```
ecormany: @jwwojak Here is where to link to Advanced Control (replacing the remainder of this section). Then take the text from my gist and put it in `new_advanced_running.rst` between the Running a Previously Written Protocol and Using Custom Labware sections.

ecormany (approved):
Looks good. There's one stray backtick and a little redundancy in the transfer of the one section to Advanced Control. I'll make one last push to clean that up and then :shipit:

PR #13288, opened 2023-08-11
feat(robot-server): block POSTing runs when estop is not disengaged

fsinapi:
# Overview

This PR adds a mechanism to return an error whenever a client attempts to POST to the `/runs` or `/maintenance_runs` endpoints if the Estop is not engaged or is not present.

These specific routes are blocked:
- POST `/maintenance_runs`
- POST `/maintenance_runs/{run_id}/commands`
- POST `/runs`
- POST `/runs/{runId}/commands`
- POST `/runs/{runId}/actions`

# Test Plan

Tested on FLEXY:
* Try performing a run with the estop in the normal state - it should work normally
* Try POSTing to the above endpoints with the estop in any of the invalid states, should get an error back

# Changelog

- Added FastAPI dependency to check that estop state is valid - returns True on OT-2 at all times
- Added aforementioned dependency to POST endpoints for runs & maintenance runs
- Added tests for the dependency

# Review requests

Any other endpoints we want to block out? We are specifically not including Module control (stateless commands) or sending protocols.

# Risk assessment

If this improperly blocks functionality that is Not Good.

### DIFF
--- a/robot-server/robot_server/errors/robot_errors.py
+++ b/robot-server/robot_server/errors/robot_errors.py
@@ -2,6 +2,7 @@
 
 from typing_extensions import Literal
 from .error_responses import ErrorDetails
+from opentrons_shared_data.errors.codes import ErrorCodes
 
 
 class InstrumentNotFound(ErrorDetails):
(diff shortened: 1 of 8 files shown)


sfoster1 (approved):
Nice, looks good to me!

PR #13289, opened 2023-08-11
feat(protocol-designer, step-generation): wire up hover labware and remove duplicate btn

jerader:
closes RQA-1273 and RQA-1272

# Overview

This PR wires up a few things related to the `MoveLabware` timeline step

# Test Plan

Sandbox: https://sandbox.designer.opentrons.com/pd_movelabware-timeline-fixes/

Check these 2 things:

- Create a Move Labware step in PD. When you highlight it in the timeline, make sure the correct labware is highlighted on the deck map
- right click on the Move Labware step in the timeline, there should only be the option to "delete" and **no** option to "duplicate" the step.

Bonus:
- if you add a H-S to the deck and then move a labware in/out of the H-S, there should be a timeline error that pops up saying the latch is closed. Previously, my logic was a bit wrong and didn't account for if the latch was null

In the screenshot, notice that only the Delete step option is available and in the deckmap, notice that the labware that is being moved is highlighted

Screen Shot 2023-08-11 at 1 45 17 PM

# Changelog

- add logic to tell what the step type is for the selected step in the timeline to remove the duplicate option if its a move labware step - this is done in `ContextMenu`
- add the highlighted labware to the deck map by adding a `moveLabware` step condition in the `getHoveredStepLabware` selector, add a test case
- tiny logic fix in `MoveLabware` atomic command

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/steplist/ContextMenu.tsx
+++ b/protocol-designer/src/components/steplist/ContextMenu.tsx
@@ -11,6 +11,7 @@ import { actions as steplistActions } from '../../steplist'
 import { Portal } from '../portals/TopPortal'
 import styles from './StepItem.css'
 import { StepIdType } from '../../form-types'
+import { getSavedStepForms } from '../../step-forms/selectors'
 
 const MENU_OFFSET_PX = 5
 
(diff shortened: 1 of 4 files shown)


koji (approved):
The sandbox worked as expected

PR #13290, opened 2023-08-11
fix(app, components, shared-data): update TC gen 2 rendering

shlokamin:
# Overview
This PR updates the TC gen 2 module definition's 2d rendering key to reflect its actual geometry. Renders of the TC gen 2 with labware on top should now look... normal! I also updated storybook to be a little more configurable so we can render different TC generations on different deck types.

closes RLAB-352

# Changelog

- Update TC gen 2 rendering

# Review requests

Upload the protocol below into the app that uses a TC gen 2 and loads labware on it (i've attached one below). You should see that labware gets rendered correctly. 

[TCgen2.py.zip](https://github.com/Opentrons/opentrons/files/12324268/TCgen2.py.zip)

# Risk assessment

Low


jerader (commented):
Is there a reason why you chose such a dark grey color for the render? or are we gonna modify the other modules to match that? The contrast in grey seems like a lot to me

Screen Shot 2023-08-11 at 3 06 13 PM

shlokamin:
good q — @mmencarelli decided the color! any insights here mel?

shlokamin:
update from slack: replace #989898 with #E6E6E6. just made this change

thank's for catching @jerader!

components/src/hardware-sim/Module/Module.stories.tsx
```diff
@@ -38,9 +42,10 @@ const Template: Story<{
   model: ModuleModel
   orientation: 'left' | 'right'
   hasLabware: boolean
+  deckType: typeof OT2_STANDARD_DECKID | typeof OT3_STANDARD_DECKID
```
b-cooper, line 45: We updated the name of the robot type constants to reflect the correct product name. This would be a good time to replace `OT3_STANDARD_MODEL` and `OT3_STANDARD_DECKID` with `FLEX_STANDARD_MODEL` and `FLEX_STANDARD_DECKID`.  The values can stay the same

b-cooper (approved):
🌡️ 🌀 🥈

components/src/hardware-sim/Module/Module.stories.tsx
shlokamin: good call, just changed those out

PR #13291, opened 2023-08-11
fix(app): fix NetworkSettings DQA

koji:
# Overview
This PR addresses RobotSettings NetworkingSettings comments
close RAUT-476

# Test Plan

# Changelog
- update RobotSettings components

# Review requests
I added position fixed and marginTop to fix wi-fi child navigation, but I feel that there would be more efficient ways to do this. So I'd like to hear about any ideas if you have a suggestion.

# Risk assessment
low


app/src/organisms/NetworkSettings/DisplayWifiList.tsx
```diff
@@ -24,6 +24,17 @@ import { DisplaySearchNetwork } from './DisplaySearchNetwork'
 import type { WifiNetwork } from '../../redux/networking/types'
 
 const NETWORK_ROW_STYLE = css`
+  display: ${DISPLAY_FLEX};
+  width: 100%;
+  height: '5rem';
```
jerader:
```suggestion
  height: 5rem;
```

app/src/organisms/RobotSettingsDashboard/NetworkSettings/WifiConnectionDetails.tsx
```diff
@@ -112,17 +118,18 @@ export function WifiConnectionDetails({
                 <Flex
                   alignItems={ALIGN_CENTER}
                   flexDirection={DIRECTION_ROW}
-                  gridGap={SPACING.spacing12}
+                  gridGap={SPACING.spacing12} // maybe need to add 10px as a const
```
jerader: does this comment mean this gridGap doesn't match designs?

app/src/organisms/RobotSettingsDashboard/NetworkSettings/WifiConnectionDetails.tsx
```diff
@@ -112,17 +118,18 @@ export function WifiConnectionDetails({
...
                     size="2.5rem"
                     name="info"
                     aria-label={`${activeSsid}_info_icon`}
+                    color={COLORS.darkBlack100}
                   />
                   <StyledText
                     as="p"
                     fontWeight={TYPOGRAPHY.fontWeightSemiBold}
-                    color={COLORS.darkBlack70}
+                    color={COLORS.darkBlack100}
```
jerader: do we need this? doesn't StyledText default to `darkBlack100`?

app/src/organisms/ChildNavigation/index.tsx
```diff
@@ -40,6 +41,11 @@ export function ChildNavigation({
       justifyContent={JUSTIFY_SPACE_BETWEEN}
       paddingX={SPACING.spacing40}
       paddingY={SPACING.spacing32}
+      position={POSITION_FIXED}
+      top="0"
+      left="0"
+      width="100%"
```
jerader, line 49: hmm i think this approach is okay

jerader (approved):
left some cleanup comments but lgtm

app/src/organisms/NetworkSettings/DisplayWifiList.tsx
koji: good catch, thanks!

app/src/organisms/RobotSettingsDashboard/NetworkSettings/WifiConnectionDetails.tsx
koji: Yeah, the design uses 10px but we don't have 10px as a const.
I think we would need to use constants as much as possible for the future DS unification.

app/src/organisms/RobotSettingsDashboard/NetworkSettings/WifiConnectionDetails.tsx
koji: oh good catch, we don't need `darkBalck100` for the text

app/src/organisms/ChildNavigation/index.tsx
koji: Do you have any better ideas on this?
Actually this isn't ideal since I needed to add marginTop to some components since adding marginBottom didn't work.

app/src/organisms/RobotSettingsDashboard/NetworkSettings/WifiConnectionDetails.tsx
koji: I added spacing10 to my inline notification branch since I noticed another component uses 10px.
I will update this after merging the branch to 7.0.0

koji:
use 10px for the above comment

app/src/organisms/ChildNavigation/index.tsx
koji: Need to find out a way to fix the header part efficiently

PR #12725, opened 2023-05-17
feat(app, robot-server): add protocol receipt toasts and protocol ids endpoint

shlokamin:
# Overview

This fUlLsTaCk PR adds an endpoint that returns all protocol ids on the robot server, and uses it to render protocol receipt toasts in the app. 

Closes RCORE-490 and RCORE-714

# Test Plan

-  Sent protocol to a flex
- Verified a protocol receipt toast shows up

# Changelog

- Add protocol ids endpoint to robot server
- Render protocol receipt toast on ODD when a new protocol is added to a Flex

# Review requests

See test plan

# Risk assessment

Low

PR #13292, opened 2023-08-11
refactor(robot-server): Followups to GET /protocols/ids

SyntaxColoring:
# Overview

Some low-priority follow-ups to the `GET /protocols/ids` endpoint, from the discussion in #12725.

# Test Plan

None needed. Changes are just to docs and tests.

# Changelog

The public-facing documentation of `GET /protocols/ids` was accidentally showing an internal Python thing. Replace it with something that says this endpoint is only meant for internal use. https://github.com/Opentrons/opentrons/pull/12725#discussion_r1200693447. I suspect we should replace this endpoint with a [partial](https://philsturgeon.com/a-happy-compromise-between-customization-and-cacheability/) on `GET /protocols` someday, so I'd rather not have external people start to depend on it. (Though it wouldn't be too hard to add a redirect, I guess.)

Also add an integration test. `GET /protocols/ids` has an ambiguity with `GET /protocols/{id}`. For `GET /protocols/ids` to take precedence, we need to define the endpoint functions in a particular order. It's only possible to cover this with a Tavern test. https://github.com/Opentrons/opentrons/pull/12725#discussion_r1200673851

# Review requests

None in particular.

# Risk assessment

No risk.

### DIFF
--- a/robot-server/robot_server/protocols/router.py
+++ b/robot-server/robot_server/protocols/router.py
@@ -307,7 +307,14 @@ async def get_protocols(
 
 @protocols_router.get(
     path="/protocols/ids",
-    summary="Get uploaded protocol ids",
+    summary="[Internal] Get uploaded protocol ids",
+    description=(
+        "Get the IDs of all protocols stored on the server."
+        "\n\n"
+        "**Warning:**"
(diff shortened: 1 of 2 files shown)


Laura-Danielle (approved):
lgtm!

PR #13295, opened 2023-08-12
test(api,robot-server): Test backwards-compatible validation of metadata and requirements

SyntaxColoring:
# Overview

This PR adds some tests ahead of #13258.

RSS-264 and RSS-265 add validation to prevent Python protocols from having weird `metadata` and `requirements` dicts. We do want that validation, but a lot of our internal testing robots already have protocols that don't follow those rules. We want to implement a sort of compatibility mode to avoid disturbing them. This PR adds tests for that.

This work should be undone at some point in the future, probably after the Flex release. RSS-306.

# Test Plan

None needed. Changes are just to tests.

# Changelog

* If a robot already has a stored protocol with weird `metadata` or `requirements`, we want it to still be rerunnable. Cover this with a persistence snapshot test.
* We want robots to reject uploads of additional protocols that have weird `metadata` or `requirements`. Cover this with a Tavern integration test.
    * Currently `xfail`'d. The robot will *not* reject these uploads right now. That's coming in the next PR.
* We want the local analysis engine used by the Opentrons App to raise an error if you give it a protocol that has weird `metadata` or `requirements`. Cover this with an integration test.
    * Currently `xfail`'d. The local analysis engine will *not* reject these protocols right now. That's coming in the next PR.
* Rename the `v6.2.0Large` snapshot to `v6.2.0_large`, just for consistency.

# Review requests

None in particular.

# Risk assessment

No risk.


TamarZanzouri (approved):
this is amazing!

PR #13296, opened 2023-08-14
feat(hardware): add can message to get gripper jaw state from firmware

ahiuchingau:
# Overview
The new can message allows us to get the firmware to report the real-time actual brushed motor state. 

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
lgtm but I was curious if there is an accompanying firmware PR that I missed?

sfoster1 (approved):
cool cool!

PR #13297, opened 2023-08-14
fix(protocol-designer): get labware dropdown from current deck state

jerader:
closes RQA-1276

# Overview

When you move a labware that is originally on top of a module to a new location, the labware dropdown info still shows the module in the prefix. And vice versa, if you move a labware to be on top of a module, the module prefix never shows up. This PR fixes that and gets the labware location from the current state instead of the starting deck state.

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_labware_dropdown_mod-prefix/

Follow the steps outlined in the ticket. But basically, add a labware on top of a module in the initial deck state and move it to a deck slot location. Then move the labware again and notice that the labware's module location is no longer there.

Also, make an ot-2 protocol and add a magnetic module. add a magnet form step and make sure the whole labware with module is shown - since the prefix change also affected this area as well

# Changelog

- modify module prefixes and rename to module since it isn't a prefix anymore
- modify `getModuleUnderLabware` util to update the labware's current location instead of only getting the starting location
- update tests

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/ui/labware/selectors.ts
+++ b/protocol-designer/src/ui/labware/selectors.ts
@@ -68,13 +68,13 @@ export const getLabwareOptions: Selector<Options> = createSelector(
           savedStepForms,
           labwareId
         )
-        const prefix = moduleOnDeck
+        const module = moduleOnDeck
           ? i18n.t(
               `form.step_edit_form.field.moduleLabwarePrefix.${moduleOnDeck.type}`
             )
           : null
-        const nickName = prefix
-          ? `${nicknamesById[labwareId]} in ${prefix}`
+        const nickName = module
+          ? `${nicknamesById[labwareId]} in ${module}`
           : nicknamesById[labwareId]
 
         if (!moveLabwarePresavedStep) {
--- a/protocol-designer/src/ui/labware/selectors.ts
+++ b/protocol-designer/src/ui/labware/selectors.ts
@@ -65,7 +65,7 @@ export const getLabwareOptions: Selector<Options> = createSelector(
           labwareEntity.def.metadata.displayCategory === 'aluminumBlock'
         const moduleOnDeck = getModuleUnderLabware(
           initialDeckSetup,
-          savedStepForms,
+          savedStepForms ?? {},
           labwareId
         )
         const module =
--- a/protocol-designer/src/ui/labware/selectors.ts
+++ b/protocol-designer/src/ui/labware/selectors.ts
@@ -68,14 +68,16 @@ export const getLabwareOptions: Selector<Options> = createSelector(
           savedStepForms,
           labwareId
         )
-        const module = moduleOnDeck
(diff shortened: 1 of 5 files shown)


protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -42,22 +42,27 @@ export const getLabwareOptions: Selector<Options> = createSelector(
...
         labwareId: string
       ): Options => {
-        const moduleOnDeck = getModuleUnderLabware(initialDeckSetup, labwareId)
-        const prefix = moduleOnDeck
+        const moduleOnDeck = getModuleUnderLabware(
+          initialDeckSetup,
+          savedStepForms,
+          labwareId
+        )
+        const module = moduleOnDeck
```
koji:
```suggestion
        const module = moduleOnDeck != null
```

protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -42,22 +42,27 @@ export const getLabwareOptions: Selector<Options> = createSelector(
...
           ? i18n.t(
               `form.step_edit_form.field.moduleLabwarePrefix.${moduleOnDeck.type}`
             )
           : null
-        const nickName = prefix
-          ? `${prefix} ${nicknamesById[labwareId]}`
+        const nickName = module
```
koji:
```suggestion
        const nickName = module != null
```

protocol-designer/src/ui/modules/utils.ts
```diff
@@ -6,11 +6,13 @@ import {
   ModuleType,
 } from '@opentrons/shared-data'
 import { Options } from '@opentrons/components'
-import {
+import type {
   ModuleOnDeck,
   LabwareOnDeck,
   InitialDeckSetup,
 } from '../../step-forms/types'
+import { SavedStepFormState } from '../../step-forms'
```
koji: Can you move this line before import type?

protocol-designer/src/ui/modules/utils.ts
```diff
@@ -29,11 +31,25 @@ export function getLabwareOnModule(
...
   initialDeckSetup: InitialDeckSetup,
+  savedStepFormState: SavedStepFormState,
   labwareId: string
 ): ModuleOnDeck | null | undefined {
+  //  latest moveLabware step related to labwareId
+  const moveLabwareStep = Object.values(savedStepFormState)
+    .filter(
+      state =>
+        state.stepType === 'moveLabware' &&
+        labwareId != null &&
```
koji, line 42: Maybe this line isn't required because of L35?
Or maybe L35 will need to be the following?

```
labwareId?: string | null
```

koji (approved):
The sandbox worked as expected 
I left a couple of comments

PR #13299, opened 2023-08-14
fix(app): navigate to run preview tab after starting run

brenthagen:
# Overview

after clicking start run, users should be routed to the run preview tab on the protocol run page. this is necessary in order to show intervention modals. fixes a regression from prior desktop app behavior

# Test Plan

 - manually confirmed navigation to run preview tab after start, and intervention modals rendering

# Changelog

 - Navigates to run preview tab after starting run

# Review requests

# Risk assessment

low


b-cooper (approved):
📑

PR #13300, opened 2023-08-14
refactor(api, engine): remove loadAdapter command from Protocol Engine

jbleon95:
# Overview

This PR removes loadAdapter from the Protocol Engine (but not any PAPI facing changes).

Why is this happening? The addition of a separate `loadAdapter` method and having it's own `adapterId` parameter, but otherwise treating it as a labware has caused some confusion on both the backend and in frontend implementation of fully supporting adapters. It was decided that in order to move more in the direction of treating adapters as labware, which for all intensive purposes to the PE they are, we would remove `loadAdapter` and load adapters through the existing `loadLabware` command.

Since we are still supporting `load_adapter` from the protocol API, some logic had to be moved around. The two main pieces of logic that had to be refactored is checking to make sure that an adapter is not being loaded on a labware (since it now uses the `LabwareLocation` type for the command, so this now has to be checked) and validating if a loaded definition is an adapter or a labware, since this now isn't being done in the PE side which sees no difference between the two for purposes of "can it be loaded at all".

# Test Plan

Testing existing protocols with `load_adapter` from PAPI to ensure that they still pass and behave identically, and ensuring that `loadLabware` can correctly load adapters just through PE via Postman.

# Changelog

- Removed `loadAdapter` from Protocol Engine
- Refactored `loadLabware` to accept loading adapters
- Moved validating whether a labware definition is a labware or an adapter to PAPI implementation

# Review requests

Just the general: is this being done as cleanly as possible and is there any possible edge cases I'm missing?

# Risk assessment

Medium. This is removing a command from Protocol Engine which always has an inherent danger of it in requiring a database reset for any robot that has a run or analysis that has this in it. There are two things that mitigate this slightly, one being that this only affects any robot which has used `load_adapter` which has not been proliferated to far to my knowledge, and two, that this only affects internal releases and will provide no danger to anybody upgrading an OT-2 in the future from 6.3.1. Waiting any longer to remove this will take away those two mitigating factors.


api/src/opentrons/protocol_api/validation.py
```diff
@@ -162,6 +172,22 @@ def ensure_lowercase_name(name: str) -> str:
     return name.lower()
 
 
+def ensure_definition_is_adapter(definition: LabwareDefinition) -> None:
+    """Ensure that one of the definition's allowed roles is `adapter`."""
+    if LabwareRole.adapter not in definition.allowedRoles:
+        raise LabwareDefinitionIsNotAdapterError(
+            f"Labware {definition.parameters.loadName} does not defined as an adapter."
```
SyntaxColoring:
```suggestion
            f"Labware {definition.parameters.loadName} does is not an adapter."
```

Labware adapter doesn't afraid of anything.

api/src/opentrons/protocol_api/validation.py
```diff
@@ -162,6 +172,22 @@ def ensure_lowercase_name(name: str) -> str:
...
+        )
+
+
+def ensure_definition_is_labware(definition: LabwareDefinition) -> None:
+    """Ensure that one of the definition's allowed roles is `labware` or that that field is empty."""
+    if definition.allowedRoles and LabwareRole.labware not in definition.allowedRoles:
+        raise LabwareDefinitionIsNotLabwareError(
+            f"Labware {definition.parameters.loadName} is not defined as a labware."
+        )
```
SyntaxColoring: Just an error message suggestion. Otherwise it sounds like this labware definition is missing entirely.

```suggestion
        raise LabwareDefinitionIsNotLabwareError(
            f"Labware {definition.parameters.loadName} is an adapter,"
            f" not a normal labware."
        )
```

SyntaxColoring (approved):
Nice! Glad we're getting this simplification in before it hits the outside world.

api/tests/opentrons/protocol_api/test_validation.py
```diff
@@ -141,6 +148,74 @@ def test_ensure_lowercase_name_invalid() -> None:
...
+                parameters=LabwareDefinitionParameters.construct(loadName="Foo"),  # type: ignore[call-arg]
+            ),
+            pytest.raises(subject.LabwareDefinitionIsNotLabwareError),
+        ),
+    ],
+)
+def test_ensure_definition_is_labware(
+    definition: LabwareDefinition, expected_raise: ContextManager[Any]
+) -> None:
+    """It should check if the Labware Definition is defined as an adapter."""
```
sanni-t:
```suggestion
    """It should check if the Labware Definition is defined as a normal labware."""
```

api/tests/opentrons/protocol_engine/state/test_labware_view.py
```diff
@@ -1196,6 +1196,24 @@ def test_get_all_labware_definition_empty() -> None:
     assert result == []
 
 
+def test_raise_if_labware_cannot_be_stacked_stacking_adapter() -> None:
```
sanni-t:
```suggestion
def test_raise_if_labware_being_stacked_is_adapter() -> None:
```
maybe?

sanni-t (approved):
Thanks for the quick change!

b-cooper (approved):
🏁

PR #13304, opened 2023-08-14
fix(api): allow loading a flex trash into column 3 only

TamarZanzouri:
# Overview

part of https://opentrons.atlassian.net/browse/RSS-255. 
restricting flex trash into column 3. will need a follow up pr for restricting column 1 and adding slot transforms. 

# Test Plan

- Load a `opentrons_1_trash_3200ml_fixed` as a labware to any slot other than column 3.
- Make sure it's raising an error that location is not allowed. 

# Changelog

- Added a validation method to check if that load name is a flex trash
- Make sure the load command is not in the forbidden slot.

this should fail
```
from typing import cast

requirements = {
    "apiLevel": "2.15",
    "robotType": "OT-3",
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    pipette = ctx.load_instrument(instrument_name="p50_multi_flex", mount=types.Mount.LEFT)
    pipette.trash_container = ctx.load_labware("opentrons_1_trash_3200ml_fixed", "A2")
```

this should pass
```
from typing import cast

requirements = {
    "apiLevel": "2.15",
    "robotType": "OT-3",
}

def run(ctx: protocol_api.ProtocolContext) -> None:
    pipette = ctx.load_instrument(instrument_name="p50_multi_flex", mount=types.Mount.LEFT)
    pipette.trash_container = ctx.load_labware("opentrons_1_trash_3200ml_fixed", "B3")
```
# Review requests

1. how much do you hate `is_flex_trash`? could just do a string check but this seems nicer.
2. anything else I'm missing?

# Risk assessment

low. adding validation to a specific load name.


api/src/opentrons/protocol_engine/commands/load_labware.py
```diff
@@ -92,6 +95,15 @@ def __init__(
 
     async def execute(self, params: LoadLabwareParams) -> LoadLabwareResult:
         """Load definition and calibration data necessary for a labware."""
+        if (
+            labware_validation.is_flex_trash(params.loadName)
+            and isinstance(params.location, DeckSlotLocation)
+            and self._state_view.geometry.get_slot_column(params.location.slotName) != 3
+        ):
```
sanni-t, line 103: I'd just add a TODO here to remove this check once RSS-258 is addressed and RSS-255 is correctly implemented in PAPI.

sanni-t (approved):
LGTM!

PR #13305, opened 2023-08-15
feat(shared-data): Modify shared data to support low volume on the p50

Laura-Danielle:
# Overview

Closes RSS-310 & RSS-311. This is the first step in [supporting low volumes](https://opentrons.atlassian.net/browse/RSS-309) on the p50s. Slight modification from the structure described in the [proposal doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3778609180/Flex+Low-Volume+Configuration). See the [comment](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3778609180/Flex+Low-Volume+Configuration#Shared-Data-and-Configuration) regarding this change.

# Test Plan

Throw on a robot and make sure things don't break, but should be fairly well tested.

# Changelog

You should be able to follow the commits for the changes.

- Add a new plunger position called `low_volume_bottom`
- Add a new parameter (and remove the top level keys) of min/max volume that changes based on the plunger position.

# Review requests

Maybe throw on a robot and make sure things are still functioning as expected when simulating a protocol.

# Risk assessment

Medium/low. Only changing how we access min/max volume currently.

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
@@ -15,6 +15,7 @@ from opentrons_shared_data.pipette.pipette_definition import (
     TipHandlingConfigurations,
     PipetteNameType,
     PipetteModelVersionType,
+    Volumes,
 )
 from ..instrument_abc import AbstractInstrument
 from ..instrument_helpers import (
@@ -73,6 +74,9 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         self._max_channels = self._config.channels
         self._backlash_distance = config.backlash_distance
 
+        self._volume_configurations = config.volume_breakpoints
+        self._volumes = self._volume_configurations.default
+
         # TODO (lc 12-05-2022) figure out how we can safely deprecate "name" and "model"
         self._pipette_name = PipetteNameType(
             pipette_type=config.pipette_type,
@@ -87,7 +91,7 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         )
         self._nozzle_offset = self._config.nozzle_offset
         self._current_volume = 0.0
-        self._working_volume = float(self._config.max_volume)
+        self._working_volume = float(self._volumes.max_volume)
         self._current_tip_length = 0.0
         self._current_tiprack_diameter = 0.0
         self._has_tip = False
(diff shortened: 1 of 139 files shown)


shared-data/pipette/definitions/2/liquid/eight_channel/p10/1_0.json
```diff
@@ -48,8 +48,16 @@
...
     "opentrons/eppendorf_96_tiprack_10ul_eptips/1": 1.0
   },
-  "maxVolume": 10,
-  "minVolume": 1.0,
+  "volumeBreakpoints": {
+    "default": {
+      "maxVolume": 10,
+      "minVolume": 1.0
+    },
+    "bottomLowVolume": {
```
sfoster1, line 56: what does it mean that this exactly overlaps with `default`?

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -212,14 +220,14 @@ def reload_configurations(self) -> None:
...
-        self._working_volume = float(self._config.max_volume)
+        self._working_volume = float(self._volumes.max_volume)
         self._current_tip_length = 0.0
         self._current_tiprack_diameter = 0.0
         self._has_tip = False
         self.ready_to_aspirate = False
         #: True if ready to aspirate
         self._active_tip_settings = self._config.supported_tips[
-            pip_types.PipetteTipType(self._config.max_volume)
+            pip_types.PipetteTipType(self._volumes.max_volume)
```
sfoster1, line 230: when we switch modes, won't this stop working since it will return "45"?

shared-data/pipette/definitions/2/general/eight_channel/p10/1_0.json
```diff
@@ -21,6 +21,7 @@
   "plungerPositionsConfigurations": {
     "top": 19.5,
     "bottom": 2.0,
+    "bottomLowVolume": 2.0,
```
sfoster1, line 24: Can we make this a mapping by breakpoint names or something? I really don't like having this different key in there. like
```
plungerPositionsConfigurationsByBreakpoint: {
   default: {
     top: ...
   },
   lowVolume: {
      top: ...
    }
``` or somethign

sfoster1 (changes requested):
I get the desire to keep things simple, but I think we need a little more complexity than this. Maybe we don't need _everything_ but we should 
- use maps by breakpoint name instead of magic keys in plunger positions
- not have `lowVolume` breakpoints if they're meaningless, and adapt code to handle optionally-present breakpoints
- Maybe a have separate trigger volumes for breakpoint selection with hysteresis, like `switchInIfBelow` and `switchOutIfAbove` volumes to prevent aspirating 5 and then 6 and then 5 and then 6 swapping back and forth every time

PR #13306, opened 2023-08-15
build: robot builds for all release types

sfoster1:
Fix a couple build issues in the release branch:
- Some failures in the app yaml that were evident in the cold light of day, like incorrect glob patterns and bad step `if`s (this commit is in the release branch already unfortunately, you can see it here: https://github.com/Opentrons/opentrons/commit/59a588959c924567538bceb58bcfba00d0e3dec8 )
- robot builds actually need to be triggered on regular release branches and tags

### DIFF
--- a/scripts/deploy/create-release.js
+++ b/scripts/deploy/create-release.js
@@ -177,6 +177,15 @@ async function createRelease(token, tag, project, version, changelog, deploy) {
   }
 }
 
+function truncateAndAnnotate(changelog, limit, prevtag, thistag) {
+  if (changelog.length < limit) {
+    return changelog
+  }
+  const truncated = changelog.substring(0, limit)
+  const linkmessage = `...and more! Log link: https://github.com/${REPO_DETAILS.owner}/${REPO_DETAILS.repo}/compare/${prevtag}...${thistag}`
+  return truncated + linkmessage
+}
+
 async function main() {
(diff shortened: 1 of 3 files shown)


vegano1 (approved):
makes sense.

b-cooper (approved):
⛓️ 🌲

scripts/deploy/create-release.js
```diff
@@ -177,6 +177,15 @@ async function createRelease(token, tag, project, version, changelog, deploy) {
   }
 }
 
+function truncateAndAnnotate(changelog, limit, prevtag, thistag) {
+  if (changelog.length < limit) {
+    return changelog
+  }
+  const truncated = changelog.substring(0, limit)
+  const linkmessage = `...and more! Log link: https://github.com/${REPO_DETAILS.owner}/${REPO_DETAILS.repo}/compare/${prevtag}...${thistag}`
```
vegano1: Wouldn't this still fail here? since we are adding this message to the max message size dictated by `limit`?

sfoster1: yeah but I'm passing a value smaller than the actual limit to `limit`. i should probably document that

sfoster1: i made it take the length of the addon message into account specifically

vegano1: yep, that works.

PR #13313, opened 2023-08-15
fix(app): show robot moving loader when exiting gripper calibration

smb2268:
fix RQA-753

# Overview
When exiting a gripper flow from a pin attach/removal screen the wizard should display the "robot in motion" loader not the calibration in progress loader

# Test Plan
Early exit from gripper calibration on the attach, move, or remove probe screen and confirm you do not see the calibration in progress modal while exiting

# Changelog
If `isExiting` is true show the normal spinner

# Review requests
Review test plan

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/GripperWizardFlows/__tests__/MovePin.test.tsx
+++ b/app/src/organisms/GripperWizardFlows/__tests__/MovePin.test.tsx
@@ -170,4 +170,13 @@ describe('MovePin', () => {
     })[0]
     getByText('Stand Back, Robot is in Motion')
   })
+
+  it('renders correct loader for early exiting', () => {
+    const { getByText } = render({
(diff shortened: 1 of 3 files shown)


app/src/organisms/GripperWizardFlows/__tests__/MovePin.test.tsx
```diff
@@ -47,6 +47,7 @@ describe('MovePin', () => {
           createRunCommand={mockCreateRunCommand}
           errorMessage={null}
           setErrorMessage={mockSetErrorMessage}
+          isExiting={false}
```
jerader, line 50: you can add a test case for if `isExisting` is true

jerader (approved):
tested this on my dev environment, lgtm

PR #13314, opened 2023-08-15
fix(app): Replace calibration probe image

smb2268:
fix RQA-1232

# Overview
Replace pipette calibration probe image with the PVT probe
Screen Shot 2023-08-15 at 4 54 58 PM

# Test Plan
Initiate a pipette flow and see that the new image is properly reflected

# Changelog
Swap image

# Review requests
Look at test plan

# Risk assessment
Low


jerader (approved):
it sounds like from the ticket that Mel gave you the png, but i wonder if it needs to be higher res? We have been downloading images from figma in 2x. But if Mel approves then this lgtm!

smb2268:
> it sounds like from the ticket that Mel gave you the png, but i wonder if it needs to be higher res? We have been downloading images from figma in 2x. But if Mel approves then this lgtm!

@jerader I think the image Mel sent me is the 2x png!

PR #13316, opened 2023-08-15
fix(app): memoize pipette name on detach pipette screen

smb2268:
fix RQA-1236

# Overview
On detach pipette flows we show the pipette name of the pipette to detach. We need to memoize this so that we don't lose the name after the pipette is removed.

# Test Plan
Go through a replace or detach pipette flow. After removing the pipette, wait 5 seconds before proceeding to the next screen and ensure the pipette name does not change to "undefined" 

# Changelog
Memoize attached pipette info on DetachPipette screen

# Review requests
Look over test plan

# Risk assessment
Low


jerader (approved):
oh cool, i keep forgetting about `useMemo()`, makes sense to me! can't test this without a robot though

koji (approved):
The changes look good to me.
Tested this PR on a Flex and it worked as expected.

PR #13318, opened 2023-08-16
refactor(api): hardware controller use error codes

ahiuchingau:
# Overview
We are now using the error codes in the hardware controller api. Let me know if i forgot anything!

### DIFF
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -763,7 +763,7 @@ class API(
             if fail_on_not_homed:
                 raise PositionUnknownError(
                     message="Cannot make a relative move because absolute position"
-                            " is unknown.",
+                    " is unknown.",
                     detail={
                         "mount": str(mount),
                         "fail_on_not_homed": fail_on_not_homed,
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -641,14 +641,20 @@ class API(
         plunger_ax = Axis.of_plunger(mount)
         position_axes = [Axis.X, Axis.Y, z_ax, plunger_ax]
 
-        if fail_on_not_homed and (
-            not self._backend.is_homed([ot2_axis_to_string(a) for a in position_axes])
-            or not self._current_position
-        ):
-            raise PositionUnknownError(
-                message=f"Current position of {str(mount)} pipette is unknown, please home."
-            )
-
+        if fail_on_not_homed:
+            if not self._current_position:
+                raise PositionUnknownError(
+                    message=f"Current position of {str(mount)} pipette is unknown,"
+                    " please home.",
+                    detail={"mount": str(mount), "missing_axes": position_axes},
+                )
+            axes_str = [ot2_axis_to_string(a) for a in position_axes]
+            if not self._backend.is_homed(axes_str):
+                unhomed = self._backend._unhomed_axes(axes_str)
+                raise PositionUnknownError(
+                    message=f"{str(mount)} pipette axes ({unhomed}) must be homed.",
+                    detail={"mount": str(mount), "unhomed_axes": unhomed},
+                )
         elif not self._current_position and not refresh:
             raise PositionUnknownError(
                 message="Current position is unknown; please home motors."
@@ -733,7 +739,8 @@ class API(
         The effector of the pipette mount axis are the mount critical points but only in z.
         """
         raise UnsupportedHardwareCommand(
-            message="move_axes is not supported on the OT-2."
+            message="move_axes is not supported on the OT-2.",
+            detail={"axes_commanded": list(position.keys())},
         )
 
     async def move_rel(
@@ -752,23 +759,32 @@ class API(
         # TODO: Remove the fail_on_not_homed and make this the behavior all the time.
         # Having the optional arg makes the bug stick around in existing code and we
         # really want to fix it when we're not gearing up for a release.
-        mhe = PositionUnknownError(
-            message="Cannot make a relative move because absolute position is unknown"
-        )
         if not self._current_position:
             if fail_on_not_homed:
-                raise mhe
+                raise PositionUnknownError(
+                    message="Cannot make a relative move because absolute position"
+                            " is unknown.",
+                    detail={
+                        "mount": str(mount),
+                        "fail_on_not_homed": fail_on_not_homed,
+                    },
+                )
             else:
                 await self.home()
 
         target_position = target_position_from_relative(
             mount, delta, self._current_position
         )
+
         axes_moving = [Axis.X, Axis.Y, Axis.by_mount(mount)]
-        if fail_on_not_homed and not self._backend.is_homed(
-            [ot2_axis_to_string(axis) for axis in axes_moving if axis is not None]
-        ):
-            raise mhe
+        axes_str = [ot2_axis_to_string(a) for a in axes_moving]
+        if fail_on_not_homed and not self._backend.is_homed(axes_str):
+            unhomed = self._backend._unhomed_axes(axes_str)
+            raise PositionUnknownError(
+                message=f"{str(mount)} pipette axes ({unhomed}) must be homed.",
+                detail={"mount": str(mount), "unhomed_axes": unhomed},
+            )
+
         await self._cache_and_maybe_retract_mount(mount)
         await self._move(
             target_position,
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -601,10 +601,13 @@ class API(
         # No internal code passes OT3 axes as arguments on an OT2. But a user/ client
         # can still explicitly specify an OT3 axis even when working on an OT2.
         # Adding this check in order to prevent misuse of axes types.
-        if axes and any(axis not in Axis.ot2_axes() for axis in axes):
-            raise UnsupportedHardwareCommand(
-                message=f"At least one axis in {axes} is not supported on the OT2."
-            )
+        if axes:
+            unsupported = list(axis not in Axis.ot2_axes() for axis in axes)
+            if any(unsupported):
+                raise UnsupportedHardwareCommand(
+                    message=f"At least one axis in {axes} is not supported on the OT2.",
+                    detail={"unsupported_axes": unsupported},
+                )
         self._reset_last_mount()
         # Initialize/update current_position
         checked_axes = axes or [ax for ax in Axis.ot2_axes()]
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -20,6 +20,10 @@ from typing import (
     cast,
 )
 
+from opentrons_shared_data.errors.exceptions import (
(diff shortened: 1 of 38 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -182,12 +182,12 @@
...
 
 def requires_update(func: Wrapped) -> Wrapped:
-    """Decorator that raises FirmwareUpdateRequired if the update_required flag is set."""
+    """Decorator that raises FirmwareUpdateRequiredError if the update_required flag is set."""
 
     @wraps(func)
     async def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
         if self.update_required and self.initialized:
-            raise FirmwareUpdateRequired()
+            raise FirmwareUpdateRequiredError
```
sfoster1: the idea with these errors is that you can pack details in them, and we should generally take advantage of that. so here for instance, you could put in the `details` argument a dict containing like
- `func.__name__`
- `update_required` and `initialized`
- maybe even poke into `self.subsystems` and figuring out which subsystems need the update

api/src/opentrons/hardware_control/errors.py
```diff
@@ -1,26 +1,25 @@
...
 from .types import OT3Mount
+from opentrons_shared_data.errors.exceptions import MotionPlanningFailureError, InvalidInstrumentData
 
 
-class OutOfBoundsMove(RuntimeError):
+class OutOfBoundsMove(MotionPlanningFailureError):
     def __init__(self, message: str):
-        self.message = message
-        super().__init__()
+        super().__init__(message=message)
```
sfoster1: we should also pass the details in and store them in `details`

api/src/opentrons/hardware_control/instruments/ot3/gripper_handler.py
```diff
@@ -93,9 +91,13 @@ def save_instrument_offset(self, delta: Point) -> GripperCalibrationOffset:
 
     def get_critical_point(self, cp_override: Optional[CriticalPoint] = None) -> Point:
         if not self._gripper:
-            raise GripperNotAttachedError()
+            raise GripperNotPresentError
```
sfoster1:
```suggestion
            raise GripperNotPresentError()
```

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -478,7 +480,7 @@ async def update_firmware(
             raise UpdateOngoingError(e.msg) from e
         except Exception as e:
             mod_log.exception("Firmware update failed")
-            raise FirmwareUpdateFailed() from e
+            raise FirmwareUpdateFailedError() from e
```
sfoster1: We probably want to let enumerated errors pass through here. What this code is trying to do is let higher layers (like robot-server) catch exceptions explicitly without having to import lower layers (like hardware) to get the exception classes. So if the lower layers are using exceptions from `opentrons_shared_data`, we don't have that concern, and can let the exception pass.

That said, if that also requires modifying robot server code where it is catching the exact exception class (e.g. `UpdateOngoingError`) then we can leave it if we want.

Either way, the way to let this pass through would be something like
```
except EnumeratedError:
     raise
except BaseException as e:
    raise FirmwareUpdateFailedError(message='Update failed because of uncaught error', wrapping=[PythonException(e)]) from e

File: api/src/opentrons/hardware_control/ot3api.py [hunk 5]
```diff
@@ -839,15 +841,15 @@ async def current_position_ot3(
...
+                message=f"Cannot return position for {mount} if no gripper is attached"
             )
 
         if refresh:
             await self.refresh_positions()
         elif not self._current_position:
-            raise MustHomeError(
-                f"Motor positions for {str(mount)} are missing; must first home motors."
+            raise PositionUnknownError(
+                message=f"Motor positions for {str(mount)} are missing; must first home motors."
```
sfoster1: we should put which mount this was in the detail

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -882,16 +884,16 @@ def _assert_motor_ok(self, axes: Sequence[Axis]) -> None:
         invalid_axes = self._backend.get_invalid_motor_axes(axes)
         if invalid_axes:
             axes_str = ",".join([ax.name for ax in invalid_axes])
-            raise MustHomeError(
-                f"Motor position of axes ({axes_str}) is invalid; please home motors."
+            raise PositionUnknownError(
+                message=f"Motor position of axes ({axes_str}) is invalid; please home motors."
```
sfoster1: mount in the detail

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -882,16 +884,16 @@ def _assert_motor_ok(self, axes: Sequence[Axis]) -> None:
...
             )
 
     def _assert_encoder_ok(self, axes: Sequence[Axis]) -> None:
         invalid_axes = self._backend.get_invalid_motor_axes(axes)
         if invalid_axes:
             axes_str = ",".join([ax.name for ax in invalid_axes])
-            raise MustHomeError(
-                f"Encoder position of axes ({axes_str}) is invalid; please home motors."
+            raise PositionUnknownError(
```
sfoster1, line 952: mount in the detail

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -917,13 +919,13 @@ async def encoder_current_position_ot3(
         if refresh:
             await self.refresh_positions()
         elif not self._encoder_position:
-            raise MustHomeError(
-                f"Encoder positions for {str(mount)} are missing; must first home motors."
+            raise PositionUnknownError(
```
sfoster1, line 980: mount in the detail

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -917,13 +919,13 @@ async def encoder_current_position_ot3(
...
+                message=f"Encoder positions for {str(mount)} are missing; must first home motors."
             )
 
         if mount == OT3Mount.GRIPPER and not self._gripper_handler.has_gripper():
-            raise GripperNotAttachedError(
-                f"Cannot return encoder position for {mount} if no gripper is attached"
+            raise GripperNotPresentError(
```
sfoster1, line 986: mount in the detail

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1045,7 +1047,7 @@ async def move_axes(  # noqa: C901
 
         for axis in position.keys():
             if not self._backend.axis_is_present(axis):
-                raise AxisNotPresentError(f"{axis} is not present")
+                raise InvalidActuator(message=f"{axis} is not present")
```
sfoster1: axis in the detail

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -398,7 +398,7 @@ def __init__(
         wrapping: Optional[Sequence[EnumeratedError]] = None,
     ) -> None:
         """Build a MustHomeError."""
-        super().__init__(ErrorCodes.GENERAL_ERROR, message, details, wrapping)
+        super().__init__(ErrorCodes.POSITION_UNKNOWN, message, details, wrapping)
```
sfoster1, line 401: whoops! thank you

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -712,7 +712,9 @@ def __init__(
         wrapping: Optional[Sequence[EnumeratedError]] = None,
     ) -> None:
         """Build a LocationIsOccupiedError."""
-        super().__init__(ErrorCodes.GENERAL_ERROR, message, details, wrapping)
+        super().__init__(
+            ErrorCodes.FIRMWARE_UPDATE_REQUIRED, message, details, wrapping
```
sfoster1, line 742: too much copy pasting on my part

sfoster1 (commented):
Inline feedback!

api/src/opentrons/protocol_engine/execution/gantry_mover.py
```diff
@@ -107,7 +107,7 @@ async def get_position(
                 critical_point=pipette_location.critical_point,
                 fail_on_not_homed=fail_on_not_homed,
             )
-        except HardwareMustHomeError as e:
+        except PositionUnknownError as e:
             raise MustHomeError(str(e)) from e
```
sfoster1: this should link the lower level error with `wraps=[e]`

api/src/opentrons/protocol_engine/execution/gantry_mover.py
```diff
@@ -167,7 +167,7 @@ async def move_relative(
                 critical_point=critical_point,
                 fail_on_not_homed=True,
             )
-        except HardwareMustHomeError as e:
+        except PositionUnknownError as e:
             raise MustHomeError(str(e)) from e
```
sfoster1: should link the error with `wraps=[e]`

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -877,15 +886,17 @@ async def current_position_ot3(
...
             )
 
         if refresh:
             await self.refresh_positions()
         elif not self._current_position:
-            raise MustHomeError(
-                f"Motor positions for {str(mount)} are missing; must first home motors."
+            raise PositionUnknownError(
+                message=f"Motor positions for {str(mount)} are missing; must first home motors.",
+                detail={"mount": str(mount)},
```
sfoster1: can we put the exact missing axis positions in the mount?

api/src/opentrons/hardware_control/api.py
```diff
@@ -645,12 +645,14 @@ async def current_position(
             not self._backend.is_homed([ot2_axis_to_string(a) for a in position_axes])
             or not self._current_position
         ):
-            raise MustHomeError(
-                f"Current position of {str(mount)} pipette is unknown, please home."
+            raise PositionUnknownError(
+                message=f"Current position of {str(mount)} pipette is unknown, please home."
```
sfoster1: maybe put in the details whether the problem is that the axis isn't homed or that position hasn't been pulled yet

api/src/opentrons/hardware_control/api.py
```diff
@@ -602,8 +602,8 @@ async def home(self, axes: Optional[List[Axis]] = None) -> None:
         # can still explicitly specify an OT3 axis even when working on an OT2.
         # Adding this check in order to prevent misuse of axes types.
         if axes and any(axis not in Axis.ot2_axes() for axis in axes):
-            raise NotSupportedByHardware(
-                f"At least one axis in {axes} is not supported on the OT2."
+            raise UnsupportedHardwareCommand(
+                message=f"At least one axis in {axes} is not supported on the OT2."
```
sfoster1: maybe worth putting in which axis it is, at least in the details

sfoster1 (changes requested):
Couple nitpicks but looking good!

sfoster1 (approved):
Looks great!

PR #13320, opened 2023-08-16
fix(api): Fix protocol.deck not accepting non-native deck slot spellings

SyntaxColoring:
# Overview

Fixes RSS-315.

This fixes `protocol.deck["A1"]` on an OT-2 and `protocol.deck[1]` on a Flex. It likewise fixes other methods of `protocol.deck`, like `.position_for()`.

# Test Plan

See the steps to reproduce in RSS-315.

# Changelog

Without this PR, when `protocol.deck` takes a slot as input, it merely validates it as matching *any* slot for *any* robot. This causes lookups to behave weirdly, because lower-level things are expecting the slot to match the robot's native format.

This PR fixes it by adding normalization on all input.

# Review requests

Double-check the `Deck` class. All methods should:

* Accept deck slots in any format—`5`, `"5"`, or `"C2"`. The `"C2"` format requires a recent `apiLevel`.
* Do internal lookups based on the `DeckSlotName` enum, with the robot's native format—like `DeckSlotName.SLOT_5` for OT-2, or `DeckSlotName.SLOT_C2` for Flex.
* Return to the user strings with the robot's native format—like `"5"` for OT-2 or `"C2"` for Flex.

In hindsight, it's possible that this normalization logic should be pushed down into Protocol Engine. That's how we've been doing it for `load_labware()` and `load_module()`. But the `Deck` class is full of cruft and I guess I didn't want to bloat Protocol Engine with supporting it. Any thoughts? 

I guess we know Protocol Engine *commands* should normalize input slots, but there's a question of whether the getters on `ProtocolEngine.state` should normalize input slots.

See also RSS-259. We clearly need more static type-safety here.

# Risk assessment

Medium? We don't really have good infrastructure for integration-testing this kind of thing [anymore](https://github.com/Opentrons/opentrons/blob/8c37860593493bd192d9bb19b53e306a0983be3c/api/tests/opentrons/protocol_api_old/__init__.py#L3-L4). We have plenty of unit tests, but I don't think those are sufficient for catching bugs like this.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/module_core.py
+++ b/api/src/opentrons/protocol_api/core/engine/module_core.py
@@ -81,7 +81,7 @@ class ModuleCore(AbstractModuleCore):
 
     def get_deck_slot_id(self) -> str:
         slot_name = self.get_deck_slot()
-        return validation.ensure_deck_slot_string(
+        return validation.internal_slot_to_public_string(
             slot_name, robot_type=self._engine_client.state.config.robot_type
         )
 
@@ -137,7 +137,7 @@ class NonConnectedModuleCore(AbstractModuleCore):
 
     def get_deck_slot_id(self) -> str:
         slot_name = self.get_deck_slot()
-        return validation.ensure_deck_slot_string(
+        return validation.internal_slot_to_public_string(
             slot_name, robot_type=self._engine_client.state.config.robot_type
         )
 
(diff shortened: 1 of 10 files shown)


sanni-t (approved):
LGTM! 
As we discussed on the call, you can add a python integration test with all virtual hardware & simulator for improved testing, but this is good as is too.

PR #13323, opened 2023-08-16
fix(api): Do not move/home RIGHT mount when 96ch is attached

andySigler:
# Overview

This PR checks if the RIGHT mount is being commanded to move/home while 96ch is attached.

Both the hold and run currents for the right mount are set to a very low value when a 96ch is attached, both because it is not safe to move it and because higher currents would unnecessarily increase heat build-up. However, the hardware-controller would still allow moving and homing, for example when calling `api.home()` with no axes specified.

# Test Plan

# Changelog

# Review requests

# Risk assessment

Low risk, this would primarily just make it so our test scripts run more seamlessly.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -998,8 +998,14 @@ class OT3API(
         realmount = OT3Mount.from_mount(mount)
         axes_moving = [Axis.X, Axis.Y, Axis.by_mount(mount)]
 
-        if self.gantry_load == GantryLoad.HIGH_THROUGHPUT and realmount == OT3Mount.RIGHT:
-            raise RuntimeError("unable to move RIGHT mount while 96CH is attached")
+        if (
+            self.gantry_load == GantryLoad.HIGH_THROUGHPUT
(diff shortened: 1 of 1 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1106,6 +1109,9 @@ async def move_rel(
         realmount = OT3Mount.from_mount(mount)
         axes_moving = [Axis.X, Axis.Y, Axis.by_mount(mount)]
 
+        if self.gantry_load == GantryLoad.HIGH_THROUGHPUT and realmount == OT3Mount.RIGHT:
+            raise RuntimeError("unable to move RIGHT mount while 96CH is attached")
```
sfoster1: can't have this be the error message probably, wants to be the actual pipette name

sfoster1 (changes requested):
I think we either should change the error message to mention the gantry load, or change the error condition to be "96 is attached". right now it's sort of a little of both and not great.

sfoster1 (approved):
Looks good to me, thanks!

PR #13324, opened 2023-08-16
fix(app):  Fix ODD pressed state CSS

mjhuff:
closes RAUT-520

# Overview

This PR harmonizes the pressed state CSS styling various ODD buttons. Dark gray bounding boxes no longer appear, and pressed state is added to several pages that were missing it. There are also a couple of CSS fixes, like the long press ProtocolDashboard menu also showing pressed state for buttons not on the modal, and touchscreen sleep buttons having bounding boxes that did not fit the shape of the button. Additionally, the inconsistent UI behavior of pressed state for existing routes is fixed.

Note that certain hover effects will no longer appear when emulating the ODD on a computer. Also, I removed some .focus-visible properties, as these could not be triggered previously on the physical ODD and were causing UI bugs (rendered boxes not fitting the shape of the buttons).

### What Causes Grey Bounding Boxes?

The behavior occurs when the bounding box does not perfectly fit the size of the text, and any psuedoclass container other than active changes the cursor property to anything but default. Ex: `&.hover: {cursor: pointer;}`

The major culprit is the Btn primitive, which is now wrapped in a media query to prevent misuse, however, there are other custom buttons on various pages. These have been changed as well. 

In order to implement bounding boxes on icons (or anything else), active properties with cursor: default seem to work.

# Test Plan

Must be tested on a physical ODD.
1. Long pressing the navbar links and overflow button on RobotDashboard should not cause grey bounding boxes to appear.
2. Navigating to Settings > Touchscreen Brightness and clicking the +/- should not cause grey bounding boxes to appear.
3. Settings > Touchscreen Sleep. Long pressing these buttons should not cause a visible gray bounding box that doesn't fit the dimensions of the buttons.
4. All Protocols. These buttons now have pressed state. Pinned protocols should also have pressed state.
5. All Protocols > Protocol (details page) > Start Setup (setup page). These buttons now have pressed state.
6.  Instruments Page. Pressed state is now consistent with other pressed state (previously was pressed for too long).
7. Start setup and then click instrument calibration. The wizard exit and back buttons should have state when pressed.
8. Network settings page now has pressed state for connected (green) networks and gray for unconnected networks.

# Changelog

- Fixed grey bounding boxes on existing pressed state.
- Harmonized behavior for existing pressed state.
- Created new pressed state behavior for buttons without it.

# Risk assessment

low

### DIFF
--- a/app/src/organisms/ChildNavigation/index.tsx
+++ b/app/src/organisms/ChildNavigation/index.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react'
+import styled from 'styled-components'
 
 import {
   ALIGN_CENTER,
-  Btn,
   COLORS,
   Flex,
   Icon,
@@ -10,7 +10,9 @@ import {
   JUSTIFY_SPACE_BETWEEN,
   SPACING,
   TYPOGRAPHY,
+  RESPONSIVENESS,
 } from '@opentrons/components'
+import { ODD_FOCUS_VISIBLE } from '../../atoms/buttons/constants'
 
 import { SmallButton } from '../../atoms/buttons'
 import { InlineNotification } from '../../atoms/InlineNotification'
@@ -42,9 +44,9 @@ export function ChildNavigation({
       paddingY={SPACING.spacing32}
     >
       <Flex gridGap={SPACING.spacing16} justifyContent={JUSTIFY_FLEX_START}>
-        <Btn onClick={onClickBack}>
+        <IconButton onClick={onClickBack}>
           <Icon name="back" size="3rem" color={COLORS.darkBlack100} />
-        </Btn>
+        </IconButton>
         <StyledText as="h2" fontWeight={TYPOGRAPHY.fontWeightBold}>
           {header}
         </StyledText>
@@ -66,3 +68,20 @@ export function ChildNavigation({
     </Flex>
   )
 }
+
+const IconButton = styled('button')`
+  border-radius: ${SPACING.spacing4};
+  max-height: 100%;
+  background-color: ${COLORS.white};
+
+  &:focus-visible {
+    box-shadow: ${ODD_FOCUS_VISIBLE};
+    background-color: ${COLORS.darkBlack20};
+  }
+  &:disabled {
+    background-color: transparent;
+  }
+  @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
+    cursor: default;
+  }
+`
--- a/app/src/organisms/Navigation/index.tsx
+++ b/app/src/organisms/Navigation/index.tsx
@@ -20,6 +20,7 @@ import {
   BORDERS,
   RESPONSIVENESS,
 } from '@opentrons/components'
+import { ODD_FOCUS_VISIBLE } from '../../atoms/buttons/constants'
 
 import { useNetworkConnection } from '../../pages/OnDeviceDisplay/hooks'
 import { getLocalRobot } from '../../redux/discovery'
@@ -168,10 +169,16 @@ const IconButton = styled('button')`
   max-height: 100%;
   background-color: ${COLORS.white};
 
+  &:active {
+    background-color: ${COLORS.darkBlack20};
+  }
+  &:focus-visible {
+    box-shadow: ${ODD_FOCUS_VISIBLE};
+    background-color: ${COLORS.darkBlack20};
+  }
   &:disabled {
     background-color: transparent;
   }
-
   @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
     cursor: default;
   }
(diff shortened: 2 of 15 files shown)


app/src/organisms/Navigation/index.tsx
```diff
@@ -157,25 +157,30 @@ const TouchNavLink = styled(NavLink)`
   &.active > div {
     background-color: ${COLORS.highlightPurple1};
   }
+
+  @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
+    cursor: default;
+
+    &:hover {
```
mjhuff, line 164: The hover effect doesn't impact the ODD here - I decided to leave it in.

app/src/organisms/Navigation/index.tsx
```diff
@@ -157,25 +157,30 @@ const TouchNavLink = styled(NavLink)`
...
+      cursor: pointer;
+    }
+  }
```
mjhuff, line 167: The hover property doesn't affect the ODD here. I left it in for DX.

app/src/organisms/Navigation/index.tsx
```diff
@@ -157,25 +157,30 @@ const TouchNavLink = styled(NavLink)`
...
     background-color: transparent;
   }
+
+  @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
+    cursor: default;
+
+    &:hover {
+      cursor: pointer;
+    }
+  }
```
mjhuff, line 185: See other comment.

app/src/organisms/RobotSettingsDashboard/TouchscreenBrightness.tsx
```diff
@@ -28,16 +27,6 @@ import {
 import type { Dispatch } from '../../redux/types'
 import type { SetSettingOption } from '../../pages/OnDeviceDisplay/RobotSettingsDashboard'
 
-const BUTTON_STYLE = css`
```
koji, line 31: Actually we need to keep pseudo class for this button.
https://www.figma.com/file/OIdG64Q5cgvEw82ish5Eee/Design-System-(Flex)?node-id=715%3A18873&mode=dev

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -113,6 +113,18 @@ export function ProtocolSetupStep({
...
+  const PUSHED_STATE_STYLING = css`
+    &:active {
+      background-color: ${!disabled
+        ? status === 'general'
+          ? COLORS.darkBlack40
+          : status === 'ready'
+          ? COLORS.green3Pressed
+          : COLORS.yellow3Pressed
+        : ''};
+    }
```
koji: Personally I like `Optional chaining` but we avoid using nested `Optional chaining`.
Maybe creating a new const and if-else/switch and passing the color value?

app/src/organisms/RobotSettingsDashboard/TouchscreenBrightness.tsx
mjhuff: @koji Great catch! Thanks for linking this.

mjhuff:
I've double checked all missing button pressed state, and we should be good to go! See the discussion on RAUT-520 for more of the specific missing pressed states that are resolved with this PR. Directly adjusting the Btn primitive solves buttons that implement it.

app/src/organisms/ChildNavigation/index.tsx
```diff
@@ -66,3 +68,20 @@ export function ChildNavigation({
...
+  &:focus-visible {
+    box-shadow: ${ODD_FOCUS_VISIBLE};
+    background-color: ${COLORS.darkBlack20};
+  }
+  &:disabled {
+    background-color: transparent;
+  }
+  @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
+    cursor: default;
+  }
```
koji, line 86: question
Is this media-query for `dev-odd` mode?

mjhuff: It impacts dev-odd if that's what you're asking. I noticed that on some pressed state, it was impacting the physical ODD, which is why I added it 👍

app/src/organisms/RobotSettingsDashboard/NetworkSettings/index.tsx
```diff
@@ -97,13 +98,22 @@ function NetworkSettingButton({
   networkName,
   onClick,
 }: NetworkSettingButtonProps): JSX.Element {
+  const PUSHED_STATE_STYLING = css`
```
koji: Maybe STYLE since other styles that are defined by `css` use `_STYLE`.

```suggestion
  const PUSHED_STATE_STYLE = css`
```

koji (approved):
I left a couple of comments but the PR lgtm

mjhuff:
> I left a couple of comments but the PR lgtm

Thanks for checking all of this. I know it was a lot. Much appreciated!

PR #13326, opened 2023-08-16
fix(app): Update receiving new protocol toast

mjhuff:
closes RAUT-485 and RAUT-625

# Overview

Updates margin for all toasts. Protocol receiving toasts to include a close button and prevents close until clicked. Animation logic now supports non-stacking toasts. All toasts (desktop & ODD) no longer stack but will auto-fade if a new toast appears. 

### Animation Artifacts

This PR addresses animation artifacts that are often present in ODD animations (currently only Toasts make use of animations on the ODD). CSS properties that rely on the GPU, such as transform and opacity, do not coerce hard acceleration if utilized in a CSS animation, resulting in increased screen tearing on the ODD as the animations run on the CPU.

By using translate3d and filter: opacity, we can coerce hardware acceleration. Adding additional properties such as backface-visibility, perspective, and will-change improve performance. Some of these optimizations are [black magic](https://stackoverflow.com/questions/10014461/why-does-enabling-hardware-acceleration-in-css3-slow-down-performance/10133679#10133679), however the impact will-change is well [documented](https://blog.logrocket.com/when-how-use-css-will-change/). Transform-related CSS properties seem to cause tearing more than opacity, so I reduced the distance the toast travels when fading in/out.

I do notice that tearing occasionally occurs (to a much reduced degree) when newly uploaded protocols do not correctly render on the ProtocolDashboard page. However, I believe this issue is separate from the toast animation logic itself. 

### Toast Refactoring

Toast and ToasterOven components are becoming fragile and due for refactoring if more functionality is to be added. Components should be restructured to better handle state changes, animation logic, and more robust conditional rendering. Additionally, test coverage for useToaster needs implementation.

------------------

Previous exit behavior

https://github.com/Opentrons/opentrons/assets/64858653/0ddd9be5-7bc8-44ff-b5ab-c73d93f3608e

New exit behavior

https://github.com/Opentrons/opentrons/assets/64858653/4c06e174-922d-4a67-aaf6-d78a82097cb9

Previous stacking animation behavior

https://github.com/Opentrons/opentrons/assets/64858653/efbd0836-e531-4c41-8763-15a9e2dc673e

New stacking animation behavior

https://github.com/Opentrons/opentrons/assets/64858653/67273d9f-774f-4f67-8b2a-6bc56de2f25d

# Test Plan

1. Start robot server.
2. Start Desktop app.
3. Start ODD app.

**Exit behavior**
1. Add new protocol.
2. Protocol remains until clicked.

**Non-stacking animation behavior**
1. Add new protocol.
2. Add second protocol.
3. New protocol replaces old protocol. Old protocol transitions out.

**Verify Animation Improvements**
1. Run the above tests on the physical ODD.

# Changelog

- Add useProtocolReceiptToast makeToast options.
- Update Toast component animation for ODD.
- Update ToasterOven bottom margin.
- Optimize Toast ODD animations.
- Toasts no longer stack.

# Review requests

# Risk assessment

low


app/src/atoms/Toast/index.tsx
```diff
@@ -44,9 +44,12 @@ export interface ToastProps extends StyleProps {
   duration?: number
   heading?: string
   displayType?: 'desktop' | 'odd'
+  exitNow?: boolean
 }
 
-const TOAST_ANIMATION_DURATION = 500
+// TODO: (jh: 08/10/23) refactor toast component and render logic.
+
+export const TOAST_ANIMATION_DURATION = 500
```
jerader, line 57: nice, good idea to export this constant to use it in the tests

app/src/atoms/Toast/index.tsx
```diff
@@ -81,10 +85,19 @@ export function Toast(props: ToastProps): JSX.Element {
...
       ? t('close')
       : ''
-  const DESKTOP_ANIMATION_IN = css`
+
+  const ANIMATION_OVERFLOW = `
+  overflow: hidden;
+  `
+  const ODD_ANIMATION_OPTIMIZATIONS = `
+  backface-visibility: hidden;
+  perspective: 1000;
```
jerader, line 99: wow, i've never used this css property before 🤩

app/src/atoms/Toast/index.tsx
```diff
@@ -109,40 +122,70 @@ export function Toast(props: ToastProps): JSX.Element {
...
+    }
+  `
+
+  const oddAnimation = isClosed
+    ? exitNow
+      ? ODD_ANIMATION_FADE_UP_AND_OUT
+      : ODD_ANIMATION_SLIDE_DOWN_AND_OUT
+    : exitNow
+    ? css``
+    : ODD_ANIMATION_SLIDE_UP_AND_IN
```
jerader: this nested ternary is a bit hard to follow, we usually try to avoid them and make if/else or switch statements instead.

jerader: also, the empty `css` is a bit hard to follow since the other css consts are clearly named (such as `ODD_ANIMATION_SLIDE_UP_AND_IN`), maybe you can make this a const that you call here as well.

jerader (approved):
looks great! 🤩 Code is really organized and easy to read - just left a comment about the nested ternary.

app/src/atoms/Toast/index.tsx
```diff
@@ -109,40 +127,88 @@ export function Toast(props: ToastProps): JSX.Element {
...
+    }
+  }
+
+  oddAnimation = isClosed
+    ? exitNow
+      ? ODD_ANIMATION_FADE_UP_AND_OUT
+      : ODD_ANIMATION_SLIDE_DOWN_AND_OUT
+    : exitNow
+    ? css``
+    : ODD_ANIMATION_SLIDE_UP_AND_IN
```
jerader: i think you forgot to delete this!

PR #13328, opened 2023-08-16
docs(api): docstring for new `del deck` behavior

ecormany:
# Overview

Adds description, a warning, and version changed notices to the `deck` entry of the API Reference. This explains things changing w.r.t #13239 

# Test Plan

Check the API Reference in the sandbox.

# Changelog

- description of new behavior
- new warning not to do this in regular protocols
- two `.. versionchanged::` notices

# Review requests

Is this accurate for planned behavior? Is the notice for v2.14 proper?

# Risk assessment

None, docs. (Negative if you include warning people against making their robot go 💥 )

### DIFF
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -924,16 +924,16 @@ class ProtocolContext(CommandPublisher):
         and :py:attr:`loaded_modules` to get a dict of modules.
 
         For :ref:`advanced-control` *only*, you can delete an element of the ``deck`` dict.
-        This only works for deck slots that contain labware objects. For example, if slot 
-        1 contains a labware, ``del protocol.deck['1']`` will free the slot so you can 
+        This only works for deck slots that contain labware objects. For example, if slot
+        1 contains a labware, ``del protocol.deck['1']`` will free the slot so you can
         load another labware there.
-        
+
         .. warning::
-            Deleting labware from a deck slot does not pause the protocol. Subsequent 
-            commands continue immediately. If you need to physically move the labware to 
-            reflect the new deck state, add a :py:meth:`.pause` or use 
-            :py:meth:`.move_labware` instead. 
-        
+            Deleting labware from a deck slot does not pause the protocol. Subsequent
+            commands continue immediately. If you need to physically move the labware to
+            reflect the new deck state, add a :py:meth:`.pause` or use
+            :py:meth:`.move_labware` instead.
+
         .. versionchanged:: 2.15
            ``del`` sets the corresponding labware's location to ``OFF_DECK``.
         """
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -923,22 +923,19 @@ class ProtocolContext(CommandPublisher):
         you can also use :py:attr:`loaded_labwares` to get a dict of labwares
         and :py:attr:`loaded_modules` to get a dict of modules.
 
-        For :ref:`advanced-control` `only`, you can delete an element of the ``deck`` dict.
+        For :ref:`advanced-control` *only*, you can delete an element of the ``deck`` dict.
         This only works for deck slots that contain labware objects. For example, if slot 
         1 contains a labware, ``del protocol.deck['1']`` will free the slot so you can 
         load another labware there.
         
         .. warning::
-            Do not delete labware from a deck slot for protocols run in the Opentrons App 
-            or on the Flex touchscreen. The protocol will continue immediately without 
-            giving you a chance to physically remove the labware. 
-            Use :py:meth:`.move_labware` instead.
+            Deleting labware from a deck slot does not pause the protocol. Subsequent 
+            commands continue immediately. If you need to physically move the labware to 
+            reflect the new deck state, add a :py:meth:`.pause` or use 
+            :py:meth:`.move_labware` instead. 
         
-        .. versionchanged:: 2.14
-           Deleting deck elements raises an error.
-
         .. versionchanged:: 2.15
-           ``del`` sets the labware location to ``OFF_DECK`` instead of deleting it.
+           ``del`` sets the corresponding labware's location to ``OFF_DECK``.
         """
         return self._deck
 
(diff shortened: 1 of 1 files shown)


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -918,14 +919,26 @@ def deck(self) -> Deck:
...
         Rather than filtering the objects in the deck map yourself,
-        you can also use :py:attr:`loaded_labwares` to see a dict of labwares
-        and :py:attr:`loaded_modules` to see a dict of modules.
-
-        For advanced control, you can delete an item of labware from the deck
-        with e.g. ``del protocol.deck['1']`` to free a slot for new labware.
+        you can also use :py:attr:`loaded_labwares` to get a dict of labwares
+        and :py:attr:`loaded_modules` to get a dict of modules.
+
+        For :ref:`advanced-control` `only`, you can delete an element of the ``deck`` dict.
```
SyntaxColoring: Although the backticks around `` `only` `` make it italic, I think it's getting treated semantically as a cross-reference with a missing target. In the rendered HTML, it's a `<cite>` tag.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -918,14 +919,26 @@ def deck(self) -> Deck:
...
+        This only works for deck slots that contain labware objects. For example, if slot 
+        1 contains a labware, ``del protocol.deck['1']`` will free the slot so you can 
+        load another labware there.
+        
+        .. warning::
+            Do not delete labware from a deck slot for protocols run in the Opentrons App 
+            or on the Flex touchscreen. The protocol will continue immediately without 
+            giving you a chance to physically remove the labware. 
+            Use :py:meth:`.move_labware` instead.
```
SyntaxColoring: This feels a bit too severe to me. I don't think it has to do with whether or not you run the protocol through the app; I think it has to do with how the protocol is written and what surrounds the `del` statement.

I would say something like:

> **Warning:**
>
> Unlike `move_labware()`, this will not pause the robot to give you a chance to physically reach in and remove the labware. Subsequent commands will continue immediately. If you need a pause, either add one with `pause()`, or use `move_labware()` instead of `del`.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -918,14 +919,26 @@ def deck(self) -> Deck:
...
+        
+        .. versionchanged:: 2.14
+           Deleting deck elements raises an error.
+
+        .. versionchanged:: 2.15
+           ``del`` sets the labware location to ``OFF_DECK`` instead of deleting it.
```
SyntaxColoring: I've kind of been viewing this as an implementation detail. Do you think this is a meaningful distinction to protocol authors?

I guess here is one trick that's potentially useful:

```python
labware = deck[1]
del deck[1]  # Take it off.
protocol.move_labware(labware, 1)  # Put it back.
```

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -918,14 +919,26 @@ def deck(self) -> Deck:
...
+        load another labware there.
+        
+        .. warning::
+            Do not delete labware from a deck slot for protocols run in the Opentrons App 
+            or on the Flex touchscreen. The protocol will continue immediately without 
+            giving you a chance to physically remove the labware. 
+            Use :py:meth:`.move_labware` instead.
+        
+        .. versionchanged:: 2.14
+           Deleting deck elements raises an error.
```
SyntaxColoring: I'll talk to RSS people about this tomorrow, but at the time of writing, it's still up in the air whether we want to treat this as an officially preserved `apiLevel` thing, or just treat it as a bug and have robot software v7.0 retroactively fix it for apiLevel 2.14.

https://github.com/Opentrons/opentrons/pull/13239#discussion_r1296295121

If you think either one is easier to explain, definitely chime in there.

SyntaxColoring: If we do end up treating it as an official `apiLevel` thing, I might clarify here that deleting deck elements only raises an error in v2.14 *exactly,* not v2.14 *and versions going forward.* In that sense, this is different from our other uses of `versionchanged`.

SyntaxColoring (commented):
Nice! Thanks!

api/src/opentrons/protocol_api/protocol_context.py
ecormany: I don't know if anyone has done something like this, but you would also have to change code like:
```
labware = deck[1]
protocol.pause("time to remove slot 1 labware")
del deck[1]
if not labware:
    # i put in this check because i am smart and don't want to crash
    # some actions that would be done in 2.13 but not now
```

SyntaxColoring: Ah—

I don't think `del deck[1]` has any effect on how `if not labware` evaluates. Not in any `apiLevel`, and neither with this PR nor without it. Is that what you're getting at?

The way it works is something like this:

`if not labware` evaluates whether the object referenced by the `labware` name is truthy. In our system today, a `Labware` object will always be truthy.

`deck[1]` points to the same underlying object as `labware`. Doing `del deck[1]` removes it from `deck`, making it so that `deck[1]` no longer refers to that object. But it doesn't actually delete the underlying object itself, and it doesn't cause the underlying object to stop evaluating as truthy.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: sphinnnnnnx. fixed with asterisks.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: Removing this for now. If it's an error, the error should speak for itself. If it's back-fixed, then the behavior isn't changed from 2.13.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: Leaving this detail. Whether it's material to protocol authors or not, it's a good reminder to _ourselves_ that this implementation changed.

SyntaxColoring (approved):
Looks great. Thank you!

PR #13331, opened 2023-08-17
feat(hardware-testing): add patches for ot3 fixtures

fsinapi:
# Overview

There are a couple of changes we need specifically for the OT-3 fixtures running our QC/diagnostic tests.

We will also have to change the SOP for each fixture to use the new command `make push-ot3-fixture`, similar to how the command for the gravimetric setup is specific to that setup.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ahiuchingau (approved):
LGTM

PR #13332, opened 2023-08-17
fix(hardware-testing): various fixture updates

fsinapi:
# Overview

All fixture patches
- Changed ot3api to never block movement based on estop

Z Stage:
- Disable stall detection
- Refresh position after each stall
- Update positions based on fixture in Shenzhen

Robot assembly:
- Refresh position only after estop is released to avoid exceptions

Robot Stress
- Home for each new settings group
- Added ability to run overnight with `nohup` by ignoring input

Pipette assembly:
- Check & Print the PCBA version for operator review
- Probing deck precision loosened to 0.3mm
- Include deck recalibration in final results
- Check XYZ instead of just X
- Current/speed tests many more points, and has 5 trials for each setting

**NOTE that all fixture software must be uploaded with the command `make push-ot3-fixture` in order to include patches for fixture-specific fixes now.**

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/hardware-testing/fixture_overrides/hardware_fixtures.patch
+++ b/hardware-testing/fixture_overrides/hardware_fixtures.patch
@@ -1,5 +1,5 @@
 diff --git a/api/src/opentrons/config/feature_flags.py b/api/src/opentrons/config/feature_flags.py
-index f46674fb56..f59f3826de 100644
+index f46674fb5..f59f3826d 100644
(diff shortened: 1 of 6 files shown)


ahiuchingau (approved):
makes sense to me

andySigler (approved):
thank you and hope you had fun!

PR #13334, opened 2023-08-17
fix(app): ui changes to moduleInfo and liquid color icon

jerader:
closes RQA-1301 and RQA-1298

# Overview

This PR does a few things
1. adds a white background color to the module info only for thermocyclers.
2. remove the hover from the liquid icon in Protocol details liquids and Setup liquids (I removed it from both places since that icon doesn't actually act like a button in either of those areas)

# Test Plan

in the app, can do this in a dev environment/dev kit, upload a protocol with a thermocycler gen2 and liquids, in the liquid details, hover over the liquid icon and it shouldn't have a hover state. Then go to protocol setup and go to the liquid setup and hover over the liquid icons again, there shouldn't be a hover state.

then go to module setup, check the thermocycler module info, it should be legible now. **Note:** I changed it for both thermocyclers because we might change the thermocycler gen1 render to be an open thermocycler svg so it doesn't hurt to change it in both places

# Changelog

- change css state in `SetupLiquidsList` for the liquids icon
- add a background color to thermocycler modules in `ModuleInfo`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/organisms/Devices/ModuleInfo.tsx
+++ b/app/src/organisms/Devices/ModuleInfo.tsx
@@ -17,6 +17,9 @@ import {
   getModuleDisplayName,
   getModuleDef2,
   MAGNETIC_BLOCK_V1,
+  THERMOCYCLER_MODULE_V2,
+  getModuleType,
+  THERMOCYCLER_MODULE_TYPE,
 } from '@opentrons/shared-data'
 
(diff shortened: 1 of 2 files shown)


koji (approved):
lgtm
thank you for fixing the map view issue.

PR #13335, opened 2023-08-17
feat(shared-data): Modify shared data to support different liquid 'classes'

Laura-Danielle:
# Overview

Since we have to potentially support many liquid properties for low volumes on the p50, we should just bite the bullet and split the liquid directory into different liquid classes. Default will be the fallback in all circumstances while `lowVolumeBottom` (should we just name it `lowVolume` or `lowVolumeDefault`??) will support lower liquid volumes. 

# Test Plan

Throw on a robot to make sure nothing breaks, but it should be a fairly low stakes change.

# Changelog

- Changed directory structure of shared-data
- Updated code to reflect this change

# Review requests

Ok with this structure?

# Risk assessment

Low/medium. Should not be changing internal behavior and test coverage is fairly sufficient.


sfoster1 (approved):
I like it and I think it's a great solution! We gotta test it though this is a lot of changes

SyntaxColoring:
Does this need to merge into the v7.0.0 release branch instead of edge?

Laura-Danielle:
> Does this need to merge into the v7.0.0 release branch instead of edge?

Yes. Good call out.

PR #11654, opened 2022-11-01
docs(api): hide certain private methods

ecormany:
# Overview

Hides a number of methods and parameters that aren't intended for (and likely aren't useful to) API users.

# Changelog

- Adds `:exclude-members:` entries to the API reference page.
- Does not use `:meta private:` in source docstrings because this is [not yet supported by autosummary](https://github.com/sphinx-doc/sphinx/issues/8922).

# Review requests

[Sandbox link](http://sandbox.docs.opentrons.com/RTC-40-hide-private-methods/v2/new_protocol_api.html)
- Check that all the methods that should be hidden are hidden.

# Risk assessment

v low

PR #13336, opened 2023-08-17
docs(api): Document `ProtocolContext.commands()`

SyntaxColoring:
11654# Overview

`ProtocolContext.commands()` is not currently listed on docs.opentrons.com. Before #11654, it was listed, but with an empty body.

It turns out that this method is actually public, in that our software team has recommended it to our support team ([1](https://opentrons.slack.com/archives/C0JLZLPL3/p1575922002338300?thread_ts=1575920726.337400&cid=C0JLZLPL3), [2](https://opentrons.slack.com/archives/C03TM47CW/p1590690760242600)) and our support team has recommended it to our customers ([3](https://support.opentrons.com/s/article/Running-the-robot-using-Jupyter-Notebook)). This PR documents and re-publishes it so it's clearer to everyone that it's official.

# Review requests

http://sandbox.docs.opentrons.com/document_protocolcontext_commands_method/v2/new_protocol_api.html#opentrons.protocol_api.ProtocolContext.commands (may still be building)

Does this definition make sense? Is the terminology good? Anything we want to add? Anything we want to leave out?

Note that things like `opentrons_simulate` also have a "run log" concept, so we should be wary of explaining too much here, unless we try to centralize the explanation somewhere.

# Risk assessment

No risk to functional behavior. Changes are just to documentation.

### DIFF
--- a/api/docs/v2/new_protocol_api.rst
+++ b/api/docs/v2/new_protocol_api.rst
@@ -13,7 +13,7 @@ Protocols and Instruments
 
 .. autoclass:: opentrons.protocol_api.ProtocolContext
    :members:
-   :exclude-members: location_cache, cleanup, clear_commands, commands
+   :exclude-members: location_cache, cleanup, clear_commands
 
 .. autoclass:: opentrons.protocol_api.InstrumentContext
    :members:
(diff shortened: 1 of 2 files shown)


ecormany (approved):
LGTM, thanks for clarifying what this does!

PR #13337, opened 2023-08-17
fix(app): display trash name on droptip command text

ncdiehl11:
# Overview
fix [RAUT-355](https://opentrons.atlassian.net/browse/RLAB-355)

Drop tip display text in the run log currently does not support trash labware other than the OT-2's `opentrons_1_trash_1100ml_fixed`. The command text is incorrectly displayed as "Dropping tip into A1 of ".

Note that the checks for displaying "fixed trash" in the protocol run log  references a hard-coded subset of labware URIs. This is a workaround for the current state in which trash labwares are listed as `loadedLabwares` but not in `loadLabware` commands of protocol analysis. In the future, we will reference the trash's labware display name directly from the `loadLabware` command, as is done with all other labware types.

# Test Plan
- Ensure drop tip command text in run log can display "Dropping tip into A1 of Fixed Trash" for all current trash options (in both App and ODD)

# Changelog
- added all current fixed trash definition URIs when getting labware name from run data

# Review requests
General

# Risk assessment
low

[RAUT-355]: https://opentrons.atlassian.net/browse/RAUT-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/CommandText/utils/getLabwareName.ts
+++ b/app/src/organisms/CommandText/utils/getLabwareName.ts
@@ -7,7 +7,11 @@ import {
 } from '@opentrons/shared-data'
 import { getLabwareDefinitionsFromCommands } from '../../LabwarePositionCheck/utils/labware'
 
-const FIXED_TRASH_DEF_URI = 'opentrons/opentrons_1_trash_1100ml_fixed/1'
+const FIXED_TRASH_DEF_URIS = [
+  'opentrons/opentrons_1_trash_850ml_fixed/1',
+  'opentrons/opentrons_1_trash_1100ml_fixed/1',
+  'opentrons/opentrons_1_trash_3200ml_fixed/1',
+]
 export function getLabwareName(
   analysis: CompletedProtocolAnalysis,
   labwareId: string
@@ -15,7 +19,7 @@ export function getLabwareName(
   const loadedLabware = getLoadedLabware(analysis, labwareId)
   if (loadedLabware == null) {
     return ''
-  } else if (loadedLabware.definitionUri === FIXED_TRASH_DEF_URI) {
+  } else if (FIXED_TRASH_DEF_URIS.includes(loadedLabware.definitionUri)) {
     return 'Fixed Trash'
   } else if (loadedLabware.displayName != null) {
     return loadedLabware.displayName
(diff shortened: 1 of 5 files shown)


app/src/organisms/CommandText/utils/getLabwareName.ts
```diff
@@ -7,15 +7,19 @@ import {
 } from '@opentrons/shared-data'
 import { getLabwareDefinitionsFromCommands } from '../../LabwarePositionCheck/utils/labware'
 
-const FIXED_TRASH_DEF_URI = 'opentrons/opentrons_1_trash_1100ml_fixed/1'
+const FIXED_TRASH_DEF_URIS = [
+  'opentrons/opentrons_1_trash_850ml_fixed/1',
+  'opentrons/opentrons_1_trash_1100ml_fixed/1',
+  'opentrons/opentrons_1_trash_3200ml_fixed/1',
+]
```
shlokamin, line 14: not introduced by your PR, but rather than hard coding these def uris we should be determining whether or not a labware is considered "fixed trash" by reading from its `quirks` parameter. 

i think this would be a good time to make a `isFixedTrash` util in `shared-data` and have it do this logic for us so we can share it in other places

app/src/organisms/InterventionModal/utils/getLabwareNameFromRunData.ts
```diff
@@ -8,7 +8,11 @@ import { getLabwareDefinitionsFromCommands } from '../../LabwarePositionCheck/ut
 
 import type { RunData } from '@opentrons/api-client'
 
-const FIXED_TRASH_DEF_URI = 'opentrons/opentrons_1_trash_1100ml_fixed/1'
+const FIXED_TRASH_DEF_URIS = [
```
shlokamin, line 11: same thing here: this could import a hypothetical `isFixedTrash` util

shlokamin (commented):
this totally works, but the labware def should be the source of truth as to whether or not a labware is a fixed trash or not. whoever initially wrote this code probably just didn't know about the `quirks` field in our labware defs. also it looks like JS checks are failing because of the type changes we made to `GripperData` but should be simple to fix. just need to do a null check

app/src/organisms/CommandText/utils/getLabwareName.ts
ncdiehl11: Got it— I will work on this now

jerader: I agree with Shlok ^ the `quirks` param comes from the labware definition. So you can pull the `const labwareDef` out of the if/else if and find `labwareDef.parameters.quirks`

shlokamin (approved):

PR #13338, opened 2023-08-17
fix(app): bring users to top of setup when they press Back to Top

smb2268:
fix RQA-1305

# Overview
The "Back to Top" button at the bottom of the run setup step was bringing users to the top of the run preview tab instead of the setup tab

# Test Plan
Go to setup for run for a protocol and press "Back to Top" - you should not be redirected to the run preview tab

# Changelog
Change redirect to /setup from /run-preview

# Review requests
See test plan

# Risk assessment
Low


koji (approved):
lgtm

PR #13339, opened 2023-08-17
fix(api): fix run cancellation race condition bug

sanni-t:
Closes RSS-305, RQA-1261

# Overview

Prior to this PR, run cancellations would result in protocol engine cleanup errors. This was mainly due to a bad sequence of things happening in the engine-
1. `engine.stop` was dispatching a `StopAction` and cancelling the queue worker, and then sending a `hardware.halt` command, in that order. This would result in the engine finish/cleanup method being called as soon as the queue worker was cancelled and `hardware.stop()` would be called as part of the cleanup.
2. Some background info:
    - `hardware.halt()` sets the execution manager to a cancelled state, which disallows any further movements until the execution manager is reset.
    - `hardware.stop()` resets the execution manager and attempts to home the robot.
    - both execution_manager methods- `cancel` & `reset`- wait on `asyncio.Condition` to execute
3. Now, as mentioned in (1), the engine cleanup begins as soon as queue worker cancellation is processed. Which means that `hardware.halt()` begins execution at the same time as `engine.finish`, and around the same time as the call to `hardware.stop()`. Depending on the time it takes to execute other cleanup tasks before getting to hardware stop, sometimes, the sequence would be as follows:
    i. `engine.finish()` calls `hardware.stop()`, which makes `execution_manager.reset` gain control of `asyncio.Condition` lock
    ii. immediately after, `hardware.halt` is called, which calls `execution_manager.cancel` but needs to wait on the condition lock.
    iii. the condition lock is released after reset, `execution_manager.cancel` gets control of the condition and sets the state to Cancelled.
    iv. right around this time, `hardware.stop()` calls `hardware.home()`, but now the state has been set to Cancelled and hence results in an `ExecutionCancelledError` which makes the engine abandon the cleanup and raise a `RunAndFinishFailed` error.

SO, this PR gets rid of this race condition by doing two things-
1. moving `hardware.halt` from inside `engine.stop`, and into engine finish method, before `hardware.stop` is called
2. making sure `hardware.halt` is fully executed before moving onto `hardware.stop` to avoid getting into the race condition again. We achieve this by waiting until execution manager cancellation is completed instead of starting the cancellation coroutine and moving onto the next thing before making sure it's done and the state has been updated correctly.

# Test Plan

- Since this PR changes the exit sequence of a run, we should run a protocol on both Flex & OT2 and confirm that:
    - if you cancel the protocol during a move, then the protocol cancels before the move completes
    - the cancellation completes cleanly- ie, the gantry homes in the end and you see a ‘RunCancelledError’ instead of protocol failed errors
    - if an error is raised in the protocol during runtime, the protocol still cancels correctly, drops tip and homes

# Review requests

- make sure this is not going to cause any change in behavior for run cancellations or error handling from 6.3.1

# Risk assessment

High. Touches a pretty important and a bit hairy part of protocol execution & cleanup.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -740,7 +740,7 @@ class OT3API(
         await self._backend.halt()
 
     async def _cancel_execution_and_running_tasks(self) -> None:
-        """Cancel execution man."""
+        """Cancel execution manager and all running (hardware module) tasks."""
         await self._execution_manager.cancel()
 
     async def halt(self) -> None:
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -735,26 +735,27 @@ class OT3API(
 
         asyncio.run_coroutine_threadsafe(_chained_calls(), self._loop)
 
-    async def stop_motors(self) -> None:
+    async def _stop_motors(self) -> None:
         """Immediately stop motors."""
         await self._backend.halt()
 
-    def stop_modules(self) -> None:
+    async def _stop_modules(self) -> None:
         """Immediately stop modules."""
-        asyncio.run_coroutine_threadsafe(self._execution_manager.cancel(), self._loop)
+        await self._execution_manager.cancel()
 
     async def halt(self) -> None:
         """Immediately disengage all present motors and clear motor and module tasks."""
-        await self.disengage_axes(
-            [ax for ax in Axis if self._backend.axis_is_present(ax)]
-        )
-        await self.stop_motors()
-        self.stop_modules()
+        # TODO (spp): check if disengaging motors is really required
+        # await self.disengage_axes(
+        #     [ax for ax in Axis if self._backend.axis_is_present(ax)]
+        # )
+        await self._stop_motors()
+        await self._stop_modules()
 
     async def stop(self, home_after: bool = True) -> None:
         """Stop motion as soon as possible, reset, and optionally home."""
-        await self.stop_motors()
-        self._log.info("Recovering from halt")
+        await self.halt()
+        self._log.info("Resetting OT3API")
         await self.reset()
 
         if home_after:
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -754,7 +754,6 @@ class OT3API(
 
     async def stop(self, home_after: bool = True) -> None:
         """Stop motion as soon as possible, reset, and optionally home."""
-        await self.halt()
         self._log.info("Resetting OT3API")
         await self.reset()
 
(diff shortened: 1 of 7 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -735,26 +735,27 @@ async def _chained_calls() -> None:
...
-    async def stop_motors(self) -> None:
+    async def _stop_motors(self) -> None:
         """Immediately stop motors."""
         await self._backend.halt()
 
-    def stop_modules(self) -> None:
+    async def _stop_modules(self) -> None:
         """Immediately stop modules."""
-        asyncio.run_coroutine_threadsafe(self._execution_manager.cancel(), self._loop)
+        await self._execution_manager.cancel()
```
sanni-t: `execution_manager.cancel` does more than just stop modules. Rename this method and update docstring.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -292,7 +292,6 @@ async def stop(self) -> None:
         action = self._state_store.commands.validate_action_allowed(StopAction())
         self._action_dispatcher.dispatch(action)
         self._queue_worker.cancel()
-        await self._hardware_stopper.do_halt()
```
sanni-t, line 295: Test if removal of the halt as part of cancellation delays the stopping of the robot motion. (try adding delays between queue worker cancellation & hardware stop from engine.finish)

sanni-t:
Tested the changes on OT2 and verified:
- [x] if you cancel the protocol during a move, then the protocol cancels before the move completes
- [x] the cancellation completes cleanly- ie, the gantry homes in the end and you see a ‘RunCancelledError’ instead of protocol failed errors
- [x] if an error is raised in the protocol during runtime, the protocol still cancels correctly, drops tip and homes

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -735,28 +735,29 @@ async def _chained_calls() -> None:
...
-    async def stop_motors(self) -> None:
+    async def _stop_motors(self) -> None:
         """Immediately stop motors."""
         await self._backend.halt()
 
-    def stop_modules(self) -> None:
-        """Immediately stop modules."""
-        asyncio.run_coroutine_threadsafe(self._execution_manager.cancel(), self._loop)
+    async def _cancel_execution_and_running_tasks(self) -> None:
+        """Cancel execution man."""
```
sanni-t:
```suggestion
        """Cancel execution manager and all running (hardware module) tasks."""
```

api/src/opentrons/protocol_engine/protocol_engine.py
sanni-t: It does. So moved halt as a separate call but inside `engine.finish`, before calling `hardware.stop`

jbleon95 (approved):
Reviewed code live review, and was able to test PR on Flex

- [x] Cancelled protocol during move operation and protocol cancelled before move was complete
- [x] Cancellation completes cleanly, homing and ending with a `RunStoppedError`
- [x] Letting a protocol fail during a run after tip pick up has the protocol home, drop tips, and end the run

PR #13341, opened 2023-08-17
fix(app): fix protocol run header banner render issue

koji:
# Overview
The current E-stop related components don't support well. This PR fix that issue.
close RQA-1314 and RQA-1325

# Test Plan
run a protocol on OT-2 and `Run Failed` banner from E-stop won't show up in the header

# Changelog
- update EstopTakeover component to return null when a robot is OT-2
- update ProtocolRunHeader component to avoid rendering `Run Failed.` banner when running a protocol on OT-2

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/App/DesktopApp.tsx
+++ b/app/src/App/DesktopApp.tsx
@@ -144,7 +144,7 @@ function RobotControlTakeover(): JSX.Element | null {
   const isOT3 = useIsOT3(robotName)
 
   // E-stop is not supported on OT2
-  if (isOT3) return null
+  if (!isOT3) return null
 
   if (deviceRouteMatch == null || robot == null || robotName == null)
     return null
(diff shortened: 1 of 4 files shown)


app/src/App/DesktopApp.tsx
```diff
@@ -141,6 +141,10 @@ function RobotControlTakeover(): JSX.Element | null {
   const params = deviceRouteMatch?.params as DesktopRouteParams
   const robotName = params?.robotName
   const robot = useRobot(robotName)
+  const isOT3 = useIsOT3(robotName)
+
+  // E-stop is not supported on OT2
+  if (isOT3) return null
```
jerader: don't we want this to be the opposite? 

```suggestion
  if (!isOT3) return null
```

koji: oooooooh, thanks.
You are absolutely right!

koji:
@jerader thank you for pointing out the issue.
I will close this PR since I think using this in [the following PR](https://github.com/Opentrons/opentrons/pull/13359) would be efficient.

PR #13343, opened 2023-08-18
feat(shared-data): Modify plunger position shape to support different liquid class requirements

Laura-Danielle:
# Overview

We need to be able to switch between plunger positions based on changing liquid classes.

# Test Plan
- Put this on a robot and make sure things work as expected

# Changelog

- Update the plunger position shape
- Set everything to use the default plunger positions for now

# Review requests

Please let me know if there is anything you wish to change

# Risk assessment

This change could be high. I don't know how many hardware testing scripts might be reaching into the internals of the hardware configs directly.


sfoster1 (approved):
Looks great, thank you!
Let's not merge this into its current base pr - let's wait until that's merged and then merge this so we get two commits.

SyntaxColoring:
Tested with this protocol.

```python
requirements = {
    "apiLevel": "2.15",
    "robotType": "Flex",
}


def run(protocol):
    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_200ul", "C3")
    reservoir = protocol.load_labware("nest_12_reservoir_15ml", "D2")

    left = protocol.load_instrument("flex_1channel_50", mount="left")
    right = protocol.load_instrument("flex_8channel_1000", mount="right")

    left.pick_up_tip(tip_rack["A1"])
    right.pick_up_tip(tip_rack["A2"])

    for v in [25, 50]:
        left.aspirate(v, reservoir.wells()[0])
        left.dispense(v / 2, reservoir.wells()[1])
        left.blow_out(reservoir.wells()[2])

    for v in [100, 200]:
        right.aspirate(v, reservoir.wells()[0])
        right.dispense(v / 2, reservoir.wells()[1])
        right.blow_out(reservoir.wells()[2])

    left.drop_tip(tip_rack["A1"])
    right.drop_tip(tip_rack["A2"])
```

It looked to me like it was working, but it's hard for me to tell if the blowouts are actually blowing out farther than the normal bottom position.

PR #13344, opened 2023-08-18
fix(app): copy QA from August 17

ecormany:
# Overview

This PR contains a number of localization changes (and corresponding tests) found in a copy QA session done on Flex hardware on August 17. The goal is to have in-app copy match designs in Figma as exactly as possible (without significant refactoring for very small gains, like capitalization only).

The changes here are strings that I could easily find and was reasonably confident only applied to the screens we were looking at on-device. Some more complicated changes that may require refactoring will be ticketed separately.

Partially addresses [RUXD-1061](https://opentrons.atlassian.net/browse/RUXD-1061).

# Test Plan

Check that these strings render properly in the touchscreen app.
We've already verified the new text against what's in Figma.

# Changelog

- string changes across about a dozen localization and organism files
- copy-paste text to corresponding tests

# Review requests

1. Verify the new text appears in the appropriate places in the touchscreen app.
2. Make sure we haven't polluted the desktop app with text that only applies to touchscreen use.
3. Make sure we haven't broken anything else across either app due to highly shared strings.

# Risk assessment

Low. This barely touches code, but there's always the possibility of breaking an interface element or otherwise worsening the user experience.

[RUXD-1061]: https://opentrons.atlassian.net/browse/RUXD-1061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/assets/localization/en/labware_position_check.json
```diff
@@ -14,8 +14,8 @@
   "confirm_position_and_move": "Confirm position, move to slot {{next_slot}}",
   "confirm_position_and_pick_up_tip": "Confirm position, pick up tip",
   "confirm_position_and_return_tip": "Confirm position, return tip to Slot {{next_slot}} and home",
-  "ensure_nozzle_is_above_tip": "Ensure that the pipette nozzle furthest from you is centered above and level with the top of the tip in the A1 position. If it isn't, use the controls below or your keyboard to jog the pipette until it is properly aligned.",
-  "ensure_tip_is_above_well": "Ensure that the pipette tip furthest from you is centered above and level with well A1 on the labware. If it isn't, use the controls below or your keyboard to jog the pipette until it is properly aligned.",
+  "ensure_nozzle_is_above_tip": "Ensure that the pipette nozzle furthest from you is centered above and level with the top of the tip in the A1 position.  If it isn't, tap <bold>Move pipette</bold> and then jog the pipette until it is properly aligned.",
```
ecormany: These might be trouble if they're shared between app and touchscreen. If so, refactor to use old text in app and new on touchscreen.

b-cooper: These are shared between desktop and the touchscreen.  We can add some branching logic so that the previous text shows on Desktop, and your new revision only shows on the Touchscreen if that's what you'd prefer?

app/src/assets/localization/en/app_settings.json
```diff
@@ -33,8 +33,8 @@
   "discovery_timeout": "Discovery timed out.",
   "download_update": "Download app update",
   "enable_dev_tools_description": "Enabling this setting opens Developer Tools on app launch, enables additional logging and gives access to feature flags.",
-  "enable_dev_tools": "Enable Developer Tools",
-  "error_boundary_description": "You need to restart your robot. Then download the robot logs from the Opentrons App and send them to support@opentrons.com for assistance.",
+  "enable_dev_tools": "Developer Tools",
```
koji, line 35: This change has an impact on Desktop app.
Do we want to change Desktop app and Touchscreen app both?

ecormany: Yes, I think that's good. I find settings that start with "Enable" to be redundant and harder to read.

koji: Okay, thanks!

koji:
@ecormany 
Shlok told me that we decided to allow users to use more than 17 characters for a robot name.
Have you heard that?

koji (approved):
lgtm, thank you for updating copies!

PR #13346, opened 2023-08-18
docs(api): Flex revisions for Advanced Control

jwwojak:
# Overview

This is for revisions to the API doc, [Advanced Control](https://docs.opentrons.com/v2/new_advanced_running.html) (`new_advanced_running.rst`). Changes make this less OT-2 specific, incorporate Flex into the doc, and are very light.

# Test Plan

* Our usual review process.

* Sandbox:  http://sandbox.docs.opentrons.com/advanced-controls-revisions/v2/new_advanced_running.html

# Changelog

Changes include removing single OT-2 mentions and including both robots in text. Also, we're using generic terms for both like "robots".

Some minor text changes.

# Review requests

No specific requests.

# Risk assessment

Low. Changes are trivial.

### DIFF
--- a/api/docs/v2/new_advanced_running.rst
+++ b/api/docs/v2/new_advanced_running.rst
@@ -7,22 +7,22 @@ Advanced Control
 
 As its name implies, the Python Protocol API is primarily designed for creating protocols that you upload via the Opentrons App and execute on the robot as a unit. But sometimes it's more convenient to control the robot outside of the app. For example, you might want to have variables in your code that change based on user input or the contents of a CSV file. Or you might want to only execute part of your protocol at a time, especially when developing or debugging a new protocol.
 
-The OT-2 offers two ways of issuing Python API commands to the robot outside of the app: through Jupyter Notebook or on the command line with ``opentrons_execute``.
+The Python API offers two ways of issuing commands to the robot outside of the app: through Jupyter Notebook or on the command line with ``opentrons_execute``.
 
 Jupyter Notebook
 ----------------
 
-The OT-2 runs a `Jupyter Notebook <https://jupyter.org>`_ server on port 48888, which you can connect to with your web browser. This is a convenient environment for writing and debugging protocols, since you can define different parts of your protocol in different notebook cells and run a single cell at a time.
+The Flex and OT-2 run `Jupyter Notebook <https://jupyter.org>`_ servers on port 48888, which you can connect to with your web browser. This is a convenient environment for writing and debugging protocols, since you can define different parts of your protocol in different notebook cells and run a single cell at a time.
 
 .. note::
-    The Jupyter Notebook server only supports Python Protocol API versions 2.13 and earlier. Use the Opentrons App to run protocols that require functionality added in newer versions.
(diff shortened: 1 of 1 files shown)


api/docs/v2/new_advanced_running.rst
```diff
@@ -84,8 +84,9 @@ Creating the dummy protocol requires you to:
 For example, the following dummy protocol will use a Flex 50 µL pipette to enable Labware Position Check for a Flex tip rack, NEST reservoir, and NEST flat well plate.
 
 .. code-block:: python
+    :substitutions:
 
-    requirements = {'apiLevel': '|apiLevel|', 'robotType': 'Flex'}
+    requirements = {'robotType': 'Flex','apiLevel': '|apiLevel|'}
```
ecormany:
```suggestion
    requirements = {'robotType': 'Flex', 'apiLevel': '|apiLevel|'}
```
tiny style change, won't affect functionality

api/docs/v2/new_advanced_running.rst
```diff
@@ -84,8 +84,9 @@ Creating the dummy protocol requires you to:
...
 
     def run(protocol):
         tips = protocol.load_labware('opentrons_flex_96_tiprack_50ul', 'D1')
```
ecormany: Oops, I may have gotten a bit out over my skis here. Since Jupyter support is still forthcoming for 2.15, this is 🔮 code, not current reality.

In the meantime, we could revert this to the example from the previous iteration (updating the API level to 2.13):
https://github.com/Opentrons/opentrons/blob/a7a198e3254c6372cf3be8014c578027c6e438de/api/docs/v2/robot_position.rst?plain=1#L133-L141

api/docs/v2/new_advanced_running.rst
```diff
@@ -99,13 +100,13 @@ After importing this protocol to the Opentrons App, run Labware Position Check t
 
 .. code-block:: python
 	
-    labware_1 = protocol.load_labware("opentrons_flex_96_tiprack_200ul", location="D1")
+    labware_1 = protocol.load_labware('opentrons_flex_96_tiprack_200ul', location='D1')
```
ecormany: …and here we should match what we do above. if we want to match the previous (OT-2, 2.13) revision:

https://github.com/Opentrons/opentrons/blob/a7a198e3254c6372cf3be8014c578027c6e438de/api/docs/v2/robot_position.rst?plain=1#L147-L154

either way, we should probably keep the double quotes instead of single quotes — yes, they don't match the other code we wrote for this page, but this is _autogenerated_ and is meant to represent exactly what you would get from the Get Labware Offset Data feature in the app.

ecormany (approved):
Looks good. See comments about maybe rolling the example back to OT-2 / API version 2.13 for now.

api/docs/v2/new_advanced_running.rst
jwwojak: I recall seeing this while going through the doc:

>  requirements = {'robotType': 'Flex', 'apiLevel': '|apiLevel|'}

and thought, "hmmm, with 2.15, this isn't going to work given the note about versions" but just went with it anyway. Will reset to 2.13 and change the code to match an OT-2.

api/docs/v2/new_advanced_running.rst
jwwojak: Fixing to match output.

PR #13347, opened 2023-08-18
fix(app-shell): match opentrons usb devices on Windows

shlokamin:
# Overview
This PR fixes usb interaction between an Opentrons app running on an Windows machine and a Flex.

Serialport exposes vendor ids and product ids as string casted hex digits with letters lowercased on Mac and Linux. For Windows, vendor ids and product ids get exposed as string casted hex digits with letters uppercased 🙃 

closes RQA-1317

# Changelog

- Match Opentrons usb devices on Windows

# Review requests

I verified this works, but if you've got a windows machine handy go ahead and try to connect it to a flex and see that the robot shows up in the app and you can interact with it. 

# Risk assessment

Low, its already broken

### DIFF
--- a/app-shell/src/usb.ts
+++ b/app-shell/src/usb.ts
@@ -115,8 +115,12 @@ function startUsbHttpRequests(dispatch: Dispatch): void {
     .then((list: PortInfo[]) => {
       const ot3UsbSerialPort = list.find(
         port =>
-          port.productId === DEFAULT_PRODUCT_ID &&
-          port.vendorId === DEFAULT_VENDOR_ID
+          port.productId?.localeCompare(DEFAULT_PRODUCT_ID, 'en-US', {
+            sensitivity: 'base',
+          }) === 0 &&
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me! Very goofy

brenthagen (approved):
🔠 🔡

koji (commented):
Tested this branch with parallels.
For Windows worked as expected
For Linux(Ubuntu) didn't recognize the connection

shlokamin:
> For Linux(Ubuntu) didn't recognize the connection

😭

shlokamin:
testing with koji rn, we are able to see serial port detect an opentrons usb device with the right vid and pid, and we're seeing all of the right machinery get called. seems like it might be an issue cuz of parallels. 

we also confirmed it also does not work in edge, so this is not a regression. we'll merge for now and test on a real linux machine (if we an find one...)

PR #13348, opened 2023-08-18
fix(app): use all commands query for ODD intervention modal

brenthagen:
# Overview

use all commands query data for ODD intervention modal, as desktop app does

closes RQA-1286

# Test Plan

 - manually verified ODD pause intervention modal timer matches desktop app, counts up from time of pause

# Changelog

 - Fixes ODD intervention modal pause timer start time

# Review requests

# Risk assessment

low


b-cooper (approved):
🦅 👀   good catch

PR #13333, opened 2023-08-17
docs(api): new page on adapting OT-2 protocols for Flex

ecormany:
# Overview

New page of the PAPI docs, explaining the minimal steps to turn an OT-2 protocol into a Flex protocol. Based on some similar material I wrote for the Flex manual, but adjusted to fit into these docs, including cross-references to other pages, methods, etc.

# Test Plan

- Check prose and formatting in the [sandbox](http://sandbox.docs.opentrons.com/docs-adapt-ot2-protocols/v2/adapting_ot2_flex.html)
- Check that all standalone code snippets pass analysis

# Changelog

- Brand new page, Adapting OT-2 Protocols for Flex
- Insert new page in the TOC

# Review requests

👍 ❓ 

# Risk assessment

None, docs.

PR #13349, opened 2023-08-19
docs(api): including Flex and OT-2 on homepage and in sidebar

ecormany:
# Overview

Some final housekeeping for deploying docs that include Flex: this is not the "OT-2 API" anymore!

In general, this includes:
- Cleaning up language that assumes the OT-2 is the only robot. This was mostly on the Welcome page, but also on the Building Block Commands page and _passim_.
- Changing the site name from `OT-2 API V2` to `Python Protocol API v2`. For this to look good at all, I had to make the sidebar wider, which was a big 🐰 🕳️ — there's a bunch of template changes and CSS overrides.
- Bidding adieu to `writing.rst`. This page was already deliberately orphaned. There are no internal links to it, and all external links to it will automatically redirect to the docs homepage (which honestly has a pretty robust description of what it takes to write a basic protocol).

# Test Plan

- Did a cross-file search for `OT-2` and it looks like we are up to date just about everywhere (except `new_advanced_running.rst`, but that's because this branch hasn't pulled in changes from #13346)
- Tried my best to check all permutations of browsers, window sizes, and page elements to make sure my CSS changes aren't breaking things badly. In particular:
  - Tried Safari, Chrome, Firefox, Edge (caveat: all for Mac).
  - Freely resized windows.
  - Scrolled all pages to look for elements that needed new `clear: none` attributes. Inline elements like `blockquote` and certain `div`s wouldn't appear above the height of the sidebar otherwise, making giant gaps in the page.

# Changelog

- change project name in sidebar
- make sidebar wider to fit new name
- bunch o' CSS hacks to make the sidebar sticky and scroll when too tall for viewport
  - including using a flexbox to get the sidebar at the bottom of the page on mobile breakpoints
  - and ditching the double droplet logo (it's already in the header)
- prose changes on Welcome page
- removed some links, fixed other links on Welcome page
- prose changes on Building Block Commands page
- _damnatio memoriae_ for Using Python for Protocols page

# Review requests

Please check a bunch of browsers and viewports, because these CSS changes could work in one but not another. I want to deploy an improved docs site, not a broken one!

# Risk assessment

Low overall. Medium for docs given the amount of template hackery.


api/docs/v2/conf.py
```diff
@@ -204,7 +205,7 @@
 # The name of an image file (relative to this directory) to place at the top
 # of the sidebar.
 #
-html_logo = '../img/logo.png'
+# html_logo = '../img/logo.png'
```
jerader, line 208: can you delete this line now instead of commenting it out?

api/docs/static/override_sphinx.css
```diff
@@ -50,6 +50,34 @@ div.body h1 + p {
...
+
+pre, div[class*="highlight-"], blockquote, blockquote::after, div.admonition::after {
+    clear: none;
+}
+
+div.sphinxsidebar {
+    position: sticky;
+    top: 150px;
+    align-self: flex-start;
+    max-height: calc(100vh - 150px);
```
jerader, line 71: the 150px is subtracting the length of the header?

jerader (approved):
i think the css looks good, nice work!

api/docs/static/override_sphinx.css
jwwojak, line 1: Sticky side bar code, I'm guessing. Such a great feature.

api/docs/v1/conf.py
jwwojak, line 1: ah `fixed_sidebar`. nice

api/docs/v2/new_atomic_commands.rst
```diff
@@ -6,7 +6,7 @@
 Building Block Commands
 #######################
 
-Building block, or basic, commands are the smallest individual actions that can be completed on an OT-2.
+Building block, or basic, commands are the smallest individual actions that can be completed on an Opentrons robot.
```
jwwojak, line 9: Did I accidentally leave "OT-2" in this intro sentence? This change is in https://github.com/Opentrons/opentrons/pull/13333

@ecormany

jwwojak (approved):
Looks _fantastic_ to me. Ship it!

api/docs/v2/conf.py
ecormany: This whole section, including this line, comes pre-commented in the conf template. So I think it makes sense to leave it in.
https://github.com/sphinx-doc/sphinx/blob/6183b6abee561fc71c88fc66416d31a8f8ca009f/doc/_static/conf.py.txt#L140-L143

SyntaxColoring (approved):
Tested briefly on desktop and iOS Chrome. Appears to be working as intended. I'm probably not qualified to review the CSS. :)

-----

Generally, I'd appreciate comments in the .css file explaining what we're overriding from the template and why. If I'm reading this file, I'm not sure what's safe to adjust and what would break some obscure part of the layout.

-----

The width of the content is messed up on some pages ([1](https://github.com/Opentrons/opentrons/assets/3236864/7755e0ae-9e04-4371-afc4-c26702c1b370), [2](https://github.com/Opentrons/opentrons/assets/3236864/b4d0f1d5-7afc-4cda-b9f4-74f86642a082)), but that appears to have been a problem before this PR.

-----

From a UX point of view, I'm a bit concerned that when the sidebar scrolls separately from the main content, it's too easy to miss that there's content below the fold. Especially on a Mac where the scrollbars are hidden by default.

-----

I also noticed these remaining mentions of "OT-2":

`api/docs/v2/new_atomic_commands.rst`:

> If no location is specified, the **OT-2** will drop the tip in the fixed trash in slot 12 of the deck.

> The door safety switch feature flag has been added to the **OT-2** software since the 3.19.0 release.

ecormany:
> Generally, I'd appreciate comments in the .css file explaining what we're overriding from the template and why. If I'm reading this file, I'm not sure what's safe to adjust and what would break some obscure part of the layout.

Fair. Will add some notes.
 
> The width of the content is messed up on some pages ([1](https://github.com/Opentrons/opentrons/assets/3236864/7755e0ae-9e04-4371-afc4-c26702c1b370), [2](https://github.com/Opentrons/opentrons/assets/3236864/b4d0f1d5-7afc-4cda-b9f4-74f86642a082)), but that appears to have been a problem before this PR.

Ooh it looks like `get_protocol_api` has something really long with no punctuation or spaces. Fortunately you can zoom to the correct width on iOS.

> From a UX point of view, I'm a bit concerned that when the sidebar scrolls separately from the main content, it's too easy to miss that there's content below the fold. Especially on a Mac where the scrollbars are hidden by default.

Yeah, I'm willing to pass the buck to Apple there. Even setting `scroll` instead of `auto` won't make the bar appear — it's totally controlled by the system setting.

> > If no location is specified, the **OT-2** will drop the tip in the fixed trash in slot 12 of the deck.

The current Flex behavior is similar but will likely change soon. So this is deliberately attributed to OT-2.
 
> > The door safety switch feature flag has been added to the **OT-2** software since the 3.19.0 release.

And Flex can't turn off the door pause switch. Will add that info to the page soon.

PR #13351, opened 2023-08-21
refactor(app): adapter split test coverage and cleanup components

jerader:
closes RAUT-624

# Overview

This is a follow up PR to the adapter split work. This PR does a few things:
1. increase test coverage of app adapter split components
2. cleans up `offsetTable` in `ResultSummary` to not have `protocolData` as a prop
3. cleans up `getDisplayLocation` to not have that optional boolean and instead pass down `i18n` to `titleCase` locations that are only `slotName`

# Test Plan

- review the tests
- does my change to `getDisplayLocation` make sense?

# Changelog

- add test coverage through the app adapter split components
- remove `protocolData` as as prop in `offsetTable`
- refactor `getDisplayLocation` to not have an optional prop and pass down `i18n` as a prop to capitalize the slot name locations

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/SetupLabwarePositionCheck/CurrentOffsetsTable.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/SetupLabwarePositionCheck/CurrentOffsetsTable.tsx
@@ -66,7 +66,7 @@ export function CurrentOffsetsTable(
   props: CurrentOffsetsTableProps
 ): JSX.Element {
   const { currentOffsets, commands, labware, modules } = props
-  const { t } = useTranslation(['labware_position_check', 'shared'])
+  const { t, i18n } = useTranslation(['labware_position_check', 'shared'])
   const defsByURI = getLoadedLabwareDefinitionsByUri(commands)
   const isLabwareOffsetCodeSnippetsOn = useSelector(
     getIsLabwareOffsetCodeSnippetsOn
@@ -95,7 +95,7 @@ export function CurrentOffsetsTable(
                   offset.location,
                   getLabwareDefinitionsFromCommands(commands),
                   t,
-                  true // capitalize Slot
+                  i18n
                 )}
               </OffsetTableDatum>
               <OffsetTableDatum>{labwareDisplayName}</OffsetTableDatum>
--- a/app/src/organisms/Devices/ProtocolRun/utils/__tests__/getLabwareOffsetLocation.test.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/utils/__tests__/getLabwareOffsetLocation.test.tsx
@@ -9,8 +9,8 @@ import type {
   LoadedLabware,
   LoadedModule,
   ProtocolAnalysisFile,
+  LabwareDefinition2,
 } from '@opentrons/shared-data'
-import type { LabwareDefinition2 } from '@opentrons/shared-data'
 
 jest.mock('../getLabwareLocation')
 jest.mock('../getModuleInitialLoadInfo')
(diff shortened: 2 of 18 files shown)


app/src/organisms/LabwarePositionCheck/utils/getDisplayLocation.ts
```diff
@@ -1,23 +1,23 @@
...
-  //  TODO: (jr, 8/18/23): perhaps find a new route here and refactor this,
-  //  only some instances of slot need to be capitalized depending on the parent component
-  capitalizeSlot?: boolean
+  i18n: i18n
 ): string {
-  const slotDisplayLocation = t('slot_name', { slotName: location.slotName })
+  const slotDisplayLocation = i18n.format(
+    t('slot_name', { slotName: location.slotName }),
+    'titleCase'
+  )
```
jerader, line 20: the reason why i18n had to be passed down as a prop instead of keeping it in the parent level is because the display location can be either 1) just slotname 2) module + slot name 3) adapter + module + slot name. And we want to `titleCase` the display location's slot name.

app/src/organisms/Devices/ProtocolRun/utils/__tests__/getLabwareOffsetLocation.test.tsx
```diff
@@ -1,14 +1,27 @@
...
 import { getLabwareOffsetLocation } from '../getLabwareOffsetLocation'
 import { getLabwareLocation } from '../getLabwareLocation'
 import { getModuleInitialLoadInfo } from '../getModuleInitialLoadInfo'
-import type { LoadedModule, ProtocolAnalysisFile } from '@opentrons/shared-data'
+import {
+  getLabwareDefURI,
+  LoadedLabware,
+  LoadedModule,
+  ProtocolAnalysisFile,
+} from '@opentrons/shared-data'
```
koji: can we move these lines after `jest-when`?

app/src/organisms/Devices/ProtocolRun/SetupLabwarePositionCheck/CurrentOffsetsTable.tsx
```diff
@@ -95,7 +95,7 @@ export function CurrentOffsetsTable(
                   offset.location,
                   getLabwareDefinitionsFromCommands(commands),
                   t,
-                  true // capitalize Slot
+                  i18n
```
koji, line 98: Question

Is there any reason to avoid using `useTranslation` in a child component?

app/src/organisms/LabwarePositionCheck/utils/getDisplayLocation.ts
koji: Maybe using `titleCase` directly would be another option since I think less props are always easy for us.
Maybe asking about this to the team?

jerader: Hmm, I don't think that works since a location of labware on an adapter on a module = `{{adapter}} on {{module}} in {slot}}`. If we used `titleCase`, those `on` and `in` would be capitalized as well.

Brian left this comment on my adapter split PR, he spoke to Ed about it being acceptable to always capitalize `Slot` but we don't want to capitalize every word in the string. https://github.com/Opentrons/opentrons/pull/13237#discussion_r1298795872

koji: titleCase can leave `on` and `in` as lowerCases and that is what titleCase does.
https://titlecase.com/

koji (approved):
🧪

PR #13353, opened 2023-08-21
fix(app, shared-data): return tip in LPC to default wellLocation

jerader:
closes RAUT-638

# Overview

During the schemav7 migration, I changed the wellLocation type to not include `default` but this is only the case for pipetting commands that are NOT `dropTip`. So this resulted in issues with LPC. But this PR fixes it and allows the `dropTip` `wellLocation` to be `default`

# Test Plan

Test on LPC, test early exit and completion when you return the tip.

# Changelog

- add `default` type to `wellLocation` for `dropTip` commands
- change well location in `ReturnTip` and early exit in LPC to be `default`, fix test

# Review requests

see test plan

# Risk assessment

low


b-cooper (approved):
Looks good to me! thanks for the speedy PR 🚤

SyntaxColoring:
Tested on RSSBot. All looks good. Thank you!

PR #13355, opened 2023-08-21
refactor(api): split backend tip_action into two functions

caila-marashaj:
## Overview
Currently, the backend `tip_action` function is used both for regular tip action moves, as well as for homing the gear motors at a constant velocity. This makes the code a little bit cluttered, as the backend function takes in either a list of moves or a distance and velocity to accommodate this, and then needs to determine which one it's received. 

Instead, the api can just use two separate functions- one for homing, and one for all other moves involving the gear motor.

## Changelog
- create a `home_gear_motors` function in `ot3api.py` that fast homes the gear motors if possible, or homes at constant velocity if not
- create a `home_gear_motors` function in `ot3controller.py` that is used only for homing at constant velocity
- `ot3controller.py::tip_action` will only take in a list of moves

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -642,12 +642,11 @@ class OT3Controller:
             positions = await asyncio.gather(*coros)
         # TODO(CM): default gear motor homing routine to have some acceleration
         if Axis.Q in checked_axes:
-            await self.tip_action(
+            await self.home_gear_motors(
                 distance=self.axis_bounds[Axis.Q][1] - self.axis_bounds[Axis.Q][0],
                 velocity=self._configuration.motion_settings.max_speed_discontinuity.high_throughput[
                     Axis.to_kind(Axis.Q)
                 ],
-                tip_action="home",
             )
         for position in positions:
             self._handle_motor_status_response(position)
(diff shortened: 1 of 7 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -642,12 +642,11 @@ async def home(
             positions = await asyncio.gather(*coros)
         # TODO(CM): default gear motor homing routine to have some acceleration
         if Axis.Q in checked_axes:
-            await self.tip_action(
+            await self.home_gear_motors(
```
Laura-Danielle: let's call this `home_tip_motors` since we refer to them as `tip` motors in the backend.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -665,26 +664,33 @@ def _filter_move_group(self, move_group: MoveGroup) -> MoveGroup:
             )
         return new_group
 
+    async def home_gear_motors(
+        self,
+        distance: float,
+        velocity: float,
+        back_off: bool = True,
+    ) -> None:
+        move_group = create_gear_motor_home_group(
```
Laura-Danielle: let's call this function `create_tip_motor_home_group`

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -665,26 +664,33 @@ def _filter_move_group(self, move_group: MoveGroup) -> MoveGroup:
...
+            float(distance), float(velocity), back_off
```
Laura-Danielle: why do you need this cast?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -814,6 +814,41 @@ async def home_plunger(self, mount: Union[top_types.Mount, OT3Mount]) -> None:
...
+
+        # if position is not known, move toward limit switch at a constant velocity
+        if not any(self._backend.gear_motor_position):
+            await self._backend.home_gear_motors(
+                distance=self._backend.axis_bounds[Axis.Q][1],
+                velocity=homing_velocity,
+            )
+            return
+
+        current_pos_float = axis_convert(self._backend.gear_motor_position, 0.0)[
```
Laura-Danielle, line 830: Is this keyed by `Axis.Q`? Could we just index in with Axis.Q?

Laura-Danielle (commented):
Couple initial comments, but overall looks great!

api/src/opentrons/hardware_control/ot3api.py
caila-marashaj: this is because the backend `gear_motor_position` is a `Dict` with `NodeId` keys, so the value is keyed by `NodeId.pipette_left`

PR #13357, opened 2023-08-21
docs(api,robot-server): Update readmes to avoid overspecifying as "OT-2"

SyntaxColoring:
# Changelog

* Update `api` and `robot-server` readmes to mention Flex, not just OT-2.
* Fix up product names to use their official spelling. "OT-2" instead of "OT 2", "OT-One" instead of "OT-1", etc.

# Review requests

None in particular.

# Risk assessment

No risk. Docs only.

### DIFF
--- a/api/README.rst
+++ b/api/README.rst
@@ -16,13 +16,13 @@ Opentrons
 Introduction
 ------------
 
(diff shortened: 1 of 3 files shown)


b-cooper (approved):
Looks good to me 🖊️

PR #13358, opened 2023-08-21
fix(components): if a labware has no wells, labels should default properly

b-cooper:
# Overview

Instead of iterating over undefined, default label column to an empty array if no wells on the targeted labware.

Closes RQA-1350

# Risk assessment
low


shlokamin (approved):
weird, i wonder if other type systems catch this

PR #13360, opened 2023-08-21
fix(engine): fix calculation for labware height on modules

sanni-t:
# Overview

I had forgotten to account for stacking overlap during calculation of the grip point when moving labware from/ to modules. This PR fixes that and adds a test to check for it.

# Test Plan

Check that the gripping point of the same labware is same no matter if it's being moved to/ from:
- [ ] a slot 
- [ ] a module with stacking overlap
- [ ] a module without stacking overlap
- [ ] a module with adapter
- [ ] a module without adapter.

# Changelog

- updated the grip point calculation to fetch the labware origin coordinates from the existing geometry getter (instead of duplicating the logic, which I was wrongly doing before)
- updated tests

# Review requests

- make sure I'm not missing anything else
- @thassyopinto has tested on a mag block and confirmed that it fixed the issue being seen before. Testing on thermocycler was pending (did it complete yet?) 

# Risk assessment

None if done correctly.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -136,7 +136,10 @@ class GeometryView:
     def _get_labware_position_offset(
         self, labware_id: str, labware_location: LabwareLocation
     ) -> LabwareOffsetVector:
-        """Gets the offset vector of a labware on the given location."""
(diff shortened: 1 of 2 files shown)


sanni-t:
Creating a new PR based on 7.0.0

PR #13363, opened 2023-08-22
fix(app-shell-odd, discovery-client): add option for MDNS in discovery client for use in odd

b-cooper:
# Overview

Because the ODD app shell is only interacting with the local robot, allow it to configure it's discovery client instance to ignore the MDNS poll.

# Changelog

- add `enableMDNS` option to discovery client (defaults to true)
- pass `false` to `enableMDNS` option when app-shell-odd constructs its discovery client instance

# Review requests

- confirm desktop app behaves as expected when discovery robots
- confirm ODD behaves as expected by finding robot at localhost and not polling MDNS for others 

# Risk assessment
med

### DIFF
--- a/app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
+++ b/app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
@@ -339,7 +339,8 @@ function PrepareToRun({
   const { data: attachedInstruments } = useInstrumentsQuery()
   const protocolName =
     protocolRecord?.data.metadata.protocolName ??
-    protocolRecord?.data.files[0].name ?? ''
+    protocolRecord?.data.files[0].name ??
+    ''
   const mostRecentAnalysis = useMostRecentCompletedAnalysis(runId)
   const { launchLPC, LPCWizard } = useLaunchLPC(runId)
   const { setODDMaintenanceFlowInProgress } = useMaintenanceRunTakeover()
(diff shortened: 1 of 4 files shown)


discovery-client/src/types.ts
```diff
@@ -169,6 +169,8 @@ export interface DiscoveryClientOptions {
   onListChange: (robots: DiscoveryClientRobot[]) => unknown
   /** Optional logger */
   logger?: Logger
+  /** Optional logger */
```
vegano1: copy/paste comment

vegano1 (approved):
perfect! nice and easy
good to go once the linter passes.

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -339,7 +339,8 @@ function PrepareToRun({
   const { data: attachedInstruments } = useInstrumentsQuery()
   const protocolName =
     protocolRecord?.data.metadata.protocolName ??
-    protocolRecord?.data.files[0].name
+    protocolRecord?.data.files[0].name ??
+    ''
```
shlokamin, line 343: i think this happened because `protocolRecord.data` wasn't yet populated in the query cache. i bet if we fold checking for that into the `isLoading` boolean we won't have to do this

shlokamin (approved):
lgtm

PR #13364, opened 2023-08-22
fix(engine): fix calculation for labware height on modules

sanni-t:
# Overview

I had forgotten to account for stacking overlap during calculation of the grip point when moving labware from/ to modules. This PR fixes that and adds a test to check for it.

# Test Plan

Check that the gripping point of the same labware is same no matter if it's being moved to/ from:
- [ ] a slot 
- [ ] a module with stacking overlap
- [ ] a module without stacking overlap
- [ ] a module with adapter
- [ ] a module without adapter.

# Changelog

- updated the grip point calculation to fetch the labware origin coordinates from the existing geometry getter (instead of duplicating the logic, which I was wrongly doing before)
- updated tests

# Review requests

- make sure I'm not missing anything else
- @thassyopinto has tested on a mag block and confirmed that it fixed the issue being seen before. Testing on thermocycler was pending (did it complete yet?) 

# Risk assessment

None if done correctly.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -439,10 +439,13 @@ class GeometryView:
         )
         location_slot: DeckSlotName
 
-        if isinstance(location, (ModuleLocation, OnLabwareLocation)):
+        if isinstance(location, DeckSlotLocation):
+            location_slot = location.slotName
+            offset = LabwareOffsetVector(x=0, y=0, z=0)
+        else:
             if isinstance(location, ModuleLocation):
                 location_slot = self._modules.get_location(location.moduleId).slotName
-            else:
+            else:  # OnLabwareLocation
                 location_slot = self.get_ancestor_slot_name(location.labwareId)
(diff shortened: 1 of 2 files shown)


thassyopinto (approved):

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -136,7 +136,10 @@ def get_labware_parent_nominal_position(self, labware_id: str) -> Point:
     def _get_labware_position_offset(
         self, labware_id: str, labware_location: LabwareLocation
     ) -> LabwareOffsetVector:
-        """Gets the offset vector of a labware on the given location."""
+        """Gets the offset vector of a labware on the given location.
+
+        NOTE: Not to be confused with LPC offset.
```
SyntaxColoring, line 141: If this isn't the LPC offset, could we spell out what this does and does not include here?

SyntaxColoring (approved):
LGTM to merge, given the test approval. Thanks!

I am confused by the code, though. One docstring request that might help.

PR #13366, opened 2023-08-23
feat(shared-data): Support new liquid shape for the update script

Laura-Danielle:
# Overview

Forgot to fix the script to update pipette configurations so that the new shape of the liquid values is supported. 

# Risk assessment

None. Internal script.


sfoster1 (approved):
thanks!

PR #13303, opened 2023-08-14
fix(app): add loading state to protocol routes

mjhuff:
closes RQA-1235 and RAUT-466

# Overview

This PR addresses loading state skeletons/fixes/refactoring to the protocol routes.

### ProtocolDetails Route
Adds new loading skeleton to ProtocolDetails route. Adds a spinner to the Start Setup when clicked.

**New Skeleton**

protocolDetailsSkeleton
**Make Setup Spinner**
makeSetupSpinner
### ProtocolSetup
Fixes jitter/styling errors with the ProtocolSetup route skeleton to better match existing page components and adds missing testing to the skeleton. Fixes ProtocolSetupStep button sizing issues when enabled/disabled.
**Updated Skeleton**
updated skeleton
**Consistent Button Sizing**
sizing

# Test Plan

ProtocolDetails Skeleton
1. Upload a new protocol and click it (ProtocolDetails route).
2. If loading occurs, there should be a skeleton, and the Make Setup button should be disabled.
3. After loading completes, the skeleton should disappear and the make setup button should be selectable.
4. Click Make Setup - there should be a spinner until the redirect occurs.

ProtocolSetup Skeleton
1. If any loading occurs on this page, there should be a skeleton. No jittering should occur. 
2. Disabled setup buttons should be the same size as enabled buttons.

# Changelog

- Created ProtocolDetailsSkeleton and testing.
- ProtocolSetupSkeleton refactored to own file and testing added.
- Refactored props within ProtocolDetails to interfaces to promote consistency.
- Added loading spinner on Make Setup button.
- Added a transparent icon in ProtocolSetup for the ProtocolSetupStep buttons.

# Review requests

# Risk assessment

PR #13367, opened 2023-08-23
fix(app): Fix white screen on RobotDashboard when clicking protocol

mjhuff:
closes RQA-1235

# Overview

Addresses white screening and non-responsive clicks when selecting a protocol from the "Run Again" menu on RobotDashboard. 

While PR #13303 added loading state when the /runs/:runId/setup route was hit, the request logic for accessing this screen when clicking on a recently run protocol is different. There are two issues: 

1. A recently run protocol requires that the previous run record be fetched before routing to the ProtocolSetup page. This period takes 1-2 seconds usually, and the lack of user feedback requires addressing.
2. There is a bug in the routing where a non-existent route is hit before the ProtocolSetup route is hit. This causes the temporary white screen.

Although it seems superfluous to have the onSuccessCallback direct to the same route as useCreateRunMutation, there are bizzare rendering issues prior to the second redirect if it's not included. The route is hit twice, but this seems necessary without further investigation.

### Fix

https://github.com/Opentrons/opentrons/assets/64858653/925f0ee9-5d35-471b-8114-27f9b0288a49

# Test Plan

- Verified fix on physical ODD testing several different protocols. 

# Changelog

- Added a pressed state to RecentRunProtocolCard.
- Corrected callback route to the correct route.

# Risk assessment

low

### DIFF
--- a/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
@@ -171,7 +171,7 @@ export function ProtocolWithLastRun({
             aria-label="icon_ot-spinner"
             spin={true}
             size="2.5rem"
-            color={COLORS.darkBlackEnabled}
+            color={COLORS.darkBlack100}
           />
         )}
       </Flex>
--- a/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
@@ -6,11 +6,13 @@ import { formatDistance } from 'date-fns'
 
 import {
   Flex,
+  Icon,
   COLORS,
   SPACING,
   TYPOGRAPHY,
   DIRECTION_COLUMN,
   BORDERS,
+  JUSTIFY_SPACE_BETWEEN,
 } from '@opentrons/components'
 import { useProtocolQuery } from '@opentrons/react-api-client'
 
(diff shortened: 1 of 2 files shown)


app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
```diff
@@ -144,13 +158,22 @@ export function ProtocolWithLastRun({
...
           background={false}
           text={i18n.format(chipText, 'capitalize')}
         />
+        {showSpinner && (
+          <Icon
+            name="ot-spinner"
+            aria-label="icon_ot-spinner"
+            spin={true}
+            size="2.5rem"
+            color={COLORS.darkBlackEnabled}
```
koji: darkBlackEnabled and darkBlack100 are the same but this component is for ODD so let's use darkBlack100

```suggestion
            color={COLORS.darkBlack100}
```

app/src/organisms/OnDeviceDisplay/RobotDashboard/__tests__/RecentRunProtocolCard.test.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from 'react'
+import { COLORS, renderWithProviders } from '@opentrons/components'
 import { fireEvent } from '@testing-library/react'
 import { formatDistance } from 'date-fns'
 import { when, resetAllWhenMocks } from 'jest-when'
 import { MemoryRouter } from 'react-router-dom'
 
 import { RUN_STATUS_FAILED } from '@opentrons/api-client'
-import { renderWithProviders } from '@opentrons/components'
+
```
koji: let's move back the following line to line8
import { COLORS, renderWithProviders } from '@opentrons/components'

sorry we haven't talked about this to you.
https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3265364497/Code+Cook+Book+for+App#import-statement

koji (approved):
left a couple of comments, but the changes worked as expected.
this is great!

app/src/organisms/OnDeviceDisplay/RobotDashboard/__tests__/RecentRunProtocolCard.test.tsx
mjhuff: Thank you for linking this! I didn't know it existed. Huge help.

PR #13370, opened 2023-08-23
fix(app): find previous labware location from all commands for run preview text

b-cooper:
# Overview

In the absence of a way to request timeline-based location information from the protocol engine, search through all previous commands to reduce labware location by duplicating PE concerns on the front end. 

Note: This is a brittle duplication of information that is already managed (but not exposed) by the Protocol Engine.

re RSS-301

Before: 
Screen Shot 2023-08-23 at 12 17 57 PM

After:
Screen Shot 2023-08-23 at 12 18 16 PM

# Review requests

- Run Preview on Desktop and Running Protocol screen on ODD should render the correct "old location"
 when describing a `moveLabware` command. 
  
# Risk assessment
low


shlokamin (approved):
we reviewed this together yesterday but forgot to approve. did not test on robot but we tested together!

PR #13371, opened 2023-08-23
fix(app): labware setup renders and location fixes

jerader:
closes RQA-1385 closes RQA-1383

# Overview

PR fixes 2 bugs found in alpha 4:
1. the location of a labware on an adapter on a module
2. labware rendered on top of a module for protocols that don't specify the `displayName` of the nested labware

# Test Plan

Test on the protocol mentioned in the ticket. Look at the labware list view and the labware map view on the desktop/odd, should should correct information now.

# Changelog

- change `labwareListView` to match the moduleId from` results` instead of `params`. TBH idk how i missed this, must be that PD protocols render those `moduleId`s as the same? 
- don't assume every labware has a `displayName` - so i changed `SetupLabwareMap` to not have that extra check
- fix test
- fix `LabwareInfoOverlay` to account for the offset

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/LabwareInfoOverlay.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/LabwareInfoOverlay.tsx
@@ -104,8 +104,8 @@ export const LabwareInfoOverlay = (
   const height = definition.dimensions.yDimension
   return (
     <RobotCoordsForeignDiv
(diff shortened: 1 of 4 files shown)


mjhuff (approved):
Tested and works! lgtm 🚀

PR #13372, opened 2023-08-23
refactor(shared-data): update plunger motor configs

caila-marashaj:
Due to mechanical changes, the PVT single channel pipette needs to run at a lower current than other pipette models. 

Additionally, after some period without use, all pipette plungers, except the 96 channels, will need to home using 1A motor current to overcome their static friction.

This branch is just these couple of changes, however I'd like a quick sanity check before merging to make sure that speed won't be an issue during homing. It looks like homing for the first time after attaching will happen using the `_build_home_pipettes_runner` function, in which case the speed chosen _should_ be 5 mm/sec.


vegano1:
Nice!
Let's retarget this for the 7.0.0 release branch

shared-data/python/opentrons_shared_data/pipette/dev_types.py
```diff
@@ -155,6 +155,7 @@ class PipetteModelSpec(TypedDict, total=False):
     pickUpPresses: PipetteCustomizableConfigElementInt
     pickUpSpeed: PipetteCustomizableConfigElementFloat
     plungerCurrent: PipetteCustomizableConfigElementFloat
+    plungerHomingCurrent: float
```
vegano1: should probably make this the same as `PipetteCustomizableConfigElementFloat`

shared-data/pipette/definitions/2/general/eight_channel/p1000/1_0.json
```diff
@@ -1,6 +1,6 @@
 {
   "$otSharedSchema": "#/pipette/schemas/2/pipettePropertiesSchema.json",
-  "displayName": "Flex 8-Channel 1000 μL",
+  "displayName": "Flex 8-Channel 1000 \u03bcL",
```
vegano1: Not sure how this was generated but should be μL

shared-data/python/opentrons_shared_data/pipette/scripts/build_json_script.py
```diff
@@ -97,6 +97,7 @@ def _build_motor_configurations(
     if model_configurations:
         run = model_configurations["plungerCurrent"]["value"]
         idle = model_configurations.get("idleCurrent", run)
+        homing = model_configurations["plungerHomingCurrent"]["value"]
```
Laura-Danielle: In this instance, you wouldn't use the `build_json_script.py` but the `update_configuration_files.py` when you're updating files. (Just for future reference).

shared-data/pipette/definitions/2/general/single_channel/p50/3_3.json
```diff
@@ -1,6 +1,6 @@
 {
   "$otSharedSchema": "#/pipette/schemas/2/pipettePropertiesSchema.json",
-  "displayName": "Flex 1-Channel 50 μL",
+  "displayName": "Flex 1-Channel 50 \u03bcL",
```
Laura-Danielle: This was probably accidentally generated by the script that you used. We should make sure to fix the areas that this happened.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1295,6 +1295,16 @@ async def _retrieve_home_position() -> Tuple[
             await self._backend.home([axis], self.gantry_load)
         else:
             # both stepper and encoder positions are invalid, must home
+            if axis in Axis.pipette_axes():
+                # if we're homing a pipette for the first time, should do it
+                # at the homing current
+                mount = Axis.to_ot3_mount(axis)
+                instr = self._pipette_handler.hardware_instruments[mount]
+
+                if instr is not None:
```
Laura-Danielle: `if instr` is cleaner.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1295,6 +1295,16 @@ async def _retrieve_home_position() -> Tuple[
...
+                    await self._backend.set_active_current(
```
Laura-Danielle:
```suggestion
async with self._backend.restore_current():
                    await self._backend.set_active_current(
```

(The home will also need to be wrapped in this)

Laura-Danielle (changes requested):
1. Gotta update the json schema to reflect the new change to plunger positions
2. Don't forget to make sure we remove the auto-generated junk that replaced `uL` in some of the pipette definitions.

shared-data/pipette/definitions/2/general/eight_channel/p1000/3_0.json
```diff
@@ -1,6 +1,6 @@
 {
   "$otSharedSchema": "#/pipette/schemas/2/pipettePropertiesSchema.json",
-  "displayName": "Flex 8-Channel 1000 μL",
+  "displayName": "Flex 8-Channel 1000 \u03bcL",
```
sfoster1: whoa buddy

sfoster1 (changes requested):
agree with laura's comments, need to fix that mu. probably can do it with find and replace in your editor.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1295,6 +1295,16 @@ async def _retrieve_home_position() -> Tuple[
             await self._backend.home([axis], self.gantry_load)
         else:
```
andySigler: `homing` current should be used for all homing actions, not just the initial homing motion. This is because the static friction issue can happen regardless of if the pipette is powered on or off, it just needs to be still for the problem to happen.

Because non-initial homing motions include that faster move to get close to the endstop, we'll need a per-pipette homing-speed that is used as a motion constraint for the plunger axis during this faster movement.

The currently used default max-speed for 1ch/8ch plungers is `70` while 96ch is `15`. So for example, during homing we can instead constrain these to be `30` for 1ch/8ch and `5` for 96ch

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_5.json
```diff
@@ -17,7 +17,8 @@
   },
   "plungerMotorConfigurations": {
     "idle": 0.3,
-    "run": 0.8
+    "run": 0.8,
+    "homing": 0.8
```
andySigler: let's set this to `2.2` then test on a few 96ch to see if it needs changing. Ideally we can use the motor's max rated current so we can get the most torque out of it to overcome any static-friction that may have built up.

Same goes for the `v3_4.json` 96ch

shared-data/pipette/definitions/2/general/single_channel/p10/1_0.json
```diff
@@ -17,7 +17,11 @@
     "increment": 0.0,
     "distance": 0.0
   },
-  "plungerMotorConfigurations": { "idle": 0.05, "run": 0.3 },
+  "plungerMotorConfigurations": {
+    "idle": 0.05,
+    "run": 0.3,
+    "homing": 1
```
andySigler: I don't remember what the OT2 does for homing, but ideally this PR doesn't affect it at all and the configurations don't describe values that aren't accurate.

andySigler (changes requested):

api/src/opentrons/hardware_control/ot3api.py
andySigler: Oh and we should also be setting the pipette's `run` current during all other plunger motions, which right now does not include the movement that happens inside `OT3API._move_to_plunger_bottom()` but it should.

PR #13373, opened 2023-08-23
fix(app): ensure current run id is one created before issuing commands

smb2268:
fix RLIQ-1377

# Overview
We are often seeing a few second mismatch between the returned `currentMaintenanceRun` id and the maintenance run that is created within a flow. This is causing a number of issues including the flow closing when the previous run id gets deleted and attempting to issue commands to a non-existent maintenance run.

# Test Plan
Try to launch a pipette flow and a gripper flow. See that the flow doesn't close right away and that no commands get a 404 error

# Changelog
1. Ensure `currentMaintenanceRunId` is equal to the resulting id of the `createMaintenanceRun` mutation before proceeding to issue any commands

# Review requests
Run through test plan

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/GripperWizardFlows/index.tsx
+++ b/app/src/organisms/GripperWizardFlows/index.tsx
@@ -90,12 +90,14 @@ export function GripperWizardFlows(
   // this will close the modal in case the run was deleted by the terminate
   // activity modal on the ODD
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


b-cooper (approved):
🦺

TamarZanzouri (approved):
Nice! thank you for this!

PR #13377, opened 2023-08-24
feat(app, component): make ODD components' screen size onDeviceDisplay size

koji:
# Overview
make ODD components' screen size onDeviceDisplay size
@jerader added viewport for touchscreen app and we can switch the screen size manually but once we change it, we will need to change the option to switch the screen size manually.

The entire Storybook layout will be update in the future.

**Estop modals are broken right now. I will figure out the issue and fix in the future.**

https://github.com/Opentrons/opentrons/assets/474225/dc3d3168-ee64-4188-9d11-bc01878b7a9a

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/atoms/Chip/Chip.stories.tsx
+++ b/app/src/atoms/Chip/Chip.stories.tsx
@@ -1,18 +1,13 @@
 import * as React from 'react'
 import { Flex, COLORS, SPACING } from '@opentrons/components'
-import { customViewports } from '../../../../.storybook/preview'
+import { touchScreenViewport } from '../../DesignTokens/constants'
 import { Chip } from '.'
 import type { Story, Meta } from '@storybook/react'
 
 export default {
   title: 'ODD/Atoms/Chip',
   component: Chip,
-  parameters: {
-    viewport: {
-      viewports: customViewports,
-      defaultViewport: 'onDeviceDisplay',
-    },
-  },
+  parameters: touchScreenViewport,
 } as Meta
 
 interface ChipStorybookProps extends React.ComponentProps<typeof Chip> {
--- a/app/src/atoms/Chip/Chip.stories.tsx
+++ b/app/src/atoms/Chip/Chip.stories.tsx
@@ -1,11 +1,18 @@
 import * as React from 'react'
 import { Flex, COLORS, SPACING } from '@opentrons/components'
+import { customViewports } from '../../../../.storybook/preview'
 import { Chip } from '.'
 import type { Story, Meta } from '@storybook/react'
 
 export default {
   title: 'ODD/Atoms/Chip',
   component: Chip,
+  parameters: {
+    viewport: {
+      viewports: customViewports,
+      defaultViewport: 'onDeviceDisplay',
+    },
+  },
 } as Meta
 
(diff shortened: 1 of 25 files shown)


koji:
- [ ] fix estop modals' stories

jerader:
@koji , the reason why you changed it to specify the viewport  directly in each story is because we are planning to remove the `touchscreen app` viewport option? It would be great if you can make a ticket that outlines why we are planning to remove the `touchscreen app` viewport, I was not aware 🤔

koji:
> @koji , the reason why you changed it to specify the viewport directly in each story is because we are planning to remove the `touchscreen app` viewport option? It would be great if you can make a ticket that outlines why we are planning to remove the `touchscreen app` viewport, I was not aware 🤔

@jerader No, we won't remove the viewport.
I added the parameter to display touchscreen size directly since a few of us didn't know touchscreen viewport existed in our Storybook. (Actually, Brian didn't know) I guess the design team also doesn't know since most folks don't use Storybook regularly.

This PR is to set touch screen size as default for the odd components.

jerader:
> @jerader No, we won't remove the viewport. I added the parameter to display touchscreen size directly since a few of us didn't know touchscreen viewport existed in our Storybook. (Actually, Brian didn't know) I guess the design team also doesn't know since most folks don't use Storybook regularly.
> 
> This PR is to set touch screen size as default for the odd components.

@koji got it, so this is for easier use for designers/devs who are unaware of the viewport selections. Thank you for clarifying!

app/src/atoms/Chip/Chip.stories.tsx
```diff
@@ -1,11 +1,18 @@
 import * as React from 'react'
 import { Flex, COLORS, SPACING } from '@opentrons/components'
+import { customViewports } from '../../../../.storybook/preview'
```
jerader: I wonder if there is a better place to store this constant. But since we still need it in `preview.js` maybe it is okay as is.

app/src/atoms/Chip/Chip.stories.tsx
```diff
@@ -1,11 +1,18 @@
...
 
 export default {
   title: 'ODD/Atoms/Chip',
   component: Chip,
+  parameters: {
+    viewport: {
+      viewports: customViewports,
+      defaultViewport: 'onDeviceDisplay',
+    },
+  },
```
jerader: maybe we can leave `customViewports` where it is and instead make this a constant that lives in `app/src/` somewhere

jerader (approved):
nice! I have a proposal for cleaning this up a bit more. 

We can make 
`parameters: {
    viewport: {
      viewports: customViewports,
      defaultViewport: 'onDeviceDisplay',
    },
  }`
a constant that lives in `app/src/` somewhere and then import that since its a bunch of repeating lines. And that way `customViewport` const that is being imported all the way from `storybook/preview.js` would only be imported once for the new const.

app/src/atoms/Chip/Chip.stories.tsx
koji: yeah, making a const for `viewport` makes sense.

koji:
- [x] create a const

PR #13381, opened 2023-08-24
refactor(api): split backend tip_action into two functions

caila-marashaj:
## Overview
Currently, the backend `tip_action` function is used both for regular tip action moves, as well as for homing the gear motors at a constant velocity. This makes the code a little bit cluttered, as the backend function takes in either a list of moves or a distance and velocity to accommodate this, and then needs to determine which one it's received. 

Instead, the api can just use two separate functions- one for homing, and one for all other moves involving the gear motor.

## Changelog
- create a `home_gear_motors` function in `ot3api.py` that fast homes the gear motors if possible, or homes at constant velocity if not
- create a `home_gear_motors` function in `ot3controller.py` that is used only for homing at constant velocity
- `ot3controller.py::tip_action` will only take in a list of moves

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -40,7 +40,7 @@ from .ot3utils import (
     create_gripper_jaw_home_group,
     create_gripper_jaw_hold_group,
     create_tip_action_group,
-    create_gear_motor_home_group,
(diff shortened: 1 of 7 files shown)


Laura-Danielle (approved):
Make sure this is tested on a robot beforehand, but looks good overall.

vegano1 (approved):
Makes sense,
go to go once we get some testing on the robot.

PR #13382, opened 2023-08-24
feat(shared-data): Increase p1000 single dispense flow rates

andySigler:
# Overview

Gives us both better performance and faster protocol runs, win win.

Updated flow-rates are based on testing with PVT units at the factory. The update was also applied to the DVT definitions, because they are still mostly used by testing in BK.

# Test Plan

# Changelog

# Review requests

# Risk assessment


Laura-Danielle (approved):
lgtm!

ryanthecoder (approved):
merge it

PR #13383, opened 2023-08-24
docs(api): refactor-tip-handling

jwwojak:
# Overview

This PR is for managing changes made to the "Manipulating Pipette Tips" section of the doc, [Building Block Commands](https://docs.opentrons.com/v2/new_atomic_commands.html). Following the plan laid out in [this Figma file](https://www.figma.com/file/WsKj7gklIgDx68DkzegdaN/Commands-PAPI-docs-merge-strategies?type=whiteboard&node-id=0%3A1&t=souDTlzeZ3A0tmOj-1). This PR is for the first, small branch section in Merge Strategy 1.

The manipulating tips section has 4 parts:

- Picking Up a Tip
- Dropping a Tip
- Returning a Tip
- Iterating Through Tips

# Test Plan

-  Check code in the Black formatter and use that style.
- Make sure links work.
- The usual grammar, style, clarity stuff.

# Changelog

This refactoring introduces significant changes to the text used in the manipulating tips section. This includes things like:

- new code samples
- new and revised text
- some shortcuts/anchors
- fix section headers to use proper style

# Review requests

The usual suspects.

# Risk assessment

Low risk. Docs changes only.

### DIFF
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -218,18 +218,21 @@ Working With Used Tips
 
 API versions 2.2 or higher consider tips as "used" after they've been picked up, even if you return a tip to the tip rack. For example, if you pick up a tip from rack location A1 and then return it to the same location, the robot will not attempt to pick up this tip again, unless explicitly specified. Instead, the robot will pick up a tip starting from rack location B1. For example::
 
-    pipette.pick_up_tip()  # picks up tip from rack location A1
-    pipette.return_tip()   # drops tip in rack location A1
-    pipette.pick_up_tip()  # picks up from rack location B1
+    pipette.pick_up_tip()                # picks up tip from rack location A1
+    pipette.return_tip()                 # drops tip in rack location A1
+    pipette.pick_up_tip(tiprack_1['A1']) # picks up tip from rack location A1
+    pipette.drop_tip()                   # drops tip in trash bin
+    pipette.pick_up_tip()                # picks up tip from rack location B1
 
 .. can this be removed?
 Also in API Version 2.2, the return tip height was corrected to utilize values determined by hardware testing. This is more in-line with return tip behavior from Python Protocol API Version 1.
 
-API versions 2.0 and 2.1 treated returned tips as unused items and they could be picked up again. For example:: 
+API versions 2.0 and 2.1 treated returned tips as unused items and they could be picked up again without an explicit argument. For example:: 
 
     pipette.pick_up_tip()  # picks up tip from rack location A1
     pipette.return_tip()   # drops tip in rack location A1
     pipette.pick_up_tip()  # picks up tip from rack location A1
+
 
 ****************
 Liquid Control
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -149,13 +149,14 @@ If you omit the ``tip_rack`` argument from the ``pipette`` variable, then you mu
 
 However, coding the location of each tip is inefficient. Let's use a ``for`` loop to automate a sequential tip pick up process.
 
+Coding the location of each tip is inefficient. Instead, let the API track and manage the tip pickup location for you. Adding a ``for`` loop to  your protocol is a great way to take advantage of API tip tracking.
+
 .. versionadded:: 2.0
 
-.. needs better title?
 Loops and Tip Pick Up
 ---------------------
 
-A ``for`` loop and Python's :py:class:`range` class gives you a better way to automate the tip pickup process. It eliminates the need to call ``pick_up_tip()`` multiple times. For example, this snippet tells the robot to sequentially use all the tips in a 96-tip rack::
+A ``for`` loop and Python's :py:class:`range` class brings automation to the tip pickup and tracking process. It eliminates the need to call ``pick_up_tip()`` multiple times. For example, this snippet tells the robot to sequentially use all the tips in a 96-tip rack::
 
     for i in range(96):
         pipette.pick_up_tip()
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -32,29 +32,35 @@ The examples in this section would be added to the following:
 
 This loads a `Corning 96 Well Plate <https://labware.opentrons.com/corning_96_wellplate_360ul_flat>`_ in slot 2 and a `Opentrons 300 µL Tiprack <https://labware.opentrons.com/opentrons_96_tiprack_300ul>`_ in slot 3, and uses a P300 Single GEN2 pipette.
 
+.. _atomic-tip-manipulation:
 
-************
-Tip Handling
-************
+Manipulating Pipette Tips
+=========================
 
-By default, the robot constantly exchanges old, used tips for new ones to prevent cross-contamination between wells. Tip handling uses the functions :py:meth:`.InstrumentContext.pick_up_tip`, :py:meth:`.InstrumentContext.drop_tip`, and :py:meth:`.InstrumentContext.return_tip`.
+The API provides three basic functions that help the robot manipulate pipette tips during a protocol run. These are :py:meth:`.InstrumentContext.pick_up_tip`, :py:meth:`.InstrumentContext.drop_tip`, and :py:meth:`.InstrumentContext.return_tip`. 
+Respectively, they tell the robot to pick up a tip from a tip rack, drop a tip into the trash (or another location), and return a tip to its location in the tip rack. Let's look at examples for each of these functions.
 
-Pick Up Tip
-===========
+Picking Up a Tip
+----------------
 
-Before any liquid handling can be done, your pipette must have a tip on it. The command :py:meth:`.InstrumentContext.pick_up_tip` will move the pipette over to the specified tip, then press down into it to create a vacuum seal. The below example picks up the tip at location ``'A1'`` of the tiprack previously loaded in slot 3.
-
-.. code-block:: python
-
-   pipette.pick_up_tip(tiprack['A1'])
-
-If you have associated a tiprack with your pipette such as in the :ref:`new-pipette` or :ref:`protocol_api-protocols-and-instruments` sections, then you can simply call
-
-.. code-block:: python
-
+Your robot needs to attach a disposable tip to the pipette before it can aspirate or dispense any liquids. Given sample code above, call the :py:meth:`~.InstrumentContext.pick_up_tip` method without any arguments::
+    
     pipette.pick_up_tip()
 
-This will use the next available tip from the list of tipracks passed in to the ``tip_racks`` argument of :py:meth:`.ProtocolContext.load_instrument`.
+This works because the variable ``tiprack_1`` includes the on-deck location of the tip rack (Flex ``location="D3"``, OT-2 ``location=3``). Next, notice how the ``pipette`` variable includes the argument ``tip_racks=[tiprack_1]``. Given this information, the robot automatically knows where to go and what to do. For example:
+
+- Flex moves the pipette to the tip rack in deck slot D3 and picks up a tip from rack location A1.
+- OT-2 moves the pipette to the tip rack in deck slot 3 and picks up a tip from rack location A1.
+
+On subsequent calls to ``pick_up_tip()``, the robot will use the next available tip in the rack. For example::
+
+    pipette.pick_up_tip()  # picks up tip in rack location A1
+    pipette.drop_tip()     # drops tip in trash bin
+    pipette.pick_up_tip()  # picks up tip in rack location B1 
+
+If you omit the ``tip_rack`` argument from the :py:meth:`~.InstrumentContext.load_instrument` method, then you must pass in the tip rack's location information to ``pick_up_tip`` like this::
+    
+    pipette.pick_up_tip(tiprack_1['A1']) 
 
 .. versionadded:: 2.0
 
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -159,9 +159,12 @@ A ``for`` loop and Python's :py:class:`range` class gives you a better way to au
 
     for i in range(96):
         pipette.pick_up_tip()
+        # liquid handling commands
         pipette.drop_tip()
 
-If your protocol requires a lot of tips, add a second tip rack to the protocol, and sum the tip count in the range. The robot will loop through both racks. For example, let's add another tip rack to the sample protocol::
+If your protocol requires a lot of tips, add a second tip rack to the protocol. Then associate it with your pipette and increase the number of repetitions in the loop. The robot will loop through both racks. 
+
+First, add another tip rack to the sample protocol::
 
     tiprack_2 = protocol.load_labware(
         load_name="opentrons_flex_96_tiprack_1000ul",
@@ -173,7 +176,7 @@ Next, revise the pipette's ``load_instrument()`` method to include the new tip r
     pipette = protocol.load_instrument(
         instrument_name="flex_1channel_1000",
         mount="left",
-        tip_racks=[tiprack_1, tip_rack_2],
+        tip_racks=[tiprack_1, tiprack_2],
     ) 
 
 Finally, sum the tip count in the range::
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -178,7 +178,7 @@ Set the new range in the ``for`` loop to use all the tips in each tip rack::
         pipette.pick_up_tip()
         pipette.drop_tip()
 
-For a more advanced example using a loop, take a moment to review the :ref:`off-deck location protocol <off-deck-location>` on the :ref:`_moving-labware` page. This example uses a ``for`` loop to iterate through a tip rack, but also includes other commands that pause the protocol and let you replace an on-deck tip rack with another rack stored in an off-deck location.
+For a more advanced "real-world" example, take a moment to review the :ref:`off-deck location protocol <off-deck-location>` on the :ref:`_moving-labware` page. This example uses a ``for`` loop to iterate through a tip rack, but also includes other commands that pause the protocol and let you replace an on-deck tip rack with another rack stored in an off-deck location.
 
 Dropping a Tip
 --------------
@@ -190,7 +190,7 @@ To drop a tip in the trash bin, call the :py:meth:`~.InstrumentContext.drop_tip`
 You can also specify where to drop the tip by passing in a location. For example, this code drops a tip in the trash bin and drops another tip in its original tip rack location::
 
     pipette.pick_up_tip()            # picks up tip from rack location A1
-    pipette.drop_tip                 # drops tip in trash bin 
+    pipette.drop_tip()               # drops tip in trash bin 
     pipette.pick_up_tip()            # picks up tip from rack location B1
     pipette.drop_tip(tiprack['B1'])  # drops tip in rack location B1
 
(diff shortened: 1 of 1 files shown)


api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-===========
-
-Before any liquid handling can be done, your pipette must have a tip on it. The command :py:meth:`.InstrumentContext.pick_up_tip` will move the pipette over to the specified tip, then press down into it to create a vacuum seal. The below example picks up the tip at location ``'A1'`` of the tiprack previously loaded in slot 3.
+Manipulating Pipette Tips
+=========================
 
-.. code-block:: python
-
-   pipette.pick_up_tip(tiprack['A1'])
+Your robot needs to attach a disposable tip to the pipette before it can aspirate or dispense liquids. The API provides three basic functions that help the robot attach and manage pipette tips during a protocol run. These methods are :py:meth:`.InstrumentContext.pick_up_tip`, :py:meth:`.InstrumentContext.drop_tip`, and :py:meth:`.InstrumentContext.return_tip`. Respectively, these methods tell the robot to pick up a tip from a tip rack, drop a tip into the trash (or another location), and return a tip to its location in the tip rack. Let's look at examples for each of these functions. Each code snippet works with the sample code shown above.
```
ecormany: 👍 
you can link to 
```
:ref:`protocol-template`
```
even though it'll just show up as plain text for now.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-    pipette.drop_tip(tiprack['A1'])  # drop back in A1 of the tiprack
-    pipette.pick_up_tip()
-    pipette.drop_tip()  # drop in the fixed trash on the deck
+If you omit the ``tip_rack`` argument from the ``pipette`` variable, then you must pass in the tip rack's location to ``pick_up_tip`` like this::
+    
+    pipette.pick_up_tip(tiprack_1['A1'])
+    pipette.drop_tip()
+    pipette.pick_up_tip(tiprack_1['B1'])
 
+However, coding the location of each tip is inefficient. Let's use a ``for`` loop to automate a sequential tip pick up process.
```
ecormany: I would mention here that if you call `pick_up_tip()` without associating tip racks with the pipette, the API will raise an error.

Then this sentence could go into the next section, and be reframed as 
```
Coding the location of each tip is inefficient. It's usually best to let the API 
track pickup location for you. 
```
_next subsection_
```
For example, let's use a ``for`` loop to automate picking up and dropping 
a large number of tips from a rack.
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-
-.. code-block:: python
-
-    tip_rack = protocol.load_labware(
-            'opentrons_96_tiprack_300ul', 1)
-    pipette = protocol.load_instrument(
-        'p300_single_gen2', mount='left', tip_racks=[tip_rack])
+    for i in range(96):
+        pipette.pick_up_tip()
+        pipette.drop_tip()
```
ecormany, line 50: could add a comment line here
```suggestion
        # liquid handling commands
        pipette.drop_tip()
```
…to make it not seem like we're just throwing a way a whole tip rack for no reason

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
 
-    pipette.pick_up_tip() # picks up tip_rack:A1
-    pipette.return_tip()
-    pipette.pick_up_tip() # picks up tip_rack:B1
+If your protocol requires a lot of tips, add a second tip rack to the protocol, and sum the tip count in the range. The robot will loop through both racks. For example, let's add another tip rack to the sample protocol::
```
ecormany: "and sum the tip count in the range" feels out of order, since it doesn't come until a couple snippets later. maybe
```suggestion
If your protocol requires a lot of tips, add a second tip rack to the protocol. Then associate it with your pipette and increase the number of repetitions in the loop. The robot will loop through both racks. 

First, add another tip rack to the sample protocol::
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-            'opentrons_96_tiprack_300ul', 4)
-        pipette = protocol.load_instrument(
-            'p300_single_gen2', mount='left', tip_racks=[tip_rack_1, tip_rack_2])
-
-This loads a `Corning 96 Well Plate <https://labware.opentrons.com/corning_96_wellplate_360ul_flat>`_ in slot 2 and two `Opentrons 300ul Tiprack <https://labware.opentrons.com/opentrons_96_tiprack_300ul>`_ in slots 3 and 4 respectively, and uses a P300 Single GEN2 pipette.
-
-When a list of tip racks is associated with a pipette in its ``tip_racks`` argument, the pipette will automatically pick up the next unused tip in the list whenever you call :py:meth:`.InstrumentContext.pick_up_tip`. The pipette will first use all tips in the first tiprack, then move on to the second, and so on:
+        instrument_name="flex_1channel_1000",
+        mount="left",
+        tip_racks=[tiprack_1, tip_rack_2],
```
ecormany:
```suggestion
        tip_racks=[tiprack_1, tiprack_2],
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
     pipette.pick_up_tip()
 
-To change the location of the first tip used by the pipette, you can use :py:obj:`.InstrumentContext.starting_tip`:
+You can also specify where to drop the tip by passing in a location. For example, this code drops a tip in the trash bin and returns another tip to the tip rack::
 
-.. code-block:: python
+    pipette.pick_up_tip()            # picks up tip from rack location A1
+    pipette.drop_tip()               # drops tip in trash bin 
+    pipette.pick_up_tip()            # picks up tip from rack location B1
+    pipette.drop_tip(tiprack['B1'])  # drops tip in rack location B1
```
ecormany: i'm not sure why you would use `drop_tip` to put a tip back where it came from, instead of `return_tip`. maybe change to `A1` here to show that it's possible to move a tip to a _different_ well of the rack.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-         pipette.return_tip()
+To return a tip to its original location, call the :py:meth:`~.InstrumentContext.return_tip` method without any arguments::
 
-    # Reset the tip tracker
-    pipette.reset_tipracks()
+    pipette.return_tip()
 
-    # Picks up a tip from well A1 of the first tip rack
-    pipette.pick_up_tip()
+.. this section below was a long note
```
ecormany:
```
..
   fun fact
   you can make a multi-line RST comment like this

   it's all a comment

now this isn't a comment
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-To check whether you should pick up a tip or not, you can utilize :py:meth:`.InstrumentContext.has_tip`:
+    pipette.pick_up_tip()                # picks up tip from rack location A1
+    pipette.return_tip()                 # drops tip in rack location A1
+    pipette.pick_up_tip()                # picks up tip from rack location B1
+    pipette.drop_tip()                   # drops tip in trash bin
+    pipette.pick_up_tip(tiprack_1['A1']) # picks up tip from rack location A1
 
-.. code-block:: python
+.. can this be removed, is it helpful?
+Also in API Version 2.2, the return tip height was corrected to utilize values determined by hardware testing. This is more in-line with return tip behavior from Python Protocol API Version 1.
```
ecormany: this is ancient history, scrap it imo

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
 
-.. versionadded:: 2.0
+API versions 2.2 or higher consider tips as "used" after being picked up. For example, if the robot picked up a tip from rack location A1 and then returned it to the same location, it will not attempt to pick up this tip again, unless explicitly specified. Instead, the robot will pick up a tip starting from rack location B1. For example::
 
-To check whether you should pick up a tip or not, you can utilize :py:meth:`.InstrumentContext.has_tip`:
+    pipette.pick_up_tip()                # picks up tip from rack location A1
+    pipette.return_tip()                 # drops tip in rack location A1
+    pipette.pick_up_tip()                # picks up tip from rack location B1
+    pipette.drop_tip()                   # drops tip in trash bin
+    pipette.pick_up_tip(tiprack_1['A1']) # picks up tip from rack location A1
```
ecormany, line 111: add a `.. versionchanged: 2.2` here

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
+Also in API Version 2.2, the return tip height was corrected to utilize values determined by hardware testing. This is more in-line with return tip behavior from Python Protocol API Version 1.
 
-    for block in range(3):
-        if block == 0 and not pipette.has_tip:
-            pipette.pick_up_tip()
-        else:
-            m300.mix(mix_repetitions, 250, d)
-            m300.blow_out(s.bottom(10))
-            m300.return_tip()
+API versions 2.0 and 2.1 treated returned tips as unused items. They could be picked up again without an explicit argument. For example::
```
ecormany: A slight reframe possibility here.

At the top of the section, instead of talking about versions, begin `The API considers tips...`. Then include the version changed notice I flagged above. Then put this sentence and the snippet below inside a note.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-
-To return the tip to the original location, you can call :py:meth:`.InstrumentContext.return_tip`. The example below will automatically return the tip to ``'A3'`` on the tip rack.
-
-.. code-block:: python
-
-    pipette.pick_up_tip(tiprack['A3'])
-    pipette.return_tip()
-
-.. note:
+.. needs better title?
```
ecormany: `Automating Pickup`?

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
+Loops and Tip Pick Up
+---------------------
 
-    In API Version 2.0 and 2.1, the returned tips are added back into the tip-tracker and thus treated as `unused`. If you make a subsequent call to `pick_up_tip` then the software will treat returned tips as valid locations.
-    In API Version 2.2, returned tips are no longer added back into the tip tracker. This means that returned tips are no longer valid locations and the pipette will not attempt to pick up tips from these locations.
-    Also in API Version 2.2, the return tip height was corrected to utilize values determined by hardware testing. This is more in-line with return tip behavior from Python Protocol API Version 1.
+A ``for`` loop and Python's :py:class:`range` class gives you a better way to automate the tip pickup process. It eliminates the need to call ``pick_up_tip()`` multiple times. For example, this snippet tells the robot to sequentially use all the tips in a 96-tip rack::
```
ecormany: This makes it sound a little like you should _always_ use loops with `pick_up_tip` or that it's some kind of best practice. A loop works great with `pick_up_tip` when it meets the needs of the protocol. It only suits our needs here because we want to quickly show burning through a whole rack of tips.

I basically don't want to discourage users from calling `pick_up_tip()` sans argument piecemeal throughout their protocol. It's meant for that too.

api/docs/v2/new_atomic_commands.rst
ecormany: there may also be some Py-pedantry here about `function` vs `method`

api/docs/v2/new_atomic_commands.rst
ecormany: it may also be useful to see whether you can drop tips in arbitrary labware. like can i make a bonus trash container out of a 1-well reservoir? i forget what limits the API has there.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
 
-.. code-block:: python
+Return Tip
+===========
 
-    # Use up all tips
-    for _ in range(96+96):
-         pipette.pick_up_tip()
-         pipette.return_tip()
+To return a tip to its original location, call the :py:meth:`~.InstrumentContext.return_tip` method without any arguments::
```
ecormany: maybe worth noting that it _never_ takes any position argument

api/docs/v2/new_atomic_commands.rst
```diff
@@ -116,182 +116,122 @@ Depending on your robot model and/or API version, the foundation of a basic prot
...
-            m300.blow_out(s.bottom(10))
-            m300.return_tip()
+API versions 2.0 and 2.1 treated returned tips as unused items. They could be picked up again without an explicit argument. For example:: 
 
-.. versionadded:: 2.7
+    pipette.pick_up_tip()  # picks up tip from rack location A1
+    pipette.return_tip()   # drops tip in rack location A1
+    pipette.pick_up_tip()  # picks up tip from rack location A1
 
-**********************
```
ecormany, line 120: We may also want a subsection explaining `InstrumentContext.has_tip` — it's useful for avoiding "hey you've already got a tip" errors with `pick_up_tip()`. And we should state somewhere that it's purely based on API-internal state. OT-2 pipettes have no way of knowing whether a tip is attached. And I don't think the Flex sensors have been wired up to `has_tip`. So it really just keeps track of whether the API _thinks_ the pipette _should_ have a tip attached based on previous instructions.

ecormany (commented):
This is cleaning up nicely. Left some comments to smooth out the flow even more. We can also make a couple more strategic exclusions and inclusions so this section is complete without going into the weeds.

api/docs/v2/new_atomic_commands.rst
jwwojak: > i'm not sure why you would use drop_tip to put a tip back where it came from, instead of return_tip. maybe change to A1 here to show that it's possible to move a tip to a different well of the rack.

I changed the text for `drop_tip` to well A1 to demonstrate returning it to a previously used well.

> it may also be useful to see whether you can drop tips in arbitrary labware. like can i make a bonus trash container out of a 1-well reservoir? i forget what limits the API has there.

I tried with an Agilent 1 well reservoir (290 ml) and an Axigen reservoir (90 ml). Validation returns a "Labware . . . has no tip length defined" error. Not sure how to add `tip_length` to the sample code. Will ask, but to keep things going, skipping for now. I get the potential use case though. Maybe you'd want to use a reservoir instead of contaminating the trash bin -- something like that.

api/docs/v2/new_atomic_commands.rst
jwwojak: Thank you. I wasn't sure how to do this. Will try it.

api/docs/v2/new_atomic_commands.rst
jwwojak: Removed.

api/docs/v2/new_atomic_commands.rst
jwwojak: Added version changed indicator.

api/docs/v2/new_atomic_commands.rst
jwwojak: Slight text revision:

> To return a tip to its original location, call the ... method with no arguments::

Also, the text provides a link to the method. I'm gonna say let 'em click the link and check the original to see that.

api/docs/v2/new_atomic_commands.rst
jwwojak: Made changes to try and reduce this impression.

api/docs/v2/new_atomic_commands.rst
jwwojak: Adding text about API errors:

> If you omit the ``tip_rack`` argument from the ``pipette`` variable, the API will throw an error. You must pass in the tip rack's location to ``pick_up_tip`` like this::

Conclusion to this subsection revised too:

> If coding the location of each tip seems inefficient or tedious, try using a ``for`` loop to automate a sequential tip pick up process. When using a loop, the API tracks and manages the tip pickup locations automatically. But ``pick_up_tip`` is still a powerful feature. It lets you maintain control over tip use when that’s important in your protocol.

Leads in to the section on `for` loops.

api/docs/v2/new_atomic_commands.rst
jwwojak: Using that.

api/docs/v2/new_atomic_commands.rst
jwwojak: I don't quite follow this suggestion. If the objection is to mentioning specific versions, lets try this:

Starting sentence:

> Currently, the API considers tips as "used" after being picked up.

Next paragraph about API v2.0, 2.1:

> Early API versions treated returned tips as unused items.

Also, discussing how the API treats tips as "used" seems like good and interesting info. I'd suggest a Note isn't quite right and recommend keeping this in its own section.

Subsection ends with:

> .. versionchanged: 2.2

api/docs/v2/new_atomic_commands.rst
jwwojak: Isn't there an optical sensor on the Flex pipette that knows if a tip is attached? I think it's the OT-2 that doesn't know about tips in the same way as a Flex does. So is it the OT-2 that only knows if `has_tip` is yes or no based on previous commands?  I'm not sure what to do here.

api/docs/v2/new_atomic_commands.rst
ecormany: oh hey look what i found: RSS-277

jwwojak: This drove me a little crazy at the end of yesterday. I was trying all sorts of things.

`pipette.drop_tip(reservoir.wells_by_name()['A1'])`

setting weird variables:

```
# test alt trash
alt_trash_reservoir = reservoir
# and then
pipette.drop_tip(alt_trash_reservoir)
```

And even enlisted Jeremy in the struggle.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -134,27 +29,29 @@ This simple statement works because the variable ``tiprack_1`` in the sample pro
     pipette.pick_up_tip()  # picks up tip from rack location B1
     pipette.drop_tip()     # drops tip in trash bin 
 
-If you omit the ``tip_rack`` argument from the ``pipette`` variable, then you must pass in the tip rack's location to ``pick_up_tip`` like this::
+If you omit the ``tip_rack`` argument from the ``pipette`` variable, the API will throw an error. You must pass in the tip rack's location to ``pick_up_tip`` like this::
```
ecormany: we always `raise` instead of `throw` errors, because 🐍

api/docs/v2/new_atomic_commands.rst
ecormany: Ticketed as RTC-334 so we can revisit it later.

ecormany (approved):
:shipit: …to the feature branch.

PR #12292, opened 2023-03-14
feat(app,app-shell): usb request via app-shell

brenthagen:
This (large) PR enables http communication between the opentrons desktop app and an OT-3 robot via usb serial port connection. It adds a new package `usb-bridge/node-client` and uses it as follows:

1) `app` detects OT-3 usb device connection
2) `app-shell` fetches OT-3 serial port path associated with OT-3
3) `app-shell` creates a custom serial port http agent singleton
3) `discovery client` polls health for assigned hostname `opentrons-usb` using serial port http agent
4) for `opentrons-usb` hostname, `app` passes [electron two-way IPC invoke](https://www.electronjs.org/docs/latest/tutorial/ipc#pattern-2-renderer-to-main-two-way) requestor function to `ApiHostProvider` host context
5) `api-client` uses provided requestor function for http requests
6) `app-shell` handles IPC http requests, makes axios requests using serial port http agent

a handful of work left for this or follow-on PRs:
 - handle usb device removal correctly, so robot remains in unavailable list (currently it is removed entirely)
 - pass response status, ok via IPC
 - download robot logs (`electron-dl`) doesn't work with a custom agent - either inactivate for a usb connection or refactor how app-shell downloads
 - unit testing coverage
 - `usb-bridge/node-client` documentation
 - remove some cruft

PR #13388, opened 2023-08-25
fix(app,app-shell): download robot logs from from app not app-shell

brenthagen:
# Overview

in the initial [USB http request PR](https://github.com/Opentrons/opentrons/pull/12292) downloading robot logs via USB was left as follow-on work because `electron-dl` (our previous method of downloading logs) does not support a custom http agent. This is a partial implementation of that follow-on work.

this PR moves the robot logs download to the desktop app to use the existing HTTP request patterns for USB requests. the download is achieved by:
 1) on click, downloads and zips log files in parallel (typical network request) or serial (USB request) using `jszip` package
 2) when done, saves blob `${robotName}_logs.zip`

as tested in development, this approach works for wifi requests. however, USB log HTTP requests fail to settle dependent on two conditions:
 1) when the electron node process is stressed from excessive logging - either from debugging logs or the serialport readline parser
 2) robot-server status - some versions of 0.14.0 based off of edge work intermittently, but 7.0.0 alphas seem not to work. also, changing the response type of the logs endpoint from plain/text to application/json (which is analytically incorrect) seems to help.

our suspicion is that client USB agent software, possibly our subclassed custom http agent, either is not able to handle the lengthy single requests for logs (up to 50MB) or is not correctly recognizing the end of a logs request.

pending further investigation on the robot-server and usb-bridge side, we're opening this PR as an incremental step towards RQA-1225

# Test Plan

 - manually verified robot log downloads over network and (intermittently) USB
 - updated unit tests

# Changelog

 - download robot logs from from app instead of app-shell
 - removes electron-dl and related redux robot logs state

# Review requests

check async download approach in `Troubleshooting.tsx`

# Risk assessment

low-medium, it's a new way of downloading robot logs


sfoster1 (approved):
per an internet-less @shlokamin : looks good to me!

PR #13390, opened 2023-08-25
feat(hardware-testing): Update and separates increment volume tables per channels

andySigler:
# Overview

This PR updates the increment volume table for 8ch pipettes, to be ready for the PVT build.

We need different increment volume tables per pipette model, so this PR refactors the gravimetric slightly so that there is a unique table for each pipette+tip combo.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
approved

PR #13391, opened 2023-08-25
fix(app): update flex robot software available slideout

mjhuff:
Partially closes RAUT-501

# Overview

_This PR is one in a series of many addressing the new desktop app update flows._

Adds missing alert banners and disabled button status to robot slideout - this functionality was missing from the OT3 slideout. Updates display text in several locations to match updated design.

Screenshot 2023-08-25 at 4 42 17 PM

# Test Plan

- Verified out of date OT3s now properly render text and disabled button.

# Changelog

- Updated i18n text.
- Added selector logic for updates on OT3 slideout.

# Risk assessment

low


app/src/assets/localization/en/protocol_details.json
koji, line 1: Actually we use VSCode sorting function which is not perfect.
And that the reason why the order is like below lol

I guess all of us using VScode so maybe we can use a specific plugins or just write a simple script to fix this weird sorting in the future.

```
"labware_name": "Labware Name",
 "labware": "labware",
```

https://opentrons.slack.com/archives/CSCLVUW3C/p1683831830189419

app/src/organisms/SendProtocolToOT3Slideout/__tests__/SendProtocolToOT3Slideout.test.tsx
```diff
@@ -246,7 +246,7 @@ describe('SendProtocolToOT3Slideout', () => {
       onCloseClick: jest.fn(),
       isExpanded: true,
     })
-    const sendButton = getByRole('button', { name: 'Send' })
+    const sendButton = getByRole('button', { name: 'Proceed to setup' })
```
koji: The name is not Send so changing would be better.

```suggestion
    const proceedButton = getByRole('button', { name: 'Proceed to setup' })
```

app/src/organisms/SendProtocolToOT3Slideout/index.tsx
```diff
@@ -38,10 +38,23 @@ export function SendProtocolToOT3Slideout(
...
+  const isSelectedRobotOnWrongVersionOfSoftware = [
+    'upgrade',
+    'downgrade',
+  ].includes(
+    useSelector((state: State) => {
+      const value =
+        selectedRobot != null
+          ? getRobotUpdateDisplayInfo(state, selectedRobot.name)
+          : { autoUpdateAction: '' }
+      return value
```
koji:
```suggestion
        return selectedRobot != null
          ? getRobotUpdateDisplayInfo(state, selectedRobot.name)
          : { autoUpdateAction: '' }
```

or 

`value` => `updateInfo` something (I don't have any good names  🤔 )

koji (approved):
The changes look good to me.
I just left a couple of comments but they are small things.

app/src/assets/localization/en/protocol_details.json
mjhuff: Yes, I agree. My built in sorter configuration seems to be opposite 🤔

PR #13394, opened 2023-08-28
feat(app): adapter info in labware & liquid setup 

jerader:
closes RAUT-628

# Overview

This PR accomplishes 2 things:
1. refactors `getSlotLabwareName`, an old util name that no longer makes sense with what it is used for
2. changes protocol setup to match the new designs for adapters

# Test Plan

on the desktop app:

1. upload a protocol with adapters and liquids in it, go to the labware setup tab, see that the adapters are listed with their labware and that the adapter+labware combos are listed first
2. go to the liquid setup tab, see that the liquids mention the adapter and module info if applicable

on an odd:
1. send that protocol to the ODD, go to labware setup tab and see that the adapter+labware combos are listed together and if there is a H-S latch, the extra text regarding it is no longer there.

Screen Shot 2023-08-27 at 9 15 21 PM
Screen Shot 2023-08-27 at 9 15 42 PM
Screen Shot 2023-08-27 at 9 18 36 PM

# Changelog

- refactored `getSlotLabwareName` to `getLocationInfoNames` which now can return slotName, labware name, adapter name, and module model, wrote a test.
- created `getNestedLabwareInfo` which can return `nestedLabwareDisplayName`, `sharedSlotId`, `nestedLabwareDefinition`, and `nestedLabwareNickName` for the labware that is nested on top of the adapter, wrote a test.
- refactored `SetupLabwareList` and `ProtocolSetupLabware` to display the labware on top of adapter with the adapter item, fixed tests (NOTE: `ProtocolSetupLabware` needs to be cleaned up and refactored which will be done in a separate ticket https://opentrons.atlassian.net/browse/RAUT-631)
- refactored `SetupLiquidList` to display the adapter + module info if applicable.

# Review requests

see test plan

# Risk assessment

low


app/src/organisms/ProtocolSetupLabware/index.tsx
```diff
@@ -396,13 +400,23 @@ export function ProtocolSetupLabware({
           </Flex>
         </Flex>
         {[...onDeckItems, ...offDeckItems].map((labware, i) => {
-          return mostRecentAnalysis != null ? (
+          const labwareOnAdapter = onDeckItems.find(
```
b-cooper, line 403: nit: it looks like were just testing here to see "that this labware is not loaded onto an adapter initially". If we don't actually need the found item, this could be a some?

jerader: I don't think changing this to `some()` works. We are mapping each labware from `onDeckItems` individually in `RowLabware`. This check is to make sure if the labware is loaded onto an adapter initially then it isn't mapped in `RowLabware` because instead, it would be mapped with the adapter as a nested labware in `nestedLabwareInfo`

b-cooper (approved):
Code looks good 👍 , didn't have a chance to verify on robot

PR #13395, opened 2023-08-28
fix(app): fix non deterministic run command query

sanni-t:
# Overview

Prior to this PR, non-deterministic protocols, or runs whose analysis is stale and no longer matches the command keys in the actual run, would mistakenly query the command by its key instead of command ID. This would result in the app querying for non-existent commands and get `CommandDoesNotExistError` in response. This PR fixes that by changing the query to use command IDs for such queries.

# Test Plan

Run a protocol that has a different set of commands during analysis vs. during actual run. On the current 7.0.0 branch, you should get `CommandDoesNotExistError` errors during the run (either by checking the app's network request & response view or by checking the robot's system logs).

Those errors should not show up with this PR.

# Review requests

- does this line change affect anything other than non-deterministic protocols?

# Risk assessment

Low. Doesn't change the underlying working of a run, just corrects command status querying.


b-cooper (approved):
👍

PR #13396, opened 2023-08-28
docs(api): Refactor liquid control section

jwwojak:
# Overview

This PR is for managing changes made to the "Liquid Control" section of the doc, [Building Block Commands](https://docs.opentrons.com/v2/new_atomic_commands.html). 

Following the plan laid out in [this Figma file](https://www.figma.com/file/WsKj7gklIgDx68DkzegdaN/Commands-PAPI-docs-merge-strategies?type=whiteboard&node-id=0%3A1&t=souDTlzeZ3A0tmOj-1). This PR is for the second small branch section in Merge Strategy 1 of the Figma file.

The original manipulating tips section has 6 parts:

- Aspirate
- Dispense
- Blow out
- Touch tip
- Mix
- Air gap

 # Test Plan

- Check code in the Black formatter and use that style.
- Make sure links work.
- The usual grammar, style, clarity stuff.

# Changelog

- Adding text about the protocol template the code samples work with.
- Change headers to match current style.
- Revise text and code examples in each section.

# Review requests

The usual suspects.

# Risk assessment

Low, docs changes only.

### DIFF
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -495,15 +495,15 @@ The touch speed controls how fast the pipette moves in mm/s during a touch actio
 Mix
 ---
 
-The :py:meth:`.InstrumentContext.mix` method performs a series of aspirate and dispense on a single location. It's designed to help mix the contents of a well together using a single command rather than using multiple ``aspirate()`` and ``dispense()`` calls. This method includes arguments that let you specify the number of times to mix, the volume (in µL) of liquid, and the location of a well that contains the liquid you want to mix.   
+The :py:meth:`.InstrumentContext.mix` method performs a series of aspirate and dispense on a single location. It's designed to help mix the contents of a well together using a single command rather than using multiple ``aspirate()`` and ``dispense()`` calls. This method includes arguments that let you specify the mix repetitions, the amount of liquid (in µL) to mix, and the location of a well that contains the liquid you want to mix. Here are some examples::  
 
-.. code-block:: python
+    # mix from the pipette's current location 
+    pipette.mix(repetitions=3, volume=50)
 
-    # mix 3 times, 50 µL, current location
-    pipette.mix(3, 50)
-    # mix 4 times, 100 µL, from well A2
+    # mix 100 µL 4 times from well A2
     pipette.mix(4, 100, plate['A2'])
-    # mix 2 times, use the pipette's max volume, current location
+
+The volume amount is optional. If omitted, the pipette will use its maximum rated volume as the amount. For example, 
     pipette.mix(2)
 
 .. note::
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -338,7 +338,7 @@ The ``aspirate()`` method includes a ``location`` parameter that accepts either 
 
 **By Well**
 
-If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of a well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below. See also, :ref:`new-default-op-positions` for information about controlling pipette height for based on well location.
+If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of the well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below. See also, :ref:`new-default-op-positions` for information about controlling pipette height for based on well location.
 
 .. code-block:: python
 
@@ -353,8 +353,8 @@ You can also aspirate on a point along the center, z-axis within a well. For exa
 .. code-block:: python
 
     pipette.pick_up_tip()
-    depth = plate['A1'].bottom(z=2) # aspirate 2mm above well bottom
-    pipette.aspirate(200, depth)
+    depth = plate['A1'].bottom(z=2) # set pipette tp 2mm above well bottom
+    pipette.aspirate(200, depth)    # aspirate 2mm above well bottom
 
 Aspiration Flow Rates
 ^^^^^^^^^^^^^^^^^^^^^
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -517,15 +517,17 @@ The :py:meth:`~.InstrumentContext.mix` method performs a series of aspirate and 
 Air Gap
 -------
 
-The :py:meth:`.InstrumentContext.air_gap` method tells the pipette to aspirate some air after a liquid. Creating an air gap can help prevent a liquid from leaking out of the pipette after drawing it from a well. Calling the ``air_gap()`` method without any arguments uses the remaining volume in the pipette tip. For example, if you load a 1000 µL pipette on a Flex and tell the robot to aspirate 500 µL of liquid, it will use the remaining 500 µL for the air gap::
+The :py:meth:`~.InstrumentContext.air_gap` method tells the pipette to draw in air before and/or after aspiration. Creating an air gap helps keep liquid from seeping out of a pipette after drawing it from a well. Calling the this method without any arguments uses the remaining volume in the pipette tip for the air gap. For example, if you load a 1000 µL pipette on a Flex and tell the robot to aspirate 200 µL, it will use the remaining 800 µL for the air gap.
 
-    pipette.aspirate(500, plate['A1'])
-    pipette.air_gap() # draws 500 µL of air
+.. code-block:: python
 
-The ``air_gap()`` method also accepts the ``volume`` and ``height`` arguments. These let you control the amount of air (in µL) drawn in and the height (in mm) above the well for aspirating air. By default, the pipette will move 5 mm above a well before creating the air gap. For example, this code aspirates 200 µL of liquid from well A1. The aspiration action is followed by an air gap of 75 µL taken from 20 mm above the well::
+    pipette.aspirate(200, plate['A1']) # aspirate 200 µL from well A1
+    pipette.air_gap()                  # aspirate 800 µL of air
+
+The ``air_gap()`` method also accepts ``volume`` and ``height`` arguments. These let you control the amount of air (in µL) drawn in and the pipette's height (in mm) above the well. By default, the pipette moves 5 mm above a well before creating the air gap. For example, this code tells the robot to draw 200 µL of liquid from well A1 and create an air gap of 75 µL at 20 mm above well A1::
     
     pipette.aspirate(200, plate['A1'])
-    pipette.air_gap(75, 20)
+    pipette.air_gap(volume=75, height=20)
 
 .. versionadded:: 2.0
 
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -311,7 +311,7 @@ To draw liquid up into a pipette tip, call the :py:meth:`.InstrumentContext.aspi
     pipette.pick_up_tip()
     pipette.aspirate(200, plate['A1'])
 
-If the pipette doesn't move, you can also specify an additional aspiration action without including a location. To demonstrate, lets pause the protocol, automatically resume it, and aspirate a second time from ``plate['A1'])``.
+If the pipette doesn't move, you can also specify an additional aspiration action without including a location. To demonstrate, lets pause the protocol, automatically resume it, and aspirate a second time from ``plate['A1']``).
 
 .. code-block:: python
 
@@ -342,13 +342,13 @@ You can also aspirate from a location along the center vertical axis within a we
 See also:
 
 - :ref:`new-default-op-positions` for information about controlling pipette height for a particular pipette.
-- :ref:`position-relative-labware` for formation about controlling pipette height from within a well.
+- :ref:`position-relative-labware` for information about controlling pipette height from within a well.
 - :ref:`move-to` for information about moving a pipette to any reachable deck location.
 
 Aspiration Flow Rates
 ^^^^^^^^^^^^^^^^^^^^^
 
-Flex and OT-2 pipettes aspirate at :ref:`default flow rates <new-plunger-flow-rates>` measured in µL/s. Adding a number to the ``rate`` parameter multiplies the flow rate by that value. As a best practice, don't set the flow rate higher than 3x the default. For example, this code causes the pipette to aspirate at twice its normal rate::
+Flex and OT-2 pipettes aspirate at :ref:`default flow rates <new-plunger-flow-rates>` measured in µL/s. Specifying the ``rate`` parameter multiplies the flow rate by that value. As a best practice, don't set the flow rate higher than 3x the default. For example, this code causes the pipette to aspirate at twice its normal rate::
 
     pipette.aspirate(200, plate['A1'], rate=2.0)
 
@@ -480,7 +480,8 @@ This example uses the current well and sets the speed to 80 mm/s::
 
 .. versionadded:: 2.0
 
-.. there was a recommendation to remove the note
+.. versionchanged:: 2.4
+    Lowered minimum speed to 1 mm/s.
 
 .. _mix:
 
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -437,11 +437,10 @@ Flex and OT-2 pipettes dispense at :ref:`default flow rates <new-plunger-flow-ra
 Blow Out
 --------
 
-To blow an extra amount of air through the pipette's tip, call the :py:meth:`~.InstrumentContext.blow_out` method. A blow out is useful to clear the pipette of any remaining liquid. When using this method, you can use a specific well in a well plate or reservoir as a blow out location. If no location is specified, the pipette will blow out from its current well position. For example::
+To blow an extra amount of air through the pipette's tip, call the :py:meth:`~.InstrumentContext.blow_out` method. You can use a specific well in a well plate or reservoir as the blow out location. If no location is specified, the pipette will blow out from its current well position. For example::
 
     pipette.blow_out()            # blow out from the current location
     pipette.blow_out(plate['B1']) # blow out into well B1
-
 
 .. versionadded:: 2.0
 
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -297,23 +297,14 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
 Liquid Control
 ==============
 
-After attaching a tip, your robot is ready to aspirate, dispense, and perform other liquid handling tasks. The API provides the following methods that help you manipulate liquids in your protocols. These include:
-
-- :py:meth:`.InstrumentContext.aspirate`
-- :py:meth:`.InstrumentContext.dispense`
-- :py:meth:`.InstrumentContext.blow_out`
-- :py:meth:`.InstrumentContext.touch_tip`
-- :py:meth:`.InstrumentContext.mix`
-- :py:meth:`.InstrumentContext.air_gap`
-
-The following sections demonstrate how to use each method and include sample code. The examples used here assume that you've loaded the pipettes and labware from the basic :ref:`protocol template <protocol-template>`. 
+After attaching a tip, your robot is ready to aspirate, dispense, and perform other liquid handling tasks. The API includes methods that help you perform these actions and the following sections show how to use them. The examples used here assume that you've loaded the pipettes and labware from the basic :ref:`protocol template <protocol-template>`. 
 
 .. _new-aspirate:
 
 Aspirate
 --------
 
-To draw liquid up into a pipette tip, call the :py:meth:`~.InstrumentContext.aspirate` method. This method lets you specify the aspiration volume in µL, the well location, and pipette flow rate. Other parameters let you position the pipette within a well. For example, this snippet tells the robot to remove 200 µL from well location A1.
+To draw liquid up into a pipette tip, call the :py:meth:`.InstrumentContext.aspirate` method. This method lets you specify the aspiration volume in µL, the well location, and pipette flow rate. Other parameters let you position the pipette within a well. For example, this snippet tells the robot to aspirate 200 µL from well location A1.
 
 .. code-block:: python
 
@@ -334,11 +325,9 @@ Now our pipette holds 300 µL.
 Aspirate by Well or Location
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The ``aspirate()`` method includes a ``location`` parameter that accepts either a ``Well`` or a ``Location``. 
+The :py:meth:`~.InstrumentContext.aspirate` method includes a ``location`` parameter that accepts either a ``Well`` or a ``Location``. 
 
-**By Well**
-
-If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below.
+If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well (see also, :ref:`new-default-op-positions`). To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below. 
 
 .. code-block:: python
 
@@ -346,19 +335,13 @@ If you specify a well, like ``plate['A1']``, the pipette will aspirate from a de
     pipette.well_bottom_clearance.aspirate=2 # place tip 2 mm above well bottom
     pipette.aspirate(200, plate['A1'])
 
-See also, :ref:`new-default-op-positions` for information about controlling pipette height for a particular well.
-
-**By Location**
-
-You can also aspirate from a point along the center vertical axis within a well using the :py:obj:`.Well.top` and :py:meth:`.Well.bottom` methods. These methods move the pipette to a specified distance relative to the top or bottom center of a well. For example, you could change the default aspirate height as shown below.
+You can also aspirate from a point along the center vertical axis within a well using the :py:obj:`.Well.top` and :py:meth:`.Well.bottom` methods (see also, :ref:`position-relative-labware`). These methods move the pipette to a specified distance relative to the top or bottom center of a well. For example, you could change the default aspirate height as shown below.
 
 .. code-block:: python
 
     pipette.pick_up_tip()
     depth = plate['A1'].bottom(z=2) # place tip 2 mm above well bottom
     pipette.aspirate(200, depth)
-
-See also, :ref:`position-relative-labware` for information about on controlling pipette height from within a well.
 
 Aspiration Flow Rates
 ^^^^^^^^^^^^^^^^^^^^^
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -436,9 +436,9 @@ The :py:meth:`.InstrumentContext.mix` method performs a series of aspirate and d
 
 .. code-block:: python
 
-    # mix 3 times, 50uL, current location
+    # mix 3 times, 50 uL, current location
     pipette.mix(3, 50)
-    # mix 4 times, 100uL, from well A2
+    # mix 4 times, 100 uL, from well A2
     pipette.mix(4, 100, plate['A2'])
     # mix 2 times, use the pipette's max volume, current location
     pipette.mix(2)
@@ -454,13 +454,16 @@ The :py:meth:`.InstrumentContext.mix` method performs a series of aspirate and d
 Air Gap
 =======
 
-When dealing with certain liquids, you may need to aspirate air after aspirating the liquid to prevent it from sliding out of the pipette's tip. A call to :py:meth:`.InstrumentContext.air_gap` with a volume in µL will aspirate that much air into the tip. ``air_gap`` takes up to two arguments: ``air_gap(volume, height)``:
+The :py:meth:`.InstrumentContext.air_gap` method tells the pipette to aspirate some air after a liquid. Creating an air gap can help prevent a liquid from leaking out of the pipette after drawing it from a well. Calling the ``air_gap()`` method without any arguments uses the remaining volume in the pipette tip. For example, if you load a 1000 uL pipette on a Flex and tell the robot to aspirate 500 uL of liquid, it will use the remaining 500 uL for the air gap::
 
-.. code-block:: python
+    pipette.aspirate(500, plate['A1'])
+    pipette.air_gap() # draws 500 uL of air
 
-    pipette.aspirate(100, plate['B4'])
-    pipette.air_gap(20)
-    pipette.drop_tip()
+The ``air_gap()`` method also accepts the ``volume`` and ``height`` arguments. These let you control the amount of air (in uL) drawn in and the height (in mm) above the well for aspirating air. By default, the pipette will move 5 mm above a well before creating the air gap. For example, this code aspirates 200 uL of liquid from well A1. The aspiration action is followed by an air gap of 75 uL taken from 20 mm above the well::
+
+    pipette.pick_up_tip()
+    pipette.aspirate(200, plate['A1'])
+    pipette.air_gap(75, 20)
 
 .. versionadded:: 2.0
 
--- a/api/docs/v2/new_atomic_commands.rst
+++ b/api/docs/v2/new_atomic_commands.rst
@@ -318,7 +318,7 @@ If the pipette doesn't move, you can also specify an additional aspiration actio
     pipette.pick_up_tip()
     pipette.aspirate(200, plate['A1'])
     protocol.delay(seconds=5) # pause for 5 seconds
-    pipette.aspirate(100)     # aspirate 100 µL from current position
+    pipette.aspirate(100)     # aspirate 100 µL at current position
 
 Now our pipette holds 300 µL.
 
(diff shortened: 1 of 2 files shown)


api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
 
 **********************
 
-****************
+
 Liquid Control
-****************
+==============
+
+After attaching a tip, your robot is ready to aspirate, dispense, and perform other liquid handling tasks. The API provides the following methods that help you manipulate liquids in your protocols. These include:
```
ecormany: Some thoughts on this list…

- Either `the following methods` _or_ `These include` depending on whether this list is exhaustive — it at least appears to be w.r.t. what this page covers.
- For the list itself, we probably shouldn't repeat `InstrumentContext.` every time, for readability. 
- I'm torn on whether each item should link directly to the reference, or whether it should link down this page. Each subsection begins with a `:py:meth:` link to the reference, for anyone who's looking for reference entries rather than prose explanations.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
-The examples in this section should be inserted in the following:
+The following sections demonstrate how to use each method and include sample code. The examples used here assume that you've loaded the pipettes and labware from the basic :ref:`protocol template <protocol-template>`. 
+
+.. _new-aspirate:
+
+Aspirate
+--------
+
+To draw liquid up into a pipette tip, call the :py:meth:`~.InstrumentContext.aspirate` method. This method lets you specify the aspiration volume in µL, the well location, and pipette flow rate. Other parameters let you position the pipette within a well. For example, this snippet tells the robot to remove 200 µL from well location A1.
```
ecormany: `remove 200 µL from well location A1`
Probably trying to not repeat "aspirate" over and over here, but I think the repetition is OK. Or to paraphrase, something like `draw 200 µL into the tip` may be more precise than `remove`.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
-        # example code goes here
+If the pipette doesn't move, you can also specify an additional aspiration action without including a location. To demonstrate, lets pause the protocol, automatically resume it, and aspirate a second time from ``plate['A1'])``.
 
+.. code-block:: python
 
-This loads a `Corning 96 Well Plate <https://labware.opentrons.com/corning_96_wellplate_360ul_flat>`_ in slot 2 and a `Opentrons 300ul Tiprack <https://labware.opentrons.com/opentrons_96_tiprack_300ul>`_ in slot 3, and uses a P300 Single GEN2 pipette.
+    pipette.pick_up_tip()
+    pipette.aspirate(200, plate['A1'])
+    protocol.delay(seconds=5) # pause for 5 seconds
+    pipette.aspirate(100)     # aspirate 100 uL from current position
```
ecormany:
```suggestion
    pipette.aspirate(100)     # aspirate 100 µL from current position
```
pedantry

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
-.. _new-aspirate:
+Aspirate by Well or Location
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Aspirate
-========
+The ``aspirate()`` method includes a ``location`` parameter that accepts either a ``Well`` or a ``Location``. 
 
-To aspirate is to pull liquid up into the pipette's tip. When calling :py:meth:`.InstrumentContext.aspirate` on a pipette, you can specify the volume to aspirate in µL, where to aspirate from, and how fast to aspirate liquid.
+**By Well**
```
ecormany: oh no, we're down into H5 territory or something. but that won't be the case when this become its own page. all the headers will get promoted up a level (Liquid Control will be an H1) so this can be an H4 then. but it will show in the sidebar.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
-.. code-block:: python
+If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below.
 
-    pipette.aspirate(50, plate['A1'], rate=2.0)  # aspirate 50uL from plate:A1
+.. code-block:: python
 
-Now the pipette's tip is holding 50 µL.
+    pipette.pick_up_tip
+    pipette.well_bottom_clearance.aspirate=2 # place tip 2 mm above well bottom
```
ecormany:
```suggestion
    pipette.well_bottom_clearance.aspirate = 2 # place tip 2 mm above well bottom
```
⬛️

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
-Aspirate
-========
+The ``aspirate()`` method includes a ``location`` parameter that accepts either a ``Well`` or a ``Location``. 
 
-To aspirate is to pull liquid up into the pipette's tip. When calling :py:meth:`.InstrumentContext.aspirate` on a pipette, you can specify the volume to aspirate in µL, where to aspirate from, and how fast to aspirate liquid.
+**By Well**
 
-.. code-block:: python
+If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below.
```
ecormany: technically not a call because clearance is an attribute 🙃 
```suggestion
If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well. To change the default clearance, set :py:obj:`.well_bottom_clearance`:
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
-    pipette.aspirate(50, plate['A1'], rate=2.0)  # aspirate 50uL from plate:A1
+.. code-block:: python
 
-Now the pipette's tip is holding 50 µL.
+    pipette.pick_up_tip
+    pipette.well_bottom_clearance.aspirate=2 # place tip 2 mm above well bottom
+    pipette.aspirate(200, plate['A1'])
 
-The ``location`` parameter is either a well (like ``plate['A1']``) or a position within a well, like the return value of ``plate['A1'].bottom``.
+See also, :ref:`new-default-op-positions` for information about controlling pipette height for a particular well.
```
ecormany:
```suggestion
See also, :ref:`new-default-op-positions` for information about controlling aspiration height for a particular pipette.
```
…since this is a pipette parameter.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
-The ``rate`` parameter is a multiplication factor of the pipette's default aspiration flow rate. The default aspiration flow rate for all pipettes is in the :ref:`defaults` section.
+**By Location**
 
-You can also simply specify the volume to aspirate, and not mention a location. The pipette will aspirate from its current location (which we previously set as ``plate['A1'])``.
+You can also aspirate from a point along the center vertical axis within a well using the :py:obj:`.Well.top` and :py:meth:`.Well.bottom` methods. These methods move the pipette to a specified distance relative to the top or bottom center of a well. For example, you could change the default aspirate height as shown below.
```
ecormany: use `obj` or `meth` (probably the latter) for both. this currently renders:
>[Well.top](http://sandbox.docs.opentrons.com/docs-liquid-control/v2/new_protocol_api.html#opentrons.protocol_api.Well.top) and [Well.bottom()](http://sandbox.docs.opentrons.com/docs-liquid-control/v2/new_protocol_api.html#opentrons.protocol_api.Well.bottom)

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
 .. code-block:: python
 
-    pipette.aspirate(50)                     # aspirate 50uL from current position
+    pipette.pick_up_tip()
+    depth = plate['A1'].bottom(z=2) # place tip 2 mm above well bottom
+    pipette.aspirate(200, depth)
 
-Now our pipette's tip is holding 100 µL.
+See also, :ref:`position-relative-labware` for information about on controlling pipette height from within a well.
```
ecormany: worth linking to section on `move_to()` in case users need to change X or Y position for aspirate?

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
+    pipette.dispense(200, plate['B1'])
 
+If the pipette doesn’t move, you can also specify an additional dispense action without including a location. To demonstrate, lets assume the pipette currently holds 200 µL. Next, we'll pause the protocol, automatically resume it, and dispense a second time into well location B1.
 
-The ``location`` parameter is either a well (like ``plate['A1']``) or a position within a well, like the return value of ``plate['A1'].bottom``.
+.. code-block:: python
+    
+    pipette.dispense(100, plate['B1'])
+    protocol.delay(seconds=5) # pause for 5 seconds
+    pipette.aspirate(100)     # aspirate 100 uL from current position
```
ecormany:
```suggestion
    pipette.dispense(100)     # dispense 100 µL at current position
```
📋 🍝

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,140 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
-    :ref:`v2-location-within-wells`) every time you call ``dispense``, or - if you want to change
-    the default throughout your protocol - you can change the default offset with
-    :py:obj:`.InstrumentContext.well_bottom_clearance` (see :ref:`new-default-op-positions`).
+If you specify a well, like ``plate['B1']``, the pipette will dispense from a default position 1 mm above the bottom center of that well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance` as shown below. 
 
-.. note::
+.. code-block:: python
+
+    pipette.well_bottom_clearance.dispense=2 # place tip 2 mm above well bottom
+    pipette.aspirate(200, plate['B1'])
```
ecormany:
```suggestion
    pipette.dispense(200, plate['B1'])
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
 
-To blow out is to push an extra amount of air through the pipette's tip, to make sure that any remaining droplets are expelled.
-
-When calling :py:meth:`.InstrumentContext.blow_out`, you can specify a location to blow out the remaining liquid. If no location is specified, the pipette will blow out from its current position.
-
-.. code-block:: python
-
-    pipette.blow_out()            # blow out in current location
-    pipette.blow_out(plate['B3']) # blow out in current plate:B3
+To blow an extra amount of air through the pipette's tip, call the :py:meth:`~.InstrumentContext.blow_out` method. You can use a specific well in a well plate or reservoir as the blow out location. If no location is specified, the pipette will blow out from its current well position. For example::
```
ecormany: `blowout` as a modifier? i think so. so `blowout location`.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
 
+    pipette.blow_out()            # blow out from the current location
+    pipette.blow_out(plate['B1']) # blow out into well B1
```
ecormany: probably also useful to demonstrate 
```
pipette.blow_out(protocol.fixed_trash['A1'])  # blow out into the trash
```
i think it's a pretty common use case.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
                       v_offset=-2)
 
-
 .. versionadded:: 2.0
 
 .. note:
 
-    It is recommended that you change your API version to 2.4 to take advantage of new
-    features added into `touch_tip` such as:
+    We recommend changing your API version to 2.4 (or higher) to take advantage of new ``touch_tip()``
```
ecormany: i think this level of detail could be saved for the reference. 2.4 is pretty old too.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
-    pipette.touch_tip(plate['B1'], speed=100) # touch tip within plate:B1 at 100 mm/s
-    pipette.touch_tip(plate['B1'], # touch tip in plate:B1, at 75% of total radius and -2mm from top of well
+    # touch tip in well B1
+    pipette.touch_tip(plate['B1'])
+    # touch tip 2mm below the top of the current location
+    pipette.touch_tip(v_offset=-2) 
+    # touch tip in well B1 at 100 mm/s
+    pipette.touch_tip(plate['B1'], speed=100)
+    # touch tip in well B1, at 75% of total radius and -2mm from top of well
+    pipette.touch_tip(plate['B1'],
```
ecormany, line 282: kinda weird this is on 3 lines, it'll fit on 1. the snippet gets a little dense then, but we could put blank lines between each comment–command pair.

ecormany: it would also be nice to figure out why adjusting radius is helpful. it's not self-explanatory to me. if you reduce it, might the tip not actually touch? unless there are big droplets maybe? 🤷

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
 .. _mix:
 
 Mix
-===
+---
 
-To mix is to perform a series of ``aspirate`` and ``dispense`` commands in a row on a single location. Instead of having to write those commands out every time, you can call :py:meth:`.InstrumentContext.mix`.
-
-The ``mix`` command takes up to three arguments: ``mix(repetitions, volume, location)``:
+The :py:meth:`.InstrumentContext.mix` method performs a series of aspirate and dispense on a single location. It's designed to help mix the contents of a well together using a single command rather than using multiple ``aspirate()`` and ``dispense()`` calls. This method includes arguments that let you specify the number of times to mix, the volume (in uL) of liquid, and the location of a well that contains the liquid you want to mix.
```
ecormany: possibility for first sentence: 
```suggestion
The :py:meth:`~.InstrumentContext.mix` method aspirates and dispenses repeatedly in a single location. It's designed to help mix the contents of a well together using a single command rather than using multiple ``aspirate()`` and ``dispense()`` calls. This method includes arguments that let you specify the number of times to mix, the volume (in µL) of liquid, and the well that contains the liquid you want to mix.   
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
 
-    # mix 4 times, 100uL, in plate:A2
-    pipette.mix(4, 100, plate['A2'])
-    # mix 3 times, 50uL, in current location
+    # mix 3 times, 50 uL, current location
     pipette.mix(3, 50)
-    # mix 2 times, pipette's max volume, in current location
+    # mix 4 times, 100 uL, from well A2
+    pipette.mix(4, 100, plate['A2'])
+    # mix 2 times, use the pipette's max volume, current location
```
ecormany: maybe split this out as a separate snippet and explain this behavior in a sentence (in addition to the comment)

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
     pipette.mix(2)
 
 .. note::
 
-    In API Versions 2.2 and earlier, mixes consist of aspirates and then immediate dispenses. In between these actions, the pipette moves up and out of the target well. In API Version 2.3 and later, the pipette will not move between actions. 
+    In API Versions 2.2 and earlier, during a mix, the pipette moves up and out of the target well. In API Version 2.3 and later, the pipette does not move between aspirate and dispense mixing actions.
```
ecormany: downcase `version` — feels like old style (or lack thereof)

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
 .. note::
 
-    In API Versions 2.2 and earlier, mixes consist of aspirates and then immediate dispenses. In between these actions, the pipette moves up and out of the target well. In API Version 2.3 and later, the pipette will not move between actions. 
+    In API Versions 2.2 and earlier, during a mix, the pipette moves up and out of the target well. In API Version 2.3 and later, the pipette does not move between aspirate and dispense mixing actions. 
 
 .. versionadded:: 2.0
 
 .. _air-gap:
 
 Air Gap
```
ecormany, line 338: `uL` -> `µL` throughout this section

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +435,86 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
-=======
+-------
 
-When dealing with certain liquids, you may need to aspirate air after aspirating the liquid to prevent it from sliding out of the pipette's tip. A call to :py:meth:`.InstrumentContext.air_gap` with a volume in µL will aspirate that much air into the tip. ``air_gap`` takes up to two arguments: ``air_gap(volume, height)``:
+The :py:meth:`.InstrumentContext.air_gap` method tells the pipette to aspirate some air after a liquid. Creating an air gap can help prevent a liquid from leaking out of the pipette after drawing it from a well. Calling the ``air_gap()`` method without any arguments uses the remaining volume in the pipette tip. For example, if you load a 1000 uL pipette on a Flex and tell the robot to aspirate 500 uL of liquid, it will use the remaining 500 uL for the air gap::
```
ecormany: maybe `aspirate air instead of liquid`. it typically comes after aspirating liquid but doesn't have to. i've heard that FAS uses "pre-air-gapping" as a technique for certain applications.

ecormany (changes requested):
Lots of little stuff but in general this section covers what it needs to!

api/docs/v2/new_atomic_commands.rst
ecormany: and on the flip side, we should probably have a `.. warning::` about large values of `radius` 💥

api/docs/v2/new_atomic_commands.rst
jwwojak: I was trying to avoid over-use of "aspirate." Will put it back in that last sentence.

> For example, this snippet tells the robot to aspirate 200 µL from well A1.

api/docs/v2/new_atomic_commands.rst
jwwojak: No no, not  pedantry at all. An important style issue/consideration. Thank you for highlighting that. I usually fix these closer to the end of the drafting process. I think all the instances of "uL" have been replaced by "µL". 

Same with any occurrences of other measurement units and numbers (2mm vs 2 mm).

api/docs/v2/new_atomic_commands.rst
jwwojak: Oh no indeed!  I will re-write to remove the bold headers. Didn't realize these would be H5s, which is horrible.

api/docs/v2/new_atomic_commands.rst
jwwojak: I was uncomfortable with the bullet list too. Let's try something different, a bit more general without links to methods or page sections. I will summarize what the page covers and let the section headers and table of contents do the work of links in a bullet list.

api/docs/v2/new_atomic_commands.rst
jwwojak: > kinda weird this is on 3 lines, it'll fit on 1.

I want to put all these in a table, so you're lucky we get this. haha!

api/docs/v2/new_atomic_commands.rst
jwwojak: These have been fixed already.

api/docs/v2/new_atomic_commands.rst
jwwojak: Yes, will use separate sentences and code samples and replace the large code block/comments combo.

api/docs/v2/new_atomic_commands.rst
jwwojak: fixed

api/docs/v2/new_atomic_commands.rst
jwwojak: fixed

api/docs/v2/new_atomic_commands.rst
jwwojak: great use case. Adding.

api/docs/v2/new_atomic_commands.rst
jwwojak: Removed

api/docs/v2/new_atomic_commands.rst
jwwojak: Updating text, added warning, and checking around about _why_ set the radius to less than 1.

jwwojak: Added warning, revised text and code samples to put them more with/in text.

api/docs/v2/new_atomic_commands.rst
jwwojak: Adopted

api/docs/v2/new_atomic_commands.rst
jwwojak: Let's try this maybe:

> The :py:meth:`.InstrumentContext.air_gap` method tells the pipette to draw in air before or after a liquid.

api/docs/v2/new_atomic_commands.rst
jwwojak: Fixed.

api/docs/v2/new_atomic_commands.rst
jwwojak: fixed

api/docs/v2/new_atomic_commands.rst
jwwojak: added shortcut to `move_to()`

api/docs/v2/new_atomic_commands.rst
jwwojak: Revisions include suggested change.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,119 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
-        pipette = protocol.load_instrument('p300_single_gen2', mount='left', tip_racks=[tiprack])
-        pipette.pick_up_tip()
-        # example code goes here
+.. code-block:: python
 
+    pipette.pick_up_tip()
+    pipette.aspirate(200, plate['A1'])
 
-This loads a `Corning 96 Well Plate <https://labware.opentrons.com/corning_96_wellplate_360ul_flat>`_ in slot 2 and a `Opentrons 300ul Tiprack <https://labware.opentrons.com/opentrons_96_tiprack_300ul>`_ in slot 3, and uses a P300 Single GEN2 pipette.
+If the pipette doesn't move, you can also specify an additional aspiration action without including a location. To demonstrate, lets pause the protocol, automatically resume it, and aspirate a second time from ``plate['A1'])``.
```
ecormany: Maybe `If you don't want the pipette to move` or `If you want to aspirate at the pipette's current location`. Current phrasing makes it sound like the pipette or the API is in control, not the protocol author.

Also a li'l format fix:
```suggestion
If the pipette doesn't move, you can also specify an additional aspiration action without including a location. To demonstrate, lets pause the protocol, automatically resume it, and aspirate a second time from ``plate['A1']``).
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,119 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
+.. code-block:: python
 
-.. _new-aspirate:
+    pipette.pick_up_tip()
+    pipette.aspirate(200, plate['A1'])
+    protocol.delay(seconds=5) # pause for 5 seconds
```
ecormany, line 145: Here and everywhere, two spaces before the `#` that begins an inline comment. Because Black.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,119 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
 
-To aspirate is to pull liquid up into the pipette's tip. When calling :py:meth:`.InstrumentContext.aspirate` on a pipette, you can specify the volume to aspirate in µL, where to aspirate from, and how fast to aspirate liquid.
+Aspirate by Well or Location
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: python
+The :py:meth:`~.InstrumentContext.aspirate` method includes a ``location`` parameter that accepts either a ``Well`` or a ``Location``. 
 
-    pipette.aspirate(50, plate['A1'], rate=2.0)  # aspirate 50uL from plate:A1
+If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well. To change the default clearance, you would call :py:obj:`.well_bottom_clearance`::
```
ecormany:
```suggestion
If you specify a well, like ``plate['A1']``, the pipette will aspirate from a default position 1 mm above the bottom center of that well. To change the default clearance, first set the ``aspirate`` attribute of :py:obj:`.well_bottom_clearance`:: 
```
Adding a little clarity on how `well_bottom_clearance` operates and the importance of ordering here.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,119 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
+    pipette.pick_up_tip()
+    depth = plate['A1'].bottom(z=2) # tip is 2 mm above well bottom
+    pipette.aspirate(200, depth)
 
-.. code-block:: python
+See also:
 
-    pipette.aspirate(50)                     # aspirate 50uL from current position
+- :ref:`new-default-op-positions` for information about controlling pipette height for a particular pipette.
+- :ref:`position-relative-labware` for formation about controlling pipette height from within a well.
```
ecormany:
```suggestion
- :ref:`position-relative-labware` for information about controlling pipette height from within a well.
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,119 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
+- :ref:`move-to` for information about moving a pipette to any reachable deck location.
 
-Now our pipette's tip is holding 100 µL.
+Aspiration Flow Rates
+^^^^^^^^^^^^^^^^^^^^^
 
-.. note::
+Flex and OT-2 pipettes aspirate at :ref:`default flow rates <new-plunger-flow-rates>` measured in µL/s. Adding a number to the ``rate`` parameter multiplies the flow rate by that value. As a best practice, don't set the flow rate higher than 3x the default. For example, this code causes the pipette to aspirate at twice its normal rate::
```
ecormany:
```suggestion
Flex and OT-2 pipettes aspirate at :ref:`default flow rates <new-plunger-flow-rates>` measured in µL/s. Specifying the ``rate`` parameter multiplies the flow rate by that value. As a best practice, don't set the flow rate higher than 3x the default. For example, this code causes the pipette to aspirate at twice its normal rate::
```
To avoid confusion with arithmetic addition.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -293,97 +293,119 @@ To check whether you should pick up a tip or not, you can utilize :py:meth:`.Ins
...
+--------
+
+To dispense liquid from a pipette tip, call the :py:meth:`.InstrumentContext.dispense` method. This method lets you specify the dispense volume in µL, the well location, and pipette flow rate. Other parameters let you position the pipette within a well. For example, this snippet tells the robot to dispense 200 µL into well location B1.
+
+.. code-block:: python
+
+    pipette.dispense(200, plate['B1'])
 
-To dispense is to push out liquid from the pipette's tip. The usage of :py:meth:`.InstrumentContext.dispense` in the Protocol API is similar to :py:meth:`.InstrumentContext.aspirate`, in that you can specify volume in µL and location, or only volume.
+If the pipette doesn’t move, you can also specify an additional dispense action without including a location. To demonstrate, lets pause the protocol, automatically resume it, and dispense a second time from location B1.
```
ecormany: Mirror changes to language in Aspirate § above (here and in following paragraphs).

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +414,120 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
-.. code-block:: python
+You can also specify a particular well as the blowout location::
+
+    pipette.blow_out(plate['B1'])
 
-    pipette.blow_out()            # blow out in current location
-    pipette.blow_out(plate['B3']) # blow out in current plate:B3
+Many protocols use trash bin for blowing out the pipette. You can specify the trash bin as the blowout location by using the :py:meth:`.ProtocolContext.fixed_trash` method::
 
+    pipette.blow_out(protocol.fixed_trash['A1'])
```
ecormany, line 254: This works fine on both robots for now, but we should monitor progress of RSS-290. Ticketed on our side as RTC-333.

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +414,120 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
-    pipette.touch_tip(plate['B1'], # touch tip in plate:B1, at 75% of total radius and -2mm from top of well
+    pipette.touch_tip(v_offset=-2) 
+
+This example moves the pipette 75% of well's total radius and 2 mm below the top of well::
+
+    pipette.touch_tip(plate['B1'], 
                       radius=0.75,
                       v_offset=-2)
 
+When ``radius=1``, the robot moves the centerline of the pipette’s plunger axis to the exact edge of the well. Sometimes this means a pipette tip may hit the well wall or edge first and bends inwards. Setting the radius to < 1 helps avoid hitting the well wall prematurely. If you need to use it, test the radius argument first before starting a protocol production run.
```
ecormany: Maybe a chance for some levity here, along the lines of:
"The purpose of `touch_tip` is to _touch_ the edges of the well, not to crash into them. Using a smaller radius and lower speed produces gentler motion. Droplets of different types of liquid behave differently, so test out these parameters in a single well before performing a full protocol run."

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +414,120 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
-        moving directly to the height offset specified.
+    pipette.touch_tip(plate['B1'], speed=20)
+
+This example uses the current well and sets the speed to 80 mm/s::
+
+    pipette.touch_tip(speed=80)
+
+.. versionadded:: 2.0
+
+.. there was a recommendation to remove the note
```
ecormany: we could compress the note rather than removing it:
```suggestion
.. versionchanged:: 2.4
    Lowered minimum speed to 1 mm/s.
```

api/docs/v2/new_atomic_commands.rst
```diff
@@ -392,93 +414,120 @@ The ``rate`` parameter is a multiplication factor of the pipette's default dispe
...
-=======
+-------
 
-When dealing with certain liquids, you may need to aspirate air after aspirating the liquid to prevent it from sliding out of the pipette's tip. A call to :py:meth:`.InstrumentContext.air_gap` with a volume in µL will aspirate that much air into the tip. ``air_gap`` takes up to two arguments: ``air_gap(volume, height)``:
+The :py:meth:`.InstrumentContext.air_gap` method tells the pipette to draw in air before or after a liquid. Creating an air gap helps keep liquids from seeping out of a pipette after drawing it from a well. This method includes arguments that let you control the amount of air to aspirate and the pipette's height (in mm) above the well. By default, the pipette moves 5 mm above a well before aspirating air. Calling :py:meth:`~.InstrumentContext.air_gap` with no arguments uses the entire remaining volume in the pipette.
 
-.. code-block:: python
+This example aspirates 200 µL of air 5 mm above the current well::
+
+    pipette.air_gap(volume=200)
```
ecormany, line 345: 200 is a biiiiiiig air gap. Maybe use 20, like the existing example.

api/docs/v2/new_atomic_commands.rst
ecormany, line 1: Check for `lets` --> `let's` throughout.

ecormany (commented):
This is a straightforward section, and it's definitely cleaner than before. A few more touchups and it'll be good to go!

api/docs/v2/new_atomic_commands.rst
jwwojak: Lets as in 3rd person singular is ok, right? As in giving you the ability, capability, or permission to do whatever it is. This method "allows you to do the thing"


> This method lets you . . .  
not

> This method let's you . . .

ecormany: Yeah, not one for find and replace! Regular expressions can't save you here 🤪

api/docs/v2/new_atomic_commands.rst
jwwojak: The section is based on, and linked to, those sample protocol files on another page. I selected 200 b/c the linked examples use a 1000 uL pipette. Drawing 20 ul of air from a 1000 ul pipette doesn't seem like a good example (or even accurately possible). That's why this uses such a large amount of air.

From the section intro:

> The examples used here assume that you've loaded the pipettes and labware from the basic `:ref:protocol template <protocol-template>.`

api/docs/v2/new_atomic_commands.rst
jwwojak: Ooof, that's what passes for humor? Borscht Belt here we come! Time for some Pocono Palaver. Tip your waitstaff, enjoy the veal, I'm here all week. badum tsssss.

I'd like to keep some of Seth's explanation in here. How about trying to combine some of these:

**Revised**
The  `touch_tip` feature allows the pipette to touch the edges of a well gently instead of crashing into them. When `radius=1` the robot moves the centerline of the pipette’s plunger axis to the edge of the well. This means a pipette tip may sometimes hit the well wall too early, causing the tip to bend inwards. A smaller radius helps avoid a premature wall collision and a lower speed produces gentler motion. Different liquid droplets behave differently, so test out these parameters in a single well before performing a full protocol run.

ecormany (approved):
New changes look good, this is ready to merge into the feature branch.

PR #12648, opened 2023-05-05
fix(robot-server): memory cache analyses

sfoster1:
Getting analysis results can really bog down the robot server - in fact, it can block the event loop such that new http requests get stalled, and that means that apps will mark the server unhealthy.

Recently, the app started asking for analyses much more frequently, which would cause the robot server to bog down more frequently, which would cause the app to mark the robot unhealthy more frequently. This was an unintended change, and reverting it fixes the robot unhealthiness; but getting analysis results still threatens to bog down the server, so this memcache should fix the problem on the server.

We have similar problems (and solutions) for other heavy resources like runs and protocols; but there we do the caching with functools.lru_cache. This is not possible for analysis results because we get analysis results with async functions, for a very good reason; so instead, we need a custom cache implementation.

This cache implementation uses a python dict to hold the result objects; since there are limited places to add results, we can add things to the cache when they're added to the database and add them to the cache when we get things from the database. We use a deque to maintain a FIFO for cache eviction.

## Review requests
Caches like this aren't really my wheelhouse; think about the behavior here and whether it's correct.

## Testing
- [ ] put this on a robot with a lot of protocols stored and retrieve analyses and see if it falls over

Supercedes #12640

PR #13397, opened 2023-08-28
perf(robot-server): Handle analysis requests serially to take more advantage of cache

SyntaxColoring:
# Overview

When an HTTP client requests a protocol analysis, the server:

1. Retrieves a binary blob of the analysis from the database.
2. Parses and validates the blob into a Python `CompletedAnalysisResource` objects.
3. Serializes the `CompletedAnalysisResource` down to JSON and returns it to the client.

Step 2 is a known major compute bottleneck. PR #12648 put it behind a cache, which helps a lot. But that implementation has a hole: if there are two simultaneous requests for the same uncached analysis, they'll *both* parse the analysis from scratch. Ideally, one of the requests would wait for the other one to complete, and just reuse its result. This PR makes it do that.

This was an attempt to treat RSS-328. For unknown reasons, it doesn't improve the overall *user-facing* time, even though our *benchmarks* do show an improvement.

# Test plan

Behaviorally, I think we're adequately covered by existing integration and unit tests.

Performance-wise, I've tested this by running [this benchmarking script](https://gist.github.com/SyntaxColoring/8555a2dc0c36a7bc1a654b85c82f982b) from the discussion in RSS-328. I've confirmed that when running it in parallel 3 times, the "getting each protocol analysis" phase is ~3x faster: ~160s, down from ~500s.

# Changelog

* If an analysis request might be cached or add to the cache, handle it serially with respect to other such requests.
    * Add an `asyncio.Lock` to `CompletedAnalysisStore`.
    * Acquire that lock every time something calls `.get_by_id()` or `.get_by_protocol()`.

# Review requests

* Does this fix make sense?
* Can the comments be made more clear or concise?
* This implementation has a limitation where small requests will get blocked behind big ones. This doesn't affect overall throughput, but it does affect latency. There are ways around this, but they're significantly more complicated, so I left them out. Does this sound acceptable?

# Risk assessment

Low, but see the note above about small requests getting blocked behind big ones.

### DIFF
--- a/robot-server/robot_server/protocols/analysis_memcache.py
+++ b/robot-server/robot_server/protocols/analysis_memcache.py
@@ -1,5 +1,6 @@
 """A simple size-limited memory cache used for large resources."""
-from typing import Generic, TypeVar, Deque, Type, Dict
+import asyncio
+from typing import Awaitable, Callable, Coroutine, Generic, TypeVar, Deque, Type, Dict
 from logging import getLogger
 
 from collections import deque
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
This looks like it will help and be a step forward, but I think the future is probably to offload a lot of this work to multiprocessing and sharedmem communication so we can parallelize to some degree and also nice down those processes so they don't block.

SyntaxColoring:
Discussed with RSS people in-person.

PR #13398, opened 2023-08-28
fix(hardware-testing): Update z-stage test to latest PVT version

andySigler:
# Overview

Commits were added to print results at end of test, and to set the device-under-test to be "other" so the CSV doesn't set the test's device ID to be the SOM's serial.

# Test Plan

# Changelog

# Review requests

# Risk assessment


fsinapi (approved):
this looks good to me 👍

PR #13399, opened 2023-08-28
fix(hardware-testing): Reduce columns in Grav CSV to 3 by disabling CSVReport meta-data checking

andySigler:
# Overview

This PR adds an optional argument to `CSVReport` to disable meta-data validation, and then sets that arg in gravimetric tests so that the number of columns in the CSV doesn't increase from 3 to 5.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
lgtm

PR #13400, opened 2023-08-28
fix(api): Always engage backlash-compensation when moving to plunger bottom

andySigler:
# Overview

While debugging 1uL issues, we found this bug in the backlash-compensation logic where the plunger wasn't guaranteed to be engaged in the aspirate position each time it was at `bottom`. This PR changes `OT3API._move_to_plunger_bottom()` so that it guarantees the plunger is engaged in the aspirate direction every time.

NOTE: This bug-fix was found to be critical in tested for 96ch only, so this does not need to be included in 7.0.0

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1542,7 +1542,8 @@ class OT3API(
         Possible events where this occurs:
 
         1. After homing the plunger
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Okay, this makes sense to me, nice!

ryanthecoder (approved):
lgtm

PR #13402, opened 2023-08-28
feat(hardware-testing): Adds --same-tip and --mix as experimental arguments to gravimetric test

andySigler:
# Overview

Some new experimental options were added last week to gravimetric testing, and they are useful enough that we should probably keep them around forever:

- Argument `--same-tip` where the pipette will never replace the current tip(s) during the entire test
- Argument `--mix` actually does something now, where the pipette will mix 5x times before every aspiration

A couple of small bug fixes too:

- Remove unused `--inspect` argument
- Fix retract bug where specified millimeter height was being added to top of vial, instead of being added to liquid surface
  - because of above bug, increases retract distance from 3mm to 5mm so it is guaranteed to leave liquid

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/hardware-testing/hardware_testing/gravimetric/__main__.py
+++ b/hardware-testing/hardware_testing/gravimetric/__main__.py
@@ -322,6 +322,7 @@ def build_gravimetric_cfg(
     isolate_channels: List[int],
     extra: bool,
     jog: bool,
+    same_tip: bool,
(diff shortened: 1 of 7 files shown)


ryanthecoder (approved):
lgtm

PR #13405, opened 2023-08-28
feat(api,robot-server): Don't scan for wifi networks on every request and expose option via query parameter

vegano1:
# Overview

Scanning for wifi networks is an expensive operation. Since we have multiple clients including the ODD, that can often request the wifi list numerous times a second, this can lead to the wifi driver on the Flex crashing. Since we call out to `nmcli` (NetworkManager Command Line Interface) command via a python subprocess which when scanning can take about 10-20 seconds to return, this can lead to the creation of many file descriptors which can lead to a resource accumulation problem. 

So let's make the triggering of network scans optional so it does not happen on every `/wifi/list` request but instead exposed via a query parameter `rescan` which is `False` by default, allowing the client to control when they want a new scan. This means that we won't perform a network scan every time we request the wifi list, instead, [nmcli](https://developer-old.gnome.org/NetworkManager/stable/nmcli.html) will return its cached list of wifi networks which is internally refreshed at least every 30 seconds.

Closes: [RQA-1238](https://opentrons.atlassian.net/browse/RQA-1238)

# Test Plan

- [x] GET `/wifi/list` without a query parameter returns cached nmcli wifi networks.
- [x] GET `/wifi/list?rescan=false` returns cached nmcli wifi networks.
- [x] GET `/wifi/list?rescan=true` triggers a network rescan, returning the latest wifi networks.
- [x] Test on the OT-2
- [x] Test on the Flex

# Changelog
- Make nmcli wifi rescan optional and expose it via a query parameter
- Update command descriptions to say `robot` instead of just `OT2` since we also have the Flex.

# Review requests
- is there anything I'm missing?

# Risk assessment
- Medium, affects OT-2 and Flex 
- wifi networks from `/wifi/list` will be slightly stale but will increase performance.

[RQA-1238]: https://opentrons.atlassian.net/browse/RQA-1238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/service/legacy/routers/networking.py
+++ b/robot-server/robot_server/service/legacy/routers/networking.py
@@ -4,10 +4,11 @@ import subprocess
 
 from starlette import status
 from starlette.responses import JSONResponse
-from fastapi import APIRouter, HTTPException, File, Path, UploadFile
+from typing import Optional
+from fastapi import APIRouter, HTTPException, File, Path, UploadFile, Query
+
 from opentrons_shared_data.errors import ErrorCodes
 from opentrons.system import nmcli, wifi
-
 from robot_server.errors import LegacyErrorResponse
 from robot_server.service.legacy.models import V1BasicResponse
 from robot_server.service.legacy.models.networking import (
@@ -36,7 +37,7 @@ router = APIRouter()
 @router.get(
     "/networking/status",
     summary="Query the current network connectivity state",
-    description="Gets information about the OT-2's network interfaces "
+    description="Gets information about the robot's network interfaces "
     "including their connectivity, their "
     "addresses, and their networking info",
     response_model=NetworkingStatus,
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
This makes sense to me. I think we should add something to the endpoint description that explicitly says "only use rescan=True based on a user need (i.e. "clicked a scan networks button") and not on a general poll" or something.

robot-server/robot_server/service/legacy/routers/networking.py
```diff
@@ -211,7 +221,7 @@ async def get_eap_options() -> EapOptions:
 
 @router.post(
     "/wifi/disconnect",
-    summary="Disconnect the OT-2 from Wi-Fi",
+    summary="Disconnect the robot from Wi-Fi",
```
SyntaxColoring, line 226: Thanks for this "OT-2" -> "robot" cleanup.

SyntaxColoring (approved):
Agreed with Seth's comment, but the HTTP API otherwise looks good to me. Thank you!

PR #13407, opened 2023-08-28
refactor(protocol-designer, step-generation): adapter support in moveLabware and text cleanup

jerader:
closes RAUT-621, RAUT-642, RAUT-643, RAUT-645, RQA-1441, RQA-1440

# Overview

A bit of PD adapter followup, this PR does 4 things:
1. fixes the `add adapter` text and has `add adapter`, `add labware` or `add adapter and labware`: RAUT-643 & RQA-1441 & RQA-1440
2. fixes copy for deleting adapter: RAUT-645
3. fixes moving a labware on top of the adapter: RAUT-642
4. add labware and location names to error string and dropdown option in Move Labware form: RAUT-621

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_adapter-split-touchups/

Verify that each of the items mentioned in the ticket works:
 1. make a protocol and add labware on top of module, on the deck and on top of adapter. It should have correct copy of either `add adapter`, `add labware` or `add adapter and labware`
 2. delete an adapter from the deck, check the copy of the confirm deletion modal
 3. add a move labware command of moving a labware on/off an adapter. it should work as expected and render on the deck correctly
 4. if you move a labware to a new position that is not compatible, make sure the error string mentions the labware and the new location

# Changelog

- added logic for string for adding labware/adapter 
- in `LabwareLocationField`, added logic for the error string
- wired up the labwareId in newLocation in `forMoveLabware`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/localization/en/deck.json
+++ b/protocol-designer/src/localization/en/deck.json
@@ -1,7 +1,7 @@
 {
   "warning": {
     "gen1multichannel": "No GEN1 8-Channel access",
-    "cancelForSure": "Are you sure you want to permanently delete this adapter?"
+    "cancelForSure": "Are you sure you want to remove this adapter?"
   },
   "blocked_slot": {
     "MODULE_INCOMPATIBLE_SINGLE_LABWARE": "Labware incompatible with this module",
@@ -29,7 +29,8 @@
       "add_labware": "Add Labware",
       "drag_to_new_slot": "Drag To New Slot",
       "place_here": "Place Here",
-      "add_adapter": "Add Adapter"
+      "add_adapter_or_labware": "Add Labware or Adapter",
+      "add_adapter": "Add adapter"
     }
   },
   "inactive_deck": "hover on a step to see deck state"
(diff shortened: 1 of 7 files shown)


protocol-designer/src/localization/en/deck.json
```diff
@@ -1,7 +1,7 @@
 {
   "warning": {
     "gen1multichannel": "No GEN1 8-Channel access",
-    "cancelForSure": "Are you sure you want to permanently delete this adapter?"
+    "cancelForSure": "Are you sure you want to remove this adapter?"
```
koji: question

when deleting a labware, the alert shows the labware name. For adapter, not need to display the adapter name?

koji (approved):
lgtm
tested all cases and worked as expected
I just left a question on the alert for deleting an adapter.

protocol-designer/src/localization/en/deck.json
jerader: good point, i should probably show the adapter name here. I think i was being lazy and didn't wire it up 😅

PR #13410, opened 2023-08-29
docs(api): Opengraph tags for module subpages

ecormany:
# Overview

This PR adds Opengraph tags to the individual module pages that are part of the Python docs. These tags weren't added when these pages were originally created (by splitting apart the Hardware Modules page into several subpages).

# Test Plan

- Check [sandbox](http://sandbox.docs.opentrons.com/docs-module-opengraph-tags/v2/new_modules.html) URLs with a tool like [opengraph.xyz](https://www.opengraph.xyz).

# Changelog

- added `:og:description:` information at the top of each of 7 module subpages

# Review requests

Do these descriptions accurately capture the pages? Will they help human and robot readers?

# Risk assessment

None, docs.

### DIFF
--- a/api/docs/v2/modules/magnetic_module.rst
+++ b/api/docs/v2/modules/magnetic_module.rst
@@ -1,3 +1,5 @@
+:og:description: How to engage and disengage the Magnetic Module for the OT-2 in a Python protocol.
+
 .. _magnetic-module:
 
 ***************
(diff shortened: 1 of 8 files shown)


api/docs/v2/modules/magnetic_module.rst
jwwojak, line 1: I realize this may be simple or obvious, but I like the OT-2 mention in here. Helps reinforce the idea that this module only works with that robot. Just very solid and on point.

api/docs/v2/modules/setup.rst
jwwojak, line 1: There's that double "and" but it can't be avoided without contortions here that might make the description longer.

jwwojak (approved):
Ship it!

2 comments, but nothing that should stop this from moving forward.

api/docs/v2/modules/setup.rst
ecormany: found a better and shorter solution!

PR #13411, opened 2023-08-29
fix(usb-bridge): worker thread for sending data from tcp to usb

fsinapi:
# Overview

The USB bridge seemed to be writing data too fast over USB when pulling logs from the robot server, as the TCP throughput outpaced the app's maximum throughput over USB. Even without that issue, we should really be tolerant of failures to send data over USB. 

This PR adds a worker thread to the TCP/USB bridge to ensure that usb requests are able to be retried indefinitely without clogging handling of the TCP data. The worker thread holds a queue that receives tuples of a serial handle and an array of bytes; this is used because the serial handle is destructed & reconstructed when the USB host unplugs and replugs, so we need to share that information across different threads.

# Test Plan

Test downloading logs over USB before and after this change. It should fail before and succeed after.

I tested successfully on Flexy, and was able to download the logs multiple times in a row from Brent's app-fix branch.

# Changelog

- Add retries when sending data over USB

# Review requests

# Risk assessment


vegano1 (approved):
This makes sense to me

usb-bridge/ot3usb/serial_thread.py
```diff
@@ -0,0 +1,47 @@
...
+        try:
+            sent += serial.write(data[sent:])
+        except Exception as e:
+            # Any exception means we need to quit
+            print(f"Failed to write: {e}")
+            return
+        if sent < len(data):
+            tries += 1
+            # Extremely short sleep to try to avoid battering the CPU
+            print(f"DELAYED: sent {sent} of {len(data)} (try {tries})")
```
sfoster1: should this be a log or are we capturing stdout

sfoster1 (approved):
Looks good as is; can we add some kind of logging if the queue gets excessively many items though? i dunno, maybe every time we add something to the queue and its size is mod 100?

usb-bridge/ot3usb/serial_thread.py
fsinapi: classic case of "I Committed My Debugging Print Statements"

usb-bridge/ot3usb/serial_thread.py
```diff
@@ -0,0 +1,45 @@
...
+
+def _worker(queue: QUEUE_TYPE) -> None:
+    while True:
+        ser, data = queue.get()
+        _try_write_all_data(ser, data)
+
+
+def create_worker_thread() -> Tuple[threading.Thread, QUEUE_TYPE]:
+    """Create a serial worker thread. Returns the comms queue."""
+    queue: QUEUE_TYPE = Queue(0)
```
SyntaxColoring: I'm concerned about using an infinite-size queue here because it will remove [backpressure](https://medium.com/@jayphelps/backpressure-explained-the-flow-of-data-through-software-2350b3e77ce7) from the USB connection.

If the receiving side is unable to handle data at the rate the sending side is throwing at it, I think we want the normal TCP flow control mechanisms to block the sending side from giving it any more.

fsinapi: Yeah a limit on here seems very reasonable! Primarily concerned about memory usage when letting the buffer grow unbounded.

PR #13413, opened 2023-08-29
fix(api,hardware): add retry mechanism for fw update data packets and can updates.

vegano1:
# Overview

We have seen that sometimes a CAN subsystem firmware update will fail when all of the subsystems are being updated at the same time, typically after a system update with a new firmware version. The reason is that the subsystem data ack times out because of all the data being sent over CAN and sometimes we hit the maximum can bandwidth and we miss some messages. [Previously](https://github.com/Opentrons/opentrons/pull/13263), we increased the timeout of the data ACK from 1s -> 5s, but we observed that some ACKs were still coming back more than 20 seconds after the message was sent.

Instead of arbitrarily increasing the date ack's timeout value, let's add a retry mechanism for data messages that will retry a data message up to 3 times before deeming the update as failed.

Closes: [RQA-1425](https://opentrons.atlassian.net/browse/RQA-1425)

# Test Plan

- [x] Make sure CAN subsystem updates no longer fail when one data packet times out
- [x] Make sure we retry a data packet 3 times before declaring the update as failed

# Changelog

- Add a simple retry loop for CAN firmware data messages, so they are resent when they timeout up to N times.

# Review requests
- Is there anything I'm missing?

# Risk assessment
Med, this touches the firmware update mechanism on the Flex so will need to be tested.

[RQA-1425]: https://opentrons.atlassian.net/browse/RQA-1425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
If it fixes it that's good, but I don't like having multiple levels of retry - retrying _both_ individual packets _and_ the download process as a whole. I think having both confuses things a bit, and it's worth committing to only one of them.

PR #13416, opened 2023-08-29
refactor(protocol-designer): OT-2 module default selection set module…

jerader:
… model

closes RAUT-649

# Overview

As outlined in the ticket, you get a white screen on the design tab when you add a temperature or magnetic module to an OT-2 protocol with the default model. This is because on click, we were deleting the model for some reason (this is a legacy component so idk why we needed to have that functionality to begin with). Anyway, this PR fixes that.

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_ot2-module-selection-bug-fix/

Go through the create wizard for an OT-2 protocol, when you get to the Modules page, add a temperature module or magnetic module and keep the default model that is selected. Click the continue button to create the protocol. Then click on the design tab to the left. You should not see a white screen and should work as expected

# Changelog

- delete the `onSetFieldValue(targetToClear, null)` in `handleOnDeckChange` and delete unneeded code in the tests and parent components

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/FilePipettesModal/ModuleFields.tsx
+++ b/protocol-designer/src/components/modals/FilePipettesModal/ModuleFields.tsx
@@ -58,7 +58,6 @@ export interface ModuleFieldsProps {
       }
   values: FormModulesByType
   onFieldChange: (event: React.ChangeEvent) => unknown
-  onSetFieldValue: (field: string, value: string | null) => void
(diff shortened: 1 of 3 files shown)


koji (approved):
tested and didn't see white-screen

Maybe adding `Error boundary` would a good way to avoid showing white-screen to users? 🤔

PR #13417, opened 2023-08-29
fix(api): normal home if fast home stalls

sfoster1:
If a motor axis says it has a valid encoder position then when we home we'll try and update the stepper from the encoder, fast move to near home, and then home. But if the encoder position was wrong - if it missed some overflows, for instance - then quite possibly that first fast motion will fail, which right now would fail the home.

Luckily, the fast move is just an optimization. So let's catch a stall error during the fast move and fall back to a normal home if it happens. This should be very rare, which is good because "detecting a stall in the fast move" really means "the fast move crashed into the side of the robot", but you don't really notice since it's fast.

Closes RQA-1432

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -34,6 +34,7 @@ from opentrons_shared_data.pipette import (
 )
 from opentrons_shared_data.gripper.constants import IDLE_STATE_GRIP_FORCE
 from opentrons_shared_data.robot.dev_types import RobotType
+from opentrons_shared_data.errors.exceptions import StallOrCollisionDetectedError
 
 from opentrons import types as top_types
 from opentrons.config import robot_configs, feature_flags as ff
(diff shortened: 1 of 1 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1288,11 +1289,18 @@ async def _retrieve_home_position() -> Tuple[
...
+                    await self._backend.move(
+                        origin,
+                        moves[0],
+                        MoveStopCondition.none,
+                    )
+                except StallOrCollisionDetectedError:
+                    self._log.warning(
+                        f"Stall on {axis} during fast home, encoder may have missed an overflow"
+                    )
+                    await self._backend.home([axis], self.gantry_load)
```
fsinapi: Do we need this extra call to `_backend.home`? I would think that the call on line 1304 should be enough, just need to catch & log the error so this function doesn't bail out

fsinapi (approved):
I _think_ this would make a stalled axis double-home but catching the stall error sounds really good to me

vegano1 (approved):

api/src/opentrons/hardware_control/ot3api.py
vegano1: Yeah i think you're right, home on 1304 always happens, nice catch

sfoster1: Oh yup I totally missed that I think I only saw the one in the other branch of the `else`

PR #13420, opened 2023-08-29
chore(api): remove spammy log

sfoster1:
This happens a _lot_ now and it is meaningless most of the time so only log when it's meaningful.


vegano1 (approved):
yep

fsinapi (approved):
🙏 🙏 🙏

PR #13421, opened 2023-08-29
fix(api): base plunger speeds on working volume

sfoster1:
When 62962c599b8b71bccc4c6b18a59a798f8b8989a3 added "liquid classes", we had to change the way we look up various volumes, and we used the liquid class max volume as the volume to look up ul/mm transforms for determining speeds. The thing is, the liquid class max volume didn't depend on the tip type, but the ul/mm transform did - so if you were using a 50ul tip on a p1000, you'd try and get the ul/mm for 1000ul on a 50ul tip and get an error.

We can use the working volume instead, which is the maximum volume of the tip, and fix this up but I think we need a better fix for this that moves volume constraints into tip definitions.

This change is only for the flex, for two reasons. First, the per-tip configurations on the OT-2 have ul/mm tables that go up to the pipette max volume on all tips, so it's not needed; and second, this change would actually (extremely slightly) change the flowrates on the non-default tips on old api versions, so we can't do it.

## Review requests
- does this look right? It feels a bit awkward, `working_volume` is this sort of weird add-on from the gen2 pipette days and I think we need to promote max and min volumes to properties of tips

## Testing and risk

This is in pretty core code, but it's also fixing a straightforward bug. We should run a protocol that uses various kinds of tips - i.e. 200ul tips on a 1000, other stuff like that - and check that (a) we don't get exceptions and (b) the flow rates match the configs.

Did the test and it all looks good.

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot2/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot2/pipette.py
@@ -125,15 +125,7 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
             pip_types.PipetteTipType(self._liquid_class.max_volume)
         ]
         self._fallback_tip_length = self._active_tip_settings.default_tip_length
-        self._aspirate_flow_rates_lookup = (
-            self._active_tip_settings.default_aspirate_flowrate.values_by_api_level
-        )
-        self._dispense_flow_rates_lookup = (
-            self._active_tip_settings.default_dispense_flowrate.values_by_api_level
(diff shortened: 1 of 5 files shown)


TamarZanzouri (commented):
Changes LGTM! there is old api tests that are failing for OT-2. I guess the value is different now. once that is fixed its good to merge but yes I think we need to find a better solution for this.

Laura-Danielle (approved):
Yes this fix makes sense to me.

PR #13422, opened 2023-08-29
feat(api): implement push_out arg in the hardware controller

TamarZanzouri:
# Overview

implement hardware controller push_out arg in the dispense command.
closes https://opentrons.atlassian.net/browse/RET-1373.

# Test Plan

- make sure you are able to use python protocols with a push_out arg in the dispense command and make sure it is pushing out the argument you supplied or the default arg specified in the pipette config. 
- same for posting a dispense command.
- same for JSON protocols.
- make sure Gen1 and Gen2 pipettes do not have this additional push_out it should use a default of 0 if not specified. 

# Changelog

- added a push_out prop to `Pipette` 
- added a method to check if the `push_out` volume supplied is not larger than the max volume allowed for blow out.
- `plan_check_dispense` now fetches the default volume if the push_out arg is not supplied. and changes plunger position to move to according to that.  
- added `pushOut` to `protocol_schema_v7.Params`

# Review requests

1. am I missing logic?
2. does `ok_to_push_out` logic make sense?
3. does it make sense to update the plunger dist this way? 

# Risk assessment

Medium. added an additional push_out force for the FLEX pipettes. should not affect OT-2.


api/src/opentrons/hardware_control/instruments/ot2/pipette.py
```diff
@@ -443,6 +447,11 @@ def working_volume(self, tip_volume: float) -> None:
         self._active_tip_settings = self.liquid_class.supported_tips[tip_size_type]
         self._fallback_tip_length = self._active_tip_settings.default_tip_length
 
+    @property
+    def minimum_volume(self) -> float:
+        """The smallest controllable volume the pipette can handle."""
```
SyntaxColoring: If I'm reading this code correctly, this is the smallest controllable volume the pipette can handle *in its current mode* / *given its current liquid class.* Can we clarify that in this docstring? Otherwise, it sounds like the minimum volume under any circumstances.

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -442,6 +444,11 @@ def working_volume(self, tip_volume: float) -> None:
         self._fallback_tip_length = self._active_tip_settings.default_tip_length
         self._tip_overlap_lookup = self.liquid_class.tip_overlap_dictionary
 
+    @property
+    def minimum_volume(self) -> float:
+        """The smallest controllable volume the pipette can handle."""
```
SyntaxColoring: Ditto my other comment about clarifying that this is for the current mode / liquid class.

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -463,6 +470,11 @@ def remove_current_volume(self, volume_incr: float) -> None:
     def ok_to_add_volume(self, volume_incr: float) -> bool:
         return self.current_volume + volume_incr <= self.working_volume
 
+    def ok_to_push_out(self, push_out_dist_mm: float) -> bool:
+        return push_out_dist_mm <= (
+            self.plunger_positions.blow_out - self.plunger_positions.bottom
+        )
+
```
SyntaxColoring, line 477: Is there a sign error here?

`self.plunger_positions.blow_out < self.plunger_positions.bottom` (I think?)

so `self.plunger_positions.blow_out - self.plunger_positions.bottom < 0`

so `push_out_dist_mm` will never be `<` than that

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
```diff
@@ -464,6 +473,11 @@ def remove_current_volume(self, volume_incr: float) -> None:
     def ok_to_add_volume(self, volume_incr: float) -> bool:
         return self.current_volume + volume_incr <= self.working_volume
 
+    def ok_to_push_out(self, push_out_dist_mm: float) -> bool:
+        return push_out_dist_mm <= (
+            self.plunger_positions.bottom - self.plunger_positions.blow_out
+        )
+
```
SyntaxColoring, line 480: Ditto my other comment about the possible sign error?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -437,13 +438,11 @@ def critical_point_for(
         If `cp_override` is specified, and that critical point actually exists,
         it will be used instead. Invalid `cp_override`s are ignored.
         """
-        pip = self._attached_instruments[mount]
+        pip = self._attached_instruments[OT3Mount.from_mount(mount)]
```
SyntaxColoring, line 446: Is this doing anything? `mount` is already an `OT3Mount`.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -437,13 +438,11 @@ def critical_point_for(
...
         if pip is not None and cp_override != CriticalPoint.MOUNT:
             return pip.critical_point(cp_override)
         else:
-            # This offset is required because the motor driver coordinate system is
-            # configured such that the end of a p300 single gen1's tip is 0.
-            return top_types.Point(0, 0, 30)
+            return top_types.Point(0, 0, 0)
```
SyntaxColoring, line 450: Why is it okay to remove this?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -460,7 +459,7 @@ def plunger_position(
         self, instr: Pipette, ul: float, action: "UlPerMmAction"
     ) -> float:
         mm = ul / instr.ul_per_mm(ul, action)
-        position = mm + instr.plunger_positions.bottom
+        position = instr.plunger_positions.bottom - mm
```
SyntaxColoring, line 467: What does this do? If you're aspirating 10 µL from bottom, I would have expected the returned `position` to be `some_positive_number + instr.plunger_positions.bottom`. Are we inverting the polarity of `plunger_position()`?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -576,10 +576,38 @@ def plan_check_dispense(
 
         # Ensure we don't dispense more than the current volume
         disp_vol = min(instrument.current_volume, disp_vol)
+        is_full_dispense = abs(instrument.current_volume - disp_vol) < (
+            instrument.minimum_volume / 10
+        )
```
SyntaxColoring: Hmm where does this logic come from? Is this to catch rounding errors, or something else?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -545,6 +544,7 @@ def plan_check_dispense(
         mount: OT3Mount,
         volume: Optional[float],
         rate: float,
+        push_out: Optional[float],
```
SyntaxColoring, line 552: Is there a semantic distinction beween `push_out=0` and `push_out=None`?

Instead of `Optional[float]`s, should we just make these `float`s? Possibly defaulting to `=0`?

It looks like all the logic is already treating `None` as equivalent to `0`, e.g. `if not is_full_dispense and push_out`.

SyntaxColoring (commented):
Various questions as I make my way through this. I see commits are still being pushed, so sorry if any of this is already undergoing revision!

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
sfoster1: no, `blow_out > bottom` on the flex, where home (top) is 0(ish) and positions increase from there

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
sfoster1: ditto response

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: probably not - legacy before getting rid of the typevar

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: So the reason I made this change (and the below change) is exactly because this was weird and confusing.

The way this code originally worked is that there was a `PipetteHandlerProvider` that had all the ot-2 behavior. 99% of that was shared with the flex, but specifically `plunger_position()` and `critical_point_for()` were overridden for the flex machine coordinate system, where we always home to 0 and move positive from there. Then, at some point, someone wanted the flex and ot2 code to be fully separate, and they did this by copying the shared code into both places, and copying the flex override into the flex side.

So this function before this change actually never ran, ditto `plunger_position` - because the flex always created the child class that used to be at the bottom of this file, and that overrides both methods.

The new behavior is the content of those overriding functions copied into the main class, which is now named something different, to prevent people thinking the ot2 behavior was what the flex was using.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: see above, and generally check out e.g. `shared-data/pipettes/definitions/2/general/single_channel/p50/3_3.json` - flex machine coordinates have 0 as home and positive positions from there so we don't have to do zero crossing math in the stepper controllers due to laziness, so in flex machine coordinates `top > bottom > blow_out > drop_tip`

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: yeah pretty much, and it's more semantically relevant. I could make it an approx check.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: Oof you're totally right, and there is a difference, and I need to change that.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
SyntaxColoring: Yeah, I guess I'm worried about the edge case if ever `instrument.minimum_volume == 0`.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
SyntaxColoring: Aha. Thanks for that cleanup and the explanation. This seems much better.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: that's a very good point!

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -576,10 +577,39 @@ def plan_check_dispense(
...
         if disp_vol == 0:
             return None
 
+        if is_full_dispense:
+            if push_out is None:
+                push_out_ul = instrument.push_out_volume
+            else:
+                push_out_ul = push_out
+        else:
+            if push_out is not None and push_out != 0:
```
sfoster1, line 595: @SyntaxColoring does this code help with your concerns about the push_out tristating

api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py
```diff
@@ -589,6 +595,34 @@ def plan_check_dispense(  # type: ignore[no-untyped-def]
...
+            raise CommandPreconditionViolated(
+                message="Cannot push_out on a dispense that does not leave the pipette empty",
+                detail={
+                    "command": "dispense",
+                    "remaining-volume": instrument.current_volume - disp_vol,
+                },
+            )
+        if push_out is None and not is_full_dispense:
+            push_out_ul = instrument.push_out_volume
+        elif push_out and not is_full_dispense:
```
ahiuchingau: Should this actually be `elif push_out and is_full_dispense`?

ahiuchingau (commented):
Left a comment on the OT2 pipette handler push out volume logic – seems like if we follow the changes for the Flex part then we should be good. 

And this is more of a stylistic question – it seems like we're allowing people to pass in a negative push out volume, the `ok_to_push_out` method would stop us from actually pushing out the wrong value but should we raise a proper error for that? Ignore me if this would have already been caught on the protocol level.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: @SyntaxColoring After thinking about this more I think the right thing to do is just make this logic `<=` or change it to a full approx check but keep it linked to minimum_volume. A minimum_volume of 0 means the pipette can handle arbitrarily small volumes and therefore it would not be correct to push_out unless it was exactly empty.

api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py
sfoster1: yeah, you're right, fixed

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
SyntaxColoring: Mmmmm I think I'm still confused, sorry.

The line `if push_out is not None and push_out != 0` makes it clear that this code is handling the case where `push_out=None` *and* the case where `push_out=0`. But from the caller's perspective, what's the difference between `push_out=None` and `push_out=0`? When should we use one or the other?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
SyntaxColoring: > A minimum_volume of 0 means the pipette can handle arbitrarily small volumes and therefore it would not be correct to push_out unless it was exactly empty.

If I do this:

1. aspirate 0.3
2. dispense 0.1
3. dispense 0.1
4. dispense 0.1

I would intuitively consider the pipette "exactly empty," though floating point arithmetic would think it has -2.7755575615628914e-17 µL left.

Are you saying if the pipette has a `minimum_volume` of 0, it only ever makes sense to provide a `push_out` volume of 0?

Or are you saying if the pipette has a `minimum_volume` of 0, it only makes sense to provide a `push_out` volume if the pipette's remaining volume is 0? Because if it's this one, I'm not sure that will reliably happen with the code as implemented, given floating point rounding situations like the one described above.

sfoster1: yeah i just made it `isclose`

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: `push_out=None` when you want to use the default push out, whatever it is
`push_out=0` when you want to not push out

sfoster1: also though this comment is tagged on line 593, it really applies to the span 587-601 inclusive. The logic is as follows:
- If you don't specify `push_out` (`push_out=None`) then you get the default behavior, which is definitely no pushout if this is not a full dispense and maybe some pushout if it is a full dispense
- If you do specify `push_out` to a non-zero value, then it will get executed if it is a full dispense and error if it is not a full dispense
- If you specify `push_out` to a zero value, then there will be zero pushout (which is a trivial case if it is not a full dispense)

ahiuchingau (approved):
Looks good to me!

SyntaxColoring (approved):
Thanks!

PR #11536, opened 2022-10-05
test(robot-server): Snapshot-test the SQL statements that we use to create tables

SyntaxColoring:
This work is part of RSS-98.

# Overview

[We use SQLAlchemy to define our database schema in Python](
https://github.com/Opentrons/opentrons/blob/2d8ef75d2e968850bfca23153523365058c4d058/robot-server/robot_server/persistence/tables.py#L5-L109).

When `robot-server` boots for the first time, SQLAlchemy automatically compiles these Python definitions into raw `CREATE TABLE` SQL statements and sends them to our database.

For example, SQLAlchemy turns this:

```python
protocol_table = sqlalchemy.Table(
    "protocol",
    _metadata,
    sqlalchemy.Column("id", sqlalchemy.String, primary_key=True),
    sqlalchemy.Column("created_at", UTCDateTime, nullable=False),
    sqlalchemy.Column("protocol_key", sqlalchemy.String, nullable=True),
)
```

Into this:

```
CREATE TABLE protocol (
    id VARCHAR NOT NULL,
    created_at DATETIME NOT NULL,
    protocol_key VARCHAR,
    PRIMARY KEY (id)
)
```

This PR adds a snapshot test for these raw `CREATE TABLE` SQL statements. Effectively, this is a snapshot test for our database schema at the raw SQL level. The initial snapshot is from release v6.0.0.

# Motivation

Sometimes, we want to refactor our Python-side table definitions without changing the SQL-level database schema. For example, PR #11006 made a new Python type to access the raw SQL data in a more convenient way.

Without a test like this, accidental schema changes are hard to detect, because they don't always cause immediate problems. For example, a column's [type affinity](https://www.sqlite.org/datatype3.html) might change, or a constraint might be renamed.

# Review requests

* Is the purpose of this test understandable from the in-code comments?
* Does this seem maintainable?

# Risk assessment

No risk to production code. Changes are only to our test suite.

PR #11448, opened 2022-09-08
test(robot-server): Test compatibility with databases created in v6.0

SyntaxColoring:
# Overview

This adds a basic test for persistence backwards compatibility, which includes backwards compatibility in our SQL database. The test makes sure that the current `robot-server` code can read some resources that were created by a v6.0.1 `robot-server`.

Closes RSS-102.

This is a humongous diff—sorry! Most of it is opaque fixture data. The actual code changes are fairly small.

# Changelog

* Copy a pre-populated persistence directory into the repository, intended to be used as an opaque test fixture. I created it by manually issuing HTTP requests to a v6.0.1 dev server and saving its persistence directory. See the added readme for details.
* Add a test that runs a dev server underpinned by that persistence directory, issues a bunch of HTTP `GET` requests to it, and makes sure none of the requests come back with an obvious error.

  For simplicity, we only check the responses in a *very shallow* way. Basically, we just check that the HTTP status is successful, and that the response data isn't empty. I believe this is sufficient to catch the bugs that we've been prone to so far.

# Review requests

* Will these tests give us sufficient confidence to rework our persistence code according to the plan in RSS-98?
* Try making a change in `robot-server`, `api`, or `shared-data` that would break backwards compatibility. For example, try making an `Optional` field in a persisted Pydantic model non-`Optional`. Then, run `make -C robot-server test`—does it catch the problem?
    * They definitely catch the bug that was fixed by #11422, so that's cool.
* Do you have any ideas for future improvements? If so, please record them in RSS-100.

# Risk assessment

No risk to production code. Changes are only to tests.

PR #10309, opened 2022-05-17
refactor(robot-server): add basic SQL migrations to persistence layer

mcous:
## Overview

This PR adds a basic SQL migration system to the robot-server's protocol/run persistence layer. Since the database has changed slightly since the 5.1.0-beta.0 release, **this PR is necessary for the robot server to continue functioning** if you've installed that version or any copy of `edge` from after SQL was added but before 2022-05-17.

#10187 was merged, and it (intentionally) included some database schema changes. This left us with a few options to make sure we didn't break the robot server for too long:

1. Force anyone with a previous (unreleased to prod) version of the SQL database to wipe it out before using the latest version
    - Ruled this out because #10238 is not yet completed, and `5.1.0-beta.0` is live
2. Make edits to #10187 to not modify the existing tables
    - Ruled this would've made the persistence logic less robust
3. Add a migration system to #10187 
    - Ruled this out because the PR was massive and we'd already been through a big review cycle
4. Merge #10187, then add a migration system in a new PR
    - AKA this PR

## Changelog

- add basic SQL migrations to persistence layer
- tweak a column name to better match what we came up with during review of #10187, since we're in here adding migrations anyway

## Review requests

Hopefully the code is at least a little self-evident, but I;ll be working on spinning up some docs about this system. This PR adds a `migration` table, that is used track any time the SQL database is migrated to a new version. New versions are necessitated by column changes to existing tables. These migration capabilities were driven entirely by what was easy / exactly what we need for #10187.

| change | needs migration | notes
| --- | --- | --- |
| new table | no | SQLAlchemy will add the table if it's not there |
| new column | yes | Must be nullable or have a default value |
| change column type | yes, but | **We can't / shouldn't do this** |
| drop column | yes, but | **We can't / shouldn't do this** |
| drop table | yes, but | **This is a pretty bad idea** |

In the past, we talked about using Alembic for SQL migrations, and I hope this PR doesn't close that door if we ever need more sophisticated migration capabilities. For where we're at, though, (and after talking with @TamarZanzouri), Alembic felt really heavy weight for our needs.

Instead, this PR implements a really lightweight, up-migration-only system that simply executes raw SQL strings to add columns. It gets the job done, especially considering SQLAlchemy offers no `ALTER TABLE` API that I could find.

Please also look at the unit tests and let me know what you think.

### How migration works

Migration is run at SQLAlchemy `Engine` creation, after tables have been added.

1. Find the current version 
[...]

PR #13425, opened 2023-08-30
perf(app,robot-server): Download analyses as raw JSON documents

SyntaxColoring:
# Overview

This dramatically improves the load time for certain views in the Opentrons App and Flex on-device display.

For example, under the circumstances described in RSS-328, it reduces load time of the Flex's home page from ~5 minutes to <5 seconds.

This closes RSS-328 and RSS-160. It probably closes some other performance-related tickets. It also helps a little bit with RSS-98. 

# Architecture

Protocol analysis on the robot is a major known compute bottleneck for us. One of the problems is the overhead of storing the completed analyses in the database and then extracting them. It goes through several layers of translation:

```mermaid
flowchart LR
    subgraph robot-server
        analysis_engine((Analysis engine))
        subgraph SQL database
            bytes
        end
        dict
        pydantic[Pydantic object]
        fastapi((FastAPI endpoint))
        json_str
    end
    analysis_engine --> pydantic
    bytes <-->|pickle| dict
    dict <--> pydantic
    pydantic --> fastapi --> json_str[JSON string] --> to_client((HTTP client))
```

This scheme adds a *lot* of overhead. Creating Pydantic objects is especially slow. (This scheme is also very brittle—see RSS-98.)

To improve this, this PR adds a second path that's basically a direct read from the database:

```mermaid
flowchart LR
    subgraph robot-server
        analysis_engine((Analysis engine))
        subgraph SQL database
            bytes
            json_str_db[JSON string]
        end
        dict
        pydantic[Pydantic object]
        old_fastapi((FastAPI endpoint))
        new_fastapi((FastAPI endpoint))
        json_str
    end
    analysis_engine --> pydantic
    bytes <-->|pickle| dict
    dict <--> pydantic
    pydantic --> old_fastapi --> json_str[JSON string] --> to_client((HTTP client))
    json_str_db <--- pydantic
    json_str_db --> new_fastapi --> to_client
    linkStyle 7 stroke:red
    linkStyle 8 stroke:red
```

We expose this through an experimental new HTTP endpoint, `GET /protocols/:id/analyses/:id/asDocument`. We leave the existing endpoints, such as `GET /protocols/:id/analyses/:id`, untouched for now.

This requires a database migration to add a new column for the JSON string, so we do that.

# Detailed changelog

## Server

* Add a new endpoint. Document it as experimental.
* Add a new column to the `analysis` table: `completed_analysis_as_document`, which stores the serialized JSON as a `VARCHAR`.
* Introduce a new database migration to support the new column.
    * The migration includes copying over all existing data, so the new column should never be `NULL`.
    * Per previous discussion in RSS-130, we report an in-progress migration with an HTTP 503 status code on `GET /health` and on the endpoints that need the database.
    * Expected migration times:
        * Post-release Flexes: None, if we get this in for v7.0.0.
        * Internal Flexes: 5 minutes, from testing with the data in RSS-328.
        * OT-2s: 1 minute on average (from my old notes in RSS-130), or 10 minutes if heavily-loaded (extrapolating from RSS-328). There's some room for optimizing this.

## Desktop app and ODD

@b-cooper can provide more details, but roughly, we're switching a bunch of stuff to query the new fast endpoint and avoid querying the old slow endpoints.

Helpers for the new `GET /protocols/:protocolId/analyses/:analysisId/asDocument` endpoint have been added `api-client` and `react-api-client`.  The new `useProtocolAnalysisAsDocumentQuery` has been substituted across all locations in the Desktop and ODD where we were formerly requesting `GET /protocols/:protocolId/analyses` (old non-performant un-pickling path) and picking the latest entry.
 
# Test Plan

@SyntaxColoring tested on an OT-2 and Flex:

* [x] It should be faster. :) See RSS-328 for one test case.
* [x] A Flex's on-device display should say "initializing..." while the migration is in progress.
* [x] Per prior discussion in RSS-130, the desktop app may say something like "This robot's API server is not responding correctly to requests..." while the migration is in progress. This is potentially confusing, but it's what it's historically shown while the robot boots up. The only difference now is that bootup is taking longer.
* [x] Power-cycle a robot in the middle of the migration and make sure it doesn't cause any errors. The migration should start from scratch.
* [x] You should be able to freely upgrade and downgrade across this commit without anything breaking.

# Review requests

Architecturally, is there anything you think we need to do now in order to make this less risky, or avoid headaches in the long run?

# Risk assessment

Medium.

* Supporting both strategies simultaneously adds complexity to the server, which will be bug-prone if we let it linger.

* Any kind of robot-server database migration with our homegrown system is inherently kind of scary, and this is the first real one that we're doing in production.
  
  We have good automated tests for it (e.g. [test_tables.py](https://github.com/Opentrons/opentrons/blob/663f6d83fb13e2fc4411a98648731bc5f5f1bb46/robot-server/tests/persistence/test_tables.py) and [test_persistence.py](https://github.com/Opentrons/opentrons/blob/663f6d83fb13e2fc4411a98648731bc5f5f1bb46/robot-server/tests/integration/http_api/persistence/test_compatibility.py)), but things can still sneak in. Examples:

    * The transaction bug fixed in #13424
    * [This thing about a mismatched column type](https://github.com/Opentrons/opentrons/blob/663f6d83fb13e2fc4411a98648731bc5f5f1bb46/robot-server/robot_server/persistence/_migrations.py#L156-L160)
    * [This quirk in how we stamp the database with schema versions](https://opentrons.slack.com/archives/CSPMY9048/p1693446038126749)


app/src/organisms/TakeoverModal/MaintenanceRunTakeover.tsx
```diff
@@ -6,6 +6,7 @@ import {
 import { TakeoverModal } from './TakeoverModal'
 import { TakeoverModalContext } from './TakeoverModalContext'
 
+const MAINTENANCE_RUN_POLL_MS = 10000
```
b-cooper, line 9: Backing off this refetch interval to a slower poll.  There are separate interval instantiated within wizards where we need more up to date info

robot-server/tests/persistence/test_migrations.py
SyntaxColoring, line 1: I'm deleting this file because these tests were difficult to update for new database schemas, and I don't think they were helpful.

These were old tests, added in #10309 when the database was quite new. Since then, we've added:

* Integration tests with snapshots of old persistence directories. PR #11448
* Safeties against accidentally changing SQL schema. PR #11536

Combined, I think those newer approaches supersede this old one.

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -193,6 +193,26 @@ async def get(self, analysis_id: str) -> ProtocolAnalysis:
...
+            AnalysisNotFoundError
+        """
+        pending_analysis = self._pending_store.get(analysis_id=analysis_id)
+        completed_analysis_document = await self._completed_store.get_by_id_as_document(
+            analysis_id=analysis_id
+        )
+
+        if pending_analysis is not None:
+            # TODO: Probably want this to be a separate HTTP status code so the pending case
+            # doesn't have to pollute the completed analysis schema.
```
SyntaxColoring: Reminder to self to resolve this.

vegano1:
Sweet!

robot-server/tests/protocols/test_protocols_router.py
```diff
@@ -663,3 +664,70 @@ async def test_get_protocol_analysis_by_id_analysis_not_found(
...
+    assert exc_info.value.status_code == 404
+    assert exc_info.value.content["errors"][0]["id"] == "ProtocolNotFound"
+
+
+async def test_get_protocol_analysis_as_document_analysis_not_found(
+    decoy: Decoy,
+    protocol_store: ProtocolStore,
+    analysis_store: AnalysisStore,
+) -> None:
+    """It should get a single full analysis by ID."""
```
jbleon95:
```suggestion
    """It should 404 if the analysis does not exist."""
```

jbleon95 (approved):
Reviewed part of PR in live review, Python side looks good to me.

robot-server/tests/protocols/test_protocols_router.py
SyntaxColoring: 🤦 Oops. Thanks for your attention to detail.

smb2268 (approved):
All of the JS code changes look sound to me. I put this branch on app&ui bot and everything loaded very quickly!

PR #13426, opened 2023-08-30
fix(app): add extra protection from run 404 on ODD

smb2268:
fix RQA-1424, RQA-1421

# Overview
Button disabling for wizard flows wasn't being reflected on ODD, so users were able to click continue and issue a command before the run was created.

# Test Plan
Try to run gripper and pipette flows from ODD, see that the button is disabled when the flow first mounts and then becomes enabled. You should never see a 404 error for a non-existent run

# Changelog
Disable button on ODD when disabled on app
Disable buttons if maintenance run id isn't the run created within the flow as an extra layer of protection

# Review requests
Any ideas on how to fix the Pipette flow test? The issue is that we now need the `createdRunId` to exist within the test instance and I'm having trouble getting that to work - it does not work to resolve `mockCreateMaintenanceRun` since the call happens in a child component.. If there's no quick fix I'm tempted to delete this test file for now since we have pretty robust coverage on all of the children.

# Risk assessment
Low


b-cooper (approved):
Code looks good 🦦

b-cooper:
It looks like you'd have to make the mock for `createMaintenanceRun` more complex (or not mock it at all) in order to ensure the state get set properly.  I agree that this may be pointing to the fact that this test is a little overkill in the face of coverage lower down.

PR #13427, opened 2023-08-30
refactor(shared-data,hardware-testing): remove calibration adapter and logic for magnetic module.

vegano1:
# Overview

The magnetic module is not supported on the Flex, and although we do support the magnetic block which is what the `opentrons_calibration_adapter_magnetic_module` definition was actually for, we don't perform a calibration on the magnetic block since it sits on top of the deck. So let's remove the definition and hardware-testing scripts that use the magnetic module.
# Test Plan

# Changelog

- Remove `opentrons_calibration_adapter_magnetic_module` definition
- Remove magneticModule usage in module calibration test script

# Risk assessment

Low, unused code


sfoster1 (approved):
noice

PR #13428, opened 2023-08-30
feat(engine): disallow gripper movement for combination adapter labware

jbleon95:
# Overview

Closes RLAB-363.

This PR makes it so that combination labware/adapter labware cannot be moved with the gripper. This should not be done in the first place since the gripper does not have logic to correctly pick up and place down these labware, and using the split adapter and labware definitions should be encouraged. Manual moves are still allowed.

In order to accomplish this, a new labware quirk `gripperIncompatible` was added to all combination adapter/labware definitions and a check in the PE `moveLabware` execution was added. Since gripper movement will be new to api level 2.15 and the addition of the new quirk will not change previous behavior, the labware definitions were not bumped up to a new version.

# Test Plan

Tested analysis with the following protocol

```
metadata = {
    'protocolName': 'PAPI Gripper incompatible',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.15"
}


def run(context):
    heater_shaker_1 = context.load_module("heaterShakerModuleV1", location="D1")
    heater_shaker_1.open_labware_latch()

    heater_shaker_2 = context.load_module("heaterShakerModuleV1", location="D3")
    heater_shaker_2.open_labware_latch()

    pcr_adapter = heater_shaker_1.load_adapter("opentrons_96_pcr_adapter")
    armadillo_plate = pcr_adapter.load_labware("armadillo_96_wellplate_200ul_pcr_full_skirt")

    plate_with_adapter = heater_shaker_2.load_labware("opentrons_96_pcr_adapter_armadillo_wellplate_200ul")

    # This move should work
    context.move_labware(armadillo_plate, new_location="C1", use_gripper=True)

    # If commented back in this should fail
    #context.move_labware(plate_with_adapter, new_location="C3", use_gripper=True)

    # This move should work since it is manual
    context.move_labware(plate_with_adapter, new_location="C3", use_gripper=False)
```

# Changelog

- Added `gripperIncompatible` to combination adapter/labware definitions
- Added check for the quirk in `moveLabware` execution if using the gripper, which if found will raise an error.

# Review requests

Mostly the error text for this new check, want to make it clear to users that they should be using the separate adapter and labware loads if they want to use the gripper, otherwise use manual movement.

# Risk assessment

Codewise low, it's a small change and contained only the relevant definitions and `moveLabware` with a gripper, but this change should be made clear since it will cause protocols that were previously passing to fail (although they were incorrect in the sense that the gripper would not function as intended during execution).

### DIFF
--- a/shared-data/labware/definitions/2/opentrons_24_aluminumblock_generic_2ml_screwcap/2.json
+++ b/shared-data/labware/definitions/2/opentrons_24_aluminumblock_generic_2ml_screwcap/2.json
@@ -23,6 +23,7 @@
   },
   "parameters": {
     "format": "irregular",
+    "quirks": ["gripperIncompatible"],
     "isTiprack": false,
     "isMagneticModuleCompatible": false,
     "loadName": "opentrons_24_aluminumblock_generic_2ml_screwcap"
--- a/api/src/opentrons/protocol_engine/commands/move_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/move_labware.py
@@ -125,7 +125,7 @@ class MoveLabwareImplementation(
                     message="Labware movement using a gripper is not supported on the OT-2",
                     details={"strategy": params.strategy},
                 )
-            if labware_validation.validate_gripper_compatible(
+            if not labware_validation.validate_gripper_compatible(
                 current_labware_definition
             ):
                 raise LabwareMovementNotAllowedError(
@@ -137,7 +137,7 @@ class MoveLabwareImplementation(
                 current_labware_definition
             ):
                 raise LabwareMovementNotAllowedError(
-                    f"Cannot move adapter {params.labwareId} with gripper."
+                    f"Cannot move adapter '{current_labware_definition.parameters.loadName}' with gripper."
                 )
 
             validated_current_loc = (
--- a/api/src/opentrons/protocol_engine/commands/move_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/move_labware.py
@@ -125,6 +125,14 @@ class MoveLabwareImplementation(
                     message="Labware movement using a gripper is not supported on the OT-2",
                     details={"strategy": params.strategy},
                 )
+            if labware_validation.validate_gripper_compatible(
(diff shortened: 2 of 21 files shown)


ecormany (commented):
Error text looks good :shipit: 

As for the behavior, I'm curious about the final line of the test protocol. Will LPC let you check multiple combo labware on the same H-S? If no, it should probably error and we should force you to end your protocol instead of _unscrewing_ the H-S adapter and replacing it mid-run.

In general, any manual intervention that says "hey go pick up that aluminum block" without being wired into the module status reporting whether it's ⚠️ hot to the touch should be disallowed.

shared-data/labware/definitions/2/opentrons_24_aluminumblock_generic_2ml_screwcap/2.json
SyntaxColoring, line 1: No opinion from me either way, but are we sure we want to disallow gripper movements for these labware that represent individual tubes or PCR strips inside aluminum blocks? These combined definitions are the only way to load them; there is no separate "just 2ml screwcap tubes" definition.

SyntaxColoring (approved):
Implementation looks great. Thanks!

SyntaxColoring:
> Will LPC let you check multiple combo labware on the same H-S?

I haven't tested this, but I suspect it doesn't have any special handling for when labware adapters need to be switched out.

> If no, it should probably error and we should force you to end your protocol instead of unscrewing the H-S adapter and replacing it mid-run.
> 
> In general, any manual intervention that says "hey go pick up that aluminum block" without being wired into the module status reporting whether it's ⚠️ hot to the touch should be disallowed.

I agree that this is something we should tell the protocol author or operator about, but I worry about disallowing it and getting in the way of people who know what they're doing. This seems like a good case for nonfatal warnings, which Python protocols currently have no way of surfacing. I've recorded this in https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3794108438/Things+that+would+make+good+nonfatal+warnings. If we accumulate enough of them, maybe we can push for that to be prioritized.

PR #13432, opened 2023-08-31
fix(usb-bridge): use correct USB gadget configuration strings

fsinapi:
# Overview

- Use the serial number on the filesystem, and default to a serial number that follows our serial pattern
- Call the devices a Flex instead of OT3 in the public-facing device information

# Test Plan

Looked at the USB info with System Information.

* Uploaded to a robot that hadn't had the serial flashed, it shows up with the default serial
* Uploaded to a robot that _does_ have a serial number, it shows up as that serial number
* Both robots show up as a Flex instead of OT3

# Changelog

- Use the serial number on the filesystem, and default to a serial number that follows our serial pattern
- Call the devices a Flex instead of OT3 in the public-facing device information

# Review requests

# Risk assessment

### DIFF
--- a/usb-bridge/ot3usb/__main__.py
+++ b/usb-bridge/ot3usb/__main__.py
@@ -4,7 +4,8 @@ import logging
 import time
 from typing import NoReturn
 
-from . import cli, usb_config, default_config, usb_monitor, tcp_conn, listener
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Looks good to me!

vegano1 (approved):
Sick!
Very good 👍

PR #13433, opened 2023-08-31
fix(app): remove hover from empty state in slideouts

b-cooper:
# Overview
 
Remove gratuitous hover border color change from the empty state of the protocol and robot slideouts.

Closes RQA-1247

# Risk assessment
low


jerader (approved):
🎨

PR #13435, opened 2023-08-31
feat(protocol-designer, step-generation): fix labware & Heater-shaker...

jerader:
… errorCreator

closes RQA-1455

# Overview

Fix the 3 bugs listed in the ticket related to errors Rob saw during protocol analysis

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_error-creator-tweaks/

1. make a new protocol (either OT-2 or Flex) in PD, add a heater-shaker, add a `Opentrons 96 Deep Well Adapter` on top of the heater-shaker, when adding a labware on top of the adapter, make sure the only option is `NEST 96 Deep well plate 2ML`. Add that labware

2. try to add a transfer step into the labware on the Heater-shaker. There should be a timeline error saying that the labware latch needs to be closed before pipetting into the labware

3. delete the transfer step and add a move labware step to the labware on the heater-shaker. There should be a timeline error saying that the labware latch needs to be open before moving the labware (NOTE: this bug was actually addressed in a different PR but just mentioning it here since Rob documented it in the ticket)

# Changelog

- change the labware that is compatible with the 96 deep well adapter
- tweak `pipetteIntoHeaterShakerLatchOpen` to account for the labware being on top of an adapter

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/utils/labwareModuleCompatibility.ts
+++ b/protocol-designer/src/utils/labwareModuleCompatibility.ts
@@ -87,9 +87,7 @@ export const COMPATIBLE_LABWARE_ALLOWLIST_FOR_ADAPTER: Record<
   string,
   string[]
 > = {
-  [DEEP_WELL_ADAPTER_LOADNAME]: [
-    'opentrons/nest_96_wellplate_100ul_pcr_full_skirt/2',
-  ],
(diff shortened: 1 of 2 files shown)


koji (approved):
tested without heater-shaker action and with heater-shaker action.
both of cases worked as expected.

PR #13440, opened 2023-08-31
fix(app): fix run setup pipette bugs

smb2268:
fix RQA-1428, RQA-1433

# Overview

We were grabbing pipette compatibility for run setup from the `/pipettes` endpoint which doesn't poll as often as `/instruments`. This was causing a delay in registering that the pipette is now attached or calibrated in the run setup Robot Calibration tab.

# Test Plan
Swap out a pipette during run setup, get through the flow successfully and see results reflected once flow is closed. See video:
https://github.com/Opentrons/opentrons/assets/14302493/9f3cd608-9805-4b00-8e94-b3e90af852c9

# Changelog
1. Separate out `SetupPipetteCalibrationItem` for OT2 and Flex because of the high divergence in logic and data. Create `SetupFlexPipetteCalibrationItem` that grabs all data from `/instruments` instead of `pipettes`
2. Remove `moveToMaintenancePosition` command between detach and attach in the pipette replace flow. The mount will already be in this position so we don't need an extra call, and this was failing because the motor needed to be homed before making the request, which we already do after attaching a pipette.

# Review requests
See if I missed anything in the file separation

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/SetupFlexPipetteCalibrationItem.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/SetupFlexPipetteCalibrationItem.tsx
@@ -1,0 +1,132 @@
+import * as React from 'react'
+import { useTranslation } from 'react-i18next'
+import {
+  Flex,
+  ALIGN_CENTER,
+  DIRECTION_ROW,
+  SPACING,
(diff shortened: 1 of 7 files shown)


b-cooper (approved):
Code changes look good to me 🐕

PR #13441, opened 2023-08-31
feat(api): Rotate the calibrated module offset if the module was physically moved and rotated.

vegano1:
# Overview

When a module is calibrated, the generated module offset vector is valid for the current location the module is in. However, if the module is physically moved and rotated across the deck then the original calibrated offset vector is no longer valid. So we need to apply a rotation of 180 degrees to the calibrated module offset vector if the module was ever moved and rotated.

# Test Plan

- [x] Make sure we can still perform a module calibration on all supported modules (temp, heater-shaker, thermocycler)
- [x] Make sure we are correctly saving/loading the calibrated module offset data
- [x] Make sure we now return the location data along with the vector offset when executing the `CalibrateModule` PE command over HTTP.
- [x] Make sure this does not make well positions and labware grip points less accurate
- [x] Make sure that the module calibration offset is rotated when the module is physically moved across the deck. The module calibration offset tends to be small <1mm, let's manually edit the stored calibration file in disk to get an exaggerated offset that is easy to visually see a change in.

# Changelog

- Added new `ModuleOffsetData` type to contain `ModuleOffsetVector` and `DeckSlotLocation` data.
- The `CalibrateModuleResult` PE command now returns `DeckSlotLocation` in addition to `moduleOffset` to be used in `ModuleStore`.
- The `ModuleStore` now uses `ModuleOffsetData` instead of `ModuleOffsetVector` since we want to store both the offset and the deck slot location.
- The `get_module_calibration_offset` function now rotates the calibrated module offset by 180 degrees if the module was physically moved and rotated to a qualifying deck slot different from the original calibrated location.
- Make the `slot` member in the `ModuleCalibrationOffset` class into a required field since you won't have an offset without performing a calibration.
- Refactor module calibration loading mechanism so we don't need to pass in a slot arg as this is loaded from disk and will always be present.
- Loading module calibration now returns `Optional[ModuleCalibrationOffset]` instead of a default object, as some modules just haven't been calibrated yet and we don't want to assume the slot they were calibrated in.

# Review requests

- I'm a bit iffy on the implementation for the `_needs_rotation` function as this is a fixed set of locations we use to determine if a rotation is needed or not, so it feels rather brittle. I would love some feedback on this, maybe there's a better way of doing this.

# Risk assessment
- Low, module calibration has not been exposed yet.


api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -108,6 +109,24 @@ class SlotTransit(NamedTuple):
...
+        DeckSlotName.SLOT_A1,
+    )
+    B = (DeckSlotName.SLOT_D3, DeckSlotName.SLOT_C3, DeckSlotName.SLOT_B3)
+    if calibrated_slot in A and module_slot in B:
+        return True
+    elif calibrated_slot in B and module_slot in A:
+        return True
+    return False
+
+
```
SyntaxColoring: Re your review request:

> I'm a bit iffy on the implementation for the _needs_rotation function as this is a fixed set of locations we use to determine if a rotation is needed or not, so it feels rather brittle. I would love some feedback on this, maybe there's a better way of doing this.

Theoretically, I think the "proper" way to do this would be use the module definitions' `slotTransforms`. We would do something like `new_offset = new_slot_transform * (old_slot_transform^-1) * original_offset`. (I *think* that's how the matrix math works, anyway?)

I'm not sure off the top of my head how attainable that is, but I'm refamiliarizing myself with the code now.

api/src/opentrons/protocol_engine/commands/calibration/calibrate_module.py
```diff
@@ -46,6 +46,10 @@ class CalibrateModuleResult(BaseModel):
         ..., description="Offset of calibrated module."
     )
 
+    location: DeckSlotLocation = Field(
+        ..., description="The deck slot this module was calibrated in."
+    )
+
```
SyntaxColoring, line 52: Beware that this will break things if robot-server's database has any stored runs with `CalibrateModuleResult`s. Fortunately, I don't think this is the case. If anything, these will be in special non-persisted maintenance runs, not regular runs. Is that correct?

api/src/opentrons/protocol_engine/types.py
```diff
@@ -380,6 +380,13 @@ class ModuleOffsetVector(BaseModel):
     z: float
 
 
+class ModuleOffsetData(BaseModel):
+    """Module calibration offset data."""
+
+    moduleOffsetVector: ModuleOffsetVector
+    location: DeckSlotLocation
+
+
```
SyntaxColoring: Is this ever exposed over HTTP or stored on the filesystem? If not, we can make this a standard Python `@dataclass` instead of an icky `BaseModel`. :)

api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -108,6 +109,24 @@ class SlotTransit(NamedTuple):
...
+    module_id: str, calibrated_slot: DeckSlotName, module_slot: DeckSlotName
+) -> bool:
+    """Determine if the calibrated module offset needs to be rotated 180 degrees."""
+    A = (
+        DeckSlotName.SLOT_D1,
+        DeckSlotName.SLOT_C1,
+        DeckSlotName.SLOT_B1,
+        DeckSlotName.SLOT_A1,
+    )
+    B = (DeckSlotName.SLOT_D3, DeckSlotName.SLOT_C3, DeckSlotName.SLOT_B3)
```
SyntaxColoring: If we do stick with this basic idea of implementing `_needs_rotation()` with a fixed set of locations, can we:

1. Change `A` and `B` to `COL_1` and `COL_3`? "A" and "B" are confusing to me because they look like they're referring to row A and row B.
2. Add `DeckSlotName.SLOT_A3`, where the fixed trash currently is? I don't want us to forget it when the trash is made movable, and I think it should be harmless to add it now.

api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -652,12 +678,27 @@ def get_dimensions(self, module_id: str) -> ModuleDimensions:
...
+                if _needs_rotation(module_id, calibrated_slot, module_slot):
+                    pre_rotation = array([offset.x, offset.y, offset.z])
+                    rotation_vector = array([[-1, 0, 0], [0, -1, 0], [0, 0, 1]])
+                    offset_vector = dot(pre_rotation, rotation_vector)  # type: ignore[no-untyped-call]
+                    offset = ModuleOffsetVector(
+                        x=offset_vector[0],
+                        y=offset_vector[1],
+                        z=offset_vector[2],
+                    )
+        return offset
```
SyntaxColoring: Could we add a unit test to cover this conditional rotation logic?

SyntaxColoring (commented):
[Lookin' good.](https://www.youtube.com/watch?v=PGNiXGX2nLU)

api/src/opentrons/protocol_engine/state/modules.py
sfoster1: yeah this is what the slot transform is for. the general case fix would be to make the offset be `saved_offset * (original slot transform)^-1 * (new slot transform)`, basically what max said but the last multiplication is the new transform not the offset

api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -652,12 +678,27 @@ def get_dimensions(self, module_id: str) -> ModuleDimensions:
...
+            if offset_data:
+                offset = offset_data.moduleOffsetVector
+                calibrated_slot = offset_data.location.slotName
+                module_slot = self.get_location(module_id).slotName
+                # NOTE (ba, 2023-08-31): If the module was physically moved and
+                # rotated to a different deck slot location, then we also need to
+                # rotate the calibrated module offset by 180 degrees around the Z axis.
+                if _needs_rotation(module_id, calibrated_slot, module_slot):
+                    pre_rotation = array([offset.x, offset.y, offset.z])
+                    rotation_vector = array([[-1, 0, 0], [0, -1, 0], [0, 0, 1]])
```
sfoster1: Yeah I would really prefer using the slot transform here

sfoster1 (changes requested):
Can we use the slot transform for this? That would simplify a lot of the logic. The explicit "which slots need rotation" considerations are baked into the module definitions (though they may need added entries for flex-style names).

The slot transforms in the module definitions are 3d affine transforms in encoded matrix form; you can either use them as-is by augmenting the calibration offset to 1x4 with a `0` to get only the linear-transform-part or deaugment the affine transforms back to 3d linear transforms and multiply as normal.

api/src/opentrons/protocol_engine/commands/calibration/calibrate_module.py
vegano1: Yeah, this is correct, we have not exposed module calibration yet.

sanni-t:
> Can we use the slot transform for this? That would simplify a lot of the logic. The explicit "which slots need rotation" considerations are baked into the module definitions (though they may need added entries for flex-style names).
> 
> The slot transforms in the module definitions are 3d affine transforms in encoded matrix form; you can either use them as-is by augmenting the calibration offset to 1x4 with a `0` to get only the linear-transform-part or deaugment the affine transforms back to 3d linear transforms and multiply as normal.

Discussed in person with @vegano1 and @sfoster1 : we can't use slot transforms for rotating the calibration offset since that's not what the slot transforms are used as. The slot transforms are entities that will convert the default `labwareOffsetVector` to one appropriate for a particular slot on a particular machine. It will include rotational components only if the default vector needs to be rotated, which is not true of all modules on the OT2 and not true of any modules on the Flex since the module caddies take care of it for the modules that can be placed in multiple locations (temperature & H/S modules).

So the decision is to apply a separate rotation to just the calibration offset if the saved calibration was done on a slot in a different column than the current location. Then this rotated calibration offset will be added to the nominal, slot transformed module offset to get the final, calibrated offset vector for a labware on the module.

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -186,13 +188,43 @@ def _get_labware_position_offset(
...
             )
 
+    def _normalize_module_calibration_offset(
+        self,
+        module_location: DeckSlotLocation,
+        offset_data: Optional[ModuleOffsetData],
+    ) -> ModuleOffsetVector:
+        """Normalize the module calibration offset depending on the module location."""
+        offset = ModuleOffsetVector(x=0, y=0, z=0)
+        if offset_data:
```
sfoster1: can we invert this so it's like `if not offset_data: return offset` and then dedent the rest of the function

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -186,13 +188,43 @@ def _get_labware_position_offset(
...
+            offset = offset_data.moduleOffsetVector
+            calibrated_slot_column = self.get_slot_column(offset_data.location.slotName)
+            current_slot_column = self.get_slot_column(module_location.slotName)
+            # make sure that we have valid colums since we cant have modules in the middle of the deck
+            assert set([calibrated_slot_column, current_slot_column]).issubset(
```
sfoster1: should be, or be caught by, an enumerated error

api/src/opentrons/hardware_control/modules/module_calibration.py
```diff
@@ -35,37 +34,27 @@ class ModuleCalibrationOffset:
...
-                status=CalibrationStatus(
-                    markedAt=module_offset_data.status.markedAt,
-                    markedBad=module_offset_data.status.markedBad,
-                    source=module_offset_data.status.source,
-                ),
-            )
-    return module_cal_obj
+    module_offset: Optional[ModuleCalibrationOffset] = None
+    module_offset_data = get_module_offset(module_type, module_id)
+    if module_offset_data:
```
sfoster1: if we invert this logic so it's `if not module_offset_data: return None`, then we can `return ModuleCalibrationOffset(...)` directly

sfoster1 (approved):
Looks good once we get some tests for the rotation process and logic in there!

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2084,11 +2084,15 @@ async def save_module_offset(
...
-        elif self._pipette_handler.has_pipette(mount):
+        if self._pipette_handler.has_pipette(mount):
             instrument_id = self._pipette_handler.get_pipette(mount).pipette_id
+        elif mount == OT3Mount.GRIPPER and self._gripper_handler.has_gripper():
+            instrument_id = self._gripper_handler.get_gripper().gripper_id
+        else:
+            self._log.warning(
+                f"Could not save calibration: no instrument found for {mount}"
+            )
+            return None
```
sanni-t, line 2095: Nice

api/src/opentrons/hardware_control/modules/module_calibration.py
```diff
@@ -35,37 +34,27 @@ class ModuleCalibrationOffset:
...
+        module_offset = ModuleCalibrationOffset(
+            module=module_type,
+            module_id=module_id,
+            slot=module_offset_data.slot,
```
sanni-t: Saving the slot data in module calibration is already implemented in 7.0.0, ya? Just making sure that there won't be any calibration files in the field that don't have the slot field which could lead to file parsing errors.

api/src/opentrons/protocol_engine/state/geometry.py
sanni-t: I think an assertion is fine because this will never be false since you can't load connected modules in any columns other than 1 & 3. The only way I can think of making this assertion fail is by messing with the saved calibration file.

The assertion here is just making super sure that the data is relevant for the calculation that follows since we don't do this check anywhere in calibration flow. Also in a way, it also provides a context as to why we're performing rotation.

(We usually use asserts instead of explicit errors when we are either just placating mypy or when the condition we are asserting is practically impossible but is only implied so from other parts of the code)

api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -689,19 +688,14 @@ def get_nominal_module_offset(
...
-        offset_vector = self.get_nominal_module_offset(module_id, deck_type)
-
-        # add the calibrated module offset if there is one
-        cal_offset = self.get_module_calibration_offset(module_id)
-        return LabwareOffsetVector(
-            x=offset_vector.x + cal_offset.x,
-            y=offset_vector.y + cal_offset.y,
-            z=offset_vector.z + cal_offset.z,
-        )
+    def get_calibration_module_offset(
```
sanni-t: Just a nitpick (feel free to discard), but I think this should be called `get_module_calibration_offset`. 

This is how these two sound to me-
`calibration_module_offset` -> offset of a calibration module
`module_calibration_offset` -> offset from a module calibration process

sanni-t (approved):
LGTM!
As @sfoster1 mentioned, we should add a test that tests the rotation of the offsets.

PR #13443, opened 2023-08-31
chore: Update v7.0.0 release notes for PR #13425

SyntaxColoring:
# Overview

PR #13425 introduces a database migration. When robots in the field upgrade to it, it will take longer for them to restart. Call this out in the release notes.

I guess in practice, this will only affect OT-2 users, and it will only affect them after we reunify the Flex and OT-2 releases. So arguably, this should be in the v7.1.0 release notes, not v7.0.0? I don't know. It at least seems prudent to add this note here now so we don't forget later on.

Also, PR #13425 should make various things load faster. Call this out as an "improved feature."

# Review requests

* Does this make sense to add to the release notes now?
* words good?

### DIFF
--- a/api/release-notes.md
+++ b/api/release-notes.md
@@ -32,7 +32,6 @@ This update may take longer than usual. Allow **approximately 15 minutes** for y
 
 - The API relaxes placement restrictions for the Heater-Shaker Module on Flex.
 - Pipettes drop tips in multiple locations above the trash bin to prevent tips from stacking up.
-- Load times for various parts of the Opentrons App have been reduced.
 
 ### Known Issues
 
--- a/api/release-notes.md
+++ b/api/release-notes.md
@@ -10,11 +10,7 @@ log][]. For a list of currently known issues, please see the [Opentrons issue tr
 
 Welcome to the v7.0.0 release of the Opentrons robot software! This release adds support for the Opentrons Flex robot, instruments, modules, and labware.
 
-### Upgrade Note: Allow Extra Time for Restart
-
-When you first upgrade a robot to this version, it will take longer to restart, because it needs to convert some internal storage to a more efficient format.
-
-How much longer depends on how many different protocols you've recently run and how big they are. A lightly-loaded robot should only take an additional minute, but a heavily-loaded robot can take 10 minutes.
+This update may take longer than usual. Allow **approximately 15 minutes** for your robot to restart. This delay will only happen once.
 
 ### New Features
 
--- a/api/release-notes.md
+++ b/api/release-notes.md
@@ -10,6 +10,12 @@ log][]. For a list of currently known issues, please see the [Opentrons issue tr
 
 Welcome to the v7.0.0 release of the Opentrons robot software! This release adds support for the Opentrons Flex robot, instruments, modules, and labware.
 
+### Upgrade Note: Allow Extra Time for Restart
+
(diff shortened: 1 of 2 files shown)


api/release-notes.md
```diff
@@ -10,6 +10,12 @@ log][]. For a list of currently known issues, please see the [Opentrons issue tr
 
 Welcome to the v7.0.0 release of the Opentrons robot software! This release adds support for the Opentrons Flex robot, instruments, modules, and labware.
 
+### Upgrade Note: Allow Extra Time for Restart
```
ecormany: This is an important note, and I don't want users to be caught unawares (and contact support thinking their robot is broken), but I would prefer that it didn't bump the bigger news (Flex!) very far down the page.

Maybe instead of 2¶ under an H3, we can drop this to a couple short sentences with some bold, e.g.:
`
This update may take longer than usual. Allow **up to ten minutes** for your robot to restart. This delay will only happen once.
`

api/release-notes.md
```diff
@@ -30,6 +36,7 @@ Welcome to the v7.0.0 release of the Opentrons robot software! This release adds
 
 - The API relaxes placement restrictions for the Heater-Shaker Module on Flex.
 - Pipettes drop tips in multiple locations above the trash bin to prevent tips from stacking up.
+- Load times for various parts of the Opentrons App have been reduced.
```
ecormany: Either we should be more specific (the app is hitting certain endpoints) or leave this to the app release notes only.

app-shell/build/release-notes.md
```diff
@@ -10,6 +10,12 @@ log][]. For a list of currently known issues, please see the [Opentrons issue tr
 
 Welcome to the v7.0.0 release of the Opentrons App! This release adds support for the Opentrons Flex robot, instruments, modules, and labware.
 
+### Upgrade Note: Allow Extra Time for Restart
```
ecormany: I think we can pull this entirely. The delay won't happen until the robot update.

app-shell/build/release-notes.md
```diff
@@ -21,6 +27,10 @@ Welcome to the v7.0.0 release of the Opentrons App! This release adds support fo
   - Manually move labware around the deck during protocols. The app shows animated instructions for which labware to move, and lets you resume the protocol when movement is complete.
   - See when your protocol will pause. During a run, marks on the protocol timeline show all pauses that require user attention, including labware movement.
 
+### Improved Features
+
+- Load times for various pages have been reduced.
```
ecormany: Active?
`The app loads various pages faster.`

ecormany (commented):
Just comments for now — I think the big question is what release this is timed with. We should probably leave this open until we have a firm answer.

api/release-notes.md
SyntaxColoring: Leaving this to the app release notes only since we're treating the endpoints as experimental for now.

SyntaxColoring:
> I think the big question is what release this is timed with. We should probably leave this open until we have a firm answer.

I discussed this with @laviera. We agreed that to avoid overthinking it, since these are changes introduced in the 7.0.0 release, they should be in the 7.0.0 release notes. Even if people will only see it later, in practice. Fundamentally, release notes that describe migrations will always need to be viewed cumulatively.

ecormany (approved):
Looks good, and I agree that we should get these notices in at the soonest moment they become applicable.

There may be one more round of revisions on release notes before stable. If so, they should only add to what we have here.

PR #13444, opened 2023-08-31
fix(robot-server): Resolve LabwareRole pickle warnings

SyntaxColoring:
# Overview

PR #12811 added a new type, `LabwareRole`, that gets pickled in robot-server's database.

robot-server is warning us that we haven't specifically acknowledged that, so if we're not careful, we might rename or move `LabwareRole` and break database compatibility.

> Unpickling unknown type "LabwareRole"...this may cause problems with reading records created by older versions of this robot software. This should be reported to Opentrons and investigated. 

This PR adds `LabwareRole` to the list in `legacy_pickle.py` to resolve the warning.

This is a result of not having resolved RSS-98. See that ticket for background.

# Test Plan

None needed.

# Review requests

None in particular.

# Risk assessment

Low.


jbleon95 (approved):
Good catch!

PR #13448, opened 2023-09-01
fix(discovery-client): deep compare robot addresses to restart health poller

brenthagen:
# Overview

uses lodash isEqual to deep compare previous and current robot addresses to restart health poller. fixes overeager restart that was causing memory management issues in the electron main process.

# Test Plan

confirmed greatly reduced health poller restarts by monitoring logging in the electron main process.

# Changelog

 - deep compares previous and current robot addresses to restart health poller

# Review requests

confirm that the health polling restarts are reduced, and that health polling still restarts when needed. check usb connection and disconnection.

# Risk assessment

medium-ish, need to confirm that discovery still updates for robots that start/stop broadcasting.

### DIFF
--- a/discovery-client/src/discovery-client.ts
+++ b/discovery-client/src/discovery-client.ts
@@ -1,3 +1,5 @@
+import isEqual from 'lodash/isEqual'
+
 import { DEFAULT_PORT } from './constants'
 import { createHealthPoller } from './health-poller'
 import { createMdnsBrowser } from './mdns-browser'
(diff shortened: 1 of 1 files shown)


mjhuff (dismissed):
Yeah, the differences are pretty noticable... 
Screenshot 2023-09-01 at 11 59 55 AM
vs
Screenshot 2023-09-01 at 12 01 19 PM

mjhuff (approved):
Tested per review requests - all looks good. Also tested that discovery updates for robots that start/stop broadcasting works as expected. 💯

PR #13450, opened 2023-09-01
fix(api): remove gratuitous tip state check prior to drop tip

b-cooper:
# Overview

Before dropping tip in the hardware controller for ot2 and flex, we previously asserted the pipette
hc tip state is has_tip=False. This provides no extra security as that tip state could be stale. in
the worst case it prevents an intentional command to descend the tip ejector from occuring because
of an inaccurate sw state.

# Review requests

Issue a drop_tip command to a pipette.  The ejector sheath should descend, regardless of the pipette hc tip tracking state.

# Risk assessment
medium, technically changes behavior of OT-2's


api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py
```diff
@@ -859,7 +859,6 @@ def plan_check_drop_tip(  # type: ignore[no-untyped-def]
         home_after,
     ):
         instrument = self.get_pipette(mount)
-        self.ready_for_tip_action(instrument, HardwareAction.DROPTIP)
```
SyntaxColoring: Mentioned in-person: we'll need to do this for the OT-3 version of this file, too.

SyntaxColoring: I think we'll want to remove `DROPTIP` from the `HardwareAction` enum. Otherwise, it could look like a bug that we're *not* calling `ready_for_tip_action()`.

SyntaxColoring (approved):
I agree with removing this check, as a separation-of-concerns thing.

I would wait for confirmation from embedded or HW-testing people to make sure it's okay to remove this validation.

api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py
b-cooper: That seems a little weird to me.  Dropping tip is still a "hardware action", it's just not something that we need to check tip state prior to. Is that just a weirdly named enum then?

sfoster1 (approved):
Yeah that makes sense to me. I think it's fine to either keep or remove the enum entry, I don't mind either way.

api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py
SyntaxColoring: > Is that just a weirdly named enum then?

Maybe.

Here's what I'm looking at:

https://github.com/Opentrons/opentrons/blob/f81ea9675572cd0025026dca1a7acdf0e4a948f7/api/src/opentrons/hardware_control/instruments/ot2/pipette_handler.py#L433-L442

This is saying to me: "for any action in `HardwareAction` (`DROPTIP`, `ASPIRATE`, `DISPENSE`, `BLOWOUT`, `PREPARE_ASPIRATE`, `LIQUID_PROBE`), it's wrong to do it if you don't have a tip." So one way to fix that would be to remove `DROPTIP` from the enum. A different way would be to special-case it in `ready_for_tip_action()`. But `DROPTIP` isn't used for anything else, as far as I can tell.

robot-server/tests/integration/protocols/runtime_error.py
```diff
@@ -6,4 +6,4 @@
 def run(ctx: ProtocolContext) -> None:
     tip_rack = ctx.load_labware("opentrons_96_tiprack_300ul", 1)
     pipette = ctx.load_instrument("p300_single", "right", [tip_rack])
-    pipette.drop_tip(tip_rack.wells()[0])
+    pipette.aspirate(location=tip_rack.wells()[0], volume=100)
```
SyntaxColoring, line 9: Hm, this is a bit concerning. If this change in validation is visible all the way up in PAPI protocol land, that means people will be able to write protocols that will raise confusing errors on older software.

I'm totally down to do this if we need to. Removing validation is much less of a problem than adding it. But we should be deliberate about it. Hypothetically, how annoying would it be to retain the validation for old `apiLevel`s but remove it for new ones?

PR #13451, opened 2023-09-01
fix(hardware): separate very small components of moves into their own steps

fsinapi:
# Overview

After applying all of the different offsets on the robot's coordinate system (LPC, belt cal, etc), movements across the deck will sometimes include a very small movement component in an axis that would otherwise stay still. We have found that these movements can have a low enough velocity that you can see evidence (by looking at the encoder signals) of oscillation in the stepper rotor around the microstepping target, possibly due to the lack of inertia during such a slow movement.

Fixing movement as such a low velocity would probably require finer microstepping or some other tuning that may not be optimal for our regular speeds. Therefore, we will instead "de-diagonalize" these movements by splitting the small component into its own movement step. For now, the threshold for each component in the unit vector will be 0.1%; if the magnitude of a component in the vector is below this threshold, it will be separated out into its own vector.

## Followup PR's

- This PR adds in the chance for a move group to use a lot more steps than before! I will open a PR in `ot3-firmware` that bumps up the max steps per move, because the current value of 5 is too low if multiple axes have components to be broken out: https://github.com/Opentrons/ot3-firmware/pull/717
- I think we can comfortably drop the minimum move displacement requirement out of the unit vector function. I think it makes sense to do this in a separate followup PR for commit history legibility.

# Test Plan

The end-of-line stress test was the only thing we've been running recently that shows encoder drift as a result of these tiny movements. I ran that script again with this PR, using a robot that has shown a lot of clear drift after 20 cycles. After running with this PR pushed, there was no drift at all in the results.

I also checked the CAN logs to make sure that the movements were generated with more steps, and with more sane velocity/accel values for the axis making the tiny movements.

# Changelog

# Review requests

# Risk assessment

Decently high! This is touching core Flex move generation code, if it's wrong then we can break a lot of stuff. On the other hand, that makes it relatively easy to test by just running normal protocols.

### DIFF
--- a/hardware/opentrons_hardware/hardware_control/motion_planning/move_utils.py
+++ b/hardware/opentrons_hardware/hardware_control/motion_planning/move_utils.py
@@ -159,6 +159,38 @@ def limit_max_speed(
     return max_linear_speed * scale
 
 
+def _unit_vector_to_move(
+    unit_vector: Coordinates[AxisKey, np.float64],
+    distance: np.float64,
+    max_speed: np.float64,
+    constraints: SystemConstraints[AxisKey],
+) -> Move[AxisKey]:
+    speed = limit_max_speed(unit_vector, max_speed, constraints)
+    third_distance = np.float64(distance / 3)
+    return Move(
+        unit_vector=unit_vector,
+        distance=distance,
+        max_speed=speed,
+        blocks=(
+            Block(
+                distance=third_distance,
+                initial_speed=speed,
+                acceleration=np.float64(0),
+            ),
+            Block(
+                distance=third_distance,
+                initial_speed=speed,
+                acceleration=np.float64(0),
+            ),
+            Block(
+                distance=third_distance,
+                initial_speed=speed,
+                acceleration=np.float64(0),
+            ),
+        ),
+    )
+
+
 def targets_to_moves(
     initial: Coordinates[AxisKey, CoordinateValue],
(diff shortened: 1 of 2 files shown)


hardware/opentrons_hardware/hardware_control/motion_planning/move_utils.py
```diff
@@ -96,33 +172,39 @@ def targets_to_moves(
...
-                    initial_speed=speed,
-                    acceleration=np.float64(0),
-                ),
-                Block(
-                    distance=third_distance,
-                    initial_speed=speed,
-                    acceleration=np.float64(0),
-                ),
-            ),
+        initial_unit_vector, initial_distance = get_unit_vector(
```
sfoster1, line 207: for readability i'd slightly prefer making the body that we pushed down (from old line 99 to line 125) be in its own function but it's a very slight preference.

sfoster1 (approved):
Looks awesome to me, very slight style request but I wouldn't mind if we merge as is since it's tested.

hardware/opentrons_hardware/hardware_control/motion_planning/move_utils.py
fsinapi: https://github.com/Opentrons/opentrons/pull/13451/commits/7935189bb266180f2b8c7cb78dd68804c41f599d

PR #13452, opened 2023-09-01
feat(api): hardware: add set_liquid_class

sfoster1:
The other step after `push_out` for low volume handling is to change the `bottom` position of the pipette. Since this can cause motion, we won't do it automatically; instead, we'll have a new command and a new hardware control method to make the configuration change, and the position will change the next time someone calls `prepare_for_aspirate`.

This PR adds the hardware control side of that, including liquid class switching and some changes to tip settings. You call the new method `set_liquid_class` which sets the liquid class on a pipette, with appropriate errors if the class doesn't exist or isn't valid.

## Testing
- In a protocol, drill down to calling `hardware.set_liquid_class` with class `lowVolumeDefault` on a 50ul pipette on a flex and check that the bottom position changes appropriately.

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
@@ -599,8 +599,14 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
 
     def set_liquid_class_by_name(self, class_name: str) -> None:
         """Change the currently active liquid class."""
+        if self.current_volume > 0:
+            raise CommandPreconditionViolated(
+                "Cannot switch liquid classes when liquid is in the tip"
+            )
         try:
             new_name = pip_types.LiquidClasses[class_name]
+            if new_name == self._liquid_class_name:
+                return
             new_class = self._config.liquid_properties[self._liquid_class_name]
         except KeyError:
             raise InvalidLiquidClassName(
@@ -630,12 +636,15 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         self._active_tip_settings = self._liquid_class.supported_tips[
             self._active_tip_setting_name
         ]
+        self.ready_to_aspirate = False
 
     def set_tip_type_by_volume(self, tip_volume: float) -> None:
         """Change the currently active tip type."""
         intified = float(int(tip_volume))
         try:
             new_name = pip_types.PipetteTipType(int(intified))
+            if new_name == self._active_tip_setting_name:
+                return
             new_tips = self._liquid_class.supported_tips[new_name]
         except (ValueError, KeyError) as e:
             raise InvalidLiquidClassName(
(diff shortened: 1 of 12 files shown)


api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -565,6 +597,97 @@ def as_dict(self) -> "Pipette.DictType":
...
+                detail={
+                    "requested-class-name": class_name,
+                    "current-tip": str(self._active_tip_setting_name),
+                },
+            )
+        self._liquid_class_name = new_name
+        self._liquid_class = new_class
+        if not self.has_tip:
+            self._active_tip_setting_name = pip_types.PipetteTipType(
+                self._liquid_class.max_volume
```
Laura-Danielle: If we're going off of liquid_class max_volume this would actually be a problem now that I think about it 🤔 . The max volume is 45ul for the low volume parameter and that is not a supported value for the pipette tip type.

I am not a huge fan of changing the tip type to include a "45ul" tip because that's technically not a real thing. For this use-case can we fall back to the default max volume? I suppose we'll just have to save it as an attribute on the pipette.

Laura-Danielle (approved):
One thing we need to fix about the max volume for `lowDefault` but otherwise everything looks great.

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
sfoster1: oh that is a good and really gross point. honestly i think i'd rather either
- allow the configured tip setting to be `None` (lot of work)
- add a `default_tip_kind` attribute to each liquid class (going to do this one)

sfoster1: @Laura-Danielle fixed this by just taking the first element in an arbitrarily (but consistently) sorted list of pipette tip types from the LQ

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -237,6 +242,39 @@ def reset_state(self) -> None:
 
         self._tip_overlap_lookup = self.liquid_class.tip_overlap_dictionary
 
+    def _update_feeds_and_speeds_defaults(
```
caila-marashaj: function name goes hard

Laura-Danielle (approved):
nice looks good

caila-marashaj (approved):

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
sfoster1: sadly i had to get rid of it

Laura-Danielle (approved):
changes look good.

fsinapi (approved):
LGTM!

vegano1 (approved):

PR #13459, opened 2023-09-04
feat(api) Add ability to configure liquid class in Protocol Engine

Laura-Danielle:
# Overview

This should encapsulate some of the work required to add an additional command to update the liquid class in the protocol engine. I did not edit the user facing API yet because I thought that would be better suited in a follow-up PR, though I expect it to be a much faster follow-up than this PR.

# To Do

- [ ] Fix failing tests (which might be due to bug in code I wrote)
- [ ] Rebase on latest 7.0.0
- [ ] Ensure that the command for updating liquid class is in all the correct places without being utilized by `core` yet

# Test Plan
Make sure you can still successfully simulate and run a protocol

# Changelog

- Add the ability to change the liquid class type in `prepare_for_aspirate`
- Add a `model_version` in the pipette dict so that the protocol engine can use that value to look up static configs
- Modified the static configs to pass in the liquid class and look up the rest of the data in the `state/pipettes.py` file

# Review requests

Check that the code makes sense and such.

# Risk assessment

Medium-ish. Theoretically there should be no information change (aside from updating the run-time static config).

### DIFF
--- a/api/src/opentrons/hardware_control/dev_types.py
+++ b/api/src/opentrons/hardware_control/dev_types.py
@@ -92,6 +92,7 @@ class PipetteDict(InstrumentDict):
     liquid_properties: Dict[
         pip_types.LiquidClasses, pipette_definition.PipetteLiquidPropertiesDefinition
     ]
+    model_version: pipette_definition.PipetteModelVersionType
 
 
 class PipetteStateDict(TypedDict):
--- a/api/src/opentrons/hardware_control/dev_types.py
+++ b/api/src/opentrons/hardware_control/dev_types.py
@@ -15,11 +15,7 @@ from opentrons_shared_data.pipette.dev_types import (
     PipetteName,
     ChannelCount,
 )
-from opentrons_shared_data.pipette.types import PipetteTipType
-from opentrons_shared_data.pipette.pipette_definition import (
-    PipetteConfigurations,
-    SupportedTipsDefinition,
-)
+from opentrons_shared_data.pipette import pipette_definition, types as pip_types
 from opentrons_shared_data.gripper import (
     GripperModel,
     GripperDefinition,
@@ -43,7 +39,7 @@ class GripperSpec(InstrumentSpec):
 
 
 class AttachedPipette(TypedDict):
-    config: Optional[PipetteConfigurations]
+    config: Optional[pipette_definition.PipetteConfigurations]
     id: Optional[str]
 
 
@@ -93,7 +89,9 @@ class PipetteDict(InstrumentDict):
     ready_to_aspirate: bool
     has_tip: bool
     default_blow_out_volume: float
-    supported_tips: Dict[PipetteTipType, SupportedTipsDefinition]
+    liquid_properties: Dict[
+        pip_types.LiquidClasses, pipette_definition.PipetteLiquidPropertiesDefinition
+    ]
 
 
 class PipetteStateDict(TypedDict):
--- a/api/src/opentrons/hardware_control/instruments/ot2/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot2/pipette.py
@@ -191,9 +191,13 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
     def liquid_class(self) -> PipetteLiquidPropertiesDefinition:
         return self._liquid_class
 
-    def configure_liquid_class(self, liquid_class: pip_types.LiquidClasses) -> None:
-        self._liquid_class = self._config.liquid_properties[liquid_class]
-        self._update_active_settings_from_liquid_class()
+    def configure_liquid_class(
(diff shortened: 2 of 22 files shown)


api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -120,21 +139,55 @@ def handle_action(self, action: Action) -> None:
             self._state.movement_speed_by_id[action.pipette_id] = action.speed
         elif isinstance(action, AddPipetteConfigAction):
             config = action.config
+            liquid_class = LiquidClasses.default
```
sfoster1, line 142: wait so doing a pipette settings override implicitly swaps back to the default liquid class?

sfoster1: and default tip setting?

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -174,37 +227,44 @@ def _handle_command(self, command: Command) -> None:  # noqa: ANN101, C901
...
-            static_config = self._state.static_config_by_id.get(pipette_id)
-            if static_config:
-                tip_configuration = static_config.tip_configuration_lookup_table[
-                    static_config.max_volume
+            liquid_class = self._state.liquid_class_definition_by_id.get(pipette_id)
+            if liquid_class:
+                self._state.tip_configuration_by_id[
+                    pipette_id
+                ] = _get_tip_configuration_from_volume(
+                    liquid_class, liquid_class.max_volume
```
sfoster1, line 257: this probably wants to be some default value since a liquid class's max volume might not be that of a valid tip

api/src/opentrons/hardware_control/dev_types.py
```diff
@@ -93,7 +89,10 @@ class PipetteDict(InstrumentDict):
     ready_to_aspirate: bool
     has_tip: bool
     default_blow_out_volume: float
-    supported_tips: Dict[PipetteTipType, SupportedTipsDefinition]
+    liquid_properties: Dict[
```
sfoster1, line 92: I really don't like dumping the entire property set in here. I know that we need some of the data available upstream especially for the whole PE-only analysis thing but this is feeling like a whole lot, especially since it doesn't even say _which_ liquid class is active.

api/src/opentrons/hardware_control/instruments/ot2/pipette.py
```diff
@@ -191,6 +191,14 @@ def config(self) -> PipetteConfigurations:
     def liquid_class(self) -> PipetteLiquidPropertiesDefinition:
         return self._liquid_class
 
+    def configure_liquid_class(
+        self, liquid_class: Optional[pip_types.LiquidClasses]
+    ) -> None:
+        if liquid_class:
```
sfoster1, line 197: we should not handle an `optional` like this. either this argument should not be `optional` or this function should have a fallback if the argument isn't specified.

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -120,21 +139,55 @@ def handle_action(self, action: Action) -> None:
...
+                default_blow_out=default_tip_configuration.default_blowout_flowrate.values_by_api_level,
+                default_aspirate=default_tip_configuration.default_aspirate_flowrate.values_by_api_level,
+                default_dispense=default_tip_configuration.default_dispense_flowrate.values_by_api_level,
+            )
+        elif isinstance(action, UpdateLiquidClassAction):
+            self._state.liquid_class_by_id[action.pipette_id] = LiquidClasses(
+                action.liquid_class
+            )
+
+            new_liquid_properties = config.liquid_properties[liquid_class]
```
sfoster1, line 173: The more I read this the more I think this stuff shouldn't live in state directly. I think this command should be calling the liquid class switcher, and since it's separate I think the liquid class switcher should _not_ be prepare for aspirate, and which liquid class is currently active should be remembered by hardware and mirrored in the data provider

sfoster1 (changes requested):
I think we should use https://github.com/Opentrons/opentrons/pull/13452 with some changes for the hardware side of this - i'll change its `configure_for_volume` implementation to take a liquid class by name.

i think this PR looks ok on the PE side of things but in my opinion the big thing we should change is: let's not hold the active liquid class and tip settings in PE state. doing that means we duplicate state and logic with the hardware controller while running pipettes. instead, let's have hardware handle it as in the other PR, and i'll refactor that other pr to have some reusable statics that we can implement in the pipette data provider.

api/src/opentrons/protocol_engine/state/pipettes.py
Laura-Danielle: You can only override settings with the new `UpdateLiquidClass` setting. Otherwise this is always a 'fresh start' command if you will.

Laura-Danielle:
Closing this in favor of a different branch.

PR #13461, opened 2023-09-05
fix(app): Update ODD maintenance run take over logic

mjhuff:
Closes RAUT-650

# Overview

This PR fixes a bug in maintenance run flow in which the ODD takeover modal would not always appropriately render when a maintenance run was initiated after an initiated maintenance run on the ODD.

# Test Plan

- Verified that beginning a maintenance run on the ODD (such as pipette calibration) and then initiating a maintenance run on the desktop causes expected takeover modal rendering.
- Verified takeover modals would not render unexpectedly on the ODD when initiating a maintenance run on the ODD. 

# Changelog

- Added context to track an ODD id via a wrapper around CreateRunMutation().
- Added relevant testing.

# Risk assessment
low


jerader (commented):
This works for the most part. but i noticed that if i start a maintenance run on the ODD, exit, then start a maintenance run on the ODD again, the takeover modal pops up for a few seconds before automatically dismissing.

jerader:
Here is a screen recording of what i mean in my comment above: 

https://github.com/Opentrons/opentrons/assets/66035149/e51e1f19-a958-40b7-b4ab-415a9146aa41

jerader (approved):
woo! seems to be working properly now, nice!! Logic also is a lot more straightforward since we aren't calling `useMaintenanceRunTakeover` in all those components that open up the maintenance runs.

PR #13463, opened 2023-09-05
feat(shared-data): add plunger homing configurations

caila-marashaj:
## Overview
We want to be able to specify a higher speed or current for pipette homing in order to overcome any potential static friction.

## Changelog

- add new `plungerHomingConfig` to shared-data
- apply the speed and current from that config to plunger homing in the hardware controller
  - break pipette plunger homing out of `OT3API._home_axis()` a little bit to apply the current and speed

## Testing
make sure homing works the same as it has on all axes


sfoster1 (approved):
Looks good to me, nice!

PR #13464, opened 2023-09-05
fix(app): update send protocol to Flex slideout text

ncdiehl11:
# Overview

[RQA=1448](https://opentrons.atlassian.net/browse/RQA-1448)

We want to change the text displayed in the slideout menu when sending a protocol to the Flex. Instead of the past title and button texts of "Choose robot to run" and "Proceed to run," respectively, we want to display "Send protocol to Opentrons Flex" and "Send."

# Test Plan

- Verify that title and button text reflect designs on slideout for sending protocols to Flex

# Changelog
- change slideout title and button text
- add translation for new text
- update tests

# Risk assessment

low


app/src/organisms/SendProtocolToOT3Slideout/index.tsx
```diff
@@ -38,7 +38,7 @@ export function SendProtocolToOT3Slideout(
     srcFiles,
     mostRecentAnalysis,
   } = storedProtocolData
-  const { t } = useTranslation(['protocol_details', 'shared'])
+  const { t } = useTranslation(['protocol_details', 'protocol_list', 'shared'])
```
jerader: `shared` can be removed from the array since it isnt in use

```suggestion
  const { t } = useTranslation(['protocol_details', 'protocol_list'])
```

app/src/organisms/SendProtocolToOT3Slideout/index.tsx
```diff
@@ -152,7 +150,7 @@ export function SendProtocolToOT3Slideout(
           onClick={handleSendClick}
           width="100%"
         >
-          {t('shared:proceed_to_setup')}
+          {t('protocol_details:send')}
```
jerader, line 157: should this be `protocol_list`?  Also, can you capitalize it through the i18n formatter function instead of directly in the i18n value? Since the string is 1 word long, its better to keep it all lower case to make it more reusable. You can look at the different formatter functions we have here: `app/src/i18n.ts`

```suggestion
          {i18n.format(t('protocol_list:send'), 'capitalize')}
```

jerader (commented):
nice, left a few comments.

app/src/organisms/SendProtocolToOT3Slideout/index.tsx
jerader: Oh it looks like there is already a `send` i18n key in `protocol_details.json` - feel free to use that instead but please delete it from `protocol_list`!

app/src/assets/localization/en/protocol_list.json
```diff
@@ -10,7 +10,8 @@
   "reanalyze_or_view_error": "<analysisLink>Reanalyze</analysisLink> protocol or view <errorLink>error details</errorLink>",
   "right_mount": "right mount",
   "robot": "robot",
-  "send_to_ot3": "Send to Opentrons Flex",
+  "send": "Send",
+  "send_to_ot3": "Send protocol to Opentrons Flex",
```
jerader: lets double check if `Opentrons Flex` should be a const instead of being translated in here since its a proper noun

ncdiehl11: By adding a const to shared-data/js/constants.js possibly?

shared-data/js/constants.ts
```diff
@@ -48,6 +48,9 @@ export const GRIPPER_V1_1: 'gripperV1.1' = 'gripperV1.1'
 export const GRIPPER_V1_2: 'gripperV1.2' = 'gripperV1.2'
 export const GRIPPER_MODELS = [GRIPPER_V1, GRIPPER_V1_1, GRIPPER_V1_2]
 
+// robot display name
+export const FLEX_DISPLAY_NAME: 'Opentrons Flex' = 'Opentrons Flex'
```
jerader, line 52: nice!

app/src/assets/localization/en/protocol_list.json
```diff
@@ -10,7 +10,7 @@
   "reanalyze_or_view_error": "<analysisLink>Reanalyze</analysisLink> protocol or view <errorLink>error details</errorLink>",
   "right_mount": "right mount",
   "robot": "robot",
-  "send_to_ot3": "Send to Opentrons Flex",
+  "send_to_ot3": "Send protocol to {{robot_display_name}}",
```
jerader, line 14: hm, i think using camel case for consts in i18n values is what we typically do (at least what I typically do). But it seems like we use camel case and snake case interchangeably so this is 👌

jerader (approved):
nice! 🦞

app/src/organisms/SendProtocolToOT3Slideout/index.tsx
```diff
@@ -5,6 +5,8 @@ import { useSelector } from 'react-redux'
 
 import { useCreateProtocolMutation } from '@opentrons/react-api-client'
 
+import { FLEX_DISPLAY_NAME } from '@opentrons/shared-data/js/constants'
```
jerader: oops didn't notice this until now but you might be able to import this directly from `@opentrons/shared-data`

app/src/assets/localization/en/protocol_list.json
ncdiehl11: my bad! thanks for the advice. seeing now that there is a slight difference to the language used in the overflow menu vs. the flex slideout: "Send to Opentrons Flex" vs. "Send *protocol* to Opentrons Flex". I'll fix now.

app/src/organisms/SendProtocolToOT3Slideout/index.tsx
ncdiehl11: good catch

app/src/assets/localization/en/protocol_list.json
jerader: oh good catch! i did not notice that!

jerader (approved):
lgtm!

PR #13466, opened 2023-09-05
fix(robot-server): Pin pickle protocol version to v4

SyntaxColoring:
# Overview

This is a change to make it easier for us to debug and do autopsies on robot-server databases. This helps when debugging issues like RQA-1473.

There are [several versions](https://docs.python.org/3/library/pickle.html#data-stream-format) of the `pickle` data format. Currently, the version that robot-server uses changes depending on its environment:

| Environment | Python version | `pickle` protocol version |
|-------------|----------------|-----------------------------------|
| Local dev   | 3.7            | 4                                 |
| OT-2        | 3.7            | 4                                 |
| Flex        | 3.8            | 5                                 |

This is annoying because it means if you save `/var/lib/opentrons-robot-server` from a Flex, you can't just point a local dev server to it with `make -C robot-server dev OT_ROBOT_SERVER_PERSISTENCE_DIRECTORY=...`. Our local dev servers run Python 3.7, so they'll raise errors when they try to read the pickle v5 records.

To fix that, this PR normalizes all robot-server environments to use pickle protocol v4, the least common denominator.

# Test Plan

* [x] Make sure existing run records on Flexes are still returned without error. i.e. we should still be able to *read* the pickle v5 records even though we're not adding new ones.

# Review requests

None in particular.

# Risk assessment

There's some risk that this will harm performance because with the v5 format, ["one copy is avoided on the serialization path".](https://peps.python.org/pep-0574/#improved-in-band-performance)

### DIFF
--- a/robot-server/robot_server/persistence/_tables.py
+++ b/robot-server/robot_server/persistence/_tables.py
@@ -2,6 +2,7 @@
 import sqlalchemy
 
 from . import legacy_pickle
+from .pickle_protocol_version import PICKLE_PROTOCOL_VERSION
 from ._utc_datetime import UTCDateTime
 
 _metadata = sqlalchemy.MetaData()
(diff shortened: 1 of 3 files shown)


jbleon95 (approved):
LGTM

TamarZanzouri (approved):
LGTM!

PR #13467, opened 2023-09-05
chore: Fix outdated team names in CODEOWNERS

SyntaxColoring:
# Overview

Fix some broken references in our GitHub CODEOWNERS file because of outdated team names.

# Changelog

* `hmg` -> `embedded-sw`
* `cpx` -> `robot-svcs`
* `spddrs` -> `app-and-uis`
* `discovery-client` is now jointly owned by `app-and-uis` and `robot-svcs`
* `shared-data/protocol` is now jointly owned by `app-and-uis` and `robot-svcs`
* Removed `api/src/opentrons` because I think its ownership is more mixed these days, and better decided by per-PR review requests.

# Review requests

Do we agree with these ownerships? Anything we want to remove or add?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,18 +9,17 @@
 /webpack-config @Opentrons/js
 
 # subprojects - those with clear team ownership should have appropriate notifications
-/protocol-designer @Opentrons/spddrs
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
okay! thank you

PR #13468, opened 2023-09-05
fix(app): refactor instrument cards for bad instruments

smb2268:
fix RQA-1478, RQA-1315, RQA-1252

# Overview
Instrument cards for bad instruments were rendered outside of the `GripperCard` and `PipetteCard` components, which is where attachment flows are rendered. This caused flows to close when an instrument that needed a subsystem update was attached during a flow. 
This PR refactors `GripperCard` and `PipetteCard` to handle instruments that need subsystem updates so that no longer happens and reflects in progress updates on both desktop and ODD.

# Test Plan
Go through an attachment flow with an instrument that will need a firmware update. See that the attachment flow no longer closes. See that an in progress modal appears on the ODD.

# Changelog
1. Refactor `GripperCard` and `PipetteCard` to handle instruments that need subsystem updates
2. Show dynamic text and color for subsystem update banner if an update is already in progress
3. Show ODD in progress modal anytime an externally initiated update is in progress

# Review requests
Ensure refactor looks sound

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/Devices/InstrumentsAndModules.tsx
+++ b/app/src/organisms/Devices/InstrumentsAndModules.tsx
@@ -23,8 +23,6 @@ import {
 
 import { StyledText } from '../../atoms/text'
 import { Banner } from '../../atoms/Banner'
-import { UpdateBanner } from '../../molecules/UpdateBanner'
-import { InstrumentCard } from '../../molecules/InstrumentCard'
 import { useCurrentRunId } from '../ProtocolUpload/hooks'
 import { ModuleCard } from '../ModuleCard'
 import { FirmwareUpdateModal } from '../FirmwareUpdateModal'
@@ -74,11 +72,7 @@ export function InstrumentsAndModules({
 
   const attachedGripper =
     (attachedInstruments?.data ?? []).find(
-      (i): i is GripperData => i.instrumentType === 'gripper' && i.ok
-    ) ?? null
-  const badGripper =
-    (attachedInstruments?.data ?? []).find(
-      (i): i is BadGripper => i.subsystem === 'gripper' && !i.ok
+      (i): i is GripperData | BadGripper => i.subsystem === 'gripper'
     ) ?? null
   const attachedLeftPipette =
     attachedInstruments?.data?.find(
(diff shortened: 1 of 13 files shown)


app/src/organisms/FirmwareUpdateModal/FirmwareUpdateTakeover.tsx
```diff
@@ -3,43 +3,97 @@ import * as React from 'react'
...
+  })
   const isUnboxingFlowOngoing = useIsUnboxingFlowOngoing()
 
+  const {
+    data: currentSubsystemsUpdatesData,
+  } = useCurrentAllSubsystemUpdatesQuery({
+    refetchInterval: POLL_INTERVAL_5000_MS,
+  })
+  const externalSubsystemUpdateExists =
+    currentSubsystemsUpdatesData?.data.some(
```
jerader: can this be a `.find()` instead of a `some()`?

jerader (approved):
code makes sense to me, not able to test it though

koji:
During firmware update, pipette card doesn't show a pipette image?

app/src/organisms/Devices/InstrumentsAndModules.tsx
```diff
@@ -193,68 +187,32 @@ export function InstrumentsAndModules({
...
-                        attachedGripper != null
-                          ? attachedGripper.serialNumber
-                          : 'no_gripper'
-                      }
-                      setShowBanner={() => null}
-                      updateType="firmware"
-                      handleUpdateClick={() => setSubsystemToUpdate('gripper')}
-                    />
+                  isCalibrated={
+                    attachedGripper?.ok === true &&
```
koji, line 212:
```suggestion
                    attachedGripper?.ok &&
```

since you do `isOT3 && attachedLeftPipette?.ok` in line 198

app/src/organisms/Devices/InstrumentsAndModules.tsx
```diff
@@ -275,42 +233,25 @@ export function InstrumentsAndModules({
...
-                        attachedRightPipette != null
-                          ? attachedRightPipette.serialNumber
-                          : 'no_right_pipette'
-                      }
-                      setShowBanner={() => null}
-                      updateType="firmware_important"
-                      handleUpdateClick={() => setSubsystemToUpdate('gripper')}
-                    />
-                  }
+                  pipetteIs96Channel={false}
```
koji, line 252: Probably this does't have any impact but I guess using `is96ChannelAttached` would be better?

app/src/organisms/FirmwareUpdateModal/FirmwareUpdateTakeover.tsx
```diff
@@ -3,43 +3,91 @@ import * as React from 'react'
...
 } from '@opentrons/react-api-client'
-
+import { Portal } from '../../App/portal'
 import { useIsUnboxingFlowOngoing } from '../RobotSettingsDashboard/NetworkSettings/hooks'
+import { UpdateInProgressModal } from './UpdateInProgressModal'
 import { UpdateNeededModal } from './UpdateNeededModal'
+import type { Subsystem } from '@opentrons/api-client'
 
-const INSTRUMENT_POLL_INTERVAL = 5000
+const POLL_INTERVAL_5000_MS = 5000
```
koji: If there isn't any specific reason to use 5000, removing it would be better since we basically don't include the number for this type of a const.

```suggestion
const POLL_INTERVAL_MS = 5000
```

app/src/organisms/Devices/PipetteCard/index.tsx
```diff
@@ -153,98 +173,139 @@ export const PipetteCard = (props: PipetteCardProps): JSX.Element => {
...
+                {pipetteModelSpecs === null ? null : (
+                  <InstrumentDiagram
+                    pipetteSpecs={pipetteModelSpecs}
+                    mount={mount}
+                    //  pipette images for Flex are slightly smaller so need to be scaled accordingly
+                    transform={isOt3 ? 'scale(0.4)' : 'scale(0.3)'}
+                    size="3.125rem"
+                    transformOrigin={isOt3 ? '-50% -10%' : '20% -10%'}
+                  />
+                )}
```
koji:
```suggestion
                {pipetteModelSpecs !== null ? (
                  <InstrumentDiagram
                    pipetteSpecs={pipetteModelSpecs}
                    mount={mount}
                    //  pipette images for Flex are slightly smaller so need to be scaled accordingly
                    transform={isOt3 ? 'scale(0.4)' : 'scale(0.3)'}
                    size="3.125rem"
                    transformOrigin={isOt3 ? '-50% -10%' : '20% -10%'}
                  />
                ): null}
```

smb2268:
> During firmware update, pipette card doesn't show a pipette image?

@koji yep - when an instrument needs a firmware update we can't tell what pipette it is so we don't show an image or any other specific information. The shape that we get back from the `/instruments` endpoint in this scenario is `BadPipette` https://github.com/Opentrons/opentrons/blob/edge/api-client/src/instruments/types.ts#L63-L69

app/src/organisms/Devices/InstrumentsAndModules.tsx
smb2268: We want this to be false regardless of whether a 96 channel is attached because the 96 will be rendered on the `left` instrument card and this is the `right card`

PR #13469, opened 2023-09-06
fix(app): fix module calibration overflowmenu bug

koji:
# Overview
Use `formattedPipetteOffsetCalibrations` as well as pipette calibration table since the previous implementation always gets empty.
Check pipettes are attached and calibrated -> allow users to calibrate modules

closes RQA-1485

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


mjhuff:
Tested code and works as expected, but I wonder if we should show some sort of feedback to let people know they need to calibrate pipettes first? Seems hard to tell it's necessary on this page. IDK, this is probably a design question.

https://github.com/Opentrons/opentrons/assets/64858653/9ef8b99c-8d31-493c-8a60-86f80f6086b4

mjhuff (approved):
We talked about this in person - lgtm 🚀

PR #13470, opened 2023-09-06
fix(app): fix home gantry on restart setting

mjhuff:
Closes RQA-1458

# Overview

The "Home gantry on restart" button on the ODD (RobotSettingsDashboard) is not properly toggling although state is updating. This PR fixes the toggle and changes the default display to "on", since this matches expected behavior given the button subtext.

https://github.com/Opentrons/opentrons/assets/64858653/65435767-da54-438a-8111-095c3c11e164

# Test Plan

- Verified pressing the button toggles on/off correctly and that state is appropriately updated. 

# Changelog

- Changed reflected button state to match the correct setting.

# Risk assessment

low


koji (approved):
Thank you for fixing this!

PR #13471, opened 2023-09-06
fix(shared-data): Migrate units for mutable configs

Laura-Danielle:
# Overview

In the older pipette config code, there was a mix of `mm/s` and `mm/sec` as units. If someone had an older file that contained `mm/sec`, it would fail because the enum was expecting exact string match.

This PR fixes that issue and adds a small test.


jerader (approved):
yay! OT-2 simulation works for me now. thank you!

PR #13473, opened 2023-09-06
feat(api): add command to configure pipette volume in Protocol Engine

Laura-Danielle:
# Overview

Add the configure for liquid class command to protocol engine and allow virtual configs to return different liquid class configurations.

# Changelog

- Add new command to configure liquid for volume
- Add a new equipment command that will configure the volume in the hardware controller and update the pipette state store
- Update the hardware controller to again have a command to configure liquid for volume (a copy of work @sfoster1 had previously done)

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
@@ -136,6 +136,28 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         else:
             self._pipetting_function_version = PIPETTING_FUNCTION_LATEST_VERSION
 
+    def get_liquid_class_for_volume(self, volume: float) -> str:
+        """Get the liquid class required for the specified volume.
+        Some pipettes have different liquid classes for different volumes. If applicable,
+        and if the volume is in the different range, this will return that liquid class.
+        Otherwise, it will return the current liquid class.
+        This function takes into account the current liquid class to provide hysteresis.
+        """
+        # For now, until we add more liquid classes, we're going to hardcode the default
+        # and lowVolumeDefault liquid classes as the ones to switch between.
+        has_lvd = (
+            pip_types.LiquidClasses.lowVolumeDefault in self._config.liquid_properties
+        )
+        if not has_lvd:
+            return self._liquid_class_name.name
+        if self._liquid_class_name == pip_types.LiquidClasses.lowVolumeDefault:
+            if volume > (self.liquid_class.max_volume * 0.75):
+                return pip_types.LiquidClasses.default.name
+        if self._liquid_class_name == pip_types.LiquidClasses.default:
+            if volume < (self.liquid_class.max_volume * 0.5):
+                return pip_types.LiquidClasses.lowVolumeDefault.name
+        return self._liquid_class_name.name
+
     @property
     def config(self) -> PipetteConfigurations:
         return self._config
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
@@ -136,28 +136,6 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         else:
             self._pipetting_function_version = PIPETTING_FUNCTION_LATEST_VERSION
 
-    def get_liquid_class_for_volume(self, volume: float) -> str:
-        """Get the liquid class required for the specified volume.
-        Some pipettes have different liquid classes for different volumes. If applicable,
-        and if the volume is in the different range, this will return that liquid class.
-        Otherwise, it will return the current liquid class.
-        This function takes into account the current liquid class to provide hysteresis.
-        """
-        # For now, until we add more liquid classes, we're going to hardcode the default
-        # and lowVolumeDefault liquid classes as the ones to switch between.
-        has_lvd = (
-            pip_types.LiquidClasses.lowVolumeDefault in self._config.liquid_properties
-        )
-        if not has_lvd:
-            return self._liquid_class_name.name
-        if self._liquid_class_name == pip_types.LiquidClasses.lowVolumeDefault:
-            if volume > (self.liquid_class.max_volume * 0.75):
-                return pip_types.LiquidClasses.default.name
-        if self._liquid_class_name == pip_types.LiquidClasses.default:
-            if volume < (self.liquid_class.max_volume * 0.5):
-                return pip_types.LiquidClasses.lowVolumeDefault.name
-        return self._liquid_class_name.name
-
     @property
     def config(self) -> PipetteConfigurations:
         return self._config
@@ -165,6 +143,10 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
     @property
     def liquid_class(self) -> PipetteLiquidPropertiesDefinition:
         return self._liquid_class
+
+    @property
+    def liquid_class_name(self) -> pip_types.LiquidClasses:
+        return self._liquid_class_name
 
     @property
     def channels(self) -> pip_types.PipetteChannelType:
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette.py
@@ -16,6 +16,7 @@ from opentrons_shared_data.pipette.pipette_definition import (
     PipetteNameType,
     PipetteModelVersionType,
     PipetteLiquidPropertiesDefinition,
+    default_tip_for_liquid_class,
 )
 from opentrons_shared_data.errors.exceptions import (
     InvalidLiquidClassName,
@@ -110,9 +111,7 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         )
         self.ready_to_aspirate = False
 
-        self._active_tip_setting_name = pip_types.PipetteTipType(
-            self._liquid_class.max_volume
-        )
+        self._active_tip_setting_name = default_tip_for_liquid_class(self._liquid_class)
         self._active_tip_settings = self._liquid_class.supported_tips[
             self._active_tip_setting_name
         ]
@@ -231,7 +230,7 @@ class Pipette(AbstractInstrument[PipetteConfigurations]):
         self.ready_to_aspirate = False
         #: True if ready to aspirate
         self.set_liquid_class_by_name("default")
-        self.set_tip_type_by_volume(self.liquid_class.max_volume)
+        self.set_tip_type(default_tip_for_liquid_class(self._liquid_class))
 
         self._aspirate_flow_rate = (
             self._active_tip_settings.default_aspirate_flowrate.default
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
@@ -899,3 +899,11 @@ class OT3PipetteHandler:
     async def set_liquid_class(self, mount: OT3Mount, liquid_class: str) -> None:
         pip = self.get_pipette(mount)
         pip.set_liquid_class_by_name(liquid_class)
+
+    async def configure_for_volume(self, mount: OT3Mount, volume: float) -> None:
+        pip = self.get_pipette(mount)
+        if pip.current_volume > 0:
+            # Switching liquid classes can't happen when there's already liquid
+            return
+        new_class = pip.get_liquid_class_for_volume(volume + pip.current_volume)
+        pip.set_liquid_class_by_name(new_class)
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
@@ -16,9 +16,13 @@ from typing import (
 from typing_extensions import Final
 import numpy
 from opentrons_shared_data.pipette.dev_types import UlPerMmAction
+
 from opentrons_shared_data.errors.exceptions import (
     CommandPreconditionViolated,
     CommandParameterLimitViolated,
+)
+from opentrons_shared_data.pipette.pipette_definition import (
+    liquid_class_for_volume_between_default_and_defaultlowvolume,
 )
 
 from opentrons import types as top_types
@@ -905,5 +909,9 @@ class OT3PipetteHandler:
         if pip.current_volume > 0:
             # Switching liquid classes can't happen when there's already liquid
             return
-        new_class = pip.get_liquid_class_for_volume(volume + pip.current_volume)
-        pip.set_liquid_class_by_name(new_class)
+        new_class_name = liquid_class_for_volume_between_default_and_defaultlowvolume(
+            volume,
+            pip.liquid_class_name,
+            pip.config.liquid_properties,
+        )
+        pip.set_liquid_class_by_name(new_class_name.name)
--- a/api/src/opentrons/protocol_engine/actions/actions.py
+++ b/api/src/opentrons/protocol_engine/actions/actions.py
@@ -14,8 +14,7 @@ from opentrons.hardware_control.modules import LiveData
 
 from opentrons_shared_data.errors import EnumeratedError
 
-from ..resources import pipette_data_provider
-from ..commands import Command, CommandCreate
+from ..commands import Command, CommandCreate, CommandPrivateResult
 from ..types import LabwareOffsetCreate, ModuleDefinition, Liquid
 
 
@@ -111,6 +110,7 @@ class UpdateCommandAction:
     """Update a given command."""
 
     command: Command
+    private_result: CommandPrivateResult
 
 
 @dataclass(frozen=True)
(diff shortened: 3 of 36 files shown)


api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -189,6 +189,15 @@ class AddPipetteConfigAction:
     config: pipette_data_provider.LoadedStaticPipetteData
 
 
+@dataclass(frozen=True)
+class UpdateLiquidClassAction:
+    """Update a pipette's current static config class based on a new liquid class."""
+
+    pipette_id: str
+    serial_number: str
```
sfoster1: the serial number definitely shouldn't be in here - that would make it impossible for the action to exist in a json protocol that won't know the serial number

api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -189,6 +189,15 @@ class AddPipetteConfigAction:
...
+    config: pipette_data_provider.LoadedStaticPipetteData
```
sfoster1: This also doesn't seem right - the payload here should be the liquid class name and not much else. That will get executed by being passed down to the hardware.

sfoster1 (changes requested):
I don't think this is quite what we want. We want an action that changes the liquid class by sending over the new name, and is handled by calling the `set_liquid_class` method in hardware and then updating the loaded pipette config from hardware (or calling an equivalent function when simulating).

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -136,6 +136,28 @@ def __init__(
...
+        has_lvd = (
+            pip_types.LiquidClasses.lowVolumeDefault in self._config.liquid_properties
+        )
+        if not has_lvd:
+            return self._liquid_class_name.name
+        if self._liquid_class_name == pip_types.LiquidClasses.lowVolumeDefault:
+            if volume > (self.liquid_class.max_volume * 0.75):
+                return pip_types.LiquidClasses.default.name
+        if self._liquid_class_name == pip_types.LiquidClasses.default:
+            if volume < (self.liquid_class.max_volume * 0.5):
```
sfoster1: I think this wants to be < than half the max volume of the low volume default

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +340,106 @@ async def load_module(
...
+            pipette_name: The pipette name.
+            mount: The mount on which pipette must be attached.
+            pipette_id: An optional identifier to assign the pipette. If None, an
+                identifier will be generated.
+            volume: The volume to configure the pipette for
+
+        Returns:
+            A LoadedConfiguredVolumeData object.
+        """
+        # TODO (spp, 2023-05-10): either raise error if using MountType.EXTENSION in
```
sfoster1: copypasta

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +340,106 @@ async def load_module(
...
+
+            # TODO(mc, 2020-10-18): calling `cache_instruments` mirrors the
+            # behavior of protocol_context.load_instrument, and is used here as a
+            # pipette existence check
+            try:
+                await self._hardware_api.cache_instruments(cache_request)
+            except RuntimeError as e:
+                raise FailedToLoadPipetteError(str(e)) from e
+
+            pipette_dict = self._hardware_api.get_attached_instrument(
```
sfoster1: don't need to do this

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
```diff
@@ -0,0 +1,89 @@
...
+    mount: MountType = Field(
+        ...,
+        description="The mount the pipette should be present on.",
+    )
+
+
+class ConfigureForVolumeResult(BaseModel):
+    """Result data from execution of an ConfigureForVolume command."""
+
+    pipetteId: str = Field(
```
TamarZanzouri, line 41: why do we need the pipetteId in the result?

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +340,106 @@ async def load_module(
...
+                    other_pipette.pipetteName.value
+                    if isinstance(other_pipette.pipetteName, PipetteNameType)
+                    else other_pipette.pipetteName
+                )
+
+            # TODO(mc, 2020-10-18): calling `cache_instruments` mirrors the
+            # behavior of protocol_context.load_instrument, and is used here as a
+            # pipette existence check
+            try:
+                await self._hardware_api.cache_instruments(cache_request)
```
sfoster1: don't need this

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +340,106 @@ async def load_module(
...
+        pipette_id = pipette_id or self._model_utils.generate_id()
+
+        if not use_virtual_pipettes:
+            cache_request = {mount.to_hw_mount(): pipette_name_value}
+
+            # TODO(mc, 2022-12-09): putting the other pipette in the cache request
+            # is only to support protocol analysis, since the hardware simulator
+            # does not cache requested virtual instruments. Remove per
+            # https://opentrons.atlassian.net/browse/RLIQ-258
+            other_mount = mount.other_mount()
```
sfoster1: don't need the other mount and cache pipette stuff

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +340,106 @@ async def load_module(
...
+            serial_number = pipette_dict["pipette_id"]
+            static_pipette_config = pipette_data_provider.get_pipette_static_config(
+                pipette_dict
+            )
+
+        else:
+            is_p50_flex_pipette = (
+                "p50" in pipette_name_value and "flex" in pipette_name_value
+            )
+            if volume < 5 and is_p50_flex_pipette:
```
sfoster1: we should use the exact same logic - maybe make `liquid_class_for_volume` a free function taking 
- the requested volume
- the current liquid class
and reutrning a liquid class, that can get called both here and in hardware?

sfoster1 (changes requested):
Looks like exactly what we want, save the copypasta in the command implementation, sharing the logic, and fixing that one hysteresis logic defn

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +340,106 @@ async def load_module(
...
+            # is only to support protocol analysis, since the hardware simulator
+            # does not cache requested virtual instruments. Remove per
+            # https://opentrons.atlassian.net/browse/RLIQ-258
+            other_mount = mount.other_mount()
+            other_pipette = self._state_store.pipettes.get_by_mount(other_mount)
+            if other_pipette is not None:
+                cache_request[other_mount.to_hw_mount()] = (
+                    other_pipette.pipetteName.value
+                    if isinstance(other_pipette.pipetteName, PipetteNameType)
+                    else other_pipette.pipetteName
```
TamarZanzouri: as we spoke about in the morning call. I guess we can extract this logic to a method bc we are doing the same for the `load_pipette`?

sfoster1: we don't need to do this logic in here at all I think

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
sfoster1: we don't really, might remove it

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -899,3 +903,15 @@ def get_pipette(self, mount: OT3Mount) -> Pipette:
     async def set_liquid_class(self, mount: OT3Mount, liquid_class: str) -> None:
         pip = self.get_pipette(mount)
         pip.set_liquid_class_by_name(liquid_class)
+
+    async def configure_for_volume(self, mount: OT3Mount, volume: float) -> None:
+        pip = self.get_pipette(mount)
+        if pip.current_volume > 0:
+            # Switching liquid classes can't happen when there's already liquid
+            return
```
TamarZanzouri, line 911: do we want to raise here instead?

TamarZanzouri (approved):
thank you for the changes! this looks great to me!

api/src/opentrons/protocol_engine/commands/command_unions.py
```diff
@@ -500,3 +514,5 @@
     calibration.CalibrateModuleResult,
     calibration.MoveToMaintenancePositionResult,
 ]
+
+CommandPrivateResult = Union[None, LoadPipettePrivateResult]
```
TamarZanzouri: are we missing `ConfigureForVolumePrivateResult` ?

api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -178,3 +180,41 @@ def __init__(
     async def execute(self, params: CommandParamsT) -> CommandResultT:
         """Execute the command, mapping data from execution into a response model."""
         ...
+
+
+class AbstractCommandWithPrivateResultImpl(
+    ABC,
+    Generic[CommandParamsT, CommandResultT, CommandPrivateResultT],
+):
+    """Abstract command creation and execution implementation.
```
jbleon95: Doc strings for this new Abstract Implementation should be updated to not just be copypasta'd from above.

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
```diff
@@ -0,0 +1,96 @@
...
+class ConfigureForVolumePrivateResult(PipetteConfigUpdateResultMixin):
+    """Result sent to the store but not serialized."""
+
+    pass
+
+
+class ConfigureForVolumeResult(BaseModel):
+    """Result data from execution of an ConfigureForVolume command."""
+
+    pipetteId: str = Field(
```
jbleon95: Is there a reason why were are returning pipetteId in the result, because unless it's changing I don't think it should be included.

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +342,52 @@ async def load_module(
...
 
+    async def configure_for_volume(
+        self,
+        pipette_id: str,
+        volume: float,
+    ) -> LoadedConfigureForVolumeData:
+        """Ensure the requested volume can be configured for the given pipette.
+
+        Args:
+            pipette_id: An optional identifier to assign the pipette. If None, an
```
jbleon95: Not optional for this method

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -181,9 +187,9 @@ def _handle_command(self, command: Command) -> None:  # noqa: ANN101, C901
                         attached_tip.volume
                     ]
                 except KeyError:
-                    tip_configuration = static_config.tip_configuration_lookup_table[
-                        static_config.max_volume
-                    ]
+                    tip_configuration = list(
```
TamarZanzouri, line 195: do we want to leave a comment about defaulting?

api/src/opentrons/protocol_engine/execution/equipment.py
```diff
@@ -330,6 +342,52 @@ async def load_module(
...
+                pipette_dict
+            )
+
+        else:
+            model = self._state_store.pipettes.get_model_name(pipette_id)
+            self._virtual_pipette_data_provider.configure_virtual_pipette_for_volume(
+                pipette_id, volume, model
+            )
+
+            serial_number = self._model_utils.generate_id(prefix="fake-serial-number-")
```
jbleon95, line 379: the fact that we are regenerating serial number here for the simulated case feels a bit iffy even though in practice I don't believe it will have any actionable impact on how we do analysis. The only place we are using serial number is for `get_calibrated_tip_length` which already is referencing a fake serial number in the first place. Still just calling it out here for maybe a comment or not explaining this.

TamarZanzouri:
looks great to me! had a few questions but overall love the private setting change :-)

api/tests/opentrons/protocol_engine/execution/test_equipment_handler.py
```diff
@@ -69,11 +69,27 @@ def _make_config(use_virtual_modules: bool) -> Config:
...
 def patch_mock_pipette_data_provider(
-    decoy: Decoy, monkeypatch: pytest.MonkeyPatch
+    decoy: Decoy,
+    monkeypatch: pytest.MonkeyPatch,
+    virtual_pipette_data_provider: pipette_data_provider.VirtualPipetteDataProvider,
 ) -> None:
     """Mock out move_types.py functions."""
     for name, func in inspect.getmembers(pipette_data_provider, inspect.isfunction):
         monkeypatch.setattr(pipette_data_provider, name, decoy.mock(func=func))
+    monkeypatch.setattr(
```
jbleon95: Since `VirtualPipetteDataProvider` is now a class we should avoid monkeypatching it here and instead make the below pytest fixture a decoy mock of that class, unless there is some functional reason why we are doing it this way.

jbleon95 (approved):
Reviewed the Protocol Engine portions of this PR, all look good to me outside of minor comments and fixing up the failing CI tests. Really like the private result addition, we can definitely use that in the future to clean up some result cruft.

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
Laura-Danielle: All the pipette commands include the pipette id.

Laura-Danielle (commented):
Cannot approve my own pull request but I like the private results update a lot.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
sfoster1: yeah probably

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
SyntaxColoring: Mm, `loadPipette` has a `result.pipetteId` because that's the thing that creates the `pipetteId`. Other pipetting commands, like `aspirate` and `blowOut`, do not. (They instead take a `pipetteId` as input, in `params`.)

I agree with @jbleon95—let's remove `pipetteId` from this result model.

api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -25,6 +25,8 @@
 
 CommandResultT = TypeVar("CommandResultT", bound=BaseModel)
 
+CommandPrivateResultT = TypeVar("CommandPrivateResultT", bound=BaseModel)
+
```
SyntaxColoring: Low-priority nitpick: I would not bind this to `BaseModel`, because `BaseModel` is for stuff that's serialized, and private results are deliberately never serialized.

And I guess I'd make `ConfigureForVolumePrivateResult` a `@dataclass` instead of a `BaseModel`.

api/src/opentrons/protocol_engine/commands/load_pipette.py
```diff
@@ -46,22 +57,32 @@ class LoadPipetteResult(BaseModel):
...
         )
 
-        return LoadPipetteResult(pipetteId=loaded_pipette.pipette_id)
+        return LoadPipetteResult(
+            pipetteId=loaded_pipette.pipette_id
+        ), LoadPipettePrivateResult(
+            pipette_id=loaded_pipette.pipette_id,
+            serial_number=loaded_pipette.serial_number,
+            config=loaded_pipette.static_config,
+        )
```
SyntaxColoring, line 85: Nitpick for slightly faster analysis:
 
```suggestion
        return LoadPipetteResult.construct(
            pipetteId=loaded_pipette.pipette_id
        ), LoadPipettePrivateResult.construct(
            pipette_id=loaded_pipette.pipette_id,
            serial_number=loaded_pipette.serial_number,
            config=loaded_pipette.static_config,
        )
```

SyntaxColoring (commented):
Generally looks great to me. Thanks!

sfoster1 (approved):
get rid of my own negative review

api/src/opentrons/protocol_engine/commands/command_unions.py
sfoster1: we were!

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
sfoster1: done

api/src/opentrons/protocol_engine/commands/command.py
sfoster1: that's a good point, there's no reason for it to be like that

sfoster1: honestly there's not really a reason to have a bound at all in that case tbh

PR #13474, opened 2023-09-06
feat(app): acknowledge animation when Flex receives a new protocol

fsinapi:
# Overview

The design spec for the status bar calls for the Confirmation animation to play when a new protocol is uploaded. We already have 1) a [stateless protocol engine command](https://github.com/Opentrons/opentrons/pull/12890) to run that animation, and 2) a hook in the ODD app to detect when a new protocol is uploaded. This PR just glues those two things together.

# Test Plan

Pushed this to Flexy and uploaded a protocol.
* If the robot is idle (status bar is plain white, not running a protocol) when it is uploaded, the animation plays
* If the robot is in the middle of a protocol run, the status bar does _not_ play the animation, instead showing the status of the running protocol

# Changelog

- Added definitions for the Status Bar Animation command to schemaV7 in shared-data
- In the hook listening for new protocol uploads, send a status bar animation command to run the `confirm` animation if a new protocol was uploaded

# Review requests

- Does the command definition stuff in shared-data need to be filled in anywhere else that I'm missing?

# Risk assessment

Pretty low, not core behavior

### DIFF
--- a/app/src/App/hooks.ts
+++ b/app/src/App/hooks.ts
@@ -109,7 +109,7 @@ export function useProtocolReceiptToast(): void {
           createLiveCommand({
             command: animationCommand,
           }).catch((e: Error) =>
-            console.warn(`cannot run status bar animtion: ${e.message}`)
+            console.warn(`cannot run status bar animation: ${e.message}`)
           )
         })
         .catch((e: Error) => {
--- a/app/src/App/hooks.ts
+++ b/app/src/App/hooks.ts
@@ -9,6 +9,7 @@ import {
   useAllRunsQuery,
   useHost,
   useRunQuery,
+  useCreateLiveCommandMutation,
 } from '@opentrons/react-api-client'
 import {
   getProtocol,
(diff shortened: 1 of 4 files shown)


shlokamin (approved):
lgtm! only complaint is the `Incidental` naming cuz its kinda vague, but honestly i dont have anything better and we can change the name later if we want. thx frank!

app/src/App/hooks.ts
```diff
@@ -98,6 +105,13 @@ export function useProtocolReceiptToast(): void {
               console.error(`error invalidating protocols query: ${e.message}`)
             )
         })
+        .then(() => {
+          createLiveCommand({
+            command: animationCommand,
+          }).catch((e: Error) =>
+            console.warn(`cannot run status bar animtion: ${e.message}`)
```
jerader:
```suggestion
            console.warn(`cannot run status bar animation: ${e.message}`)
```

fsinapi: yes very much indeed!

PR #13477, opened 2023-09-06
feat(app): create module instruction modal and delete H-S wizard

jerader:
closes RQA-1299

# Overview

This deletes the H-S wizard and creates a module instruction modal instead. 

# Test Plan

Test that the modal shows up correctly via the:
1. heater-shaker module overflow menu for the Ot-2
2. heater-shaker module overflow menu's test shake slideout
3. module setup list (go to protocol setup with a protocol with a H-S)

Double check that the i18n keys and images I deleted are not in use.

# Changelog

- delete old code for the wizard that is no longer needed
- create new modal `ModuleSetupModal` and test

# Review requests

see test plan

# Risk assessment

low


ecormany:
Copy looks good in app. Link text could just say "Module setup instructions" instead of filling in the module name, since it goes to the modules landing page on the help center site. But OK as-is.

The QR code still goes to example.com and needs to be updated to target the same URL as the text link.

jerader:
> The QR code still goes to example.com and needs to be updated to target the same URL as the text link.

Thanks @ecormany ! Ya, idk who is in charge of the QR code but I left a comment on the ticket.

mjhuff (approved):
Code lgtm and tested per your instructions. Got to love all that red! 💯

PR #13478, opened 2023-09-06
feat(hardware): allow variable number of tip presence messages

caila-marashaj:
## Overview
Currently, the hardware controller is only able to process one tip presence notification. This changes `tip_presence.py` to wait for and return either 1 or 2 notifications, as a `List[int]` rather than an `int`. 

At the hardware controller level, we check that both responses are the same, and then handle the tip presence status as a single value.

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -6,7 +6,6 @@ from contextlib import asynccontextmanager
 from functools import wraps
 import logging
 from copy import deepcopy
-from typing_extensions import Literal
 from typing import (
     Any,
     Awaitable,
@@ -857,12 +856,11 @@ class OT3Controller:
         expected_responses = 2 if expect_multiple_responses else 1
         node = sensor_node_for_mount(OT3Mount(mount.value))
         assert node != NodeId.gripper
-        res = await get_tip_ejector_state(
-            self._messenger, node, expected_responses
-        )
+        res = await get_tip_ejector_state(self._messenger, node, expected_responses)  # type: ignore[arg-type]
         vals = list(res.values())
         if not all([r == vals[0] for r in vals]):
-            raise UnmatchedTipPresenceStates()
+            states = {int(sensor): res[sensor] for sensor in res.keys()}
+            raise UnmatchedTipPresenceStates(states)
         tip_present_state = bool(vals[0]) and bool(vals[-1])
         return tip_present_state
 
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -6,6 +6,7 @@ from contextlib import asynccontextmanager
 from functools import wraps
 import logging
 from copy import deepcopy
+from typing_extensions import Literal
 from typing import (
     Any,
     Awaitable,
@@ -125,7 +126,6 @@ from opentrons.hardware_control.types import (
     SubSystem,
     TipStateType,
     FailedTipStateCheck,
-    UnmatchedTipStates,
     EstopState,
 )
 from opentrons.hardware_control.errors import (
@@ -164,6 +164,7 @@ from opentrons_shared_data.gripper.gripper_definition import GripForceProfile
 from opentrons_shared_data.errors.exceptions import (
     EStopActivatedError,
     EStopNotPresentError,
+    UnmatchedTipPresenceStates,
 )
 
 from .subsystem_manager import SubsystemManager
@@ -848,15 +849,19 @@ class OT3Controller:
             raise FailedTipStateCheck(tip_state, res)
 
     async def get_tip_present_state(
-        self, mount: OT3Mount, expect_multiple_responses: bool = False
+        self,
+        mount: OT3Mount,
+        expect_multiple_responses: bool = False,
     ) -> bool:
         """Get the state of the tip ejector flag for a given mount."""
         expected_responses = 2 if expect_multiple_responses else 1
+        node = sensor_node_for_mount(OT3Mount(mount.value))
+        assert node != NodeId.gripper
         res = await get_tip_ejector_state(
-            self._messenger, sensor_node_for_mount(OT3Mount(mount.value)), expected_responses  # type: ignore
+            self._messenger, node, Literal[expected_responses]
         )
         if res[0] != res[-1]:
-            raise UnmatchedTipStates()
+            raise UnmatchedTipPresenceStates()
         tip_present_state = bool(res[0]) and bool(res[-1])
         return tip_present_state
 
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -836,7 +836,7 @@ class OT3Controller:
         res = await get_limit_switches(self._messenger, motor_nodes)
         return {node_to_axis(node): bool(val) for node, val in res.items()}
 
-    async def get_tip_present(
+    async def check_for_tip_presence(
         self,
         mount: OT3Mount,
         tip_state: TipStateType,
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -861,7 +861,7 @@ class OT3Controller:
         if not all([r == vals[0] for r in vals]):
             states = {int(sensor): res[sensor] for sensor in res}
             raise UnmatchedTipPresenceStates(states)
-        tip_present_state = bool(vals[0]) and bool(vals[-1])
+        tip_present_state = bool(vals[0])
         return tip_present_state
 
     @staticmethod
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -835,16 +835,24 @@ class OT3Controller:
         res = await get_limit_switches(self._messenger, motor_nodes)
         return {node_to_axis(node): bool(val) for node, val in res.items()}
 
-    async def get_tip_present(self, mount: OT3Mount, tip_state: TipStateType) -> None:
+    async def get_tip_present(
+        self,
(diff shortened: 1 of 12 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -835,16 +835,24 @@ async def get_limit_switches(self) -> OT3AxisMap[bool]:
         res = await get_limit_switches(self._messenger, motor_nodes)
         return {node_to_axis(node): bool(val) for node, val in res.items()}
 
-    async def get_tip_present(self, mount: OT3Mount, tip_state: TipStateType) -> None:
+    async def get_tip_present(
+        self,
+        mount: OT3Mount,
+        tip_state: TipStateType,
+        is_ninety_six_channel: bool = False,
```
sfoster1: i'd love to see this changed to like `expect_multiple_responses` or something

sfoster1 (approved):
Looks good, though I would like to change the name of the backend argument - I feel like ideally it's just the hardware controller that knows about the 96

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1870,11 +1870,11 @@ async def pick_up_tip(
         # TODO: implement tip-detection sequence during pick-up-tip for 96ch,
         #       but not with DVT pipettes because those can only detect drops
 
-        if (
-            self.gantry_load != GantryLoad.HIGH_THROUGHPUT
-            and ff.tip_presence_detection_enabled()
-        ):
-            await self._backend.get_tip_present(realmount, TipStateType.PRESENT)
+        if ff.tip_presence_detection_enabled():
```
Laura-Danielle: we're not quite ready to enable 96 channel tip presence. We need the motor routine to do this.

Laura-Danielle (dismissed):
We should still not be checking tip presence for the 96 chan until we implement the motor routine.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -835,16 +835,24 @@ async def get_limit_switches(self) -> OT3AxisMap[bool]:
...
         """Raise an error if the expected tip state does not match the current state."""
-        res = await self.get_tip_present_state(mount)
-        if res != tip_state.value:
-            raise FailedTipStateCheck(tip_state, res)
+        res = await self.get_tip_present_state(mount, expect_multiple_responses)
+        if res[0] != tip_state.value or res[0] != res[1]:
+            raise FailedTipStateCheck(tip_state, res[0])
 
-    async def get_tip_present_state(self, mount: OT3Mount) -> int:
+    async def get_tip_present_state(
```
Laura-Danielle, line 850: Imo I don't think we should return a list here. We should `AND` the result if there are two values in the list. And then we can keep the rest of the logic.

Otherwise, you'll have to do some funky stuff with getting the correct state for `pipette_state_for_mount`

caila-marashaj: agree, that makes sense

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -835,16 +835,24 @@ async def get_limit_switches(self) -> OT3AxisMap[bool]:
         res = await get_limit_switches(self._messenger, motor_nodes)
         return {node_to_axis(node): bool(val) for node, val in res.items()}
 
-    async def get_tip_present(self, mount: OT3Mount, tip_state: TipStateType) -> None:
+    async def get_tip_present(
```
ahiuchingau: I know this func was created before this PR, but if it's not too much work we really should rename this function since it's not returning a value

ahiuchingau (approved):
One nit otherwise looks good to me!

api/src/opentrons/hardware_control/backends/ot3controller.py
Laura-Danielle: can call it `check_for_tip_presence` or something

api/src/opentrons/hardware_control/types.py
```diff
@@ -642,3 +642,13 @@ def __init__(self, tip_state_type: TipStateType, actual_state: int) -> None:
             f"Failed to correctly determine tip state for tip {str(tip_state_type)} "
             f"received {bool(actual_state)} but expected {bool(tip_state_type.value)}"
         )
+
+
+class UnmatchedTipStates(RuntimeError):
```
sfoster1: We should make this an enumerated error - something new in the 1xxx series of errors in `shared-data/errors/exceptions` and codes and such.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -835,18 +836,29 @@ async def get_limit_switches(self) -> OT3AxisMap[bool]:
...
+    ) -> bool:
         """Get the state of the tip ejector flag for a given mount."""
+        expected_responses = 2 if expect_multiple_responses else 1
         res = await get_tip_ejector_state(
-            self._messenger, sensor_node_for_mount(OT3Mount(mount.value))  # type: ignore
+            self._messenger, sensor_node_for_mount(OT3Mount(mount.value)), expected_responses  # type: ignore
         )
-        return res
+        if res[0] != res[-1]:
+            raise UnmatchedTipStates()
```
sfoster1: this exception should contain the two conflicting states explicitly

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -835,18 +836,29 @@ async def get_limit_switches(self) -> OT3AxisMap[bool]:
...
+        self, mount: OT3Mount, expect_multiple_responses: bool = False
+    ) -> bool:
         """Get the state of the tip ejector flag for a given mount."""
+        expected_responses = 2 if expect_multiple_responses else 1
         res = await get_tip_ejector_state(
-            self._messenger, sensor_node_for_mount(OT3Mount(mount.value))  # type: ignore
+            self._messenger, sensor_node_for_mount(OT3Mount(mount.value)), expected_responses  # type: ignore
         )
-        return res
+        if res[0] != res[-1]:
```
sfoster1: i like this, it's elegant. however, i think it's better to be explicit here and check that all responses are the same value by doing something like `if not all([r == res[0] for r in res])`

api/src/opentrons/hardware_control/types.py
```diff
@@ -642,3 +642,13 @@ def __init__(self, tip_state_type: TipStateType, actual_state: int) -> None:
...
+    """Error raised if a 96 channel pipette receives two differing tip presence statuses."""
+
+    def __init__(self) -> None:
+        """Iniitialize FailedTipStateCheck error."""
```
sfoster1:
```suggestion
        """Initialize FailedTipStateCheck error."""
```

hardware/opentrons_hardware/hardware_control/tip_presence.py
```diff
@@ -21,35 +27,45 @@
 async def get_tip_ejector_state(
     can_messenger: CanMessenger,
     node: Literal[NodeId.pipette_left, NodeId.pipette_right],
-) -> int:
+    expected_responses: int = 1,
```
sfoster1: one thing we might want to consider here is typing this function with overloads like this:
```python
@typing.overload
async def get_tip_ejector_state(can_messenger: .., node: Literal[...], expected_responses: Literal[1] timeout: ...) -> Tuple[int]
    ...

@typing.overload
async def get_tip_ejector_state(can_messenger: ..., node: Literal[...], expected_responses: Literal[2], timeout: ...) -> Tuple[int, int]:
    ...

async def get_tip_ejector_state(): ...code...
```

This makes it explicit that callers can expect the same length return value as the number they fed in.

sfoster1 (approved):
Looks good once @Laura-Danielle 's changes are addressed, some nitpicks/improvements

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -835,18 +836,33 @@ async def get_limit_switches(self) -> OT3AxisMap[bool]:
...
-        return res
+        expected_responses = 2 if expect_multiple_responses else 1
+        node = sensor_node_for_mount(OT3Mount(mount.value))
+        assert node != NodeId.gripper
+        res = await get_tip_ejector_state(self._messenger, node, expected_responses)  # type: ignore[arg-type]
+        vals = list(res.values())
+        if not all([r == vals[0] for r in vals]):
+            states = {int(sensor): res[sensor] for sensor in res}
+            raise UnmatchedTipPresenceStates(states)
+        tip_present_state = bool(vals[0]) and bool(vals[-1])
```
sfoster1: at this point we know all entries in `vals` are the same (if not, we would have raised the exception) so we can just do `bool(vals[0])`

shared-data/errors/definitions/1/errors.json
```diff
@@ -181,6 +181,10 @@
     "4003": {
       "detail": "Not supported on this robot type",
       "category": "generalError"
+    },
+    "4004": {
+      "detail": "Unmatched tip presence states",
+      "category": "roboticsControlError"
```
sfoster1: if this is in the `roboticsControlerror` category, its error number should be in that category's error number - presumably 2012, as befits its apocalyptic nature

shared-data/python/opentrons_shared_data/errors/codes.py
```diff
@@ -75,6 +75,7 @@ class ErrorCodes(Enum):
     ROBOT_IN_USE = _code_from_dict_entry("4001")
     API_REMOVED = _code_from_dict_entry("4002")
     NOT_SUPPORTED_ON_ROBOT_TYPE = _code_from_dict_entry("4003")
+    UNMATCHED_TIP_PRESENCE_STATES = _code_from_dict_entry("4004")
```
sfoster1: `2012` probably, or matching whatever you change it to above

sfoster1 (changes requested):
Couple more small changes, only really important one is the error code

sfoster1 (approved):
Noice! Looks good to me if tests pass.

PR #13479, opened 2023-09-06
chore(api): go back to full blowout

sfoster1:
We added a default blowout volume to fix essentially the same issue that
push_out fixes better. Now that we have push_out, let's go back to the
full-length blowout as default (though we retain the ability to override
it on command).

This removes the `defaultBlowoutVolume`  from the pipette configs and makes the code go back to doing a full-volume blowout, on the recommendation of hardware.


Laura-Danielle (approved):
nice, lgtm.

PR #13480, opened 2023-09-06
fix(update-server): check that total rootfs size is greater and not equal to the partition size.

vegano1:
# Overview

The rootfs size is equal to the partition size, so let's make sure we only fail if the new rootfs is greater than the partition size.

# Test Plan

- [x] Create a build larger than the rootfs partition size and make sure the update fails
- [x] Create a build smaller or equal to the rootfs partition size and make sure the update passes 

# Changelog

- Check that rootfs size is greater and not greater than or equal to rootfs partition size

# Review requests

# Risk assessment
Low


sfoster1 (approved):
Looks good to me!

PR #13483, opened 2023-09-06
fix(app): Fix overeager maintenance run takeover modal

mjhuff:
# Overview

Fixes hanging maintenance run takeover modals after a delete maintenance run request has been sent to the server. Awaiting a successful response is redundant, since the currentRunId is polled and will be null upon successful deletion. Checking the status introduces odd edge cases where the currentRunId clears, but the takeover modal hangs unnecessarily.

# Test Plan
- Bug identified on the physical flex. While recreating the exact conditions in which this bug appeared are difficult, it has been confirmed that this change does not impact normal functionality.

# Changelog

- Removed deleteMaintenanceRun status check.

# Risk assessment

low

### DIFF
--- a/app/src/organisms/TakeoverModal/MaintenanceRunTakeover.tsx
+++ b/app/src/organisms/TakeoverModal/MaintenanceRunTakeover.tsx
@@ -51,7 +51,7 @@ export function MaintenanceRunTakeoverModal(
   }
 
   React.useEffect(() => {
-    if (currentRunId == null && status === 'success') {
+    if (currentRunId == null) {
       setIsLoading(false)
       setShowConfirmTerminateModal(false)
       reset()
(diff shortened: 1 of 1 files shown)


app/src/organisms/TakeoverModal/MaintenanceRunTakeover.tsx
```diff
@@ -51,7 +51,7 @@ export function MaintenanceRunTakeoverModal(
   }
 
   React.useEffect(() => {
-    if (currentRunId == null && status === 'success') {
+    if (currentRunId == null) {
```
koji, line 50: We don't need to add `status` as a dependency, do we?

mjhuff: Woops, good catch!

koji (approved):
lgtm!

PR #13384, opened 2023-08-25
docs(api): refactor and rewrite Complex Commands

ecormany:
# Overview

Bringing the docs page about our most-used Python API command (and friends) up to par with the rest of the docs site.

# Test Plan

- [x] Proofread prose in the [sandbox](http://sandbox.docs.opentrons.com/docs-refactor-complex-commands/v2/new_complex_commands.html).
- [x] Verify that all RST anchors in deleted lines were properly moved/replaced.

# Changelog

- Replace long Complex Commands page with landing page
- Three new pages, linked from the landing page:
  - Sources and Destinations: new and more accurate description of how transfer, distribute, and consolidate match up source and destination wells (with tables!)
  - Order of Operations: page dedicated to the order of all the possible actions in complex commands, and how they do or don't repeat when the pipette has to be refilled
  - Complex Liquid Handling Parameters: way way more about these kwargs than could fit in a docstring (2000 words 😅). **lots of detailed discussion of nuanced and unintuitive behavior here** 🐉 
- Move boilerplate protocol code to new section of Protocol Examples. We plan to use this across the docs site.
- API Reference
  - Clean up `transfer()` docstring, including links to new docs pages.
  - Remove two features from `transfer()` docstring because they haven't worked since API v1 😬 
  - Clean up `distribute()` and `consolidate()` docstrings, still largely relying on link to `transfer()`

# Review requests

This is a massive rewrite. TC can handle the prose proofreading.

If you've worked on some feature of complex commands as an engineer, please drill down to the relevant section. Feedback appreciated on how easy these pages are to navigate, and whether my rewrites are accurate to API behavior.

And an API reference style question: how should we order long lists of kwargs? Logically (i.e., matching the Complex Liquid Handling Parameters page)? Alphabetically?

# Risk assessment

Low in general. Mid for docs (much potential for broken links).

PR #13485, opened 2023-09-06
feat(api): add configure_for_volume

sfoster1:
This function configures a pipette in the appropriate mode to handle a specific volume. It's needed for p50 gen3 pipettes to handle very low volumes by changing a number of settings, and needs to be exposed as a separate function because it can cause motion like prepare_for_aspirate().

## Testing

- Run on a robot! Make sure that the low volume pipetting profile gets selected.
Note: this will be tested on hardware much more... extensively.

### DIFF
--- a/api/docs/v2/new_pipette.rst
+++ b/api/docs/v2/new_pipette.rst
@@ -331,7 +331,7 @@ In a protocol that handles many different volumes, it's a good practice to call 
     for i in range(8):
         pipette50.pick_up_tip()
         pipette50.configure_for_volume(volumes[i])
-        pipette50.aspirate(volume=volume, location=sources[i])
+        pipette50.aspirate(volume=volumes[i], location=sources[i])
         pipette50.dispense(location=destinations[i])
         pipette50.drop_tip()
 
--- a/api/docs/v2/new_pipette.rst
+++ b/api/docs/v2/new_pipette.rst
@@ -303,6 +303,41 @@ The OT-2 works with the GEN1 and GEN2 pipette models. The newer GEN2 pipettes ha
 
 The single- and multi-channel P50 GEN1 pipettes are the exceptions here. If your protocol uses a P50 GEN1 pipette, there is no backward compatibility with a related GEN2 pipette. To replace a P50 GEN1 with a corresponding GEN2 pipette, edit your protocol to load a P20 Single-Channel GEN2 (for volumes below 20 µL) or a P300 Single-Channel GEN2 (for volumes between 20 and 50 µL).
 
+.. _pipette-volume-modes:
+
+Volume Modes
+============
+
+The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a low-volume mode to accurately dispense 1 µL of liquid. Set low-volume mode by calling :py:meth:`.InstrumentContext.configure_for_volume` with the amount of liquid you plan to aspirate, in µL::
+
+    pipette50.configure_for_volume(1)
+    pipette50.pick_up_tip()
+    pipette50.aspirate(1, plate["A1"])
+    
+.. versionadded:: 2.15
+
+.. note::
+    The pipette must not contain liquid when you call ``configure_for_volume()``, or the API will raise an error.
+    
+    Also, if the pipette is in a well location that may contain liquid, it will move upward to ensure it is not immersed in liquid before changing its mode.
+
+In a protocol that handles many different volumes, it's a good practice to call this function immediately before each :py:meth:`.transfer` or :py:meth:`.aspirate`, specifying the volume that you are about to handle. When operating with a list of volumes, nest ``configure_for_volume()`` inside a ``for`` loop to ensure that the pipette is properly configured for each volume:
+
+.. code-block:: python
+    
+    volumes = [1, 2, 3, 4, 1, 5, 2, 8]
+    sources = plate.columns()[0]
+    destinations = plate.columns()[1]
+    for i in range(8):
+        pipette50.pick_up_tip()
+        pipette50.configure_for_volume(volumes[i])
+        pipette50.aspirate(volume=volume, location=sources[i])
+        pipette50.dispense(location=destinations[i])
+        pipette50.drop_tip()
+
+If you know that all your liquid handling will take place in a specific mode, then you can call ``configure_for_volume()`` just once with a nominal volume. Or if all the volumes correspond to the pipette's default mode, you don't have to call ``configure_for_volume()`` at all.
+
+
 .. _new-plunger-flow-rates:
 
 Pipette Flow Rates
@@ -343,9 +378,13 @@ Now let's change the flow rates for each action::
         pipette.aspirate(200, plate['A1'])  #  50 µL/s
         pipette.dispense(200, plate['A2'])  # 100 µL/s
         pipette.blow_out()                  #  75 µL/s
+        
+These flow rates will remain in effect until you change the ``flow_rate`` attribute again *or* call ``configure_for_volume()``. Calling ``configure_for_volume()`` always resets all pipette flow rates to the defaults for the mode that it sets.
+
+.. TODO add mode ranges and flow defaults to sections below
 
 .. note::
-    In API version 2.13 and earlier, :py:obj:`.InstrumentContext.speed` offered similar functionality. It attempted to set the plunger speed in mm/s. Due to technical limitations, that speed could only be approximate. You must use ``.flow_rate`` in version 2.14 and later, and you should consider replacing older code that sets ``.speed``.
+    In API version 2.13 and earlier, :py:obj:`.InstrumentContext.speed` offered similar functionality to ``.flow_rate``. It attempted to set the plunger speed in mm/s. Due to technical limitations, that speed could only be approximate. You must use ``.flow_rate`` in version 2.14 and later, and you should consider replacing older code that sets ``.speed``.
 
 .. versionadded:: 2.0
 
--- a/api/src/opentrons/protocol_api/core/instrument.py
+++ b/api/src/opentrons/protocol_api/core/instrument.py
@@ -228,5 +228,13 @@ class AbstractInstrument(ABC, Generic[WellCoreType]):
     ) -> None:
         ...
 
+    def configure_for_volume(self, volume: float) -> None:
+        """Configure the pipette for a specific volume.
+
+        Args:
+            volume: The volume to preppare to handle.
+        """
+        ...
+
 
 InstrumentCoreType = TypeVar("InstrumentCoreType", bound=AbstractInstrument[Any])
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1585,77 +1585,30 @@ class InstrumentContext(publisher.CommandPublisher):
 
     @requires_version(2, 15)
     def configure_for_volume(self, volume: float) -> None:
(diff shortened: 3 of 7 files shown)


Laura-Danielle (commented):
I'm good with the interface. Will re-review when implementation is done

TamarZanzouri (commented):
structure LGTM! will review again when its complete

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1578,3 +1582,93 @@ def __repr__(self) -> str:
...
+        if self._core.get_current_volume():
+            raise CommandPreconditionViolated(
+                message=f"Cannot switch modes of {str(self)} while it contains liquid"
+            )
+        if volume < 0:
+            raise CommandParameterLimitViolated(
+                command_name="configure_for_volume",
+                parameter_name="volume",
+                limit_statement="must be greater than 0",
+                actual_value=str(volume),
```
SyntaxColoring, line 1622: Ideally, this validation would live in Protocol Engine, right?

Not saying that has to happen in this PR. Just getting a sense of the ideal separation of responsibilities.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1578,3 +1582,93 @@ def __repr__(self) -> str:
 
     def __str__(self) -> str:
         return "{} on {} mount".format(self._core.get_display_name(), self.mount)
+
+    @requires_version(2, 15)
+    def configure_for_volume(self, volume: float) -> None:
+        """Configure the pipette to handle a specific volume of liquid.
+
+        Some pipettes have multiples of operation to handle certain volumes of liquid accurately.
+        For instance, the 50 µL GEN3 Single and Eight Channel pipettes must be placed in a low
```
SyntaxColoring: Nitpick: "GEN3" terminology is deprecated in favor of "Flex," I think.

api/src/opentrons/protocol_api/core/instrument.py
```diff
@@ -228,5 +228,13 @@ def set_flow_rate(
     ) -> None:
         ...
 
+    def configure_for_volume(self, volume: float) -> None:
+        """Configure the pipette for a specific volume.
+
+        Args:
+            volume: The volume to preppare to handle.
```
SyntaxColoring, line 235:
```suggestion
            volume: The volume to prepare to handle.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1578,3 +1582,93 @@ def __repr__(self) -> str:
 
     def __str__(self) -> str:
         return "{} on {} mount".format(self._core.get_display_name(), self.mount)
+
+    @requires_version(2, 15)
+    def configure_for_volume(self, volume: float) -> None:
+        """Configure the pipette to handle a specific volume of liquid.
+
+        Some pipettes have multiples of operation to handle certain volumes of liquid accurately.
```
SyntaxColoring:
```suggestion
        Some pipettes have multiple modes of operation to handle certain volumes of liquid accurately.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1578,3 +1582,93 @@ def __repr__(self) -> str:
...
+            instr = protocol.load_instrument('p50_single_gen3', 'left', tip_racks=[tiprack])
+            # this list might have been loaded from a CSV
+            for volume in [1, 2, 3, 4, 1, 5, 2, 8]:
+                instr.pick_up_tip()
+                # configuring for each volume before aspiration ensures the pipette is always in the
+                # right mode
+                instr.configure_for_volume(volume)
+                instr.aspirate(volume=volume, location=sample_plate['A1'])
+                instr.dispense(location=sample_plate['A1'])
+                instr.drop_tip()
```
SyntaxColoring: Nitpick: `tip_rack` instead of `tiprack`

```suggestion
            tip_rack = protocol.load_labware('opentrons_flex_96_tiprack_50ul', 'D1')
            sample_plate = protocol.load_labware('nest_1_reservoir_195ml', 'D2')
            instr = protocol.load_instrument('p50_single_gen3', 'left', tip_racks=[tip_rack])
            # this list might have been loaded from a CSV
            for volume in [1, 2, 3, 4, 1, 5, 2, 8]:
                instr.pick_up_tip()
                # configuring for each volume before aspiration ensures the pipette is always in the
                # right mode
                instr.configure_for_volume(volume)
                instr.aspirate(volume=volume, location=sample_plate['A1'])
                instr.dispense(location=sample_plate['A1'])
                instr.drop_tip()
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1578,3 +1582,93 @@ def __repr__(self) -> str:
...
+        .. code-block:: python
+
+            # A p300_multi_gen2 has only a default mode; you do not need to call this function
+            tiprack = protocol.load_labware('opentrons_96_tiprack_300ul', '1')
+            sample_plate = protocol.load_labware('nest_1_reservoir_195ml', '2')
+            instr = protocol.load_instrument('p300_multi_gen2', 'left', tip_racks=[tiprack])
+            for volume in [50, 10, 25, 300, 8]:
+                instr.pick_up_tip()
+                instr.aspirate(volume=volume, location=sample_plate['A1'])
+                instr.drop_tip()
```
SyntaxColoring: Nitpick: `tip_rack` instead of `tiprack`

```suggestion
            # A p300_multi_gen2 has only a default mode; you do not need to call this function
            tip_rack = protocol.load_labware('opentrons_96_tiprack_300ul', '1')
            sample_plate = protocol.load_labware('nest_1_reservoir_195ml', '2')
            instr = protocol.load_instrument('p300_multi_gen2', 'left', tip_racks=[tip_rack])
            for volume in [50, 10, 25, 300, 8]:
                instr.pick_up_tip()
                instr.aspirate(volume=volume, location=sample_plate['A1'])
                instr.drop_tip()
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1578,3 +1582,93 @@ def __repr__(self) -> str:
...
+
+            Changing the mode of a pipette may require a plunger movement. The pipette may move to ensure it
+            is not immersed in liquid.
+
+        :param volume:  - The volume to configure the pipette to handle
+        :type volume: float
+
+        :raises CommandPreconditionViolated: If called when liquid is held in the pipette
+        :raises CommandParameterLimitViolated: If called with a negative volume or a volume that the pipette
+                                               cannot handle in any mode.
```
SyntaxColoring: I think we should not document these exact exception types because it could look as if it's okay for users to `import` them and catch them in `except` blocks. These are a bit too deep for us to commit to maintaining them as part of the public Python Protocol API.

We could either:

* Document it vaguely, like "an exception," and keep raising what you have now.
* Document it precisely, but make the exception types into official public stable parts of the Python Protocol API. That means making them importable like `opentrons.protocol_api.FooException` or maybe `opentrons.protocol_api.exceptions.FooException`.
* Raise standard Python exception types instead, and document those? `CommandParameterLimitViolated` could be `ValueError`, but I'm not sure if `CommandPreconditionViolated` has a good substitute.

SyntaxColoring (approved):
Interface looks great, thanks. A few minor comments.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: I removed these and replaced them with general statements of when-errors-happen.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: Updated these to official product names.

api/docs/v2/new_pipette.rst
```diff
@@ -303,6 +303,41 @@ The OT-2 works with the GEN1 and GEN2 pipette models. The newer GEN2 pipettes ha
 
 The single- and multi-channel P50 GEN1 pipettes are the exceptions here. If your protocol uses a P50 GEN1 pipette, there is no backward compatibility with a related GEN2 pipette. To replace a P50 GEN1 with a corresponding GEN2 pipette, edit your protocol to load a P20 Single-Channel GEN2 (for volumes below 20 µL) or a P300 Single-Channel GEN2 (for volumes between 20 and 50 µL).
 
+.. _pipette-volume-modes:
+
+Volume Modes
+============
+
+The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a low-volume mode to accurately dispense 1 µL of liquid. Set low-volume mode by calling :py:meth:`.InstrumentContext.configure_for_volume` with the amount of liquid you plan to aspirate, in µL::
```
sfoster1, line 311: extra `:` at the end of the line unless it's a rst thing i don't know

api/docs/v2/new_pipette.rst
```diff
@@ -303,6 +303,41 @@ The OT-2 works with the GEN1 and GEN2 pipette models. The newer GEN2 pipettes ha
...
+
+.. code-block:: python
+    
+    volumes = [1, 2, 3, 4, 1, 5, 2, 8]
+    sources = plate.columns()[0]
+    destinations = plate.columns()[1]
+    for i in range(8):
+        pipette50.pick_up_tip()
+        pipette50.configure_for_volume(volumes[i])
+        pipette50.aspirate(volume=volume, location=sources[i])
```
sfoster1: `volumes[i]`

api/src/opentrons/protocol_api/instrument_context.py
sfoster1: @SyntaxColoring fyi @ecormany is taking a look at the docs for this, but thank you!

api/src/opentrons/protocol_api/instrument_context.py
sfoster1: Yeah. I had been implementing this as-if we were going to keep the PE command as "set liquid class", so the logic of which liquid class to use would be here; but in the event, we made it `ConfigureForVolume` so the validators should indeed be in the command.

api/docs/v2/new_pipette.rst
ecormany: it's an RST shorthand. you can do [double colon](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks) instead of explicitly declaring a `.. code-block`. it renders as a single colon.

api/docs/v2/new_pipette.rst
ecormany: ah yep, i got too clever. will fix.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: these got moved into docs pages, and the offending load statements have been removed. after #13384 merges, we're going to have a basic protocol template (with Flex and OT-2 flavors) on the Protocol Examples page, which we'll cross-link to from other pages. that will define what `pipette`, `plate`, etc. are across the site. 

that PR [does have](https://github.com/Opentrons/opentrons/pull/13384/files#diff-7ce4841635c3e08c388ceee76b0df6e89d78d8585495566d3559a360d8636d73R81) `tiprack` tho, so i can get that fixed over there. [update: maybe not, we've got a ton of tokens of `tiprack`, and it _is_ in the load names, which aren't changing]

api/docs/v2/new_pipette.rst
sfoster1: oh sick

PR #13495, opened 2023-09-07
fix(app,robot-server): Display less error barf

SyntaxColoring:
# Overview

This removes the more verbose error details from run failures.

Before this PR:

4cc079c9-9647-4b2f-a563-814447935329

With this PR: 

IMG_6568

Different error, but you get the idea. The `JSON.stringify(.errorInfo)` part has been removed.

Closes RQA-1491.

# Test Plan

I've tested this on a Flex on-device display. It should work exactly the same on the desktop app.

# Changelog

* On-device display: Don't show the `errorInfo` field.
* Desktop app: Don't show the `errorInfo` field.
* HTTP API: More specifically specify the fields of our error model.

  Going forward, reporting errors in a consistently good way will depend on us having a strong interface contract for them. This is an attempt to define that.

  The new specification does not necessarily match the robot's current behavior. In particular, I think the `detail` string sometimes be a gross concatenated frankenstein that redundantly includes the `errorCode` and `errorType`. Going forward, we would treat those contract/behavior mismatches as behavioral robot server bugs.

# Review requests

* Do we agree with these new specifications of `detail`, etc.?
* Can the new specifications of `detail`, etc. be made more specific?
* Any thoughts on the `# todo` comments for `errorType`?

# Risk assessment

Medium. This may inadvertently stop displaying information that people need, which would be frustrating.

### DIFF
--- a/api/src/opentrons/protocol_engine/errors/error_occurrence.py
+++ b/api/src/opentrons/protocol_engine/errors/error_occurrence.py
@@ -37,17 +37,55 @@ class ErrorOccurrence(BaseModel):
         )
 
     id: str = Field(..., description="Unique identifier of this error occurrence.")
-    errorType: str = Field(..., description="Specific error type that occurred.")
     createdAt: datetime = Field(..., description="When the error occurred.")
-    detail: str = Field(..., description="A human-readable message about the error.")
+
     errorCode: str = Field(
         default=ErrorCodes.GENERAL_ERROR.value.code,
-        description="An enumerated error code for the error type.",
+        description=(
+            "An enumerated error code for the error type."
+            " This is intended to be shown to the robot operator to direct them to the"
+            " correct rough area for troubleshooting."
+        ),
     )
+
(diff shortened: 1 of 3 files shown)


ecormany:
Can we also do something to make the support email address not break across the line like that? Especially now that this PR greatly reduces how much we have to cram into the modal.

SyntaxColoring:
> Can we also do something to make the support email address not break across the line like that? Especially now that this PR greatly reduces how much we have to cram into the modal.

:+1: Ticketed for the app+UI team as RAUT-704.

koji (approved):
app-side changes make sense to me.

vegano1:
sick!

sfoster1 (approved):
As long as error info is in there i'm happy. i could wish we had a way to expand something to see the info or a button to click to copy it but it's in the logs so it'll be ok.

SyntaxColoring:
> i wish we had a way to expand something to see the info or a button to click to copy it

Strongly agree.

PR #13497, opened 2023-09-08
fix(app): commandText for dropTip and adapters

jerader:
closes RAUT-698

# Overview

Fixes the copy for returning a tip to a location and adds some missing i18n keys for adapters

# Test Plan

upload a protocol where the drop tip happens in a labware location other than the trash. See in the run log that it displays the text correctly.

# Changelog

- add logic for returning tip to location other than trash
- add back i18n keys that somehow were missing??
- add test

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/assets/localization/en/protocol_command_text.json
+++ b/app/src/assets/localization/en/protocol_command_text.json
@@ -1,5 +1,5 @@
 {
-  "adapter_in_mod_in_slot": "{{adapter}} in {{module}} in {{slot}}",
+  "adapter_in_mod_in_slot": "{{adapter}} on {{module}} in {{slot}}",
   "adapter_in_slot": "{{adapter}} in {{slot}}",
   "aspirate": "Aspirating {{volume}} µL from well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
   "blowout": "Blowing out at well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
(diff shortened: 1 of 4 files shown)


app/src/assets/localization/en/protocol_command_text.json
```diff
@@ -1,4 +1,6 @@
 {
+  "adapter_in_mod_in_slot": "{{adapter}} in {{module}} in {{slot}}",
```
ecormany:
```suggestion
  "adapter_on_mod_in_slot": "{{adapter}} on {{module}} in {{slot}}",
```

app/src/assets/localization/en/protocol_command_text.json
```diff
@@ -34,6 +36,7 @@
   "pause_on": "Pause on {{robot_name}}",
   "pause": "Pause",
   "pickup_tip": "Picking up tip from {{well_name}} of {{labware}} in {{labware_location}}",
+  "return_tip": "Returning tip in {{well_name}} of {{labware}} in {{labware_location}}",
```
ecormany:
```suggestion
  "return_tip": "Returning tip to {{well_name}} of {{labware}} in {{labware_location}}",
```

ecormany (commented):
Prepositions amirite? 😅

jerader:
> Prepositions amirite? 😅

@ecormany yes, thank you!! What would we do without you 😢😆

app/src/organisms/CommandText/PipettingCommandText.tsx
```diff
@@ -74,11 +75,18 @@ export const PipettingCommandText = ({
...
     }
     case 'dropTip': {
-      return t('drop_tip', {
-        well_name: wellName,
-        labware: getLabwareName(robotSideAnalysis, labwareId),
-        labware_location: displayLocation,
-      })
+      const labwareDisplayName = getLabwareName(robotSideAnalysis, labwareId)
+
+      return labwareDisplayName === 'Fixed Trash'
```
b-cooper: It seems like the opposite check would be more stable. Instead of checking if the labware is a fixed trash, let's check if the labware is a tip rack.  Also, the display name is not a great field for this kind of comparison. Let's use the `parameters.isTiprack` to determine if we're "returning a tip".

koji (approved):
The changes look good to me!

PR #13499, opened 2023-09-08
fix(api,shared-data): Reference stable deck slot IDs instead of unstable enum names

SyntaxColoring:
# Overview

Labware definitions can adjust their gripper offsets depending on which slot they're in.

Currently, they specify this with names like `SLOT_D3`. But those names are internal and unstable. This PR switches them to be like `D3`, which is more stable, and exactly matches the keys in the deck definition.

# Test Plan

* [x] Confirm the offsets still apply.

# Review requests

None in particular.

# Risk assessment

Low.


sanni-t (approved):
Glad you caught this! LGTM!

PR #13501, opened 2023-09-08
refactor(protocol-designer): filter out 96-channel tiprack adapter fo…

jerader:
…r now

# Overview

We don't want the 96-channel tip rack adapter to be selectable right now until we support the 96-channel. So i filtered out the def in `labwareSelectionModal`

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_remove-96-channel-adapter/

Make a protocol for ot-2 or Flex. Go to the design tab, on the deck setup view, add a labware directly to the deck map. Click on the adapter dropdown option in the labware selection modal. Make sure only 1 adapter is available and that it is the 96 well aluminum block. The 96 channel adapter should NOT be available

# Changelog

- filter out adapter

# Review requests

see test plan

# Risk assessment

low


mjhuff (approved):
lgtm!

PR #13503, opened 2023-09-08
refactor(app, app-shell-odd): update configuration translation

shlokamin:
# Overview

Currently when running `make -C app dev-odd` you see the `welcome to your opentrons flex` pop up. This is because the makefile specifies the unfinished unboxing flow route to `'/dashboard'` What we want, really, is to set it to `null`, which indicates that the unboxing flow has been completed and we don't need any specific redirect. 

The problem is that the make commands are run in the OS shell environment, not in a JS environment. This means there is no way to convey core JS data types like `null`.  So, what i did for the dev odd target was pass in "0" as the unfinished unboxing flow route, and i have the selector that grabs app configs explicitly check for string "0" and convert it to `null`.

Don't love this, so if anyone has some cleaner way to go about it I'm all ears

# Review requests

Run the `dev-odd` make command and make sure the `welcome to your opentrons flex` pop up does not show up


ecormany:
The main behavior works as intended, but for some reason on this branch the window is the wrong size. It should be 1024x600 (@ 2x if on a high-resolution display) but instead it's 800x600.
Screenshot 2023-10-06 at 11 41 15 AM

This also has some weird text size knock-on effects:
Screenshot 2023-10-06 at 11 46 36 AM

koji:
@ecormany 
You see the weird text size because right now we switch the text size by Windows size if the Window's size 1024x600, the app uses ODD text definition.
Probably we will add min-height (601px) to avoid seeing this text weirdness.

koji:
I tried to `make -C dev-odd OT_APP_UI__WIDTH=1024` but the window's size wasn't changed.
However, `make -C dev OT_APP_UI__WIDTH=1200` worked as expected.
Not sure why `dev-odd` is facing this issue.

b-cooper (approved):
✔️

shlokamin:
oh yeah this is an actual issue  — i changed the library we use to parse command flags to NOT parse numbers as numbers which messing up setting the screen size. ill push up a fix

PR #13505, opened 2023-09-08
feat(app): run Disco animation after first-time-setup completion

fsinapi:
# Overview

Very simple PR to add in a "disco" animation when the first time setup flow is completed, per status bar design specs. The confluence doc stated "user completes robot setup", so I just set it up to run when the user confirms the robot's name (see vid below).

# Test Plan

Tested on Flexy McFlexFace

https://github.com/Opentrons/opentrons/assets/27798632/8dfe4fc4-3d11-48cd-ab21-591db52efcae

# Changelog

# Review requests

Mostly want to make sure this is the correct place to put this.

# Risk assessment

Pretty low

### DIFF
--- a/app/src/organisms/OnDeviceDisplay/NameRobot/ConfirmRobotName.tsx
+++ b/app/src/organisms/OnDeviceDisplay/NameRobot/ConfirmRobotName.tsx
@@ -11,13 +11,11 @@ import {
   TYPOGRAPHY,
   COLORS,
 } from '@opentrons/components'
-import { useCreateLiveCommandMutation } from '@opentrons/react-api-client'
 
 import { StyledText } from '../../../atoms/text'
 import { StepMeter } from '../../../atoms/StepMeter'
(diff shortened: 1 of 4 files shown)


shlokamin (commented):
because this component is the same rename component used in robot settings, the disco animation will also happen when you rename the robot from there. 

i think what we probably want is for the `WelcomeModal` (which is what gets rendered right after you finish set up) to trigger the disco light on mount. you can do that by using the [useEffect hook](https://react.dev/reference/react/useEffect#examples-dependencies) by passing it an empty dependency array. but also if you dont wanna deal with this someone from app ui can do it

fsinapi:
I'm pretty sure that (at least on the 7.0.0 release branch) this component is only used at the end of the setup flow - the settings page just has the one page with a textbox and then goes back to the settings page once a name is entered.

Either way, I agree that syncing this to show up with the welcome modal makes more sense, will do 👍

shlokamin (approved):
🥳

PR #13507, opened 2023-09-08
fix(app): fix module calibration while h/s is shaking

koji:
# Overview
When calibrate heater-shaker, the app will send stop shaking command.
check a module type (h/s), `currentSpeed > 0` and latch is open/close.
fix RQA-1518

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ModuleCard/__tests__/ModuleOverflowMenu.test.tsx
+++ b/app/src/organisms/ModuleCard/__tests__/ModuleOverflowMenu.test.tsx
@@ -511,4 +511,15 @@ describe('ModuleOverflowMenu', () => {
     const calibrate = getByRole('button', { name: 'Calibrate' })
     expect(calibrate).toBeDisabled()
   })
+
+  it('a mock function should be called when clicking Calibrate if pipette is ready', () => {
+    props = {
+      ...props,
+      isPipetteReady: true,
+    }
+    const { getByRole } = render(props)
+
+    getByRole('button', { name: 'Calibrate' }).click()
+    expect(props.handleCalibrateClick).toHaveBeenCalled()
+  })
 })
--- a/app/src/organisms/ModuleCard/index.tsx
+++ b/app/src/organisms/ModuleCard/index.tsx
@@ -234,9 +234,6 @@ export const ModuleCard = (props: ModuleCardProps): JSX.Element | null => {
   }
 
   const handleCalibrateClick = (): void => {
-    if (!isLatchClosed) {
-      toggleLatch()
-    }
     if (
       module.moduleType === HEATERSHAKER_MODULE_TYPE &&
       module.data.currentSpeed != null &&
@@ -256,7 +253,9 @@ export const ModuleCard = (props: ModuleCardProps): JSX.Element | null => {
         )
       })
     }
-
+    if (!isLatchClosed) {
+      toggleLatch()
+    }
     setShowCalModal(true)
   }
 
--- a/app/src/organisms/ModuleCard/index.tsx
+++ b/app/src/organisms/ModuleCard/index.tsx
@@ -63,6 +63,7 @@ import { TestShakeSlideout } from './TestShakeSlideout'
 import { ModuleWizardFlows } from '../ModuleWizardFlows'
 import { getModuleCardImage } from './utils'
 import { FirmwareUpdateFailedModal } from './FirmwareUpdateFailedModal'
+import { useLatchControls } from './hooks'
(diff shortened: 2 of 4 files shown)


app/src/organisms/ModuleCard/index.tsx
```diff
@@ -225,6 +231,22 @@ export const ModuleCard = (props: ModuleCardProps): JSX.Element | null => {
   }
 
   const handleCalibrateClick = (): void => {
+    if (module.moduleType === HEATERSHAKER_MODULE_TYPE) {
```
jerader: I would also check if the `currentSpeed` status is a number > 0, so we would only emit the command if the module is shaking. Additionally, maybe we can emit a "close latch" command as well, only if the latch is opened?

app/src/organisms/ProtocolSetupModules/index.tsx
```diff
@@ -81,8 +84,24 @@ function RenderModuleStatus({
   setShowModuleWizard,
 }: RenderModuleStatusProps): JSX.Element {
   const { i18n, t } = useTranslation('protocol_setup')
+  const { createLiveCommand } = useCreateLiveCommandMutation()
 
   const handleCalibrate = (): void => {
+    if (module.attachedModuleMatch?.moduleType === HEATERSHAKER_MODULE_TYPE) {
```
jerader: same for this!

app/src/organisms/RobotSettingsCalibration/CalibrationDetails/ModuleCalibrationOverflowMenu.tsx
```diff
@@ -49,13 +52,29 @@ export function ModuleCalibrationOverflowMenu({
...
   })
+  const { createLiveCommand } = useCreateLiveCommandMutation()
 
   const requiredAttachOrCalibratePipette =
     formattedPipetteOffsetCalibrations.length === 0 ||
     (formattedPipetteOffsetCalibrations[0].lastCalibrated == null &&
       formattedPipetteOffsetCalibrations[1].lastCalibrated == null)
 
   const handleCalibration = (): void => {
+    if (attachedModule.moduleType === HEATERSHAKER_MODULE_TYPE) {
```
jerader: and this!

app/src/organisms/ModuleCard/__tests__/ModuleOverflowMenu.test.tsx
```diff
@@ -511,4 +511,15 @@ describe('ModuleOverflowMenu', () => {
...
+
+  it('a mock function should be called when clicking Calibrate if pipette is ready', () => {
+    props = {
+      ...props,
+      isPipetteReady: true,
+    }
+    const { getByRole } = render(props)
+
+    getByRole('button', { name: 'Calibrate' }).click()
+    expect(props.handleCalibrateClick).toHaveBeenCalled()
```
jerader, line 523: maybe you can further add on to this test to see if the `useCreateLiveCommand` gets called with the correct command

jerader (commented):
looks good so far, but maybe we can add some additional logic to only omit the command when necessary.

app/src/organisms/ModuleCard/__tests__/ModuleOverflowMenu.test.tsx
koji: Testing createLiveCommand is kind of complicated so I will skip this time but created a ticket under RAUT.
I will follow it later.

app/src/organisms/ModuleCard/index.tsx
```diff
@@ -225,7 +234,30 @@ export const ModuleCard = (props: ModuleCardProps): JSX.Element | null => {
   }
 
   const handleCalibrateClick = (): void => {
-    setshowCalModal(true)
+    if (!isLatchClosed) {
```
jerader: I think we need to switch the order of these, right? If the latch is open, the module can't be shaking. So it's better to check if it is shaking then check if the latch is open, i think.

jerader: actually, i guess either way works since you won't run into a situation where you need to run both the commands in a specific order 🤔

koji: I think stop shaking and open a latch would be natural.
I will change the order.

jerader (approved):
lgtm! tested it in my dev env and saw the correct commands in network tab at the correct times

PR #12674, opened 2023-05-10
fix(app): check if mixpanel events exist before opting out #12673

shlokamin:
# Overview

This PR checks to see if a mixpanel config exists before attempting to invoke the opt out and reset methods. This is because internal testers were getting the error attached below if they opted out of analytics.

image (1)

# Review Requests

Because this issue cant be repro'd using make dev, and because we cant build mac osx versions locally because of python mismatch issues, im honestly not sure how to verify this fixes the issue other than checking the artifact CI generates to make sure the error goes away

PR #13508, opened 2023-09-11
fix(app): ensure mixpanel initialization before opting out of tracking

mjhuff:
# Overview

This PR fixes a redux store crash issue caused by opting out of analytics tracking and restarting the app. Upon restart, the redux store never properly initializes. The bug occurs because setMixpanelTracking() does not check to make sure that mixpanel.init() has been evaluated first, which is a requirement to invoke other mixpanel methods. 

This PR enables the simpler pre #12674 approach for conditionally checking user opt-in status.

Test Build: https://builds.opentrons.com/app/Opentrons-v7.0.0-alpha.8-mac-b34431-app_mixpanel_init_tracking_fix-app-build.dmg

NOTE: This issue will not appear in the HMR dev environment, because a Mixpanel ID is not passed in by default. This only occurs on built versions of the app or when the mixpanel id flag is passed to the dev version.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/f42c6cf7-e240-46be-a241-e25c076afd28

### With Fix

https://github.com/Opentrons/opentrons/assets/64858653/43084a1e-f323-438a-a5e4-f226f50ec317

# Test Plan

- Followed the steps outlined in the videos above.

# Changelog

- Added wrapper around mixpanel init. 

# Risk assessment

high

### DIFF
--- a/app/src/redux/analytics/mixpanel.ts
+++ b/app/src/redux/analytics/mixpanel.ts
@@ -20,14 +20,14 @@ const MIXPANEL_OPTS = {
   track_pageview: false,
 }
 
+const initMixpanelInstanceOnce = initializeMixpanelInstanceOnce(MIXPANEL_ID)
+
 export function initializeMixpanel(
   config: AnalyticsConfig,
   isOnDevice: boolean | null
 ): void {
   if (MIXPANEL_ID) {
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Nice, thank you!

PR #13509, opened 2023-09-11
Add usb rsa key

vegano1:
# Overview

The Flex does not use a link-local connection to the host computer so we cannot use `curl` to add a public SSH rsa key via the `/server/ssh_keys` endpoint like we do for the OT-2. Ideally, we have a mechanism that will formulate an add-key request and send it to the Flex via the usb-bridge connection, handling it as a standard HTTP request. However, we don't yet have that feature set up and need to disable root password logins for the Yocto builds. So let's add an alternative way to securely add SSH-RSA keys that still require local access to the robot in order to add an SSH key. 

This PR accomplishes this by adding a new `/server/ssh_keys/from_local` endpoint to the update-server that, scans for USB thumb drives on the Flex, finds any .pub files, validates, and then adds them to the `~/.ssh/authorized_keys` file.

# Instructions for adding your SSH key to the Flex

1. Make sure the Flex is powered on
2. Make sure you are running a version of the software that supports the `/server/ssh_keys/from_local` endpoint
3. Copy your valid SSH RSA public key to a USB thumb drive
4. Plug the USB thumb drive into the Flex
5. Using curl or equivalent send a POST request to the `/server/ssh_keys/from_local` endpoint, see below

`curl --location --request POST 'http://10.13.11.96:31950/server/ssh_keys/from_local' --header 'opentrons-version: 3'`

7. If successful, the response status will be a `201` and the message will tell you how many keys were added
`{
    "message": "Added 1 new keys",
    "key_md5": [
        "0ca9f47168c05f6675fe1806f9063084"
    ]
}`
9. If unsuccessful, the response status will be `404`
`{
    "error": "no-key",
    "message": "No valid keys found"
}`

# Test Plan

- [x] Connect a USB thumb drive to a Flex containing a valid public rsa key file, using `curl` post to the `/server/ssh_keys/from_local` endpoint, and make sure the pub key is added to the `~/.ssh/authorized_keys` file.
- [x] Connect a USB thumb drive to a Flex with invalid .pub files, using `curl` post to the `/server/ssh_keys/from_local` endpoint, no keys should be added to the `~/.ssh/authorized_keys` file.
- [x] Disable SSH password logins by adding the '-s' arg to `DROPBEAR_EXTRA_ARGS` and make sure we can't SSH without first adding a valid public key by hitting the `/server/ssh_keys/from_local` endpoint. 

# Changelog

- Added POST `/server/ssh_keys/from_local` endpoint to the update-server which searches .pub keys in thumb drives mounted to `/media`.

# Review requests

# Risk assessment
Low


sfoster1 (changes requested):
Can we bind this to a route somewhere, like `POST /server/ssh_keys/from_local` or something in the update server? That way it's easy to make a button that does it in the ui and we can still get people to `curl` it. We don't need to limit where they can `curl` from because it will only handle things from devices physically plugged in to the machine.

Also, let's refactor this so that it raises exceptions or returns a string status or something rather than printing - when it prints like this we can't really reuse it in other code.

vegano1:
> Can we bind this to a route somewhere, like `POST /server/ssh_keys/from_local` or something in the update server? That way it's easy to make a button that does it in the ui and we can still get people to `curl` it. We don't need to limit where they can `curl` from because it will only handle things from devices physically plugged in to the machine.
> 
> Also, let's refactor this so that it raises exceptions or returns a string status or something rather than printing - when it prints like this we can't really reuse it in other code.

This is the way

update-server/otupdate/common/ssh_key_management.py
```diff
@@ -238,3 +236,48 @@ async def remove(request: web.Request) -> web.Response:
...
+    """Add a public keys from usb device to the authorized_keys file.
+
+    POST /server/ssh_keys/from_local
+    -> 201 Created
+    -> 404 Not Found otherwise
+
+    """
+
+    LOG.info("Searching for public keys in /media")
+    pub_keys = (
```
sfoster1: imo we should use `os.path.walk` rather than shelling out but this will also work

sfoster1 (approved):
Would love to see this using python for the directory search rather than subprocess but looks good!

sfoster1 (approved):
Looks good to me, thank you!

PR #13512, opened 2023-09-11
fix(api): raise an error when loading a 96c pipette when another pipette is attached

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/RSS-304.
raise an error when trying to load a 96c pipette when another pipette is attached on the right.

# Test Plan

- on a Flex load the following python protocol and make sure there is an exception raised for a pipette attached:

```
requirements = {
    "robotType": "OT-3",
    "apiLevel": "2.15",
}


def run(protocol_context):
	tip_rack_50 = ctx.load_labware("opentrons_ot3_96_tiprack_50ul", 4)
	p300 = protocol_context.load_instrument('p50_single_flex', 'right', [tip_rack_50])
	p300.pick_up_tip()
	p300.drop_tip()

	# should raise an error for a pipette attached on the right
	p1000_96c = protocol_context.load_instrument('flex_96channel_1000', 'left')

```

- should work as expected:

```

requirements = {
    "robotType": "OT-3",
    "apiLevel": "2.15",
}

def run(protocol_context):
	tip_rack_50 = ctx.load_labware("opentrons_ot3_96_tiprack_50ul", 4)
	p1000_96c = protocol_context.load_instrument('flex_96channel_1000', 'right', [tip_rack_50])
	p1000_96c.pick_up_tip()
	p1000_96c.drop_tip()
```

# Changelog

- Removed all instances of `Literal[p1000_96]` and accept PipetteNameType.
- Added logic when loading a 96c pipette check the right mount and make sure nothing is attached.

# Review requests

1. Is it sufficient to check the right mount? the left mount will be checked in the following if statement if statment and its not allowed to replace. 
2. do we want raise a specific error as part of this pr? I checked in the rest of the file and the only custom error we are raising is an `APIVersionError`
3. error message raised is fine?

# Risk assessment

low. should not change existing logic. needs to be test on a robot.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -1,5 +1,4 @@
 """ProtocolEngine-based Protocol API core implementation."""
-from typing_extensions import Literal
 from typing import Dict, Optional, Type, Union, List, Tuple
 
 from opentrons.protocol_engine.commands import LoadModuleResult
@@ -408,10 +407,8 @@ class ProtocolCore(
                 load_module_result=load_module_result, model=model
             )
 
-    # TODO (tz, 11-23-22): remove Union when refactoring load_pipette for 96 channels.
-    # https://opentrons.atlassian.net/browse/RLIQ-255
     def load_instrument(
-        self, instrument_name: Union[PipetteNameType, Literal["p1000_96"]], mount: Mount
+        self, instrument_name: PipetteNameType, mount: Mount
     ) -> InstrumentCore:
         """Load an instrument into the protocol.
 
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -773,12 +773,12 @@ class ProtocolContext(CommandPublisher):
 
         tip_racks = tip_racks or []
 
-        if is_96_channel and self._instruments[Mount.RIGHT] is not None:
-            loaded_instrument = self._instruments[Mount.RIGHT]
-            assert loaded_instrument is not None
+        # TODO (tz, 9-12-23): move validation into PE
+        on_right_mount = self._instruments[Mount.RIGHT]
+        if is_96_channel and on_right_mount is not None:
             raise RuntimeError(
                 f"Instrument already present on right:"
-                f" {loaded_instrument.name}. In order to load a 96 channel pipette both mounts need to be available."
+                f" {on_right_mount.name}. In order to load a 96 channel pipette both mounts need to be available."
             )
 
         existing_instrument = self._instruments[checked_mount]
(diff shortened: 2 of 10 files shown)


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -773,6 +773,14 @@ def load_instrument(
 
         tip_racks = tip_racks or []
 
+        if is_96_channel and self._instruments[Mount.RIGHT] is not None:
+            loaded_instrument = self._instruments[Mount.RIGHT]
+            assert loaded_instrument is not None
```
TamarZanzouri: annoying that we need this assert

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -408,10 +407,8 @@ def _get_module_core(
                 load_module_result=load_module_result, model=model
             )
 
-    # TODO (tz, 11-23-22): remove Union when refactoring load_pipette for 96 channels.
-    # https://opentrons.atlassian.net/browse/RLIQ-255
```
SyntaxColoring, line 412: 🎉

api/src/opentrons/protocol_api/protocol_context.py
SyntaxColoring: Huh, yeah, that's weird.

Apparently this works:

```python
        on_right_mount = self._instruments[Mount.RIGHT]
        if is_96_channel and on_right_mount is not None:
            raise RuntimeError(
                f"Instrument already present on right:"
                f" {on_right_mount.name}. In order to load a 96 channel pipette both mounts need to be available."
            )
```

SyntaxColoring (approved):
> Is it sufficient to check the right mount? the left mount will be checked in the following if statement if statment and its not allowed to replace.

👍 

> do we want raise a specific error as part of this pr? I checked in the rest of the file and the only custom error we are raising is an APIVersionError

No, I think it's okay to raise something generic like `RuntimeError` for now.

> error message raised is fine?

👍

api/src/opentrons/protocol_api/protocol_context.py
TamarZanzouri: Yep I just didn't want to store it if it's not a 96c but it doesn't really matter. I can do this change if it's more readable

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -773,6 +773,14 @@ def load_instrument(
...
+            raise RuntimeError(
+                f"Instrument already present on right:"
+                f" {loaded_instrument.name}. In order to load a 96 channel pipette both mounts need to be available."
+            )
```
SyntaxColoring: Now that I'm looking at this again, I think ideally, this validation would be moved inside Protocol Engine, right?

That doesn't have to happen in this PR, but we should leave a `# todo` if you agree. It might not be trivial, for the same reasons as RSS-335.

SyntaxColoring (approved):
Tested on a Flex. 👍

api/src/opentrons/protocol_api/protocol_context.py
SyntaxColoring: Yeah, I would do it to avoid the `assert` if we can.

PR #13513, opened 2023-09-11
chore(api): disable legacy calibration accessors on flex

sfoster1:
We had a lot of spammy logs in the api log about missing or malformed pipette and tip length calibration.

The missing calibrations should not be logged at info level - these are polled and this is a level state, so it just ends up being spammy. The malformed calibrations though should be, and it's weird that we were getting them.

The culprit is that <7.0.0 apps poll routes to get pipette offset and tip length because that's what you do on the ot2, and the implementations of those routes hardcode the ot2 calibration models - reasonable, since we completely replaced them for the flex. So you'd only get those logs if an old app was polling the robot.

To fix that, I made the routes shortcircuit with a 403 forbidden / not allowed on hardware error. This won't affect the flex because the app doesn't check that route on flex.

### DIFF
--- a/api/src/opentrons/calibration_storage/ot2/pipette_offset.py
+++ b/api/src/opentrons/calibration_storage/ot2/pipette_offset.py
@@ -91,7 +91,7 @@ def get_pipette_offset(
         log.debug(f"Calibrations for {pipette_id} on {mount} does not exist.")
         return None
     except (json.JSONDecodeError, ValidationError):
-        log.debug(
+        log.warning(
             f"Malformed calibrations for {pipette_id} on {mount}. Please factory reset your calibrations."
         )
         # TODO: Delete the bad calibration here maybe?
--- a/api/src/opentrons/calibration_storage/ot2/pipette_offset.py
+++ b/api/src/opentrons/calibration_storage/ot2/pipette_offset.py
@@ -88,12 +88,13 @@ def get_pipette_offset(
             **io.read_cal_file(pipette_calibration_filepath)
         )
     except FileNotFoundError:
-        log.warning(f"Calibrations for {pipette_id} on {mount} does not exist.")
+        log.debug(f"Calibrations for {pipette_id} on {mount} does not exist.")
         return None
     except (json.JSONDecodeError, ValidationError):
-        log.warning(
+        log.debug(
             f"Malformed calibrations for {pipette_id} on {mount}. Please factory reset your calibrations."
         )
+        # TODO: Delete the bad calibration here maybe?
         return None
 
 
--- a/api/src/opentrons/calibration_storage/ot3/tip_length.py
+++ b/api/src/opentrons/calibration_storage/ot3/tip_length.py
@@ -37,19 +37,22 @@ def tip_lengths_for_pipette(
 ) -> typing.Dict[str, v1.TipLengthModel]:
     tip_lengths = {}
     try:
+        # While you technically could drop some data in for tip length calibration on the flex,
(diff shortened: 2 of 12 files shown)


api/src/opentrons/calibration_storage/ot2/pipette_offset.py
```diff
@@ -88,12 +88,13 @@ def get_pipette_offset(
             **io.read_cal_file(pipette_calibration_filepath)
         )
     except FileNotFoundError:
-        log.warning(f"Calibrations for {pipette_id} on {mount} does not exist.")
+        log.debug(f"Calibrations for {pipette_id} on {mount} does not exist.")
         return None
     except (json.JSONDecodeError, ValidationError):
-        log.warning(
+        log.debug(
```
SyntaxColoring: I might leave this as a warning, unless this is actually more false-positive than true-positive in practice.

api/src/opentrons/calibration_storage/ot3/tip_length.py
```diff
@@ -37,19 +37,22 @@ def tip_lengths_for_pipette(
...
+        # taking the FileNotFoundError path.
         tip_length_filepath = config.get_tip_length_cal_path() / f"{pipette_id}.json"
         all_tip_lengths_for_pipette = io.read_cal_file(tip_length_filepath)
         for tiprack, data in all_tip_lengths_for_pipette.items():
             try:
                 tip_lengths[tiprack] = v1.TipLengthModel(**data)
             except (json.JSONDecodeError, ValidationError):
-                log.warning(
+                log.debug(
                     f"Tip length calibration is malformed for {tiprack} on {pipette_id}"
```
SyntaxColoring, line 50: I might leave this as a warning, unless this is actually more false-positive than true-positive in practice.

SyntaxColoring (approved):
🥫

api/src/opentrons/calibration_storage/ot2/pipette_offset.py
sfoster1: this is, unfortunately, a frequent occurrence when a 6.3.1 app is viewing a flex because it's polling the old pipette offset route. which i might want to just disable in the server or something

api/src/opentrons/calibration_storage/ot3/tip_length.py
sfoster1: this just never should ever print ever so i can make it a warning

robot-server/robot_server/service/pipette_offset/router.py
```diff
@@ -45,7 +48,9 @@ def _format_calibration(
     response_model=pip_models.MultipleCalibrationsResponse,
 )
 async def get_all_pipette_offset_calibrations(
-    pipette_id: Optional[str] = None, mount: Optional[pip_models.MountType] = None
+    pipette_id: Optional[str] = None,
+    mount: Optional[pip_models.MountType] = None,
+    _: API = Depends(get_ot2_hardware),
```
SyntaxColoring, line 53: To make sure I understand correctly: on a Flex, even *retrieving* pipette calibrations through these endpoints was doing the wrong thing?

SyntaxColoring (approved):
Agreed with blocking off these routes if they aren't how calibration works on the Flex. Thanks!

robot-server/robot_server/service/pipette_offset/router.py
sfoster1: Yes. This code was hardcoded to select the ot2 pipette offset loaders.

PR #13514, opened 2023-09-11
fix(api): Remove obsolete "_gen3" pipette names

SyntaxColoring:
# Overview

Removes obsolete pipette names like `p50_single_gen3`. Closes RSS-284.

# Changelog

* Remove some transitional code in various places that was accepting strings like `p50_single_gen3`.
* Update a whole bunch of tests to use the new official strings, which are either like `flex_1channel_50` for the Python Protocol API, or `p50_single_flex` for HTTP and the back-end.

# Test Plan

* [x] Analyze a protocol that does `protocol.load_instrument("p50_single_gen3", mount="left")` and confirmed that it errored out.
* [x] With the robot software, ODD, and local app on this branch, play around with uploading and running protocols and make sure nothing has broken. Also attach and detach instruments.

# Review requests

I did a global find+replace, but I'm not sure whether it's appropriate to change all of these places. I'm vaguely aware that there's some distinction now between Python Protocol API load names and the names used by things deeper in the back-end, but I'm not sure exactly where that division is drawn. See my review comments.

# Risk assessment

Medium. We need to be careful that this doesn't somehow affect the HTTP API.


robot-server/simulators/test-flex.json
```diff
@@ -6,15 +6,15 @@
...
     },
     "left": {
       "model": "p50_single_3.4",
       "id": "123",
       "max_volume": 50,
-      "name": "p50_single_gen3",
+      "name": "flex_1channel_50",
       "tip_length": 0,
       "channels": 1
     }
```
SyntaxColoring: Is it correct to update these?

protocol-designer/src/components/modals/CreateFileWizard/__tests__/ModulesAndOtherTile.test.tsx
```diff
@@ -48,7 +48,7 @@ const mockWizardTileProps: Partial<WizardTileProps> = {
       robotType: FLEX_ROBOT_TYPE,
     },
     pipettesByMount: {
-      left: { pipetteName: 'p1000_single_gen3', tiprackDefURI: 'mocktip' },
+      left: { pipetteName: 'flex_1channel_1000', tiprackDefURI: 'mocktip' },
```
SyntaxColoring: Is it correct to update this?

api/tests/opentrons/hardware_control/test_instruments.py
```diff
@@ -680,12 +680,12 @@ async def test_reset_instruments(monkeypatch, sim_and_instr):
...
             "model": "p1000_single_v3.3",
             "id": "testy",
-            "name": "p1000_single_gen3",
+            "name": "flex_1channel_1000",
         },
         types.Mount.RIGHT: {
             "model": "p1000_single_v3.3",
             "id": "testy",
-            "name": "p1000_single_gen3",
+            "name": "flex_1channel_1000",
```
SyntaxColoring: Is it correct to update these?

api-client/src/pipettes/__fixtures__/index.ts
```diff
@@ -256,7 +256,7 @@ export const pipetteSettingsResponseFixture = {
 export const pipetteDataLeftFixture = {
   data: { channels: 1, min_volume: 1, max_volume: 50 },
   instrumentModel: 'p1000_single_v3.0',
-  instrumentName: 'p1000_single_gen3',
+  instrumentName: 'flex_1channel_1000',
```
SyntaxColoring: Is it correct to update these?

TamarZanzouri: why not use `p1000_single_flex`? its just a bit more readable and the outcome should be the same

SyntaxColoring: 🤷 According to https://opentrons.atlassian.net/browse/RSS-261, `p1000_single_flex` is just not the style we chose for the final Python Protocol API load names.

TamarZanzouri: I see. thank you for explaining this!

robot-server/simulators/test-flex.json
sanni-t: I thought these were all changed to '.._flex'. Let me check how this is still working.

api-client/src/pipettes/__fixtures__/index.ts
ecormany: The new load names match the product names for Flex pipettes. They don't use the "P###" naming template that OT-2 pipettes do. Users whose first Opentrons robot is a Flex 🤞in theory🤞 should never see their pipette referred to as a "P1000" or "P50" anywhere. And with three channel configurations for Flex, the names all use numerals: 1-Channel, 8-Channel, 96-Channel, not <s>Single-channel</s> or <s>Multi-channel</s>.

SyntaxColoring:
The failing "hardware-testing protocols" check seems to be a gripper thing that's unrelated to this PR. Possibly from #13506?

Laura-Danielle (commented):
If you change the tsx files, you'll need to modify `pipetteModelSpecs.json` names as well from `p1000_single_flex` to the new name.

SyntaxColoring:
> If you change the tsx files, you'll need to modify `pipetteModelSpecs.json` names as well from `p1000_single_flex` to the new name.

@Laura-Danielle Sorry, I'm confused. I understand we want to change names like `p1000_single_gen3` to `flex_1channel_1000` in the Python Protocol API. I thought names like `p1000_single_flex` were still the modern names used by the backend. Is that not true?

sanni-t (dismissed):
All the clients should use names like `p1000_single_flex`, etc.. since those are the names we are using in backend everywhere except in the user-facing API. 
Maybe these client tests don't interface with actual backend and shared data and that's why they might not be erroring out when using the `_gen3` names..? And looks like I might have missed them when I changed all `.._gen3` instances to `.._flex` in [this PR](https://github.com/Opentrons/opentrons/pull/13082/files)

As for the use in hardware control/ hardware testing, those too should change to `.._flex` but that's less critical. You can check how many tests you'll need to update by removing `gen3` from [this line](https://github.com/Opentrons/opentrons/blob/4c47d201e023266083c6ebb5fa41959e35bb7b7e/shared-data/python/opentrons_shared_data/pipette/__init__.py#L104) and running the tests. That is the line that looks up the actual pipette model from the names.

Laura-Danielle:
@SyntaxColoring the front-end never ported over to getting its info from the new shared-data format. I think someone from JS could better confirm what relies on the old config and what is comparing things directly with the backend (if that makes sense).

shared-data/python/opentrons_shared_data/pipette/dev_types.py
```diff
@@ -31,8 +31,6 @@
     "p1000_single_gen2",
     "p1000_single_flex",
     "p1000_multi_flex",
-    # TODO (tz, 11-23-22): remove when refactoring load_pipette for 96 channels.
-    # https://opentrons.atlassian.net/browse/RLIQ-255
```
SyntaxColoring, line 35: @TamarZanzouri I think it makes sense to remove this `# TODO` comment now, right?

Even if we did RLIQ-255 and the Python Protocol API loaded pipettes by channel and volume, the 96-channel wouldn't be special. If we removed the idea of "p1000_96," we'd also remove the idea of "p1000_multi_flex" and the rest.

robot-server/simulators/test-flex.json
SyntaxColoring: It turns out that since #13104, most of these fields are just getting ignored. It looks to me like only `model` and `id` are getting read now. I'll delete the rest of them.

protocol-designer/src/components/modals/CreateFileWizard/__tests__/ModulesAndOtherTile.test.tsx
SyntaxColoring: This appears to be mock data that's just ignored. Switching it to `mockPipetteName` to be more obvious.

api/tests/opentrons/hardware_control/test_instruments.py
SyntaxColoring: Looks like the same thing as https://github.com/Opentrons/opentrons/pull/13514#discussion_r1326292771. These keys are ignored now, so I'll delete them.

api-client/src/instruments/__fixtures__/index.ts
```diff
@@ -27,3 +27,25 @@ export const instrumentsResponseFixture = {
...
+export const instrumentsResponseRightPipetteFixture = {
+  data: { channels: 1, min_volume: 1, max_volume: 50 },
+  instrumentModel: 'p1000_single_v3.0',
+  instrumentName: 'p1000_single_flex',
+  instrumentType: 'pipette',
+  mount: 'right',
+  serialNumber: 'cba',
+  subsystem: 'pipette_right',
+  ok: true,
+}
```
SyntaxColoring, line 51: These are test fixtures for the `/instruments` endpoints, but they were in the directory for the `/pipettes` endpoints. I'm moving them here and correcting their `instrumentName` from `p1000_single_gen3` to `p1000_single_flex`.

api-client/src/pipettes/__fixtures__/index.ts
SyntaxColoring: To answer my original question:

Yes, `p1000_single_gen3` needed to be changed, but it needed to be changed to `p1000_single_flex` because this is an HTTP thing, as opposed to `flex_1channel_1000` which is a Python Protocol API thing.

SyntaxColoring:
> @SyntaxColoring the front-end never ported over to getting its info from the new shared-data format. I think someone from JS could better confirm what relies on the old config and what is comparing things directly with the backend (if that makes sense).

I've taken a closer look at the TypeScript files I'm modifying here. They're used for modeling HTTP responses, which return names like `p50_single_flex`. So I've changed the TypeScript files to match.

I'm not sure if anything in the front-end is still relying on the old shared-data format, but if there is, this PR doesn't touch it.

shared-data/python/opentrons_shared_data/pipette/dev_types.py
TamarZanzouri: I removed this in my 96c pr that was merged to edge yesterday. you can remove it!

robot-server/simulators/test-flex.json
```diff
@@ -4,19 +4,11 @@
   "attached_instruments": {
     "right": {
       "model": "p1000_single_3.4",
-      "id": "321",
```
TamarZanzouri, line 7: are these changes necessary for this pr?

TamarZanzouri (approved):
LGTM! Thank you for clarifying it :-)

robot-server/simulators/test-flex.json
SyntaxColoring: I think so. See https://github.com/Opentrons/opentrons/pull/13514#discussion_r1322123345.

CaseyBatten:
Ran tests on this branch utilizing each naming convention on RSS Flex in the office. Good results overall, code samples for protocols to run can be found at the following comment under the RSS-284 Issue:
https://opentrons.atlassian.net/browse/RSS-284?focusedCommentId=88046

Note that tests for the following are commented out: p50_single_flex, p1000_single_flex, p50_multi_flex, p1000_multi_flex

Upon running the Single channel test script, the following executions occur:
>**call to p50_single_gen3 denied** - _this is due to the name not existing and the try-catch entering an exception_
Load Flex 1-Channel 50 uL in Left mount succeeds, lifts a pipette, and returns it
**call to p1000_single_gen3 denied** - _this is due to the name not existing and the try-catch entering an exception_
Load Flex 1-Channel 1000 uL in Right Mount succeeds, lifts a pipette and returns it

This is the correct behavior, old naming conventions fail their attempts to assign to the Left mount. 
If the file is modified to remove the p50_single_flex and p1000_single_flex try-catch statements from their commeted out status, the following executions occur:
>**call to p50_single_gen3 denied** - _this is due to the name not existing and the try-catch entering an exception_
Load Flex 1-Channel 50uL in left mount succeeds, lifts a pipette, and returns it. This time this is happening with the "p50_single_flex" name being assigned to the hardware.
**call to flex_1channel_50 denied** - This is an unusual case, but is correct. The Left mount is now already occupied as the name "p50_single_flex" was used in the device setup and already accepted.
**call to p1000_single_gen3 denied** -_this is due to the name not existing and the try-catch entering an exception_
Load Flex 1-Channel 1000uL in Right mount succeeds, lifts a pipette, and returns it. This time this is happening with the "p1000_single_flex" name being assigned to the hardware.
**call to flex_1channel_1000 denied** - This is the same unusual case, once again correct. The Right mount is now already occupied as the name "p50_single_flex" was used in the device setup and already accepted.

When running the attached 8 channel protocol the same results are encountered at each equivalent step, including the intermediate exception cases when the p50_multi_flex and p1000_multi_flex names are re-enabled causing the standard flex naming conventions to be denied on the next step.

Overall, each step of this worked as intended on a live flex system, and the edge case naming conventions in the middle can be handled in other ways if at all.

sanni-t (approved):
🙌

PR #13518, opened 2023-09-12
fix(app, app-shell): Support protocol backwards app compatibility

mjhuff:
Closes RQA-1509

# Overview

This PR fixes an issue where users who downgrades their app from v7.0.0 to 6.3.1 or lower whitescreen on startup due to protocols supporting modules not present in previous versions. Because a v6.3.1 and 7.0.0 version of the app must both be supported until OT-2/Flex app parity, there's need for a workaround solution.

### The Immediate Problem

In order to solve this issue without deploying a hotfix for 6.3.1, protocols for version 7.0+ are stored in a new protocols folder (within the same directory as the old). A migration occurs on startup where new protocols added in the 6.3.1 app are copied over for use within the 7.0.0 app. 

Using a copy rather than a simple folder rename prevents users who have both versions of the app from seeing none of their protocols in the v6.3.1 version.

NOTE: Very large protocol files take a while to transfer and may result in a temporary whitescreen on app startup, however, in practice this should not be an issue in practice- only significantly large protocols made with the purpose of testing protocol size will cause noticeable delay. 

### The Future Problem

To prevent the need for a stop-gap solution in the future, a simple error boundary is added.

Screenshot 2023-09-12 at 8 15 28 AM

# Test Plan

- Open the app on this branch.
- Observe protocol migration by going to /Users/user_name/Library/Application Support/Opentrons/ and opening the protocols and protocols_v7.0-supported directories. Only protocols before the v6.3.1 release date are present in the former directory, and all protocols are present in the latter directory.
- Use the protocol below, which is for a module that doesn't exist (manually add it to the new protocols directory). Observe the error message upon app restart.
[34564c7z-f57a-4ca6-9e2f-725aedeb1149.zip](https://github.com/Opentrons/opentrons/files/12589099/34564c7z-f57a-4ca6-9e2f-725aedeb1149.zip)

# Changelog

- Added temporary stop-gap to migrate protocols until app parity.
- Added protocol error boundary to prevent this issue from occurring in the future.

# Risk assessment

low

### DIFF
--- a/app-shell/src/protocol-storage/file-system.ts
+++ b/app-shell/src/protocol-storage/file-system.ts
@@ -24,8 +24,13 @@ import { analyzeProtocolSource } from '../protocol-analysis'
  *  │  ├─ analysis/
  *  │  │  ├─ 1646303906.json
  */
-
-export const PROTOCOLS_DIRECTORY_NAME = 'protocols'
+// TODO(jh, 2023-09-11): remove OLD_PROTOCOLS_DIRECTORY_PATH after
+// OT-2 parity work is completed and move all protocols back to "protocols" directory.
+export const OLD_PROTOCOLS_DIRECTORY_PATH = path.join(
+  app.getPath('userData'),
+  'protocols'
+)
+export const PROTOCOLS_DIRECTORY_NAME = 'protocols_v7.0-supported'
 export const PROTOCOLS_DIRECTORY_PATH = path.join(
   app.getPath('userData'),
   PROTOCOLS_DIRECTORY_NAME
--- a/app-shell/src/protocol-storage/index.ts
+++ b/app-shell/src/protocol-storage/index.ts
@@ -42,11 +42,60 @@ export const getProtocolSrcFilePaths = (
     })
 }
 
+// TODO(jh, 2023-09-11): remove migrateProtocolsToNewDirectory after
+// OT-2 parity work is completed.
+const migrateProtocols = migrateProtocolsToNewDirectory()
+function migrateProtocolsToNewDirectory(): () => void {
+  let hasCheckedForMigration = false
+  return function (): void {
+    if (!hasCheckedForMigration) {
+      console.log(
+        `Performing one time protocol migration to ${FileSystem.PROTOCOLS_DIRECTORY_NAME}...`
+      )
+      try {
+        migrateProtocols(
+          FileSystem.OLD_PROTOCOLS_DIRECTORY_PATH,
+          FileSystem.PROTOCOLS_DIRECTORY_PATH
+        )
+        console.log('Protocol migration complete.')
+        hasCheckedForMigration = true
+      } catch (e) {
+        console.log(
+          `Error migrating protocols to ${FileSystem.PROTOCOLS_DIRECTORY_NAME}: ${e}`
+        )
+      }
+    }
+  }
+  function migrateProtocols(src: string, dest: string): void {
+    if (fse.existsSync(src)) {
+      fse.readdirSync(src).forEach(item => {
+        const srcItem = path.join(src, item)
+        const destItem = path.join(dest, item)
+        fse.copySync(srcItem, destItem, { overwrite: false })
+        deleteIfVersion7Protocol(srcItem)
+      })
+    }
+  }
+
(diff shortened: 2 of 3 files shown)


app-shell/src/protocol-storage/file-system.ts
```diff
@@ -24,8 +24,13 @@ import { analyzeProtocolSource } from '../protocol-analysis'
  *  │  ├─ analysis/
  *  │  │  ├─ 1646303906.json
  */
-
-export const PROTOCOLS_DIRECTORY_NAME = 'protocols'
+// TODO(jh, 2023-09-11): remove OLD_PROTOCOLS_DIRECTORY_PATH after
+// OT-2 parity work is completed and move all protocols back to "protocols" directory.
```
jerader, line 28: Let's make a ticket to remind us to remove this after 7.0.0 launch!

app/src/organisms/ProtocolsLanding/ProtocolCard.tsx
```diff
@@ -3,6 +3,7 @@ import { format } from 'date-fns'
 import { useTranslation } from 'react-i18next'
 import { useSelector } from 'react-redux'
 import { useHistory } from 'react-router-dom'
+import { ErrorBoundary } from 'react-error-boundary'
```
jerader, line 6: woo hoo! you added the error boundary!! My only suggestion is we make the text red so it is a bit more noticeable? You can use `COLORS.errorEnabled`

jerader (approved):
This is awesome, thank you so much for tackling this!! ⭐ 🚀 🥇  For follow up, let's make a ticket in RAUT and label it as high priority so we don't forget about it. And once this merges, we probably want to make a post in app & ui + 7.0.0 release planning slack channels to let people know that this migration will happen.

app-shell/src/protocol-storage/index.ts
```diff
@@ -42,11 +42,60 @@ export const getProtocolSrcFilePaths = (
...
+      fse.readdirSync(src).forEach(item => {
+        const srcItem = path.join(src, item)
+        const destItem = path.join(dest, item)
+        fse.copySync(srcItem, destItem, { overwrite: false })
+        deleteIfVersion7Protocol(srcItem)
+      })
+    }
+  }
+
+  // TODO(jh, 2023-09-11): Remove  deleteIfVersion7Protocol() after
```
sfoster1: IMO if this is to handle in-house only or even the guided installs we don't need it

app-shell/src/protocol-storage/index.ts
```diff
@@ -42,11 +42,60 @@ export const getProtocolSrcFilePaths = (
...
+        hasCheckedForMigration = true
+      } catch (e) {
+        console.log(
+          `Error migrating protocols to ${FileSystem.PROTOCOLS_DIRECTORY_NAME}: ${e}`
+        )
+      }
+    }
+  }
+  function migrateProtocols(src: string, dest: string): void {
+    if (fse.existsSync(src)) {
```
sfoster1: i think that since this is called in a promise chain anyway we should be using the promise versions of these functions, so we can run all the migrations concurrently through this `forEach` and gather with `Promise.all`. in-person conversation indicates that using the sync variants and blocking the node event loop is what causes the whitescreen for large protoocls so that should fix it.

sfoster1 (changes requested):
I think let's get rid of the deletion of protocols - most likely we won't have a chance to remove it before at least a beta build, and we can tell in house people to just delete their app config - and we definitely have to use the promise versions of the fs functions. Otherwise, thanks for tracking this down and finding a fix!

app-shell/src/protocol-storage/file-system.ts
mjhuff: Will do - creating a chore ticket now 👍

mjhuff:
Thanks for the feedback @jerader and @sfoster1. Changes are in - everything appears to be working.

koji:
Let me test this branch on Windows machine

koji (approved):
Tried this branch on Windows machine and it worked as expected!
🚀

sfoster1 (approved):
Thank you! Looks good!

PR #13519, opened 2023-09-12
feat(app): status bar Update animation during update installation

fsinapi:
# Overview

This should be the last missing status bar animation. Without this PR, the robot server will start running the Update animation when it starts up after an update and starts updating the firmware images on the robot. However, we want the status bar to pulse while the robot is updating its _software_, which starts well before the robot has restarted.

This PR adds a couple of simple hooks in the `RobotUpdateProgressModal` to start the Updating animation on a Flex that is being updated. If an error occurs, the app changes to the "idle" animation.

# Test Plan

* Updated an OT-2 to make sure that the `setStatusBar` command fails silently instead of interfering with operation
* Updated FlexyFace from a .zip file. The status bar pulsed as soon as the app start uploading the file, and pulsed during the restart.
* Updated FlexyFace and then killed the Update Server during the .zip file upload. The app turned off the pulsing animation.

# Changelog

* When the RobotUpdateProgressModal is first rendered, send a command to the robot to start the Updating animation (pulsing white) on the status bar
* When there is an error during an update, send a command to the robot to display the Idle animation (solid white) on the status bar

# Review requests

* I ended up taking basically the same approach here as with the "disco" animation during first-time setup, mostly because I don't know enough typescript to know if there's a better way to do this. If this is a silly method and/or location to add these hooks then please point me in the right direction :) 

# Risk assessment

Pretty low _as long as_ my understanding of javascript `catch` is correct - when updating a robot with software old enough to not recognize the animation commands, they should just be ignored. 

Otherwise, the biggest risk is some possibility of weird animation states on the robot if you start an update and then close your app in the middle of it. I think the only alternative here would be to manage this through the Update Server instead of the app. I'm a little wary about integrating this kind of functionality in the Update Server unless absolutely necessary, but I could be convinced.

### DIFF
--- a/app/src/organisms/Devices/RobotSettings/UpdateBuildroot/RobotUpdateProgressModal.tsx
+++ b/app/src/organisms/Devices/RobotSettings/UpdateBuildroot/RobotUpdateProgressModal.tsx
@@ -15,6 +15,7 @@ import {
   SPACING,
   BORDERS,
 } from '@opentrons/components'
+import { useCreateLiveCommandMutation } from '@opentrons/react-api-client'
 
 import { StyledText } from '../../../../atoms/text'
 import { LegacyModal } from '../../../../molecules/LegacyModal'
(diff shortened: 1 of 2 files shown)


jerader (approved):
makes sense to me! thanks for adding test coverage. You can potentially split the `startIdleAnimationIfFailed` and `startUpdatingAnimation` into their own functions/files but i don't think its necessary since it seems like these commands will only get called here.

PR #13520, opened 2023-09-12
fix(shared-data): PVT pipette functions, not including 200ul

andySigler:
# Overview

Accuracy function for 1ch and 8ch pipettes, 50ul and 1000ul tips (no 200ul functions yet).

Note that the `lowVolume` functions are copy/pasted from `default`. For these first PVT pipettes, we'll test to see if the default function is good enough, and will generate a new function for `lowVolume` only if needed.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
lgtm!

PR #13521, opened 2023-09-12
chore(push): add ssh_key option when using push-ot3

vegano1:
# Overview

We are enabling SSH RSA keys on the Flex, which means you will no longer be able to SSH to a Flex release build without a valid SSH private key added using the new `/server/ssh_keys/from_local` endpoint recently added [here](https://github.com/Opentrons/opentrons/pull/13509). Because of this, we want to make sure developers are not affected and are ready once the robots require SSH keys. In order to support this we need to change our `push-ot3` Makefile script so it now takes in an identity file `ssh_key` as the private key.

# Instructions for adding your SSH key to the Flex

1. Make sure the Flex is powered on
2. Make sure you are running a version of the software that supports the `/server/ssh_keys/from_local` endpoint
3. Copy your valid SSH RSA public key to a USB thumb drive
4. Plug the USB thumb drive into the Flex
5. Using curl or equivalent send a POST request to the `/server/ssh_keys/from_local` endpoint, see below

`curl --location --request POST 'http://10.13.11.96:31950/server/ssh_keys/from_local' --header 'opentrons-version: 3'`

7. If successful, the response status will be a `201` and the message will tell you how many keys were added
`{
    "message": "Added 1 new keys",
    "key_md5": [
        "0ca9f47168c05f6675fe1806f9063084"
    ]
}`
9. If unsuccessful, the response status will be `404`
`{
    "error": "no-key",
    "message": "No valid keys found"
}`

# Test Plan

- [x] Use push-ot3 without `ssh_key` on a robot with ssh keys enabled and make sure you get permission denied
- [x] Use push-ot3 `ssh_key` on a robot with SSH keys enabled and make sure you get permission denied if the private key is invalid
- [x] Use push-ot3 with `ssh_key` on a robot with SSH keys enabled and make sure you can SSH if you have a valid private key.

# Changelog

- Added `ssh_key` option to make-push so It takes in an identity file in order to perform SSH/SCP operations
- Remove the default `default_ssh_key` from `scripts/push.mk` so we can still ssh into a Flex without providing an identity file

# Review requests

- It should be straightforward but would like some thoughts on removing the default `default_ssh_key`, Is anything wrong with this?

# Risk assessment

Medium, this will affect developers directly or anyone needing to use make push-ot3 to SSH/SCP to the robot and if the robot version and local monorepo branch are not in sync it will cause problems.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -380,10 +380,10 @@ ssh-keygen # note the path you save the key to
 make -C robot-server install-key br_ssh_pubkey=/path/to/pubkey host=${some_other_ip_address}
 ```
 
-and subsequently, when you do `make term`, add the `br_ssh_key=/path/to/key` option:
+and subsequently, when you do `make term`, add the `ssh_key=/path/to/key` option:
 
 ```shell
-make term br_ssh_key=/path/to/privkey
+make term ssh_key=/path/to/privkey
 ```
 
 If you create the key as `~/.ssh/robot_key` and `~/.ssh/robot_key.pub` then `make term` and `make install-key` will work without arguments.
--- a/api/Makefile
+++ b/api/Makefile
@@ -49,10 +49,12 @@ test_opts ?=
 pypi_username ?=
 pypi_password ?=
 
-# Host key location for buildroot robot
-br_ssh_key ?= $(default_ssh_key)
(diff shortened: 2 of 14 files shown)


sfoster1 (approved):
Looks good to me!

PR #13522, opened 2023-09-12
fix(api): failed ot3api gripper capacitive sweep test

ahiuchingau:
# Overview
I forgot to update the gripper jaw state in the ot3simulator after performing a jaw action. This should fix those broken tests.


sfoster1 (approved):
lgtm!

PR #13524, opened 2023-09-12
feat(app, shared-data, protocol-designer, step-generation): frontend low volume support

jerader:
closes RAUT-714

# Overview

This adds low volume front-end support

# Test Plan

in PD:

- create a Flex protocol with a p50 pipette (single or 8-channel). 
- Add a transfer step and make the volume > 1uL, see that in the dispense advance settings, the `push out` option is disabled with a tooltip explaining why.
- go back and change the volume = 1uL, see that the "push out" setting is available and clickable. Click it.
- save and download the protocol and open up the json file. see that `configure_for_volume` command occurs before the `aspirate` with the correct commands. See that the `dispense` command has the `pushOut` param with the correct volume (volume should be 1, which matches the aspirate volume).
- go back and create a new PD protocol for the ot-2, see in the transfer form that the `push out` option is not available.

in the App:

This is not testable yet since there are no backend changes in `chore_release-7.0.0` yet. But just review the `commandText` for the `configureForVolume` and `dispense` command.

# Changelog

in shared-data:
- add the types for `ConfigureForVolumeCreateCommand` and ConfigureForVolumeRunTimeCommand`, add the optional `pushOut` param to `dispense` command

in app:
- add logic in `commandText` and add test cases

in PD:
- add the `dispense_pushOut` value to `transfer` form
- add logic for when the advanced setting in transfer are disabled, adding tooltip as well
- add to the migration file to migrate all `transfer` stepTypes to include the `pushOut` boolean value
- fix cypress e2e tests to include the `pushOut` value 

in Step-generation:
- create new `configureForVolume.ts` atomic command, add test for it
- add logic for emitting the command in `transfer.ts` before an `aspirate` happens and volume is 1uL and with the correct pipettes.

# Review requests

see test plan

# Risk assessment

low


app/src/assets/localization/en/protocol_command_text.json
```diff
@@ -12,6 +13,7 @@
   "degrees_c": "{{temp}}°C",
   "disengaging_magnetic_module": "Disengaging Magnetic Module",
   "dispense": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
+  "dispense_push_out": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec with push out volume {{push_out_volume}} µL",
```
jerader, line 16: @ecormany and this is the `commandText` for dispensing with the `pushOut` param

jerader:
made an oopsie when i change the targeted branch and accidentally messed up the commit history. Opening a new pr targeting the correct branch!

PR #13526, opened 2023-09-12
feat(app, shared-data, step-generation): frontend low volume support

jerader:
closes RAUT-714

# Overview

This adds low volume front-end support

# Test Plan

in PD/Step-generation:

sandbox: https://sandbox.designer.opentrons.com/app_pd-low-volume-support/

- create a Flex protocol with a p50 pipette (single or 8-channel) and a p1000 flex pipette (single or 8-channel). 
- Add 2x transfer steps (1 for each pipette)
- save and download the protocol and open up the json file. see that `configureForVolume` command occurs before the `aspirate` with the p50 pipette but not before the `aspirate` with the p1000 pipette.
- create a new Flex protocol with a p50 pipette (single or 8-channel). 
- this time add a mix step with mixing 1uL
- download the protocol, in the json file you should see only 1 `configureForVolume` command happening before the mix's `aspirate` and `dispense` commands

in the App:

- upload the protocol to the app and should pass protocol analysis. Check the run log command text for the `configureForVolume` commands.

# Changelog

in shared-data:
- add the types for `ConfigureForVolumeCreateCommand` and `ConfigureForVolumeRunTimeCommand`, add the optional `pushOut` param to `dispense` command

in app:
- add logic in `commandText` and add test cases

in PD:
- there are no visual changes to PD!

in Step-generation:
- create new `configureForVolume.ts` atomic command, add test for it
- add logic for emitting the command in `transfer.ts` before an `aspirate` happens and volume is 1uL and with the correct pipettes.
- add logic for emitting the command in `mix.ts` to happen before the set of mix `aspirate` and `dispense` commands.
- add a comment explaining that pushOut won't be specified in `dispense.ts`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/assets/localization/en/protocol_command_text.json
+++ b/app/src/assets/localization/en/protocol_command_text.json
@@ -1,7 +1,7 @@
 {
   "aspirate": "Aspirating {{volume}} µL from well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
   "blowout": "Blowing out at well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
-  "configure_for_volume": "Configure {{volume}} µL for aspirate",
+  "configure_for_volume": "Configure {{volume}} µL for aspirating with {{pipette}}",
   "closing_tc_lid": "Closing Thermocycler lid",
   "comment": "Comment",
   "confirm_and_resume": "Confirm and resume",
@@ -13,6 +13,7 @@
   "degrees_c": "{{temp}}°C",
   "disengaging_magnetic_module": "Disengaging Magnetic Module",
   "dispense": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
+  "dispense_push_out": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec with push out volume {{push_out_volume}} µL",
   "drop_tip": "Dropping tip in {{well_name}} of {{labware}}",
   "engaging_magnetic_module": "Engaging Magnetic Module",
   "fixed_trash": "Fixed Trash",
--- a/app/src/assets/localization/en/protocol_command_text.json
+++ b/app/src/assets/localization/en/protocol_command_text.json
@@ -1,6 +1,7 @@
 {
   "aspirate": "Aspirating {{volume}} µL from well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
   "blowout": "Blowing out at well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
+  "configure_for_volume": "Configure {{volume}} µL for aspirate",
   "closing_tc_lid": "Closing Thermocycler lid",
   "comment": "Comment",
   "confirm_and_resume": "Confirm and resume",
--- a/app/src/assets/localization/en/protocol_command_text.json
+++ b/app/src/assets/localization/en/protocol_command_text.json
@@ -1,7 +1,7 @@
 {
   "aspirate": "Aspirating {{volume}} µL from well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
   "blowout": "Blowing out at well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
-  "configure_for_volume": "Configure {{volume}} µL for aspirating with {{pipette}}",
+  "configure_for_volume": "Configure {{pipette}} to aspirate {{volume}} µL",
   "closing_tc_lid": "Closing Thermocycler lid",
   "comment": "Comment",
   "confirm_and_resume": "Confirm and resume",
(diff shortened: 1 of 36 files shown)


app/src/assets/localization/en/protocol_command_text.json
```diff
@@ -1,6 +1,7 @@
 {
   "aspirate": "Aspirating {{volume}} µL from well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
   "blowout": "Blowing out at well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
+  "configure_for_volume": "Configure {{volume}} µL for aspirating with {{pipette}}",
```
jerader: @ecormany , this is the commandText in the run log for configure_for_volume, does it make sense to you?

app/src/assets/localization/en/protocol_command_text.json
```diff
@@ -12,6 +13,7 @@
   "degrees_c": "{{temp}}°C",
   "disengaging_magnetic_module": "Disengaging Magnetic Module",
   "dispense": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec",
+  "dispense_push_out": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec with push out volume {{push_out_volume}} µL",
```
jerader: @ecormany and this is the commandText for dispensing with the pushOut param

step-generation/src/commandCreators/compound/transfer.ts
```diff
@@ -163,6 +165,20 @@ export const transfer: CommandCreator<TransferArgs> = (
             changeTipNow = isInitialSubtransfer || destWell !== prevDestWell
           }
 
+          const configureForVolumeCommand: CurriedCommandCreator[] =
```
jerader, line 167: note: this command emits every time for a p50 flex pipette in the beginning of a transfer. So if you were to aspirate/dispense multiple times in the same transfer (at the same volume, since you're not allowed to change volumes in the middle of a transfer in PD), this command will only happen once at the beginning. But if you were to add a new transfer right afterwards with the same volume, the command would emit again. That is where i'm not too sure how to refactor to not emit it again 🤔 , would need to do so in a follow up PR

sfoster1:
App command render looks good!

app/src/assets/localization/en/protocol_command_text.json
ecormany: To be really useful, I would like the run log text to say what _mode_ the pipette is placed in. I chatted some with @andySigler about this yesterday, although I don't think we came to a final decision re: run log.

If, at least at this point, the run log is just going to report the arguments passed to `configureForVolume`, then I'd suggest `Configure {{pipette}} to aspirate {{volume}} µL`

app/src/assets/localization/en/protocol_command_text.json
ecormany: oof this is long but i dunno that there's much we can do…
```suggestion
  "dispense_push_out": "Dispensing {{volume}} µL into well {{well_name}} of {{labware}} in {{labware_location}} at {{flow_rate}} µL/sec and pushing out {{push_out_volume}} µL",
```
do we need to state the push-out flow rate or is it constant?

mjhuff (approved):
It's hard for me to tell if conceptually these changes are correct (since I'm still learning PD), but I tested per your instructions and everything works. The JS lgtm as well! 🚀

app/src/assets/localization/en/protocol_command_text.json
jerader: Oh wow, that order of words makes way more sense than what I had 😅

sfoster1 (changes requested):
Stuff from initial testing
- ✅ app rendering of configure looks good 
   - ❓  though it says "Configuring 1 ul for aspiration with <blah blah pipette>" and maybe it should say "Configuring <blah blah> pipette to aspirate 1ul"
- ❌ PD UI for push out needs to change; we should either not
   - not have push out UI at all (my pref) because specifying it is complex (see below) and not emit the arg in step-generation so it uses default settings or
   - have push out be a check box _that defaults to checked_, such that when it's checked step-generation doesn't emit the push out arg and when it's not checked step-generation emits it with a 0 value
   - have push out be the same "default or set value" ui field that we have elsewhere where defaulted is null/not present in s-g, and a real value is set to that value. note that here it can only be shown on full dispenses and the limits for the value have to be computed from the liquid settings as (bottom-blow_out)*plunger_surface_area
- [ ] haven't tested the actual generated protocol yet though

sfoster1 (changes requested):
`configureForVolume` needs to be emitted more frequently. right now it's emitted only before 1ul aspirates, which means the pipette could never switch out of the low volume mode. we need to, on flex 50ul pipettes, emit before each aspirate that starts with the plunger empty, with the total volume we'll aspirate until the next time the plunger is empty

sfoster1 (approved):
- [x] made a protocol in PD branch with 50ul 1 and 8 channels
   - [x] with a variety of transfer volumes, always saw configure for volume
   - [x] protocol analyzed successfully in app branch
   - [x] ODD and app runlog displayed configures successfully
   - [x] configures executed properly traced through to api logs
- [x] made protocol with mix in PD branch
   - [x] saw configure before mix multistep
- [x] made protocol with 1000ul pipettes and saw no configure for volumes

Looks good to me in my dev testing!

PR #13528, opened 2023-09-12
fix(hardware): prioritize raising estop errors from failed movements

fsinapi:
# Overview

Closes RQA-1528

When the estop is pressed during a movement, it can cause some of the axes to raise errors _other_ than estop failures - most notably, `collision_detected` can be reported by some nodes. The presence of these other errors can obfuscate the actual cause of the error (which is that the estop was pressed), as the app may show an error other than `estop_detected`.

This PR changes the MoveGroupRunner to prioritize Estop-related errors. Pressing the estop during a movement really represents a single logical "error event", but each node independently reports it. Therefore, we will just return a single one of those error messages so that the logged error is not the generic message "Motion failed with multiple errors" but rather a specific Estop-related error message.

# Test Plan

Ran a Flex protocol a few times and pressed the Estop during a movement each time. After each failure, looking at the error details for the run would show an Estop error message.

# Changelog

- MoveGroupRunner will return a single estop error if one exists, instead of returning the entire list of errors.
- Added test that the estop errors are filtered properly

# Review requests

# Risk assessment

### DIFF
--- a/hardware/opentrons_hardware/hardware_control/move_group_runner.py
+++ b/hardware/opentrons_hardware/hardware_control/move_group_runner.py
@@ -10,6 +10,7 @@ from opentrons_shared_data.errors.exceptions import (
     GeneralError,
     MoveConditionNotMetError,
     EnumeratedError,
+    EStopActivatedError,
     MotionFailedError,
     PythonException,
 )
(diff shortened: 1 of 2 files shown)


vegano1 (approved):
This makes sense to me

sfoster1 (approved):
okay. i kind of wish we were fixing this by changing what we render but it'll be all right.

PR #13529, opened 2023-09-12
chore(shared-data): liquid class bound updates

sfoster1:
Change the bounds for the liquid classes to match what hardware wants and update the selection logic to remove hysteresis and just change based on the low volume of default.


Laura-Danielle (approved):
nice. LGTM

PR #13531, opened 2023-09-12
refactor(api): add idle gripper method to ot3api

ahiuchingau:
# Overview
This PR does:
1. Add gripper default idle force to shared-data gripper definitions & update schema
2. Add `idle_gripper()` to ot3api which uses the default idle force to close the gripper jaw
3. Make protocol engine call `idle_gripper()` instead


sfoster1 (approved):
lgtm if it all passes!

PR #13533, opened 2023-09-12
fix(app): fix TC allows module calibration with lid closed

koji:
# Overview
fix TC allows module calibration with lid closed

close RQA-1519

ToDo update test cases later

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ModuleCard/index.tsx
+++ b/app/src/organisms/ModuleCard/index.tsx
@@ -66,7 +66,10 @@ import { FirmwareUpdateFailedModal } from './FirmwareUpdateFailedModal'
 import { useLatchControls } from './hooks'
 import { ErrorInfo } from './ErrorInfo'
 
-import type { HeaterShakerDeactivateShakerCreateCommand } from '@opentrons/shared-data/protocol/types/schemaV7/command/module'
+import type {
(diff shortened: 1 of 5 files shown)


app/src/organisms/RobotSettingsCalibration/index.tsx
```diff
@@ -158,9 +158,6 @@ export function RobotSettingsCalibration({
       : null
   )
 
-  console.log('pipetteOffsetCalibrations', pipetteOffsetCalibrations)
```
jerader, line 161: 🔪

app/src/organisms/ModuleCard/index.tsx
```diff
@@ -256,6 +259,24 @@ export const ModuleCard = (props: ModuleCardProps): JSX.Element | null => {
     if (!isLatchClosed) {
```
jerader: I think we need additional logic in here to check if the module is a heater-shaker type, otherwise, the command is emitted even if you are launching the module cal flow for the thermocycler and the h-s attached to the robot has a closed latch.

jerader (approved):
lgtm, was able to see the TC open lid command called in the network tab. Just need the additional logic for when `toggleLatch()` is called

PR #13504, opened 2023-09-08
fix(engine): allow dropping tips in any labware

sanni-t:
Closes RSS-277

# Overview

Fixes a bug introduced in 6.3.0 that unintentionally raises an error when trying to drop tips into labware that is neither a fixed trash nor a tiprack.

Our API documentation clearly states that you can specify an arbitrary tip drop location in `instrument_context.drop_tip()`. Furthermore, the API allows assigning any labware as the designated trash container for a particular pipette, which is treated as the default tip drop location. 
For executing both these cases, the engine needs to allow dropping tips into any labware; not just tiprack or fixed trash.

This is also necessary for the upcoming 'post-run-tip-drop' flow that the app is planning on adding.

# Test Plan

This protocol should drop tips into the fixed trash, the tip rack **and a reservoir** as expected
(Can also be tested on the OT-2 by updating the protocol for OT-2)

```
requirements = {
	"robotType": "Flex",
	"apiLevel": "2.15"
}

def run(protocol_context):
	# For Flex:
	tiprack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "D2")
	pipette = protocol_context.load_instrument("flex_1channel_50", 'left', tip_racks=[tiprack1])
	
	pipette.pick_up_tip()
	pipette.drop_tip()		# Should drop tip at TOP of fixed trash in A3 (at an alternating location)

        pipette.pick_up_tip()
	pipette.drop_tip()		# Should drop tip at TOP of fixed trash in A3 (at an alternating location)

	pipette.drop_tip(pipette.trash_container['A1'].center())	# Should always drop tip in CENTER of Fixed trash

	trash_labware = protocol_context.load_labware("nest_1_reservoir_195ml", "D1")
	pipette.trash_container = trash_labware

	pipette.pick_up_tip()
	pipette.drop_tip()		# Should drop tip at CENTER TOP of Nest reservoir (NO alternating locations)

	pipette.pick_up_tip()
	pipette.drop_tip(pipette.trash_container['A1'].center())	# Should always drop tip in CENTER of Nest reservoir

	pipette.pick_up_tip()
	pipette.return_tip()	        # Should return tip to tiprack
```

# Changelog

- changed `get_checked_tip_drop_location()` to check if labware is tiprack, else drop tip to TOP of specified labware.
- updated tests

# Review requests

- Make sure I am not messing up any side effects of tip dropping by allowing any arbitrary locations

# Risk assessment

Low. Bug fix.

PR #13534, opened 2023-09-12
fix(api): disallow moving a fixed-trash labware

TamarZanzouri:
# Overview

follow up work for https://github.com/Opentrons/opentrons/pull/13504.
disallow moving a labware if it has a fixed trash quirk. 

# Test Plan

try moving a fixed trash labware. make sure it raises.
try moving a non fixed-trash labware - make sure it works as expected. 

# Changelog

- validate that the movable labware is not a fixed trash. if it is, raise. 

# Risk assessment

low. should only affect fixed-trash labware in move-labware.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/move_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/move_labware.py
@@ -92,6 +92,13 @@ class MoveLabwareImplementation(
         )
         definition_uri = current_labware.definitionUri
 
+        if labware_validation.validate_if_labware_is_fixed_trash(
+            current_labware_definition
+        ):
+            raise LabwareMovementNotAllowedError(
+                f"Cannot move fixed trash labware '{current_labware_definition.parameters.loadName}'."
+            )
+
         available_new_location = self._state_view.geometry.ensure_location_not_occupied(
             location=params.newLocation
         )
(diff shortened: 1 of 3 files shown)


SyntaxColoring (commented):
Code looks great to me. Will test today.

SyntaxColoring (approved):
Tested, works as described. 🚮

api/src/opentrons/protocol_engine/commands/move_labware.py
```diff
@@ -92,6 +92,13 @@ async def execute(self, params: MoveLabwareParams) -> MoveLabwareResult:
         )
         definition_uri = current_labware.definitionUri
 
+        if labware_validation.validate_if_labware_is_fixed_trash(
+            current_labware_definition
+        ):
```
sanni-t: I think we should just use the `labware_view.is_fized_trash()` check to test if this is the fixed trash. We really only care about the 'fixed' trash in slot 12/ A3 and want to keep that fixed/ unmovable.

TamarZanzouri: done!

sanni-t (approved):
🙌 Thanks for the quick fix!!

PR #13536, opened 2023-09-12
refactor(hardware): SensorIdField fix

caila-marashaj:
## Overview
The `build` method that was in the `PushTipPresenceNotificationPayload` causes 16 0's to get added onto the `sensor_id` field of the payload, resulting in the secondary sensor reporting its id as `65536`. Removing this method and handling the `sensor_id` the same way the `SensorPayload` does seems to fix this issue.


sfoster1 (approved):
Looks good!

PR #13537, opened 2023-09-12
fix(api): Report Python protocol line numbers for errors from Protocol Engine

SyntaxColoring:
# Overview

Fixes RSS-317.

# Test Plan

* [x] E-stop a physical run on a Flex and make sure the desktop app and ODD respond correctly.
* [x] Import the protocol from RSS-317 and make sure it shows a line number now.

# Changelog

* **Bug fix:** When an exception bubbles out of the Python protocol `exec()`, do *not* handle it specially just because that exception passed through Protocol Engine. Handle it like any other exception. This lets us attach the protocol line number to it.

  We were handling it specially before because, I guess, we wanted it to collude with [this `isinstance` check up in `ProtocolEngine.finish()`](https://github.com/Opentrons/opentrons/blob/3ffe41c25ae3deb822942019543480aba206ac02/api/src/opentrons/protocol_engine/protocol_engine.py#L337). Apparently this was unnecessary, because the `ExceptionInProtocolError` that we would wrap it with passes the `isinstance` check just the same.

* Add some analysis integration tests that make sure the line number shows up.

* Resolve some type-checking exclusions in `opentrons.protocols.execute`.

# Review requests

* Do the comments make sense?
* Do we have a reasonable way of inducing E-stop errors in CI tests?

# Risk assessment

Probably medium-high. The special E-stop handling is hard to reason about and unit-test because it works by several far-away places colluding with each other.

### DIFF
--- a/api/src/opentrons/protocol_engine/protocol_engine.py
+++ b/api/src/opentrons/protocol_engine/protocol_engine.py
@@ -329,11 +329,26 @@ class ProtocolEngine:
                 after the run is over.
         """
         if self._state_store.commands.state.stopped_by_estop:
+            # This handles the case where the E-stop was pressed while we were *not* in the middle
+            # of some hardware interaction that would raise it as an exception. For example, imagine
+            # we were paused between two commands, or imagine we were executing a very long run of
+            # comment commands.
             drop_tips_after_run = False
             post_run_hardware_state = PostRunHardwareState.DISENGAGE_IN_PLACE
             if error is None:
                 error = EStopActivatedError(message="Estop was activated during a run")
+
         if error:
+            # If the run had an error, check if that error indicates an E-stop.
+            # This handles the case where the run was in the middle of some hardware control
+            # method and the hardware controller raised an E-stop error from it.
+            #
+            # To do this, we need to scan all the way through the error tree.
+            # By the time E-stop error has gotten to us, it may have been wrapped in other errors,
+            # so we need to unwrap them to uncover the E-stop error's inner beauty.
+            #
+            # TODO(mm, 2023-09-12): Do we need to scan the exception tree like this? Instead, can we
+            # directly inspect the E-stop state through self._hardware_api.get_estop_state()?
             if isinstance(error, EnumeratedError) and self._code_in_error_tree(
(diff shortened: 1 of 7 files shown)


api/src/opentrons/protocols/execution/errors.py
SyntaxColoring, line 1: No substantive changes in here, just adding type annotations and rearranging things for readability.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -329,13 +329,28 @@ async def finish(
...
+            # If the run had an error, check if that error indicates an E-stop.
+            # This handles the case where the run was in the middle of some hardware control
+            # method and the hardware controller raised an E-stop error from it.
+            #
+            # To do this, we need to scan all the way through the error tree.
+            # By the time E-stop error has gotten to us, it may have been wrapped in other errors,
+            # so we need to unwrap them to uncover the E-stop error's inner beauty.
+            #
+            # TODO(mm, 2023-09-12): Do we need to scan the exception tree like this? Instead, can we
+            # directly inspect the E-stop state through self._hardware_api.get_estop_state()?
```
SyntaxColoring: @TamarZanzouri @sfoster1 Thoughts on this `# TODO`?

sfoster1: We need to scan the exception tree because someone in theory could have released the estop really fast after pressing it and the physical estop state would not be the same as the one that caused the error.

sfoster1 (approved):
Looks good to me! It makes sense and I agree with it. However, I think the todo is wrong - see the inline comment.

api/src/opentrons/protocol_engine/protocol_engine.py
SyntaxColoring: Mm, yeah, okay.

Would that be likely in practice? I think someone would have to physically release the E-stop button *and* `PUT /robot/control/acknowledgeEstopDisengage` before the exception finishes its journey from the hardware controller to `ProtocolEngine.finish()`. (I don't think that means it doesn't matter—just developing a mental model.)

TamarZanzouri (approved):
LGTM! Thank you!

PR #13538, opened 2023-09-12
fix(api): OT3 use retract target in pick up tip function

ahiuchingau:
# Overview
We haven't been retracting the pipette during `pick_up_tip()` in ot3api, which makes it hard for user to verify whether if tip is picked up or not during LPC. This fixes that. 

Now the behavior of pick up tip should match that of the OT2. One problem is that now we're going to retract _every time_ we pick up a tip, even outside of LPC. We really only need to do this when someone needs to physically verify if a tip is attached. I think eventually, we should move the retract call out of pick up tip and have the client move the pipette up only during LPC.


sfoster1 (approved):
Looks good to me!

PR #13544, opened 2023-09-13
chore(release): notes for 7.0.0 stable

ecormany:
# Overview

Any modifications to the release notes for the 7.0.0 stable release should be added to this PR, which ~~will be merged shortly before we have a release candidate~~ let's merge it now, so it's definitely in for the tag that gets promoted to beta and stable.

Addresses RTC-307.

# Test Plan

- Read the prose in the rich markdown preview here on GitHub.
- Preview the notes in the Opentrons App update flow, if possible. This depends on the exact timing of when this PR is merged / when the RC is tagged.

# Changelog

- @sanni-t added a bug fix for tip dropping.
- Removed nested list formatting from this and previous versions, to play nicer with our current CSS.
- Noted tip tracking fixed by [#13156](https://github.com/Opentrons/opentrons/pull/13156)
- Still a known issue that there's no `opentrons_simulate`

# Review requests

Are there any _complete or required for 7.0.0_ features not adequately described in the notes?

We can always patch later, but ideally this covers it — at the high level our release notes usually do.

# Risk assessment

For code, effectively nil. For user impact, somewhat higher — everyone sees the notes when updating.

### DIFF
--- a/api/release-notes.md
+++ b/api/release-notes.md
@@ -41,10 +41,14 @@ Python API features
 ### Bug Fixes
 
 - The API no longer raises an error when dropping tips into labware other than the fixed trash.
+- All API versions now properly track tips, including starting at a well other than A1.
 
 ### Known Issues
 
-- Tip tracking starting at a well other than A1 will not pick up tips from the intended locations.
+Some protocols can't be simulated with the `opentrons_simulate` command-line tool:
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Excellent!

PR #13550, opened 2023-09-13
fix(app, component, shared-data): fix select location on deck for tc

koji:
# Overview
fix select location on deck for TC
Added a new slot name for `A1+B1` (TC) and modify the component's text part for TC.
I think this isn't really good solution but the select function is for a single slot and that function makes things complicated when compare to other deck view usages.

close RQA-1494

h/s
Screenshot 2023-09-13 at 5 22 36 PM

tc
Screenshot 2023-09-13 at 5 25 33 PM

# Test Plan
- module calibration for TC
- module calibration for any modules except TC (not break anything)

# Changelog
- add Z1 to shared-data and components

# Review requests
The way to add a new item, `Z1` to  `shared-data/deck/definitions/3/ot3_standard.json`  is correct?
If you have more clean ways to render text in selected rectangle, please leave comments.

# Risk assessment


shared-data/deck/definitions/3/ot3_standard.json
```diff
@@ -188,6 +188,18 @@
           "heaterShakerModuleType",
           "magneticBlockType"
         ]
+      },
+      {
+        "id": "Z1",
```
jerader, line 193: hmm, i feel a little bit sus about this since we are changing the deck definition. Is there a way to make this change downstream and contained in the app? Maybe we can discuss a solution in standup

koji: The component compares the information from this json file.
https://github.com/Opentrons/opentrons/pull/13550/files#diff-4be3572eda2bf35be1fe63ebbb3126bf7976178a7f38bdb833b46e1046c8c7a1R60

PR #13555, opened 2023-09-14
fix(api, app): handle gantry backdrive during instrument attach

fsinapi:
# Overview

Closes RQA-1578

When attaching an instrument (gripper or pipette), it's possible to push a little too hard and backdrive the gantry axes a little bit. If this happens, the motor controllers detect that their position is invalid and the subsequent command to start calibrating the instrument will fail as a result.

We could fix this by unconditionally homing all of the axes before starting calibration (we used to do this on at least some of the workflows), but that is Really Annoying and most of the time it is just a waste of the user's time.

In order to handle situations where the gantry position becomes invalid, we add a new _optional_ parameter to the Home command, named `skipIfMountPositionOk`. If a mount is provided to this parameter, the protocol engine will check if that mount has a valid position (e.g. every axis has a known motor position & the encoder position has not drifted). If the position is ok, no action is performed; otherwise, the gantry homes.

# Test Plan

Push the API and ODD packages to a robot and test the following workflows:

- [x] Gripper attach
- [x] Pipette attach (left & right)
- [x] Gripper calibrate
- [x] Pipette calibrate

On each one, try it once normally and ensure the gantry doesn't home to the origin before calibrating. Then, run the workflow again but push the gantry enough that it loses position; this time, make sure the gantry homes.

# Changelog

- Added an optional parameter to the `home` command that will skip homing if the position of a mount is valid
- Before calibrating gripper or pipette, call `home` with the proper mount provided just in case the user backdrove the gantry

# Review requests

- Is there anyplace else we would want to add this call? I didn't add it to module calibration because my understanding is that there needs to be a pipette already attached through the normal pipette-attach workflow

# Risk assessment

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/home.py
+++ b/api/src/opentrons/protocol_engine/commands/home.py
@@ -4,6 +4,7 @@ from pydantic import BaseModel, Field
 from typing import TYPE_CHECKING, Optional, List, Type
 from typing_extensions import Literal
 
+from opentrons.types import MountType
 from ..types import MotorAxis
 from .command import AbstractCommandImpl, BaseCommand, BaseCommandCreate
 
@@ -25,6 +26,14 @@ class HomeParams(BaseModel):
             " to ensure accurate homing of the explicitly specified axes."
         ),
     )
+    skipIfMountPositionOk: Optional[MountType] = Field(
+        None,
+        description=(
+            "If this parameter is provided, the gantry will only be homed if the"
+            " specified mount has an invalid position. If omitted, the homing action"
+            " will be executed unconditionally."
+        ),
+    )
 
 
 class HomeResult(BaseModel):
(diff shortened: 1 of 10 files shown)


sfoster1 (approved):
This looks good to me as long as it's a serialization-safe change

api/src/opentrons/protocol_engine/commands/home.py
```diff
@@ -25,6 +26,14 @@ class HomeParams(BaseModel):
             " to ensure accurate homing of the explicitly specified axes."
         ),
     )
+    skipIfMountPositionOk: Optional[MountType] = Field(
```
SyntaxColoring, line 29: It feels kind of weird to me that this is `skipIfMountPositionOk: Optional[MountType]` instead of `skipIfAllPositionsOk: Optional[bool]`.

The PR description mentions pushing too hard when attaching pipettes. Is it possible for that to affect one mount, but not the others?

SyntaxColoring (commented):
Yep, this is serialization-safe. Thanks!

api/src/opentrons/protocol_engine/commands/home.py
fsinapi: Mostly a product of the problem mostly occurring when you're working with attach/detach flows - my thinking here was that the client is basically saying "skip if all of the axes pertinent to this mount that I'm in the process of using are good." We also don't really have a concept of the gantry's position sans a mount in the hardware controller or in any existing commands. 

When attaching a 96ch there's also the chance for the weight of the pipette to cause the left mount to drop down - however, we're homing that axis anyways in the preceding command of any relevant workflow.

mjhuff (approved):
Sorry to keep you waiting - the JS side looks good to me!

PR #13557, opened 2023-09-14
fix(app): fix various app copy

smb2268:
# Overview

This PR makes two copy changes:
1. Detaching pipette failure CTA says "Cancel Detachment" now instead of cancel attachment
2. Firmware update takeover modal on odd now says "Updating pipette firmware...." instead of <Left or Right> Pipette firmware to be less specific for 96 channel

# Test Plan
Screen Shot 2023-09-14 at 2 40 22 PM

# Changelog
See description

# Review requests
Make sure copy is updated properly

# Risk assessment
Low


app/src/assets/localization/en/firmware_update.json
```diff
@@ -5,8 +5,8 @@
   "gantry_y": "Gantry Y",
   "gripper": "Gripper",
   "head": "Head",
-  "pipette_left": "Left Pipette",
-  "pipette_right": "Right Pipette",
+  "pipette_left": "pipette",
+  "pipette_right": "pipette",
```
jerader, line 9: can these be updated into 1 i18n key now?

jerader (approved):
lgtm, just left a question/suggestion for combining the i18n keys that have the same value

PR #13561, opened 2023-09-15
fix(app-shell): delete protocols across both protocol storages 

mjhuff:
Closes RQA-1600

# Overview

Removes protocol files from both versions of the app to prevent protocol re-migration upon navigation to the ProtocolDashboard. This is part of the pre-parity stop-gap work for the app and will be removed after parity.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/e8dc4c75-3c90-4448-b9d0-e76a364f9960

### Fixed Behavior

https://github.com/Opentrons/opentrons/assets/64858653/2fa0a088-05ba-478c-b92c-972ad409f150

# Test Plan

- From the app, delete any protocol that exists within the ~/Library/Application Support/Opentrons/protocols and protocols_v7.0-supported directories.
- Navigate away from ProtocolDashboard and return. Note that the protocol is now properly deleted.

# Changelog

- Add support for deleting protocols from the old protocols directory if they exist.

# Risk assessment

low

### DIFF
--- a/app-shell/src/protocol-storage/index.ts
+++ b/app-shell/src/protocol-storage/index.ts
@@ -201,14 +201,22 @@ export function registerProtocolStorage(dispatch: Dispatch): Dispatch {
           })
         break
       }
-
+      // TODO(jh, 2023-09-15): remove the secondary removeProtocolByKey() after
+      // OT-2 parity work is completed.
       case ProtocolStorageActions.REMOVE_PROTOCOL: {
         FileSystem.removeProtocolByKey(
           action.payload.protocolKey,
           FileSystem.PROTOCOLS_DIRECTORY_PATH
-        ).then(() =>
-          fetchProtocols(dispatch, ProtocolStorageActions.PROTOCOL_ADDITION)
         )
+          .then(() =>
+            FileSystem.removeProtocolByKey(
+              action.payload.protocolKey,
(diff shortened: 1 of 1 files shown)


jerader (commented):
it works-ish, but if I have both 7.0 and 6.3 apps open, deleting requires me to navigate off the protocol cards list for 7.0? It works fine if i have just 7.0 or 6.3 open individually. And 6.3 works if i have both open at the same time. IDK if its also just my dev env being laggy.


https://github.com/Opentrons/opentrons/assets/66035149/0bb92494-b96a-4b23-bfb6-266a46abfc23

mjhuff:
> it works-ish, but if I have both 7.0 and 6.3 apps open, deleting requires me to navigate off the protocol cards list for 7.0? It works fine if i have just 7.0 or 6.3 open individually. And 6.3 works if i have both open at the same time. IDK if its also just my dev env being laggy.

Hm, interesting. Thanks for catching that. Yes, I didn't account for the case where if a protocol didn't exist on the 6.3 version, then the re-render wouldn't happen (because the delete function rejects instead of resolves). This should fix it. The logic is now "do the re-render on the 7.0 app, THEN try deleting from the 6.3 app."

sfoster1 (approved):
Tested and works including @jerader 's case!

PR #13562, opened 2023-09-15
fix(app): fix random white spaces in robot list when sending a protocol

mjhuff:
Closes RQA-1580

# Overview

Recent CSS changes to ChooseRobotSlideout error text created a superfluous Flex component for each AvailableRobot option. This component coupled with the logic of healthyReachableRobots renders an empty Flex, thus creating the white spacing. Because robots may be reordered when re-rendering the slideout, the behavior appears random.

Removing the unneeded Flex component fixes the issue.

### Current Behavior

current behavior

### Fixed Behavior

fixed behavior

# Test Plan

- On ProtocolDashboard, click the overflow menu for a protocol and click Send to Opentrons Flex. Style consistency should match that of the "fixed behavior" image above.
- Repeat a few times, as white spacing errors are inconsistent.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ChooseRobotSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotSlideout/index.tsx
@@ -223,26 +223,24 @@ export function ChooseRobotSlideout(
               selectedRobot != null && selectedRobot.ip === robot.ip
             return (
               <>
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Tested and works for me!

PR #13563, opened 2023-09-15
fix(app): sort current offsets by date to ensure order

jerader:
closes RAUT-722 and RQA-1608

# Overview

The data fetched from the runQuery hook does guarantee that the current offsets return in the same order. To fix this, I added sorting via `createdAt` from the offsets key

# Test Plan

see the ticket attached, basically run LPC on desktop with offsets already in the table. Make sure they don't reorder during LPC. The reordering starts after the initial page of the LPC flow and ends when you complete LPC

# Changelog

added sorting of the current offsets

# Review requests

see test plan

# Risk assessment

low


sfoster1 (approved):
Tested on robot, looks good to me!

koji:
Maybe create a sort function for this?
Not recommend to overwrite the data.

PR #13564, opened 2023-09-15
fix(api): disengage gripper z when not in use at maintenance position

ahiuchingau:
# Overview
Gripper Z drops during the 96-channel pipette detach flow when the right mount Z carriage gets released from the mount. Turns out it's because the gripper z hold current is too low and vibration from right mount caused it to fall. We can fix this by disengaging the z, which engages the e-brake so this will never happen.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
+++ b/api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
@@ -107,6 +107,11 @@ class MoveToMaintenancePositionImplementation(
             )
 
         if params.mount != MountType.EXTENSION:
+
+            # disable z since we don't need it
+            if ot3_api.has_gripper():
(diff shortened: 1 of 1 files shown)


api/src/opentrons/protocol_engine/commands/calibration/move_to_maintenance_position.py
```diff
@@ -107,6 +107,11 @@ async def execute(
             )
 
         if params.mount != MountType.EXTENSION:
+
+            # disable z since we don't need it
```
SyntaxColoring: Can we flesh out this comment with notes like you added in the PR description? Otherwise, it's not really clear that disengaging this axis is an actively beneficial thing to do.

SyntaxColoring (approved):
Code makes sense to me.

sfoster1 (approved):
Looks good but agree about the comment

SyntaxColoring (approved):
🦓

PR #13565, opened 2023-09-15
feat(api): After overpressure errors, leave the pipette where it is

SyntaxColoring:
# Overview

Implements a request from the hardware testing team, via @sfoster1.

When a run fails because of a pipette overpressure event, it's easier to investigate it if we handle it sort of like E-stops. Leave the pipette exactly where it is—don't home.

# Test Plan

* [ ] Make sure this doesn't break E-stop handling
* [ ] Manually trigger a pipette overpressure somehow and see that it leaves the pipette in-place?

# Review requests

None in particular.

# Risk assessment

Low.


sfoster1 (approved):
Tested and works after that latest commit, pipette just doesn't move

PR #13566, opened 2023-09-15
fix(api): gripper idling error post protocol run drop tip

ahiuchingau:
# Overview
If the gripper is unhomed, we shouldn't try to idle the gripper jaw. We saw an error when the pipette tries to drop tip during the post-protocol cleanup, the jaw had been disengaged and could not idle.


sfoster1 (approved):
nice

PR #13567, opened 2023-09-15
fix(shared-data): thermocycler and related labware offset measurements

sanni-t:
Addresses RQA-1522

# Overview

Fixes multiple issues in labwareOffsets related to thermocyclerModuleV2 and its compatible labware & calibration adapter.
Achieves that by shifting the 'labware origin Z' for labware on a thermocycler to be the top plane of the metallic cylinders on the thermocycler that house the labware. This shifts the z-labwareOffset by +10.7mm. 
It also makes the top plane similar in concept to the magnetic block's top plane.

NOTE: These changes now require all TC-compatible labware to use a stacking overlap offset for the TCV2. I have addressed the two recommended labware in this PR. I'll be following up with the other commonly used ones shortly.

# Test Plan

Test this on both Flex & OT2:
- push this branch to the robot.
- on the Flex, perform TC module calibration. Check that the calibrated offset isn't too large. No calibration on OT2 so skip this step
- upload the attached protocol that loads a TC and different labware on it
- go through LPC steps. Verify that the offsets for labware on the thermocycler are very minimal

```
metadata = {
    'ctxName': 'gripper test with thermocycler',
}

requirements = {
    'robotType': 'OT-3',
    'apiLevel': '2.15'
}


def run(protocol):
    # LABWARE AND MODULES
    thermocycler = protocol.load_module('thermocycler module gen2')
    thermocycler.open_lid()

    # DECK SETUP AND LABWARE FOR PIPETTES
    pcr_plate = thermocycler.load_labware('opentrons_96_wellplate_200ul_pcr_full_skirt')
    reservoir = protocol.load_labware('nest_12_reservoir_15ml', location = "B3")
    tiprack_50 = protocol.load_labware('opentrons_flex_96_tiprack_50ul', location = "A2")

    # LOAD PIPETTES
    m50 = protocol.load_instrument("p50_multi_gen3", "right", tip_racks=[tiprack_50])

    # PIPETTE COMMANDS
    m50.pick_up_tip()
    m50.aspirate(50, reservoir.wells_by_name()['A1'].bottom(z=3))
    m50.dispense(50, pcr_plate.wells_by_name()['A1'].bottom(z=1.5))
    m50.drop_tip()

    protocol.move_labware(
            labware=pcr_plate,
            new_location=2,
            use_gripper=True,
        )
    protocol.move_labware(
            labware=pcr_plate,
            new_location=thermocycler,
            use_gripper=True,
        )
```
# Changelog

- updated `labwareOffset` for thermocyclerV2 and updated the slot transform for Flex.
- added thermocyclerV2's `stackingOffsetWithModule` to armadillo & Opentrons 200uL PCR plates

# Review requests

Make sure the calculations are correct and testing confirms that.

# Risk assessment

Medium


api/tests/opentrons/protocol_engine/state/test_module_view.py
```diff
@@ -345,7 +345,7 @@ def test_get_module_offset_for_ot2_standard(
         (
             lazy_fixture("thermocycler_v2_def"),
             DeckSlotName.SLOT_7.to_ot3_equivalent(),
-            LabwareOffsetVector(x=-20.005, y=67.96, z=0.26),
+            LabwareOffsetVector(x=-20.005, y=67.96, z=10.96),
```
sanni-t, line 348: This is what the final z offset for a TC labware on Flex will be

shared-data/labware/definitions/2/opentrons_calibration_adapter_thermocycler_module/1.json
```diff
@@ -13,7 +13,14 @@
   "dimensions": {
     "xDimension": 116.25,
     "yDimension": 80.5,
-    "zDimension": 27.5
+    "zDimension": 26.4
```
sanni-t, line 16: This z measurement seems to have changed over different versions and 26.4 is the correct, final version.

shared-data/module/definitions/3/thermocyclerModuleV2.json
```diff
@@ -4,8 +4,8 @@
   "model": "thermocyclerModuleV2",
   "labwareOffset": {
     "x": 0,
-    "y": 68.06,
-    "z": 98.26
+    "y": 68.8,
+    "z": 108.96
```
sanni-t, line 8: Previously this z was the height at which the Opentrons/Armadillo 200uL PCR plate sat. We are changing that offset to indicate the top plane of the module's metal cylinders that house the labware, which is 10.7mm higher than the previous offset plane and hence the change in z.

The change in y just seems to be a previous measurement error.

SyntaxColoring: > We are changing that offset to indicate the top plane of the module's metal cylinders that house the labware

We started discussing this in-person. For backwards compatibility reasons, I don't think we should make this change like this.

People in the field are already using Thermocyclers with labware definitions that we don't control. So, if we do this, those movements will suddenly become ~1cm too high. To fix that, users will either have to compensate with Labware Position Check, which we know we don't want to encourage for this kind of thing, or they'll have to edit their labware definitions, which they'll have to do by hand because Custom Labware Creator doesn't support `stackingOffsetWithModule`.

EDIT: Also, what does this do for JSON protocols, which will have the old definitions embedded within them?

I do think it's a good idea to change how overlaps are measured and decouple ourselves from the Armadillo plate. But I think it should work something like this:

* Keep the interpretation of `labwareOffset` for Thermocyclers exactly as it exists currently, with reference to the Armadillo plate.
* Add a *new* `metalCylinderTopCenter` coordinate to the Thermocycler module definitions.
* Add some way for labware definitions to specify overlap offsets relative to `metalCylinderTopCenter`.

  Strawman example:

  ```js
  "stackingOffsetWithModule": {
    "thermocyclerModuleV2": {
      "mateConnector": "metalCylinderTopCenter",
      // Offsets from the origin of this labware to where the top-center
      // of the metal cylinders ends up fitting within it.
      "x": 1.23,
      "y": 4.56,
      "z": 7.89,
    }
  }
  ```

  (*Mate connector* is terminology that [I'm borrowing from Onshape](https://cad.onshape.com/help/Content/mateconnector.htm). It looks like [SolidWorks might call them](https://www.youtube.com/watch?v=U8IVL6uBl_c) *mate references.*)

* **Fallback logic:** If you're putting a labware on a Thermocycler, we calculate the overlap using `metalCylinderTopCenter` if the labware defines that. Or, otherwise, we calculate the overlap assuming the labware sits like an Armadillo plate, like we do today.

SyntaxColoring: Separately: Whatever we end up with, can we describe it in the module definition schema, in the `description` of `labwareOffset`? It appears undocumented at the moment but I think @jbleon95 is working on that.

SyntaxColoring:
[Adapted the test protocol to run on an OT-2](https://gist.github.com/SyntaxColoring/879fecb12bc0da3f61de5b032c330db3). Here are the Labware Position Check offsets that I ended up with:

Screenshot 2023-09-19 at 4 20 16 PM

Is it possible that the offset of the labware on a Thermocycler has an error of ~0.6mm for the OT-2?

sanni-t:
> Is it possible that the offset of the labware on a Thermocycler has an error of ~0.6mm for the OT-2?

Let me check with hardware team

sfoster1 (approved):
Hardware says it looks good!

shared-data/module/definitions/3/thermocyclerModuleV2.json
SyntaxColoring: Discussed in-person.

The thinking is that it's okay to change this because `stackingOffsetsWithModule` is not in any production protocols or labware. However, a follow-up *will* need to add a fallback to calculate the stacking the old way when `stackingOffsetsWithModule` is not present.

sanni-t:
**Flex testing summary for future reference:**

Before changes on this branch, module calibration offset of the thermocycler was:
`[0.3515625, -1.2890625, -1.1905000000000285]`
After changes on this branch, it changed to:
`[0.1640625, -1.1015625, 0.20574999999998766]`

LPC offset for Opentrons PCR plate on the thermocycler (using an 8-channel pipette), changed from 1.7mm to 0.4mm after changes from this branch.

Similar improvement was observed on other Flex bots.

PR #13568, opened 2023-09-15
fix(app): add retract axis sequence before dropTip when cancelling LPC

smb2268:
fix RQA-1605

# Overview

When cancelling LPC we issue a `dropTip` command for each pipette. Before this, we should issue a sequence of `retractAxis` commands to prevent any possible collisions.

# Test Plan
Run through LPC, cancel early, and ensure that the pipette is retracted to the top before moving to drip tips

# Changelog
Add `retractAxis` commands before issuing `dropTip` in the `handleCleanUpAndClose` LPC function

# Review requests
Test this out!

# Risk assessment


sfoster1 (approved):
Tested on robot, works for me!

SyntaxColoring (approved):
🆙

PR #13570, opened 2023-09-15
fix(app): labware render viewbox adjustment

jerader:
# Overview

Noticed that the labware view of adapter in the labware tab's slideouts of the desktop app was not positioned properly for adapters. I must've missed this case when i went through and fixed it for other labware renders.

# Test Plan

Open the desktop app, go to the labware tab. Click on an adapter definition. See that the adapter render in the slideout is positioned properly.

Before:
Screen Shot 2023-09-15 at 4 55 55 PM

After:
Screen Shot 2023-09-15 at 4 55 33 PM

# Changelog

changed viewbox coordinates

# Review requests

see test plan

# Risk assessment

low


sfoster1 (approved):
Tested by looking at all the labwares, seems right to me

PR #13571, opened 2023-09-15
fix(shared-data): Update low-volume accuracy functions and bottom plunger position

andySigler:
# Overview

Tested with protocol running on DVT flex, and measured %D and %CV using plate-reader. Several 8ch multis pass our spec at 1ul, have not yet tested 1ch pipettes but can assume similar results. The 1ch low-volume accuracy function has not finished validation in SZ, but this function is guaranteed to be an incremental improvement for the beta release from what we have right now.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
sweet!

PR #13572, opened 2023-09-15
fix(app): copy changes from September 11–14 QA

ecormany:
# Overview

**Desktop** copy changes based on Sue's copy QA. As always, the goal is to make the software match the designs.

# Test Plan

Tested everything I could using `make -C app dev`. Some flows aren't accessible without access to Flex hardware.

# Changelog

See the Copy QA tab of the spreadsheet linked in [RUXD-1049](https://opentrons.atlassian.net/browse/RUXD-1049). Anything marked with this PR's number in the "Ticket" column should be covered here.

- changed localization values
- changed test values when present
- changed a couple actual layout things, hoo boy

# Review requests

- Try _all_ the strings! 
- How much do we care about renaming localization keys? Everything _works_ now, even if key doesn't match value very well. And a good number of these names don't appear verbatim in the components, so it'd require a bit more refactoring to change them. I leave it to the software team to decide whether that's time well spent right now.

# Risk assessment

Low-ish. Lots of very small changes.

[RUXD-1049]: https://opentrons.atlassian.net/browse/RUXD-1049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/assets/localization/en/protocol_setup.json
```diff
@@ -62,20 +62,19 @@
   "instruments_connected_plural": "{{count}} instruments attached",
   "instruments_connected": "{{count}} instrument attached",
   "instruments": "Instruments",
-  "labware_help_example": "<block>Alice is using a labware in Slot 6. During labware position check, she adjust the pipette position while checking the A1 of the labware to 0.2mm in X, and 1.2mm in Z. Later, Bob is preparing to run the same protocol on that robot. The labware offset that Alice created in Slot 6 will be applied for Bob’s protocol unless he changes or clears it.</block>",
```
ecormany, line 65: Pretty sure this isn't referenced anywhere. I _hope_ it isn't.

jerader: nope i don't think it is.

app/src/organisms/GripperWizardFlows/Success.tsx
```diff
@@ -49,7 +49,7 @@ export const Success = (
     },
     [SUCCESSFULLY_DETACHED]: {
       header: t('gripper_successfully_detached'),
-      buttonText: t('shared:exit'),
+      buttonText: i18n.format(t('shared:exit'), 'capitalize'),
```
jerader, line 52: nice use of an i18n formatter!

app/src/assets/localization/en/gripper_wizard_flows.json
```diff
@@ -25,17 +25,17 @@
...
-  "remove_labware_to_get_started_detaching": "To get started, remove labware from the deck and clean up the working area to make detachment easier. Also gather the needed equipment shown to the right.",
-  "remove_labware_to_get_started_recalibrating": "<block>To get started, remove labware from the deck and clean up the working area to make detachment easier. Also gather the needed equipment shown to the right.</block><block>The calibration pin is included with the gripper and should be stored on its right side above the jaws.</block>",
+  "remove_labware_to_get_started_attaching": "<block>To get started, remove labware from the deck and clean up the working area to make attachment and calibration easier. Also gather the needed equipment shown to the right.</block><block>The calibration pin is included with the gripper and should be stored on its right side above the jaws.</block>",
+  "remove_labware_to_get_started_detaching": "<block>To get started, remove labware from the deck and clean up the working area to make detachment easier. Also gather the needed equipment shown to the right.</block>",
```
jerader, line 37: nice use of a tag nested into an i18n value!

jerader (approved):
lgtm! 🐎

app/src/assets/localization/en/gripper_wizard_flows.json
ecormany: This copy had been flagged as having some formatting issues. I added this to mirror similar strings on the adjacent lines. Hopefully all it needed was the tag.

app/src/assets/localization/en/protocol_setup.json
```diff
@@ -111,9 +110,9 @@
   "module_mismatch_body": "Check that the modules connected to this robot are of the right type and generation",
   "module_name": "Module Name",
   "module_not_connected": "Not connected",
-  "module_setup_step_description_plural": "Plug in and turn on the required modules via the robot's USB Ports. Place the modules as shown in the deck map.",
-  "module_setup_step_description": "Plug in and turn on the required module via the robot's USB Port. Place the module as shown in the deck map.",
-  "module_setup_step_title": "Module Setup",
+  "module_setup_step_description_plural": "Install the required modules and power them on.",
```
ecormany, line 113: @jerader This could be refactored out now that the two strings are identical. I couldn't tell exactly where the singular/plural logic is in ProtocolRunSetup.tsx

app/src/assets/localization/en/gripper_wizard_flows.json
jerader: yeah, the tag is needed because the component is using the i18n `Trans` component to give the string multiple designs. The tag isn't needed on its own but since the string is displayed using `Trans`, it is needed.

jerader: Here is the component it is being used in. Adding the block gives it the `p` tag styling.

` <Trans
          t={t}
          i18nKey={bodyI18nKey}
          components={{ block: <StyledText as="p" /> }}
        />`

app/src/assets/localization/en/protocol_setup.json
jerader: as far as i can tell, `module_setup_step_description` and `module_setup_step_description_plural` aren't being used at all lol, so we can probably delete them. But lets just keep them in for now 🤷 i'm sure at some point post-launch we will do an audit of the i18n strings that are not in use.

ecormany: Yeah, changing the `plural` string definitely changed the text here in run setup: 
image

app/src/organisms/GripperWizardFlows/Success.tsx
ecormany: i got the syntax wrong at first and blew up like 100 tests, but i figured it out 😅

PR #13577, opened 2023-09-18
feat(protocol-designer): 96 flat bottom adapter support + waitForTemp…

jerader:
… command fix

closes RAUT-726, RQA-1544, RQA-1650, RQA-1656

# Overview

Sorry to address multiple things in this PR but basically this:
1. adds support for the 96 flat bottom aluminum block adapter
2. fixes a bug where the `waitForTemperature` command was being assigned to the wrong module
3. fixes a UI issue with the step creation button
4. fixes bug where the module dropdown for which slot to go into when editing modules is wrong

# Test Plan

sandbox: https://sandbox.designer.opentrons.com/pd_96-flat-bottom-adapter-support/

For the adapter stuff:

Create a protocol for Flex or OT-2. Add a temperature module, thermocycler, magnetic module/block, and heater-shaker. In deck setup on the design tab, add the 96 Aluminum flat bottom plate. See that there are 7 options available for labware to put on top of it (6 corning and 1 NEST). Now, directly on the deck, add the 96 Aluminum flat bottom plate. See that those 7 options are also available. Now, check the modules, there SHOULD NOT be the 96 aluminum flat bottom plate option available. Also, mess around with dragging/dropping. Make sure that the only labware that can go on top of the adapter are the compatible labware.

For the `waitForTemperature` bug:

Create a Flex or OT-2 protocol with the heater-shaker and temperature module. Add a temperature command in the timeline for the temperature module and select "build pause now". Do the same thing for the heater-shaker and select "build pause now". Export the protocol. See in the json file that the `waitForTemperature` commands are assigned to the correct modules in the correct order. 

For the Step creation button bug:

Check the screenshots

Before:
Screen Shot 2023-09-18 at 12 28 19 PM

After:
Screen Shot 2023-09-18 at 12 26 49 PM

# Changelog

- add compatible labware to the new aluminum block adapter
- change the zIndex on the step creation button
- plug in the moduleId in `saveSetTempFormWithAddedPauseUntilTemp` - we forgot to plug it in so it kept defaulting to the heater-shaker moduleId if both modules were in the protocol or using the temp module if the heater-shaker wasn't present.

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/listButtons.css
+++ b/protocol-designer/src/components/listButtons.css
@@ -1,7 +1,7 @@
 @import '@opentrons/components';
 
 .list_item_button {
-  z-index: 10;
+  z-index: 1;
   position: relative;
   padding: 1rem 2rem 1.25rem 2rem;
 }
--- a/protocol-designer/src/ui/steps/actions/thunks/index.ts
+++ b/protocol-designer/src/ui/steps/actions/thunks/index.ts
@@ -242,11 +242,11 @@ export const saveSetTempFormWithAddedPauseUntilTemp: () => ThunkAction<any> = ()
       },
     })
   )
-  const tempId = unsavedSetTemperatureForm?.moduleId
+  const tempertureModuleId = unsavedSetTemperatureForm?.moduleId
   dispatch(
     changeFormInput({
       update: {
-        moduleId: tempId,
+        moduleId: tempertureModuleId,
       },
     })
   )
--- a/protocol-designer/src/ui/steps/actions/thunks/index.ts
+++ b/protocol-designer/src/ui/steps/actions/thunks/index.ts
@@ -242,6 +242,14 @@ export const saveSetTempFormWithAddedPauseUntilTemp: () => ThunkAction<any> = ()
       },
     })
   )
+  const tempId = unsavedSetTemperatureForm?.moduleId
+  dispatch(
+    changeFormInput({
+      update: {
+        moduleId: tempId,
+      },
+    })
+  )
   dispatch(
     changeFormInput({
       update: {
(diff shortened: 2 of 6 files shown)


protocol-designer/src/components/listButtons.css
```diff
@@ -1,7 +1,7 @@
 @import '@opentrons/components';
 
 .list_item_button {
-  z-index: 10;
```
jerader, line 4: this z-index being so high makes me think there was a reason for it but i can not figure out what that reason is 🤔 nothing appears to be broken after smoke testing it.

protocol-designer/src/ui/steps/actions/thunks/index.ts
```diff
@@ -242,6 +242,14 @@ export const saveSetTempFormWithAddedPauseUntilTemp: () => ThunkAction<any> = ()
       },
     })
   )
+  const tempId = unsavedSetTemperatureForm?.moduleId
```
koji: maybe can we use `tempModuleId` since if we didn't read the code carefully, we would think temporaryId.

jerader: oh ya good call!

protocol-designer/src/components/listButtons.css
koji: How about other components for PD?
Are they using z-index 1?

koji (approved):
The changes looks good to me.

One quick question.
Screenshot 2023-09-18 at 1 13 25 PM

D1 is not available for h/s, is this the right behavior?
I'm asking this because module calibration's select location for h/s allows to use `D1`

https://github.com/Opentrons/opentrons/blob/e55eb32c4a28e48a108463cb1317cb2dcb33540b/app/src/organisms/ModuleWizardFlows/constants.ts#L25

jerader:
> D1 is not available for h/s, is this the right behavior? I'm asking this because module calibration's select location for h/s allows to use `D1`
> 
> https://github.com/Opentrons/opentrons/blob/e55eb32c4a28e48a108463cb1317cb2dcb33540b/app/src/organisms/ModuleWizardFlows/constants.ts#L25

@koji ~~the h-s should work in `D1`. do you know if you had another module in `D1`? I believe the slots are filtered out of the dropdown if there is a module in its place.~~, nevermind just checked. I think this is a bug that its not available. I'll investigate. Thanks!

protocol-designer/src/components/listButtons.css
jerader: other z-indexes in PD are in modals, i don't think there's a reason why we would want the step creation button to render on top of the modals.

PR #13578, opened 2023-09-18
refactor(protocol-designer): edit module section fixes

jerader:
closes RQA-1131 and RQA-1077

# Overview

Changes up a few things in edit module section in PD
1. change the text for TC GEN2 to say "A1 + B1" instead of "B1" and GEN1 should say "Slot 7,8,10,11" instead of "Slot 7" - these changes match the app
2. change order of slot drop down in edit module modal to go from D1 -> A1 to A1 -> D1

# Test Plan

Create a protocol on the Flex with a TC GEN2, look at the position in the module section, should say `A1+B1` for the slot. Add another module, make sure in the slot dropdown, the slots are listed alphabetically from top to bottom and that the correct slot position is highlighted on the map.

Screen Shot 2023-09-18 at 3 44 27 PM
Screen Shot 2023-09-18 at 3 45 00 PM

# Changelog

- changed the order of listed slots for modules
- change slot display name for TCGEN2

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modules/ModuleRow.tsx
+++ b/protocol-designer/src/components/modules/ModuleRow.tsx
@@ -74,7 +74,7 @@ export function ModuleRow(props: Props): JSX.Element {
   // If this Module is a TC deck slot and spanning
   // populate all 4 slots individually
   if (slot === SPAN7_8_10_11_SLOT) {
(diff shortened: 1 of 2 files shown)


mjhuff (approved):
lgtm! 🍁

PR #13579, opened 2023-09-18
fix(app): update gripper probe assets, ensure size is same as pipette cal

smb2268:
fix RQA-794

# Overview
Match size of gripper calibration video to that of the pipette cal screen and update assets to final ones in drive (I couldn't tell if the final ones were in the code so I replaced them just in case)

# Test Plan
Run through gripper calibration, ensure probing video is the correct size and not big like the screenshot in the linked ticket

# Changelog
1. Add flex container to restrict video size
2. Update probe front and back jaw assets

# Review requests
Test this out!

# Risk assessment
Low


jerader (approved):
lgtm 🪴

mjhuff (approved):
Tested per instructions - lgtm!

PR #13581, opened 2023-09-18
fix(api): Removal of OT3 to Flex load name handling

CaseyBatten:
Older naming conventions for tipracks supported load names utilizing "OT3" as an alias for "Flex", which this removes the definitions and handlers for.

# Overview
OT3 tiprack naming convention support removed to satisfy launch naming conventions. Required removal of all interactions with resolve_loadname handler alongside the relevant dictionary of names. 

Mapped names removed and reference to standard loaded name used instead. Load_name test data under test_protocol_core adjusted.

Of note, issue found in test_protocol_core under **test_load_labware_on_labware**. Originally, load_name provided was "labware_some" and was given to the resolution handler. This passed in unmodified build. Removal of references to resolve_loadname cause this test to fail, as the name "labware_some" would result in the following error:

**TypeError: cannot unpack non-iterable NoneType object**

This is corrected by adjusting the test case name to "some_labware" to prevent inconsistency between subject call and load name expectations, allowing the test to pass.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/load_labware_params.py
+++ b/api/src/opentrons/protocol_api/core/engine/load_labware_params.py
@@ -38,6 +38,7 @@ _APILEVEL_2_14_OT_DEFAULT_VERSIONS: Dict[str, int] = {
 class AmbiguousLoadLabwareParamsError(RuntimeError):
     """Error raised when specific labware parameters cannot be found due to multiple matching labware definitions."""
 
+
 def resolve(
     load_name: str,
     namespace: Optional[str],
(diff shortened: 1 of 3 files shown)


TamarZanzouri (commented):
Changes look good to me! there is a CI test that is failing. once that is fixed this is good to merge

sfoster1 (approved):
Looks good to me once we fix CI!

CaseyBatten:
This PR is being closed as it has been replaced by a similar PR built off and for 7.0 Chore release alongside a few other corrections to assist with consistency and solving CI issues. New PR: https://github.com/Opentrons/opentrons/pull/13606

PR #13582, opened 2023-09-18
feat(api): Add nozzle configuration manager for partial tip support

Laura-Danielle:
# Overview

With this PR you will now be able to configure your pipette to have different nozzle configurations to support partial tip in the hardware controller. The one thing it does _not_ handle right at this moment is modifying current based on configuration. I realized during this PR that having a current scalar does not work -- especially with a 96 channel -- because it will quickly exceed max current.

# TODO

- [X]  Fix failing tests. It's possible that I'm no longer grabbing the 'exact' same location for the XY_CENTER locations.
- [X]  Decide if it still makes sense to support `NOZZLE` overrides when a tip is attached to a pipette.

# Test Plan

Please verify that all of the following operations still work as expected:
- [ ]  attaching/de-attaching pipettes
- [ ] Dropping tip in trash
- [ ] Moving to labware during LPC
- [ ] Attaching a pipette probe and performing auto-calibration

# Changelog

- Adds a nozzle configuration manager
- Adds functions in hw pipette/hw controller to update the nozzle configuration

# Review requests

Let me know if anything seems weird to you aside from physical testing on the robot. Is the interface for nozzle names OK as well?

# Risk assessment

High. This touches a critical component of the code which affects robot positioning.


api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,145 @@
...
+
+@dataclass
+class CurrentNozzleConfiguration:
+    starting_nozzle: str
+    number_of_nozzles: int
+    starting_index: int
+    ending_index: int
+
+
+class IncompatibleNozzleConfiguration(Exception):
```
sfoster1: should be an enumerated error

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,145 @@
...
+        nozzle_map: Dict[str, List[float]],
+        nozzle_configuration: CurrentNozzleConfiguration,
+        current_scalar: float,
+    ) -> None:
+        sorted_nozzlemap = list(nozzle_map.keys())
+        sorted_nozzlemap.sort(key=lambda x: int(x[1::]))
+        nozzle_map_ordered_dict = OrderedDict(
+            {k: nozzle_map[k] for k in sorted_nozzlemap}
+        )
+        self._nozzle_map: List[Tuple[str, List[float]]] = list(
```
sfoster1: feels like internally we should keep the nozzle offsets as `Point`s

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,145 @@
...
+            nozzle_map_ordered_dict.items()
+        )
+        self._nozzle_map_key_lookup = list(nozzle_map_ordered_dict.keys())
```
sfoster1: can't we just store the nozzle map as an ordered dict rather than as an array of values plus an array of keys with identical ordering? feels kind of fragile

sfoster1 (changes requested):
I like where this is going but I think there's some stuff we should change.

First, on the critical point thing - XY_CENTER is the geometric center of the pipette footprint, that's it and that's all. It doesn't, can't, and shouldn't align to the position of a specific nozzle because it's not _about_ a nozzle - it's about the pipette as a whole. It should remain that geometric center, which the nozzle manager can tell you about because you've encoded nozzle maps such that (0,0) is the geometric center.

Second, there's a number of things in nozzle manager that I don't love - like keeping the nozzles as a list of values plus list of keys rather than a dict, for instance - that seem to be done to keep a concept of a logically-linear arrangement of tips. For instance, you pick up tips starting at some nozzle and continuing for a number of tips; you set a nozzle configuration with a starting tip and a number of tips. I think a lot of that gets simpler if you instead define nozzle configurations as oriented rectangles of nozzles, and store the back-left and front-right. 

This gives you a couple good features:
- it's very easy to do something like "upper left quadrant" as a configuration, or "left two columns but not bottom row" with pretty intuitive interfaces
- you're storing the nozzle locations of semantic relevance in the back-left and front-right, which means that if you want to know their positions (e.g. for critical points) you don't have to do calculations based on a hardcoded row/column number, you just look it up from the config
- we will at some point need to treat these nozzle configurations this way anyway on some layer so we can do collision checks with tips

This does come at the cost of having more complex code to think in terms of rectangles rather than lists but I really think it's worth it.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self._current_nozzle_configuration = nozzle_configuration
+
+        self._xy_nozzle: Final[Point] = self._get_nozzle_for(CriticalPoint.XY_CENTER)
+        self._front_nozzle: Final[Point] = self._get_nozzle_for(
+            CriticalPoint.FRONT_NOZZLE
+        )
+        self._current_scalar: Final[float] = current_scalar
+
+    def _get_nozzle_for(self, critical_point: CriticalPoint) -> Point:
```
sfoster1: nit: this really isn't returning a nozzle, this is returning a point, call it something else maybe?

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+    @property
+    def nozzle_offset(self) -> Point:
+        return self._nozzle_map[self._current_nozzle_configuration.back_left_nozzle]
+
+    def update_nozzle_with_tips(
+        self, back_left_nozzle: str, front_right_nozzle: str
+    ) -> None:
+        if back_left_nozzle > front_right_nozzle:
+            raise IncompatibleNozzleConfiguration(
+                message=f"Starting nozzle {back_left_nozzle} must not be greater than the ending nozzle {front_right_nozzle}.",
```
sfoster1: can we phrase this differently it's very weird from a user perspective to consider "H1" as "greater than" "A1"

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+            front_right_nozzle="H1",
+            configuration=NozzleConfigurationType.FULL,
+        )
+        return cls(nozzle_map, starting_nozzle_config, current_scalar)
+
+    @property
+    def nozzle_offset(self) -> Point:
+        return self._nozzle_map[self._current_nozzle_configuration.back_left_nozzle]
+
+    def update_nozzle_with_tips(
```
sfoster1: should this be called this? it doesn't really seem to have anything to do with tips, it just sets the nozzle config

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+
+        self._current_nozzle_configuration = CurrentNozzleConfiguration(
+            back_left_nozzle=back_left_nozzle,
+            front_right_nozzle=front_right_nozzle,
+            configuration=self._determine_nozzle_configuration(
+                back_left_nozzle, front_right_nozzle
+            ),
+        )
+
+    def update_nozzle_with_critical_point(
```
sfoster1: this doesn't really update anything, right? it just calculates and returns the critical point

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+    ) -> None:
+        sorted_nozzlemap = list(nozzle_map.keys())
+        sorted_nozzlemap.sort(key=lambda x: int(x[1::]))
+        self._nozzle_map: Dict[str, Point] = OrderedDict(
+            {k: Point(*nozzle_map[k]) for k in sorted_nozzlemap}
+        )
+        self._first_nozzle = next(iter(self._nozzle_map.values()))
+        self._last_nozzle = next(reversed(self._nozzle_map.values()))
+
+        self._current_nozzle_configuration = nozzle_configuration
```
sfoster1: should this or its `.configuration` be accessible somewhere as like a property?

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+    QUADRANT = "QUADRANT"
+    SINGLE = "SINGLE"
+    FULL = "FULL"
+
+
+@dataclass
+class CurrentNozzleConfiguration:
+    back_left_nozzle: str
+    front_right_nozzle: str
+    configuration: NozzleConfigurationType
```
sfoster1: looking at what `NozzleConfigurationType` means, maybe this should be a property?

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            return NozzleConfigurationType.QUADRANT
+
+    @classmethod
+    def build_from_nozzlemap(
+        cls, nozzle_map: Dict[str, List[float]], current_scalar: float
+    ) -> "NozzleConfigurationManager":
+        starting_nozzle_config = CurrentNozzleConfiguration(
+            back_left_nozzle="A1",
+            # TODO gotta fix this to be a real nozzle
```
sfoster1: I think this might benefit from `nozzlemap` being its own class that you build from a dict, that has accessors like "back_left` and `front_right` and `front_left` and `configuration` (i.e. row/column/rectangle).

At that point, `NozzleManager` can hold two `NozzleMap`s - the `physical` and the `current`. `NozzleManager.__init__` takes a `NozzleMap`, and this builder becomes `build_from_nozzledict` and just builds a map and then passes it in.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+            front_right_nozzle="H1",
+            configuration=NozzleConfigurationType.FULL,
+        )
+        return cls(nozzle_map, starting_nozzle_config, current_scalar)
+
+    @property
+    def nozzle_offset(self) -> Point:
```
sfoster1: can we call this like `standard_nozzle_offset` or something, IMO it needs a further identification than just "nozzle" because the point of this class is to handle having multiple nozzles

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,190 @@
...
+        return self._nozzle_map[self._current_nozzle_configuration.back_left_nozzle]
+
+    def update_nozzle_with_tips(
+        self, back_left_nozzle: str, front_right_nozzle: str
+    ) -> None:
+        if back_left_nozzle > front_right_nozzle:
```
sfoster1, line 151: IMO these checks want to be a free function somewhere or on that `NozzleMap` class

sfoster1 (changes requested):
Liking this a lot more now; I think we can tighten up some naming stuff and we would benefit from a sort of first-class nozzlemap class though.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
Laura-Danielle: yeah, I am doing that in the follow-up current setting PR because we're not currently using it here.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
Laura-Danielle: technically it updates the critical point with the tip length, but I can change the name a little bit.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,277 @@
...
+        cls,
+        nozzle_difference: Point,
+        physical_nozzlemap_length: int,
+        current_nozzlemap_length: int,
+    ) -> "NozzleConfigurationType":
+        """
+        Determine the nozzle configuration based on the starting and
+        ending nozzle.
+
+        :param nozzle_difference: the difference between
```
sfoster1: incomplete sentence?

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,277 @@
...
+    COLUMN = "COLUMN"
+    ROW = "ROW"
+    QUADRANT = "QUADRANT"
+    SINGLE = "SINGLE"
+    FULL = "FULL"
+
+    @classmethod
+    def determine_nozzle_configuration(
+        cls,
+        nozzle_difference: Point,
```
sfoster1, line 34: I think this logic works but it really feels like it should be about nozzle names rather than points, i dunno

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def __str__(self) -> str:
+        return f"back_left_nozzle: {self.back_left} front_right_nozzle: {self.front_right} configuration: {self.configuration}"
+
+    @property
+    def starting_nozzle_offset(self) -> Point:
+        return self.map_store[self.starting_nozzle]
+
+    @property
+    def xy_nozzle(self) -> Point:
```
sfoster1: I really don't like having it called this because it doesn't actually represent the position of any nozzle. can we call it `xy_center` or something?

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -0,0 +1,277 @@
...
+        difference = self.map_store[self.front_right] - self.map_store[self.back_left]
+        return self.map_store[self.back_left] + Point(
+            difference[0] / 2, difference[1] / 2, 0
+        )
+
+    @property
+    def front_nozzle(self) -> Point:
```
sfoster1: maybe front nozzle offset?

sfoster1 (approved):
I haven't yet tested, this is just about the code.

I think this code is good to merge (once I test it) but I think pretty soon we should follow up with a refactor that treats the physical layout from the defs of the nozzles as the basis of data and gets away from the internozzle distance. If we can encode the row/column layout in the pipette general def and make decisions about backleft/front right from those rather than playing around with nozzle offsets i think the code will be cleaner

Laura-Danielle:
@sfoster1 I'll make a ticket to track those changes, but I think it should be low priority as compared to the rest of the partial tip feature set.

No rush on the testing also. We really shouldn't pick up tips until my follow-up current PR.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
Laura-Danielle: yes will fix.

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
Laura-Danielle: we can't determine the actual configuration unless we had all four corners of the square, which would still require some type of logic similar to this.

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -324,44 +310,22 @@ def critical_point(self, cp_override: Optional[CriticalPoint] = None) -> Point:
...
-            mod_offset_xy = list(offsets)
-        mod_and_tip = Point(
-            mod_offset_xy[0], mod_offset_xy[1], mod_offset_xy[2] - tip_length
-        )
 
-        cp = mod_and_tip + instr
+        instr = Point(*self._pipette_offset.offset)
+        cp_with_tip_length = self._nozzle_manager.critical_point_with_tip_length(
+            cp_override,
+            self.current_tip_length if cp_override != CriticalPoint.NOZZLE else 0.0,
```
sfoster1, line 317: let's preserve the behavior from before where this is
```
if not self.has_tip or cp_override == CriticalPoint.NOZZLE
    tip_length = 0
else:
    tip_length = self.current_tip_length
```
and then pass that in. that way we're insulated from someone forgetting to explicitly reset the tip length to 0

ahiuchingau (approved):
Sick nozzle manager, can't wait to see partial tip pickup in action

sfoster1 (approved):
Tested LPC and protocols on a flex with a reservoir and a plate, looks good to me!

PR #13583, opened 2023-09-18
feat(api,robot-server): check pipette offset consistency

sfoster1:
Add a basic instrument offset consistency check to drive UI indicating that an instrument may be miscalibrated and the user should run calibration.

This checks the consistency between the calibrated pipette offsets of the left and right pipettes. It detects a problem if
- There are two pipettes
- Both are calibrated
- Their calibration offsets differ by more than 1.5mm in any axis (this is tunable)

In other situations they succeed.

The outcome of this check shouldn't be treated as definitely meaning a calibration is bad; rather, it can serve as a hint that a user might want to retry calibration (so for instance if we display a banner on this it should be dismissable).

Closes RAUT-738

## Testing
- [x] Check that a robot with two pipettes with similar offsets passes
- [x] Check that a robot with two pipettes with different enough offsets fails
- [x] Check that a robot with one pipette passes
- [x] Take an inventory of robots and see if 1.5 is appropriate

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
@@ -1,5 +1,6 @@
 import typing
-from dataclasses import dataclass
+from typing_extensions import Literal, Final
+from dataclasses import dataclass, field
 from datetime import datetime
 
 from opentrons.config import feature_flags as ff
@@ -20,6 +21,18 @@ from opentrons.calibration_storage import (
 )
 from opentrons.hardware_control.types import OT3Mount
 
+PIPETTE_OFFSET_CONSISTENCY_LIMIT: Final = 1.5
+
+
+@dataclass
+class InconsistentPipetteOffsets:
+    kind: Literal["inconsistent-pipette-offset"]
+    offsets: typing.Dict[Mount, Point]
+    limit: float
+
+
+ReasonabilityCheckFailure = typing.Union[InconsistentPipetteOffsets]
+
 
 @dataclass
 class PipetteOffsetByPipetteMount:
@@ -31,6 +44,11 @@ class PipetteOffsetByPipetteMount:
     source: cal_top_types.SourceType
     status: cal_top_types.CalibrationStatus
     last_modified: typing.Optional[datetime] = None
+
+
+@dataclass
+class PipetteOffsetSummary(PipetteOffsetByPipetteMount):
+    reasonability_check_failures: typing.List[ReasonabilityCheckFailure] = field(default_factory=list)
 
 
 @dataclass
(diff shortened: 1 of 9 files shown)


api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
```diff
@@ -20,6 +21,18 @@
 )
 from opentrons.hardware_control.types import OT3Mount
 
+PIPETTE_OFFSET_CONSISTENCY_LIMIT: Final = 1.5
+
+
+@dataclass
+class InconsistentPipetteOffsets:
+    kind: Literal["inconsistent-pipette-offset"]
```
Laura-Danielle, line 29: why do we need `kind` if `InconsistentPipetteOffsets` is already its own class?

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
```diff
@@ -122,3 +142,25 @@ def save_gripper_calibration_offset(
...
+) -> typing.List[ReasonabilityCheckFailure]:
+    if (
+        not left_offset
+        or left_offset == Point(0, 0, 0)
+        or not right_offset
+        or right_offset == Point(0, 0, 0)
+    ):
+        return []
+    diff = left_offset - right_offset
+    if any(abs(d) > PIPETTE_OFFSET_CONSISTENCY_LIMIT for d in diff):
```
Laura-Danielle, line 158: Where does the limit come from?

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
```diff
@@ -122,3 +142,25 @@ def save_gripper_calibration_offset(
...
+        return [
+            InconsistentPipetteOffsets(
+                "inconsistent-pipette-offset",
+                {Mount.LEFT: left_offset, Mount.RIGHT: right_offset},
```
Laura-Danielle, line 162: Don't we already provide the offsets in the UI for users, could we not just pass in the difference and/or only show th e limit here?

robot-server/robot_server/instruments/instrument_models.py
```diff
@@ -26,6 +26,21 @@
...
+    """Pipette offsets are very different from each other.
+
+    This indicates that one of the pipettes (though we can't tell which one)
+    was likely calibrated with its calibration probe not fully attached, and
+    should be redone. However, it's possible that the pipettes are in fact
+    calibrated correctly and their offsets look strange, so this should be
+    taken as an advisory.
+    """
+
+    kind: Literal["inconsistent-pipette-offset"] = "inconsistent-pipette-offset"
```
Laura-Danielle, line 39: Same comments for this as per the dataclass in `instruments/ot3`

Laura-Danielle (approved):
None of my comments are blocking. In general it looks good to me, but unsure about some of the data we're providing.

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
sfoster1: It's to have future support for the thing that goes in the `reasonability_check_failures` being a discriminated union if we add more checks

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
sfoster1: discussions with hardware

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
sfoster1: yeah, but it feels kind of nice to have it in the error class (kind of like the details fields in the error codes)

robot-server/robot_server/instruments/instrument_models.py
sfoster1: same answer - adds support for a discriminated union of different kinds of checks with different payloads

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
Laura-Danielle: but we're already calling the dataclass `InconsistentPipetteOffsets` which feels restricted to me. Also what if the other checks require other data inputs that don't map to what `InconsistentPipetteOffsets` requires? I'm just not sure if we need to worry about future proofing that much at the moment.

sfoster1: Right, this dataclass is called `InconsistentPipetteOffsets`, but if you look below, the type that's bound into the `PipetteOffsetSummary` is actually `ReasonabilityCheckFailure` which is a union of (right now) one element. And having other checks have different data is exactly what the discriminated union is for - we'd definitely have to change stuff internally but we wouldn't have to change the shape of the server code or responses.

PR #13584, opened 2023-09-18
feat(robot-server): monitor door status

sfoster1:
Is it open? Is it closed? Only you can decide

## testing
- [x] put on robot
- [x] route should say open when door is open
- [x] route should say closed when door is closed

### DIFF
--- a/robot-server/robot_server/robot/control/router.py
+++ b/robot-server/robot_server/robot/control/router.py
@@ -13,6 +13,7 @@ from .models import EstopStatusModel, DoorStatusModel, DoorState
 from .estop_handler import EstopHandler
 from robot_server.hardware import get_estop_handler, get_hardware
 from opentrons.hardware_control import HardwareControlAPI
+from opentrons.config import feature_flags as ff
 
 if TYPE_CHECKING:
     from opentrons.hardware_control.ot3api import OT3API  # noqa: F401
@@ -66,6 +67,10 @@ async def put_acknowledge_estop_disengage(
     return await _get_estop_status_response(estop_handler)
 
 
+def get_door_switch_required() -> bool:
+    return ff.enable_door_safety_switch()
+
+
 @control_router.get(
     "/robot/door/status",
     summary="Get the status of the robot door.",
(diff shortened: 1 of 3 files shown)


robot-server/robot_server/robot/control/router.py
```diff
@@ -65,3 +64,21 @@ async def put_acknowledge_estop_disengage(
     """Transition from the `logically_engaged` status if applicable."""
     estop_handler.acknowledge_and_clear()
     return await _get_estop_status_response(estop_handler)
+
+
+@control_router.get(
+    "/robot/door/status",
+    summary="Get the status of the robot door.",
```
SyntaxColoring, line 76: Is there a more general term than "robot door" that we should use to account for the fact that on the OT-2, this is also wired to the top window?

I'm also fine with just going with "door," since that's what the advanced setting says.

robot-server/tests/robot/control/test_door_status.py
```diff
@@ -0,0 +1,16 @@
...
+from robot_server.robot.control.router import get_door_status
+
+
+@pytest.mark.parametrize("state", [DoorState.OPEN, DoorState.CLOSED])
+async def test_door_status(
+    state: DoorState, hardware_api: HardwareControlAPI, decoy: Decoy
+) -> None:
+    """Test the door status route."""
+    decoy.when(hardware_api.door_status).then_return(state)
+    assert (await get_door_status(hardware_api)).data.state == state.name.lower()
```
SyntaxColoring: `.data` -> `.content.data` to resolve type-checking errors.

```suggestion
    assert (
        await get_door_status(hardware_api)
    ).content.data.state == state.name.lower()
```

SyntaxColoring (approved):
Thanks! I tested this on an OT-2. This LGTM once the type-checking errors are resolved.

vegano1 (approved):
nice, this will work.
I think it would be good eventually to have eStop, door_state, and any other "safety" states come from one endpoint to cut down on chatter.

PR #13585, opened 2023-09-19
feat(api-client, app, react-api-client): Feat add robot door status check to desktop and odd

koji:
# Overview
add api-client and react-api-client to check door status via `/robot/door/status` endpoint
This PR needs https://github.com/Opentrons/opentrons/pull/13584 (Adding a new endpoint for the robot status)

For Flex, doorRequiredClosedForProtocol is always true, so in this PR the item is used as an alternative of isOT3.
```json
{
  "data": {
    "status": "closed",
    "doorRequiredClosedForProtocol": true
  }
}
```

[design]
Desktop
https://www.figma.com/file/0hYQ4lFJbAAI33jQMRt1Di/Release%3A-Desktop-June-Flex-launch?node-id=22521%3A106668&mode=dev

ODD
https://www.figma.com/file/AoTLAYuWawlaWItB1umOjr/Release%3A-Opentrons-Flex-Touchscreen?node-id=20416%3A122097&mode=dev

when a robot is missing modules/instruments or both -> not display Snackbar
when a robot is ready to run (instruments and modules are attached), but the door is open -> display Snackbar (7sec)

close RQA-1614

# Test Plan
Desktop app
1. Setup protocol
2. Open the robot door
Banner will show up in the header and the run start button is disabled

Touchscreen app
1. Setup protocol
2. Open the robot door
Snackbar will show up and the play button is disabled

# Changelog
- add getDoorStatus to `api-client` and `useDoorQuery` to `react-api-client` for the new endpoint, `/robot/door/status/`
- update ProtocolSetup components for Desktop app and ODD app
- update Snackbar type for passing duration

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/assets/localization/en/shared.json
+++ b/app/src/assets/localization/en/shared.json
@@ -6,7 +6,7 @@
   "browse": "browse",
   "cancel": "cancel",
   "clear_data": "clear data",
-  "close_robot_door": "Close the robot door before starting the run",
+  "close_robot_door": "Close the robot door before starting the run.",
   "close": "close",
   "computer_in_app_is_controlling_robot": "A computer with the Opentrons App is currently controlling this robot.",
   "confirm_placement": "Confirm placement",
--- a/app/src/assets/localization/en/shared.json
+++ b/app/src/assets/localization/en/shared.json
@@ -6,6 +6,7 @@
   "browse": "browse",
   "cancel": "cancel",
   "clear_data": "clear data",
+  "close_robot_door": "Close the robot door before starting the run",
   "close": "close",
   "computer_in_app_is_controlling_robot": "A computer with the Opentrons App is currently controlling this robot.",
   "confirm_placement": "Confirm placement",
--- a/api-client/src/robot/getDoorStatus.ts
+++ b/api-client/src/robot/getDoorStatus.ts
@@ -1,0 +1,9 @@
+import { GET, request } from '../request'
+
+import type { ResponsePromise } from '../request'
+import type { HostConfig } from '../types'
+import type { DoorStatus } from './types'
+
+export function getDoorStatus(config: HostConfig): ResponsePromise<DoorStatus> {
+  return request<DoorStatus>(GET, '/robot/door/status', null, config)
+}
--- a/api-client/src/robot/index.ts
+++ b/api-client/src/robot/index.ts
@@ -1,8 +1,10 @@
+export { getDoorStatus } from './getDoorStatus'
 export { getEstopStatus } from './getEstopStatus'
 export { acknowledgeEstopDisengage } from './acknowledgeEstopDisengage'
 export { getLights } from './getLights'
 export { setLights } from './setLights'
 export type {
+  DoorStatus,
   EstopPhysicalStatus,
   EstopState,
   EstopStatus,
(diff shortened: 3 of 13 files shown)


koji:
- [x] add test ProtocolSetup for ODD
- [x] clean up code

app/src/assets/localization/en/shared.json
```diff
@@ -6,6 +6,7 @@
   "browse": "browse",
   "cancel": "cancel",
   "clear_data": "clear data",
+  "close_robot_door": "Close the robot door before starting the run",
```
mjhuff:
```suggestion
  "close_robot_door": "Close the robot door before starting the run.",
```
Nit, seems to match the format of other full sentences in the file.

mjhuff (commented):
I'm having difficulty getting the snackbar to appear whenever a "newly uploaded protocol" toast is present. 9/10 times, the snackbar doesn't render. When it does render above the protocol added toast, it looks good!

If you can't figure out an exact reason why this is happening, I think we need something else on the screen to alert uses to close the door. Maybe a chip in the ProtocolRunHeader? 

Code is great! Just one small nit.

koji:
> I'm having difficulty getting the snackbar to appear whenever a "newly uploaded protocol" toast is present. 9/10 times, the snackbar doesn't render. When it does render above the protocol added toast, it looks good!
> 
> If you can't figure out an exact reason why this is happening, I think we need something else on the screen to alert uses to close the door. Maybe a chip in the ProtocolRunHeader?
> 
> Code is great! Just one small nit.

One question about the protocol you used.
Is the protocol was `ready to run`? (Right instruments and modules were attached?)

mjhuff:
> > I'm having difficulty getting the snackbar to appear whenever a "newly uploaded protocol" toast is present. 9/10 times, the snackbar doesn't render. When it does render above the protocol added toast, it looks good!
> > If you can't figure out an exact reason why this is happening, I think we need something else on the screen to alert uses to close the door. Maybe a chip in the ProtocolRunHeader?
> > Code is great! Just one small nit.
> 
> One question about the protocol you used. Is the protocol was `ready to run`? (Right instruments and modules were attached?)

Ah good point. One was not ready to run, the other was a magnetic block protocol. Here's the protocol I used:
[MagneticBlockTest2.py.zip](https://github.com/Opentrons/opentrons/files/12665544/MagneticBlockTest2.py.zip)

koji:
> > > I'm having difficulty getting the snackbar to appear whenever a "newly uploaded protocol" toast is present. 9/10 times, the snackbar doesn't render. When it does render above the protocol added toast, it looks good!
> > > If you can't figure out an exact reason why this is happening, I think we need something else on the screen to alert uses to close the door. Maybe a chip in the ProtocolRunHeader?
> > > Code is great! Just one small nit.
> > 
> > 
> > One question about the protocol you used. Is the protocol was `ready to run`? (Right instruments and modules were attached?)
> 
> Ah good point. One was not ready to run, the other was a magnetic block protocol. Here's the protocol I used: [MagneticBlockTest2.py.zip](https://github.com/Opentrons/opentrons/files/12665544/MagneticBlockTest2.py.zip)

Thank you.
I just tested my branch with the protocol you shared with me.
Seems that there is a bug on Magnetic module.

However, the snackbar shows up and disappear after 7 sec.
I guess that might be a timing issue. I set refetch duration 10sec.
As a result, sometimes the snackbar shows up immediately, but mostly it will take a few seconds.

mjhuff (approved):
I think the higher refetch interval will solve our current problems.

Thanks for doing all this! 🚪 🚀

PR #13587, opened 2023-09-19
refactor(protocol-designer): remove nest_96_wellplate_200ul_flat from…

jerader:
… adapter compatiblity

# Overview

upon smoke testing PD, I realized that this labware is not compatible with the universal flat adapter in the API and also a bug in the heater-shaker error creator where it didn't extend to support adapters.

# Test Plan

test that the universal flat adapter only has 1 labware available, the corning 384 well. So create a new protocol either OT-2 or Flex and add a heater-shaker. On the deck map, add the universal flat adapter. Add the labware on top and see only 1 is available.

add that labware and add a heater-shaker step to start shaking the heater-shaker. Now, try to add a transfer or mix into that labware on the heater-shaker, you should see a timeline error saying the heater-shaker is shaking.

# Changelog

- comment out the nest 96 well plate flat for now
- extend adapter support to the Heater-shaker collision util for if you're pipetting into a shaking heater-shaker

# Review requests

see test plan

# Risk assessment

low for these changes, but I'm hoping to cut a release candidate off of this branch that will hopefully be the release.

### DIFF
--- a/protocol-designer/src/utils/labwareModuleCompatibility.ts
+++ b/protocol-designer/src/utils/labwareModuleCompatibility.ts
@@ -96,13 +96,14 @@ export const COMPATIBLE_LABWARE_ALLOWLIST_FOR_ADAPTER: Record<
   ],
   [UNIVERSAL_FLAT_ADAPTER_LOADNAME]: [
     'opentrons/corning_384_wellplate_112ul_flat/2',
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Tested on PD sandbox, looks good to me!

PR #13588, opened 2023-09-19
fix(app, react-api-client): rely on subsystem status before instrument status

smb2268:
fix RQA-1585

# Overview
We can't rely on the `/instruments` endpoint response while a subsystem update is ongoing. Instead, we first need to poll for the presence of an update on a given subsystem.

# Test Plan
1. Attach an instrument that needs an update during a firmware update flow. Observe that when after the firmware update takes place, you don't see an "instrument not attached" screen before you see a success screen.
2. Update the firmware of an instrument via the ODD. On desktop, watch the instrument card for the corresponding instrument. The instrument card should display a subsystem update in progress until the update is complete, and then show the instrument card for an attached instrument.

# Changelog
1. In the firmware update modal, once the update is `done` refetch the instruments. If the instrument is `ok`, proceed to the results screen. If it is still not `ok`, wait another 10 seconds before proceeding.
2. In the gripper and pipette cards, continuously poll for a subsystem update. If an update is ongoing, rely on that information to show the "bad" instrument card and update banner. If it is not, rely on the `/instruments` endpoint response for what to show.

# Review requests
Run through the test plan

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/Devices/PipetteCard/index.tsx
+++ b/app/src/organisms/Devices/PipetteCard/index.tsx
@@ -60,7 +60,7 @@ const BANNER_LINK_CSS = css`
   cursor: pointer;
   margin-left: ${SPACING.spacing8};
 `
-const SUBSYSTEM_UPDATE_POLL_MS = 3000
+const SUBSYSTEM_UPDATE_POLL_MS = 5000
 
 export const PipetteCard = (props: PipetteCardProps): JSX.Element => {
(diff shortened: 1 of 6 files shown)


sfoster1 (approved):
Tested on a robot with a pipette with an erased application; the app follows along and immediately (well, 5s poll duration) shows that the update is complete even when it was started from the ODD.

PR #13594, opened 2023-09-19
fix(app): exit module calibration copy

jerader:
closes RQA-1653

# Overview

Fix copy from "calibration in progress" to "robot is in motion" to match what the copy says for exiting out of pipette flows

# Test Plan

Start module calibration and early exit. See that the in progress modal says the correct copy.

# Changelog

change i18n key to another key

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #13595, opened 2023-09-19
chore(build): re-add default ssh_key as robot_key

vegano1:
# Overview

- We should be using the default robot_key when ssh_key option is not provided during make push
- Fix left over `-O` option for `push-ot3-gravimetric` in hardware-testing Makefile

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
ah, yeah, good call

PR #13597, opened 2023-09-19
fix(app): mitigate ODD toast artifact generation

mjhuff:
Closes RAUT-729

# Overview

While utilizing hardware acceleration significantly reduced the rate of physical ODD toast artifact generation, it did not eliminate the underlying problem, which appears to be the high number of re-renders occurring on certain ODD pages. While there's plenty of render improvements and optimizations that need to be made for the app, "protocol addded" toasts can may be dismissed on any page (and thereby trigger the exit animation, which is what's causing the tearing), meaning that solving the animation problem requires optimizing re-renders across the whole app.

After consulting with Design, an acceptable and working solution appears to be only using a fade-out effect for exits. When screen tearing does occur, it likely occurs with an opacity of ~1%, which is hardly noticeable. The plan is to test the new animations on the next build and see if tearing is noticeable. After extensive testing on the app&ui flex, there was no noticeable tearing.  

### New Exit Animation

https://github.com/Opentrons/opentrons/assets/64858653/00b29216-1a7f-4106-97ab-645cdbc00476

# Test Plan

- Push protocols to the ODD. Observe if any tearing occurs, following the video above as a guide. 

# Changelog

- Changed Toast component exit animations.

# Review requests

- Test on a physical ODD. 

# Risk assessment
low

### DIFF
--- a/app/src/atoms/Toast/index.tsx
+++ b/app/src/atoms/Toast/index.tsx
@@ -163,11 +163,9 @@ export function Toast(props: ToastProps): JSX.Element {
 
     @keyframes slidedown {
       from {
-        transform: translate3d(0%, 0%, 0);
         filter: opacity(100%);
       }
       to {
-        transform: translate3d(0%, 50%, 0);
         filter: opacity(0);
       }
     }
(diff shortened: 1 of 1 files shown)


koji (approved):
This is great 🚀 
Thank you for fixing the issue!

PR #13600, opened 2023-09-19
fix(app): dismiss app update toast when update modal appears

mjhuff:
Closes RQA-1612
# Overview

Resolves a bug in which the app update toast remains visible when clicking on the app update banner.
 
The only real solution is to either prop drill or create context, which seems the lesser of two not great options. When more of our team is back in office, I really would like to push for [this library](https://github.com/eBay/nice-modal-react), which would greatly simplify the way we manage modals and prevent the need for such excessive context/prop drilling in the future plus aid in maintainability. Read the motivation section for an explanation of the benefits.
 
Note that the exit animation for toasts does not properly play when the update banner renders, although this doesn't seem to negatively affect UX. When toasts are refactored (a current to-do item), eatToast() should better account for returning a null Id after the toast exit animation duration, thereby allowing toasts to play while still enabling conditional removal of toasts based on toastId.
 
### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/8bad8f1f-384a-4be8-9018-2b4711f16908

### Fixed Behavior

https://github.com/Opentrons/opentrons/assets/64858653/a16355ca-dba0-45f2-a135-bd3628c1571a

# Test Plan
- Follow test steps as outlined in the video.
 

# Changelog
- Created new context and supporting hook for dismissing app updates.
- Placed hook in UpdateAppModal
- Added supporting testing

# Risk assessment

low

### DIFF
--- a/app/src/organisms/Alerts/AlertsModal.tsx
+++ b/app/src/organisms/Alerts/AlertsModal.tsx
@@ -1,0 +1,96 @@
+import * as React from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { useTranslation } from 'react-i18next'
+import head from 'lodash/head'
+
+import * as AppAlerts from '../../redux/alerts'
+import { getHasJustUpdated, toggleConfigValue } from '../../redux/config'
+import { getAvailableShellUpdate } from '../../redux/shell'
+import { SUCCESS_TOAST, WARNING_TOAST } from '../../atoms/Toast'
+import { useToaster } from '../ToasterOven'
+import { AnalyticsSettingsModal } from '../AnalyticsSettingsModal'
+import { UpdateAppModal } from '../UpdateAppModal'
+import { U2EDriverOutdatedAlert } from './U2EDriverOutdatedAlert'
+import { useRemoveActiveAppUpdateToast } from './useRemoveActiveAppUpdateToast.ts'
+
+import type { State, Dispatch } from '../../redux/types'
+import type { AlertId } from '../../redux/alerts/types'
+import type { MutableRefObject } from 'react'
+
+interface AlertsModalProps {
+  toastIdRef: MutableRefObject<string | null>
+}
+
+export function AlertsModal({ toastIdRef }: AlertsModalProps): JSX.Element {
+  const dispatch = useDispatch<Dispatch>()
+  const [showUpdateModal, setShowUpdateModal] = React.useState<boolean>(false)
+  const { t } = useTranslation('app_settings')
+  const { makeToast } = useToaster()
+  const { removeActiveAppUpdateToast } = useRemoveActiveAppUpdateToast()
+
+  // TODO(mc, 2020-05-07): move head logic to selector with alert priorities
+  const activeAlertId: AlertId | null = useSelector((state: State) => {
+    return head(AppAlerts.getActiveAlerts(state)) ?? null
+  })
+  const isAppUpdateAvailable = Boolean(useSelector(getAvailableShellUpdate))
+
+  const dismissDriverAlert = (remember?: boolean): void => {
+    if (activeAlertId != null) {
+      dispatch(AppAlerts.alertDismissed(activeAlertId, remember))
+    }
+  }
+  const isAppUpdateIgnored = useSelector((state: State) => {
+    return AppAlerts.getAlertIsPermanentlyIgnored(
+      state,
+      AppAlerts.ALERT_APP_UPDATE_AVAILABLE
+    )
+  })
+
+  const hasJustUpdated = useSelector(getHasJustUpdated)
+  const removeToast = !isAppUpdateAvailable || isAppUpdateIgnored
+  const createAppUpdateAvailableToast =
+    isAppUpdateAvailable && !isAppUpdateIgnored
+
+  // Only run this hook on app startup
+  React.useEffect(() => {
+    if (hasJustUpdated) {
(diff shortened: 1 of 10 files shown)


app/src/organisms/Alerts/AlertsModal.tsx
```diff
@@ -0,0 +1,96 @@
...
+      })
+      dispatch(toggleConfigValue('update.hasJustUpdated'))
+    }
+  }, [])
+
+  React.useEffect(() => {
+    if (createAppUpdateAvailableToast) {
+      toastIdRef.current = makeToast(
+        t('opentrons_app_update_available_variation'),
+        WARNING_TOAST,
```
koji, line 69: 👍 
we should use this instead of `warning` 🤣

app/src/organisms/Alerts/__tests__/Alerts.test.tsx
```diff
@@ -4,11 +4,13 @@ import { when } from 'jest-when'
...
 import * as AppAlerts from '../../../redux/alerts'
 import { getAvailableShellUpdate } from '../../../redux/shell'
+import { getHasJustUpdated } from '../../../redux/config'
 import { TOAST_ANIMATION_DURATION } from '../../../atoms/Toast'
-import { Alerts } from '..'
+import { AlertsModal } from '../AlertsModal'
 import { AnalyticsSettingsModal } from '../../AnalyticsSettingsModal'
 import { U2EDriverOutdatedAlert } from '../U2EDriverOutdatedAlert'
 import { UpdateAppModal } from '../../UpdateAppModal'
+import { useRemoveActiveAppUpdateToast } from '../useRemoveActiveAppUpdateToast.ts'
```
koji: we can remove `.ts`, can't we?

app/src/organisms/Alerts/__tests__/Alerts.test.tsx
```diff
@@ -24,21 +26,35 @@ jest.mock('../../UpdateAppModal', () => ({
 }))
 jest.mock('../../../redux/alerts/selectors')
 jest.mock('../../../redux/shell')
+jest.mock('../../../redux/config')
+jest.mock('../useRemoveActiveAppUpdateToast.ts')
```
koji: same as the above

app/src/organisms/UpdateAppModal/__tests__/UpdateAppModal.test.tsx
```diff
@@ -1,9 +1,11 @@
 import * as React from 'react'
 import { i18n } from '../../../i18n'
+import { when } from 'jest-when'
 import { fireEvent } from '@testing-library/react'
 import * as Shell from '../../../redux/shell'
 import { renderWithProviders } from '@opentrons/components'
 import { UpdateAppModal, UpdateAppModalProps } from '..'
+import { useRemoveActiveAppUpdateToast } from '../../Alerts/useRemoveActiveAppUpdateToast.ts'
```
koji:
```suggestion
import * as React from 'react'
import { when } from 'jest-when'
import { fireEvent } from '@testing-library/react'
import { renderWithProviders } from '@opentrons/components'
import { i18n } from '../../../i18n'
import * as Shell from '../../../redux/shell'
import { UpdateAppModal, UpdateAppModalProps } from '..'
import { useRemoveActiveAppUpdateToast } from '../../Alerts/useRemoveActiveAppUpdateToast.ts'
```

app/src/organisms/UpdateAppModal/index.tsx
```diff
@@ -97,6 +98,7 @@ export function UpdateAppModal(props: UpdateAppModalProps): JSX.Element {
   const releaseNotes = updateInfo?.releaseNotes
   const { t } = useTranslation('app_settings')
   const history = useHistory()
+  const { removeActiveAppUpdateToast } = useRemoveActiveAppUpdateToast()
 
   if (downloaded) setTimeout(() => dispatch(applyShellUpdate()), 5000)
```
koji: This is not related to this PR but could you create a const for 5000 ms?

app/src/organisms/Alerts/AlertsModal.tsx
mjhuff: Totally agree. It's on my to-do list to refactor the toast component entirely. Naming it `INFO_TOAST` would be more in-line with how we handle other yellow informing elements 👍

app/src/organisms/UpdateAppModal/__tests__/UpdateAppModal.test.tsx
mjhuff: Sorry to make you keep doing this...I'll eventually get this correct one of these days 😅

app/src/organisms/Alerts/__tests__/Alerts.test.tsx
mjhuff: I'm actually getting an error when I remove the .ts
Screenshot 2023-09-20 at 8 50 37 AM

mjhuff: Update: I found a workaround 👍

app/src/organisms/UpdateAppModal/__tests__/UpdateAppModal.test.tsx
koji: it's okay since I did this a lot last year when I joined the team 😄

koji (approved):
The changes look good to me!
Also the lib you mentioned is pretty interesting.

PR #13604, opened 2023-09-20
fix(shared-data): H/S labware offsets and TD calibration adapter measurements

sanni-t:
# Overview

We were mistakenly using the z- labware offset for labware (adapters) on H/S **on a Flex** as the 'unscrewed' position of the heater-shaker's labware bed. When you screw in an adapter, the labware bed dips down a little so this should be included in the labware offset. (so the z-offset on Flex changes from `z=18.95` to  `z=18.35`). This value was also a little off for the h/s on OT2 so corrected that as well.

Also found small mistake in temp module calibration block's well's measurements.

NOTE: This is not too critical for 7.0.0. Because these offsets/mistakes gets compensated for with module calibration, it doesn't affect actual labware positions on the Flex much.

# Test Plan

On Flex test that this PR:
- improves (reduces) the z-offset found during H/S module calibration
- improves the x, y offset found during temperature module calibration

On the OT2, test that this PR:
- reduces the z-offset for a labware on H/S during LPC

# Changelog

- updated the x, y coordinates of well A1 of the temp module calibration block
- updated the H/S `labwareOffset` z for both the OT-2 & Flex.

# Review requests

- Now that we have adapters separated out and have a stacking overlap measurement in the definitions, should we move this  'h/s bed dip' measurement as an overlap value for H/S adapters and adapter+labware combos? Or would that cause backwards compatibility issues with custom labware and json protocols?
- I'm updating the TD calibration block values in-place since this goes in the fast-follow release (and will be the official release for all, except a few we will be helping with installations). But let me know if this should get a version bump.

# Risk assessment

Very low. These are small improvements for positional accuracy. Doesn't touch any code.


SyntaxColoring:
Measurements look good so far on an OT-2.

Screenshot 2023-09-22 at 3 08 39 PM

shared-data/module/definitions/3/heaterShakerModuleV1.json
```diff
@@ -5,7 +5,7 @@
   "labwareOffset": {
     "x": -0.125,
     "y": 1.125,
-    "z": 68.275
+    "z": 68.525
```
SyntaxColoring, line 8: Can we add a note to the module schema saying what this physically corresponds to for the Heater-Shaker?

SyntaxColoring (commented):
Looks good, thanks!

> Now that we have adapters separated out and have a stacking overlap measurement in the definitions, should we move this 'h/s bed dip' measurement as an overlap value for H/S adapters and adapter+labware combos? Or would that cause backwards compatibility issues with custom labware and json protocols?

No strong opinion from me.

If the springy plate always gets depressed by the same distance, regardless of the adapter, I think it's reasonable to keep it as you have it.

However, let's definitely document this in the module schema.

> I'm updating the TD calibration block values in-place since this goes in the fast-follow release (and will be the official release for all, except a few we will be helping with installations). But let me know if this should get a version bump.

I actually think it's always going to be appropriate to change these calibration adapters in-place. It's not like they're "real labware" that people load in their protocols. They're an internal implementation detail of module calibration. They're only labware definitions because that let us share the geometry calculation code easier, I guess.

SyntaxColoring:
Also reminder to unify this with Jeremy's changes to add module offsets to slot/cutout A3

sanni-t:
This PR is planned to be merged with rest of the deck positioning work in https://opentrons.atlassian.net/browse/RSS-444. Do not merge until then.

PR #13607, opened 2023-09-20
fix(robot-server): Fix 500 error trying to start tip length calibration sessions

SyntaxColoring:
# Overview

Closes RSS-349.

# Test Plan

* [x] On an OT-2, follow the reproduction steps in RSS-349 and make sure you can complete tip length calibration now.

# Fix details

According to our labware definition JSON schema, these are the valid shapes of `gripperOffsets`:

* Totally omitted:
  
  ```
  <gripperOffsets omitted>
  ```
  
* Containing just `default`:
  
  ```json
  {
      "default": {
          "pickUpOffset": {"x": 1, "y": 2, "z": 3}
          "dropOffset": {"x": 1, "y": 2, "z": 3}
      }
  }
  ```
  
* Containing `default` and slot-specific offsets:
  
  ```json
  {
      "default": {
          "pickUpOffset": {"x": 1, "y": 2, "z": 3}
          "dropOffset": {"x": 1, "y": 2, "z": 3}
      }
      "A1": {
          "pickUpOffset": {"x": 1, "y": 2, "z": 3}
          "dropOffset": {"x": 1, "y": 2, "z": 3}
      }
  }
  ```

The problem was that this line in our Pydantic model did not match that.

https://github.com/Opentrons/opentrons/blob/76a6e54dbec634836b5d0a8e577d9631b3917533/shared-data/python/opentrons_shared_data/labware/labware_definition.py#L317

It would cause the server and local analysis engine to return labware definitions where `gripperOffsets` was normalized to `{}`. This was invalid according to our JSON schema, which required `gripperOffsets` to either contain at least a `"default"` key, or be totally omitted.

Those incorrect `{}`s are all over the database by now, so we unfortunately can't correct the Pydantic model to match the JSON schema. Instead, we need to loosen the JSON schema to accept what the Pydantic model has been emitting.

So, we're loosening the JSON schema to now accept these additional shapes:
  
* Containing nothing:
  
  ```json
  { }
  ```
  
* Containing just slot-specific offsets (this is an accidental byproduct of this implementation, but we believe it to be harmless)
  
  ```json
  {
      "A1": {
          "pickUpOffset": {"x": 1, "y": 2, "z": 3}
          "dropOffset": {"x": 1, "y": 2, "z": 3}
      }
  }
  ```  

# Review requests

Is this the best way to fix this?

Does it look like our Pydantic model now conforms to our JSON schema?

# Risk assessment

Low I guess? I tried not to touch any of the logic for deciding gripper offsets.


robot-server/tests/integration/utils.py
```diff
@@ -1,3 +1,9 @@
+"""Helper functions for use inside Tavern tests.
+
+https://tavern.readthedocs.io/en/latest/basics.html#calling-external-functions
```
sanni-t, line 3: This is pretty cool!

sanni-t (approved):
Looks great! I like the integration test idea too

PR #13608, opened 2023-09-20
feat(protocol-designer): add waste chute as option in move labware form

jerader:
closes RAUT-746 and RAUT-717

# Overview

This PR adds waste chute as an option in the move labware drop down menu and adds a dismissible hint to let users know that once they move their labware to the waste chute, they can't get it back.

# Test Plan

NOTE: we do not have the asset to display the waste chute in the deck map yet!! So it will appear empty even when you have the waste chute in your protocol!

- go to settings and make sure your FF is turned on (the deck modification ff). If you don't see it, you'll have to enable the dev ffs (type `enablePrereleaseMode()` in console)
- create a flex protocol and add a waste chute and a gripper (note: the waste chute asset is a gripper for now)
- go to deck setup and 1 labware on the deck (make sure it is a reservoir or well plate or tiprack)
- create a move labware step and select the labware without the gripper. see that the new location drop down DOES NOT include the waste chute. Select the gripper and see that the dropdown DOES include the waste chute. Click to move it to the waste chute
- save the form and observe the modal hint saying they are disgarding their labware. click on the checkbox to permanently dismiss it
- repeat the move labware step with the tiprack, moving it to the waste chute, and see that when you add it, the hint modal is not there.
- mess around with move labware and the new location dropdown, the waste chute should only be available if you are moving with the gripper, the labware is on deck, and the waste chute is present.

The hint modal:
Screen Shot 2023-09-20 at 3 26 16 PM

# Changelog

- add copy to the move labware header for moving to a waste chute
- add logic to the `getUnocuppiedLabwareLocationOptions` util for waste chute option
- add a hint for discarding of the waste chute

# Review requests

see test plan

# Risk assessment

low - shouldn't be able to access anything without the feature flag turned on.


protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -245,6 +221,30 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
           />
         )
       })}
+      <EquipmentOption
```
jerader, line 241: I decided that it makes sense to move the gripper to be after the modules so the sections are split up between with modules then additional equipment instead of gripper, modules, waste chute.

protocol-designer/src/localization/en/alert.json
```diff
@@ -84,6 +84,10 @@
 
       "body4a": "To batch edit Transfer or Mix advanced settings:",
       "body4b": "Only include Transfer or Mix steps in your selection."
+    },
+    "waste_chute_warning": {
+      "title": "Discarding of labware in the Waste Chute",
+      "body1": "You are about to move your labware to be discarded in the Waste Chute. Note that this action is irreversible unless you delete this move labware step. "
```
jerader: @ecormany, do you have any feedback on what the modal should say? This modal is for when you move a labware or tiprack to the waste chute. You are completely discarding it at this point so I created a modal to pop up to make sure users are aware of the consequences.

protocol-designer/src/localization/en/application.json
koji, line 1: nit
recommend to sort by VSCode sorting.

jerader: i dont know how to sort nested strings with VSCode sorting though, when i tried it un-nested the strings. Unless i'm doing it wrong 😆

protocol-designer/src/components/StepEditForm/fields/LabwareLocationField/index.tsx
```diff
@@ -20,18 +23,32 @@ export function LabwareLocationField(
...
+    unoccupiedLabwareLocationsOptions = unoccupiedLabwareLocationsOptions.filter(
+      option => option.value !== 'offDeck' && option.value !== WASTE_CHUTE_SLOT
+    )
+  } else if (props.useGripper || isLabwareOffDeck) {
     unoccupiedLabwareLocationsOptions = unoccupiedLabwareLocationsOptions.filter(
       option => option.value !== 'offDeck'
     )
+  } else if (!displayWasteChuteValue) {
+    unoccupiedLabwareLocationsOptions = unoccupiedLabwareLocationsOptions.filter(
+      option => option.name !== 'D3 - Waste Chute'
```
koji: The sandbox shows `Waste Chute in D3` in selectors whether I selected use gripper or not.

Tested the following code.
Or maybe using `includes('Waste Chute')

```suggestion
      option => option.name !== 'Waste Chute in D3'
```

Screenshot 2023-09-20 at 5 08 24 PM
Screenshot 2023-09-20 at 5 08 35 PM

protocol-designer/src/components/labware/utils.ts
```diff
@@ -33,3 +34,11 @@ export const wellFillFromWellContents = (
     },
     {}
   )
+
+export const getHasWasteChute = (
+  additionalEquipmentEntities: AdditionalEquipmentEntities
+): boolean => {
+  return Object.values(additionalEquipmentEntities).some(
+    aE => aE.location === WASTE_CHUTE_SLOT
```
koji: nit
maybe more specific name would be better

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -245,6 +221,30 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
...
+              'additionalEquipment',
+              without(values.additionalEquipment, 'gripper')
+            )
+          } else {
+            setFieldValue('additionalEquipment', [
+              ...values.additionalEquipment,
+              'gripper',
+            ])
+          }
+        }}
```
koji: I think creating a function for onClick would be better since it includes if-else statement if there is a specific reason to stick this way.

protocol-designer/src/components/StepEditForm/fields/LabwareLocationField/index.tsx
```diff
@@ -20,18 +23,32 @@ export function LabwareLocationField(
...
+  const additionalEquipmentEntities = useSelector(
+    getAdditionalEquipmentEntities
+  )
+  const hasWasteChute = getHasWasteChute(additionalEquipmentEntities)
   const isLabwareOffDeck =
     props.labware != null
       ? robotState?.labware[props.labware]?.slot === 'offDeck'
       : false
+  const displayWasteChuteValue =
+    props.useGripper && hasWasteChute && !isLabwareOffDeck
```
koji: If you don't mind, can you destruct props instead of using `props.`?

protocol-designer/src/localization/en/application.json
koji: oh, i see. never mind

protocol-designer/src/components/labware/utils.ts
```diff
@@ -33,3 +34,11 @@ export const wellFillFromWellContents = (
     },
     {}
   )
+
```
koji, line 37: recommend to add tests for this function.
1. true case
2. false case

koji (commented):
Followed test plan and most items worked as expected.
Only waste chute selector didn't work as expected. I left a comment for that.

Also I fixed the typo in description, `enablePrerelaseMode()` => `enablePrereleaseMode()` (I guess you utilized the first step from the previous your PR.)

protocol-designer/src/components/StepEditForm/fields/LabwareLocationField/index.tsx
jerader: oops thanks for noticing this! i forgot that i changed the string name

jerader:
> Also I fixed the typo in description, `enablePrerelaseMode()` => `enablePrereleaseMode()` (I guess you utilized the first step from the previous your PR.)

@koji oops thank you! Ya you should actually only have to enable it once? I think that it stays enabled unless you disable it 🤔 I haven't had to enable it in awhile.

koji:
> > Also I fixed the typo in description, `enablePrerelaseMode()` => `enablePrereleaseMode()` (I guess you utilized the first step from the previous your PR.)
> 
> @koji oops thank you! Ya you should actually only have to enable it once? I think that it stays enabled unless you disable it 🤔 I haven't had to enable it in awhile.

I guess I needed to run the command in console because I ran pd locally to check the waste chute condition.
For sandbox, as you said, we wouldn't need to run the command regularly.

protocol-designer/src/localization/en/alert.json
ecormany: Let's see. Based on what else we have in this file, I'm going to try to make it parallel in structure. And I think we tend to use the word "dispose" for putting stuff (blowout liquid, tips, now whole plates) in the trash. Also amping up the real-world consequences.
```suggestion
      "title": "Disposing labware",
      "body1": "Moving labware to the Waste Chute permanently discards it. You can't use this labware in later steps. During a protocol run, the labware will be dropped in the chute and become irretrievable."
```
feel free to bold **permanently discards** for extra "hey you!" ⚠️

protocol-designer/src/localization/en/application.json
jerader: I did use VSCode sorting for the clumps of un-nested strings though and tried to sort the rest manually. I'll keep in mind as i develop PD to try to alphabetize the i18n json files as I go.

protocol-designer/src/localization/en/alert.json
jerader: thank you!

protocol-designer/src/components/labware/__tests__/utils.test.ts
koji, line 1: 👍

koji (approved):
lgtm!
Tested the test plan again and the sandbox passed all items.

PR #13609, opened 2023-09-20
fix(robot-server): Say "Flex" instead of "OT-3 Standard" in upload error message

SyntaxColoring:
# Overview

Closes RQA-1673.

# Test Plan

See steps to reproduce in RQA-1673.

# Changelog

* Map the internal string "OT-3 Standard" to the user-facing string "Flex", and the internal string "OT-2 Standard" to the user-facing string "OT-2." Take care to use the proper indefinite articles: "a Flex" and "an OT-2," not "an Flex" or "a OT-2."
* Update our integration tests for this.
* Resolve an old todo comment in those integration tests about checking OT-2 -> Flex uploads, not just Flex -> OT-2 uploads.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/robot-server/robot_server/protocols/router.py
+++ b/robot-server/robot_server/protocols/router.py
@@ -224,9 +224,9 @@ async def create_protocol(
     if source.robot_type != robot_type:
         raise ProtocolRobotTypeMismatch(
             detail=(
-                f"This protocol is for {source.robot_type} robots."
+                f"This protocol is for {_user_facing_robot_type(source.robot_type)} robots."
                 f" It can't be analyzed or run on this robot,"
-                f" which is an {robot_type}."
+                f" which is {_user_facing_robot_type(robot_type, include_article=True)}."
             )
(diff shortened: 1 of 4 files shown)


SyntaxColoring:
I have a test timing out because I forgot an `ot3_only` mark, I think.

jbleon95 (approved):
We're flexin' here

TamarZanzouri (approved):
nice!

sanni-t (approved):
LGTM!
You could also test for old python protocols that use a metadata without the requirements dict to improve case coverage. But not critical for this PR.

SyntaxColoring:
> You could also test for old python protocols that use a metadata without the requirements dict to improve case coverage. But not critical for this PR.

Maybe! That should already be covered in the unit tests for `opentrons.protocols.parse`, but I'm open to rethinking our test strategy for this kind of thing.

PR #13610, opened 2023-09-20
fix(hardware-testing): Clear pipette accuracy function when using new low-volume configs

andySigler:
# Overview

Now that we are adjusting pipettes' configurations immediately before aspirating depending on what volume is aspirated, there is always the chance that we will be re-populating the pipette's accuracy function. This is bad for the increment test, because that test requires the accuracy function to be cleared (linear).

This PR fixes this by adding the ability to clear accuracy functions immediately after calling `configure_for_volume()` or `set_liquid_class()`

There are also some changes to Makefile to make it slightly easier to work with the SSH key requirements introduced in 7.0.0-alpha.9

# Test Plan

# Changelog

# Review requests

# Risk assessment


vegano1 (approved):
very nice, thank you

PR #13611, opened 2023-09-20
fix(api): do not block refresh positions

ahiuchingau:
# Overview
Just found out that if the gripper jaw is reporting a value greater than the max jaw value for whatever reason, we won't be able to home/move any axes. 

This is because we have this sanity check that raises an error whenever we try to set the jaw displacement. This is an issue for us now because whenever we home an axis, we first cache the motor positions for all available motors-- which also sets the jaw displacement value based on the jaw encoder position. 

To fix this, instead of raising an error whenever the encoder value is out-of-bound, we're just going to show a warning and set the displacement to the max value to not block any motion.


sfoster1 (approved):
noice!

PR #13614, opened 2023-09-20
feat(api): add settings option to reset ssh authorized keys

vegano1:
# Overview

We need to be able to reset the SSH authorized keys file, so we don't ship robots with internal public keys

# Test Plan

- [x] Make sure you can ssh to the Flex by using the `/server/ssh_keys/from_local` endpoint
- [x] Make sure that GET `/settings/reset/options` includes `authorizedKeys` key in the response.
- [x] Make sure that you CAN'T SSH when we POST to `/settings/reset` with the new `authorizedKeys` option

# Changelog

- Add authorizedKeys option to the list of robot settings that can be reset

# Review requests

# Risk assessment
Low

### DIFF
--- a/api/src/opentrons/config/reset.py
+++ b/api/src/opentrons/config/reset.py
@@ -15,6 +15,7 @@ from opentrons.calibration_storage import (
 
 
 DATA_BOOT_D = Path("/data/boot.d")
+AUTHORIZED_KEYS = Path(os.path.expanduser("~/.ssh/authorized_keys"))
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


api/src/opentrons/config/reset.py
```diff
@@ -39,6 +40,7 @@ class ResetOptionId(str, Enum):
     runs_history = "runsHistory"
     on_device_display = "onDeviceDisplay"
     module_calibration = "moduleCalibration"
+    authorized_keys = "authorizedKeys"
 
 
 _OT_2_RESET_OPTIONS = [
```
sfoster1, line 46: we should probably have it in the ot2 options too but that's ok for now

sfoster1 (approved):
Works for me, tested on a robot and couldn't ssh after resetting!

PR #13615, opened 2023-09-21
feat(step-generation): add wasteChute to invariantContext, add timeline warnings

jerader:
closes RAUT-749

# Overview

The bulk of this PR is adding a warning to `moveLabware` command creator so if you try to move a tiprack with tips or labware with liquid to the waste chute, a timeline warning appears. But in order to do that, I had to extend step-generation's `invariantContext` to include `additionalEquipmentEntities`. 

Also, I pulled out `WASTE_CHUTE_SLOT` from `protocol-designer` directory and moved it to the `shared-data` directory so I could use it in `step-generation`.

# Test Plan

- Make sure the correct feature flag for deck modification is turned on! (you might need to type `enablePrereleaseMode()` in console and restart in order to enable the dev feature flags)
- create a flex protocol and add a gripper and a waste chute
- in liquids, create a liquid 
- on the deck map, add a labware and add some liquid in it
- add a move labware step to move the labware to the waste chute with the gripper, hit save and see that the timeline warning shows up - it should be dismissible.
- add another move labware step and move the tiprack to the waste chute with the gripper, hit save and see that another timeline warning shows up - should also be dismissible.
- optional: repeat the steps with an empty tiprack/no liquids or even with a labware NOT to the waste chute, see that the timeline warnings don't show up.

# Changelog

- move `WASTE_CHUTE_SLOT` const to `shared-data` and fix all affected components
- fix some weirdness going on with i18n in a test (`AdditionalEquipmentRow.test.tsx`) - PD doesn't actually have a reacti18next dependency or something so we are gonna have to add that at some point and fix all of our PD unit tests. i have a ticket I made awhile back to do that but it is a huge housekeeping chore: https://opentrons.atlassian.net/browse/RAUT-542
- extend `invariantContext` type to include `AdditionalEquipmentEntities` and add to the selector. 
- create 2 timeline warnings
- add logic to `MoveLabware` so it knows when to raise those warnings
- add test coverage to `MoveLabware`
- created a few step-generation utils to `getTiprackHasTips` and `getLabwareHasLiquid` from the robot state.

# Review requests

see test plan

# Risk assessment

low


mjhuff (approved):
tested and works! still learning pd, but code seems 🔉

PR #13619, opened 2023-09-21
fix(app): fix tc module calibration issue

koji:
# Overview
Switch back the location from A1 to B1

close RQA-1682 RQA-1683

When we call module calibration wizard, we have to point out B1 for TC.
The frontend issue will be fixed when Brian is back.

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/getModulePrepCommands.ts
+++ b/app/src/organisms/Devices/getModulePrepCommands.ts
@@ -35,10 +35,6 @@ export function getModulePrepCommands(
   if (module.id != null && module.moduleType === THERMOCYCLER_MODULE_TYPE) {
     modulePrepCommands = [
       {
-        commandType: 'thermocycler/closeLid',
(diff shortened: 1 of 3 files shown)


jerader (approved):
lgtm, thanks for fixing! I confirmed that the correct sequence of commands get called on the network tab. 

One question, should this be targeting `chore_release-7.0.0`? I guess i am unsure if another alpha is getting cut?

mjhuff (approved):
Everything appears to be working now. Thank you!

koji:
> lgtm, thanks for fixing! I confirmed that the correct sequence of commands get called on the network tab.
> 
> One question, should this be targeting `chore_release-7.0.0`? I guess i am unsure if another alpha is getting cut?

ssh_key reset is not alpha14 so we will need to have another alpha.

PR #13623, opened 2023-09-21
feat(hardware-testing): save the generated gravimetric reports to the matching run_id directory

vegano1:
# Overview

The gravimetric script generates a lot of data in the form of CSV files, these files are written to the `testing_data/test_name` directory alongside other camera files. Since we generate new CSV files whenever the script is run, it can become unwieldy to manage hundreds of files over time. So instead of dumping all the files in the same location, let's create another directory with the run_id of the current run like `testing_data/test_name/run_id`.

The directory structure will now look like this for example
`testing_data/gravimetric-ot3-p1000-96/run-23-09-21-20-52-56`

Where now the last dir contains the generated reports like this one
`gravimetric-ot3-p1000-96_run-23-09-21-20-52-56_CSVReport-None-qc.csv`

# Test Plan

- [x] run simulated gravimetric tests and make sure the generated CSV files are dumped in the run_id dir
- [ ] run gravimetric test on a Flex and make sure everything still works
- [ ] run photometric test on flex and make sure everything still works

# Changelog

- added run_id arg to functions that create/append/insert to CSV files
- don't fail on appy/remove patches
- simplify hardware-testing labware scp

# Review requests

# Risk assessment
low

### DIFF
--- a/hardware-testing/Makefile
+++ b/hardware-testing/Makefile
@@ -289,19 +289,16 @@ sync-ot3: sync-sw-ot3 sync-fw-ot3
 push-ot3-gravimetric:
 	$(MAKE) apply-patches-gravimetric
 	-$(MAKE) sync-sw-ot3
+	scp $(ssh_helper_ot3) -r hardware_testing/labware root@$(host):/data/labware/v2/custom_definitions/custom_beta/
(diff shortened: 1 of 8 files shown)


andySigler (approved):
great!

PR #13625, opened 2023-09-21
fix(app): fix SetupModuleList's initial jsx element and condition

koji:
# Overview
The issue was that the implementation couldn't handle OT-2 case since the initial renderModuleStatus was the button. Update the initial renderModuleStatus and update the conditions.

close RQA-1549

after
Screenshot 2023-09-21 at 5 57 12 PM

Screenshot 2023-09-21 at 5 58 22 PM

# Test Plan
Protocol setup with OT-2/Flex

# Changelog
- update moduleStatus's initial value and conditions for it

# Review requests

# Risk assessment
low


sfoster1 (approved):
Tested on a flex, looks good to me!

PR #13628, opened 2023-09-22
fix(app): modify the text style in the software update modal

koji:
# Overview
add a couple of html tags to align with the design.

As I posted to app&ui channel, we will need to update the package and component to align with the design completely.
But right now, it's not a good time to do that. So this PR just try to align with the design as much as possible without a lot of efforts.

removed ballet points
Screenshot 2023-09-22 at 11 24 38 AM
Screenshot 2023-09-22 at 11 24 49 AM

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


mjhuff (approved):
lgtm! 🗒️

PR #13629, opened 2023-09-22
fix(api): New protocols print a JSON "run log" from opentrons_execute and opentrons.execute.execute() 

SyntaxColoring:
# Overview

Closes RSS-283.

When running PAPIv≥2.14 or JSONv≥6 protocols, `opentrons_execute` will now output a "run log" to stdout like this:

```
{"id": "056a8b70-0f46-43ca-b0eb-847671d31eff", "createdAt": "2023-09-27T18:33:48.154347+00:00", "commandType": "custom", "key": "3ca2f4eb308dc93ebab03777f50b0d85", "status": "running", "params": {"legacyCommandType": "command.COMMENT", "legacyCommandText": "Hello, world!"}, "result": null, "error": null, "startedAt": "2023-09-27T18:33:48.155772+00:00", "completedAt": null, "intent": null}
{"id": "f4d94b1d-6278-45ca-a525-8f811bbdcaa4", "createdAt": "2023-09-27T18:33:48.166984+00:00", "commandType": "loadPipette", "key": "72836212bea2e508afa3f07b3c296c6c", "status": "running", "params": {"pipetteName": "p20_multi_gen2", "mount": "left", "pipetteId": null}, "result": null, "error": null, "startedAt": "2023-09-27T18:33:48.168365+00:00", "completedAt": null, "intent": null}
```

And the same objects are exposed by `opentrons.execute.execute()`'s `emit_runlog` callback.

# Test Plan

* [x] Test this on a Flex by running `python3 -m opentrons.execute path_to_some_protocol.py` with a protocol that specifies `"apiLevel": "2.15"`.
* [x] Test this on an OT-2 by running `opentrons_execute path_to_some_protocol.py` with a protocol that specifies `"apiLevel": "2.14"` or above.

# Changelog

* Add a new helper to allow monitoring a `ProtocolEngine` for ongoing commands.

  This works by subscribing to every state update and seeing if the currently running command has changed since the last one. If it has, we emit a "command running" or "command no longer running" event. Conceptually, that event structure matches the "before" and "after" messages that [the Python Protocol API has long published](https://github.com/Opentrons/opentrons/blob/504588f84da5be81ff567b54b60d3fb9eb9425a9/api/src/opentrons/protocol_api/instrument_context.py#L424-L432).
* Wire that helper up to the `opentrons_execute` CLI and the `opentrons.execute.execute()` function.

  `opentrons.execute.execute()` has [a quirky interface](https://docs.opentrons.com/v2/new_protocol_api.html?highlight=execute#module-opentrons.execute:~:text=The%20format%20of%20the%20runlog%20entries%20is%20as%20follows%3A) that we need to preserve for now, so this takes some shoehorning. For this ticket, I'm not concerned about how the protocol's activity gets printed—just that it gets printed *somehow.* So I'm doing that by reporting each command as a "comment" where the message is the command's JSON text. There's obviously a lot of room for improvement here. RSS-320 tracks making these messages more nicely human-readable. And if there's a use case for it, we could also go the other way, officially exposing the command in a machine-readable way.

# Review requests

* Is this compatible with existing public-facing interfaces, in the sense that it won't outright break any code?
* See my inline comments.

# Risk assessment

Low. This is tricky to test, but it's also fairly nonintrusive.

### DIFF
--- a/api/src/opentrons/protocol_engine/command_monitor.py
+++ b/api/src/opentrons/protocol_engine/command_monitor.py
@@ -1,3 +1,6 @@
+"""Monitor the execution of commands in a `ProtocolEngine`."""
+
+
 from dataclasses import dataclass
 import typing
 import contextlib
@@ -8,44 +11,20 @@ from opentrons.protocol_engine import Command, ProtocolEngine
 
 @dataclass
 class RunningEvent:
+    """Emitted when a command starts running."""
+
     command: Command
 
 
 @dataclass
 class NoLongerRunningEvent:
+    """Emitted when a command stops running--either because it succeeded, or failed."""
+
     command: Command
 
 
 Event = typing.Union[RunningEvent, NoLongerRunningEvent]
-
-
 Callback = typing.Callable[[Event], None]
-
-
-class _Monitor:
-    def __init__(self, engine: ProtocolEngine, callback: Callback):
-        self._engine = engine
-        self._callback = callback
-        self._last_running_id: typing.Optional[str] = None
-
-    def handle_state_update(self) -> None:
-        running_id = self._engine.state_view.commands.get_running()
-        last_running_id = self._last_running_id
-
-        if running_id != last_running_id:
-            if last_running_id is not None:
-                self._callback(
-                    NoLongerRunningEvent(
-                        self._engine.state_view.commands.get(last_running_id)
-                    )
-                )
-
-            if running_id is not None:
-                self._callback(
-                    RunningEvent(self._engine.state_view.commands.get(running_id))
-                )
-
-        self._last_running_id = running_id
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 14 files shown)


api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -129,6 +131,36 @@ def state_view(self) -> StateView:
         """Get an interface to retrieve calculated state values."""
         return self._state_store
 
+    @property
+    def state_update_broker(self) -> ReadOnlyBroker[None]:
```
SyntaxColoring, line 135: Things I do not like about this:

* All of the footguns described in the docstring.
* It's callback-based, which can be spaghetti.

Things I like about this:

* Making the mechanism this generic—as opposed to something more specific like `wait_for_commands()`—lets us build more logic outside of `ProtocolEngine`. This is really nice for curbing `ProtocolEngine` complexity.
* This is is roughly compatible with my experiments in notifications for the HTTP API. I'm expecting various internal classes like `ProtocolEngine` and `AnalysisStore` to grow interfaces that let callers monitor them for changes in a coarse-grained way. Then, higher-level endpoints filter those notifications down and return only what the client actually cares about.
* Despite callbacks being ugly, they are suited to the actual requirements of RSS-238. The ultimate user-facing interface is callback-based. If we want async generators later, it's easier to build those atop callbacks than to do it the other way around.

jbleon95 (approved):
Nice, I think this is the most straightforward but sensible way to add this feature to PE-run CLI.

api/src/opentrons/protocol_engine/command_monitor.py
```diff
@@ -0,0 +1,65 @@
...
+    starts or stops a command.
+    """
+    # Subscribe to all state updates in protocol_engine.
+    # On every update, diff the new state against the last state and see if the currently
+    # running command has changed. If it has, emit the appropriate events.
+
+    last_running_id: typing.Optional[str] = None
+
+    def handle_state_update(_message_from_broker: None) -> None:
+        nonlocal last_running_id
```
TamarZanzouri, line 47: this is new to me :-)

TamarZanzouri (approved):
I agree with Jeremy. I like this approach. clean and straightforward. nice work!

api/src/opentrons/protocol_engine/command_monitor.py
SyntaxColoring: Yeah...Python. 😬

PR #13631, opened 2023-09-22
fix(hardware-testing): Explicitely add the commit hash to the test version string

andySigler:
# Overview

This PR explicitly places the commit has at the end of the test version string stored in the CSV file.

The test version string that is generated for QC tests uses the output of `git describe` and writes it to the CSV file. We rely on the commit hash in that output to make the data traceable to a specific software version.

A recent tag was generated that, for some reason, makes the output of `git describe` not include the commit hash.

# Test Plan

# Changelog

# Review requests

# Risk assessment


vegano1 (approved):
nice

PR #13632, opened 2023-09-22
fix(api): Allow right mount Z to home/retract when a 96-chan pipette is attached

ahiuchingau:
# Overview
When a 96-channel pipette is attached, the right mount doesn't really need to do anything other than to home/retract. Instead of disabling all movement for the right mount, we should allow it to home and retract. I've created a decorator that wraps these two functions in the hardware controller and temporarily updates the run current for the Z_R axis to allow it to move, and restores the old current once the motion is complete.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -165,7 +165,7 @@ AXES_IN_HOMING_ORDER: Tuple[Axis, Axis, Axis, Axis, Axis, Axis, Axis, Axis, Axis
 Wrapped = TypeVar("Wrapped", bound=Callable[..., Awaitable[Any]])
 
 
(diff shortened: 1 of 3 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -161,6 +162,24 @@
     Axis.Q,
 )
 
+Wrapped = TypeVar("Wrapped", bound=Callable[..., Awaitable[Any]])
+
+
+def adjust_high_throughput_z_current(func: Wrapped) -> Wrapped:
```
sfoster1: we should probably make this "private" (start the name with `_`)

sfoster1 (approved):
Looks good to me though I think we should make the decorator a private name

PR #13633, opened 2023-09-22
fix(app): add modal close logic for remote run deletion for LPC

smb2268:
# Overview
We need to add logic to the LPC modal that will cause it to close if the procedure is terminated remotely

# Test Plan
Run LPC on desktop, terminate the procedure on ODD, see that within 5-10 seconds the desktop modal closes.

# Changelog
1. Mimic logic we have in other maintenance run modals to track the `currentMaintenanceRun` and close the modal if that run is no longer present
2. Close the LPC modal even if run deletion fails, so that if the desktop modal is manually exited after the run is deleted from ODD but before we detect that, it won't get stuck in an infinite spinner

# Review requests
Test this out!

# Risk assessment
Low


sfoster1 (approved):
Tested on a flex, looks good to me!

PR #13636, opened 2023-09-22
fix(app): fix various physical odd stylistic issues

mjhuff:
Closes RQA-1513, RAUT-750, RQA-1686

# Overview

This PR fixes a number of stylistic bugs that occur only on the physical ODD. 

### Firmware Update Step

The firmware update step no longer shows in any flow if a firmware update isn't required. There were two main issues here: first, the total step counter was off - total steps would be +1 what actually occurred if a firmware update was not a part of the flow. Second, on the physical ODD, a brief flash of the firmware update screen occurs. 

This PR creates a new utility hook for conditionally filtering out steps that are unneeded. Currently, that is only firmware updates, but it will be easy to add/modify functionality in the future. 

Note that the hook hits the GET /instruments endpoint only once, since polling doesn't seem to make sense during a launched flow. 

**BEFORE**

https://github.com/Opentrons/opentrons/assets/64858653/2b692c66-62d7-4c1e-95a9-1df453a16174

**AFTER** 

https://github.com/Opentrons/opentrons/assets/64858653/66ff4b8c-c613-4083-a9a5-9bef21261fb1

### Magblock is Not Connected

The attached modules endpoint is compared against required protocol modules to determine if a module needs attachment. The problem is there is no way to detect if a magblock is attached. The easy and likely correct solution is just to assume it's attached when compared against a protocol - users are informed that it needs to be on the deck, anyway.

**BEFORE**
block not attached

**AFTER**
magblock attached

### Protocol Command Tearing & Snackbar Optimizations

Yep, you guessed it - other animations also tear on the ODD. This PR includes the ODD animation optimizations. Design helped modify the animations to be non-tearing.

I added the ODD optimizations to Snackbars as well...it can't hurt!

**BEFORE**

https://github.com/Opentrons/opentrons/assets/64858653/b0f6c8d9-1ca3-4449-81e0-7731d164c1cc

**AFTER** 

https://github.com/Opentrons/opentrons/assets/64858653/bcbac5c7-388b-4c0c-b1d1-ff837c23d329

# Test Plan

### Protocol Command Tearing
- Run any protocol and see that animations do not tear. Here's an easy one that just needs a gripper attached. [GripperMoveTesting.py.zip](https://github.com/Opentrons/opentrons/files/12704376/GripperMoveTesting.py.zip)

### Magblock is Not Connected
- Navigate to the ProtocolDetails screen for any protocol that includes a mag block. NOTE: A dev environment may not work here - the test-flex.json file needs updating (it's on my TODO). If behavior isn't correct, remove the magdeck from the test-flex.json.
- Here's an easy protocol to test: 
[MagneticBlockTest.py.zip](https://github.com/Opentrons/opentrons/files/12704436/MagneticBlockTest.py.zip)

### Firmware Update Step
- Test on the physical ODD (it's the only way to see the screen blip) and go through any attachment/module cal flow. There shouldn't be a firmware update step included in the total or a screen blip if the pipette/gripper doesn't need a firmware update.

# Changelog

- Changed command protocol animations and added ODD optimizations.
- Tweaked a Toast optimzation.
- Create a new hook for filtering wizard steps.
- Added supportive hook tests.
- Added a special case filter for the magblock connected property.
- Pre-emptive snackbar animation optimization for ODD.

# Review requests

Firmware updates and animations need testing on the physical ODD. See notes on magdeck testing if on the dev environment.

# Risk assessment
low


app/src/atoms/Snackbar/index.tsx
```diff
@@ -22,10 +22,17 @@ export interface SnackbarProps extends StyleProps {
 
 const SNACKBAR_ANIMATION_DURATION = 500
 
+const ODD_ANIMATION_OPTIMIZATIONS = `
+  backface-visibility: hidden;
+  perspective: 1000;
+  will-change: opacity;
```
koji, line 28: 👍

app/src/pages/Protocols/hooks/index.ts
```diff
@@ -14,6 +14,7 @@ import type {
 } from '@opentrons/shared-data'
 import type { LabwareSetupItem } from '../utils'
 import { getProtocolUsesGripper } from '../../../organisms/ProtocolSetupInstruments/utils'
+import { AttachedModule } from '@opentrons/api-client'
```
koji: i guess 
```suggestion
import type { AttachedModule } from '@opentrons/api-client'
```

koji: Could you move up the above line, `import { getProtocolUsesGripper } from '../../../organisms/ProtocolSetupInstruments/utils'` seems that the import isn't type.

koji (commented):
lgtm
tested this branch with a Flex and it worked as expected.

I didn't work on wizard part that much so getting another review would be good!

mjhuff:
Reopening under chore_release-0.7.1

PR #13637, opened 2023-09-22
feat(protocol-designer, step-generation, shared-data): 96-channel tip adapter + well selection support

jerader:
closes RAUT-678 RAUT-679 RAUT-697 RAUT-755

this pr is now unblocked ✨ 

# Overview

This adds support for 2 main things for PD 96-channel:
1. 96-channel tiprack adapter support
2. 96-channel well selection support

Sorry this ended up doing a lot in 1 PR! 

# Test Plan

For tiprack adapter:

1. turn on the 96-channel dev feature flag
2. create a Flex protocol and add the 96-channel
4. go to deck setup. Observe that the tiprack on the deck has an adapter underneath it (the adapter is just a white svg rectangle thing underneath the tiprack)
5. try duplicating the tiprack, the duplicate should NOT have an adapter underneath
6. select an empty deck slot and add another tiprack, it should NOT render the adapter underneath
7. select an empty deck slot and add the 96-channel adapter (under the adapter section), it should render only the adapter. Check that the only compatible labware is the tiprack.
8. add another adapter with no tiprack, try to drag and drop other labwares onto it, make sure only the tiprack is supported to drag and drop onto the adapter but that the tiprack also can be directly on the deck.
9. add a move labware command, make sure only the tiprack can move on top of the adapter .

For 96-channel well selection:

1. create a Flex protocol with the 96-channel and a tiprack and add the trash bin (the waste chute isn't fully wired up for dispensing/dropping tip into so a trash bin is required for now). Go the deck map and add a reservoir, a 96 well plate, an irregular plate, and the 384 well plate.
2. try to mix/transfer from out of each plate. See that they all work except for the irregular plate. See that the correct amount of wells are highlighted (all wells except for the 384 well plate which should highlight every other well for up to 96 wells.)
4. now create a new protocol with a single and 8-channel pipette. See that transferring/mixing works as usual and highlights either 1 or 8 wells.

# Changelog

- `createContainer` thunk to include an optional `adapterUnderLabwareURI` or `adapterId` in order to allow for duplicating the labware with the adapter underneath it. 
- extend the deck setup components to have knowledge of if the 96-channel is in the protocol
- extend `getWellNamePerMultiTip` and affecting utils to have knowledge of if the pipette has 96-channels 
- add a few helper functions to shared-data (moved `orderedWells` out of step-generation and into shared-data and created a `get96Channel384WellPlateWells` util that grabs the arrays of wells the 96-channel can aspirate from in a 384 well plate)
- fix tests
- create a `missingAdapter()` error creator to be plugged in for a follow up PR

# Review requests

see test plan

# Risk assessment

low-ish, this changes a bunch of logic

### DIFF
--- a/labware-library/src/labware-creator/utils/determineMultiChannelSupport.ts
+++ b/labware-library/src/labware-creator/utils/determineMultiChannelSupport.ts
@@ -16,7 +16,7 @@ export const determineMultiChannelSupport = (
   // allow multichannel pipette options only if
   // all 8 channels fit into the first column correctly
   const multiChannelTipsFirstColumn =
-    def !== null ? getWellNamePerMultiTip(def, 'A1') : null
+    def !== null ? getWellNamePerMultiTip(def, 'A1', 8) : null
 
   const allowMultiChannel =
     multiChannelTipsFirstColumn !== null &&
--- a/labware-library/src/labware-creator/utils/determineMultiChannelSupport.ts
+++ b/labware-library/src/labware-creator/utils/determineMultiChannelSupport.ts
@@ -15,6 +15,8 @@ export const determineMultiChannelSupport = (
 
   // allow multichannel pipette options only if
   // all 8 channels fit into the first column correctly
+  // TODO(Jr, 9/25/23): support 96-channel in labware creator then plug in
+  // channels below in getWellNamePerMultiTip
   const multiChannelTipsFirstColumn =
     def !== null ? getWellNamePerMultiTip(def, 'A1', 8) : null
 
--- a/protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
+++ b/protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
@@ -40,6 +40,7 @@ interface DNDP {
 interface OP {
   slot: DeckSlotDefinition & { id: DeckSlot } // NOTE: Ian 2019-10-22 make slot `id` more restrictive when used in PD
   moduleType: ModuleType | null
+  has96Channel: boolean
   selectedTerminalItemId?: TerminalItemId | null
   handleDragHover?: () => unknown
 }
@@ -68,6 +69,7 @@ export const SlotControlsComponent = (
     draggedItem,
     itemType,
     customLabwareDefs,
+    has96Channel,
   } = props
   if (
     selectedTerminalItemId !== START_TERMINAL_ITEM_ID ||
@@ -82,11 +84,12 @@ export const SlotControlsComponent = (
 
   let slotBlocked: string | null = null
   if (
-    isOver &&
-    moduleType != null &&
-    draggedDef != null &&
-    !getLabwareIsCompatible(draggedDef, moduleType) &&
-    !isCustomLabware
+    (isOver &&
+      moduleType != null &&
+      draggedDef != null &&
+      !getLabwareIsCompatible(draggedDef, moduleType) &&
+      !isCustomLabware) ||
+    (has96Channel && draggedDef?.parameters.isTiprack)
   ) {
     slotBlocked = 'Labware incompatible with this module'
   }
--- a/protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
+++ b/protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
@@ -1,17 +1,16 @@
 import assert from 'assert'
 import * as React from 'react'
-import { Icon, RobotCoordsForeignDiv } from '@opentrons/components'
-import { DropTarget, DropTargetConnector, DropTargetMonitor } from 'react-dnd'
-import cx from 'classnames'
 import { connect } from 'react-redux'
 import noop from 'lodash/noop'
+import { DropTarget, DropTargetConnector, DropTargetMonitor } from 'react-dnd'
+import cx from 'classnames'
+import { Icon, RobotCoordsForeignDiv } from '@opentrons/components'
 import { i18n } from '../../../localization'
 import { DND_TYPES } from '../../../constants'
 import {
   getLabwareIsCompatible,
   getLabwareIsCustom,
 } from '../../../utils/labwareModuleCompatibility'
-import { BlockedSlot } from './BlockedSlot'
 import {
   moveDeckItem,
   openAddLabwareModal,
@@ -21,13 +20,14 @@ import {
   selectors as labwareDefSelectors,
 } from '../../../labware-defs'
 import { START_TERMINAL_ITEM_ID, TerminalItemId } from '../../../steplist'
+import { BlockedSlot } from './BlockedSlot'
 
-import { BaseState, DeckSlot, ThunkDispatch } from '../../../types'
-import { LabwareOnDeck } from '../../../step-forms'
-import {
+import type {
   DeckSlot as DeckSlotDefinition,
   ModuleType,
 } from '@opentrons/shared-data'
+import type { BaseState, DeckSlot, ThunkDispatch } from '../../../types'
+import type { LabwareOnDeck } from '../../../step-forms'
 import styles from './LabwareOverlays.css'
 
 interface DNDP {
@@ -40,7 +40,6 @@ interface DNDP {
 interface OP {
   slot: DeckSlotDefinition & { id: DeckSlot } // NOTE: Ian 2019-10-22 make slot `id` more restrictive when used in PD
   moduleType: ModuleType | null
-  has96Channel: boolean
   selectedTerminalItemId?: TerminalItemId | null
   handleDragHover?: () => unknown
 }
@@ -69,7 +68,6 @@ export const SlotControlsComponent = (
     draggedItem,
     itemType,
     customLabwareDefs,
-    has96Channel,
   } = props
   if (
     selectedTerminalItemId !== START_TERMINAL_ITEM_ID ||
@@ -84,12 +82,11 @@ export const SlotControlsComponent = (
 
   let slotBlocked: string | null = null
   if (
-    (isOver &&
-      moduleType != null &&
-      draggedDef != null &&
-      !getLabwareIsCompatible(draggedDef, moduleType) &&
-      !isCustomLabware) ||
-    (has96Channel && draggedDef?.parameters.isTiprack)
+    isOver &&
+    moduleType != null &&
+    draggedDef != null &&
+    !getLabwareIsCompatible(draggedDef, moduleType) &&
+    !isCustomLabware
   ) {
     slotBlocked = 'Labware incompatible with this module'
   }
--- a/protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
+++ b/protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
@@ -193,7 +193,7 @@ export const LabwareSelectionModal = (props: Props): JSX.Element | null => {
       const smallYDimension = labwareDef.dimensions.yDimension < 85.48
       const irregularSize = smallXDimension && smallYDimension
       const adapter = labwareDef.metadata.displayCategory === 'adapter'
-      const adapter96Channel =
+      const isAdapter96Channel =
         labwareDef.parameters.loadName === ADAPTER_96_CHANNEL
       return (
         (filterRecommended &&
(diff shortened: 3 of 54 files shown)


labware-library/src/labware-creator/utils/determineMultiChannelSupport.ts
```diff
@@ -16,7 +16,7 @@ export const determineMultiChannelSupport = (
   // allow multichannel pipette options only if
   // all 8 channels fit into the first column correctly
   const multiChannelTipsFirstColumn =
-    def !== null ? getWellNamePerMultiTip(def, 'A1') : null
+    def !== null ? getWellNamePerMultiTip(def, 'A1', 8) : null
```
jerader, line 21: since this is for labware creator, i defaulted to only having the 8-channel since it doesn't support creating labware for a 96-channel yet.

step-generation/src/errorCreators.ts
```diff
@@ -12,6 +12,13 @@ export function insufficientTips(): CommandCreatorError {
   }
 }
 
+export function missingAdapter(): CommandCreatorError {
+  return {
+    type: 'MISSING_96_CHANNEL_TIPRACK_ADAPTER',
+    message: 'A 96-channel cannot pick up tips fully without an adapter',
+  }
+}
+
```
jerader, line 21: NOTE: this will be wired up to `replaceTip` in a follow up PR. Since we need to have knowledge of the current `tiprackId`, I would need to grab that data from higher up and extend the `replaceTip` args to include it. https://opentrons.atlassian.net/browse/RAUT-798

protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
```diff
@@ -82,11 +84,12 @@ export const SlotControlsComponent = (
...
-    moduleType != null &&
-    draggedDef != null &&
-    !getLabwareIsCompatible(draggedDef, moduleType) &&
-    !isCustomLabware
+    (isOver &&
+      moduleType != null &&
+      draggedDef != null &&
+      !getLabwareIsCompatible(draggedDef, moduleType) &&
+      !isCustomLabware) ||
+    (has96Channel && draggedDef?.parameters.isTiprack)
```
shlokamin: this code block looks like it's trying to prevent users from moving around any tiprack if this protocol uses a 96 channel pipette. this is probably me not understanding the product spec properly, but how come we want to do that?

also, it looks like the `Labware incompatible with this module` text shows up when you try to drag a tiprack, but you can still drop it anywhere

protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
```diff
@@ -190,7 +193,8 @@ export const LabwareSelectionModal = (props: Props): JSX.Element | null => {
       const smallYDimension = labwareDef.dimensions.yDimension < 85.48
       const irregularSize = smallXDimension && smallYDimension
       const adapter = labwareDef.metadata.displayCategory === 'adapter'
-
+      const adapter96Channel =
```
shlokamin: can this variable be prefixed with an `is` since its a boolean?

protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
jerader: oh, i think this is left over from when I had restrictions where you had to always have an adapter under the tiprack. It should be that a tiprack can be placed anywhere directly on the deck or on the adapter. Thanks for calling it out!

protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
```diff
@@ -420,37 +424,40 @@ export const LabwareSelectionModal = (props: Props): JSX.Element | null => {
             })
           ) : (
             <PDTitledList
+              data-testid="adapterCompatibleLabware"
```
shlokamin: we unfortunately arent super consistent about how we format our data test ids, but i think what we've landed on is `data-testid={ComponentName_description}` so like `data-testid={Slideout_icon_close}`

protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
```diff
@@ -420,37 +424,40 @@ export const LabwareSelectionModal = (props: Props): JSX.Element | null => {
...
-                  const Uris = Object.keys(latestDefs)
-                  const labwareDefUri = Uris.find(
-                    defUri => defUri === adapterDefUri
-                  )
-                  const labwareDef = labwareDefUri
-                    ? latestDefs[labwareDefUri]
-                    : null
+              {getLabwareCompatibleWithAdapter(
+                has96Channel ? permittedTipracks : [],
+                adapterLoadName
```
shlokamin: this is a little hard for me to follow — why are we special casing the input to `getLabwareCompatibleWithAdapter`? it looks like that function already checks if the adapter load name is `ADAPTER_96_CHANNEL`.

protocol-designer/src/components/LabwareSelectionModal/__tests__/LabwareSelectionModal.test.tsx
```diff
@@ -51,4 +61,17 @@ describe('LabwareSelectionModal', () => {
       MAX_LABWARE_HEIGHT_EAST_WEST_HEATER_SHAKER_MM
     )
   })
+  it('should display only permitted tipracks if has', () => {
```
shlokamin: looks like this test name is incomplete

shlokamin (commented):
gonna keep throwing these comments in batches so you can see them roll in!

protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
jerader: the function needs to have knowledge of the permitted tipracks only if it is a 96-channel. but i'll make refactor it to make it easier to read.

protocol-designer/src/components/LabwareSelectionModal/__tests__/LabwareSelectionModal.test.tsx
jerader: lol oops

protocol-designer/src/labware-defs/actions.ts
```diff
@@ -241,11 +245,21 @@ const _createCustomLabwareDef: (
...
-  event: React.SyntheticEvent<HTMLInputElement>
-) => ThunkAction<any> = _createCustomLabwareDef(false)
+  event: React.SyntheticEvent<HTMLInputElement>,
+  has96Channel: boolean
+) => (event: React.SyntheticEvent<HTMLInputElement>) => ThunkAction<any> = (
+  event,
+  has96Channel
+) => _createCustomLabwareDef(false, has96Channel)
+
 export const createCustomTiprackDef: (
```
shlokamin: having `has96Channel` params passed into generic functions like `createCustomLabwareDef` and `createCustomTiprackDef` make my spidey sense tingle. it looks like this parameter just keeps getting passed on until its used inside of `getIsOverwriteMismatched`. since both `createCustomLabwareDef` and `createCustomTiprackDef` are redux thunks, i think it makes more sense for them to determine if the protocol in question has a 96 channel pipette since they have access to redux state.

@b-cooper what do you think?

jerader: I think passing props down is a lot easier since some of these utilities are heavily nested. But maybe instead of calling it `has96Channel`, it can be more generic like `channels: number` or `numberOfChannels: number`. When we add partial tip support, the utils are going to have to have some knowledge of the partial tip configuration and having knowledge of the number of channels I think would work for all those configurations including whether or not there are 96 channels used or 8 channels used for the 8-channel.

shlokamin: from irl code review: we'll grab pipette entities from the getState redux function inside of the thunk so we dont have to update all calls to these helpers

protocol-designer/src/top-selectors/labware-locations/index.ts
```diff
@@ -145,13 +147,18 @@ export const getUnocuppiedLabwareLocationOptions: Selector<
...
+                    ? `${adapterDisplayName} on top of ${
+                        modIdWithAdapter != null
+                          ? getModuleDisplayName(
+                              moduleEntities[modIdWithAdapter].model
+                            )
+                          : 'unknown module'
+                      } in slot ${modSlot ?? 'unknown slot'}`
+                    : `${adapterDisplayName} on slot ${
+                        adapterSlot ?? 'unknown'
+                      }`,
```
shlokamin: these nested ternaries make me sad

protocol-designer/src/top-selectors/substep-highlight.ts
```diff
@@ -12,15 +12,23 @@ import { PipetteEntity, LabwareEntity } from '@opentrons/step-generation'
...
   pipetteEntity: PipetteEntity,
   labwareEntity: LabwareEntity,
   wells: string[]
 ): string[] {
   // `wells` is all the wells that pipette's channel 1 interacts with.
-  if (pipetteEntity.spec.channels === 8) {
+  if (pipetteEntity.spec.channels === 8 || pipetteEntity.spec.channels === 96) {
     return wells.reduce((acc: string[], well: string) => {
-      const setOfWellsForMulti = getWellNamePerMultiTip(labwareEntity.def, well)
+      const channels = pipetteEntity.spec.channels as MultiChannels
```
shlokamin: if we destructure the `channels` value you wont have to as cast this

protocol-designer/src/top-selectors/substep-highlight.ts
```diff
@@ -186,7 +195,8 @@ function _getSelectedWellsForSubstep(
       if (activeTips && activeTips.labwareId === labwareId) {
         const multiTipWellSet = getWellSetForMultichannel(
           invariantContext.labwareEntities[labwareId].def,
-          activeTips.wellName
+          activeTips.wellName,
+          8
```
shlokamin, line 197: 8 shouldnt be hardcoded right? it should be taken from the pipette entity

protocol-designer/src/utils/index.ts
```diff
@@ -127,3 +128,7 @@ export const getStagingAreaSlots = (
   //  we can assume that the location is always a string
   return stagingAreas.map(area => area.location as string)
 }
+
+export const getHas96Channel = (pipettes: PipetteEntities): boolean => {
+  return Object.values(pipettes).some(pip => pip.name === 'p1000_96')
```
shlokamin: lets read the channels property from pipette entities instead, thats safer

protocol-designer/src/utils/labwareModuleCompatibility.ts
```diff
@@ -118,15 +119,23 @@ export const COMPATIBLE_LABWARE_ALLOWLIST_FOR_ADAPTER: Record<
...
 export const getLabwareCompatibleWithAdapter = (
   adapterLoadName?: string
-): string[] =>
-  adapterLoadName != null
+): string[] => {
+  return adapterLoadName != null
     ? COMPATIBLE_LABWARE_ALLOWLIST_FOR_ADAPTER[adapterLoadName]
     : []
-
+}
```
shlokamin: implicit return?

shared-data/js/helpers/__tests__/wellSets.test.ts
```diff
@@ -3,19 +3,135 @@ import fixture_12_trough from '../../../labware/fixtures/2/fixture_12_trough.jso
...
+  'A9',
+  'A10',
+  'A11',
+  'A12',
+]
+
+const wellsFor96WellPlate = [
+  'A1',
+  'B1',
+  'C1',
```
shlokamin, line 40: we should probably have a function do this in a nested for loop or something

shared-data/js/helpers/get96Channel384WellPlateWells.ts
```diff
@@ -0,0 +1,42 @@
+export function get96Channel384WellPlateWells(
```
shlokamin, line 3: lets add a todo inside of `getWellNamePerMultiTip` to extend it to do similar math as it does for the 8 channel but also for the 96 so we dont have to special case and so we can support custom labware

shared-data/js/helpers/wellSets.ts
```diff
@@ -11,24 +11,31 @@
...
 import uniq from 'lodash/uniq'
 
 import { getWellNamePerMultiTip } from './getWellNamePerMultiTip'
+import { get96Channel384WellPlateWells, getLabwareDefURI, orderWells } from '.'
 import type { LabwareDefinition2, PipetteNameSpecs } from '../types'
 
 type WellSetByPrimaryWell = string[][]
 
 // Compute all well sets for a labware def (non-memoized)
 function _getAllWellSetsForLabware(
```
shlokamin: i dont think this fn needs to be changed if `getWellNamePerMultiTip` accounts for the 96 channel pipette

shlokamin (approved):
code reviewed together, approved pending the comments i left

PR #13638, opened 2023-09-22
fix(app): fix desktop door status check condition for run button

koji:
# Overview
The solution for this issue is to match the disabled status for run button and the condition that I set for banner.

close RQA-1702

# Test Plan
Desktop app

There are a few cases need to check
### case1
1. Finish ProtocolSetup
2. Open the door
Run button is disabled & the banner is displayed

### case 2
1. Finish running a protocol
2. Open the door
Re-run button is enabled
3. Click re-run button
Run button is disabled since the door is still open

### case 3
1. Run a protocol
2. Cancel a run
3. Open the door
Re-run button is enabled
4. Click re-run button
Run button is disabled since the door is still open

# Changelog
- match the run button disable condition and the condition for displaying banner

# Review requests
The cases that we should check for open door status are enough?
Is there any other case that prevents users' operations?

# Risk assessment
low


mjhuff (approved):
Tested and works as intended. I can't think of any other places we need to add door check logic...I spent some time trying to break the door notifications, and it seems to hold up. I think we're good! Thanks for fixing this! 🚀

PR #13645, opened 2023-09-25
fix(app):  Update conditional calibration banner rendering

mjhuff:
Closes RQA-1708, RQA-1393

# Overview

This PR fixes two bugs related to calibration banner rendering.

### Flex Pipette/Gripper banners not rendering after clearing offsets

Clearing pipette/gripper calibration now sets the calibrated offsets to zero instead of clearing the calibrated offsets property. Checking the date last calibrated property is the easiest and most accurate way to determine if a calibration is necessary.
 
**FIXED**

fixed cal banners
 
### OT-2 Calibration banners still rendering when robot offline
 
Adds a conditional check to make sure that the robot is viewable before rendering the banner. "isRobotViewable" is the same conditional check used for other elements rendered on the DeviceDetails page.

**BEFORE**

Screenshot 2023-08-24 at 10 29 54 AM

**AFTER**

ot-2 cal banner fix

# Test Plan
- For the Flex, clear pipette/gripper calibration settings via Advanced Settings and select pipette/gripper calibration. On restart, the banners should properly render. 
- For the OT-2, clear pipette calibration data. Note that the calibration banner on Devices > Robot should no longer render when the robot is not viewable (power off the robot).

# Changelog
- Pipette/Gripper card now uses offset date to determine if there's need for calibration.
- OT-2 calibration banner conditionally checks if robot is viewable before rendering.
- Added relevant testing. 

# Risk assessment
 low

### DIFF
--- a/app/src/organisms/Devices/InstrumentsAndModules.tsx
+++ b/app/src/organisms/Devices/InstrumentsAndModules.tsx
@@ -196,7 +196,8 @@ export function InstrumentsAndModules({
                 }
                 isPipetteCalibrated={
                   isOT3 && attachedLeftPipette?.ok
-                    ? attachedLeftPipette?.data?.calibratedOffset != null
+                    ? attachedLeftPipette?.data?.calibratedOffset
+                        ?.last_modified != null
                     : leftMountOffsetCalibration != null
                 }
                 mount={LEFT}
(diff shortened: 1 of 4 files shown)


koji (approved):
The changes look good to me!
Thank you for fixing them.

PR #13646, opened 2023-09-25
fix(app): fix odd module pages update issue

koji:
# Overview
The solution for this is to add refetch to `useAttacahedModules` to handle the following two cases.
This fix isn't an ideal way to resolve the issue. In the future we should refactor ODD components to remove this type of temp fix.

1. after a module calibration
2. connect/disconnect a module

close RQA-1710

Attached the video to the above ticket since the video size is too big for GitHub

# Test Plan
- protocol setup 
1. connect / disconnect a module
2. clear module cal data and calibrate a module

# Changelog
- add options to useAttachedModules
- add fetch option to useAttachedModules in ProtocolSetupModules

# Review requests

# Risk assessment
low


mjhuff (approved):
Tested and works - lgtm, thank you!

PR #13647, opened 2023-09-25
refactor(api): Why does EquipmentBroker, the largest broker, not simply eat the other brokers?

SyntaxColoring:
# Overview

We have two very similar classes:

* `opentrons.equipment_broker.EquipmentBroker`
* `opentrons.broker.Broker`

These are both message brokers / event emitters. Despite the names, `EquipmentBroker` is actually the more generic of the two. `Broker` is a holdover from before ~v5.

This PR renames them to more accurately reflect their current standing:

* `opentrons.equipment_broker.EquipmentBroker` ➡️ `opentrons.util.broker.Broker`
* `opentrons.broker.Broker` ➡️ `opentrons.legacy_broker.LegacyBroker`

And also adds missing unit tests for the class formerly known as `EquipmentBroker`.

This goes towards:

* RSS-283. Having some kind of generic broker is helpful for monitoring a `ProtocolEngine`'s command execution, and it would be weird if we used something still called `EquipmentBroker`.
* RSS-270. We're a little closer to deleting the old broker in favor of the newer more generic one.

# Test Plan

We're covered by linting and unit tests.

# Review requests

None in particular. This is largely a global find+replace. If you want, you can give it a skim and see if any call sites don't make sense anymore. For example, pytest fixtures with out-of-date names.

# Risk assessment

Very low. Just renaming and reorganizing in this PR.


sfoster1 (approved):
Hooray!

SyntaxColoring:
I'm holding off on merging this for a minute. We might want to retarget into `chore_release_7.0.1` to support resolving RSS-318 in v7.0.1. This can wait until the dust settles on v7.0.0.

PR #13653, opened 2023-09-26
feat(hardware): pass move ack id as part of the move complete info

ahiuchingau:
# Overview
On the Flex, for each move we command, we expect the following info to be reported from the axis firmware:
1. Motor position in um
2. Encoder position in um (if available)
3. Motor position ok flag
4. Encoder position ok flag
5. Ack ID

We have been checking the Ack id in the hardware move group runner; this tells us whether or not a move had reached the stop condition we set when it completed. For the most part, we want the ack id to match the move's stop condition – a home move should complete with the limit switch triggered, and if it didn't, something is really wrong and we should cancel any other scheduled moves.

But sometimes, it's okay if the ack id doesn't match the stop condition. For instance, while the robot is performing edge finding during calibration, the capacitive probe is _not_ expected to reach the stop condition when it actually finds the edge of the surface. However, when we're calibrating for the Z height, if capacitive probe should make contact with the surface, otherwise, the operation should fail. 

The complexity of the move ack/condition warrants it to be exposed to the hardware controller so the caller of the move can determine what to do with it, instead of leaving the responsibility solely to the move group runner. 

All that is to say - we can now know, for a fact (we used to check the z position at the end of the move against the hardware- limit 😅 ), whether or not the probe touched the deck and can use that info to better our calibration process. 

# Changelog
* Motor position/move complete info is now passed to the hardware controller the form of a new dataclass: `MovePositionStatus` instead of a tuple of ints and bools
* Adding a bool to the `api.capactive_probe` return values that checks if the probe made contact with a surface

### DIFF
--- a/api/src/opentrons/hardware_control/ot3_calibration.py
+++ b/api/src/opentrons/hardware_control/ot3_calibration.py
@@ -7,7 +7,7 @@ from typing import Tuple, List, Dict, Any, Optional, Union
 import datetime
 import numpy as np
 from enum import Enum
-from math import floor, copysign, isclose
+from math import floor, copysign
 from logging import getLogger
 from opentrons.util.linal import solve_attitude, SolvePoints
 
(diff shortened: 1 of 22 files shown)


sfoster1 (approved):
Looks good feature-wise and the refactor definitely looks like an improvement!

api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -222,10 +218,10 @@ async def find_edge_binary(
     final_z_height_found = slot_edge_nominal.z
     for _ in range(edge_settings.search_iteration_limit):
         LOG.info(f"Checking position {checking_pos}")
-        interaction_pos = await _probe_deck_at(
+        interaction_pos, hit_deck = await _probe_deck_at(
             hcapi, mount, checking_pos, edge_settings.pass_settings, probe=probe
         )
-        hit_deck = _deck_hit(interaction_pos, checking_pos.z, edge_settings)
+        _verify_height(interaction_pos, checking_pos.z, edge_settings)
         if hit_deck:
```
vegano1, line 225: hit_deck no longer exists

vegano1 (approved):
Just found a small typo, besides that, this is an awesome change!

We should also test this on instrument calibration to make sure it all still works.

PR #13654, opened 2023-09-26
fix(app): remove pipette settings from overflow menu if no pipette settings found

mjhuff:
Closes RQA-1231, RQA-1228

# Overview

This PR addresses an OT-2 issue in which advanced pipette settings would display an empty modal for certain pipettes. HW indicates we don't actually have more pipette function updates on the backend, so the plan is not to show "pipette settings as an option within the overflow menu (and in general) if the pipette does not have settings stored on the server.

### Before 

Screenshot 2023-08-04 at 1 13 49 PM

### After

Screenshot 2023-09-26 at 10 58 55 AM

# Test Plan
- Verify that existing pipettes on the /pipettes/settings endpoint still show an overflow menu option with an appropriate modal.
- Verify that a pipette that is not able to store settings on the /pipettes/settings endpoint does not show the overflow menu (P3HSV202019072249 on ParityBot, example).

# Changelog

- Fixed text errors displaying extra periods.
- Moved the selector logic up a few component layers to PipetteCard to conditionally render the overflow button.
- Improved the conditional check within PipetteSettingsSlideout, as there are instances where this is rendered without the overflow menu (mainly after attaching a different pipette).
- Added relevant testing.
- 

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ConfigurePipette/__tests__/ConfigurePipette.test.tsx
+++ b/app/src/organisms/ConfigurePipette/__tests__/ConfigurePipette.test.tsx
@@ -4,7 +4,6 @@ import { renderWithProviders } from '@opentrons/components'
 import { i18n } from '../../../i18n'
 import * as RobotApi from '../../../redux/robot-api'
 import { ConfigurePipette } from '../../ConfigurePipette'
-import { getAttachedPipetteSettingsFieldsById } from '../../../redux/pipettes'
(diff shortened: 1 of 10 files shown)


koji (approved):
The changes look good to me.
Tested this branch with an OT-2 and it worked as expected.
Thank you for fixing this.

PR #13657, opened 2023-09-26
fix(api): specify gripper probe

sfoster1:
462049edbc added the ability to calibrate the 8 channel pipette, but broke calibration for gripper and possibly 96 channel while doing so since those devices need to calibrate with the secondary probe by default.

This fixes that problem by explicitly forwarding the requested probe to the internals.


Laura-Danielle (approved):
lgtm

PR #13658, opened 2023-09-26
feat(protocol-designer, step-generation): remove hardcoded 'fixedTrash'

jerader:
closes RAUT-741

# Overview

Protocol designer has always had a hard coded fixed trash, this PR gets rid of that concept and instead creates a labware entity for it. This PR is mainly intended for that and doesn't support moving trash bin to a new location or removing it.

# Test Plan

With the deck modification FF turned on, create a protocol for the Flex and see that a trash bin entity is selectable in the module and other tile. I did not wire anything else up there so you can't select a location for it yet, it defaults to 'A3'

Create a protocol for the Flex and Ot-2, add a transfer or mix step to them and see that the trash bin is selectable, see in the advanced setting of dispense in transfer form that the blow out locations include the trash. Download the protocol and look in the json file, see 3 things: 1) a `loadLabware` is created for the trash, 2) a trash is included in the `labwareLocationUpdate`, 3) the trash labware definition is included. 

# Changelog

- remove all instances of defaulting the `labwareLocationUpdate` and `blowout_location` to be fixed trash and instead create a container for the trash entity to be added to the labware entities.
- add `trashBin` as an additional item in formik and add it as selectable in the modules and other tile
- add a migration `7.1.0` that migrates protocols to remove the `fixedTrash` id
- remove any instance of filtering out the trash, if applicable.
- remove the hardcoded fixed trash in `dropTip` in step generation and instead grab the trash entity and drop tip to that trash entity labwareId
- extend `additionalEquipment` type in `FormState` to include `trashBin`

# Review requests

The one thing that might be confusing is the fact that I extended the `additionalEquipment` array to include `trashBin` in the  `FormState` type. This made is easier to add/remove a trash bin in the create file wizard since that `addtionalEquipment` key already exists. Perhaps I need to make a new key for it. The thing is that `wasteChute` is part of that key and `wasteChute` and `trashBin` are pretty similar in terms of roles -> waste Chute is basically an enhanced trash bin and does all that the trash bin does + more. So if the waste chute is in `additionalEquipment` array then shouldn't trash bin go there too?

# Risk assessment

low


protocol-designer/cypress/integration/migrations.spec.js
```diff
@@ -12,61 +12,61 @@ describe('Protocol fixtures migrate and match snapshots', () => {
   })
 
   const testCases = [
+    //  TODO(jr, 9/26/23): when 7.1 is more developed, lets go back and fix these
```
jerader, line 15: made a ticket for this to fix later: https://opentrons.atlassian.net/browse/RAUT-759

koji:
>With the deck modification FF turned on, create a protocol for the Flex and see that a trash bin entity is selectable in the module and other tile. 

Does this mean we can see trash bin in here? (I could see trash bin in the modal).
Screenshot 2023-09-26 at 5 24 10 PM

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -153,7 +154,11 @@ export const DeckSetupContents = (props: ContentsProps): JSX.Element => {
     deckDef,
     robotType,
   } = props
-  const trashSlot = robotType === FLEX_ROBOT_TYPE ? FLEX_TRASH_SLOT : '12'
+  const trashSlot = Object.values(activeDeckSetup.labware).find(
+    lw =>
+      lw.labwareDefURI === OT_2_TRASH_DEF_URI ||
+      lw.labwareDefURI === FLEX_TRASH_SLOT
+  )?.slot
```
koji: nit
maybe create a function, like getTrashSlot?
But this is totally optional since only twice. If three times, def recommend to create a function for filtering slots.

protocol-designer/src/load-file/migration/7_1_0.ts
```diff
@@ -0,0 +1,125 @@
...
+  appData: ProtocolFile<DesignerApplicationData>
+): ProtocolFile => {
+  const { designerApplication, robot, commands } = appData
+  const labwareLocationUpdate: LabwareLocationUpdate =
+    designerApplication?.data?.savedStepForms[INITIAL_DECK_SETUP_STEP_ID]
+      .labwareLocationUpdate
+  const allLatestDefs = getOnlyLatestDefs()
+
+  const robotType = robot.model
+  const trashSlot = robot.model === FLEX_ROBOT_TYPE ? 'A3' : '12'
```
koji, line 33:
```suggestion
  const trashSlot = robot.model === FLEX_ROBOT_TYPE ? 'A3' : '12'
```
```suggestion
  const trashSlot = robotType === FLEX_ROBOT_TYPE ? 'A3' : '12'
```

koji: nit

```suggestion
  const trashSlot = robotType === FLEX_ROBOT_TYPE ? 'A3' : '12'
```

protocol-designer/src/ui/labware/__tests__/selectors.test.ts
```diff
@@ -321,24 +322,24 @@ describe('labware selectors', () => {
...
       ])
     })
   })
 
   describe('_sortLabwareDropdownOptions', () => {
     const trashOption = {
-      name: 'Some kinda fixed trash',
-      value: FIXED_TRASH_ID,
+      name: 'Trash Bin',
+      value: 'fixedTrash',
```
koji: Don't you want to use `export const FIXED_TRASH_ID = 'fixedTrash'`?

jerader:
> Does this mean we can see trash bin in here? (I could see trash bin in the modal). Screenshot 2023-09-26 at 5 24 10 PM

@koji , ya it will be added there eventually, i didn't wire it up yet. This PR is meant to just remove the trash from being a "fixedTrash"

protocol-designer/src/ui/labware/__tests__/selectors.test.ts
jerader: yeah, sorry this is confusing. I should call the constant something else since I'm trying to remove all usages of `FIXED_TRASH_ID`. I'll address this.

koji (approved):
Followed the test plan and each item worked as expected.

PR #6754, opened 2020-10-13
fix(api): Error instead of infinite-looping when disposal_volume is too high to make progress

SyntaxColoring:
# Overview

This PR turns code like this into an error:

```python
# disposal_volume too high.
# There would be no room left in the tip to move liquid to dests.
p300.distribute(123, source, dests, disposal_volume=300)
```

The protocol API's current behavior for code like this depends on the exact circumstances. For example, sometimes, it hangs in an infinite loop (as reported in #6170). Other times, it seems to treat the `distribute()` step as a no-op.

This PR turns one specific variant into an error, and moves towards characterizing other variants.

# Changelog

* Fix #6170. Before, the test protocol in that ticket would infinite-loop. Now, it raises a `ValueError` internally.

* Leave these related bugs *unfixed for now*:

  * When the pipette tip has a lower maximum volume than the pipette itself, it seems like an invalid disposal volume can make the `distribute()` effectively *no-op*. It picks up a tip and then drops it, without any sub-steps.

    This PR's unit tests caught this by chance. They're `xfail`'d.

    If we merge this PR without a fix for this bug, I'll ticket this bug separately.

  * Even with this PR, another variant of #6170, calling `transfer()` instead of `distribute()`, still hangs.

     ```python
     metadata = {"apiLevel": "2.7"}

     def run(protocol):
         labware = protocol.load_labware('nest_12_reservoir_15ml', 1)
         tip_rack = protocol.load_labware('opentrons_96_filtertiprack_200ul', 2)
         pipette = protocol.load_instrument('p300_single', mount='left', tip_racks=[tip_rack])
         pipette.transfer(10, labware.wells()[0], labware.wells()[1], disposal_volume=200)
     ```

     I haven't looked into why.

     I found this by manual fiddling, so it's not currently represented in the unit tests.

     If we merge this PR without a fix for this bug, I'll ticket this bug separately.

# Risk assessment

Medium.

There's a risk that this PR accidentally changes edge-case behavior in runnable protocols without an `apiLevel` bump.

This PR is intended to maintain the exact current behavior of all runnable protocols. It should merely improve how the error is presented for certain non-runnable protocols. (Instead of infinite-looping, there should be an error message.)

But suppose someone's written a "wrong," but runnable, protocol. The protocol provides an invalid value to one of our API functions, but the function currently happens not to raise any error. If this PR accidentally *does* make it raise an error, that would break our API versioning policy.

It's possible that this is the case, because:

[...]

Issue #6170, opened 2020-07-17
bug: Inappropriate disposal or air_gap volume causes hang in transfers.py


tanaes:
# Overview

Accidentally including an inappropriate disposal volume (disposal_volume=pipette_volume) causes a hang in transfers. This leads to infinite simulation times and difficulty troubleshooting. 

It appears to be due to bad assumptions in this `while` statement in `transfers.py`:

https://github.com/Opentrons/opentrons/blob/ee1b5041e69dc930a91ed330fbef17d84a0ae009/api/src/opentrons/protocol_api/transfers.py#L564-L569

I'm not familiar with the code base, but it seems like you want to handle the case where the summed value exceeds `self._max_volume`.

# Steps to reproduce

Minimal example protocol:

```python
from opentrons import protocol_api

metadata = {'apiLevel': '2.5',
            'author': 'Jon Sanders'}

def run(protocol: protocol_api.ProtocolContext):

    protocol.home()

    tiprack_reagents = protocol.load_labware('opentrons_96_tiprack_20ul', 
                                             11)
    reagents = protocol.load_labware('opentrons_96_aluminumblock_generic_pcr_strip_200ul',
                                     1, 'reagents')
    samples = protocol.load_labware('biorad_96_wellplate_200ul_pcr',
                                   4, 'samples')

    pipette_right = protocol.load_instrument('p10_multi', 
                                            'right',
                                            tip_racks=[tiprack_reagents])


    pipette_right.distribute(10,
                         reagents['A1'],
                         samples.wells(),
                         disposal_volume=10,
                         new_tip='once')

```

Out from kernel interrupt:

```
---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-8-b50bc670336c> in <module>
      1 protocol_file = open('../test.py', 'r')
      2 
----> 3 runlog, _bundle = simulate(protocol_file)
      4 
      5 protocol_file.close()

~/miniconda3/envs/py3/lib/python3.7/site-packages/opentrons/simulate.py in simulate(protocol_file, file_name, custom_labware_paths, custom_data_paths, propagate_logs, hardware_simulator_file_path, log_level)
    342         scraper = CommandScraper(stack_logger, log_level, context.broker)
    343         try:
--> 344             execute.run_protocol(protocol, context)
    345             if isinstance(protocol, PythonProtocol)\
    346                and protocol.api_level >= APIVersion(2, 0)\

[...]

PR #13659, opened 2023-09-26
fix(api): Addition of distribution error to prevent invalid disposal values

CaseyBatten:
# Overview
Prevents the distribute functionality from accepting a disposal value equal to or greater than the maximum value of the instrument. This was necessary to prevent an infinite looping issue where when disposal values greater than or equal to the maximum volume of an instrument were provided, which would cause the looping component of the _plan_distribute function to get caught in an state that was impossible to complete. This would seek to solve the issue detailed in RSS-347.

### DIFF
--- a/api/src/opentrons/protocols/advanced_control/transfers.py
+++ b/api/src/opentrons/protocols/advanced_control/transfers.py
@@ -696,11 +696,12 @@ class TransferPlan:
                *.. Aspirate -> Air gap -> Touch tip ->..
                .. Aspirate -> .....*
         """
-        self._check_valid_volume_parameters(
-            disposal_volume=self._strategy.disposal_volume,
-            air_gap=self._strategy.air_gap,
-            max_volume=self._instr.max_volume,
-        )
+        # TODO: verify if _check_valid_volume_parameters should be re-enabled here
+        # self._check_valid_volume_parameters(
+        #     disposal_volume=self._strategy.disposal_volume,
+        #     air_gap=self._strategy.air_gap,
+        #     max_volume=self._instr.max_volume,
+        # )
         plan_iter = self._expand_for_volume_constraints(
             # todo(mm, 2021-03-09): Is it right to use _instr.max_volume here?
             # Why don't we account for tip max volume, disposal volume, or air
--- a/api/src/opentrons/protocols/advanced_control/transfers.py
+++ b/api/src/opentrons/protocols/advanced_control/transfers.py
@@ -534,6 +534,12 @@ class TransferPlan:
             ]
         return sources, targets
 
+    def _check_valid_disposal_volume(self):
+        if self._strategy.disposal_volume >= self._instr.max_volume:
+            raise ValueError(
+                "The disposal volume must be less than the maximum volume of the pipette"
+            )
+
     def _plan_distribute(self):
         """
         * **Source/ Dest:** One source to many destinations
@@ -572,10 +578,7 @@ class TransferPlan:
 
         """
 
-        if self._strategy.disposal_volume >= self._instr.max_volume:
-            raise ValueError(
-                "The Disposal Volume must be less than the Maximum Volume of the Instrument"
-            )
+        self._check_valid_disposal_volume()
 
         # TODO: decide whether default disposal vol for distribute should be
         # pipette min_vol or should we leave it to being 0 by default and
--- a/api/src/opentrons/protocols/advanced_control/transfers.py
+++ b/api/src/opentrons/protocols/advanced_control/transfers.py
@@ -484,6 +484,11 @@ class TransferPlan:
             -> Blow out -> Touch tip -> Drop tip*
         """
         # reform source target lists
+        self._check_valid_volume_parameters(
+            self._strategy.disposal_volume,
+            self._strategy.air_gap,
+            self._instr._core.get_working_volume(),
+        )
         sources, dests = self._extend_source_target_lists(self._sources, self._dests)
         plan_iter = self._expand_for_volume_constraints(
             self._volumes,
@@ -534,12 +539,6 @@ class TransferPlan:
             ]
         return sources, targets
 
-    def _check_valid_disposal_volume(self, disposal_volume, working_volume):
-        if disposal_volume >= working_volume:
-            raise ValueError(
-                "The disposal volume must be less than the working volume of the pipette"
-            )
-
     def _plan_distribute(self):
         """
         * **Source/ Dest:** One source to many destinations
@@ -578,8 +577,10 @@ class TransferPlan:
 
         """
 
-        self._check_valid_disposal_volume(
-            self._strategy.disposal_volume, self._instr._core.get_working_volume()
+        self._check_valid_volume_parameters(
+            self._strategy.disposal_volume,
+            self._strategy.air_gap,
+            self._instr._core.get_working_volume(),
         )
 
         # TODO: decide whether default disposal vol for distribute should be
@@ -644,6 +645,7 @@ class TransferPlan:
         """Split a sequence of proposed transfers if necessary to keep each
         transfer under the given max volume.
         """
+        assert max_volume > 0
         for volume, target in zip(volumes, targets):
             while volume > max_volume * 2:
                 yield max_volume, target
@@ -691,6 +693,11 @@ class TransferPlan:
                *.. Aspirate -> Air gap -> Touch tip ->..
                .. Aspirate -> .....*
         """
+        self._check_valid_volume_parameters(
+            self._strategy.disposal_volume,
+            self._strategy.air_gap,
+            self._instr._core.get_working_volume(),
+        )
         plan_iter = self._expand_for_volume_constraints(
             # todo(mm, 2021-03-09): Is it right to use _instr.max_volume here?
             # Why don't we account for tip max volume, disposal volume, or air
@@ -853,6 +860,20 @@ class TransferPlan:
             return (rel_y * diff_vol) + min_v
 
         return [_map_volume(i) for i in range(total)]
+
+    def _check_valid_volume_parameters(self, disposal_volume, air_gap, working_volume):
+        if air_gap >= working_volume:
+            raise ValueError(
+                "The air gap must be less than the working volume of the pipette"
+            )
+        elif disposal_volume >= working_volume:
+            raise ValueError(
+                "The disposal volume must be less than the working volume of the pipette"
+            )
+        elif disposal_volume + air_gap >= working_volume:
+            raise ValueError(
+                "The sum of the air gap and disposal volume must be less than the working volume of the pipette"
+            )
 
     def _check_valid_well_list(self, well_list, id, old_well_list):
         if self._api_version >= APIVersion(2, 2) and len(well_list) < 1:
--- a/api/src/opentrons/protocols/advanced_control/transfers.py
+++ b/api/src/opentrons/protocols/advanced_control/transfers.py
@@ -870,15 +870,15 @@ class TransferPlan:
     ):
         if air_gap >= max_volume:
             raise ValueError(
-                "The air gap must be less than the working volume of the pipette"
+                "The air gap must be less than the maximum volume of the pipette"
             )
         elif disposal_volume >= max_volume:
             raise ValueError(
-                "The disposal volume must be less than the working volume of the pipette"
+                "The disposal volume must be less than the maximum volume of the pipette"
             )
         elif disposal_volume + air_gap >= max_volume:
             raise ValueError(
-                "The sum of the air gap and disposal volume must be less than the working volume of the pipette"
+                "The sum of the air gap and disposal volume must be less than the maximum volume of the pipette"
             )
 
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -571,6 +571,12 @@ def _plan_distribute(self):
                .. Dispense air gap -> ...*
 
         """
+
+        if self._strategy.disposal_volume >= self._instr.max_volume:
+            raise ValueError(
+                "The Disposal Volume must be less than the Maximum Volume of the Instrument"
```
ecormany: Looks good! We can probably lowercase all these terms in the error text:

```suggestion
                "The disposal volume must be less than the maximum volume of the pipette"
```

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -534,6 +534,12 @@ def _extend_source_target_lists(
             ]
         return sources, targets
 
+    def _check_valid_disposal_volume(self):
+        if self._strategy.disposal_volume >= self._instr.max_volume:
```
SyntaxColoring: * Do we need to check for the "working volume" or something like that, instead of `self._instr.max_volume` here?
* Do we need to check for negative numbers?

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -571,6 +577,9 @@ def _plan_distribute(self):
                .. Dispense air gap -> ...*
 
         """
+
+        self._check_valid_disposal_volume()
```
SyntaxColoring: Let's add tests for this. The unit tests in `test_transfers.py` are probably the right place to start, unless those turn out to be horrific.

SyntaxColoring (commented):
Looks great, thank you! A couple notes for moving forward with this.

api/src/opentrons/protocols/advanced_control/transfers.py
ecormany: Not for fixing the narrow bug: infinite loop on analysis. It has to be ≥ the pipette max volume to trigger that.

You can get into other (less bad? merely different?) bug behavior by doing things like having a Flex 1000 µL pipette pick up a 50 µL tip and then specifying disposal volume of ≥50. The whole `distribute()` command no-ops then. The same goes for 200 µL filter tips on a 300 µL OT-2 pipette, etc. See RSS-344.

SyntaxColoring: Good grief.

SyntaxColoring:
Okay, I've done some archeology and found that this was previously reported as GitHub issue #6170 and we tried to fix it in PR #6754. We closed that PR without merging it because of [these reasons](https://github.com/Opentrons/opentrons/pull/6754#issuecomment-804499335). 2.5 years later, I no longer think those reasons are good. In particular, screw this:

> ...by my understanding of our [Python Protocol API's versioning policy](https://docs.opentrons.com/v2/versioning.html), we must characterize and preserve that behavior in fixing these bugs. Even though it was undocumented, and didn't make sense, and only happened in incorrectly written protocols.

Versioning in the Python Protocol API is good to avoid flagrantly breaking people's protocols. But it shouldn't prevent us from fixing flagrant *bugs* like infinite loops. That's not helping anybody.

-----

So, for this PR, here's what we should do:

1. Let's definitely add tests. Let's start with the ones from PR #6754. We'll have to lightly adapt them to work in our modern codebase, and we'll have to address some pending `# todo` comments, but it'll be much better than starting from scratch.
2. As described in [the old issue](https://github.com/Opentrons/opentrons/issues/6170), we should solve this for *`air_gap`*, not just `disposal_volume`. We can crib from [the old PR](https://github.com/Opentrons/opentrons/pull/6754).
3. Pending the two points above, and another round of code review, let's move forward with merging this despite the versioning concerns that we had 2.5 years ago.

api/src/opentrons/protocols/advanced_control/transfers.py
SyntaxColoring: See https://github.com/Opentrons/opentrons/pull/13659#issuecomment-1738022236.

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -484,6 +484,11 @@ def _plan_transfer(self):
             -> Blow out -> Touch tip -> Drop tip*
         """
         # reform source target lists
+        self._check_valid_volume_parameters(
+            self._strategy.disposal_volume,
+            self._strategy.air_gap,
+            self._instr._core.get_working_volume(),
+        )
         sources, dests = self._extend_source_target_lists(self._sources, self._dests)
```
SyntaxColoring: A few things here:

1. For readability, let's move this `_check_valid_volume_parameters()` call below the `self._extend_source_target_lists()` call. We want to avoid interrupting the `# reform source target lists` comment, and we also want to keep the `_check_valid_volume_parameters()` call as close as possible to the `self._instr.max_volume - self._strategy.disposal_volume - self._strategy.air_gap,max_volume` calculation.
2. This is a nitpicky style thing, but I suggest providing the arguments as keywords, instead of positionally:

   ```python
    self._check_valid_volume_parameters(
        disposal_volume=self._strategy.disposal_volume,
        air_gap=self._strategy.air_gap,
        max_volume=self._instr._core.get_working_volume(),
    )
   ```

   We have 3 arguments and they're all floats, so I'm worried about them getting mixed up.
3. This particular call to `_check_valid_volume_parameters()` should use *`self._instr._core.max_volume`,* not `self._instr._core.get_working_volume()`. We need to exactly match what's in the call to `self._expand_for_volume_constraints()`.

   To make this kind of mistake harder to make, we could do the calculation and validation in one place. [Something like this.](https://gist.github.com/SyntaxColoring/c88f9c2d5b97ecf1dd2ab3cfc6f17c1c)

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -633,6 +645,7 @@ def _expand_for_volume_constraints(
         """Split a sequence of proposed transfers if necessary to keep each
         transfer under the given max volume.
         """
+        assert max_volume > 0
```
SyntaxColoring, line 651: Let's leave a comment here to explain the intent:

```suggestion
        # A final defense against an infinite loop.
        # Raising a proper exception with a helpful message is left to calling code,
        # because it has more context about what the user is trying to do.
        assert max_volume > 0
```

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -571,6 +576,13 @@ def _plan_distribute(self):
                .. Dispense air gap -> ...*
 
         """
+
+        self._check_valid_volume_parameters(
+            self._strategy.disposal_volume,
+            self._strategy.air_gap,
+            self._instr._core.get_working_volume(),
+        )
+
```
SyntaxColoring: Ditto my previous comments about:

1. Providing arguments with keyword syntax instead of positionally.
2. Using `self._instr.max_volume` instead of `self._instr._core.get_working_volume()`. And maybe combining the calculation with the validation to make that kind of mistake harder to make.

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -680,6 +693,11 @@ def _plan_consolidate(self):
                *.. Aspirate -> Air gap -> Touch tip ->..
                .. Aspirate -> .....*
         """
+        self._check_valid_volume_parameters(
+            self._strategy.disposal_volume,
+            self._strategy.air_gap,
+            self._instr._core.get_working_volume(),
+        )
```
SyntaxColoring: Ditto my previous comment about providing arguments with keyword syntax, instead of positionally.

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -843,6 +861,20 @@ def _map_volume(i):
 
         return [_map_volume(i) for i in range(total)]
 
+    def _check_valid_volume_parameters(self, disposal_volume, air_gap, working_volume):
```
SyntaxColoring: Let's give type annotations to this new function.

```suggestion
    def _check_valid_volume_parameters(self, disposal_volume: float, air_gap: float, working_volume: float) -> None:
```

Normally, forgetting the type annotations would be an error because of [our project config](https://github.com/Opentrons/opentrons/blob/c0543a6e0a89e38a06117c5124085a4bb45443c4/api/mypy.ini#L5), but I think we have that overridden for this file because it's legacy code and we haven't gotten around to fixing it up. So we *could* leave them out. But let's leave a little less work for our future selves and add them now.

api/tests/opentrons/protocols/advanced_control/test_transfers.py
```diff
@@ -796,6 +796,83 @@ def test_all_options(_instr_labware):
...
+            700,
+            lw1.columns()[0][0],
+            lw2.rows()[0][1:3],
+            _instr_labware["instr"],
+            max_volume=_instr_labware["instr"].hw_pipette["max_volume"],
+            api_version=_instr_labware["ctx"].api_version,
+            mode="distribute",
+            options=options,
+        )
+        list(plan)
```
SyntaxColoring, line 823: This iterator thing was a little tricky, so let's leave an explanatory comment. Shamelessly copy-pasting from #6754:

```suggestion
        # Exhaust the iterator in case it raises the expected exception lazily.
        list(plan)
```

And ditto for the other two test functions.

api/tests/opentrons/protocols/advanced_control/test_transfers.py
```diff
@@ -796,6 +796,83 @@ def test_all_options(_instr_labware):
...
+def test_invalid_air_gap_distribute(_instr_labware):
+    _instr_labware["ctx"].home()
+    lw1 = _instr_labware["lw1"]
+    lw2 = _instr_labware["lw2"]
+    # Supply the disposal volume assessment with the instrument followed by a disposal value equal to the max volume
+
+    options = tx.TransferOptions()
+    options = options._replace(
+        transfer=options.transfer._replace(air_gap=_instr_labware["instr"].max_volume)
+    )
```
SyntaxColoring: This creates a `tx.TransferOptions` with all default values, and then, as a second step, modifies `.transfer.air_gap` to be `._instr_labware["instr"].max_volume`.

That works, but it's more common to do it all in one step. That looks like this:

```suggestion
    options = tx.TransferOptions(
        transfer=tx.Transfer(
            air_gap=_instr_labware["instr"].max_volume
        )
    )
```

Ditto for the other two test functions.

api/tests/opentrons/protocols/advanced_control/test_transfers.py
```diff
@@ -796,6 +796,83 @@ def test_all_options(_instr_labware):
     assert xfer_plan_list == exp1
 
 
+def test_invalid_air_gap_disposal_sum_distribute(_instr_labware):
```
SyntaxColoring, line 799: **I think we should leave this to a separate PR, but:**

This adds a few test functions that differ only in the values that they provide for `disposal_volume` and `air_gap`. This would be a good use case for [pytest parametrization](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#parametrize-basics). We could write this as one function that gets parametrized with the different `disposal_volume` and `air_gap` combinations.

We could also parametrize this over the *mode* of `TransferPlan`: `"distribute"`, `"consolidate"`, or `"transfer"`. That would be good because our bugfix *does* touch those other functions.

SyntaxColoring (commented):
This looks great, thank you. Here are several minor comments, and then this looks good to merge.

api/src/opentrons/protocols/advanced_control/transfers.py
SyntaxColoring: I just noticed the #todo comment right below here that points out how `air_gap` and `disposal_volume` aren't factoring into the loop *at all* here in `_plan_consolidate()`. That seems really suspicious to me, but for our sanity, I think we need to treat that as a separate bug for a separate PR. For now, let's remove `_check_valid_volume_parameters()` from `_plan_consolidate()`. We'll keep your fix scoped to the one infinite loop that we know and understand.

CaseyBatten: Gotcha! Will do, gonna recommit with the plan_consolidate call to the volume param check removed.

CaseyBatten:
Primary fail issues cleared, formatting good, I believe this is ready for merge.

SyntaxColoring (commented):
This looks good to merge, assuming CI passes. I'm a little worried about that failing G-code test. I'm checking now to see if it's something that was introduced by this PR, or whether it's the known issue that was fixed in https://github.com/Opentrons/opentrons/pull/13670.

SyntaxColoring (approved):
G-code tests are failing on `edge`, so I'm presuming those failures are unrelated to this PR. (I'm running into obscure unrelated errors trying to run them locally, so I can't be sure.)

This looks good to merge. Thank you!

PR #13660, opened 2023-09-26
feat(api) Support changing current based on tip configuration

Laura-Danielle:
# Overview

With this PR, we should now be able to perform partial tip configuration movements in the hardware controller. 

# Test Plan

- Check that the following works:
- [ ] Full configuration pick-ups
- [ ] Single pick ups (8 channel and 96 channel)
- [ ]  Column pick ups (96 channel only)
- [ ] Quadrant pick ups (96 channel only)

We should also check that these movements persist for liquid handling functions as well.

# Changelog

- Change the current scalar in the pipette configurations to be a current map
- Utilize this new current map to determine the pick up current only (HW folks, not sure if anything should _also_ change for drop tip. Please let me know).

# Review requests

HW folks, please comment and tell me what the partial currents should _actually_ be. I made up numbers here.
SW folks, let me know if anything seems weird/gross/whatever.

# Risk assessment

Medium. We should ensure that normal configuration tip pick-ups have not deviated from normal.


api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -200,15 +200,18 @@ class NozzleConfigurationManager:
...
-        self._current_scalar: Final[float] = current_scalar
+        self._pick_up_current_map: Final[Dict[int, float]] = pick_up_current_map
 
     @classmethod
     def build_from_nozzlemap(
-        cls, nozzle_map: Dict[str, List[float]], current_scalar: float
+        cls,
+        nozzle_map: Dict[str, List[float]],
+        default_pickup_current: float,
+        pick_up_current_map: Optional[Dict[int, float]] = None,
```
sfoster1: in what situations would we not have a `pick_up_current_map`? does it have to be `= None`?

sfoster1 (changes requested):
The code looks good to me! Will have to test on a machine though

sfoster1 (approved):
Review now with appropriate resolution selected

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
Laura-Danielle: single channels do not have a pick up current map.

shared-data/pipette/definitions/2/general/eight_channel/p1000/3_5.json
```diff
@@ -40,7 +40,17 @@
...
+    "availableConfigurations": [1, 2, 3, 4, 5, 6, 7, 8],
+    "perTipPickupCurrent": {
+      "1": 0.2,
+      "2": 0.3,
+      "3": 0.4,
+      "4": 0.5,
+      "5": 0.6,
+      "6": 0.7,
+      "7": 0.8,
+      "8": 0.9
```
andySigler: are these real values? if not let's get those

andySigler: Also is the current value defined in the `pickUpTipConfigurations` being used anywhere anymore?

sfoster1 (changes requested):
From HW:
Let's change the values for the `perTipPickupCurrent` configurations such that:
- For 1-channel pickups, the current is the same as the pickup current for the single-channel version of the pipette
- For 2-7 channel pickups inclusive, the current is `Ntips / Nchannels` times the as-of-now 8channel pickup current
- For 8 channel pickups, the current is the same is the as-of-now 8channel pickup current

These will be good for the OT-2 pipettes (the different nozzle geometry means the overlap positioning won't change much), and they'll be good starting points for more in-depth testing of the Flex configurations.

From me:
We also should probably get rid of `pickUpTipConfigurations.current`.

shared-data/pipette/definitions/2/general/eight_channel/p1000/3_5.json
Laura-Danielle: no these are not real numbers. `pickUpTipConfigurations` current will not be used anymore.

shared-data/pipette/definitions/2/general/eight_channel/p1000/1_0.json
```diff
@@ -1,10 +1,9 @@
 {
   "$otSharedSchema": "#/pipette/schemas/2/pipettePropertiesSchema.json",
-  "displayName": "Flex 8-Channel 1000 μL",
+  "displayName": "Flex 8-Channel 1000 \u03bcL",
```
sfoster1:
```suggestion
  "displayName": "Flex 8-Channel 1000 μL",
```

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
```diff
@@ -256,13 +262,10 @@ def update_nozzle_configuration(
...
-    def get_current_for_tip_configuration(self) -> float:
-        # TODO While implementing this PR I realized that
-        # the current scalar truly is inefficient for
-        # encapsulating varying currents needed for
-        # pick up tip so we'll need to follow this up
-        # with a lookup table.
-        return self._current_scalar
+    def get_tip_configuration_current(self) -> float:
+        return self._pick_up_current_map[
+            len(self._current_nozzle_configuration.map_store)
```
sfoster1: nit: i would love it if the nozzle configuration objects had like `tip_count` property that did this or something but it's ok if it works

sfoster1 (commented):
Code looks good, will try and get this tested!

caila-marashaj (approved):
looks good on hardware!

sfoster1 (approved):
Ah, looks like @caila-marashaj did some in person testing so we're good - let's just fix that malformed mu

api/src/opentrons/hardware_control/instruments/nozzle_manager.py
Laura-Danielle: sure can do that

PR #13661, opened 2023-09-27
fix(hardware-testing): Photometric test minor bug fixes

andySigler:
# Overview

- Fixed bug where liquid level tracker wasn't resetting when changing plates, so SW would think plate was overflowing when it wasn't actually
- Removed superfluous homing, which is much worse when testing 96ch b/c the plunger is so slow
- Consolidate 96ch Protocol files into one for grav and one for photo

# Test Plan

# Changelog

# Review requests

# Risk assessment


vegano1 (approved):
very nice,
the consolidation makes it much cleaner and easier to understand

PR #13664, opened 2023-09-27
docs(api): specify link text for most `:ref:` links and fix broken links

ecormany:
# Overview

We've decided that nearly all `:ref:` links in the API docs should specify custom link text. Doing so provides more graceful fallbacks in case we change something at the target and don't realize that there's an inbound link.

For example, look at this inline link and target:

**Link**

```
Any :ref:`v2-custom-labware` added to your Opentrons App...
```

**Target**

```
.. _v2-custom-labware:

Custom Labware
==============
```

We accidentally broke the anchor at the target, and this text was rendering as "Any v2-custom-labware added to your Opentrons App". If we changed the header to `Custom Labware is Cool` then the text would render as "Any Custom Labware is Cool added to your Opentrons App". This didn't happen, but it would be even worse, because it's a _valid_ link that wouldn't raise a warning when compiling the docs.

Updating the link to…
```
:ref:`custom labware <v2-custom-labware>`
```
…fixes both scenarios. If we break the anchor, Sphinx renders "Any custom labware added…" with no link. If we change the header, the link text is still `custom labware`.

### What isn't changing

We're deliberately leaving plain `:ref:` links for _cross-references_ that identify a page or section, i.e., they aren't hyperlinks of an ordinary phrase in running prose.

In other words, if we want to say "See the _How to Foo_ section" and we have an anchor `.. _foo:`, we'll still use:
```
:ref:`foo`
```

# Test Plan

Searched for all uses of `:ref:` and checked they're valid links in local `make -C api docs` build.

# Changelog

- A light touch across many pages. I tried to only change link text. Sometimes this meant shifting a word or two into or out of the hyperlink.
- Changed one `:ref:` link to a `:py:meth:` reference because I thought it made more sense.

# Review requests

Check the [sandbox](http://sandbox.docs.opentrons.com/docs-improve-refs/v2/).

# Risk assessment

v low

### DIFF
--- a/api/docs/v2/adapting_ot2_flex.rst
+++ b/api/docs/v2/adapting_ot2_flex.rst
@@ -18,7 +18,7 @@ Metadata and Requirements
 Flex requires you to specify an ``apiLevel`` of 2.15 or higher. If your OT-2 protocol specified ``apiLevel`` in the ``metadata`` dictionary, it's best to move it to the ``requirements`` dictionary. You can't specify it in both places, or the API will raise an error.
 
 .. note::
-    Consult the list of :ref:`version-notes` to see what effect raising the ``apiLevel`` will have. If you increased it by multiple minor versions to get your protocol running on Flex, make sure that your protocol isn't using removed commands or commands whose behavior has changed in a way that may affect your scientific results.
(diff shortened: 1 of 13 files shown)


jwwojak (approved):
This is a good and needed change. Integrating links into text with sentence case text improves readability and makes docs look less robotic/auto-generated. It is good to keep the target title when referencing it as part of a "See, _Title Here_" or "See also, _Title Here_" and with any other type of equivalent structure as well. 2 nice style techniques, IMO.

PR #13665, opened 2023-09-27
feat(api-client,react-api-client,app): stub out deck config api

brenthagen:
# Overview

stubs out deck config api-client and react-api-client get and update functions to unblock deck config app development. adds a skeleton of a device details deck configuration section.

re RAUT-683

# Test Plan

# Changelog

 - Stubs out deck config api-client and react-api-client

# Review requests

review the stubs, ideally try out the feature flag/updates

# Risk assessment

low, behind FF


b-cooper (approved):
Changes look great! I like the stubbing approach.

PR #13668, opened 2023-09-27
fix(app): fix protocolsetup next step button issue

koji:
# Overview
The issue is that nextStep doesn't track the actual required steps that a protocol file has. This pr check protocol data  and create an array to track the required steps

close RQA-1620

# Test Plan

# Changelog
- update ProtocolSetup component to add filteredStepKeysInOrder array to track the actual required steps for the protocol setup

# Review requests
For `filteredStepKeysInOrder`, if there is a better way, please let me know.

if (protocolData?.modules.length !== 0) {
      nextStepKeysInOrder = [...nextStepKeysInOrder, MODULE_SETUP_KEY]
    } else if (protocolData.labware.length !== 0) {
      nextStepKeysInOrder = [...nextStepKeysInOrder, LABWARE_SETUP_KEY]
    } else if (protocolData.liquids.length !== 0) {
      nextStepKeysInOrder = [...nextStepKeysInOrder, LIQUID_SETUP_KEY]
    }
    setStepKeysInOrder(
      nextStepKeysInOrder.sort(
        (a, b) => targetStepsOrder.indexOf(a) - targetStepsOrder.indexOf(b)
      )
    )
```

<!--
Describe any requests for your reviewers here.
-->

# Risk assessment
low
<!--
Carefully go over your pull request and look at the other parts of the codebase it may affect. Look for the possibility, even if you think it's small, that your change may affect some other part of the system - for instance, changing return tip behavior in protocol may also change the behavior of labware calibration.

Identify the other parts of the system your codebase may affect, so that in addition to your own review and testing, other people who may not have the system internalized as much as you can focus their attention and testing there.
-->

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
@@ -79,6 +79,9 @@ export function ProtocolRunSetup({
     LPC_KEY,
     LABWARE_SETUP_KEY,
   ])
+  const [targetStepKeyInOrder, setTargetStepKeyInOrder] = React.useState<
+    StepKey[]
+  >(stepsKeysInOrder)
 
   React.useEffect(() => {
     let nextStepKeysInOrder = stepsKeysInOrder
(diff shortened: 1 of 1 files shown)


brenthagen (approved):
this makes sense and works, but I think this `targetStepKeyInOrder` and the `stepsKeysInOrder` above both do not need to be local state set within a useEffect. the step keys and target step keys can just be assigned and filtered in the body of the function.

koji:
> this makes sense and works, but I think this `targetStepKeyInOrder` and the `stepsKeysInOrder` above both do not need to be local state set within a useEffect. the step keys and target step keys can just be assigned and filtered in the body of the function.

oh yeah, thanks!

PR #13669, opened 2023-09-27
feat(api): segment 96-channel tip pickup and dropoff moves

caila-marashaj:
## Overview
The 96 channel pipette can execute tip handling more efficiently if we split the tip motor's moves into two segments. This pr changes both pickup and dropoff so that we will do this by default when using a 96 channel.

This also includes a significant refactor of the dataclasses that pass information around for tip pickup and dropoff, since it would make these changes easier.

## Changelog
- all pickup and dropoff will use a more generic `TipActionSpec`, which will contain a list of `TipActionMoveSpec` objects
  - refactored `plan_check_pick_up_tip` and `plan_check_drop_tip` to accommodate this
- setting the currents during the larger 'pick up tip' or 'drop tip' command, rather than within each `press`, `droptipmove`, or tip motor move. I did this because the currents appeared to be the same for each of these actions anyway, **please let me know if this could potentially mess anything up.**
- for the 96 channel, moving the tip motors will use a singular common helper function- this gets rid of the `_motor_pick_up_tip` helper.
- added new configs to shared data to help with this
  - `prep_move_distance` and `prep_move_speed` for the first half of tip action moves, which will be slightly faster than when directly approaching the tiprack
- changed `tiprack_up` and `tiprack_down` to `end_tip_action_retract_distance_mm` and `connect_tiprack_distance_mm`(_I still don't love this name I'm open to feedback here_), in an effort to hopefully be more self-explanatory now that the code has grown and there's a bunch of other configs and variables with the word 'tiprack' in them

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
@@ -793,6 +793,7 @@ class OT3PipetteHandler:
         return TipActionSpec(
             tip_action_moves=build_presses(),
             shake_off_moves=self._build_pickup_shakes(instrument),
+            ending_z_retract_distance=instrument.config.end_tip_action_retract_distance_mm,
         )
 
     @staticmethod
--- a/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
@@ -73,7 +73,7 @@ class LiquidActionSpec:
 @dataclass(frozen=True)
 class TipActionMoveSpec:
     distance: float
-    currents: Dict[Axis, float]
+    currents: Optional[Dict[Axis, float]]
     speed: Optional[float]
 
 
@@ -81,8 +81,8 @@ class TipActionMoveSpec:
 class TipActionSpec:
     tip_action_moves: List[
         TipActionMoveSpec
-    ]  # can be presses, gear pickup, or gear dropoff
-    shake_off_moves: List[Tuple[top_types.Point, Optional[float]]]  # this is the same
+    ]
+    shake_off_moves: List[Tuple[top_types.Point, Optional[float]]]
     z_distance_to_tiprack: Optional[float] = None
     ending_z_retract_distance: Optional[float] = None
 
@@ -711,22 +711,40 @@ class OT3PipetteHandler:
 
         return []
 
-    def plan_check_pick_up_tip(
-        self,
-        mount: OT3Mount,
-        tip_length: float,
-        presses: Optional[int],
-        increment: Optional[float],
-    ) -> Tuple[TipActionSpec, Callable[[], None]]:
+    def plan_ht_pick_up_tip(self, mount) -> TipActionSpec:
         # Prechecks: ready for pickup tip and press/increment are valid
         instrument = self.get_pipette(mount)
         if instrument.has_tip:
             raise UnexpectedTipAttachError("pick_up_tip", instrument.name, mount.name)
         self._ihp_log.debug(f"Picking up tip on {mount.name}")
 
-        def add_tip_to_instr() -> None:
-            instrument.add_tip(tip_length=tip_length)
-            instrument.set_current_volume(0)
+        tip_motor_moves = self._build_tip_motor_moves(
+            prep_move_dist=instrument.pick_up_configurations.prep_move_distance,
+            clamp_move_dist=instrument.pick_up_configurations.distance,
+            prep_move_speed=instrument.pick_up_configurations.prep_move_speed,
+            clamp_move_speed=instrument.pick_up_configurations.speed,
+            plunger_current=instrument.plunger_motor_current.run,
+            tip_motor_current=instrument.pick_up_configurations.current,
+        )
+
+        return TipActionSpec(
+            tip_action_moves=tip_motor_moves,
+            shake_off_moves=[],
+            z_distance_to_tiprack=(-1 * instrument.connect_tiprack_distance_mm),
+            ending_z_retract_distance=instrument.end_tip_action_retract_distance_mm,
+        )
+
+    def plan_lt_pick_up_tip(
+        self,
+        mount: OT3Mount,
+        presses: Optional[int],
+        increment: Optional[float],
+    ) -> TipActionSpec:
+        # Prechecks: ready for pickup tip and press/increment are valid
+        instrument = self.get_pipette(mount)
+        if instrument.has_tip:
+            raise UnexpectedTipAttachError("pick_up_tip", instrument.name, mount.name)
+        self._ihp_log.debug(f"Picking up tip on {mount.name}")
 
         if presses is None or presses < 0:
             checked_presses = instrument.pick_up_configurations.presses
@@ -739,7 +757,6 @@ class OT3PipetteHandler:
             check_incr = increment
 
         pick_up_speed = instrument.pick_up_configurations.speed
-        pipette_axis = Axis.of_main_tool_actuator(mount)
 
         def build_presses() -> List[TipActionMoveSpec]:
             # Press the nozzle into the tip <presses> number of times,
@@ -756,7 +773,9 @@ class OT3PipetteHandler:
                         distance=press_dist,
                         speed=pick_up_speed,
                         currents={
-                            pipette_axis: instrument.pick_up_configurations.current
+                            Axis.by_mount(
+                                mount
+                            ): instrument.pick_up_configurations.current
                         },
                     )
                 )
@@ -766,36 +785,14 @@ class OT3PipetteHandler:
                     TipActionMoveSpec(
                         distance=backup_dist,
                         speed=pick_up_speed,
-                        currents={pipette_axis: instrument.plunger_motor_current.run},
+                        currents=None,
                     )
                 )
             return press_moves
 
-        if instrument.channels == 96:
-            tip_motor_moves = self._build_tip_motor_moves(
-                prep_move_dist=instrument.pick_up_configurations.prep_move_distance,
-                clamp_move_dist=instrument.pick_up_configurations.distance,
-                prep_move_speed=instrument.pick_up_configurations.prep_move_speed,
-                clamp_move_speed=instrument.pick_up_configurations.speed,
-                plunger_current=instrument.plunger_motor_current.run,
-                tip_motor_current=instrument.pick_up_configurations.current,
-            )
-
-            return (
-                TipActionSpec(
-                    tip_action_moves=tip_motor_moves,
-                    shake_off_moves=[],
-                    z_distance_to_tiprack=(-1 * instrument.connect_tiprack_distance_mm),
-                    ending_z_retract_distance=instrument.end_tip_action_retract_distance_mm,
-                ),
-                add_tip_to_instr,
-            )
-        return (
-            TipActionSpec(
-                tip_action_moves=build_presses(),
-                shake_off_moves=self._build_pickup_shakes(instrument),
-            ),
-            add_tip_to_instr,
+        return TipActionSpec(
+            tip_action_moves=build_presses(),
+            shake_off_moves=self._build_pickup_shakes(instrument),
         )
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -767,6 +766,16 @@ def build_presses() -> Iterator[Tuple[float, float]]:
                 yield (press_dist, backup_dist)
 
         if instrument.channels == 96:
+            tip_motor_moves = [
+                {
+                    "distance": instrument.pick_up_configurations.prep_move_distance,
```
ahiuchingau: Since we have a fixed set of keys, we should make use of dataclass here

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -885,12 +894,26 @@ def plan_check_drop_tip(
...
-        droptip = (
-            instrument.drop_configurations.distance
-            if is_96_chan
-            else instrument.plunger_positions.drop_tip
-        )
-        speed = instrument.drop_configurations.speed
+
+        if is_96_chan:
+            drop_seq = [
+                {
```
ahiuchingau: Same as above

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1918,11 +1918,20 @@ async def _motor_pick_up_tip(
...
+            for move in pipette_spec.tip_motor_moves:
+
+                move_target = {Axis.Q: move["distance"]}
+
+                move_target_start = MoveTarget.build(
+                    position=move_target,
+                    max_speed=move["speed"],
+                )
+                _, moves = self._move_manager.plan_motion(
+                    origin={Axis.Q: gear_origin_float}, target_list=[move_target_start]
```
ahiuchingau: Instead of using separate move group runners to execute this, why not include both targets to the target_list?

ahiuchingau (changes requested):
MoveManager can take in a list of targets, i think we can make use of that here

sfoster1 (approved):
Looks good iif it works, thoguh I'm wondering if maybe we're beyond the point of `plan_check_drop_tip` handling both motor and force drop being helpful

caila-marashaj:
> Looks good iif it works, thoguh I'm wondering if maybe we're beyond the point of `plan_check_drop_tip` handling both motor and force drop being helpful

I was thinking of doing a little refactoring just for cleanliness in a separate pr. can definitely add this to that

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_0.json
```diff
@@ -53,5 +57,7 @@
     "current": 2.0,
     "speed": 5
   },
-  "tipPresenceCheckDistanceMM": "8.0"
+  "tipPresenceCheckDistanceMM": "8.0",
+  "connectTiprackDistanceMM": "7.0",
```
sfoster1, line 61: we should add these to the schema

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1998,20 +1991,34 @@ async def pick_up_tip(
         )
 
         await self._move_to_plunger_bottom(realmount, rate=1.0)
-        if spec.pick_up_motor_actions:
-            await self._motor_pick_up_tip(realmount, spec.pick_up_motor_actions)
-        else:
-            await self._force_pick_up_tip(realmount, spec)
+        async with self._backend.motor_current(
```
sfoster1: We can't set the current like this, I think (know). We purposefully set the current of the Z much lower than normal when driving into tips to pick them up so we can use the motors as a force clutch. We should _not_ use the same current when we're going back up after the main press.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -762,60 +723,54 @@ def build_presses() -> Iterator[Tuple[float, float]]:
...
             )
         return (
-            PickUpTipSpec(
-                plunger_prep_pos=instrument.plunger_positions.bottom,
-                plunger_currents={
+            TipActionSpec(
+                currents={
                     Axis.of_main_tool_actuator(
                         mount
                     ): instrument.plunger_motor_current.run
```
sfoster1: this current isn't the right one - in pick up tip, for force pickups, it's the current of the _z_ that's important, not the current of the plunger. Note that after this refactor, there's no reference to `instrument.pick_up_configurations.current` outside of `Axis.Q`.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -840,87 +795,51 @@ def _shake_off_tips_drop(
...
-            droptip,
-            {Axis.of_main_tool_actuator(mount): instrument.plunger_motor_current.run},
-            {drop_tip_current_axis: instrument.drop_configurations.current},
-            speed,
-            home_after,
-            (Axis.of_main_tool_actuator(mount),),
-            is_96_chan,
-        )
+        currents = {
+            Axis.of_main_tool_actuator(mount): instrument.plunger_motor_current.run
```
sfoster1: this current is wrong; when dropping tip, we use a much higher current on the plunger motor than normal because we need to apply enough force to pip the tips off. That's `instrument.drop_configurations.current`.

sfoster1 (changes requested):
I like the impulse to refactor this code, but this has lost a lot of management of motor currents.

For force-pickup, we use a much lower current than normal; we need to set the low current (`uinstrument.pick_up_configurations.current` on the Z axis when pressing into the tip, and the default Z run current when raising out of the tip, because the low current may not be enough to move the Z.

For force-drop, we need to be using a much higher current on the plunger when interacting with the shroud (`instrument.drop_configurations.current`)

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -72,59 +70,21 @@ class LiquidActionSpec:
...
 @dataclass(frozen=True)
-class DropTipSpec:
-    drop_moves: List[DropTipMove]
-    shake_moves: List[Tuple[top_types.Point, Optional[float]]]
-    ending_current: Dict[Axis, float]
+class TipActionSpec:
+    tip_action_moves: List[
+        TipActionMoveSpec
+    ]  # can be presses, gear pickup, or gear dropoff
+    shake_off_moves: List[Tuple[top_types.Point, Optional[float]]]  # this is the same
```
sfoster1: the same as what?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -72,59 +70,21 @@ class LiquidActionSpec:
...
+    z_distance_to_tiprack: Optional[float] = None
```
sfoster1, line 85: can these be moves?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -751,70 +739,61 @@ def add_tip_to_instr() -> None:
...
                 press_dist = (
                     -1.0 * instrument.pick_up_configurations.distance
                     + -1.0 * check_incr * i
                 )
+                press_moves.append(
+                    TipActionMoveSpec(
+                        distance=press_dist,
+                        speed=pick_up_speed,
+                        currents={
+                            pipette_axis: instrument.pick_up_configurations.current
```
sfoster1: shouldn't this be for the z axis?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -751,70 +739,61 @@ def add_tip_to_instr() -> None:
...
+                    )
+                )
                 # move nozzle back up
                 backup_dist = -press_dist
-                yield (press_dist, backup_dist)
+                press_moves.append(
+                    TipActionMoveSpec(
+                        distance=backup_dist,
+                        speed=pick_up_speed,
+                        currents={pipette_axis: instrument.plunger_motor_current.run},
```
sfoster1: shouldn't this be the z axis?

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -751,70 +739,61 @@ def add_tip_to_instr() -> None:
...
+                    )
+                )
+            return press_moves
 
         if instrument.channels == 96:
+            tip_motor_moves = self._build_tip_motor_moves(
```
sfoster1: might be a lot easier if `_build_tip_motor_moves` took a `TipClassHandlingConfiguration` so this could be `self._build_tip_motor_moves(instrument.pick_up_configurations)`

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -839,87 +818,55 @@ def _shake_off_tips_drop(
...
 
-        bottom = instrument.plunger_positions.bottom
-        droptip = (
-            instrument.drop_configurations.distance
-            if is_96_chan
-            else instrument.plunger_positions.drop_tip
-        )
-        speed = instrument.drop_configurations.speed
-        shakes: List[Tuple[top_types.Point, Optional[float]]] = []
+        if is_96_chan:
```
sfoster1: nit: better to have this be `if instrument.channels == 96` if this isn't used anywhere else

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -839,87 +818,55 @@ def _shake_off_tips_drop(
...
+
+        else:
+            drop_seq = [
+                TipActionMoveSpec(
+                    distance=instrument.plunger_positions.drop_tip,
+                    speed=None,
+                    currents={
+                        Axis.of_main_tool_actuator(
+                            mount
+                        ): instrument.plunger_motor_current.run,
```
sfoster1: this needs to be the higher drop-tip current, since it's where we're driving the plunger to pop the tips off

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1684,7 +1688,7 @@ async def _move_to_plunger_bottom(
         """
         checked_mount = OT3Mount.from_mount(mount)
         instrument = self._pipette_handler.get_pipette(checked_mount)
-        if instrument.current_volume > 0:
+        if check_current_vol and instrument.current_volume > 0:
```
sfoster1, line 1691: Why add this? Won't the current volume always be 0 after dropping a tip?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1934,52 +1938,48 @@ async def get_tip_presence_status(self, mount: OT3Mount) -> bool:
...
                 # home gear motor if position not known
                 await self.home_gear_motors()
             pipette_axis = Axis.of_main_tool_actuator(mount)
             gear_origin_float = axis_convert(self._backend.gear_motor_position, 0.0)[
                 pipette_axis
             ]
-            clamp_move_target = pipette_spec.pick_up_distance
-            clamp_moves = self._build_moves(
-                {Axis.Q: gear_origin_float}, {Axis.Q: clamp_move_target}
+            move_targets = []
```
sfoster1: `move_targets = [MoveTarget.build(position={Axis.Q: move_segment.distance}, max_speed=move_segment.speed) for move_segment in pipette_spec]` ?

I generally like using list comprehensions to make a transformation like this look like an assignment, rather than looking like mutating something

sfoster1 (changes requested):
Comments inline. Two big things
- We definitely are not getting the currents correct, still. During a force press pickup, the _z current_ is what defines the tip attach force and what we need to control during the downward stroke. During a force tip drop, the plunger current defines the applied force and what we need to control during the plunger downward stroke
- We're doing this thing where we decide if the pipette is a 96 twice, separately, in ot3api and in the pipette handler. We should do it in only one of those places and I think the right place is the ot3api, by gantry load. Then, let's make the functions in the pipette handler separate - one for 96, one for low throughput. One fewer place to change if we ever need to.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -723,23 +711,41 @@ def build_one_shake() -> List[Tuple[top_types.Point, Optional[float]]]:
 
         return []
 
-    def plan_check_pick_up_tip(
+    def plan_ht_pick_up_tip(self, mount) -> TipActionSpec:
```
sfoster1:
```suggestion
    def plan_ht_pick_up_tip(self, mount: OT3Mount) -> TipActionSpec:
```
? 
heck it might not even need the mount

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -752,71 +758,41 @@ def add_tip_to_instr() -> None:
...
+                )
+                # move nozzle back up
+                backup_dist = -press_dist
+                press_moves.append(
+                    TipActionMoveSpec(
+                        distance=backup_dist,
                         speed=pick_up_speed,
-                        relative_down=top_types.Point(0, 0, press_dist),
-                        relative_up=top_types.Point(0, 0, backup_dist),
+                        currents=None,
```
sfoster1, line 788: probably wants to set the default run current here

sfoster1 (approved):
Okay, I'm liking this a bit more, but I think we need to establish some test criteria. Why don't we say:

- Pick up a tip on a single and multi and check the canbus logs to make sure we're setting the current from config
- Drop a tip on a single and multi and check the canbus logs to make sure we're setting the current from config
- Check that 96 pickup and drop works with the segmented moves

I'm most concerned about LT regressions as you can probably tell.

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
caila-marashaj: I'm using the `None` to pass into `backend.motor_current`, which will set all the axes to their default run currents if it gets a `None`

ahiuchingau (approved):

PR #13670, opened 2023-09-27
fix(g-code-testing): Update g code testing for new stop behavior

sfoster1:
Updates our cached gcode regression tests for new stop behavior on the ot-2.


Laura-Danielle (approved):

SyntaxColoring (approved):
Thank you!

PR #13671, opened 2023-09-27
fix(app): filter LPC offsets for duplicates on ODD

smb2268:
# Overview
The "View Current Offsets" modal on the LPC step of protocol setup should be filtered on ODD the same way it is filtered on desktop

# Test Plan
1. Clone a run that has LPC offsets
2. Rerun LPC and change some of the offsets
3. Look at the applied offsets on protocol setup on the desktop app and on the ODD and see that there are the same number on each and no duplicates printed for the same labware/location combination

# Changelog
Filter offsets for ODD modal the same way we do on the desktop app

# Review requests
Run through test plan

# Risk assessment
Low


mjhuff (approved):
tested per instructions and lgtm!

PR #13674, opened 2023-09-27
feat(protocol-designer, components): allow trash to be movable

jerader:
closes RAUT-758 and RAUT-753 and RAUT-719

# Overview

This PR allows the trash to be movable for Flex protocols and also wires up a modal for editing the waste chute. Also, a user is required to select either a waste chute or a trash bin. NOTE: no trash/waste chute is not supported until post PD 7.1 since return tip is currently out of scope, you can only drop tip to a trash bin or waste chute.

# Test Plan

- Turn the deck configuration FF flag on.
- Make a Flex protocol and don't add the trash bin but add a waste chute. Create the protocol
- in the Additional Items section, see that the Trash bin is selectable. (I am stubbing in a gripper asset for now). Click add and see the slots available in the modal. Select a slot and click save.
- try editing and removing the trash slot
- mess around with adding/deleting the trash bin and waste chute, make sure the `remove` button disables with a tooltip appropriately. A user is required to add one or the other for not since return tip to tiprack will not be supported in 7.1.
- when you have the trash in a slot, go to the deck map and see that it is rendered in the correct slot that you selected
- look at the module edit modals and the waste chute modal, i changed the layout to match the trash modal. Make sure the behavior works as normal and you can edit/add/delete modules. Note that the TC GEN2 slot only highlights B1, I will go back later to fix that.
- now, go create an OT-2 protocol, see that the edit modules modal works as expected, check the TC modal and see that all 4 slots are highlighted.

# Changelog

- make `setSelectedLocation` prop optional in `DeckLocationSelect` and add an optional `theme` prop that defaults to `default` with the purple theme
- allow modules to be placed in slot A3 if they support it
- create Trash modal and update the edit modules modal to have the same design
- extend `AdditionalItemsRow` to support trash bin
- add a disabled reason in `ModuleAndOtherTile` with a tooltip requiring you to select a trash or waste chute
- update and add tests

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/modules/TrashModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+        deleteContainer({
+          labwareId: trashBinId,
+        })
+      )
+      dispatch(
+        createContainer({
+          labwareDefURI: FLEX_TRASH_DEF_URI,
+          slot: values.selectedSlot,
+        })
+      )
```
jerader, line 139: we don't have a `editContainer` action already like we do for modules and I figured since we'd only want to edit the trash bin (which is the only editable container) it made sense to just stick with `deleteContainer` + `createContainer`.

koji:
reduced the images' size a little bit.
[images.zip](https://github.com/Opentrons/opentrons/files/12766447/images.zip)

one quick question about images
One is .png and the others are .jpg. Don't we need to unify the image format? (.png no background, jpg white background)

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -72,9 +75,21 @@ export function ModulesAndOtherTile(props: WizardTileProps): JSX.Element {
...
   )
+  const [targetProps, tooltipProps] = useHoverTooltip()
+  const enableDeckModification = useSelector(
+    featureFlagSelectors.getEnableDeckModification
+  )
+  const hasATrash =
+    robotType === FLEX_ROBOT_TYPE && enableDeckModification
+      ? values.additionalEquipment.includes('wasteChute') ||
+        values.additionalEquipment.includes('trashBin')
+      : true
```
koji, line 90: one quick question

Could you tell me when hasATrash is `false`?

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -72,9 +75,21 @@ export function ModulesAndOtherTile(props: WizardTileProps): JSX.Element {
...
+
+  const buttonDisabled = !hasATrash
```
koji: Seems that `hasATrash` is only for buttonDisabled.
Maybe combine them?

protocol-designer/src/components/modals/EditModulesModal/ConnectedSlotMap.tsx
```diff
@@ -7,15 +7,20 @@ import styles from './EditModules.css'
 interface ConnectedSlotMapProps {
   fieldName: string
   robotType: RobotType
+  modal?: boolean
```
koji: nit
maybe `isModal` something since `modal` is a little bit vague.

protocol-designer/src/components/modals/EditModulesModal/index.tsx
```diff
@@ -282,73 +296,117 @@ const EditModulesModalComponent = (
...
-                      tabIndex={1}
-                    />
-                  </FormGroup>
-                </div>
-
-                <ConnectedSlotMap
-                  fieldName="selectedSlot"
-                  robotType={robotType}
+    <ModalShell width="48rem">
+      <Box marginTop={SPACING.spacing32} paddingX={SPACING.spacing32}>
```
koji: If the following works

```suggestion
    <ModalShell width="48rem" paddingTop={SPACING.spacing32}>
      <Box paddingX={SPACING.spacing32}>
```

Recently I read this article.
https://mxstbr.com/thoughts/margin/

protocol-designer/src/components/modules/TrashModal.tsx
```diff
@@ -0,0 +1,169 @@
...
+      <ModalShell width="48rem">
+        <Box marginTop={SPACING.spacing32} paddingX={SPACING.spacing32}>
+          <Text as="h2">
+            {i18n.t(`modules.additional_equipment_display_names.${trashName}`)}
+          </Text>
+        </Box>
+        <TrashModalComponent
+          onCloseClick={onCloseClick}
+          trashName={trashName}
+          trashBinSlot={trashBinSlot}
```
koji: Need this line? From L41, this line won't be used.

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
jerader: When you are on the module and other tile page, don't select the waste chute or trash bin. Make sure you've selected Flex and the feature flag is turned on.

protocol-designer/src/components/modals/EditModulesModal/ConnectedSlotMap.tsx
jerader: good call! it is super vague

protocol-designer/src/components/modules/TrashModal.tsx
jerader: oh ya i don't need it! thanks!

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
jerader: there is a chance that the sandbox isn't up to date for some reason so I recommend testing on the dev env

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
koji, line 1: just saw L59

The line might be the following since they aren't undefined from the interface definition?
```ts
formModule.onDeck && formModule.model !== null
```

```ts
export interface FormModule {
  onDeck: boolean
  model: ModuleModel | null
  slot: DeckSlot
}
```

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
koji: Thank you.
I could check the condition is changed by checking checkboxes.

koji (approved):
Followed the test plan and they worked as expected.
I just left one comment, but the changes look good to me.

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
jerader: good catch, that util was copied over from some legacy code - i'll fix it up!

PR #13679, opened 2023-09-28
fix(hardware): wrong val for dummy moves

ahiuchingau:
# Overview
We've been using the wrong values in the dummy moves that probably got added from a gnarly rebase, but this is not going to affect actual move behaviors.


caila-marashaj (approved):

fsinapi (approved):
yep

PR #13686, opened 2023-09-29
feat(shared-data): deck definition v4

TamarZanzouri:
# Overview

Closes RSS-343.

This PR adds both the schema and definitions for the new Deck Definition V4. This is a major change from V3, allowing deck configuration to be changed to support moveable trash, staging area slots, and the waste chute.

`orderedSlots` is no longer a property of the deck definition. Instead positional data is built up from three different properties in the definition. On the lowest level are `cutouts`, which define the 12 built in slots in the robot with absolute positional data. One layer above that is `cutoutFixtures`, which is slightly more abstract but defines both what `cutout` they can be loaded on, and which top-level `addressableArea` can be loaded onto that. And `addressableArea` is the highest level, which contains relative positional data, bounding boxes, compatible modules and drop tip and labware behavior and offsets.

# Test Plan

Right now, there is no actual way to test this on the robot since this has yet to be hooked up to Protocol Engine or Robot Server (which will be happening in future PRs). Coverage is instead handled by newly added schema validation tests, along with a test to ensure that the slot positions in `orderedSlots` from deck definition v3 match the new combination of positions in `addressableAreas` and `cutouts`.

# Changelog

- Added new deck schema v4
- Created new deck definitions for ot2, ot2 with short fixed trash and flex
- Added python shared data support

# Review requests

Naming of things along with a double check of all the positional data. Some of these will most likely be tweaked once further development commences, but we want these to be at least relatively accurate.

# Risk assessment

Low for now, since this isn't hooked up to anything yet.


shared-data/fixture/schemas/1.json
```diff
@@ -0,0 +1,149 @@
...
+          "description": "List of descriptions for a given fixture",
+          "items": {
+            "type": "string"
+          }
+        }
+      }
+    },
+    "brand": {
+      "$ref": "#/definitions/brandData",
+      "description": "Real-world fixfure that the definition is modeled from and/or compatible with"
```
b-cooper:
```suggestion
      "description": "Real-world fixture that the definition is modeled from and/or compatible with"
```

shared-data/deck/definitions/4/ot3_standard.json
```diff
@@ -0,0 +1,1269 @@
...
+    "addressableAreas": [
+      {
+        "id": "D1",
+        "areaType": "slot",
+        "matingSurfaceUnitVector": [-1, 1, -1],
+        "position": [0.0, 0.0, 0.0],
+        "boundingBox": {
+          "xDimension": 128.0,
+          "yDimension": 86.0,
+          "zDimension": 0
```
sfoster1, line 23: sort of tough to decide, but should the slot addressable area bounding boxes include the clips?

shared-data/deck/schemas/4.json
```diff
@@ -0,0 +1,424 @@
...
+          "description": "List of element children",
+          "items": { "$ref": "#/definitions/svgsonNode" }
+        },
+        "attributes": {
+          "type": "object",
+          "description": "Attributes for the element"
+        }
+      }
+    },
+    "cutoutSlot": {
```
sfoster1: I don't think we want this in here explicitly because these are as much a part of data as anything else. Putting this in the schema _requires_ that every deck definition fit its slot names into this enum. I think it's much better to leave this as a string.

The proper way to type ids of this kind is that the things that refer to them need valid references to something else in the document; I don't think jsonschema can do that, and in the absence of that capability I think we should keep the schema loose.

shared-data/deck/schemas/4.json
```diff
@@ -0,0 +1,424 @@
...
+        "9",
+        "10",
+        "11",
+        "12"
+      ]
+    },
+    "cutoutFixtureType": {
+      "type": "string",
+      "enum": [
+        "singleStandardSlot",
```
sfoster1: Same here - I do not think these should be controlled. In general, I think schemas should constrain values only if those values are discriminators in heterogeneous unions.

shared-data/deck/schemas/4.json
```diff
@@ -0,0 +1,424 @@
...
+              "id": {
+                "description": "Unique identifier for slot",
+                "type": "string"
+              },
+              "areaType": {
+                "description": "The type of deck item, defining allowed behavior.",
+                "type": "string",
+                "enum": ["slot", "movableTrash", "fixedTrash", "wasteChute"]
+              },
+              "position": {
```
sfoster1: can we make this `offsetFromCutoutFixture` or something

shared-data/deck/schemas/4.json
```diff
@@ -0,0 +1,424 @@
...
+                "description": "A human-readable nickname for this area e.g. \"Slot A1\" or \"Movable Trash\"",
+                "type": "string"
+              },
+              "compatibleModuleTypes": {
+                "description": "An array of module types that can be placed in this area.",
+                "type": "array",
+                "items": {
+                  "type": "string",
+                  "enum": [
+                    "magneticModuleType",
```
sfoster1: I think we should take these out (they were also defined like this in v3, and I think they shouldn't have been there either)

sfoster1 (changes requested):
Couple things about minutiae of slot definitions.

Also, I think that either we should make the inline SVG correct in this PR, or we should make it as empty an element as it can get so it's more obvious we haven't set it up yet.

shared-data/deck/schemas/4.json
SyntaxColoring: This will be moot if this is turned into a plain `string` instead of an `enum`, per @sfoster1's point, but: it looks like `singleStandardSlot` is unused, replaced by `singleLeftSlot`/`singleCenterSlot`/`singleRightSlot`?

shared-data/deck/schemas/4.json
SyntaxColoring: If we take these out, can we document where to find the valid strings?

shared-data/python/opentrons_shared_data/deck/__init__.py
```diff
@@ -36,6 +38,11 @@ class Offset(NamedTuple):
 }
 
 
+@overload
+def load(name: str, version: "DeckSchemaVersion4") -> "DeckDefinitionV4":
+    ...
+
+
```
SyntaxColoring: Here's a patch that fixes the `@overload` problems, I think.

https://github.com/Opentrons/opentrons/compare/feature-deck-def-v4...feature-deck-def-v4-fix-overload

If this doesn't work out, happy for it to be worked around with a separate function or something, like you said.

SyntaxColoring (commented):
Looks good to me, pending @sfoster1's comments and the `@overload` thing you mentioned in-person. Thank you!

shared-data/deck/definitions/4/ot3_standard.json
sfoster1: decide: no

sfoster1 (approved):
All right, looks great to me! Thanks!

SyntaxColoring (approved):
🛹

PR #13691, opened 2023-10-02
fix(app): 96-channel detach probe video

ncdiehl11:
# Overview

The detach probe module wizard step video is currently rendering the wrong animation. Here, I update to the correct animation.

# Test Plan

- Verify that during module calibration with a 96-channel pipette, the detach probe step shows the proper animation

# Risk assessment

Low


koji:
Should this PR go to `edge` instead of 7.1.0? (not sure 7.1.0 is existing right now)

PR #13692, opened 2023-10-02
docs(api): expand `Well.bottom()` warning

ecormany:
# Overview

Elaborate upon the existing warning for `Well.bottom()` on the Labware and Deck Positions page. We previously only warned against negative values. Testing has shown that values below 1 may cause problems on Flex, and the warning now (gently) cautions against them. 

# Test Plan

Check the [sandbox](http://sandbox.docs.opentrons.com/docs-overpressure-warning/v2/robot_position.html#bottom).

# Changelog

Updated warning text in `robot_position.rst`

# Review requests

Is this guidance good? Do we need to add this warning in other places (e.g., docstring)?

# Risk assessment

none, docs

### DIFF
--- a/api/docs/v2/robot_position.rst
+++ b/api/docs/v2/robot_position.rst
@@ -58,7 +58,11 @@ This is a good position for :ref:`aspirating liquid <new-aspirate>` or an activi
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks correct to me!

jwwojak (approved):
LGTM.

I got a bit of a chuckle from this:

> The protocol will continue even after a collision.

PR #13695, opened 2023-10-02
fix(app): nicknames on modules in protocol setup

jerader:
closes RAUT-768

# Overview

Previously, we were grabbing the display name instead of the nick name. it is very confusing since they are both key "displayName" and lots of utils refer to both as "displayName"

# Test Plan

upload a protocol with a nickname on a labware on deck, on adapter on deck, on module, and on adapter on module on deck. See that they render properly in labware and liquid setup. Also upload a protocol without any nicknames and make sure it works as expected showing the displayname.

see nicknames (one -> four):
Screen Shot 2023-10-02 at 2 00 24 PM

# Changelog

- change to get display name from params instead of the definition metadata

# Review requests

see test plan

# Risk assessment

low


koji (approved):
The changes look good to me.
Tested the branch with a couple of nick names and they were displayed as expected.

PR #13696, opened 2023-10-02
fix(robot-server): mutable configs for attached

sfoster1:
When something hits /settings/pipettes, we load known mutable configuration overrides. So if you don't have any for the attached pipettes, we don't return anything.

The thing is, what we actually want is to return all of the mutable configurations, with overrides applied. So now we do.

Closes RET-1376

## Testing
On an OT-2, attach a pipette that doesn't have overrides and `GET /settings/pipettes` (or view the pipette settings in the opentrons app).

### DIFF
--- a/robot-server/robot_server/service/legacy/routers/settings.py
+++ b/robot-server/robot_server/service/legacy/routers/settings.py
@@ -292,13 +292,15 @@ async def get_pipette_settings(
         get_opentrons_path("pipette_config_overrides_dir")
     ):
         # Have to convert to dict using by_alias due to bug in fastapi
-        res[pipette_id] = _pipette_settings_from_id(
+        res[pipette_id] = _pipette_settings_from_known_id(
             pipette_id,
         )
     for dct in attached_pipettes.values():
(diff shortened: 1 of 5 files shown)


robot-server/robot_server/service/legacy/routers/settings.py
```diff
@@ -376,3 +387,35 @@ def _pipette_settings_from_config(pipette_id: str) -> PipetteSettings:
...
+    :return: PipetteSettings object
+    """
+    mutable_configs = mutable_configurations.list_mutable_configs(
+        pipette_serial_number=pipette_id,
+        pipette_override_path=get_opentrons_path("pipette_config_overrides_dir"),
+    )
+    return _pipette_settings_from_mutable_configs(mutable_configs)
+
+
+def _pipette_settings_from_pipette_dict(
```
Laura-Danielle: could we call this something more like `_pipette_settings_from_pipette_default` and/or `_pipette_settings_from_default`?

Laura-Danielle (approved):
Changes look good, just one small name quibble.

Issue #7579, opened 2021-04-01
bug: from_center_cartesian needs more guidance to be used in complex commands


alexjoel42:
# Overview
The `distribute()` and `transfer()` commands (likely all of them) will dispense at slot 10 instead of the intended location if the well is using from_center_cartesian incorrectly (without warning). This will not be recorded in the run log either. 

This looks like and is likely related to bug [5106](https://github.com/Opentrons/opentrons/issues/5106) but unlike that bug which simply will ignore invalid arguments, this one will cause the OT-2 to move somewhere randomly.

# Steps to reproduce
This protocol both simulates and appears in the run log as if nothing went wrong,  but moves to the wrong location without warning. 

```python
from opentrons import protocol_api, types

metadata = {'apiLevel': '2.8'}

def run(protocol: protocol_api.ProtocolContext):
    tr2 = protocol.load_labware('opentrons_96_tiprack_20ul', '1')
    p20 = protocol.load_instrument('p20_single_gen2', 'left', tip_racks=[tr2])
    reservoir = protocol.load_labware('nest_12_reservoir_15ml', location='2')
    well = protocol.load_labware('nest_96_wellplate_100ul_pcr_full_skirt', location='4')
    p20.distribute(10, reservoir['A1'], [well['A1'].center().move(well['A1'].from_center_cartesian(x=0.1, y=0.1, z=0.1))])
```
## What happens on step 4 

 
![Image from iOS (19)](https://user-images.githubusercontent.com/65437763/113355305-91ceb380-930e-11eb-88a3-bfcd7e0e840d.jpg)

![Image from iOS (20)](https://user-images.githubusercontent.com/65437763/113355318-95fad100-930e-11eb-8b9d-a2b4ff0adccf.jpg)

# Expected behavior
This should raise either `TypeError`, which is what Python normally raises when unexpected kwargs are provided (since it is passing a location type value as opposed to a well) . Alternatively, it should follow the behavior of #5106 and ignore this argument. 

# Ideal behavior
The ability for complex commands is to throw an error that this will occur. 

# Available workaround 
Use the correct `from_center_cartesian` formatting and define the destinations like so:

    center_cart = types.Location(well['A1'].from_center_cartesian(0, 0, -1), well["A1"])

```python
from opentrons import protocol_api, types

metadata = {'apiLevel': '2.8'}

def run(protocol: protocol_api.ProtocolContext):
[...]

PR #13697, opened 2023-10-02
docs(api): updating docstrings for the Well class 

ecormany:
# Overview

Updating docstrings for the Well class. Not trying to be revolutionary, just trying to improve things for readers of the API Reference.

Closes RTC-206.

# Test Plan

- Checked formatting of [API Reference in sandbox](http://sandbox.docs.opentrons.com/docstrings-well/v2/new_protocol_api.html#opentrons.protocol_api.Well).
- Verified any new claims about how code works with `opentrons_simulate` for API version 2.13.

# Changelog

- Updated basically every docstring in the Well class.
- Added a missing docstring for `Well.diameter`.

# Review requests

Looks good? Sounds right?

# Risk assessment

low, inline docstrings

### DIFF
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -229,8 +229,8 @@ class Well:
             position in absolute deck coordinates.
 
         .. note:: Even if the absolute values of ``x``, ``y``, and ``z`` are all less
-            than 1, applying the result of ``from_center_cartesian`` to the well's
-            center may produce a location outside the physical well. For example,
+            than 1, a location constructed from the well and the result of
+            ``from_center_cartesian`` may be outside of the physical well. For example,
             ``from_center_cartesian(0.9, 0.9, 0)`` would be outside of a cylindrical
             well, but inside a square well.
 
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -182,8 +182,8 @@ class Well:
     def bottom(self, z: float = 0.0) -> Location:
         """
         :param z: An offset on the z-axis, in mm. Positive offsets are higher and
-            negative offsets are lower. 
-        
+            negative offsets are lower.
+
         :return: A :py:class:`~opentrons.types.Location` corresponding to the
             absolute position of the bottom-center of the well, plus the ``z`` offset
             (if specified).
@@ -211,10 +211,10 @@ class Well:
 
         :param x: The fraction of the distance from the well's center to its edge
             along the x-axis. Negative values are to the left, and positive values
-            are to the right. 
+            are to the right.
         :param y: The fraction of the distance from the well's center to its edge
             along the y-axis. Negative values are to the front, and positive values
-            are to the back. 
+            are to the back.
         :param z: The fraction of the distance from the well's center to its edge
             along the x-axis. Negative values are down, and positive values are up.
 
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -68,10 +68,13 @@ class OutOfTipsError(Exception):
 
 class Well:
     """
-    The Well class represents a  single well in a :py:class:`Labware`
+    The Well class represents a single well in a :py:class:`Labware`. It provides parameters and functions for three major uses:
 
-    It provides functions to return positions used in operations on the well
-    such as :py:meth:`top`, :py:meth:`bottom`
+        - Calculating positions relative to the well. See :ref:`position-relative-labware` for details.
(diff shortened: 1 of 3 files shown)


api/src/opentrons/protocol_api/labware.py
jwwojak, line 1: The changes here are a great coincidence. There are some similar schema items that define the dimensions of circular and square/rectangular wells. I'll go back through my revisions and try to make descriptions similar where appropriate.

jwwojak (approved):
LGTM

api/src/opentrons/protocol_api/labware.py
```diff
@@ -160,54 +169,64 @@ def well_name(self) -> str:
...
-            the radius/width in the Y axis
-        :param z: a float in the range [-1.0, 1.0] for a percentage of half of
-            the height above/below the center
+        Specifies a :py:class:`~opentrons.types.Point` based on fractions of the
+        distance from the center of the well to the edge along each axis.
+
+        For example, ``from_center_cartesian(0, 0, 0.5)`` specifies a point with no
+        change in the x- or y-axis, and half of the distance from the center of
+        the well to its top along the z-axis. Use :py:meth:`.move` to construct a
+        :py:class:`~opentrons.types.Location` that a pipette can move to.
```
SyntaxColoring: What do you have in mind for "use `.move` to construct a `Location` that a pipette can move to"?

SyntaxColoring (commented):
Mostly looks correct to me. One thing I'm not sure about, though:

api/src/opentrons/protocol_api/labware.py
ecormany: This method returns a Point, which is just an unmoored vector. The previous docstring seemed to describe it as returning a Location (e.g., "the resulting position…inside of the well"). The point needs to be applied to the well's `.center()` to get the intuitive "position". 

As much as I try to avoid it, perhaps we need a short code snippet embedded in the ref here. Maybe something like:
```
distance_from_center = plate["A1"].from_center_cartesian(0, 0, 0.5) 
new_location = plate["A1"].center().move(distance_from_center)
pipette.move_to(new_location)
```

I, and I think a lot of users, might try:

```
pipette.move_to(plate["A1"].from_center_cartesian(0, 0, 0.5))
# this raises a type error because move_to requires a Well or Location
```

SyntaxColoring: I haven't tested this recently, but my understanding is that this:

```python
distance_from_center = plate["A1"].from_center_cartesian(0, 0, 0.5) 
```

...doesn't actually return the distance from center. I expect it to return the *absolute* coordinate instead.

So instead of this:

```python
location = plate["A1"].center().move(from_center_cartesian(...)
```

I expect the correct incantation to be:

```python
location = Location(plate["A1"], plate["A1"].from_center_cartesian(...))
```

SyntaxColoring: @alexjoel42 and I had a good discussion about this a while ago: https://github.com/Opentrons/opentrons/issues/7579

ecormany: This is not helped by the fact that in the run log, regardless of how large an offset is applied, any of these commands will say something like `Moving to A1 of well plate on 1`

api/src/opentrons/protocol_api/labware.py
```diff
@@ -204,10 +204,17 @@ def from_center_cartesian(self, x: float, y: float, z: float) -> Point:
...
-        the well to its top along the z-axis. Use :py:meth:`.move` to construct a
-        :py:class:`~opentrons.types.Location` that a pipette can move to.
+        For example, ``from_center_cartesian(0, 0, 0.5)`` specifies a point at the
+        well's center on the x- and y-axis, and half of the distance from the center of
+        the well to its top along the z-axis. To move the pipette to that location,
+        construct a :py:class:`~opentrons.types.Location` relative to the same well::
+
+            location = Location(
+                plate["A1"], plate["A1"].from_center_cartesian(0, 0, 0.5)
+            )
```
SyntaxColoring: Oops, I had the argument order swapped in my example. My bad.

```suggestion
            location = Location(
                plate["A1"].from_center_cartesian(0, 0, 0.5), plate["A1"]
            )
```

SyntaxColoring (approved):
Looks great to me with the latest revisions. Thank you!

One fixup in the code snippet before you merge.

api/src/opentrons/protocol_api/labware.py
ecormany: It also has to be `types.Location`. Fixing now.

SyntaxColoring (approved):

PR #13701, opened 2023-10-02
fix(app,app-shell): undelete old robot update urls

sfoster1:
We removed the urls for the old OT-2 infrastructure, because we don't want them anymore. But because we removed them from the app config, and the app doesn't downgrade configs, then if you
- ran 7.0, the config gets deleted
- run 6.3.1 or previous, the config never gets restored and the app just doesn't know where to get robot system updates and therefore acts as if there are none. That in turn means you can't downgrade your robot through the app, you have to manually update from zip.

The gross but only solution is to change the config back to get those values to be present again when you downgrade. This leaves a historical hole - someone who ran 7.0.0 and downgraded to 6.3.1 would still have that behavior - but it's fixed going forward, and those affected can just remove their app config.

## Testing
- Run this branch through `make -C app dev` (just has to migrate the config, don't have to do anything with it)
- Run 6.3.1 and check that you still get downgrade banners on robots. You could make dev it, but also the installers are on opentrons.com/ot-app

### DIFF
--- a/app-shell-odd/src/config/__fixtures__/index.ts
+++ b/app-shell-odd/src/config/__fixtures__/index.ts
@@ -7,7 +7,7 @@ import type {
   ConfigV17,
   ConfigV18,
   ConfigV19,
-  ConfigV20
+  ConfigV20,
 } from '@opentrons/app/src/redux/config/types'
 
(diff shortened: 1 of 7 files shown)


b-cooper (approved):
Changes look sound to me ♻️

PR #13702, opened 2023-10-02
feat(app): add loading state to LPC when applying offsets

shlokamin:
# Overview

This PR adds a loading state to LPC when applying offsets and deleting the maintenance run. This avoids a bug where LPC appears unresponsive, and prevents users from submitting offsets over and over again.

closes RQA-1220

# Test Plan

- Tested on a desktop app + ODD

# Changelog

- Add loading state to LPC when applying offsets

# Review requests

- go through LPC on an OT-2 using the desktop app
- go through LPC on a flex using the ODD (i added a loading state there too)

# Risk assessment
Low

### DIFF
--- a/app/src/atoms/buttons/SmallButton.tsx
+++ b/app/src/atoms/buttons/SmallButton.tsx
@@ -30,8 +30,8 @@ interface SmallButtonProps extends StyleProps {
   onClick: React.MouseEventHandler
   buttonType?: SmallButtonTypes
   buttonText: React.ReactNode
-  iconPlacement?: IconPlacement
-  iconName?: IconName
+  iconPlacement?: IconPlacement | null
+  iconName?: IconName | null
   buttonCategory?: ButtonCategory // if not specified, it will be 'default'
   disabled?: boolean
 }
--- a/app/src/organisms/LabwarePositionCheck/LabwarePositionCheckComponent.tsx
+++ b/app/src/organisms/LabwarePositionCheck/LabwarePositionCheckComponent.tsx
@@ -49,6 +49,7 @@ interface LabwarePositionCheckModalProps {
   protocolName: string
   caughtError?: Error
   setMaintenanceRunId: (id: string | null) => void
+  isDeletingMaintenanceRun: boolean
 }
 
 export const LabwarePositionCheckComponent = (
@@ -62,6 +63,7 @@ export const LabwarePositionCheckComponent = (
     maintenanceRunId,
     setMaintenanceRunId,
     protocolName,
+    isDeletingMaintenanceRun,
   } = props
   const { t } = useTranslation(['labware_position_check', 'shared'])
   const isOnDevice = useSelector(getIsOnDevice)
@@ -102,6 +104,9 @@ export const LabwarePositionCheckComponent = (
   ])
 
   const [fatalError, setFatalError] = React.useState<string | null>(null)
+  const [isApplyingOffsets, setIsApplyingOffsets] = React.useState<boolean>(
+    false
+  )
   const [
     { workingOffsets, tipPickUpOffset },
     registerPosition,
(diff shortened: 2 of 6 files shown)


app/src/atoms/buttons/SmallButton.tsx
```diff
@@ -30,8 +30,8 @@ interface SmallButtonProps extends StyleProps {
   onClick: React.MouseEventHandler
   buttonType?: SmallButtonTypes
   buttonText: React.ReactNode
-  iconPlacement?: IconPlacement
-  iconName?: IconName
+  iconPlacement?: IconPlacement | null
+  iconName?: IconName | null
```
shlokamin, line 34: added `null` here so we don't have to pass in `undefined` when conditionally rendering icons

app/src/organisms/LabwarePositionCheck/LabwarePositionCheckComponent.tsx
```diff
@@ -102,6 +104,9 @@ export const LabwarePositionCheckComponent = (
   ])
 
   const [fatalError, setFatalError] = React.useState<string | null>(null)
+  const [isApplyingOffsets, setIsApplyingOffsets] = React.useState<boolean>(
+    false
+  )
```
shlokamin, line 109: another way to do this is to create a wrapper around `useCreateLabwareOffsetMutation` that applies multiple offsets and returns a loading state for the duration of the entire array of offsets being applied. the reason i can't just use the `isLoading` boolean that comes back from `useCreateLabwareOffsetMutation` is because it would flicker between each request to apply individual offsets

app/src/atoms/buttons/SmallButton.tsx
jerader, line 34: did you add `null` to these types so you could avoid explicitly saying `undefined` in the component? Feels like theres a cleaner way to approach this.

app/src/organisms/LabwarePositionCheck/ResultsSummary.tsx
```diff
@@ -168,6 +174,9 @@ export const ResultsSummary = (
           alignSelf={ALIGN_FLEX_END}
           onClick={() => handleApplyOffsets(offsetsToApply)}
           buttonText={i18n.format(t('apply_offsets'), 'capitalize')}
+          iconName={isSubmittingAndClosing ? 'ot-spinner' : null}
+          iconPlacement={isSubmittingAndClosing ? 'startIcon' : null}
```
jerader, line 178: maybe you can do this to avoid typing the props as null:

```suggestion
          iconName={isSubmittingAndClosing && 'ot-spinner'}
          iconPlacement={isSubmittingAndClosing && 'startIcon'}
```

app/src/organisms/LabwarePositionCheck/__tests__/ResultsSummary.test.tsx
```diff
@@ -33,6 +33,8 @@ describe('ResultsSummary', () => {
       protocolData: mockCompletedAnalysis,
       workingOffsets: mockWorkingOffsets,
       existingOffsets: mockExistingOffsets,
+      isApplyingOffsets: false,
+      isDeletingMaintenanceRun: false,
```
jerader, line 37: there can be more test coverage in this component but this is okay.

jerader (approved):
works as expected. I have a suggestion for not adding `null` to the `iconPlacement` and `iconName` types in the comments.

app/src/organisms/LabwarePositionCheck/ResultsSummary.tsx
shlokamin: hmm that won't either because `isSubmittingAndClosing && 'ot-spinner'` is a logical expression that will either return `ot-spinner` (valid) or `false` (not valid)

app/src/organisms/LabwarePositionCheck/__tests__/ResultsSummary.test.tsx
shlokamin: nahh you're right i was honestly being lazy, thanks for checking me. just added two cases to handle when offsets being applied and the maintenance run being deleted

PR #13703, opened 2023-10-02
feat(api): check for tip presence on 96 channel pipette

caila-marashaj:
## Overview
This pr adds a function the api that will return the tip presence status for any pipette. If the pipette in question is a 96 channel, this code will move the gear motors downward a distance specified in the pipette configs by `tip_presence_check_distance` before asking for the status. Something to note is that the tip presence responses returned by the 96 channel pipette are reversed coming from the firmware side compared to what we expect, so this function will invert the response before returning if the pipette is a 96 channel.


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1894,6 +1897,35 @@ async def blow_out(
...
+            )
+            await self._backend.tip_action(moves=clamp_moves[0])
+        try:
+            tip_status = await self._backend.get_tip_present_state(
+                mount=checked_mount, expect_multiple_responses=high_throughput
+            )
+            # FIXME: tip presence responses from the 96 channel hardware need to be inverted
+            if high_throughput:
+                tip_status = not tip_status
+        except UnmatchedTipPresenceStates:
```
sfoster1: as it is, this try/except isn't really doing anything, but I think we want it to: specifically, we want to move the tip motors back to their neutral positions if we get either no error, or an `UnamtchedTipPresenceStates` (since that's sort of an "expected" error that doesn't indicate that something larger is wrong). If we get another error, we can just pass it through.

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1894,6 +1897,35 @@ async def blow_out(
...
             blowout_spec.instr.ready_to_aspirate = False
 
+    async def get_tip_presence_status(self, mount: OT3Mount) -> bool:
+        """
+        Check tip presence status. If a high throughput pipette is present,
+        move the tip motors down before checking the sensor status.
+        """
+        checked_mount = OT3Mount.from_mount(mount)
+        high_throughput = self._gantry_load == GantryLoad.HIGH_THROUGHPUT
+        if high_throughput:
```
sfoster1, line 1907: is it safe to do this if we have a rack of tips clamped? will it cause the seal on the tips to get messed up if we drop and come back?

sfoster1 (changes requested):
I think we need to get the clamp back to fully-clamped after the action.

I'm also a little worried about what happens if we do this when we have a rack of tips clamped in place - does it mess up the seals? Does the sensor work the same?

api/tests/opentrons/hardware_control/backends/test_ot3_utils.py
```diff
@@ -84,6 +84,7 @@ def test_filter_zero_duration_step() -> None:
         moves=moves,
         present_nodes=present_nodes,
     )
+    breakpoint()
```
sfoster1: accidentally checked in?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1895,6 +1897,47 @@ async def blow_out(
...
+            if tip_motor_pos_float < tip_presence_check_target:
+                clamp_moves = self._build_moves(
+                    {Axis.Q: tip_motor_pos_float}, {Axis.Q: tip_presence_check_target}
+                )
+                await self._backend.tip_action(moves=clamp_moves[0])
+        try:
+            tip_status = await self._backend.get_tip_present_state(
+                mount=checked_mount, expect_multiple_responses=high_throughput
+            )
+            # FIXME: tip presence responses from the 96 channel hardware need to be inverted
```
sfoster1: can we just fix this, right now, in firmware?

sfoster1 (changes requested):
Couple little changes

api/src/opentrons/hardware_control/ot3api.py
caila-marashaj: yeah I can just invert that in firmware

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1895,6 +1897,44 @@ async def blow_out(
...
+            if tip_motor_pos_float < tip_presence_check_target:
+                clamp_moves = self._build_moves(
+                    {Axis.Q: tip_motor_pos_float}, {Axis.Q: tip_presence_check_target}
+                )
+                await self._backend.tip_action(moves=clamp_moves[0])
+        try:
+            tip_status = await self._backend.get_tip_present_state(
+                mount=checked_mount, expect_multiple_responses=high_throughput
+            )
+        except Exception:
```
sfoster1: we should remove this try/except entirely, it's not doing anything except adding another layer of indentation

sfoster1 (approved):
Looks good but let's remove that try/except

api/tests/opentrons/hardware_control/backends/test_ot3_utils.py
```diff
@@ -72,7 +72,7 @@ def test_filter_zero_duration_step() -> None:
         Axis.P_R: 0,
     }
     moves = [Move.build_dummy([Axis.X, Axis.Y, Axis.Z_L, Axis.Z_R, Axis.P_L])]
-    for block in (moves[0].blocks[0], moves[0].blocks[1]):
+    for block in moves[0].blocks:
```
fsinapi, line 75: Is this change related to the PR?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1895,6 +1897,41 @@ async def blow_out(
...
+                clamp_moves = self._build_moves(
+                    {Axis.Q: tip_motor_pos_float}, {Axis.Q: tip_presence_check_target}
+                )
+                await self._backend.tip_action(moves=clamp_moves[0])
+        tip_status = await self._backend.get_tip_present_state(
+            mount=checked_mount, expect_multiple_responses=high_throughput
+        )
+
+        # return tip motors to neutral position
+        await self.home_gear_motors()
```
fsinapi: Assuming that this function is supposed to work on low-throughput pipettes as well, we should only `home_gear_motors()` if `high_throughput == True`

fsinapi (dismissed):
Coupla small things

PR #13707, opened 2023-10-03
test(robot-server): Fix race condition in integration test

SyntaxColoring:
# Overview

This fixes a test that was flaky because of a race condition.

We were waiting for a run to succeed by:

1. Polling until its status indicated "done."
2. Checking that the final status was `succeeded`.

But we would erroneously proceed from step 1 to step 2 when the run transitioned from `running` to `finishing`, and `finishing` is of course not `succeeded`, so the test would sometimes fail depending on timing.

# Test Plan

# Risk assessment

No risk.

### DIFF
--- a/robot-server/tests/integration/http_api/persistence/test_compatibility.py
+++ b/robot-server/tests/integration/http_api/persistence/test_compatibility.py
@@ -1,4 +1,3 @@
-import asyncio
 from dataclasses import dataclass, field
 from pathlib import Path
 from shutil import copytree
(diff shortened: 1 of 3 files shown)


CaseyBatten:
Looks good to me, appropriate change to increase the scope of that check. Thinking on potential alternative cases, would it be beneficial to adjust the check to include "stop-requested" as a third valid pre-end state on this check? so this:
```
if latest_status not in {"running", "stop-requested", "finishing"}:
            return latest_status  # type: ignore[no-any-return]
```

CaseyBatten (approved):

SyntaxColoring:
> Looks good to me, appropriate change to increase the scope of that check. Thinking on potential alternative cases, would it be beneficial to adjust the check to include "stop-requested" as a third valid pre-end state on this check? so this:
> 
> ```
> if latest_status not in {"running", "stop-requested", "finishing"}:
>             return latest_status  # type: ignore[no-any-return]
> ```

That would be more appropriate, yes.

And there's a very similar polling loop in a different test file:

https://github.com/Opentrons/opentrons/blob/cd064e66d178a58669eb5bd8e939e33cd4cbc3b1/robot-server/tests/integration/http_api/runs/test_labware_offsets_on_compatible_modules.py#L35-L49

That one takes the approach of listing out the "done" cases, instead of the "not done" cases.

This should be a shared function somewhere. I will do that in this PR.


-----

Edit: Done.

SyntaxColoring:
Got a 👍 from @CaseyBatten in person.

PR #13708, opened 2023-10-03
fix(app): update the condition to display the banner for open door status

koji:
# Overview
Update the condition to display the banner for open door status for the turn-off case.

close RQA-1672 partially
Need RSS side fix to solve this issue completely 

# Test Plan

# Changelog

# Review requests

# Risk assessment


koji:
- [x] fix test errors
- [ ] test with rss's fix

PR #13709, opened 2023-10-03
fix(api): Support simulating PAPIv≥2.14

SyntaxColoring:
# Overview

Closes RSS-322. Lets you run new Python protocols through `opentrons_simulate`, `opentrons.simulate.simulate()`, and `opentrons.simulate.get_protocol_api()`.

Also fixes a straggling instance of RSS-156, incorrectly using the machine's current deck type in simulation, instead of the correct deck type to match the protocol.

# Test Plan

See [this document](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3836739643/Manual+test+plan+for+opentrons+simulate+and+friends+in+v7.0.1).

# Changelog

Notable changes:

* Rework `simulate.py` to support PAPIv≥2.14 protocols, via Protocol Engine. This rework mirrors what we did to `execute.py` in #12970.
* Rework the `adapt_protocol_source()` helper to cope with protocols for which we don't know the filename. This is necessary to support `opentrons.simulate.simulate()`, whose `filename` argument is optional.

  Unfortunately, this requires modifications to `opentrons.protocols.parse` to make the filename more pass-through, so `adapt_protocol_source()` can tell when it wasn't provided.

Plus various smaller refactors for type safety, deduplication, and consistency.

# Code review requests

I suggest code-reviewing this in two ways:

1. Read the latest `simulate.py` file side-by-side with the latest `execute.py` file. They should mirror each other almost exactly.
    * Known discrepancy: They have minor differences because of differences in their interfaces. For example, the stuff in `simulate.py` *returns* the run log, whereas the stuff in `execute.py` sends run log events to a callback.
    * Known discrepancy: `execute.py` has a global hardware controller and global `ProtocolEngine`s, whereas `simulate.py` only has the global `ProtocolEngine`s.
2. To get a more semantic sense of what changed, go commit-by-commit. Many of the commits are trivial refactors and can be glossed over. The commit messages should mark when this is the case.

Some special things to watch out for:

* When rewriting `simulate.py` to make it look like `execute.py`, make sure I didn't accidentally drop anything unique to the old `simulate.py` that was important.
* Double-check my work on `adapt_protocol_source()` and `opentrons.protocols.parse`. 

# Risk assessment

Medium.

This touches some subtle parts of `opentrons.protocols.parse` and `opentrons.protocols.execute`. The touches are light, but those areas are critical to protocol execution.

It also touches the simulation of PAPIv≤2.13 and JSONv≤5 protocols, so we need regression testing on those. See the test plan.

### DIFF
--- a/api/src/opentrons/protocols/parse.py
+++ b/api/src/opentrons/protocols/parse.py
@@ -217,11 +217,16 @@ def _parse_python(
     extra_labware: Optional[Dict[str, "LabwareDefinition"]] = None,
 ) -> PythonProtocol:
     """Parse a protocol known or at least suspected to be python"""
-    filename_checked = filename or "<protocol>"
-    if filename_checked.endswith(".zip"):
+    if filename is None:
+        # The fallback "<protocol>" needs to match what opentrons.protocols.execution.execute_python
+        # looks for when it extracts tracebacks.
+        ast_filename = "<protocol>"
+    elif filename.endswith(".zip"):
+        # The extension ".zip" and the fallback "protocol.ot2.py" need to match what
+        # opentrons.protocols.execution.execute_python looks for when it extracts tracebacks.
         ast_filename = "protocol.ot2.py"
     else:
-        ast_filename = filename_checked
+        ast_filename = filename
 
     # todo(mm, 2021-09-13): By default, ast.parse will inherit compiler options
     # and future features from this module. This may not be appropriate.
@@ -244,7 +249,7 @@ def _parse_python(
 
     static_info = _extract_static_python_info(parsed)
     protocol = compile(parsed, filename=ast_filename, mode="exec")
-    version = _get_version(static_info, parsed, filename_checked)
+    version = _get_version(static_info, parsed, ast_filename)
     robot_type = _robot_type_from_static_python_info(static_info)
 
     if version >= APIVersion(2, 0):
@@ -257,7 +262,7 @@ def _parse_python(
 
     result = PythonProtocol(
         text=protocol_contents,
-        filename=getattr(protocol, "co_filename", "<protocol>"),
+        filename=filename,
         contents=protocol,
         metadata=static_info.metadata,
         api_level=version,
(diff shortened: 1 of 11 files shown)


SyntaxColoring:
Notes to self:

* [x] Make sure all entry points are consistently validating MAX_SUPPORTED_VERSION.
* [ ] ~~Add tests for, and clean up, adapt_protocol_source()~~
  * This is good enough for now.
* [x] Use deck type for declared robot type, not for current machine
* [x] Check fixme and todo comments
* [x] Resolve `simulate()` with no filename not working
* [x] Fix error message printing
* [x] Flip on tests that were commented out

api/src/opentrons/protocols/parse.py
```diff
@@ -257,7 +262,7 @@ def _parse_python(
 
     result = PythonProtocol(
         text=protocol_contents,
-        filename=getattr(protocol, "co_filename", "<protocol>"),
+        filename=filename,
```
SyntaxColoring, line 265: Falling back to `"<protocol>"` at this level was lossy. If we weren't given a filename, we want to preserve that information for the caller, by returning `None`, so they can choose their own fallback.

ecormany:
@SyntaxColoring If this will actually make it into the 7.0.1 release, can you remove these lines (or at least line 20) from the release notes as part of this PR? https://github.com/Opentrons/opentrons/blob/8a2ac3dfb4057aa971e114171c094c953e7662f4/api/release-notes.md?plain=1#L15-L21

This is in the base branch as of now.

CaseyBatten:
Running locally, so far so good, need to test standard and JSON protocols on a Flex, might I recommend a change here:
https://github.com/Opentrons/opentrons/blob/ca8da73d1f81e80d9bcb9e917726f90978219e9d/api/src/opentrons/simulate.py#L928

Should say something more like this to clarify support for the Flex:
```
parser = argparse.ArgumentParser(
        prog="opentrons_simulate", description="Simulate an OT-2/Flex protocol"
    )
```

I'll update again once I have run this on actual Flex hardware.

jbleon95 (approved):
Went through code and tested in a live review, looks good. Welcome back opentrons_simulate.

sanni-t (approved):
🙌 Great work! Changes make sense

CaseyBatten (approved):
Was capable of running multiple Flex python and JSONv5 protocols via the simulation CLI. Did not interrupt standard Flex operation.

PR #13710, opened 2023-10-03
fix(app): render labware on module correctly in moveLabware modal

jerader:
closes RAUT-772

# Overview

Fix labware moving away from on top of the module in move labware intervention modal

# Test Plan

Run a protocol with a move labware manually step from on top of a module to another location. See that the deck map in the intervention modal shows that there is no labware on top of the module.

# Changelog

- fix logic to not render the nested labware if it matches the `labwareId` in the move labware command

# Review requests

see test plan

# Risk assessment

low


koji (approved):
The changes look good to me.

PR #13711, opened 2023-10-03
chore(monorepo): add make test targets for frontend projects

shlokamin:
# Overview

This PR adds make targets for our frontend projects so developers don't need to use jest directly. I also swapped out our ci workflows to use these new make targets.

One behavior change worth noting: the app build test deploy workflow was running tests on components and shared data. i changed it so that it only runs tests on the app, plus its native packages (app shell, app shell odd, and discovery client)

# Changelog

-  Add make test targets for frontend projects

# Review requests

Make sure my changes to the CI workflows make sense. I specifically wanna make sure i didn't break codecov.

# Risk assessment

Low


Makefile
```diff
@@ -196,11 +199,17 @@ test-py: test-py-windows
 
 .PHONY: test-js
 test-js:
-	yarn jest \
-		--coverage=$(cover) \
-		--watch=$(watch) \
-		--updateSnapshot=$(updateSnapshot) \
-		--ci=$(if $(CI),true,false)
+	$(MAKE) -C $(API_CLIENT_DIR) test
```
sfoster1: I think you don't want to do this part because I think one top-level jest process is gonna be more efficient doing its own scheduling than a jest process per subdir.

sfoster1 (approved):
Heck yeah this looks really good

sfoster1 (approved):
Looks good to me, I just wish we could centralize some test and coverage opts

shlokamin:
> Looks good to me, I just wish we could centralize some test and coverage opts

im gonna merge this for now but yeah fair point. i guess we could `include` shared test + coverage opts from a shared place and use them in each of these projects

PR #13712, opened 2023-10-03
fix(app): update move gantry text in change pipette flow

shlokamin:
# Overview

This PR updates the pipette attach/detach flows to use updated copy when the gantry is moving.
closes RQA-1224

# Risk assessment

Low, just a copy change


smb2268 (approved):
lgtm!

PR #13714, opened 2023-10-03
fix(app): fix calibration health check result summary condition issue

koji:
# Overview
The issue was the condition was the opposite. Fix that issue and rename the variable name since the previous one didn't describe itself well.

close RQA-1736

# Test Plan
- calibrate deck, pipette offset, and tip length

# Changelog
- update ResultSummary component and rename the variable
- update ResultSummary tests

# Review requests

# Risk assessment
low


jerader (approved):
nice! 🐑

PR #13716, opened 2023-10-03
feat(protocol-designer): wire up loadFixture for waste chute in fileCreator

jerader:
closes RAUT-687 RAUT-723

# Overview

Now that `loadFixtureCreateCommand` type exists in v7 command types, I wired up creating the command in file creator and importing as well.

# Test Plan

- make sure the deck configuration FF is turned on and create a protocol for the Flex. Go through the form and add a waste chute. Export the protocol. Look at the json file and see that there is a `loadFixture` command with the proper params.
- Change the version from `7.0.0` to `7.1.0` in the json file (this is so the protocol doesn't go through a migration)
- import the protocol back into PD. See that the waste chute entity exists in the Additional Items section

# Changelog

- extend createSelector to support 15 selectors
- add loadFixture command support to `fileCreator` and the step form reducer
- fix test
- also wire up nicknames for the labware as the display name if it exists. apparently nicknames weren't being carried over to the json protocol this whole time 😅 

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/step-forms/reducers/index.ts
+++ b/protocol-designer/src/step-forms/reducers/index.ts
@@ -22,6 +22,7 @@ import {
   LoadFixtureCreateCommand,
   STAGING_AREA_LOAD_NAME,
   STANDARD_SLOT_LOAD_NAME,
+  TRASH_BIN_LOAD_NAME,
 } from '@opentrons/shared-data'
 import type { RootState as LabwareDefsRootState } from '../../labware-defs'
 import { rootReducer as labwareDefsRootReducer } from '../../labware-defs'
@@ -1332,7 +1333,8 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
           //  until it is supported
           if (
             loadName === STAGING_AREA_LOAD_NAME ||
-            loadName === STANDARD_SLOT_LOAD_NAME
+            loadName === STANDARD_SLOT_LOAD_NAME ||
+            loadName === TRASH_BIN_LOAD_NAME
           ) {
             return acc
           }
--- a/protocol-designer/src/step-forms/reducers/index.ts
+++ b/protocol-designer/src/step-forms/reducers/index.ts
@@ -1350,7 +1350,7 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
         {}
       )
       const hasGripper = gripperCommands.length > 0
-      const isOt3 = file.robot.model === FLEX_ROBOT_TYPE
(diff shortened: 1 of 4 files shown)


koji:
Is this expected behavior?
For pipette part, the wizard shows the selected item when clicking back button.

https://github.com/Opentrons/opentrons/assets/474225/6a670428-67a2-4e99-920a-60e1ecb10661

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1309,23 +1313,57 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
...
+        (command): command is LoadFixtureCreateCommand =>
+          command.commandType === 'loadFixture'
+      )
+      const fixtures = fixtureCommands.reduce(
+        (
+          acc: NormalizedAdditionalEquipmentById,
+          command: LoadFixtureCreateCommand
+        ) => {
+          const { fixtureId, loadName, location } = command.params
+          const id = fixtureId ?? ''
```
koji, line 1331: Do you think we can change `id` to something else?

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1309,23 +1313,57 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
...
+              id: id,
+              name: loadName,
+              location: location.cutout,
+            },
+          }
+        },
+        {}
+      )
+      const hasGripper = gripperCommands.length > 0
       const isOt3 = file.robot.model === FLEX_ROBOT_TYPE
```
koji: This is unrelated to this PR but maybe we can use isFlex instead of `isOt3`?
RSS won't change the robotType from OT-3 to Flex but in FE we avoid using OT-3 recently so it might be good for the future.

koji (approved):

jerader:
> Is this expected behavior? For pipette part, the wizard shows the selected item when clicking back button.
> 
>  Screen.Recording.2023-10-05.at.5.01.11.PM.mov

@koji oh good catch! I didn't notice. It should probably keep the staging slot behavior. Also I think you meant to post the video in my other PR 😅 https://github.com/Opentrons/opentrons/pull/13723

protocol-designer/src/step-forms/reducers/index.ts
jerader: i kept it as just `id` since the type also encompasses `gripper`. But maybe it can be changed in the future if it gets too confusing.

PR #13717, opened 2023-10-03
fix(app, app-shell): Add app layer robot update flow error handling

mjhuff:
Closes RQA-1562, RQA-1714, RQA-1715, RQA-1680, RQA-1536, RQA-1747

# Overview

This PR adds a layer of enforced consistency between robot update stages to reduce intermittent bugs that occur when updating and somewhat gracefully handle errors when they occur. Users are no longer indefinitely locked into update flows if the robot fails to respond - users may restart the update (except during the robot restart step) or exit the update if the app does not receive a robot update session event in 10 minutes (timer is debounced on new events).

The redux store now includes robot update session logic to delineate between robot updates initiated from a manual zip file vs. the S3 downloaded zip file, which is utilized when retrying a new update. We currently assume the S3 downloaded file is always correct, but users are prompted to select a new upload file if they previously selected an update file and the update failed.

This PR provides a general refactoring to update flows. Previously, update sessions were preserved in the redux store whenever an error occurred, leading to some erroneous assumptions about current robot update progress when a new update flow was launched - this PR now ensure a fresh session each time the flows are launched, and any sudden dismissal of the flows clears a session. Additionally, update flows are no longer conditionally rendered by the existence of a session/robot connection, which prevents flows suddenly closing (although this behavior still occurs in a few spots that will be addressed in a future PR). Lastly, progress bar logic has been made more scalable as more update step progress percentages will be added in future PRs (providing a more accurate & granular picture of update progress). 

A bonus refactor of the app software update flows is included to bring them in-line with Figma. 

NOTE: It is possible when uploading files manually to error out when restarting the update from within an existing update. It's non-blocking to the user (trying again AFTER the error works), but this will be addressed in a following PR. 

### Restarting an Update Via Zip File

https://github.com/Opentrons/opentrons/assets/64858653/eddcc0e9-40cd-4a18-9cc7-6603e5645136

Note that the update debounce interval was adjusted for this video.

# Test Plan
- Build this branch, launch it, then upload an older robot version via zip file to a robot (probably best to test on an OT-2, since they tend to be more error prone). Verify this works.
- Update the robot to the latest version via S3 download (clicking on a banner). Verify this works.
- Try reinstalling via Robot Settings > Advanced. Verify this works.
- Set the TIME_BEFORE_ALLOWING_EXIT_MS within RoboUppateProgressModal to something like 5000, verify that the restart logic works. Verify that restarting the update does NOT render if the robot is on the "restarting" step

# Changelog
- Added "retry the update" logic for robot updates.
- Refactored existing update flows to guard update transitions, reducing error occurrence.  

# Review requests
- Test on real robots please!

# Risk assessment
low-ish


mjhuff:
Bringing back the cleanup function - I initially removed it when debugging some OT-2 update issues (caused by fun Wifi issues - not related to the PR), but it's absolutely essential to prevent some infinite re-rendering bugs caused by suddenly closing update banners (which will be addressed separately in a follow-up PR).

mjhuff:
No clue what the lint issue is - my prettier doesn't seem to change anything when I run it on useRobotUpdateInfo. Will look into this.

sfoster1 (changes requested):
looks good to me based on current testing plus tuesday's!

sfoster1 (approved):
Wrong button sorry!

PR #13720, opened 2023-10-04
fix(api, robot-server, shared-data): Flex and OT2 door check disparity correction

CaseyBatten:
# Overview
This branch aims to fix the issue where OT2 would refuse to run with the door open. Full issue detailed in: https://opentrons.atlassian.net/browse/RSS-369

# Test Plan
Make push the changes to an OT2 and do the following:

1. In APP, disable "Pause protocol when robot door opens"
2. Run a Protocol with the door open, ensure process is not interrupted
3. After protocol ends, re-enable "Pause protocol when robot door opens"
4. Run protocol again, ensure process is interrupted.

Make push the changes to a Flex and do the following:
1. Begin to run a protocol
2. Open the door, ensure the Flex pauses and interrupts the process

### DIFF
--- a/api/src/opentrons/config/feature_flags.py
+++ b/api/src/opentrons/config/feature_flags.py
@@ -1,6 +1,5 @@
 from opentrons.config import advanced_settings as advs
 from opentrons_shared_data.robot.dev_types import RobotTypeEnum
-from opentrons_shared_data.robot.dev_types import RobotType
 
 
 def short_fixed_trash() -> bool:
@@ -21,17 +20,8 @@ def use_old_aspiration_functions() -> bool:
     )
 
 
-def enable_door_safety_switch(robot_type: RobotType) -> bool:
-    if robot_type == "OT-2 Standard":
-        return advs.get_setting_with_env_overload(
-            "enableDoorSafetySwitch", RobotTypeEnum.OT2
-        )
-    elif robot_type == "OT-3 Standard":
-        return advs.get_setting_with_env_overload(
-            "enableDoorSafetySwitch", RobotTypeEnum.FLEX
-        )
-    else:
-        raise KeyError("Invalid Robot Type during Door Saftey switch check")
+def enable_door_safety_switch(robot_type: RobotTypeEnum) -> bool:
+    return advs.get_setting_with_env_overload("enableDoorSafetySwitch", robot_type)
 
 
 def disable_fast_protocol_upload() -> bool:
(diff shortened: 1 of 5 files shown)


api/src/opentrons/config/feature_flags.py
```diff
@@ -20,10 +21,17 @@ def use_old_aspiration_functions() -> bool:
...
+    if robot_type == "OT-2 Standard":
+        return advs.get_setting_with_env_overload(
+            "enableDoorSafetySwitch", RobotTypeEnum.OT2
+        )
+    elif robot_type == "OT-3 Standard":
+        return advs.get_setting_with_env_overload(
+            "enableDoorSafetySwitch", RobotTypeEnum.FLEX
+        )
+    else:
+        raise KeyError("Invalid Robot Type during Door Saftey switch check")
```
sfoster1: This is a "this should never happen" type error since it means some other opentrons coder has put in the wrong value, so let's make it a little more dev-oriented by including the invalid value that got passed in. Also we probably want to make it one of the enumerated errors.

sfoster1 (approved):
Looks good to me, though I think we should change the error message a bit

api/src/opentrons/config/feature_flags.py
SyntaxColoring: We can restructure the code to avoid the possibility of this runtime error.

First, in this file, let's stick with just `RobotTypeEnum`, and not introduce the `RobotType` `"OT-2 Standard"` / `"OT-3 Standard"` literal. That lets us avoid the `if`/`elif`/`else` here; we can just pass our `robot_type` argument straight down to `advs.get_setting_with_env_overload()`.

That pushes the conversion burden up to `EngineStore` and `MaintenanceEngineStore`. To help them, let's add a utility function to `shared-data/python/opentrons_shared_data/robot/dev_types.py` to convert `RobotType`s to `RobotTypeEnum`s. Putting it there will make it hard to fall out of sync, because that's where we define the underlying types.

Finally, if we write it something like this:

```python
def robot_type_to_robot_type_enum(robot_type: RobotType) -> RobotTypeEnum:
    if robot_type == "OT-2 Standard":
        return RobotTypeEnum.OT2
    elif robot_type == "OT-3 Standard":
        return RobotTypeEnum.FLEX
    # Note: deliberately no final `else` statement here!
```

Then we use a form of mypy's [exhaustiveness checking](https://mypy.readthedocs.io/en/stable/literal_types.html#exhaustiveness-checking), which will statically ensure that we haven't forgotten any cases. So no need to `raise KeyError`.

SyntaxColoring (approved):
Looks great, thank you!

PR #13721, opened 2023-10-04
fix(app): fix conditional rendering of cal dashboard completion screen

mjhuff:
Closes RQA-1727

# Overview

This PR addresses an issue when running through all  OT-2 calibration wizard flows yields a "using current calibrations" completion screen at the very end instead of the correct "calibrations complete" screen.

The deck cal wizard incorrectly marks deck cal as being aborted when clicking the exit buttons on the final step of the flow. The fix is simply to conditionally mark the deck cal flow as aborted only if the exit occurs on a step that's not the success step.

### Current  Behavior
Screenshot 2023-10-02 at 11 58 25 AM

### With Fix
Screenshot 2023-10-04 at 2 16 18 PM

# Test Plan
- On an OT-2, run through deck calibration and all pipette calibration flows that are required. At the end of the final pipette calibration, there should be a "calibrations complete" message.
- Start deck recalibration, click past the "get started" screen, then exit the wizard. Observe the "using current calibrations" message.

# Risk assessment
low

### DIFF
--- a/app/src/assets/localization/en/robot_calibration.json
+++ b/app/src/assets/localization/en/robot_calibration.json
@@ -21,7 +21,6 @@
   "calibration_recommended": "Calibration recommended",
   "calibration_status": "Calibration Status",
   "calibration_status_description": "For accurate and precise movement, calibrate the robot's deck, pipette offsets, and tip lengths.",
-  "calibrations_aborted": "Using current calibrations.",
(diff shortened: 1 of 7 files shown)


koji (approved):
followed the test plan and the app worked as expected.
the changes look good to me.

PR #13722, opened 2023-10-04
feat(app): add useLoadedFixturesConfigStatus to diff protocol and deck config

brenthagen:
# Overview

adds a hook `useLoadedFixturesConfigStatus` to provide the configuration status for each loaded protocol fixture according to the current robot deck config

closes RAUT-695

# Test Plan

 - added unit tests
 - manually verified output with stubs in app

# Changelog

 - Adds useLoadedFixturesConfigStatus hook

# Review requests

look at shape of hook input/output, will this work

# Risk assessment

low

### DIFF
--- a/app/src/resources/deck_configuration/hooks.ts
+++ b/app/src/resources/deck_configuration/hooks.ts
@@ -1,0 +1,44 @@
+import { useDeckConfigurationQuery } from '@opentrons/react-api-client'
+
+import type { Fixture, LoadFixtureRunTimeCommand } from '@opentrons/shared-data'
+
+export const CONFIGURED = 'configured'
+export const CONFLICTING = 'conflicting'
+export const NOT_CONFIGURED = 'not configured'
+
+type LoadedFixtureConfigurationStatus =
+  | typeof CONFIGURED
+  | typeof CONFLICTING
+  | typeof NOT_CONFIGURED
+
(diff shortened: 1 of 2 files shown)


app/src/resources/deck_configuration/hooks.ts
```diff
@@ -0,0 +1,44 @@
+import { useDeckConfigurationQuery } from '@opentrons/react-api-client'
+
+import type { Fixture, LoadFixtureRunTimeCommand } from '@opentrons/shared-data'
+
+export const CONFIGURED = 'configured'
+export const CONFLICTING = 'conflicting'
+export const NOT_CONFIGURED = 'not configured'
```
jerader, line 7: should these go into i18n? I guess the `conflicting` doesn't translate to the i18n string correctly but maybe it'd be nice for the hook to do the i18n translation for us in the return.

jerader (approved):
lgtm! just left a question

app/src/resources/deck_configuration/hooks.ts
brenthagen: i was thinking of these as data values that we want to remain consistent - i think it's better to let the individual components determine their own translated display values

jerader: yup that makes sense to me!

PR #13723, opened 2023-10-04
feat(protocol-designer): add staging area tile to create file wizard

jerader:
closes RAUT-690

# Overview

This PR creates a staging area tile in the create file wizard that allows you to select where you want to place your staging areas. It only renders on the flex.

# Test Plan

Turn on the deck configuration FF and create a Flex protocol. See that the Staging Area tile exists and works as expected to add the staging areas. After you create the file, open up dev tools and see in redux that the `additionalEquipmentInvariantProperties` in `stepForms` shows the entities exist.

Try it again for an Ot-2 and see that the tile doesn't render. And then turn off the FF and try it again for Flex and see that the page doesn't render.

NOTE: the steps in the wizard are wrong, that will be fixed in a follow up ticket
ANOTHER NOTE: this doesn't wire up the ability to edit yet in the Additional Items section

Screen Shot 2023-10-04 at 4 47 30 PM

# Changelog

- create `StagingAreaTile` and test and add the new step to Create file wizard
- add logic for goBack depending on what the user selected
- expand the types to include `stagingArea`
- create the deck fixtures in `CreateFileWizard`
- fix a small bug where you could still enter and call the button even when disabled

# Review requests

see test plan

# Risk assessment

low


ecormany:
Should we label column 4 on the selection screen? Slightly larger labels might be nice too.

Also, I expected to be able to click anywhere in the blue rectangle to add, not just the smaller + target. (It makes sense that you have to click exactly on the X to remove, though.)

jerader:
> Should we label column 4 on the selection screen? Slightly larger labels might be nice too.
> 
> Also, I expected to be able to click anywhere in the blue rectangle to add, not just the smaller + target. (It makes sense that you have to click exactly on the X to remove, though.)

@ecormany, those are good call outs. This component is actually shared with the app which I think is still changing? 🤔 but i'll keep note of these, i agree that it would be easier to click anywhere other than just the X. Here is the usage in the app:

Screen Shot 2023-10-11 at 10 00 46 AM

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -169,14 +169,18 @@ export function ModulesAndOtherTile(props: WizardTileProps): JSX.Element {
...
-                props.values.pipettesByMount.left.pipetteName === 'p1000_96'
-              ) {
-                goBack(3)
-              } else if (
-                props.values.pipettesByMount.right.pipetteName === ''
-              ) {
-                goBack(2)
+              if (!enableDeckModification || robotType === OT2_ROBOT_TYPE) {
+                if (
+                  props.values.pipettesByMount.left.pipetteName === 'p1000_96'
```
koji: I think you can remove props since you destruct props in the above.
```suggestion
                  values.pipettesByMount.left.pipetteName === 'p1000_96'
```

protocol-designer/src/components/modals/CreateFileWizard/StagingAreaTile.tsx
```diff
@@ -0,0 +1,155 @@
+import * as React from 'react'
+import { useSelector } from 'react-redux'
+import without from 'lodash/without'
+import { i18n } from '../../../localization'
```
koji: this line can be after `@opentrons/shared-data`

protocol-designer/src/components/modals/CreateFileWizard/StagingAreaTile.tsx
```diff
@@ -0,0 +1,155 @@
...
+  Flex,
+  Text,
+  SPACING,
+  JUSTIFY_SPACE_BETWEEN,
+  ALIGN_CENTER,
+  PrimaryButton,
+  DeckConfigurator,
+} from '@opentrons/components'
+import {
+  DeckConfiguration,
```
koji: I guess this can be import type { DeckConfiguration } from '@opentrons/shared-data'

```suggestion
  DeckConfiguration,
```

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
```diff
@@ -62,6 +62,7 @@ import { WizardHeader } from './WizardHeader'
 
 import type { NormalizedPipette } from '@opentrons/step-generation'
 import type { FormState } from './types'
+import { StagingAreaTile } from './StagingAreaTile'
```
koji: Can you move this line up?

koji (approved):
lgtm
left a couple of comments.

PR #13724, opened 2023-10-04
refactor(api): Monitor PAPI commands, not PE commands

SyntaxColoring:
# Overview

Closes RSS-320 for Python protocols (but not for JSON protocols). Also moves us towards other things in RSS-318.

# Test Plan

No manual tests needed. This is covered by unit tests now, and it's going to be covered by other manual tests as work continues on RSS-318.

If you do test it manually, note that there's at least one unrelated bug that causes the run log text to be wrong: RSS-368.

# Changelog

#13629 added a mechanism to monitor the progress of Protocol Engine commands. It turns out that that isn't a great approach for RSS-318.

When `opentrons_simulate` and `opentrons_execute` print commands, we want them to preserve the structure of the original protocol file: transfers contain aspirates and dispenses, and so on. If we only see a flat sequence of Protocol Engine commands, we lose that structure.

Also, we don't have any mechanism right now for converting raw Protocol Engine commands into human-readable strings. That's RSS-368, and we will probably need it for JSON protocols, but we would prefer to not have to deal with it to get `opentrons_simulate` to work for Python protocols again.

So, instead, we go back to how things worked in our olden RPC times. Instead of a *flat* sequence of *Protocol Engine* commands, `opentrons_simulate` and `opentrons_execute` will see a *nested sequence* of *Python Protocol API* commands.

Specifically, we:

* Remove `ProtocolEngine.state_update_broker` and friends. This reverts most of #13629.
* Replace it with `ProtocolRunner.broker`.
    * Instead of returning Protocol Engine commands, this currently returns the old `opentrons.commands` `TypedDict`s. This will probably change eventually.

# Review requests

Is this a good approach, architecturally?

# Risk assessment

Low, but there is a tradeoff going forward.

Although these messages give us human readability by default, they are *not* exhaustive by default. `ProtocolContext`, `InstrumentContext`, and `ModuleContext` need to explicitly publish activity that they want to print. We will need to make sure they do this for the new commands like `move_labware()`. Let's track that as RSS-377.


CaseyBatten (approved):
This looks good, local tests all pass as well. One comment to note for the future of this and things related to it, app team seems interested in the notion of "generic annotations" inside PE going forward (so wrapping multiple aspirations under a parent transfer command at the highest level of the log and only having the individual aspirates appear once you looked into the log contents of that transfer, if my understanding is correct). So long as this supports that kind of "readability" change in the future I see no reason why, architecturally, this wouldn't be acceptable.

PR #13728, opened 2023-10-05
fix(api): Flag pipette as not ready to aspirate after pushing out air

andySigler:
# Overview

Set `ready_to_aspirate = False` if a push-out volume is used during dispense. The plunger would then be below the `bottom` position, and so would not be ready to aspirate.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1842,7 +1842,9 @@ class OT3API(
             raise
         else:
             dispense_spec.instr.remove_current_volume(dispense_spec.volume)
-            if push_out or dispense_spec.instr.push_out_volume:
+            bottom = dispense_spec.instr.plunger_positions.bottom
+            plunger_target_pos = target_pos[Axis.of_main_tool_actuator(realmount)]
+            if (push_out is not None and push_out > 0) or  plunger_target_pos > bottom:
                 dispense_spec.instr.ready_to_aspirate = False
 
     async def blow_out(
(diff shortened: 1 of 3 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1842,6 +1842,8 @@ async def dispense(
             raise
         else:
             dispense_spec.instr.remove_current_volume(dispense_spec.volume)
+            if push_out or dispense_spec.instr.push_out_volume:
```
sfoster1: probably better for this to just be this since it's what actually controls things
```suggestion
            if dispense_spec.instr.push_out_volume:
```

sfoster1 (changes requested):
Also we should add a test

api/src/opentrons/hardware_control/ot3api.py
andySigler: if caller sets `push_out=0` then the pipette is still ready-to-aspirate, so we need to either check this argument or we could specify what push-out volume was used within the `dispense_spec` and check that instead

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1842,6 +1842,10 @@ async def dispense(
             raise
         else:
             dispense_spec.instr.remove_current_volume(dispense_spec.volume)
+            bottom = dispense_spec.instr.plunger_positions.bottom
+            plunger_target_pos = target_pos[Axis.of_main_tool_actuator(realmount)]
+            if (push_out is not None and push_out > 0) or plunger_target_pos > bottom:
```
ryanthecoder: Can we change this to just the `plunger_target_pos > bottom` ? that should be the only condition we care about if i'm thinking about this right

andySigler: makes sense to me

ryanthecoder (approved):

sfoster1 (approved):
Looks good to me!

PR #13729, opened 2023-10-05
fix(app): only show leveling screen for gen2 multi-channel pipettes

smb2268:
fix RQA-1735

# Overview
Pipette leveling was introduced for GEN2 pipettes but we were trying to show the leveling screen for GEN1 pipettes. Missing GEN1 assets were causing a white screen. To solve this, we'll only show the pipette leveling step for GEN2 pipettes.

# Test Plan
Attach a Gen1 multi channel pipette and see that there is no longer a white screen nor are you prompted to level the pipette

# Changelog
Add generation check for leveling screen

# Review requests

Quick code check!
# Risk assessment
Low


jerader (approved):
logic makes sense to me!

koji (approved):
The changes look good to me.
Once pass Dereck's check, this pr is good to go.

PR #13730, opened 2023-10-05
fix(app): fix remove probe asset for module cal with 96 channel

smb2268:
fix RQA-1511

# Overview
Use detach probe asset for module calibration with 96 channel instead of detach pipette asset

# Test Plan
Run through module calibration with a 96-channel and see that the correct asset is now present for the probe detach screen

# Changelog
Updated asset URL

# Review requests
Make sure asset URL matches correct asset

# Risk assessment
Low


smb2268:
Closing, this is already fixed in edge

PR #13731, opened 2023-10-05
fix(shared-data): Remove v7 parameters from the v6 schema

CaseyBatten:
# Overview
Relates to: https://opentrons.atlassian.net/browse/RSS-293
v6 Schema edited to remove references to v7 Schema parameters. These parameters were in place during a stage in development when the two had yet to fully separate. Adapter ID also removed from V7 due to deprecation.

### DIFF
--- a/shared-data/python/opentrons_shared_data/protocol/models/protocol_schema_v6.py
+++ b/shared-data/python/opentrons_shared_data/protocol/models/protocol_schema_v6.py
@@ -53,8 +53,6 @@ class Params(BaseModel):
     offset: Optional[OffsetVector]
     profile: Optional[List[ProfileStep]]
     radius: Optional[float]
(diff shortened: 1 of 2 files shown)


sanni-t (approved):
✅

SyntaxColoring (commented):
Code changes look good—thank you!

I think to reduce release risk, though, we should retarget this to `edge` instead of `chore_release-7.0.1`.

CaseyBatten:
@jerader Would you mind double checking to ensure that no PD schema V6 protocols use newLocation/strategy from the V6 schema and the adapterID from V7 that have been removed. If not, then we can continue with the merge to 7.0.1.

jerader (approved):
lgtm! this shouldn't affect any PD protocols

PR #13733, opened 2023-10-05
fix(app): clear redux wifi status data on disconnect

brenthagen:
# Overview

provides a fallback to clear redux wifi interface data when the wifi disconnect modal is in a state where it has presumed a successful disconnection. based on observation of connected/disconnected state, provides a wifi interface reducer value of:
```
       [wifiInterfaceKey]: {
          ipAddress: null,
          macAddress: 'unknown',
          gatewayAddress: null,
          state: 'disconnected',
          type: 'wifi'
        },
```
This should clear the existing stale data in the case that wifi data hangs around too long due to lack of healthy HTTP communication. 

In the case that HTTP communication is healthy, the reducer value above should be replaced by the next successful `networking/status` call.

closes RQA-1745

# Test Plan

 - updated unit tests for redux and disconnect modal

# Changelog

 - Clears desktop app redux wifi status data on disconnect

# Review requests

 - check behavior of wifi disconnect on robot: 1) when only wifi connected 2) when also connected by ethernet. check for unexpected side effects of declaring reducer data 
 

# Risk assessment

low


shlokamin (approved):
🚢

PR #13736, opened 2023-10-06
test(api): make `test_filter_zero_duration_step`filter a zero-duration step

fsinapi:
# Overview

This test hasn't actually been testing anything for a while. This PR restores the original intention of the test, which is to ensure that move groups omit steps with a duration of 0.

# Test Plan

Ran the test and it worked

# Changelog

- Configure `test_filter_zero_duration_step` to filter out a step from a move group

# Review requests

# Risk assessment

None


sfoster1 (approved):
Missed this one somehow.

PR #13737, opened 2023-10-06
feat(api): Replace obsolete `machine` args with `robot_type`

SyntaxColoring:
# Overview

Closes RSS-266 and RSS-267.

# Test Plan

I will:

* [x] Test simulating an OT-2 protocol locally.
* [x] Test simulating an OT-2 protocol on an OT-2.
* [ ] Test simulating a Flex protocol on an OT-2. This should fail with an error message describing the robot type mismatch.
    * Turns out this can't be tested until #13709 merges. Deferring this test until after then.
* [x] Test simulating an OT-2 protocol on a Flex. This should fail with an error message describing the robot type mismatch.

# Changelog

Main items, RSS-266 and RSS-267:

* Remove the `machine` argument of the `opentrons.simulate.simulate()` function and `opentrons_simulate` CLI. Instead, we automatically infer it from the contents of the protocol.
* Remove the `machine` argument of `opentrons.simulate.get_protocol_api()`. Replace it with `robot_type`, and make its values match the `robotType` key in the `requirements` dict of Python protocols. 

Other stuff:

* Move `execute.py`'s `find_jupyter_labware()` helper to somewhere it can be shared between `simulate.py` and `execute.py`.
* Fix bugs in the test cases that could cause false passes.
* Add various notes and comments.

# Review requests

None in particular.

# Risk assessment

The removal of the `machine` arguments is technically a breaking change for any OT-2 user who's gone out of their way to discover and use them. I'm betting nobody has done this, and even if they have, the fix is trivial. But if anybody disagrees, we can add some kind of backwards compatibility shim.


api/src/opentrons/simulate.py
```diff
@@ -276,6 +296,32 @@ def _build_protocol_context(
...
+    elif robot_type != current_robot_type:
+        # Match robot server behavior: raise an early error if we're on a robot and the caller
+        # tries to simulate a protocol written for a different robot type.
+
+        # FIXME: This exposes the internal strings "OT-2 Standard" and "OT-3 Standard".
+        # https://opentrons.atlassian.net/browse/RSS-370
+        raise RuntimeError(
+            f'This robot is of type "{current_robot_type}",'
+            f' so it can\'t simulate protocols for robot type "{robot_type}"'
+        )
```
jbleon95, line 322: Rather than expose that in this error message, until we fix this it might be better if we just say something along the lines of "robotType defined in protocol cannot be simulated on this machine."

jbleon95 (approved):
Looks good, just have one comment but other than that good to merge 🤖

api/src/opentrons/simulate.py
SyntaxColoring: Yeahhhhhhh, maybe.

I'm going to leave it as-is for now, to try to stay closer to robot-server, which provides these details and does it properly. Ideally, the logic to generate the human-readable error would be shared between them.

PR #13741, opened 2023-10-09
fix(app, components): fix module calibration slot selection rendering for tc

koji:
# Overview
Add a new g tag for TC.
The current implementation is a little bit messy, so I will refactor related components later.

Screenshot 2023-10-09 at 12 34 34 PM

Screenshot 2023-10-09 at 4 13 56 PM

# Test Plan
- calibrate/recalibrate TC
Check the calibration steps are done properly

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/components/src/hooks/useSelectDeckLocation/index.tsx
+++ b/components/src/hooks/useSelectDeckLocation/index.tsx
@@ -11,7 +11,10 @@ import { Text } from '../../primitives'
 import { ALIGN_CENTER, JUSTIFY_CENTER } from '../../styles'
 import { DeckSlotLocation } from '../../hardware-sim/DeckSlotLocation'
 
-const X_CROP_MM = 60
+const X_CROP_MM = 0
+const X_ADJUSTMENT_FOR_TC = '-50'
+const Y_ADJUSTMENT_FOR_TC = '214'
+
 export function useDeckLocationSelect(
   robotType: RobotType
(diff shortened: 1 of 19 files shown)


components/src/hooks/useSelectDeckLocation/index.tsx
```diff
@@ -60,22 +62,18 @@ export function DeckLocationSelect({
...
+                d="M-97.8,496.6h239c2.3,0,4.2-1.9,4.2-4.2v-282c0-2.3-1.9-4.2-4.2-4.2h-239c-2.3,0-4.2,1.9-4.2,4.2v282 C-102,494.7-100.1,496.6-97.8,496.6z"
+              />
               <RobotCoordsForeignDiv
-                x={slot.position[0]}
-                y={slot.position[1]}
+                x="-50"
+                y="214"
                 width={slot.boundingBox.xDimension}
-                height={slot.boundingBox.yDimension}
+                height="282"
```
jerader: maybe these can be consts declared at the top of the component

jerader (approved):
it works but i agree that it can use a bit of refactoring later to increase readability. we are also going to have to add this to PD's usage of `SelectLocation` later.

components/src/hooks/useSelectDeckLocation/index.tsx
koji: addressed the feedback, thank you.

PR #13743, opened 2023-10-09
feat(api): add resetAuthorizedKeys option for the OT-2

vegano1:
# Overview

We want to be able to reset the authorized_keys file for the OT-2 as well, so let's add the option to do so.

# Test Plan

- [x] Make sure that we can't ssh into an OT-2 after hitting the `/settings/reset` endpoint with the `AuthorizedKeys` option

# Changelog

- add `AuthorizedKeys` option to list of resettable options for the OT-2

# Review requests

# Risk assessment


sfoster1 (approved):
Looks good to me!

PR #13745, opened 2023-10-09
feat(api): add AuthorizedKeys reset option for the OT-2 + fixes

vegano1:
# Overview

This change should have been merged into the `chore_release-7.0.1` branch but was merged to `edge` instead. So let's Cherry-pick the commit from edge and retarget to `7.0.1` release branch. Original pr can be found [here](https://github.com/Opentrons/opentrons/pull/13743). This pr also adds a few improvements to the `/from_keys` function, we now skip hidden (.) files, we handle errors so we don't stop processing all pub keys found if one of them fails to open, we now accept `ecdsa` keys like the OT-2.

# Test Plan

- [x] Make sure we can still use the `ssh_keys/from_local` endpoint to add ssh keys from a USB drive
- [x] Clear all ssh keys on the robot, make sure you can't ssh, now use the `ssh_keys/from_local` endpoint and make sure you can ssh.
- [x] Create a binary file that ends in `.pub` and make sure we skip it
- [x] Make sure we skip hidden files because MacOS likes to create hidden binary files that are invalid to us
- [x] Make sure we don't stop processing all valid public keys we find if one of them is invalid.

# Changelog

- Skip hidden files when searching the USB drive, MacOS likes to create hidden binary files in the USB drive
- Added `ecdsa` as a valid public key type to match the OT-2
- Handle exceptions so only skip the invalid key and not all subsequent potentially valid keys.

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/config/reset.py
+++ b/api/src/opentrons/config/reset.py
@@ -49,6 +49,7 @@ _OT_2_RESET_OPTIONS = [
     ResetOptionId.pipette_offset,
     ResetOptionId.tip_length_calibrations,
     ResetOptionId.runs_history,
+    ResetOptionId.authorized_keys,
 ]
 _FLEX_RESET_OPTIONS = [
     ResetOptionId.boot_scripts,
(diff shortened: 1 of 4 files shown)


mjhuff (approved):
lgtm, thank you!

sfoster1 (approved):
The changes to handling keys look good to me!

mjhuff (approved):
Tested on an OT-2 and works! Code makes sense per in-person convo.

PR #13746, opened 2023-10-09
fix(odd): pretend to be online all the time

sfoster1:
For some reason, when the ODD starts and the robot boots while not connected to any network, some (some!) requests to localhost will fail. It's really strange. One thing that seems to help is pretending to be online by overriding the `navigator.onLine` browser api to always return true.

## Testing
- Boot the odd with no network connected and see if stuff works I guess

Closes RQA-1754


app-shell-odd/Makefile
```diff
@@ -48,7 +48,7 @@ setup:
 
 .PHONY: clean
 clean:
-	shx rm -rf lib dist
+	shx rm -rf lib dist opentrons-robot-app.tar.gz
```
shlokamin, line 51: lol ty

app/src/App/OnDeviceDisplayApp.tsx
```diff
@@ -47,9 +47,13 @@ import { useCurrentRunRoute, useProtocolReceiptToast } from './hooks'
 
 import { OnDeviceDisplayAppFallback } from './OnDeviceDisplayAppFallback'
 
+import { hackWindowNavigatorOnLine } from './hacks'
+
 import type { Dispatch } from '../redux/types'
 import type { RouteProps } from './types'
 
+hackWindowNavigatorOnLine()
+
```
shlokamin, line 58: can we throw a comment before this fn call?

app/src/App/OnDeviceDisplayApp.tsx
```diff
@@ -257,7 +261,7 @@ export const OnDeviceDisplayApp = (): JSX.Element => {
 
   // TODO (sb:6/12/23) Create a notification manager to set up preference and order of takeover modals
   return (
-    <ApiHostProvider hostname="localhost">
+    <ApiHostProvider hostname="127.0.0.1">
```
shlokamin, line 266: i guess this doesnt hurt but probably not necessary anymore

shlokamin (approved):
lgtm, what a journey

app/src/App/OnDeviceDisplayApp.tsx
sfoster1: yeah... i like having it though

PR #13747, opened 2023-10-09
refactor(app): list experimental protocol stats on detail view for lifetime testing

b-cooper:
# Overview

Create a feature flag "protocolStats" that gates access to a new "stats" tab in the protocol details
page.

Provide Quick Counts of the following:

- Grips
- aspirates per pipette
- dispenses per pipette
- pick up tips per pipette

# Risk assessment
low, behind FF


app/src/organisms/ProtocolDetails/ProtocolStats.tsx
```diff
@@ -0,0 +1,187 @@
...
+  JUSTIFY_CENTER,
+  SIZE_AUTO,
+  SPACING,
+  TYPOGRAPHY,
+} from '@opentrons/components'
+import { StyledText } from '../../atoms/text'
+import {
+  getPipetteNameSpecs,
+  ProtocolAnalysisOutput,
+} from '@opentrons/shared-data'
```
jerader, line 19: import can go higher

app/src/organisms/ProtocolDetails/index.tsx
```diff
@@ -74,6 +74,7 @@ import type { JsonConfig, PythonConfig } from '@opentrons/shared-data'
 import type { StoredProtocolData } from '../../redux/protocol-storage'
 import type { State, Dispatch } from '../../redux/types'
 import { useFeatureFlag } from '../../redux/config'
+import { ProtocolStats } from './ProtocolStats'
```
jerader: imports can go higher

.gitignore
```diff
@@ -158,3 +158,6 @@ opentrons-robot-app.tar.gz
 
 # ignore linux swap files
 *.swp
+
+# asdf versions file
+*.tool-versions
```
jerader, line 163: why do you need this?

jerader (approved):
nice, this is actually pretty neat. One thing i noticed is we show grip count for OT-2 or if there are no move with gripper commands, but since this is for lifetime testing only it is probably fine.

Screen Shot 2023-10-11 at 11 05 46 AM

.gitignore
b-cooper: this addition has to do with us using `asdf` in the `protocol-library-reborn` project as an alternative to `nvm` which is used in the mono-repo. We will be developing across this boundary going forward, so I added an ignore here until we're ready to recommend one tool in both projects. That way, developers on our team aren't forced to switch up their tooling when changing between development areas of the same flavor of code.

PR #13749, opened 2023-10-09
fix(app): RunProgressMeter status issue when run is completed

koji:
# Overview
updated the text to align with https://github.com/Opentrons/opentrons/pull/13749#issuecomment-1756266771

https://github.com/Opentrons/opentrons/assets/474225/db486c92-79eb-4b8b-b9a7-34f5cac4b8b6

close RQA-1632

# Test Plan
- run a protocol
RunProgressMeter shows Protocol completed when run is completed.
 

# Changelog
- remove protocol completed condition and update current step part condition to add final step
- add a test for final step (the test for this component will be update in the future)

# Review requests

# Risk assessment

### DIFF
--- a/app/src/assets/localization/en/run_details.json
+++ b/app/src/assets/localization/en/run_details.json
@@ -38,6 +38,7 @@
   "error_info": "Error {{errorCode}}: {{errorType}}",
   "error_type": "Error: {{errorType}}",
   "failed_step": "Failed step",
+  "final_step": "Final Step",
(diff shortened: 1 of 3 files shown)


b-cooper:
Screen Shot 2023-10-10 at 5 10 38 PM

re: conversation with design. Instead of "Protocol completed" let's add swap the "Current Step" label for the "Final Step" header if the run status is terminal

jerader (approved):
lgtm!

PR #13753, opened 2023-10-10
refactor(app-shell): Migrate protocols to "protocols" directory

mjhuff:
Closes RAUT-716

# Overview

This PR is a follow-up to #13561, which addressed pre-parity protocol directory compatibility if a 7.0.0 and <7.0.0 version of the app were used on the same computer. Now that there will only be one supported app soon and all protocols can go in the same directory again, it makes sense to migrate protocols back to the more semantic and familiar directory name, "protocols". 

More importantly, there's no need to support a function that checks all protocols in an old directory every time the app is launched - let's migrate all the protocols from the temp 7.0 directory to "protocols" then delete the folder. That way, our only necessary act of atonement moving forward is to check to see if the temp directory exists every startup (asynchronously!). 

The one downside to this approach is if two protocols have the same name UUID - the older protocol in the protocol directory will be overwritten. Considering this scenario could only occur for the handful of users utilizing both versions of the app, odds are pretty good.

The error boundary added in the earlier PR will prevent this exact scenario from occurring again!

# Test Plan
- DO NOT OPEN THE APP YET.
- Navigate to the parent directory of "protocols". /Users/Library/Application Support/Opentrons/ on Mac.
- To test different behavior, make a copy of protocols-v7.0-supported in some other directory.
- Open the app. Notice that any protocols in the v7.0-supported folder are moved into the protocols folder if they didn't exist there already. Notice that the v7.0-supported folder is deleted.
- Protocols folder should be created if it doesn't already exist. 

# Changelog
- Removed pre-app parity protocol migration.
- Added post-app parity protocol migration. 

# Review requests
- Double check my reasoning. I think this is the most effective long term solution... 

# Risk assessment
low but long term implications


app/src/organisms/ProtocolDetails/index.tsx
```diff
@@ -344,6 +345,17 @@ export function ProtocolDetails(
     setShowChooseRobotToRunProtocolSlideout(true)
   }
 
+  const UNKNOWN_ATTACHMENT_ERROR = `${protocolDisplayName} protocol uses 
+  instruments  or modules from a future version of Opentrons software. Please update
```
koji: The space between `instruments` and `or` is 2-space. Is this on purpose?
If not, ↓

```suggestion
  instruments or modules from a future version of Opentrons software. Please update 
```

app/src/organisms/ProtocolsLanding/ProtocolCard.tsx
```diff
@@ -79,13 +79,14 @@ export function ProtocolCard(props: ProtocolCardProps): JSX.Element | null {
   )
 
   const UNKNOWN_ATTACHMENT_ERROR = `${protocolDisplayName} protocol uses 
-  instruments  or modules from a future version of the app. Please update 
+  instruments  or modules from a future version of Opentrons software. Please update
```
koji: same (2 spaces)

koji (approved):
The changes look good to me.
I tested this branch with macOS and Windows and both of them worked as expected.

PR #13755, opened 2023-10-10
fix(app) fix robotName spacing issue in RobotStatusHeader

koji:
# Overview
Fix [this issue](https://opentrons.slack.com/archives/CSCLVUW3C/p1696620573224269).
Had a conversation on protocolName with Mel then we decided that we will fix protocolName part in the future.
I think after updating Electron version, we can use container query and that allows us to make RobotCard more flexible.

This PRs add the following rule.

When the card shows `Go to Run`
windows size less than 678px -> name space 105px (if the name is longer than the space, display ....)
otherwise use 150px  (if the name is longer than the space, display ....)

When no `Go to Run`
Display the name without using `...`

design
https://www.figma.com/file/l7BAJAGICr10ELsSqQD0Sr/Desktop-App-Component-Library?type=design&node-id=220-8246&mode=design&t=MrjSqY9f0J0zX7Qq-0

In the design, Mel mentioned to a specific length but we cannot do this right now since we decided to use css only to control the layout.

without Go to Run
Screenshot 2023-10-10 at 4 06 31 PM

with Go to Run
Screenshot 2023-10-10 at 4 06 48 PM

# Test Plan
1. Device landing page
2. Set up a protocol
3. Back to Device landing page

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/app/src/organisms/Devices/RobotStatusHeader.tsx
+++ b/app/src/organisms/Devices/RobotStatusHeader.tsx
@@ -138,7 +138,7 @@ export function RobotStatusHeader(props: RobotStatusHeaderProps): JSX.Element {
 
   const RobotNameContainer = styled.div`
     max-width: ${(props: RobotNameContainerProps) =>
-      props.isGoToRun ? `149px` : undefined};
+      props.isGoToRun ? `150px` : undefined};
     @media screen and (max-width: 678px) {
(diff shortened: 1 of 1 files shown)


mjhuff (approved):
Tested and works. Code looks good - looks much better now!

PR #13756, opened 2023-10-10
fix(app): use current labware location instead of loaded location

smb2268:
fix RQA-1770

# Overview
Currently in the run log command text for pipetting commands we show the loaded labware location - we should show the current labware location

# Test Plan
Upload a protocol that has a move labware command followed by some pipetting commands. Verify that the location shown in the pipette commands in the run log is the current location, not the location prior to the move
I tested this by looking at DVT1ABR3 running Illumina DNA Enrichment.py DRYRUN and it is working as expected

# Changelog
Use `getFinalLabwareLocation` util to grab labware location instead of the corresponding `loadLabware` command

# Review requests
Test this out by looking at a long protocol run and see if you notice a performance decrease with this commit

# Risk assessment
Medium - this could be visibly less performant since we have to parse through previous commands for `moveLabware` commands


jerader (approved):
lgtm!

PR #13759, opened 2023-10-11
feat(protocol-designer): wire up staging areas to loadFixture commands

jerader:
closes RAUT-744 RAUT-770

# Overview

Wires up the staging areas to create `loadFixture` commands and when you import a protocol back to PD, the `loadFixture` commands get populated back as staging areas in redux

# Test Plan

Make sure the deck configuration ff is turned on and add a staging area in the create file wizard. Export the protocol. Open the json file and see that there is a `loadFixture` command associated with the staging area and location that you added.

Change the version in the json file from `7.0.0-candidate-d` to `7.1` so the protocol doesn't unnecessarily go through a migration. Then import the protocol back into PD. Open dev tools and look at `StepForms` -> `additionalEquipmentInvariantProperties` and see that the staging area entities are there.

# Changelog

- add logic for import/export to manage the staging areas
- fix the import order on the files

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/file-data/selectors/fileCreator.ts
+++ b/protocol-designer/src/file-data/selectors/fileCreator.ts
@@ -14,8 +14,6 @@ import {
   SPAN7_8_10_11_SLOT,
   Cutout,
 } from '@opentrons/shared-data'
-import { getFileMetadata, getRobotType } from './fileFields'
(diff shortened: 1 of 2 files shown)


mjhuff (approved):
🚀

PR #13761, opened 2023-10-11
feat(app, app-shell): add file upload progress to robot update

mjhuff:
# Overview

This PR adds upload file progress to the progress bar during a robot update. 

# Test Plan
- Start a robot update, whether via zip file or a download banner (requires building the app via make -C app-shell package and then running the dist). Note in the console that there is a wall of Redux actions, robotUpdate:FILE_UPLOAD_PROGRESS - the progress bar will start filling earlier than in previous alphas. 
- The update should proceed as normal.

# Changelog
- Added upload file progress to robot update flows.  

# Risk assessment
lowish. There are some very bizzare errors that can occur due to event listeners, but I _believe_ I've stamped them out.


app-shell/src/http.ts
```diff
@@ -86,31 +87,67 @@ export function postFile(
...
+      if (seenDataLength === size) {
+        readStream.removeListener('data', onData)
+        readStream.removeListener('error', handleError)
+      }
+    }
 
     readStream.once('error', handleError)
+    readStream.on('data', onData)
 
     function handleSuccess(): void {
```
mjhuff, line 130: handleSuccess() returns before the stream ends, so the error listener wasn't kept alive while we asynchronously read from it. Remove it when the data is read and we no longer can run into an error - this way we actually use it!

app-shell/src/http.ts
```diff
@@ -86,31 +87,67 @@ export function postFile(
...
+      seenDataLength += chunk.length
+      if (
+        size !== Infinity &&
+        seenDataLength / size > notifiedDataLength / size + 0.01
+      ) {
+        progress?.(seenDataLength / size)
+        notifiedDataLength = seenDataLength
+      }
+
+      if (seenDataLength === size) {
```
mjhuff, line 121: Given the way our file upload works, we need to close the onData listener here. Closing it elsewhere causes socket hangups or Errno22s to occur.

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/useRobotUpdateInfo.ts
```diff
@@ -4,24 +4,100 @@ import type { RobotUpdateSession } from '../../../../redux/robot-update/types'
...
+    sessionStep == null
+  ) {
+    return progressPercent
+  }
+
+  const stepAndStage = `${sessionStep}-${sessionStage}`
+  // Ignored because 0-100 is too fast to be worth recording.
+  const IGNORED_STEPS_AND_STAGES = ['processFile-awaiting-file']
+  // Each stepAndStage is an equal fraction of the total steps.
+  const TOTAL_STEPS_WITH_PROGRESS = fileInfo?.isManualFile ? 3 : 3
```
koji: question

Do we need this ternary?
The value is always `3`

koji (approved):
The changes look good to me.
Followed the test plan and the app worked as expected.
I just left one comment.

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/useRobotUpdateInfo.ts
mjhuff: Good catch! I'll end up using this in a future PR but it's not needed as of yet.

mjhuff:
> The changes look good to me. Followed the test plan and the app worked as expected. I just left one comment.

Thank you for the quick review!

PR #13762, opened 2023-10-11
fix(api): Addition of catch for missing argument for feature flag test

CaseyBatten:
# Overview
When door safety check under feature flags was updated to receive RobotTypeEnum as a parameter, the decoy test that called it was never updated to account for this. This caused it to raise the following warning:
```
.../opentrons/api/tests/opentrons/conftest.py:108: IncorrectCallWarning: missing a required argument: 'robot_type'
    decoy.when(mock_get_ff()).then_return(False)
```
With the addition of a parameter check case, we will handle these tests correctly now. Also, in the future if we add the RobotTypeEnum parameter to any of the other feature flags in the future, this should catch those cases as well so long as they use the "robot_type" parameter naming convention.

### DIFF
--- a/api/tests/opentrons/conftest.py
+++ b/api/tests/opentrons/conftest.py
@@ -32,6 +32,7 @@ try:
 except (OSError, ModuleNotFoundError):
     aionotify = None
 
+from opentrons_shared_data.robot.dev_types import RobotTypeEnum
 from opentrons_shared_data.protocol.dev_types import JsonProtocol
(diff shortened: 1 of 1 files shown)


SyntaxColoring (approved):
Yep, this works given:

* The fact that there are a million feature flag functions that we don't want to mock out individually and explicitly
* The fact that, at the moment, they don't all have the same signature (which we think is something that we want to fix)

Thank you!

PR #13763, opened 2023-10-11
fix(app): keep run controls enabled if run is idle

smb2268:
fix RQA-1784

# Overview
Pipette card controls should remain accessible until the run is running.

# Test Plan
1. On an OT-2, proceed to run setup
2. Navigate back to the Devices page
3. Look at the overflow menu and see controls enabled
4. Start the run
5. See that the controls are now disabled

# Changelog
Use `isRunRunning` instead of `!isRunTerminal` to determine whether to disable the pipette overflow menu controls

# Review requests
Test this out

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/Devices/InstrumentsAndModules.tsx
+++ b/app/src/organisms/Devices/InstrumentsAndModules.tsx
@@ -61,7 +61,7 @@ export function InstrumentsAndModules({
   )?.data ?? { left: undefined, right: undefined }
   const isRobotViewable = useIsRobotViewable(robotName)
   const currentRunId = useCurrentRunId()
-  const { isRunTerminal } = useRunStatuses()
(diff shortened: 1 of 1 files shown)


mjhuff (commented):
Appears that the overflow menu is disabled before the run is started

Screenshot 2023-10-11 at 3 42 00 PM

smb2268:
> Appears that the overflow menu is disabled before the run is started
> 
> Screenshot 2023-10-11 at 3 42 00 PM

@mjhuff sorry the description wasn't clear! those shouldn't be affected - this is just for the instrument card overflow menus which are working as expected from my testing: 
Screen Shot 2023-10-11 at 3 54 01 PM

mjhuff (approved):
Works great! 🚀

PR #13764, opened 2023-10-11
fix(api): Fully stringify well names

SyntaxColoring:
# Overview

Fixes RSS-368.

# Test Plan

Run some PAPIv≥2.14 protocols through `opentrons_simulate` and check the run log. You shouldn't see the string `None` anywhere. The location strings should match PAPIv≤2.13.

Here's a test case demonstrating how we print a well of a labware on an adapter on a module on a deck slot. It also demonstrates how we prefer to print the user-defined labware label, but if that doesn't exist, we fall back to the display name specified by the labware definition.

```python


requirements = {"robotType": "Flex", "apiLevel": "2.15"}


    heater_shaker = protocol.load_module("heaterShakerModuleV1", "D1")
    source = heater_shaker.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt",
        adapter="opentrons_96_pcr_adapter",
    )
    destination = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "A2", label="destination"
    )

    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_50ul", "B1")

        "flex_1channel_50", mount="right", tip_racks=[tip_rack]
    )

    heater_shaker.close_labware_latch()
    pipette.transfer(50, source.wells()[0], destination.wells()[0])
```

```bash
$ pipenv run opentrons_simulate ~/Desktop/complicated_locations.py
/Users/maxmarrone/.opentrons/robot_settings.json not found. Loading defaults
Belt calibration not found.
Latching labware on Heater-Shaker
Transferring 50.0 from A1 of Opentrons Tough 96 Well Plate 200 µL PCR Full Skirt on Opentrons 96 PCR Adapter on Heater-Shaker Module GEN1 on D1 to A1 of destination on A2
	Picking up tip from A1 of Opentrons Flex 96 Tip Rack 50 µL on B1
	Aspirating 50.0 uL from A1 of Opentrons Tough 96 Well Plate 200 µL PCR Full Skirt on Opentrons 96 PCR Adapter on Heater-Shaker Module GEN1 on D1 at 8.0 uL/sec
	Dispensing 50.0 uL into A1 of destination on A2 at 8.0 uL/sec
	Dropping tip into A1 of Opentrons Fixed Trash on A3
```

# Details

This fixes two problems with how PAPIv≥2.14 was stringifying well names when they were printed to the `opentrons_simulate` run log:

1. We were *only* printing the *user-specified* labware `label`. That's it. If the user didn't specify a labware label, we were printing `None`, without making an attempt to fall back to the display name specified by the *labware definition.*
2. We were just printing the name of the labware itself, without descending down to the labware's parents. So, the name of the module or deck slot that the labware was on wouldn't get printed.

We fix both of these by adding a helper function to `opentrons.protocol_api.core.engine` that queries the `ProtocolEngine` for full location information and builds a human-readable string out of it. It works recursively, so it supports deep locations like "well A1 of labware Foo on adapter Bar on module Baz on C2."

This brings PAPIv≥2.14's user-facing behavior in line with that of PAPIv≤2.13.

For comparison, PAPIv≤2.13 accomplished this through a combination of [this helper function](https://github.com/Opentrons/opentrons/blob/3f85e13f30d32ae60bbe8e60d02683e0a3062fe2/api/src/opentrons/commands/helpers.py#L36), and [computing the human-readable strings eagerly](https://github.com/Opentrons/opentrons/blob/3f85e13f30d32ae60bbe8e60d02683e0a3062fe2/api/src/opentrons/protocol_api/core/legacy/legacy_labware_core.py#L70) as `Well` objects were constructed. That approach isn't sufficient for us anymore because labware can move around throughout the protocol.

# Review requests

Any other stringifications that need to be fixed?

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/stringify.py
+++ b/api/src/opentrons/protocol_api/core/engine/stringify.py
@@ -10,7 +10,7 @@ from opentrons.protocol_engine.types import (
 def well(engine_client: SyncClient, well_name: str, labware_id: str) -> str:
     """Return a human-readable string representing a well and its location.
 
-    For example: "A1 on My Cool Labware on C2".
+    For example: "A1 of My Cool Labware on C2".
     """
     labware_location = OnLabwareLocation(labwareId=labware_id)
     return f"{well_name} of {_labware_location_string(engine_client, labware_location)}"
--- a/api/src/opentrons/protocol_api/core/engine/stringify.py
+++ b/api/src/opentrons/protocol_api/core/engine/stringify.py
@@ -1,0 +1,56 @@
+from opentrons.protocol_engine.clients.sync_client import SyncClient
+from opentrons.protocol_engine.types import (
+    DeckSlotLocation,
+    LabwareLocation,
+    ModuleLocation,
+    OnLabwareLocation,
+)
+
+
+def well(engine_client: SyncClient, well_name: str, labware_id: str) -> str:
+    """Return a human-readable string representing a well and its location.
+
+    For example: "A1 on My Cool Labware on C2".
+    """
+    labware_location = OnLabwareLocation(labwareId=labware_id)
+    return f"{well_name} of {_labware_location_string(engine_client, labware_location)}"
+
+
+def _labware_location_string(
+    engine_client: SyncClient, location: LabwareLocation
+) -> str:
+    if isinstance(location, DeckSlotLocation):
+        # Returning just the ID, like "5" or "C2", matches historical behavior.
+        # Ideally, we might want to use the display name specified by the deck definition?
+        return location.slotName.id
+
+    elif isinstance(location, ModuleLocation):
+        module_name = engine_client.state.modules.get_definition(
+            module_id=location.moduleId
+        ).displayName
+        module_on = engine_client.state.modules.get_location(
+            module_id=location.moduleId
+        )
+        module_on_string = _labware_location_string(engine_client, module_on)
+        return f"{module_name} on {module_on_string}"
+
+    elif isinstance(location, OnLabwareLocation):
+        labware_name = _labware_name(engine_client, location.labwareId)
+        labware_on = engine_client.state.labware.get_location(
+            labware_id=location.labwareId
+        )
+        labware_on_string = _labware_location_string(engine_client, labware_on)
+        return f"{labware_name} on {labware_on_string}"
+
+    elif location == "offDeck":
+        return "[off-deck]"
+
+
+def _labware_name(engine_client: SyncClient, labware_id: str) -> str:
+    """Return the user-specified labware label, or fall back to the display name from the def."""
+    user_name = engine_client.state.labware.get_display_name(labware_id=labware_id)
+    definition_name = engine_client.state.labware.get_definition(
+        labware_id=labware_id
+    ).metadata.displayName
+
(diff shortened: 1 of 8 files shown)


api/src/opentrons/protocol_api/core/engine/stringify.py
```diff
@@ -0,0 +1,56 @@
...
+)
+
+
+def well(engine_client: SyncClient, well_name: str, labware_id: str) -> str:
+    """Return a human-readable string representing a well and its location.
+
+    For example: "A1 on My Cool Labware on C2".
+    """
+    labware_location = OnLabwareLocation(labwareId=labware_id)
+    return f"{well_name} of {_labware_location_string(engine_client, labware_location)}"
```
SyntaxColoring: I'm working on a unit test for this now. We might want to merge this without waiting for that, to get this into v7.0.1-alpha.3.

api/src/opentrons/protocol_api/core/engine/stringify.py
```diff
@@ -0,0 +1,56 @@
...
+    LabwareLocation,
+    ModuleLocation,
+    OnLabwareLocation,
+)
+
+
+def well(engine_client: SyncClient, well_name: str, labware_id: str) -> str:
+    """Return a human-readable string representing a well and its location.
+
+    For example: "A1 on My Cool Labware on C2".
```
jbleon95:
```suggestion
    For example: "A1 of My Cool Labware on C2".
```

api/src/opentrons/protocol_api/core/engine/stringify.py
```diff
@@ -0,0 +1,56 @@
...
+    return f"{well_name} of {_labware_location_string(engine_client, labware_location)}"
+
+
+def _labware_location_string(
+    engine_client: SyncClient, location: LabwareLocation
+) -> str:
+    if isinstance(location, DeckSlotLocation):
+        # Returning just the ID, like "5" or "C2", matches historical behavior.
+        # Ideally, we might want to use the display name specified by the deck definition?
+        return location.slotName.id
```
jbleon95: Maybe we should make this return `f'slot {location.slotName.id}'` to reduce ambiguity between Flex coordinate deck slot names and well names?

jbleon95 (approved):
Curious about if performance is affected by the increased amount of engine client calls, but otherwise looks good outside of my couple suggestions below.

api/src/opentrons/protocol_api/core/engine/stringify.py
SyntaxColoring: Yeah. That introduces an inconsistency between PAPIv≤2.13 and PAPIv≥2.14, but I think it's worth it.

SyntaxColoring:
> Curious about if performance is affected by the increased amount of engine client calls

Hm, yeah.

It seems like ideally, we would maybe only compute the parts of the run log that are actually required by the caller. For `robot-server`, that's basically none of it, whereas for `opentrons_simulate`, it would include this stuff...

SyntaxColoring:
Waiting for #13709 to hit `chore_release-7.0.1` before we merge this into `chore_release-7.0.1`.

PR #13766, opened 2023-10-11
docs(api): Change "OT-2" to "Opentrons robots" in opentrons_simulate help text

SyntaxColoring:
# Overview

Another good catch from @CaseyBatten.

`opentrons_simulate --help` said it would simulate "an OT-2 protocol." It should say it'll simulate "a protocol for an Opentrons robot" instead, to account for the Flex.

# Test Plan

None needed.

# Risk assessment

Very low.


ecormany (approved):
Nice catch! :shipit:

PR #13767, opened 2023-10-11
fix(api): cancellation bug on legacy core api

sanni-t:
Closes RQA-1553

# Overview

Fixes a bug that was causing protocols running on PAPI legacy core would freeze in 'stop-requested' state upon cancellation.

The issue was caused because #13339 readjusted the sequence of events that happen when cancelling a run such that it doesn't cause race conditions that were present before. But in that fix, we forgot that protocols running on legacy core are not controlled by the engine directly and hence need to be stopped differently- i.e., by invoking the hardware control's execution manager to cancel the execution.

This PR fixes this issue by doing 2 things-
1. makes the hardware controller run all blocking hardware access methods in cancellable tasks
2. calls `cancel()` on these tasks as soon as a `stop()` request comes in to the engine.

The combination of these two things makes the run cancel as soon as the running task is able to be cancelled, and calls the cleanup function (`engine.finish()`)

# Test Plan

1. Run any protocol on an OT-2 with API level <= 2.13 and cancel it mid-run
    - [x] verify that if a cancellation comes in mid-movement, the run cancels before the movement ends (can test with a slow-speed pipette movement)
    - [x] verify that if you cancel protocol with a tip on the pipette, then the protocol cancels and drops tips in fixed trash
    - [x] verify that if you cancel the protocol while it's on a non-api-based, `time.delay()`, the protocol ends after the delay is over
2. Run any protocol on an OT-2 & Flex with api level >= 2.14 and cancel it mid-run and verify that all the above still happen

# Changelog

- added 'taskification' of movement (& delay) functions of the hardware controller, inside the `wait_for_is_running` wrapper
- added option to taskify the functions based on the type of run
- call task cancellation from `engine.stop()` for runs with 'taskified' functions

# Review requests

Usual stuff. Make sure the PR's not causing any regression.

# Risk assessment

High for protocols running on legacy core API since it moves the behavior of core movement functions inside tasks. 
Shouldn't have any effect on engine-backed protocols.

### DIFF
--- a/api/src/opentrons/hardware_control/execution_manager.py
+++ b/api/src/opentrons/hardware_control/execution_manager.py
@@ -135,6 +135,10 @@ class ExecutionManagerProvider:
             if not inst._em_simulate:
                 await inst.execution_manager.wait_for_is_running()
             if inst.taskify_movement_execution:
+                # Running these functions inside cancellable tasks makes it easier and
+                # faster to cancel protocol runs. In the higher, runner & engine layers,
+                # a cancellation request triggers cancellation of the running move task
+                # and hence, prevents any further communication with hardware.
                 decorated_task: "asyncio.Task[DecoratedReturn]" = asyncio.create_task(
                     decorated(inst, *args, **kwargs)
                 )
--- a/api/src/opentrons/hardware_control/execution_manager.py
+++ b/api/src/opentrons/hardware_control/execution_manager.py
@@ -92,6 +92,15 @@ class ExecutionManagerProvider:
     def __init__(self, simulator: bool) -> None:
         self._em_simulate = simulator
         self._execution_manager = ExecutionManager()
+        self._taskify_movement_execution: bool = False
+
+    @property
+    def taskify_movement_execution(self) -> bool:
+        return self._taskify_movement_execution
+
+    @taskify_movement_execution.setter
+    def taskify_movement_execution(self, cancellable: bool) -> None:
+        self._taskify_movement_execution = cancellable
 
     @property
     def execution_manager(self) -> ExecutionManager:
@@ -125,7 +134,14 @@ class ExecutionManagerProvider:
         ) -> DecoratedReturn:
             if not inst._em_simulate:
                 await inst.execution_manager.wait_for_is_running()
-            return await decorated(inst, *args, **kwargs)
+            if inst.taskify_movement_execution:
+                decorated_task: "asyncio.Task[DecoratedReturn]" = asyncio.create_task(
+                    decorated(inst, *args, **kwargs)
+                )
+                inst.execution_manager.register_cancellable_task(decorated_task)
+                return await decorated_task
+            else:
+                return await decorated(inst, *args, **kwargs)
 
         return cast(DecoratedMethodReturningValue, replace)
 
--- a/api/src/opentrons/hardware_control/protocols/motion_controller.py
+++ b/api/src/opentrons/hardware_control/protocols/motion_controller.py
@@ -215,11 +215,11 @@ class MotionController(Protocol):
         ...
 
     def is_movement_execution_taskified(self) -> bool:
-        """Get whether move tasks are being executed inside cancellable tasks."""
+        """Get whether move functions are being executed inside cancellable tasks."""
         ...
 
     def should_taskify_movement_execution(self, taskify: bool) -> None:
-        """Specify whether move methods should be executed inside cancellable tasks."""
+        """Specify whether move functions should be executed inside cancellable tasks."""
         ...
 
     async def cancel_execution_and_running_tasks(self) -> None:
--- a/api/src/opentrons/hardware_control/protocols/motion_controller.py
+++ b/api/src/opentrons/hardware_control/protocols/motion_controller.py
@@ -213,3 +213,15 @@ class MotionController(Protocol):
     async def retract_axis(self, axis: Axis) -> None:
         """Retract the specified axis to its home position."""
         ...
+
+    def is_movement_execution_taskified(self) -> bool:
+        """Get whether move tasks are being executed inside cancellable tasks."""
+        ...
+
+    def should_taskify_movement_execution(self, taskify: bool) -> None:
+        """Specify whether move methods should be executed inside cancellable tasks."""
+        ...
+
+    async def cancel_execution_and_running_tasks(self) -> None:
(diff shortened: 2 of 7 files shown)


sanni-t:
Tested on Otie:
- [x] verify that if a cancellation comes in mid-movement, the run cancels before the movement ends (can test with a slow-speed pipette movement)
- [x]  verify that if you cancel protocol with a tip on the pipette, then the protocol cancels and drops tips in fixed trash
- [x] verify that if you cancel the protocol while it's on a non-api-based, time.delay(), the protocol ends after the delay is over

api/src/opentrons/hardware_control/execution_manager.py
```diff
@@ -125,7 +134,18 @@ async def replace(
...
+                # faster to cancel protocol runs. In the higher, runner & engine layers,
+                # a cancellation request triggers cancellation of the running move task
+                # and hence, prevents any further communication with hardware.
+                decorated_task: "asyncio.Task[DecoratedReturn]" = asyncio.create_task(
+                    decorated(inst, *args, **kwargs)
+                )
+                inst.execution_manager.register_cancellable_task(decorated_task)
+                return await decorated_task
+            else:
+                return await decorated(inst, *args, **kwargs)
```
SyntaxColoring, line 148: Should we update the name `@wait_for_running` at this point? Maybe `@flow_controlled` or something?

CaseyBatten (approved):
Ran through all three test cases, passed all conditions. LGTM.

api/src/opentrons/hardware_control/protocols/motion_controller.py
```diff
@@ -213,3 +213,15 @@ async def retract(self, mount: Mount, margin: float = 10) -> None:
...
+        """Get whether move functions are being executed inside cancellable tasks."""
+        ...
+
+    def should_taskify_movement_execution(self, taskify: bool) -> None:
+        """Specify whether move functions should be executed inside cancellable tasks."""
+        ...
+
+    async def cancel_execution_and_running_tasks(self) -> None:
+        """Cancel all tasks and set execution manager state to Cancelled."""
+        ...
```
SyntaxColoring, line 227: Could you elaborate on how this is different from `stop()` and what's the proper way to use it?

My perspective as a user of the hardware API is that its tasks and execution manager are implementation details. I don't know what its tasks are, or what they're doing, so I don't know when it's appropriate to cancel them, or how I need to follow that up.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -324,6 +324,15 @@ async def stop(self) -> None:
...
         self._queue_worker.cancel()
+        if self._hardware_api.is_movement_execution_taskified():
+            # We 'taskify' hardware controller movement functions when running protocols
+            # that are not backed by the engine. Such runs cannot be stopped by cancelling
+            # the queue worker and hence need to be stopped via the execution manager.
+            # `cancel_execution_and_running_tasks()` sets the execution manager in a CANCELLED state
+            # and cancels the running tasks, which raises an error and gets us out of the
+            # run function execution, just like `_queue_worker.cancel()` does for
+            # engine-backed runs.
+            await self._hardware_api.cancel_execution_and_running_tasks()
```
SyntaxColoring, line 335: Am I correct in assessing that this sequence of events is now possible?

1. `ProtocolEngine.stop()` is called
2. `ProtocolEngine.stop()` calls `self._hardware_api.cancel_execution_and_running_tasks()`
3. That does an asyncio cancel of the current `API.move_to()` task, or whatever the protocol was most recently doing
4. The `asyncio.CancelledError` bubbles up through the Python protocol thread and kills it
5. **But the hardware is still moving,** because we merely cancelled the task waiting for its completion, not the actual motion

Is that correct? If so, is that different from before, and are we worried about that?

api/src/opentrons/hardware_control/protocols/motion_controller.py
```diff
@@ -213,3 +213,15 @@ async def retract(self, mount: Mount, margin: float = 10) -> None:
     async def retract_axis(self, axis: Axis) -> None:
         """Retract the specified axis to its home position."""
         ...
+
+    def is_movement_execution_taskified(self) -> bool:
+        """Get whether move functions are being executed inside cancellable tasks."""
+        ...
```
SyntaxColoring, line 219: We talked about this a bit on the call: at some point, I hope we can unify things so there is only one mode—either always taskify, or never. I know having two modes is intended to reduce risk by avoiding touching too many things at once, but it seems like a substantial addition of complexity, from the caller's point of view. (This is on my mind because I'm guilty of the same kind of thing in #13425, and I sort of regret it.)

Is that unification part of the (short-, medium-, long-?)-term plan?

SyntaxColoring (approved):
Code review only—mostly questions.

The `ProtocolEngine` changes seem good if we're okay with the changes to the hardware API, which I'm probably not qualified to assess.

api/src/opentrons/protocol_engine/protocol_engine.py
sfoster1: That is correct and different from before, but the key is that the PE stop behavior then calls the smoothie halt.

api/src/opentrons/hardware_control/protocols/motion_controller.py
sfoster1: For my money the unification will occur when we deprecate and remove pre-2.14 code, or when we cut over the hardware controller to a separate process or something.

api/src/opentrons/hardware_control/protocols/motion_controller.py
sanni-t: `stop()` is something that would disable motors, cleanup the hardware state and put it in a state appropriate for end of run. While `cancel_execution_and_running_tasks` is only necessary to stop execution of a protocol mid-way.

api/src/opentrons/protocol_engine/protocol_engine.py
sanni-t: Yep, because the python protocol thread is killed, we immediately jump to `engine.finish` which halts the motion immediately. We've made the actual motion halt as part of the cleanup rather than `engine.stop` because doing the other way round was what lead to the race condition bug. Since the motion halt & reset happens in one cleanup process, we are able to make sure that it happens in the correct order.

PR #13768, opened 2023-10-12
chore(api): Fix Sphinx language warning

SyntaxColoring:
# Overview

Fixes this warning when running `make docs`:

> WARNING: Invalid configuration value found: 'language = None'. Update your configuration to a valid langauge code. Falling back to 'en' (English).

# Test Plan

I've done this:

* [x] Run `make docs` and make sure the warning's gone away.
* [x] Spot-check the API docs (APIv1 and APIv2) and make sure they're not flagrantly broken.

# Risk assessment

Very low.

### DIFF
--- a/api/docs/hardware/conf.py
+++ b/api/docs/hardware/conf.py
@@ -93,7 +93,7 @@ max_apiLevel = str(MAX_SUPPORTED_VERSION)
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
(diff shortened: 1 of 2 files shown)


ecormany (approved):
Thanks for doing a super-back-burner chore here. Someday we will kill all the warnings (today is not that day). But while we're at it we should do the same at https://github.com/Opentrons/opentrons/blob/9396efe3dc88ea1c053ebedbc235588b0519dce8/api/docs/hardware/conf.py#L96

Also, it's extremely funny to me that we're pinned on a version of Sphinx after they added this warning but before they fixed the `langauge` typo in the warning text.

PR #13771, opened 2023-10-12
fix(app): use single source of truth for data in pipette settings slideout

shlokamin:
# Overview

This PR changes the pipette settings slideout to both read from react query and write from react query, which should slow down the time it takes for the server to reflect accurate pipette settings after a user changes them

closes RQA-1768

# Changelog

- use single source of truth for data in pipette settings slideout

# Review requests

- In the app go to the device details page of an OT-2 with 7.0.1 installed on it
- Open up the pipette settings slideout from the pipette card on the instruments page
- Change a config value, and make sure the updated value is reflected in the pipette settings slideout

# Risk assessment

Low/med

### DIFF
--- a/api-client/src/pipettes/index.ts
+++ b/api-client/src/pipettes/index.ts
@@ -1,5 +1,6 @@
 export { getPipettes } from './getPipettes'
 export { getPipetteSettings } from './getPipetteSettings'
+export { updatePipetteSettings } from './updatePipetteSettings'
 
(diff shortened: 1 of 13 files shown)


ncdiehl11:
Tested successfully on a 7.0.1 OT-2 👍

smb2268 (approved):
read through this and it all lgtm! happy to have this all in react-query

PR #13775, opened 2023-10-12
feat(protocol-designer, components): prep work for deck config in PD

jerader:
partially addresses RAUT-743 RAUT-688

# Overview

Some deck prep and reorganization for protocol designer and wires up a bit of the deck map.

Screen Shot 2023-10-12 at 1 31 19 PM

# Test Plan

Review how i reorganized the `deckSetup` component in PD. I separated out the constants and utils so the component is more readable. 

With the deck modification ff turned on, create a flex protocol and add some stuff to the deck: add some modules and the staging areas/waste chute/trash bin. See how it looks. not all the logic is in place yet so it is possible to accidentally overlap the trash with a module. that will be wired up in a follow up ticket.

# Changelog

- create a `RobotCoordinateSpaceWithDOMCoords` component for PD 
- fix some of the index files so the fixtures can be exported easier
- add row 4 to slot labels in PD
- clean up the `deckSetup` component
- add some logic to render the waste chute/trash bin/staging areas

# Review requests

see test plan

# Risk assessment

low


components/src/hardware-sim/RobotCoordinateSpace/RobotCoordinateSpaceWithDOMCoords.tsx
```diff
@@ -0,0 +1,65 @@
...
+  viewBox?: string | null
+  deckDef?: DeckDefinition
+  children?: (
+    props: RobotCoordinateSpaceWithDOMCoordsRenderProps
+  ) => React.ReactNode
+}
+
+type GetRobotCoordsFromDOMCoords = RobotCoordinateSpaceWithDOMCoordsRenderProps['getRobotCoordsFromDOMCoords']
+
+export function RobotCoordinateSpaceWithDOMCoords(
```
jerader, line 24: I needed to create a new component for PD since I need the `getRobotCoordsFromDOMCoords` to be able to drag/drop labware. So this is basically a mix between `RobotCoordinateSpace` and `RobotWorkSpace`

components/src/hardware-sim/BaseDeck/index.ts
koji, line 1: nit
maybe ordering lines via vscode would be good.

components/src/hardware-sim/Deck/index.tsx
koji, line 1: nit
same as the above

components/src/hardware-sim/RobotCoordinateSpace/RobotCoordinateSpaceWithDOMCoords.tsx
```diff
@@ -0,0 +1,65 @@
...
+  props: RobotCoordinateSpaceWithDOMCoordsProps
+): JSX.Element | null {
+  const { children, deckDef, viewBox, ...restProps } = props
+  const wrapperRef = React.useRef<SVGSVGElement>(null)
+  const getRobotCoordsFromDOMCoords: GetRobotCoordsFromDOMCoords = (x, y) => {
+    if (!wrapperRef.current) return { x: 0, y: 0 }
```
koji: I guess doing null check might be general 🤔 if there is no strong reasons to do use `!wrapperRef.current`.

```suggestion
    if (wrapperRef.current == null) return { x: 0, y: 0 }
```

components/src/hardware-sim/RobotCoordinateSpace/RobotCoordinateSpaceWithDOMCoords.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const cursorPoint = wrapperRef.current.createSVGPoint()
+
+    cursorPoint.x = x
+    cursorPoint.y = y
+
+    return cursorPoint.matrixTransform(
+      wrapperRef.current.getScreenCTM()?.inverse()
+    )
+  }
+  if (!deckDef && !viewBox) return null
```
koji: deckDef would be fine, but for viewBox, doing null check would be better because of its type.  

```suggestion
  if (deckDef == null && viewBox == null) return null
```

components/src/hardware-sim/RobotCoordinateSpace/RobotCoordinateSpaceWithDOMCoords.tsx
jerader: ya i basically copied and pasted that logic from `RobotWorkSpace`, so the logic is a bit dated, I'll update it to your suggestion. I think it makes more sense to check for if it isn't null instead.

components/src/hardware-sim/RobotCoordinateSpace/RobotCoordinateSpaceWithDOMCoords.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    deckSlotsById = deckDef.locations.orderedSlots.reduce(
+      (acc, deckSlot) => ({ ...acc, [deckSlot.id]: deckSlot }),
+      {}
+    )
+    wholeDeckViewBox = `${viewBoxOriginX} ${viewBoxOriginY} ${deckXDimension} ${deckYDimension}`
+  }
+  return (
+    <Svg
+      viewBox={viewBox || wholeDeckViewBox}
```
koji, line 57: nit

```suggestion
      viewBox={viewBox != null || wholeDeckViewBox}
```

protocol-designer/src/components/DeckSetup/NullDeckState.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <div className={styles.deck_row}>
+      <div className={styles.deck_wrapper}>
+        <RobotWorkSpace
+          deckLayerBlocklist={DECK_LAYER_BLOCKLIST}
+          deckDef={deckDef}
+          viewBox={`${VIEWBOX_MIN_X} ${VIEWBOX_MIN_Y} ${VIEWBOX_WIDTH} ${VIEWBOX_HEIGHT}`}
+          width="100%"
+          height="100%"
+        >
+          {() => (
```
koji, line 34: question
Do we need `{() => <></>)}` in stead of React fragment?

koji:
One quick question.
As you know we use `Trash Bin` for Deck Configuration. Don't we need to unify the word?

koji (approved):
lgtm
I left a couple of comments but ran through the test plan and I didn't see any issues.

jerader:
> One quick question. As you know we use `Trash Bin` for Deck Configuration. Don't we need to unify the word?

Good catch @koji! I think you're right, it would be more clear to say `Trash Bin`. Let's align Brent on this as well since I used the trash fixture that he made.

koji:
@jerader 
maybe Waste chute image is needed to update.
https://www.figma.com/file/eMAHM8g07dtTKr2RsFjkzB/OCT-Release%3A-Touchscreen-%5BDeck-configuration-%26-Drop-tips%5D?type=design&node-id=85-121378&mode=design&t=PzSyuUinsjlHfaKX-0

Issue #9988, opened 2022-04-18
api: Thermocycler execute profile bug in opentrons_simulate 5.0.2


alexjoel42:
# Overview

We have an issue where there is a key error for any "execute_profile." 

# Steps to reproduce

1. Include tc_execute_profile with any quantity of steps on 5.x.x
2. Receive a software error 

```python3


metadata = {
    'protocolName': 'Thermocycler Example Protocol',
    'author': 'Opentrons <protocols@opentrons.com>',
    'source': 'Protocol Library',
    'apiLevel': '2.12'
    }


def run(protocol):
    tc_mod = protocol.load_module('thermocycler')
    # run profile
    profile = [
        {'temperature': 90, 'hold_time_seconds': 10}
    ]

    tc_mod.execute_profile(steps=profile, repetitions=1)


```

# Current behavior
```
Traceback (most recent call last):
  File "/Users/alexcopperman/opt/anaconda3/bin/opentrons_simulate", line 8, in <module>
    sys.exit(main())
  File "/Users/alexcopperman/opt/anaconda3/lib/python3.9/site-packages/opentrons/simulate.py", line 605, in main
    print(format_runlog(runlog))
  File "/Users/alexcopperman/opt/anaconda3/lib/python3.9/site-packages/opentrons/simulate.py", line 391, in format_runlog
    + command["payload"].get("text", "").format(**command["payload"])
KeyError: "'temperature'

```

# Expected behavior

Thermocycler starting 1 repetitions of cycle composed of the following steps: [{'temperature': 90, 'hold_time_seconds': 10}]

# work around 
install 4.7.0

amitlissack:
Same as #9799

FankangMeng:
Same problem for me when simulation protocol with a thermocycler profile. Any solutions?🐶

Traceback (most recent call last):
  File "/Users/fankangmeng/opt/anaconda3/bin/opentrons_simulate", line 8, in <module>
    sys.exit(main())
  File "/Users/fankangmeng/opt/anaconda3/lib/python3.8/site-packages/opentrons/simulate.py", line 605, in main
    print(format_runlog(runlog))
  File "/Users/fankangmeng/opt/anaconda3/lib/python3.8/site-packages/opentrons/simulate.py", line 391, in format_runlog
    + command["payload"].get("text", "").format(**command["payload"])
KeyError: "'temperature'"

mcous:
[...]

PR #13776, opened 2023-10-12
docs(api): Fix description of run log payload

SyntaxColoring:
# Overview

While working on #13709, I noticed some outdated documentation for `opentrons.simulate.simulate()` and `opentrons.execute.execute()`.

The docs say you're supposed to do this:

```python
command = ...  # get from opentrons.simulate.simulate() or opentrons.execute.execute()

run_log_text = command["payload"]["text"].format(command["payload"])
```

However, it's apparently been simplified to this now:

```python
command = ... # get from opentrons.simulate.simulate() or opentrons.execute.execute()

run_log_text = command["payload"]["text"]
```

I'm not sure how long this has been the case. PR #9799 suggests it's at least since v5.0.2 but doesn't provide further details ("`text` field is not a format string anymore"). I tried some Git archeology, but I couldn't get to the bottom of it quickly.

Unfortunately, if you try to do the old `.format()` thing today, it's actively harmful. If the string happens to contain any `{` or `}` characters, it will confuse `.format()` and raise a `KeyError`. This happens for Thermocycler commands (see issue #9988), and of course it can happen for `ProtocolContext.comment()` commands.

# Test Plan

* Preview changes on http://sandbox.docs.opentrons.com/fix_payload_docs/v2/.
   * http://sandbox.docs.opentrons.com/fix_payload_docs/v2/new_protocol_api.html#opentrons.simulate.simulate
   * http://sandbox.docs.opentrons.com/fix_payload_docs/v2/new_protocol_api.html#opentrons.execute.execute

# Changelog

* Document that you should just do `command["payload"]["text"]`.
* Document that you should *no longer* do `command["payload"]["text"].format(command["payload"])`.

# Review requests

* Do you see any other places where we encourage the `.format()` thing? Anything in the Support help center? Any demo files floating around?
* Does my flowery prose inspire joy in the depths of your heart?

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -328,9 +328,9 @@ def execute(  # noqa: C901
             'name': command_name,
             'payload': {
               'text': string_command_text,
-               # The rest of this struct is
-               # command-dependent; see
-               # opentrons.commands.commands.
+              # The rest of this struct is
+              # command-dependent; see
+              # opentrons.commands.commands.
              }
           }
 
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -338,7 +338,7 @@ def execute(  # noqa: C901
           In older software versions, ``payload["text"]`` was a
           `format string <https://docs.python.org/3/library/string.html#formatstrings>`_.
           To get human-readable text, you had to do ``payload["text"].format(**payload)``.
-          That usage is deprecated now. If ``payload["text"]`` happens to contain any
+          Don't do that anymore. If ``payload["text"]`` happens to contain any
           ``{`` or ``}`` characters, it can confuse ``.format()`` and cause it to raise a
           ``KeyError``.
 
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -316,11 +316,32 @@ def execute(  # noqa: C901
                       ``"debug"``, ``"info"``, ``"warning"``, or ``"error"``.
                       Defaults to ``"warning"``.
     :param emit_runlog: A callback for printing the runlog. If specified, this
-                        will be called whenever a command adds an entry to the
-                        runlog, which can be used for display and progress
-                        estimation. If specified, the callback should take a
-                        single argument (the name doesn't matter) which will
-                        be a dictionary (see below). Default: ``None``
+        will be called whenever a command adds an entry to the
+        runlog, which can be used for display and progress
+        estimation. If specified, the callback should take a
+        single argument (the name doesn't matter) which will
+        be a dictionary:
+
+        .. code-block:: python
+
+          {
+            'name': command_name,
+            'payload': {
+              'text': string_command_text,
+               # The rest of this struct is
+               # command-dependent; see
+               # opentrons.commands.commands.
+             }
+          }
+
+        .. note::
+          In older software versions, ``payload["text"]`` was a
+          `format string <https://docs.python.org/3/library/string.html#formatstrings>`_.
+          To get human-readable text, you had to do ``payload["text"].format(**payload)``.
+          That usage is deprecated now. If ``payload["text"]`` happens to contain any
+          ``{`` or ``}`` characters, it can confuse ``.format()`` and cause it to raise a
+          ``KeyError``.
+
     :param custom_labware_paths: A list of directories to search for custom labware.
                                  Loads valid labware from these paths and makes them available
                                  to the protocol context. If this is ``None`` (the default), and
@@ -333,22 +354,6 @@ def execute(  # noqa: C901
                               non-recursive contents of specified directories
                               are presented by the protocol context in
                               ``ProtocolContext.bundled_data``.
-
-    The format of the runlog entries is as follows:
-
-    .. code-block:: python
-
-        {
-            'name': command_name,
-            'payload': {
-                 'text': string_command_text,
-                  # The rest of this struct is command-dependent; see
-                  # opentrons.commands.commands. Its keys match format
-                  # keys in 'text', so that
-                  # entry['payload']['text'].format(**entry['payload'])
-                  # will produce a string with information filled in
-             }
-        }
     """
     stack_logger = logging.getLogger("opentrons")
     stack_logger.propagate = propagate_logs
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -315,9 +315,9 @@ def execute(  # noqa: C901
     :param log_level: The level of logs to emit on the command line:
                       ``"debug"``, ``"info"``, ``"warning"``, or ``"error"``.
                       Defaults to ``"warning"``.
-    :param emit_runlog: A callback for printing the runlog. If specified, this
+    :param emit_runlog: A callback for printing the run log. If specified, this
         will be called whenever a command adds an entry to the
-        runlog, which can be used for display and progress
+        run log, which can be used for display and progress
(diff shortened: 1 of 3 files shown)


api/src/opentrons/simulate.py
```diff
@@ -440,14 +440,22 @@ def simulate(
     Each dict element in the run log has the following keys:
 
         - ``level``: The depth at which this command is nested - if this an
```
ecormany: space hyphen space, le sigh. 
```suggestion
        - ``level``: The depth at which this command is nested. If this an
```

api/src/opentrons/simulate.py
```diff
@@ -440,14 +440,22 @@ def simulate(
...
-                     aspirate inside a mix inside a transfer, for instance,
-                     it would be 3.
-        - ``payload``: The command, its arguments, and how to format its text.
-                       For more specific details see
-                       ``opentrons.commands``. To format a message from
-                       a payload do ``payload['text'].format(**payload)``.
+          aspirate inside a mix inside a transfer, for instance, it would be 3.
```
ecormany, line 443: monospace the `3`? it's an integer value in a key–value pair. but on the other hand, it's just the number 3.

api/src/opentrons/simulate.py
```diff
@@ -440,14 +440,22 @@ def simulate(
...
+
+        - ``payload``: The command. The human-readable run log text is available at
+          ``payload["text"]``. The other keys of ``payload`` are command-dependent;
+          see ``opentrons.commands``.
+
+          .. note::
+            In older software versions, ``payload["text"]`` was a
```
ecormany, line 450: i think the PR description indicates that we may not know the exact history here, but do we at least have enough information to say whether this is older Python API versions, older robot software versions, or both?

api/src/opentrons/simulate.py
```diff
@@ -440,14 +440,22 @@ def simulate(
...
+            `format string <https://docs.python.org/3/library/string.html#formatstrings>`_.
+            To get human-readable text, you had to do ``payload["text"].format(**payload)``.
+            That usage is deprecated now. If ``payload["text"]`` happens to contain any
+            ``{`` or ``}`` characters, it can confuse ``.format()`` and cause it to raise a
+            ``KeyError``.
+
         - ``logs``: Any log messages that occurred during execution of this
-                    command, as a logging.LogRecord
+          command, as a :py:obj:`logging.LogRecord`.
```
ecormany: i like the link, but i wasn't expecting it to go to the Python standard library. maybe it should read: 
```suggestion
          command, as a standard Python :py:obj:`~logging.LogRecord`.
```

api/src/opentrons/execute.py
```diff
@@ -316,11 +316,32 @@ def execute(  # noqa: C901
                       ``"debug"``, ``"info"``, ``"warning"``, or ``"error"``.
                       Defaults to ``"warning"``.
     :param emit_runlog: A callback for printing the runlog. If specified, this
-                        will be called whenever a command adds an entry to the
-                        runlog, which can be used for display and progress
-                        estimation. If specified, the callback should take a
-                        single argument (the name doesn't matter) which will
-                        be a dictionary (see below). Default: ``None``
+        will be called whenever a command adds an entry to the
+        runlog, which can be used for display and progress
```
ecormany: i think we prefer `run log`, but this isn't consistent everywhere. we could follow up on it later.

api/src/opentrons/execute.py
```diff
@@ -316,11 +316,32 @@ def execute(  # noqa: C901
...
+            'payload': {
+              'text': string_command_text,
+               # The rest of this struct is
+               # command-dependent; see
+               # opentrons.commands.commands.
+             }
+          }
+
+        .. note::
+          In older software versions, ``payload["text"]`` was a
```
ecormany, line 338: see comments on simulate. i think the two notes should be identical in the end.

api/src/opentrons/execute.py
```diff
@@ -316,11 +316,32 @@ def execute(  # noqa: C901
...
+        single argument (the name doesn't matter) which will
+        be a dictionary:
+
+        .. code-block:: python
+
+          {
+            'name': command_name,
+            'payload': {
+              'text': string_command_text,
+               # The rest of this struct is
```
ecormany: should the `#` go one space further left? 🧐

ecormany (approved):
This looks good to me. The most important part is the inclusion of the new notes to ward people away from writing code that might break. All the little comments are possible improvements, but completely optional.

I would like to rewrite the prose earlier in each of these docstrings, but I think that's already covered in RTC-211.

api/src/opentrons/execute.py
```diff
@@ -316,11 +316,32 @@ def execute(  # noqa: C901
...
+               # command-dependent; see
+               # opentrons.commands.commands.
+             }
+          }
+
+        .. note::
+          In older software versions, ``payload["text"]`` was a
+          `format string <https://docs.python.org/3/library/string.html#formatstrings>`_.
+          To get human-readable text, you had to do ``payload["text"].format(**payload)``.
+          That usage is deprecated now. If ``payload["text"]`` happens to contain any
```
jwwojak: A question about text on line 341, "That usage is deprecated now."

- Do we mean "not recommended" (a common use of "deprecated") or something stronger, more along the lines of "you shouldn't use it at all" because it raises an error?

If this is a "don't do this" warning (instead of a "you can, but shouldn't" warning), then maybe make the text stronger -- "That usage|method|technique is obsolete and should not be used."

jwwojak (commented):
Added a comment about the use of "deprecated". Does the text need a stronger warning?

api/src/opentrons/execute.py
ecormany: Interesting point, especially since "deprecated" can't mean "slated for removal" from the API, because we're talking about using methods from the Python standard library. You can `format()` anything, it's a free country.

api/src/opentrons/simulate.py
SyntaxColoring: Unclear to me, sorry. :(

api/src/opentrons/execute.py
SyntaxColoring: Yeah, this is a weird one because most of the time, the old usage will happen to have no effect. Only sometimes, it has the effect of raising an error.

I think Joe is on the right track. We should tell people what to do, directly and clearly. How about:

> In older software versions, `payload["text"]` was a format string. To get human-readable text, you had to do ``payload["text"].format(**payload)``. Don't do that anymore. If ``payload["text"]`` happens to contain any ``{`` or ``}`` characters, it can confuse ``.format()`` and cause it to raise a``KeyError``.

"Don't do that anymore" is the instruction. "If `payload["text"]` happens to contain..." is its supporting reasoning.

api/src/opentrons/execute.py
SyntaxColoring: Updated this instance and the other obvious ones to "run log."

api/src/opentrons/execute.py
SyntaxColoring: 🕵️‍♂️

api/src/opentrons/simulate.py
SyntaxColoring: Heh, that's a fun question.

I'mmmmm going to leave it alone for now, I guess? But no opinion from me either way, going forward.

SyntaxColoring:
Failing tests appear to all be flaky Thermocycler integration tests.

PR #13778, opened 2023-10-12
refactor(app): Remove robot admin restart logic

mjhuff:
Closes RAUT-771, RQA-1788

# Overview

This PR removes the old timeout logic for checking if a robot takes too long to reconnect during an update. It has already replaced by new, generalized timeout logic that isn't specific to only restarts. This older code intermittently causes an issue where the robot registers as taking too long to reconnect just seconds after the restart step initiates, so removing this code is more of a priority.

### Old
10_12 update did not successfully reconnect

### New 

Screenshot 2023-10-12 at 2 51 05 PM

# Test Plan
- Change TIME_BEFORE_ALLOWING_EXIT_MS in RobotUpdateProgressModal to 1 second, then initiate a robot update. During the restart step, the new text should display.

# Changelog
- Cleaned up robot update restart timeout logic.

# Risk assessment
low

### DIFF
--- a/app/src/assets/localization/en/device_settings.json
+++ b/app/src/assets/localization/en/device_settings.json
@@ -226,7 +226,7 @@
   "resets_cannot_be_undone": "Resets cannot be undone",
   "restart_now": "Restart now?",
   "restart_robot_confirmation_description": "<span>It will take a few minutes for <bold>{{robotName}}</bold> to restart.</span>",
-  "restart_taking_too_long": "{{robotName}} restart is taking longer than expected. If the robot appears unresponsive, try restarting the robot manually.",
+  "restart_taking_too_long": "{{robotName}} restart is taking longer than expected. Check your robot's settings page to verify whether or not the update was successful. If the robot appears unresponsive, try restarting the robot manually.",
   "restarting_robot": "Install complete, robot restarting...",
   "resume_robot_operations": "Resume robot operations",
   "returns_your_device_to_new_state": "This returns your device to a new state.",
--- a/app/src/assets/localization/en/device_settings.json
+++ b/app/src/assets/localization/en/device_settings.json
@@ -226,7 +226,7 @@
   "resets_cannot_be_undone": "Resets cannot be undone",
   "restart_now": "Restart now?",
   "restart_robot_confirmation_description": "<span>It will take a few minutes for <bold>{{robotName}}</bold> to restart.</span>",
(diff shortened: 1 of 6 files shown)


app/src/assets/localization/en/device_settings.json
```diff
@@ -226,7 +226,7 @@
   "resets_cannot_be_undone": "Resets cannot be undone",
   "restart_now": "Restart now?",
   "restart_robot_confirmation_description": "<span>It will take a few minutes for <bold>{{robotName}}</bold> to restart.</span>",
-  "restart_taking_too_long": "{{robotName}} restart is taking longer than expected. If the robot appears unresponsive, try restarting the robot manually.",
+  "restart_taking_too_long": "{{robotName}} restart is taking longer than expected. Check your robot's settings page to verify whether or not the update was successful. If the robot appears unresponsive, try restarting the robot manually.",
```
mjhuff: @ecormany Totally open to any ideas on how to phrase this!

ecormany: Hmm. The robot settings page is already visible behind the modal. What exactly is the next action we want users to try in this scenario?

We should also keep in mind that we have a warning in our latest release notes that restarts may take a while due to database migration from 6.x -> 7.x. This string should be as general as possible — it should apply to all future versions — without contradicting this special case that we know a lot of users will have to go through soon.

mjhuff: So this text can also appear from the Devices page by clicking an update banner - a user can begin the robot update flow without necessarily being on the settings page. We'd ideally like users to check the update page to see if the update was applied (sometimes a robot won't properly tell the app that the update applied successfully, and we have to handle those cases). However, if the robot is offline, there's no way to tell the update went through without first restarting the robot... at this point, a manual restart would likely be required. 

And yes, completely agree about the migration notice. This string is meant to be generic and covers cases when the robot restarts and is unable to connect (wifi issues, hardware issues, etc etc).

shlokamin (approved):
makes sense to me

app/src/assets/localization/en/device_settings.json
ecormany: OK this is pure suggestion, or you can take Shlok's GTG and merge as is 😁 
```suggestion
  "restart_taking_too_long": "{{robotName}} is taking longer than expected to restart. Check the Advanced tab of its settings page to see whether it finished updating. If the robot is unresponsive, restart it manually.",
```

mjhuff: I like it, thanks! 📝

PR #13780, opened 2023-10-12
docs(api): Refactor instrument-context docstrings

jwwojak:
# Overview

This PR is for managing suggested revisions to the docstrings in `instrument_context.py`. This is descriptive, customer/user-facing text that appears in [Protocols and Instruments section](https://docs.opentrons.com/v2/new_protocol_api.html) of the API v2 reference. JIRA issue is RTC-204.

These are lightweight changes that revise text, grammar, style, punctuation. Changes also include anchors to longer, more expository API docs throughout because many of the docstrings are terse and/or brief. Linking to other docs helps provide readers with more context, information, code samples, and broader understanding of our API methods. In all cases, we've tried to change as little text as possible.

Excluded: any text related to warnings and errors (``raises``, ``runtime``, ``type``, etc.) or anything that is a combination of static text and dynamic code. Just revisions to main definitions text only.

# Test Plan

Sandbox: http://sandbox.docs.opentrons.com/instrument-context-docstrings/v2/new_protocol_api.html#

Testing consisted of making text revisions, pushing to the working branch, and reviewing text results in the sandbox.

# Changelog

Definition text changes throughout. Things like:

- Align text with corresponding method descriptions in the larger API docs.
- Adding anchor links to other sections.
- Punctuation (sentences ended without periods)
- Replacing units of measurement in text with abbreviations (µL replaces "microliter") and getting measurements styled properly (e.g. "20 mm" replaces "20mm"). 
- Really tried to be as conservative/light-touch as possible throughout.

# Review requests

A lot of this is small and intricate. Any second (or third or _n_) set of eyes reviewing helps.

# Risk assessment

Low, docs/text change only, at least I hope I've only changed text and not working code. heh.


jwwojak:
@ecormany 

Picking up and trying again. Ran Black. Output suggests this might be ok:

> joewojak@admins-Air-2 protocol_api % black instrument_context.py --diff --color
  All done! ✨ 🍰 ✨
  1 file would be left unchanged.

Also, resolved small text conflict around `air_gap`.

ecormany:
> Picking up and trying again. Ran Black. Output suggests this might be ok:

yep, and the "API test/lint/deploy" tests are passing in CI ✅

jwwojak:
Changing status from draft.

api/src/opentrons/protocol_api/instrument_context.py
ecormany, line 1: I think, in the spirit of "write the introduction last", it would be good to come back and revise the docstring for the class as a whole (lines 58–71) as well now.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -170,10 +171,9 @@ def aspirate(
                          robot will aspirate from the exact specified location.
                          If unspecified, the robot will aspirate from the
                          current position.
-        :param rate: A relative modifier for how quickly to aspirate liquid.
-                     The flow rate for this aspirate will be
-                     `rate` * :py:attr:`flow_rate.aspirate <flow_rate>`.
-                     If not specified, defaults to 1.0.
+        :param rate: How quickly a pipette aspirates liquid, measured in µL/s.
```
ecormany: This parameter is *not* measured in µL/s. 200 sets it to 200x the default rate, not 200 µL/s. 

The old copy is a little clunky though.
```suggestion
        :param rate: A multiplier for the default flow rate of the pipette.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -170,10 +171,9 @@ def aspirate(
...
+                     Calculated as `rate` * :py:attr:`flow_rate.aspirate <flow_rate>`.
```
ecormany: Picky, but I'd love to get rid of the unformatted asterisk indicating multiplication here. (It would be fine in a code snippet, but it's not). `times` or `multiplied by` would suffice.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +256,37 @@ def dispense(
...
-                         current position.
-        :param rate: A relative modifier for how quickly to dispense liquid.
-                     The flow rate for this dispense will be
-                     `rate` * :py:attr:`flow_rate.dispense <flow_rate>`.
-                     If not specified, defaults to 1.0.
+        :param location: Where to dispense liquid held in the pipette.
+                        If ``location`` is a :py:class:`.Well`,
+                        the robot will dispense into
+                        :py:obj:`well_bottom_clearance.dispense <well_bottom_clearance>`
+                        mm above the bottom of the well.
```
ecormany: "into clearance millimeters above the bottom of the well" has me 😵 …I don't think it's even grammatical. Maybe: 
```suggestion
                        If ``location`` is a :py:class:`.Well`,
                        the pipette will dispense at or above the center bottom of the well.
                        The distance from the bottom is specified by                 
                        :py:obj:`well_bottom_clearance.dispense <well_bottom_clearance>`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +256,37 @@ def dispense(
...
+                        If ``location`` is  :py:class:`.Location`
+                        (i.e. the result of :py:meth:`.Well.top` or
```
ecormany:
```suggestion
                        (e.g., the result of :py:meth:`.Well.top` or
```
this list is non-exhaustive. and a comma. 🤓

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +256,37 @@ def dispense(
...
+                        :py:meth:`.Well.bottom`), the robot will dispense into
+                        the specified position. If only a ``location`` is passed
```
ecormany: I think the sentence that starts on this line needs to go last, otherwise the antecedent of the null pronoun in `If unspecified...` gets totally lost.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +256,37 @@ def dispense(
...
+                        (e.g., ``instr.dispense(location=wellplate['A1'])``),
+                        all of the liquid aspirated into the pipette
+                        will be dispensed (amount is accessible through
+                        :py:attr:`current_volume`). If unspecified,
+                        the robot will dispense into the current position.
+        :param rate: How quickly a pipette dispenses liquid. Measured in µL/s.
```
ecormany, line 295: same as for `aspirate()`

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +256,37 @@ def dispense(
...
+                     Calculated as `rate` * :py:attr:`flow_rate.dispense <flow_rate>`.
+                     If not specified, defaults to 1.0. See also, :ref:`new-plunger-flow-rates`.
         :type rate: float
-        :param push_out: Continue past the plunger bottom to guarantee all liquid
-                        leaves the tip. Specified in microliters. By default, this value is None.
+        :param push_out: Continue past the plunger bottom to help ensure all liquid
```
ecormany, line 299: 👍 under-promise, over-deliver.

api/src/opentrons/protocol_api/instrument_context.py
ecormany, line 1: note at the end of `dispense()`: we could update it to match our style for including parentheses in method references, and to use `pipette` as our default `InstrumentContext` object in examples.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -377,25 +379,26 @@ def mix(
...
+
+        :param repetitions: Number of times to mix (default is 1).
+        :param volume: The volume to mix, measured in µL. If 0 or unspecified,
+                       defaults to the maximum volume for the pipette and its
+                       attached tip.
+        :param location: A well or a position relative to well
+                         (e.g, ``plate.rows()[0][0].bottom()``).
+                         If unspecified, the pipette will mix from
+                         its current position.
         :type location: types.Location
```
ecormany: this seems wrong. it's Well or Location. and it doesn't render anyway, so we should probably remove the line.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -125,20 +124,21 @@ def starting_tip(self, location: Union[labware.Well, None]) -> None:
...
+        """The speed at which the robot's gantry moves in mm/s.
 
-        By default, 400 mm/s. Changing this value will change the speed of the
-        pipette when moving between labware. In addition to changing the
+        The default speed for Flex varies between 300 and 350 mm/s.
+        The OT-2 default is 400 mm/s. In addition to changing the
         default, the speed of individual motions can be changed with the
-        ``speed`` argument to :py:meth:`InstrumentContext.move_to`.
+        ``speed`` argument of the :py:meth:`InstrumentContext.move_to` method.
+        See also, :ref:`gantry_speed`.
```
ecormany: super nit, i don't think there should ever be a comma after `See also`.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +523,31 @@ def touch_tip(
     ) -> InstrumentContext:
         """
         Touch the pipette tip to the sides of a well, with the intent of
-        removing left-over droplets
+        removing left-over droplets. See also, :ref:`touch-tip`.
```
ecormany: closed compound. and the comma thing.
```suggestion
        removing leftover droplets. See also :ref:`touch-tip`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +523,31 @@ def touch_tip(
...
 
         :param location: If no location is passed, pipette will
```
ecormany: would say "the pipette" here.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +523,31 @@ def touch_tip(
...
                        radius. When `radius=1.0`, the pipette tip will move to
-                       the edge of the target well; when `radius=0.5`, it will
+                       the edge of the target well. When `radius=0.5`, it will
                        move to 50% of the well's radius. Default: 1.0 (100%)
         :type radius: float
-        :param v_offset: The offset in mm from the top of the well to touch tip
-                         A positive offset moves the tip higher above the well,
-                         while a negative offset moves it lower into the well
-                         Default: -1.0 mm
+        :param v_offset: The offset in mm from the top of the well to touch tip.
```
ecormany: Is it clearer to say something like `How far above or below the well to touch the tip, measured in mm.`?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +523,31 @@ def touch_tip(
...
+                         A positive offset moves the tip higher above the well.
+                         A negative offset moves the tip lower into the well.
+                         Default is -1.0 mm.
         :type v_offset: float
         :param speed: The speed for touch tip motion, in mm/s.
-                      Default: 60.0 mm/s, Max: 80.0 mm/s, Min: 20.0 mm/s
+                      Default: 60.0 mm/s, Max: 80.0 mm/s, Min: 1.0 mm/s.
```
ecormany: maybe better as bullets

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +523,31 @@ def touch_tip(
...
         :type speed: float
         :raises: ``UnexpectedTipRemovalError`` -- if no tip is attached to the pipette
         :raises RuntimeError: If no location is specified and location cache is
```
ecormany, line 559: same as note on `blow_out()` above

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -602,25 +604,29 @@ def air_gap(
...
+        :param volume: The amount of air to aspirate, measured in µL.
+                       Calling ``air_gap()`` with no arguments uses
+                       the entire remaining volume in the pipette.
         :type volume: float
 
-        :param height: The number of millimeters to move above the current Well
-                       to air-gap aspirate. (Default: 5 mm above current Well)
+
+        :param height: The height, in mm, to move above the current well
+                       before aspirating air. The default is 5 mm above current well.
```
ecormany: terminology question here. we use `draw` as a more general verb to describe `asiprate()` above. it would make sense to mirror that here, instead of `pull`. 

i definitely want to avoid using `aspirate` as a verb for describing what `air_gap()` is doing. i think that introduces unnecessary confusion.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -602,25 +604,29 @@ def air_gap(
...
+
         :type height: float
 
         :raises: ``UnexpectedTipRemovalError`` -- if no tip is attached to the pipette
 
         :raises RuntimeError: If location cache is None.
```
ecormany, line 638: same as other errors above

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -654,7 +660,7 @@ def return_tip(self, home_after: Optional[bool] = None) -> InstrumentContext:
 
         This will not reset tip tracking, so the well flag will remain ``False``.
```
ecormany, line 676: I think this means `has_tip`? But we never use the term `well flag` anywhere else, so we should track down what this means for sure and state it clearly.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -681,8 +687,9 @@ def pick_up_tip(
     ) -> InstrumentContext:
         """
         Pick up a tip for the pipette to run liquid-handling commands.
+        See also, :ref:`basic-tip-pickup`.
 
-        If no location is passed, the Pipette will pick up the next available
+        If no location is passed, the pipette will pick up the next available
         tip in its :py:attr:`InstrumentContext.tip_racks` list.
```
ecormany: prob add a tilde here for readability.

api/src/opentrons/protocol_api/instrument_context.py
ecormany, line 1: in `drop_tip()` there are several instances of the word Location surrounded by single backticks instead of double backticks. (lines 691-699)

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -700,8 +707,8 @@ def pick_up_tip(
           was not specified when creating the :py:class:`.InstrumentContext`.
```
ecormany: the `tip_racks` property is mutable, so this isn't really about creation time so much as it's what the elements of the `tip_racks` list are right now.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -700,8 +707,8 @@ def pick_up_tip(
...
 
         * If you want to pick up the next available tip(s) in a specific
```
ecormany: i'd love to kill the `(s)` if possible

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -700,8 +707,8 @@ def pick_up_tip(
...
-          tip rack, you may use the tip rack directly:
-          e.g. ``instr.pick_up_tip(tiprack)``
+          tip rack, you may use the tip rack directly
+          (e.g. ``instr.pick_up_tip(tiprack)``).
 
         * If the position to move to in the well needs to be specified,
```
ecormany, line 725: this is 🤯 wild. i wonder if anyone uses this, or if we even really want to support it. a matter for another day, probably.

nevertheless, `:type location:` needs to be edited to not display weirdly. `presses`, `increment`, and `prep_after` also need edits.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: also, all of this info about ``location`` should be _within_ the description of ``:param location:``.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -862,7 +869,7 @@ def drop_tip(
         home_after: Optional[bool] = None,
     ) -> InstrumentContext:
         """
-        Drop the current tip.
+        Drop the current tip. See also, :ref:`pipette-drop-tip`.
 
         If no location is passed, the Pipette will drop the tip into its
```
ecormany:
```suggestion
        If no location is passed, the pipette will drop the tip into its
```
as above

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -862,7 +869,7 @@ def drop_tip(
...
         :py:attr:`trash_container`, which if not specified defaults to
```
ecormany: the following text needs some Flex vs. OT-2 clarification

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -877,19 +884,19 @@ def drop_tip(
 
             - If the only thing to specify is which well into which to drop
               a tip, `location` can be a :py:class:`.Well`. For instance,
```
ecormany: more double backticks needed here, and a few lines up

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -877,19 +884,19 @@ def drop_tip(
...
-              if you have a tip rack in a variable called `tiprack`, you can
+              if you have a tip rack in a variable called ``tiprack``, you can
               drop a tip into a specific well on that tiprack with the call
-              `instr.drop_tip(tiprack.wells()[0])`. This style of call can
+              ``instr.drop_tip(tiprack.wells()[0])``. This style of call can
               be used to make the robot drop a tip into arbitrary labware.
             - If the position to drop the tip from as well as the
               :py:class:`.Well` to drop the tip into needs to be specified,
```
ecormany: this is such a weird and wordy phrasing. can definitely be tightened up.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -877,19 +884,19 @@ def drop_tip(
...
               for instance to tell the robot to drop a tip from an unusually
-              large height above the tiprack, `location`
+              large height above the tiprack, ``location``
               can be a :py:class:`.types.Location`; for instance, you can call
               `instr.drop_tip(tiprack.wells()[0].top())`.
```
ecormany: is this unusually high? i would have assumed this was the default! we should find out and say what the default is.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -877,19 +884,19 @@ def drop_tip(
...
+              if you have a tip rack in a variable called ``tiprack``, you can
               drop a tip into a specific well on that tiprack with the call
-              `instr.drop_tip(tiprack.wells()[0])`. This style of call can
+              ``instr.drop_tip(tiprack.wells()[0])``. This style of call can
               be used to make the robot drop a tip into arbitrary labware.
             - If the position to drop the tip from as well as the
               :py:class:`.Well` to drop the tip into needs to be specified,
               for instance to tell the robot to drop a tip from an unusually
-              large height above the tiprack, `location`
+              large height above the tiprack, ``location``
```
ecormany:
```suggestion
              large height above the tip rack, ``location``
```
🧐

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -877,19 +884,19 @@ def drop_tip(
...
               can be a :py:class:`.types.Location`; for instance, you can call
               `instr.drop_tip(tiprack.wells()[0].top())`.
 
         :param location:
-            The location to drop the tip
+            The location to drop the tip.
         :type location:
             :py:class:`.types.Location` or :py:class:`.Well` or None
```
ecormany:
```suggestion
            :py:class:`.types.Location` or :py:class:`.Well` or ``None``
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -904,12 +911,12 @@ def drop_tip(
 
                 * You're using a GEN2 pipette, not a GEN1 pipette.
                 * You've tested ``home_after=False`` extensively with your
-                  particular pipette and your particular tips.
+                  pipette and tips.
                 * You understand the risks described below.
 
             The ejector shroud that pops the tip off the end of the pipette is
```
ecormany: another shroud! let's send it to the grave too.
```suggestion
            The ejector that pops the tip off the end of the pipette is
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -980,7 +987,7 @@ def drop_tip(
 
     @requires_version(2, 0)
     def home(self) -> InstrumentContext:
-        """Home the robot.
+        """Home the robot. See also, :ref:`utility-homing`.
```
ecormany: "Home the robot" sounds like it homes _everything_. Does the left pipette vertical axis home if you call this on the right pipette? OK to follow up later if we don't have a quick answer.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1342,7 +1349,7 @@ def move_to(
     @property  # type: ignore
     @requires_version(2, 0)
     def mount(self) -> str:
-        """Return the name of the mount this pipette is attached to"""
+        """Return the name of the mount this pipette is attached to."""
```
ecormany: I presume the only possibilities are `left` and `right`? Might be worth listing if so.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1458,7 +1465,7 @@ def trash_container(self, trash: labware.Labware) -> None:
     @requires_version(2, 0)
     def name(self) -> str:
         """
-        The name string for the pipette (e.g. 'p300_single')
+        The name string for the pipette (e.g. 'p300_single).
```
ecormany:
```suggestion
        The name string for the pipette (e.g., ``p300_single``).
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1479,10 +1486,10 @@ def min_volume(self) -> float:
     @requires_version(2, 0)
     def max_volume(self) -> float:
```
ecormany, line 1484: we should also document `min_volume()` above, unless it's not supported for some reason.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1524,7 +1531,7 @@ def hw_pipette(self) -> PipetteDict:
     @property  # type: ignore
     @requires_version(2, 0)
     def channels(self) -> int:
-        """The number of channels on the pipette."""
+        """The number of channels on the pipette. See also, :ref:`new-pipette`."""
```
ecormany: add that it's 1, 8, or 96?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1524,7 +1531,7 @@ def hw_pipette(self) -> PipetteDict:
...
         return self._core.get_channels()
 
     @property  # type: ignore
```
ecormany, line 1536: i'd never heard of this, but it's just a getter, so it's probably fine to leave as is. except `tiprack` in line 1540.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1584,27 +1591,21 @@ def __str__(self) -> str:
...
     @requires_version(2, 15)
     def configure_for_volume(self, volume: float) -> None:
-        """Configure a pipette to handle a specific volume of liquid, specified in µL. Depending on the
-        volume, the pipette will enter a certain pipetting mode. Changing pipette modes alters properties
-        of the instance of :py:class:`.InstrumentContext`, such as default flow rate, minimum volume, and
-        maximum volume. The pipette will remain in the mode set by this function until it is called again.
-
-        The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a low-volume mode
-        to accurately dispense 1 µL of liquid. Low-volume mode can only be set by calling this function.
+        """Configure a pipette to handle a specific volume of liquid, measured in µL. Depending on the volume, the pipette will enter a unique pipetting mode.
```
ecormany: not sure i like "unique" here. there's two modes, and that makes it sound like there's a lot more. stet or `The pipette enters a volume mode depending on the volume provided.`

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1584,27 +1591,21 @@ def __str__(self) -> str:
...
+        Changing pipette modes alters properties of the instance of
+        :py:class:`.InstrumentContext`, such as default flow rate, minimum volume,
+        and maximum volume. The pipette remains in the mode set by this function until it is called again.
 
-        For more information on available modes for certain pipettes, see :ref:`pipette-volume-modes`.
+        The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a low-volume mode to accurately dispense 1 µL of liquid. Low-volume mode can
+        only be set by calling ``configure_for_volume``. See also, :ref:`pipette-volume-modes`.
```
ecormany: i like the demonstrative pronoun better. indicates "you are already here". i could see reading this and thinking "oh i have to look up another function"

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1584,27 +1591,21 @@ def __str__(self) -> str:
...
 
         .. note ::
 
-            Changing a pipette's mode will reset its :ref:`flow rates <new-plunger-flow-rates>`.
+            Changing a pipette's mode will reset its :ref:`flow rate <new-plunger-flow-rates>`.
```
ecormany: stet. all the flow rates (aspirate, dispense, blowout). yes, these happen to be identical for the Flex pipettes in question, but that could change.

ecormany (commented):
This is big, so I left lots of comments. Mostly small stuff on the edited parts. A few methods and parameters were passed over entirely and need a first round of edits.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: A great topic for discussion, and a mushy one as this might be more style than rules.

We ought to settle on this as a style - comma or no comma. I've always used `See<comma> rest of sentence` as if it were a signal like "For example, . . . " so "See,  . . . " or "See also, . . . " is ok. To me it's the difference between mashing things together or separating them:

- SeeTitleofThingHere
vs
- See `<comma pause>` Title of Thing Here.

I think lawyerly types may not use this much e.g., "See _name of case law thing here_", but we're not doing legal writing.

This is one of those things like using a comma with i.e. or e.g.  Either way is acceptable, IMO.

- i.e.,   or    i.e.
- e.g., or e.g.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Yikes, italics.  _location_

Should be ``location``

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: changed to remove commas _and_ these should just be "See _name of thing_" and not "See also _name of thing_".

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Fixing. Will use "multiplied by".  Found multiple instances and some hidden single quote for code instead of ``.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Fixing `dispense()`.

Keeping as reminder to check `aspirate()`.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: In the words of the inimitable Mr. Torgue, "[That sentence has too many syllables](https://youtu.be/UJj85ToIC7Y?feature=shared&t=10). Apologize!"  😄

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -125,20 +124,21 @@ def starting_tip(self, location: Union[labware.Well, None]) -> None:
 
     @requires_version(2, 0)
     def reset_tipracks(self) -> None:
-        """Reload all tips in each tip rack and reset starting tip"""
+        """Reloads all tips in each tip rack and resets the starting tip."""
```
SyntaxColoring: The first line of Python docstrings is conventionally imperative ("reload"/"reset" instead of "reloads"/"resets"), unfortunately.

See [PEP 257](https://peps.python.org/pep-0257/#:~:text=The%20docstring%20is%20a%20phrase%20ending%20in%20a%20period.%20It%20prescribes%20the%20function%20or%20method%E2%80%99s%20effect%20as%20a%20command%20(%E2%80%9CDo%20this%E2%80%9D%2C%20%E2%80%9CReturn%20that%E2%80%9D)%2C%20not%20as%20a%20description%3B%20e.g.%20don%E2%80%99t%20write%20%E2%80%9CReturns%20the%20pathname%20%E2%80%A6%E2%80%9D.).

ecormany: Welp, standards gonna standard. We can comply 🫡

api/src/opentrons/protocol_api/instrument_context.py
SyntaxColoring: > this is 🤯 wild. i wonder if anyone uses this, or if we even really want to support it. a matter for another day, probably.

Picking up / dropping a tip with an offset, you mean?

In the Python Protocol API, I think it's helpful in environments like Jupyter Notebook where you want to be able to adjust things like that on-the-fly and experiment with different offsets. Outside the Python Protocol API, the system does this during Labware Position Check.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: 🧐 linguistics nerd alert 🤓 
i think they're bare vPs, not imperative clauses, but whatever. i've [ticketed](https://opentrons.atlassian.net/browse/RTC-359) this for a followup, since i spotted some other TP clauses in the docstrings of other classes (e.g., `HeaterShakerContext`):

jwwojak: 🖖 fixing this.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Revising `aspirate` to match `dispense` in the Note callout.

**Dispense**
> If ``dispense`` is called with a single argument, it will not try
            to guess whether the argument is a volume or location - it is
            required to be a volume. If you want to call ``dispense`` with only
            a location, specify it as a keyword argument:
            ``pipette.dispense(location=plate['A1'])``

**Aspirate**
> If ``aspirate`` is called with a single argument, it will not try
            to guess whether the argument is a volume or location - it is
            required to be a volume. If you want to call ``aspirate`` with only
            a location, specify it as a keyword argument:
            ``pipette.aspirate(location=plate['A1'])``

For `dispense`, we also discussed shortening the location code example in the text for both methods:

Change:
> (as in instr.dispense(location=plate['A1']))

To:
> (e.g., (location=plate['A1']))

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Re-wrote `:param location:` for `aspirate` and `dispense`. Make descriptions more parallel. `aspirate` was a bit different. I think this is better than before.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Rewrote to so `aspirate` and `dispense` are similar.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Removing line 390.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: 😨 yikes. Fixing this. 

> :param location: If no location is passed, **the** pipette will . . .

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Yes. Better than just adding a period, which is all I did.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Yes, let's try it! Touchy syntax.

Text
New line
4 space indentation

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Added ~ to shorten link to API reference:

> the pipette will pick up the next available tip in its :py:attr:`~.InstrumentContext.tip_racks` list.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Oh, I hate those, the (s). Let's kill it with fire. Just use the plural even if something can be singular. Readers will understand. Save the (s) for the lawyers.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Changed "tiprack" to "tip rack."

jwwojak: Checking throughout file for instances of "tiprack." Will only change text, not code (e.g. `tiprack`, or `if parent_labware.is_tiprack`)

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Revised using the recommended text.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: leaving text as is.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Checked in Slack (python-protocol-api). Seth says the missing text is probably an oversight. I will add a minimum description.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Yes, checked on that. Will add.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: @ecormany I wanted to tag you regarding this `drop_tip` part of the API reference.

`:param home_after:` has this _War and Peace_ length explanation. It goes into detail about problems with setting `home_after=false`. There's enough extra text this could be a standalone support article. Rather including this info in the API reference, I had this idea:

**Part 1**
For `:param home_after:` we say the following:

> Whether to home the pipette's plunger after dropping the tip. If not specified, defaults to ``True`` on an OT-2. Setting home_after=false is not recommended. See _link to new article TBD_.

(I don't know what a Flex does by default, yet.)

**Part 2**
Next, we take all the info about the problems setting `home_after=false` and put that in a new doc somewhere else. It doesn't have to be in our API docs. It could be a short support team article. I could create this and work with those guys to get it posted somewhere.

I'm recommending this because all the info makes `drop_tip` very long because of this (kinda obscure) argument that it accepts.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: ❓

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: This got hidden as "Outdated." Fixing now.

> Draw air into the pipette's tip at the current location. See :ref:`air-gap`.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Fixing.

> The amount of air to draw, measured in µL.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: I'm not sure what to look for.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: ❓

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: I think these are ok now, though I'm lost among all the changes.

> nevertheless, :type location: needs to be edited to not display weirdly. presses, increment, and prep_after also need edits.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: I think I was referring to making the parallel changes here in the explanation of the runtime error:

- [x] punctuate `eg`
- [x] double backticks around method name

api/src/opentrons/protocol_api/instrument_context.py
ecormany: should actually say 
```
``air_gap`` 
```

api/src/opentrons/protocol_api/instrument_context.py
ecormany: maybe:
```
You can use this style of call to pick up a tip from a rack that is not a member of :py:obj:`.InstrumentContext.tip_racks`.
```

api/src/opentrons/protocol_api/instrument_context.py
ecormany: just remove `to pick up a tip from.` in line 727

api/src/opentrons/protocol_api/instrument_context.py
ecormany: let's also add to the min definition:
```
This value may change based on the :ref:`volume mode <FILL IN ANCHOR>` that the pipette is currently configured for.
```

api/src/opentrons/protocol_api/instrument_context.py
ecormany: oops, i meant leave the original: `rates`

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Updating text.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Added text and link to volume mode.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: stetted 😉

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Removed, but I also don't see the exact quoted text. Very odd.  Lines 726-730:

> The number of times to lower and then raise the pipette when picking up a tip, to ensure a good seal (0 [zero] will result in the pipette hovering over the tip but not picking it up--generally not desirable, but could be used for dry-run).

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Revised text.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: From @Laura-Danielle via Slack:
> The way that the protocol api determines whether a tip ‘exists’ in a tiprack is by this concept of a well_flag. If the protocol api sees that a well is marked as ‘False’ (i.e. empty) then it will not attempt to pick up from that location. Instead it will try to find a location that has a ‘True’ (i.e. filled) location in the tiprack.

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Took a run at a generic intro for this section. It might be kind of "meh." I figure most readers _aren't_ going to rely on it to find what they need. Instead, they may get to individual class objects via search or from links on other API docs.

jwwojak:
I've seen formatting issues in the labware class too, but that's out of scope for this PR, which covers the InstrumentContext.

For example [columns_by_name](https://docs.opentrons.com/v2/new_protocol_api.html#opentrons.protocol_api.Labware.columns_by_name), code syntax using single quotes.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -55,18 +55,17 @@
 
 
 class InstrumentContext(publisher.CommandPublisher):
-    """A context for a specific pipette or instrument.
```
ecormany, line 58: We have to leave the one-line summary. This is a Python docstring format standard.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -55,18 +55,17 @@
...
-
-    This can be used to call methods related to pipettes - moves or
-    aspirates or dispenses, or higher-level methods.
+    """
+       The InstrumentContext class in the Opentrons Python API (PAPI) provides the objects,
+       attributes, and methods that allow you to use pipettes and other instruments in your
+       protocols. Similar to other class containers, InstrumentContext is temporary.
+       Class objects only exist when instantiated in your Python protocol.
```
ecormany: A couple things:
- "PAPI" is internal shorthand only.
- "temporary" is maybe too strong — you can't switch pipettes during a protocol, so they persist for its duration

Maybe replace this ¶ and the next one:
```suggestion
    """
       The InstrumentContext class provides the objects,
       attributes, and methods that allow you to use pipettes in your protocols. 
       
       Methods generally fall into one of two categories. 
       
         - They can change the state of the InstrumentContext object, such as how fast 
         it moves liquid or where it disposes of used tips). 
         - They can command the instrument to perform an action, like picking up tips, moving
         to certain locations, and aspirating or dispensing liquid. 
```
check my formatting, i composed this directly here in github.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -154,26 +153,28 @@ def aspirate(
         rate: float = 1.0,
     ) -> InstrumentContext:
         """
-        Aspirate a given volume of liquid from the specified location, using
-        this pipette.
+        Draw liquid into a pipette tip. Includes parameters that control
+        aspiration volume, well location, tip position in a well,and pipette flow rate.
+        See :ref:`new-aspirate`.
```
ecormany: Summarizing the immediately following parameter list is redundant. And let's keep the first line special.
```suggestion
        Draw liquid into a pipette tip.
        
        See :ref:`new-aspirate` for more details and examples.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -154,26 +153,28 @@ def aspirate(
...
-                     If not specified, defaults to 1.0.
+        :param location: Tells the robot where to aspirate from. The location can be a
+                         :py:class:`.Well` or a :py:class:`.Location`.
+
+                            - If the location is a ``well``, the robot will aspirate at or above
+                              the center of the bottom of the well. The distance (in mm) from the
+                              well bottom is specified by the :py:obj:`well_bottom_clearance.aspirate` object.
+
+                            - If the location is a ``location`` (e.g., the result of :py:meth:`.Well.top`
+                              or :py:meth:`.Well.bottom`), the robot will aspirate from that specified position.
```
ecormany: the classes `Well` and `Location` should be capitalized. it wouldn't matter overly much, except here lowercase `location` is the name of the _parameter_ in question. (naming things is hard.)
```suggestion
                            - If the location is a ``Well``, the robot will aspirate at or above
                              the bottom center of the well. The distance (in mm) from the
                              well bottom is specified by :py:obj:`well_bottom_clearance.aspirate`.

                            - If the location is a ``Location`` (e.g., the result of :py:meth:`.Well.top`
                              or :py:meth:`.Well.bottom`), the robot will aspirate from that specified position.
```
🐍 🤓 `well_bottom_clearance.aspirate` isn't an object, it's a parameter of an object.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -183,7 +184,7 @@ def aspirate(
             to guess whether the argument is a volume or location - it is
```
ecormany, line 190: This note could be more succinct, but it's fine.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +257,41 @@ def dispense(
         push_out: Optional[float] = None,
     ) -> InstrumentContext:
         """
-        Dispense a volume of liquid (in microliters/uL) using this pipette
-        into the specified location.
+        Dispense liquid from a pipette tip. Includes parameters that control
+        dispense volume, well location, tip position in a well, and pipette
+        flow rate. See :ref:`new-dispense`.
```
ecormany: like aspirate above
```suggestion
        Dispense liquid from a pipette tip.

        See :ref:`new-dispense` for more details and examples.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +257,41 @@ def dispense(
...
-        aspirated into the pipette will be dispensed (this volume is accessible
-        through :py:attr:`current_volume`).
+        :param location: Tells the robot where to dispense liquid held in the pipette.
+                         The location can be a :py:class:`.Well` or a
+                         :py:class:`.Location`.
 
-        :param volume: The volume of liquid to dispense, in microliters. If 0
-                       or unspecified, defaults to :py:attr:`current_volume`.
-        :type volume: int or float
+                            - If the location is a ``well``, the pipette will dispense at or above
```
ecormany, line 281: Same as with aspirate above

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -256,35 +257,41 @@ def dispense(
...
+                              the center of the bottom of the well. The distance from the well bottom
+                              (in mm) is specified by the :py:obj:`well_bottom_clearance.dispense <well_bottom_clearance>`
+                              object.
+
+                            - If the location is a ``location`` (e.g., the result of :py:meth:`.Well.top` or
+                              :py:meth:`.Well.bottom`), the robot  will dispense into the specified position.
```
ecormany:
```suggestion
                              :py:meth:`.Well.bottom`), the robot will dispense into the specified position.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -377,25 +384,25 @@ def mix(
...
-                         the pipette will mix from its current position.
-        :type location: types.Location
-        :param rate: A relative modifier for how quickly to aspirate and
-                     dispense liquid during this mix. When aspirating, the flow
-                     rate will be
-                     `rate` * :py:attr:`flow_rate.aspirate <flow_rate>`,
-                     and when dispensing, it will be
-                     `rate` * :py:attr:`flow_rate.dispense <flow_rate>`.
+        Mixes a volume of liquid by repeatedly aspirating and dispensing it in a single location.
+        See :ref:`mix`.
```
ecormany:
```suggestion
        Mix a volume of liquid by repeatedly aspirating and dispensing it in a single location.
        
        See :ref:`mix`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -377,25 +384,25 @@ def mix(
...
+
+        :param repetitions: Number of times to mix (default is 1).
+        :param volume: The volume to mix, measured in µL. If 0 or unspecified,
+                       defaults to the maximum volume for the pipette and its
+                       attached tip.
+        :param location: A well or a position relative to well
```
ecormany, line 401: It would probably be good to use the `Well or Location` terminology from other methods.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -445,25 +452,24 @@ def blow_out(
         self, location: Optional[Union[types.Location, labware.Well]] = None
     ) -> InstrumentContext:
         """
-        Blow liquid out of the tip.
+        Blows an extra amount of air through a pipette's tip to clear it.
```
ecormany:
```suggestion
        Blow an extra amount of air through a pipette's tip to clear it.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -445,25 +452,24 @@ def blow_out(
...
 
-        :param location: The location to blow out into. If not specified,
-                         defaults to the current location of the pipette
+        :param location: The blowout location. If no location is specified,
+                         the pipette will blow out from its current position.
         :type location: :py:class:`.Well` or :py:class:`.Location` or None
 
         :raises RuntimeError: If no location is specified and location cache is
-                              None. This should happen if `blow_out` is called
+                              None. This should happen if ``blow_out`` is called
```
ecormany:
```suggestion
                              ``None``. This should happen if ``blow_out`` is called
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -484,7 +490,7 @@ def blow_out(
         if isinstance(target, validation.WellTarget):
             if target.well.parent.is_tiprack:
                 _log.warning(
-                    "Blow_out being performed on a tiprack. "
+                    "Blow_out being performed on a tip rack. "
```
ecormany: This isn't a docstring and could have an effect on tests. Revert.
```suggestion
                    "Blow_out being performed on a tiprack. "
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +527,34 @@ def touch_tip(
...
+                         A positive offset moves the tip higher above the well.
+                         A negative offset moves the tip lower into the well.
+                         Default is -1.0 mm.
         :type v_offset: float
         :param speed: The speed for touch tip motion, in mm/s.
-                      Default: 60.0 mm/s, Max: 80.0 mm/s, Min: 20.0 mm/s
+
+                        - Default: 60.0 mm/s
+                        - Max: 80.0 mm/s
+                        - Min: 1.0 mm/s
```
ecormany: I like spelling these out.
```suggestion
                        - Maximum: 80.0 mm/s
                        - Minimum: 1.0 mm/s
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -521,31 +527,34 @@ def touch_tip(
...
         :type speed: float
         :raises: ``UnexpectedTipRemovalError`` -- if no tip is attached to the pipette
         :raises RuntimeError: If no location is specified and location cache is
-                              None. This should happen if `touch_tip` is called
+                              None. This should happen if ``touch_tip`` is called
```
ecormany:
```suggestion
                              ``None``. This should happen if ``touch_tip`` is called
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -579,7 +588,7 @@ def touch_tip(
             return self
         if parent_labware.is_tiprack:
             _log.warning(
-                "Touch_tip being performed on a tiprack. Please re-check your code"
+                "Touch_tip being performed on a tip rack. Please re-check your code"
```
ecormany: Not a docstring, could affect tests. Revert.
```suggestion
                "Touch_tip being performed on a tiprack. Please re-check your code"
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -602,25 +611,28 @@ def air_gap(
         self, volume: Optional[float] = None, height: Optional[float] = None
     ) -> InstrumentContext:
         """
-        Pull air into the pipette current tip at the current location
+        Draw air into the pipette's tip at the current location. See :ref:`air-gap`.
```
ecormany: Param description make clear this happens at the same _well_ but not at the same exact location (i.e., the pipette moves to get out of the well so it doesn't air-gap in a location full of liquid).
```suggestion
        Draw air into the pipette's tip at the current well.

        See :ref:`air-gap`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -602,25 +611,28 @@ def air_gap(
...
 
-        :param volume: The amount in uL to aspirate air into the tube.
-                       (Default will use all remaining volume in tip)
+        :param volume: The ``air_gap``, measured in µL.
```
ecormany: Too circular.
```suggestion
        :param volume: The amount of air, measured in µL.
```

api/src/opentrons/protocol_api/instrument_context.py
ecormany: it must be `has_tip`, that's the only bool in the Well class

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -862,75 +871,46 @@ def drop_tip(
         home_after: Optional[bool] = None,
     ) -> InstrumentContext:
         """
-        Drop the current tip.
+        Drop the current tip. See :ref:`pipette-drop-tip`.
```
ecormany:
```suggestion
        Drop the current tip.

        See :ref:`pipette-drop-tip`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -862,75 +871,46 @@ def drop_tip(
...
 
-        If no location is passed, the Pipette will drop the tip into its
-        :py:attr:`trash_container`, which if not specified defaults to
-        the fixed trash in slot 12.  From API version 2.15 on, if the trash container is
-        the default fixed trash in A3 (slot 12), the API will default to
-        dropping tips in different points within the trash container
-        in order to prevent tips from piling up in a single location in the trash.
+        If no location is passed (e.g. ``pipette.drop_tip()``), the pipette will drop the
+        attached tip into its default :py:attr:`trash_container`. For the Flex and OT-2 the
+        default slots for their in-deck trash containers are A3 and 12, respectively.
```
ecormany: i wouldn't characterize the OT-2 as having an "in-deck" trash container.
```suggestion
        default slots for their trash containers are A3 and 12, respectively.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -980,7 +960,7 @@ def drop_tip(
 
     @requires_version(2, 0)
     def home(self) -> InstrumentContext:
-        """Home the robot.
+        """Home the robot. See :ref:`utility-homing`.
```
ecormany:
```suggestion
        """Home the robot.

        See :ref:`utility-homing`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1342,7 +1322,10 @@ def move_to(
     @property  # type: ignore
     @requires_version(2, 0)
     def mount(self) -> str:
-        """Return the name of the mount this pipette is attached to"""
+        """
+        Return the name of the mount the pipette is attached to. The names are
+        ``left`` and ``right``.
```
ecormany:
```suggestion
        Return the name of the mount the pipette is attached to.

        The possible names are ``left`` and ``right``.
```

ecormany: note on which the 96-channel reports?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1388,7 +1371,7 @@ def speed(self) -> "PlungerSpeeds":
     @property  # type: ignore
     @requires_version(2, 0)
     def flow_rate(self) -> "FlowRates":
-        """The speeds (in uL/s) configured for the pipette.
+        """The speeds, in µL/s, configured for the pipette. See :ref:`new-plunger-flow-rates`.
```
ecormany:
```suggestion
        """The speeds, in µL/s, configured for the pipette.

        See :ref:`new-plunger-flow-rates`.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1403,7 +1386,7 @@ def flow_rate(self) -> "FlowRates":
           same values, setting one will override the other.
 
         For instance, to change the flow rate for aspiration on an instrument
-        you would do
+        you would do:
```
ecormany, line 1405: i think we can delete this whole example since we added a cross-reference

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1516,21 +1504,21 @@ def _has_tip(self) -> bool:
...
+        :raises: :py:class:`.types.PipetteNotAttachedError` if the pipette is
                  no longer attached (should not happen).
         """
         return self._core.get_hardware_state()
 
     @property  # type: ignore
     @requires_version(2, 0)
     def channels(self) -> int:
-        """The number of channels on the pipette."""
+        """The number of channels on the pipette. See :ref:`new-pipette`."""
```
ecormany:
```suggestion
        """The number of channels on the pipette.

        See :ref:`new-pipette`."""
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1584,27 +1572,26 @@ def __str__(self) -> str:
...
 
-        For more information on available modes for certain pipettes, see :ref:`pipette-volume-modes`.
+        The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a
+        low-volume mode to accurately dispense 1 µL of liquid. Low-volume mode can
+        only be set by calling ``configure_for_volume``. See :ref:`pipette-volume-modes`.
 
         .. note ::
 
-            Changing a pipette's mode will reset its :ref:`flow rates <new-plunger-flow-rates>`.
+            Changing a pipette's mode will reset its :ref:`flow rate <new-plunger-flow-rates>`.
```
ecormany:
```suggestion
            Changing a pipette's mode will reset its :ref:`flow rates <new-plunger-flow-rates>`.
```

ecormany (approved):
The substance is good; all that remains is pedantry! I've left many comments to make a record of why certain changes are necessary. I'll batch up the suggestions into a commit and then also do a final pass through to clean up line lengths. Then this will be ready to merge.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1458,7 +1441,7 @@ def trash_container(self, trash: labware.Labware) -> None:
     @requires_version(2, 0)
     def name(self) -> str:
         """
-        The name string for the pipette (e.g. 'p300_single')
+        The name string for the pipette (e.g., ``p300_single``).
```
ecormany:
```suggestion
        The name string for the pipette (e.g., ``"p300_single"``).
```
string value

api/src/opentrons/protocol_api/instrument_context.py
jwwojak: Restored orignal intro sentence.

> A context for a specific pipette or instrument.

PR #13781, opened 2023-10-12
feat(app): wire up chip statuses for modules & deck on odd

jerader:
closes RAUT-670

# Overview

wires up statuses for modules and fixtures in list view for the odd

# Test Plan

Send a protocol to the ODD and look at the modules & deck list view, see the various statuses for fixtures and modules.

# Changelog

- add logic to `FixtureTable` and `ProtocolSetupModules` and add test coverage

note: the `ProtocolSetupModules` component is very big and confusing? It should be refactored in a follow up at some point.

# Review requests

see test plan

# Risk assessment

low


app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
```diff
@@ -69,8 +79,45 @@ export function FixtureTable({
...
+          configurationStatus === CONFLICTING ||
+          configurationStatus === NOT_CONFIGURED
+        ) {
+          chipLabel = (
+            <>
+              <Chip
+                text={
+                  configurationStatus === CONFLICTING
+                    ? i18n.format(t('location_conflict'), 'capitalize')
+                    : i18n.format(configurationStatus, 'capitalize')
```
koji: Question
Is there any reason to use configurationStatus instead of i18n item like t('not_configured')?

app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
```diff
@@ -69,8 +79,45 @@ export function FixtureTable({
         <StyledText flex="2 0 0">{t('location')}</StyledText>
         <StyledText flex="3 0 0"> {t('status')}</StyledText>
       </Flex>
-      {/* ToDo (kk:10/10/2023) Add status will be implemented in a following PR */}
       {requiredFixtureDetails.map(fixture => {
+        const configurationStatus = configuration.find(
+          config => config.id === fixture.id
+        )?.configurationStatus
+
+        let chipLabel
```
koji: from the code, this type is `JSX.Element | string` isn't it?

app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
```diff
@@ -69,8 +79,45 @@ export function FixtureTable({
...
+                background={false}
+                iconName="connection-status"
+              />
+              <Icon name="more" size="3rem" />
+            </>
+          )
+        } else if (configurationStatus === CONFIGURED) {
+          chipLabel = (
+            <Chip
+              text={i18n.format(configurationStatus, 'capitalize')}
```
koji: same question

app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
```diff
@@ -55,6 +63,8 @@ export function FixtureTable({
         ]
       : []
 
+  const configuration = useLoadedFixturesConfigStatus(requiredFixtureDetails)
```
koji: The return type is `LoadedFixtureConfiguration[]` so probably configurations would be better.

```suggestion
  const configurations = useLoadedFixturesConfigStatus(requiredFixtureDetails)
```

app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
jerader: no reason, just that the status is already the string but you're right itd be better to use the string for translating in the future. I'll update!

app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
```diff
@@ -69,8 +79,45 @@ export function FixtureTable({
...
-      {/* ToDo (kk:10/10/2023) Add status will be implemented in a following PR */}
       {requiredFixtureDetails.map(fixture => {
+        const configurationStatus = configurations.find(
+          configuration => configuration.id === fixture.id
+        )?.configurationStatus
+
+        let chipLabel: JSX.Element
+        if (
+          configurationStatus === CONFLICTING ||
+          configurationStatus === NOT_CONFIGURED
```
brenthagen, line 89: we'll need to change the background color of the rendered item below to `COLORS.yellow3` under this condition also:

Screen Shot 2023-10-13 at 10 47 12 AM

app/src/organisms/ProtocolSetupModules/index.tsx
```diff
@@ -197,6 +216,7 @@ function RowModule({
   isLoading,
   prepCommandErrorMessage,
   setPrepCommandErrorMessage,
+  conflictedFixture,
```
brenthagen, line 219: same as in `FixtureTable`, need to change background color to `COLORS.yellow3` when a conflicted fixture exists

app/src/organisms/ProtocolSetupModules/FixtureTable.tsx
brenthagen: and a note based on the above screenshot, at some point we should add some extra `marginBottom` to the `ProtocolSetupModules` component so users can scroll past the map view button to see the last fixture when the list is longer like this

brenthagen (approved):
approach is good, just a couple background color comments -

and i definitely agree we should split up `ProtocolSetupModules` at some point. probably separate `ModulesTable` and `ModulesAndDeckMapModal` components would help readability a lot

PR #13782, opened 2023-10-12
fix(app): fix module calibration odd display

smb2268:
fix RQA-1789, RQA-1792

# Overview
There were a few problems with module calibration data being displayed on ODD:
1. Module calibration status was sometimes being shown as not ready when it was (1789)
2. Modules with ok calibrations still had "calibration required, attach pipette" text if instrument cal was incomplete

# Test Plan
1. On a Flex with some attached and not calibrated modules, setup a protocol that doesn't use all attached modules. Calibrate the required modules and see that the Modules item on the "Prepare to Run" screen is green.
2. Setup a protocol that use a module that is already calibrated. See that even before attaching and/or calibrating the instruments, the calibrated module list item is green and displays "Calibrated" text

# Changelog
1. Use `useModuleCalibrationStatus` hook to determine whether _needed_ modules are calibrated as opposed to checking all attached modules for calibration data. This logic now mirrors what is shown on desktop
3. Update `useModuleCalibrationStatus` to look at `last_modified` date instead of whole object
4. Update `RenderModuleStatus` to show "calibrated" text if module is attached and calibrated even if pipettes are not attached and/or calibrated

# Review requests
Sanity check the changes, I tested both changes locally and tested item 2 on a Flex but if there's a robot you can test item 1 on that would be great!

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/ProtocolSetupModules/index.tsx
+++ b/app/src/organisms/ProtocolSetupModules/index.tsx
@@ -132,7 +132,6 @@ function RenderModuleStatus({
 
   if (
     isModuleReady &&
-    calibrationStatus.complete &&
     module.attachedModuleMatch?.moduleOffset?.last_modified != null
   ) {
     moduleStatus = (
--- a/app/src/pages/OnDeviceDisplay/ProtocolSetup/__tests__/ProtocolSetup.test.tsx
+++ b/app/src/pages/OnDeviceDisplay/ProtocolSetup/__tests__/ProtocolSetup.test.tsx
@@ -24,6 +24,7 @@ import {
   useRunCreatedAtTimestamp,
   useModuleCalibrationStatus,
 } from '../../../../organisms/Devices/hooks'
+import { getLocalRobot } from '../../../../redux/discovery'
 import { useMostRecentCompletedAnalysis } from '../../../../organisms/LabwarePositionCheck/useMostRecentCompletedAnalysis'
 import { ProtocolSetupLiquids } from '../../../../organisms/ProtocolSetupLiquids'
 import { getProtocolModulesInfo } from '../../../../organisms/Devices/ProtocolRun/utils/getProtocolModulesInfo'
@@ -65,7 +66,6 @@ jest.mock(
 jest.mock(
   '../../../../organisms/Devices/ProtocolRun/utils/getProtocolModulesInfo'
 )
-jest.mock('../../../../redux/discovery')
 jest.mock('../../../../organisms/ProtocolSetupModules')
 jest.mock('../../../../organisms/ProtocolSetupModules/utils')
 jest.mock('../../../../organisms/OnDeviceDisplay/RunningProtocol')
@@ -73,6 +73,7 @@ jest.mock('../../../../organisms/RunTimeControl/hooks')
 jest.mock('../../../../organisms/ProtocolSetupLiquids')
 jest.mock('../../../../organisms/ModuleCard/hooks')
 jest.mock('../../../../redux/config')
+jest.mock('../../../../redux/discovery/selectors')
 jest.mock('../ConfirmAttachedModal')
 jest.mock('../../../../organisms/ToasterOven')
 
@@ -137,6 +138,9 @@ const mockUseDoorQuery = useDoorQuery as jest.MockedFunction<
 const mockUseToaster = useToaster as jest.MockedFunction<typeof useToaster>
 const mockUseModuleCalibrationStatus = useModuleCalibrationStatus as jest.MockedFunction<
   typeof useModuleCalibrationStatus
+>
+const mockGetLocalRobot = getLocalRobot as jest.MockedFunction<
+  typeof getLocalRobot
 >
 
 const render = (path = '/') => {
@@ -221,6 +225,7 @@ describe('ProtocolSetup', () => {
       <div>Mock ConfirmCancelRunModal</div>
     )
     mockUseModuleCalibrationStatus.mockReturnValue({ complete: true })
+    mockGetLocalRobot.mockReturnValue({ name: '123' } as any)
     when(mockUseRunControls)
       .calledWith(RUN_ID)
       .mockReturnValue({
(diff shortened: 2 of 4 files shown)


app/src/organisms/ProtocolSetupModules/index.tsx
```diff
@@ -132,7 +132,6 @@ function RenderModuleStatus({
 
   if (
     isModuleReady &&
-    calibrationStatus.complete &&
```
smb2268, line 135: If the module is attached and calibrated, we don't care if the instrument calibration status is complete.

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -399,9 +402,7 @@ function PrepareToRun({
     hasMissingModulesForOdd: isMissingModules,
     hasMissingCalForOdd: !areInstrumentsReady,
   })
-  const requiredCalibration = attachedModules.some(
```
smb2268, line 402: This is checking calibration of ALL attached modules, not just those needed for the protocol.

app/src/organisms/ProtocolSetupModules/index.tsx
koji: I think we need to check the calibration status since we don't allow users to calibrate a module without required instruments calibrations.

In my understanding
module status
- `connected`: instruments calibrations done + module calibration done
- `not calibrated (display calibrate button)`: instruments calibrations done + module is connected
- `require instruments calibration`: instruments are attached but not calibrated + module is connected  
- `require to attach instruments`: instruments are not attached but not calibrated + module is connected  
- `not connected`: module is not connected

smb2268: I don't think we need to in this scenario! This code is just used to show the status of an individual module during setup. Currently what a user sees when a module is attached and calibrated before proceeding to setup but not all instruments are attached and calibrated is this:
Screen Shot 2023-10-12 at 3 06 55 PM
This is misleading because we don't need the module to be recalibrated after the instruments are attached/calibrated. In this scenario, the module status item should be complete

smb2268: For every other item in this if/else chain we do care whether instruments are calibrated, but if no action is required for that module we shouldn't need to check.

koji: The screenshot you posted is from your branch or 7.0.1 branch?
For that case, the background color should be yellow and I guess the condition is used to change the background color correctly.

I could not find out the module calibration design link but the difference between the following link and module calibration is the text part. The color is the same.

https://www.figma.com/file/eMAHM8g07dtTKr2RsFjkzB/OCT-Release%3A-Touchscreen-%5BDeck-configuration-%26-Drop-tips%5D?type=design&node-id=85-121378&mode=design&t=7aqcL0fP3xvC0acX-0

smb2268: The screenshot is from 7.0.1 - the color is correct as the module is already calibrated. This change fixes the text to also reflect that

app/src/pages/OnDeviceDisplay/ProtocolSetup/__tests__/ProtocolSetup.test.tsx
```diff
@@ -64,6 +65,7 @@ jest.mock(
 jest.mock(
   '../../../../organisms/Devices/ProtocolRun/utils/getProtocolModulesInfo'
 )
+jest.mock('../../../../redux/discovery')
```
shlokamin: how come we need to mock the discovery module?

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -336,6 +337,8 @@ function PrepareToRun({
   const { t, i18n } = useTranslation(['protocol_setup', 'shared'])
   const history = useHistory()
   const { makeSnackbar } = useToaster()
+  const localRobot = useSelector(getLocalRobot)
+  const robotName = localRobot?.name != null ? localRobot.name : 'no name'
```
shlokamin, line 341: mmm it's dangerous to pretend we have a robot name here if we don't actually have one, we've gotten into trouble in the past by not short circuiting when something weird like this happens (ex passing in an empty protocol id string to a server endpoint when the frontend cant find a protocol id). i think we should either return null and console an error so its clear what went wrong, or let whatever util we're passing robot name into deal with the null case. 

practically this should never happen on an ODD, but we should still be careful to explicitly handle null checks

shlokamin (approved):
lgtm! just left some small nits

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
smb2268: yeah that makes sense! I pulled this pattern from elsewhere in ODD code but I'll update this case

app/src/pages/OnDeviceDisplay/ProtocolSetup/__tests__/ProtocolSetup.test.tsx
smb2268: now that we're using `getLocalRobot` in protocol setup, this test suite was failing since it tries to dig into that selector -- `TypeError: Cannot read properties of undefined (reading 'robotsByName')`

shlokamin: ah, if the subject of this test is protocol setup, and protocol set up is using  `getLocalRobot`,  i think we should mock `getLocalRobot` instead. its more clear whats going on and that way we're explicitly mocking modules the test subject is using, rather than downstream dependencies that the test subject shouldnt really have to worry about

shlokamin: ex. if another module the test subject imports ever depends on the discovery module down the road, debugging this would be sort of tricky

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
smb2268: It looks like theres 8 instances of this usage spread throughout the ODD - it would take me a bit to untangle for this case since there's a pretty long chain of hooks using the resulting string. Is it okay to make a follow up ticket for 7.1 that addresses all of those so we can get this in to cut the next alpha?

shlokamin: yeah def — we could also just have `ProtocolSetup` return null if we cant find a robot

PR #13783, opened 2023-10-13
fix(api): remove gratuitous tip state check prior to blow out

b-cooper:
# Overview

Do not assert, at the hardware control layer, that a tip must be present in the software tracker in order to move the plunger to the blowout position.

This is particularly important in recovery flows where the system may no longer have an accurate sense of what is attached to the pipette nozzles, and whether there is liquid inside.  In this case, there may actually be a tip present, but the system is not tracking it.  We should still allow the blowout action to occur, in the same way that we allow the drop tip action to occur.

# Review requests

- Issuing a `blowOutInPlace` command before a pick up tip should no longer fail.   

# Risk assessment
low


jbleon95 (approved):
All red diff is good diff

sfoster1 (approved):
Yeah, looks good to me!

PR #13785, opened 2023-10-13
feat(shared-data): Deck Definition V4 cutoutFixture updates

jbleon95:
# Overview

Based on a couple days of discussion between robot services and Apps/UI, we've decided to make a change in how we define cutoutFixtures. 

The waste chute, since it has four possible combinations between cover/no cover and staging slot/no staging slot, now have four definitions that each provide all usable addressableAreas. This flat structure has been chosen for simplicity and moving forward in implementing it the fastest. The names chosen for the cutoutFixture properties have been adjusted to allow an easier migration path when moving to a more robust schema in the future, if we decide to go down that path.

The `fixtureType` property was also removed for redundancy reasons.

# Test Plan

The veracity of changes should be covered by the previously added schema validation tests and python unit tests.

# Changelog

- added two more `cutoutFixtures` to represent the waste chute with a cover on, with and without the staging slot adapter
- renamed waste chute `cutoutFixtures` to semantically describe the hierarchal relationship between staging slot, waste chute, and cover in a flat structure
- renamed `cutoutLocationIds` to `mayMountTo`
- renamed `onFixtureAddressableAreas` to `providesAddressableAreas`
- renamed `1/8cWasteChute` and `96cWasteChute` addressable areas to `1and8ChannelWasteChute` and `96ChannelWasteChute`
- removed `fixtureType` property since it had identical data with the `id` property

# Review requests

We should all agree on the names being used here for cutout fixtures since they will be used going forward across multiple stacks and we'd like them to be backwards compatible with any future deck schema v5.

# Risk assessment

Low.

### DIFF
--- a/shared-data/deck/definitions/4/ot2_short_trash.json
+++ b/shared-data/deck/definitions/4/ot2_short_trash.json
@@ -367,7 +367,6 @@
         "10",
         "11"
       ],
-      "fixtureType": "singleStandardSlot",
       "displayName": "Standard Slot",
       "onFixtureAddressableAreas": {
         "1": ["1"],
@@ -386,7 +385,6 @@
     {
       "id": "fixedTrashSlot",
       "cutoutLocationIds": ["12"],
-      "fixtureType": "fixedTrashSlot",
       "displayName": "Fixed Trash",
       "onFixtureAddressableAreas": {
         "12": ["shortFixedTrash"]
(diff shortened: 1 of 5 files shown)


sfoster1 (approved):
Sweet!

SyntaxColoring (approved):
Yeah man.

CaseyBatten (approved):
Structure looks great, decision to refer to locations as "may mount to" makes it linguistically clear what will occur.

PR #13787, opened 2023-10-13
feat(server-utils): add logging configuration to server_utils to be used by servers.

vegano1:
# Overview

We want to split up the hardware_controller from the opentrons package, we are configuring the logs for each server independently. We would like to have that configuration in a common location, so let's throw it into server-utils and some other useful variables.

Closes: [RET-1381](https://opentrons.atlassian.net/browse/RET-1381)

# Test Plan

- [ ] Make sure buildroot builds succeed with this change
- [ ] Make sure oe-core builds succeed with this change
- [ ] Make sure the OT-2 and the Flex boot and all servers are started correctly
- [ ] Make sure we are correctly logging for each server

# Changelog

- Move common logging config code out of each server and keep it in server-utils
- Import server-utils to update/robot/system/ot3usb servers to configure logging
- Adds pydantic to update-server Pipfile to support ServerSettings

# Review requests
# Risk assessment

# TODO
- Fix all the tests

[RET-1381]: https://opentrons.atlassian.net/browse/RET-1381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server-utils/server_utils/__init__.py
```diff
@@ -3,3 +3,13 @@
 This package provides common utilities to be shared across any Opentrons servers
 intended to run on Opentrons robots.
 """
+
+from .constants import *
+from .logging import log_init
+
+print("Im importing server_utils")
```
fsinapi, line 10: debugging line?

sfoster1, line 10: debug print

fsinapi (commented):
My initial thought is that, if we're just going to have hard-coded configurations in here, there's little benefit over just keeping everything hard-coded in each package. I think we should either 

1. Set up a single, common config that applies to everything (including the full breadth of loggers/handlers)
2. Have a function that can be called with a set of configurable handlers & loggers to be supplied by the caller

...or option 3 is that this isn't a useful refactor because it's a minimal amount of duplicated code to set this up independently in each server. I'll leave that up to your judgement 😄

api/src/opentrons/__init__.py
```diff
@@ -140,6 +140,7 @@ async def initialize() -> ThreadManagedHardware:
     """
     Initialize the Opentrons hardware returning a hardware instance.
     """
+    # TODO (ba, 2023-10-12): Remove log init once the opentrons_hardware process is up and running
```
sfoster1, line 143: we could remove it now if we put the logging config for the hardware controller in server utils

system-server/system_server/__main__.py
```diff
@@ -4,13 +4,13 @@
...
-LOG = logging.getLogger(__name__)
+
+log = logging.getLogger(__name__)
 
 
 if __name__ == "__main__":
     args = build_root_parser().parse_args()
-    systemd.configure_logging(level=args.log_level.upper())
-    LOG.info(f"Starting system server on {args.host}:{args.port}")
+    log.info(f"Starting system server on {args.host}:{args.port}")
```
sfoster1, line 13: this now won't be logged, right? since the logging is set up when we invoke uvicorn now

update-server/Pipfile
```diff
@@ -6,6 +6,8 @@ name = "pypi"
 [packages]
 aiohttp = "==3.4.4"
 typing-extensions = "==3.10.0.0"
+pydantic = "==1.8.2"
```
sfoster1, line 9: is this necessary? on the flex, that's a full second copy of pydantic and all its transitive deps in `update-server`

sfoster1 (changes requested):
Inline comments, and I think I'm on Frank's side that this isn't quite a useful enough refactor yet. Specifically, it's weird to me that we have these fully qualified package names in server utils, you know? The configurations are so incredibly specific to these other servers that it's strange to have them in server-utils which is in theory a common-use library.

The thing is though that if we make this general enough that `server-utils` doesn't contain fully qualified package names (which I think is a good bar for "general enough"), then what _does_ it contain on the logging front? It's already pretty terse.

I like getting `config` out of the opentrons package but as discussed in person I think that's only useful if we can keep `opentrons` from having a dependency on `server_utils`, which I think needs a little more refactoring before it can happen.

vegano1:
Agreed,
The only required change to move the logging config will happen once the hardware controller is in its own package.
So for now I think ill skip this refactor.

PR #13790, opened 2023-10-16
Update 2.json

anuwrag:
"irregular" is incompatible with multi-channel.

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/shared-data/labware/schemas/2.json
+++ b/shared-data/labware/schemas/2.json
@@ -141,7 +141,7 @@
         "format": {
           "description": "Property to determine compatibility with multichannel pipette",
           "type": "string",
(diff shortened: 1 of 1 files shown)


sfoster1 (changes requested):
Ah, the idea here is that you write down the actual format in the labware def, and then code checks that to determine if it's compatible with a multi. So you would put "irregular" in there if the labware _is_ irregular. This PR will instead remove the possibility to put "irregular" in there, and we shouldn't do that.

anuwrag:
Thanks for rectifying @sfoster1! Sorry, this was incorrectly interpreted as a correct choice to make from the given list.

PR #13791, opened 2023-10-16
fix(app,api,robot-server): expose flex serial

sfoster1:
The flex doesn't store its serial number in /var/serial; instead, it stores it in an onboard eeprom. While we had the code to read it, we weren't using that code anywhere, and we should expose it to the app for display.

However, the update server doesn't (and shouldn't) know how to talk to the onboard eeprom. That's a problem, because it knows perfectly well how to read a file, and thus is the current source of truth for the serial.

So this PR
- adds the serial to the robot server health endpoint, on both ot-2 and flex
- changes the app to try and prefer the serial from the robot server health endpoint, falling back to the update server health endpoint if necessary

## risk
low, this is really only about displaying the serial number in the advanced settings page

## tests
- [x] push to a flex and use the dev app and see that you get a serial number
- [x] use the dev app and check that a flex still has an unknown serial (i.e. and not a whitescreen)
- [x] use the dev app and check that an OT-2 still displays its serial (and not unknown or null)
- [x] use the dev app and push to an OT-2 and see that it still displays the same serial (and not unknown or null)

### DIFF
--- a/app/src/redux/discovery/selectors.ts
+++ b/app/src/redux/discovery/selectors.ts
@@ -197,8 +197,10 @@ export const getDiscoverableRobotByName: (
   (robots, robotName) => robots.find(r => r.name === robotName) ?? null
 )
 
-export const getRobotSerialNumber = (robot: DiscoveredRobot): string | null =>
-  (robot.serverHealth && robot.serverHealth.serialNumber) ?? null
+export const getRobotSerialNumber = (robot: DiscoveredRobot): string | null => {
+  console.log(robot)
+  return robot.health?.robot_serial ?? robot.serverHealth?.serialNumber ?? null
+}
 
 export const getRobotApiVersion = (robot: DiscoveredRobot): string | null =>
   (robot.health && semver.valid(robot.health.api_version)) ??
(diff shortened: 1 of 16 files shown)


sanni-t (approved):
Backend changes LGTM! 
(looks like you need to run the formatter tho)

app/src/redux/discovery/selectors.ts
```diff
@@ -197,8 +197,10 @@ export const getDiscoverableRobotByName: (
   (robots, robotName) => robots.find(r => r.name === robotName) ?? null
 )
 
-export const getRobotSerialNumber = (robot: DiscoveredRobot): string | null =>
-  (robot.serverHealth && robot.serverHealth.serialNumber) ?? null
+export const getRobotSerialNumber = (robot: DiscoveredRobot): string | null => {
+  console.log(robot)
```
SyntaxColoring: Stray `console.log()` from debugging?

SyntaxColoring (approved):
This seems like the right way to do this, given our HTTP API architecture today. It's a bummer that the client has to care about this, though; the implementation details of how we divide our processes is leaking up.

app/src/redux/discovery/selectors.ts
sfoster1: yep, thank you!

ahiuchingau (approved):

mjhuff (approved):
JS lgtm!

PR #13792, opened 2023-10-16
fix(app): fix pipette wizard flows always showing "recalibrate" on results step

mjhuff:
Closes RQA-1800

# Overview

Persists hasCalData between renders so as not to return true during the attach/calibration wizard flow after the calibration step.

hasCalData is only used for displaying correct wizard text and will not impact any wizard functionality.

### Before
Screenshot 2023-10-16 at 1 29 03 PM

### After

Screenshot 2023-10-16 at 1 14 32 PM

# Test Plan
- Make sure your device has at least one mount empty, if not, detach a pipette to free the mount
- On the ODD, select instruments
- Select the empty mount, and select attach pipette
- Follow the wizard through the prompt for attach and calibrate
- Notice that on the results step (final step), the correct text is displayed.
- 

# Changelog
- Fix attach pipette calibration wizard text. 

# Risk assessment
low

### DIFF
--- a/app/src/organisms/PipetteWizardFlows/index.tsx
+++ b/app/src/organisms/PipetteWizardFlows/index.tsx
@@ -97,17 +97,15 @@ export const PipetteWizardFlows = (
   const [isFetchingPipettes, setIsFetchingPipettes] = React.useState<boolean>(
     false
   )
-  const hasCalData = React.useRef<boolean | null>(null)
-  if (hasCalData.current == null) {
-    hasCalData.current =
-      attachedPipettes[mount]?.data.calibratedOffset?.last_modified != null
-  }
   const memoizedAttachedPipettes = React.useMemo(() => attachedPipettes, [])
+  const hasCalData =
+    memoizedAttachedPipettes[mount]?.data.calibratedOffset?.last_modified !=
+    null
   const wizardTitle = usePipetteFlowWizardHeaderText({
     flowType,
     mount,
     selectedPipette,
-    hasCalData: hasCalData.current,
+    hasCalData,
     isGantryEmpty,
     attachedPipettes: memoizedAttachedPipettes,
     pipetteInfo: memoizedPipetteInfo,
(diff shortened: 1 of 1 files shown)


app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -97,14 +97,17 @@ export const PipetteWizardFlows = (
   const [isFetchingPipettes, setIsFetchingPipettes] = React.useState<boolean>(
     false
   )
-  const hasCalData =
-    attachedPipettes[mount]?.data.calibratedOffset?.last_modified != null
+  const hasCalData = React.useRef<boolean | null>(null)
+  if (hasCalData.current == null) {
+    hasCalData.current =
+      attachedPipettes[mount]?.data.calibratedOffset?.last_modified != null
+  }
```
smb2268: I think using `React.memo` would make more sense for this use case - you can pull the calibration data from  the `memoizedAttachedPipettes` variable below and I think that'll have the same effect

app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -97,14 +97,17 @@ export const PipetteWizardFlows = (
   const [isFetchingPipettes, setIsFetchingPipettes] = React.useState<boolean>(
     false
   )
-  const hasCalData =
-    attachedPipettes[mount]?.data.calibratedOffset?.last_modified != null
+  const hasCalData = React.useRef<boolean | null>(null)
```
smb2268:
```suggestion
  const hasCalData = memoizedAttachedPipettes[mount]?.data.calibratedOffset?.last_modified != null
```

app/src/organisms/PipetteWizardFlows/index.tsx
mjhuff: Ah yep, good catch 👍

smb2268 (approved):
lgtm! thanks for updating

PR #13793, opened 2023-10-16
chore: fix pypi deploys of api and shared-data

sfoster1:
These need to get uploaded to pypi and were not because of some kind of version conflict in api. With a bit of judicious pipenv relocking all is well.

In theory everybody would need to relock but really nobody's running twine so it's fine.

## Testing
- [x] do a `make deploy-pi` to test pypi and see that it runs. i did this already ( `opentrons`: https://test.pypi.org/project/opentrons/7.0.1a2/ , `opentrons_shared_data`: https://test.pypi.org/project/opentrons-shared-data/7.0.1a2/ ) which means that if you try to upload it will fail because pypi and test pypi don't allow duplicate files, but as long as it fails because of a 400 error you're good, the issue is twine wouldn't even run


SyntaxColoring (approved):
Okey dokey artichokey. 🕺

PR #13794, opened 2023-10-16
fix(api): Implicitly pre-home in opentrons.simulate.get_protocol_api()

SyntaxColoring:
# Overview

This fixes RQA-1801, an unreleased bug introduced by #13709.

Normally, the robot needs to be homed before you can command it to move, or it will raise an error. This is good.

However, there is a slight difference between `opentrons.execute.get_protocol_api()` and `opentrons.simulate.get_protocol_api()`: the `execute` variant requires the *caller* to do that home, whereas the `simulate` variant does it *for the caller.* In #13709, I accidentally dropped the implicit home from the `simulate` variant, causing it to start raising `MustHomeError`s.

# Test Plan

See RQA-1801.

# Changelog

* Add a test to catch the bug.
* Fix the bug by always doing `protocol_context.home()` in `opentrons.simulate.get_protocol_api()`, just before we return `protocol_context`.

  For PAPIv≤2.13, this is required to preserve historically expected behavior and fix RQA-1801. For PAPIv≥2.14, this is unnecessary, but harmless.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/simulate.py
+++ b/api/src/opentrons/simulate.py
@@ -316,6 +316,11 @@ def get_protocol_api(
             extra_labware=extra_labware,
         )
 
+    # Intentional difference from simulate.get_protocol_api():
+    # For the caller's convenience, we home the virtual hardware so they don't get MustHomeErrors.
+    # Since this hardware is virtual, there's no harm in commanding this "movement" implicitly.
+    context.home()
+
     return context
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Good call, thanks!

PR #13799, opened 2023-10-16
fix(app): do not check mag block in calibration status

shlokamin:
# Overview
There was a bug in the `useModuleCalibrationStatus` hook where we were looking for cal info for a mag block (which doesnt need to be calibrated).

closes RQA-1805

# Risk assessment

Low


mjhuff (approved):
Doesn't seem to affect anything else and solves the problem. LGTM!

PR #13803, opened 2023-10-17
fix(app): unload adapters after checking position of labware on adapter on heater shaker module

b-cooper:
# Overview

Properly combine dynamic logic for opening the heater shaker latch and offloading the labware and adapters in the correct order during LPC checks of labware on adapters on heater shaker modules

Closes RAUT-803

# Review requests

-  run LPC for a protocol that includes two unique labware loaded into the heatershaker on the same adapter

# Risk assessment
low

### DIFF
--- a/app/src/organisms/LabwarePositionCheck/CheckItem.tsx
+++ b/app/src/organisms/LabwarePositionCheck/CheckItem.tsx
@@ -323,7 +323,18 @@ export const CheckItem = (props: CheckItemProps): JSX.Element | null => {
         ]
 
   const handleConfirmPosition = (): void => {
-    let confirmPositionCommands: CreateCommand[] = [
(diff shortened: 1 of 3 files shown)


shlokamin (approved):
code changes lgtm, gonna test on a bot now

sanni-t:
Will this need to be updated for labware on adapters on the temperature module too?

shlokamin:
@sanni-t no we special case the HS, the adapter bit is a red herring

shlokamin:
and i just verified this works on a flex! lets merge

shlokamin:
@b-cooper i reverted your last commit that was meant to fix the mag block orange color since i think its not actually quite right (it broke a test and while trying to fix and untangle the logic i decided i didnt wanna deal with this right now)

we can fix the orange color as part as 7.1

PR #13804, opened 2023-10-17
feat(protocol-designer): error handling in create file wizard

jerader:
closes RAUT-801, RAUT-802, RAUT-786

# Overview

This PR adds a few functions to the create file wizard:
1. wires up error handling for modules/staging area slots/trash bin so things are disabled or go to the correct slot if the default slot is full
2. changes naming from `Staging Areas` -> `Staging Area Slots`
3. selects the trash bin selection by default for smoother user experience

bonus:
4. fixes a small bug in the deck setup (trash fixture rendering)

# Test Plan

Turn on the deck configuration FF in PD and create a flex protocol. Add all 4 staging area slots and try to add all the modules, see that if you add the outer slot modules (H-S, Temp, TC) and the trash bin, that the one you add last is disabled. This goes both ways for trash bin and H-S/Temp/TC. And also notice that the trash bin is selected by default.

NOTE: the disabled state for the module equipment cards is gross and designs should be updated probably, that'll be done later.

When you hit `Review file details` to create the file, obverse on the deck map that nothing is overlapping and the trash renders in the correct slot.

# Changelog

- rename staging area -> staging area slot in create file wizard and additional items section
- fix trash fixture rendering in deck setup
- create utils file in `CreateFileWizard` folder and create a few new utils for error handling (`getTrashSlot` and `getLastCheckedEquipment`)
- add test coverage
- wire up trash bin to be selected by default

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
@@ -17,7 +17,6 @@ import {
 import {
   HEATERSHAKER_MODULE_TYPE,
   MAGNETIC_MODULE_TYPE,
-  ModuleType,
   TEMPERATURE_MODULE_TYPE,
   getPipetteNameSpecs,
   PipetteName,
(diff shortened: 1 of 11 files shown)


protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -31,38 +30,36 @@ import {
...
-    formModule?.onDeck && formModule?.model != null
-      ? isModuleWithCollisionIssue(formModule.model)
-      : false
-
-  return crashableModuleOnDeck
+export const DEFAULT_SLOT_MAP: { [moduleModel in ModuleModel]?: string } = {
+  [THERMOCYCLER_MODULE_V2]: 'B1',
+  [HEATERSHAKER_MODULE_V1]: 'D1',
+  [MAGNETIC_BLOCK_V1]: 'D2',
+  [TEMPERATURE_MODULE_V2]: 'C1',
```
koji, line 55: I guess these settings make sense when all staging area slots are selected, but if a user doesn't select any staging area slots or just selects 2 slots, the user may think "Why I cannot add h/s temp and TC?".

Maybe telling these settings in the wizard makes sense? 🤔 


https://github.com/Opentrons/opentrons/assets/474225/43329958-da90-40fa-ab04-d3d08c745cfe

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -200,35 +197,14 @@ export function ModulesAndOtherTile(props: WizardTileProps): JSX.Element {
...
   const isFlex = values.fields.robotType === FLEX_ROBOT_TYPE
-  const allStagingAreasInUse =
-    values.additionalEquipment.filter(equipment =>
-      equipment.includes('stagingArea')
-    ).length === 4
-  const allModulesInSideSlotsOnDeck =
-    values.modulesByType.heaterShakerModuleType.onDeck &&
-    values.modulesByType.thermocyclerModuleType.onDeck &&
-    values.modulesByType.temperatureModuleType.onDeck
+  const trashDisabled = getTrashBinOptionDisabled(values)
```
koji: nit

```suggestion
  const trashBinDisabled = getTrashBinOptionDisabled(values)
```

protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
jerader: oh yes that makes sense! i should probably extend the logic to account for staging areas as well.

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+      : false
+
+  return crashableModuleOnDeck
+}
+
+export const getTrashBinOptionDisabled = (values: FormState): boolean => {
+  const allStagingAreasInUse =
+    values.additionalEquipment.filter(equipment =>
+      equipment.includes('stagingArea')
+    ).length === 4
```

```suggestion
    ).length === 4
```

maybe using const? 
```ts
const STAGINH_AREA_FULL = 4
```

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+  const moduleSlots: string[] = []
+  FLEX_SUPPORTED_MODULE_MODELS.map(model => {
+    const moduleType = getModuleType(model)
+    if (values.modulesByType[moduleType].onDeck) {
+      const slot = String(DEFAULT_SLOT_MAP[model])
+      return moduleType === THERMOCYCLER_MODULE_TYPE
+        ? moduleSlots.push('A1', slot)
+        : moduleSlots.push(slot)
+    }
+  })
```

```suggestion
const moduleSlots: string[] = FLEX_SUPPORTED_MODULE_MODELS.reduce((slots, model) => {
  const moduleType = getModuleType(model);
  if (values.modulesByType[moduleType].onDeck) {
    const slot = String(DEFAULT_SLOT_MAP[model]);
    return moduleType === THERMOCYCLER_MODULE_TYPE ? [...slots, 'A1', slot] : [...slots, slot];
  }
  return slots;
}, []);
```

koji:
Maybe adding padding to the div that wrapping deck map would be good? (especially right-side)

Screenshot 2023-10-18 at 4 21 54 PM

koji (approved):
Ran through the test plan and it worked as expected.
I think controlling checkbox disable is great but a few cases may confuse users 🤔.

jerader:
> Maybe adding padding to the div that wrapping deck map would be good? (especially right-side)
> 

@koji thanks for the call out! I'll address this when I fix up the deck map, it is still WIP since I am waiting for the v4 deck definition to be migrated into our `getDeckDefinition` utils.

PR #13805, opened 2023-10-17
fix(api, hardware): allow OT3Controller to disable tip motors

fsinapi:
# Overview

The assembly process for the 96ch requires the tip motors to be disabled, and it was noticed that the hardware controller will never actually disable them. While there's a workaround to use the `can_comm` script for now, it would be useful for the hardware controller to be able to do this with the functions we already have.

# Test Plan

Pushed to Flexy with a 96ch. Used ot3repl to call `disengage_axes` and `engage_axes` with `Axis.Q` specified. Watched `can_mon` and confirmed that the `GearEnableMotorRequest` and `GearDisableMotorRequest` messages got sent instead of the non-gear variants.

# Changelog

- Added `opentrons_hardware` helper functions for sending `GearEnableMotorRequest` and `GearDisableMotorRequest`
- `OT3Controller` will send gear-specific enable/disable messages if relevant
- Added tests

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -78,6 +78,8 @@ from opentrons.hardware_control.estop_state import EstopStateMachine
 from opentrons_hardware.hardware_control.motor_enable_disable import (
     set_enable_motor,
     set_disable_motor,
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Ah, nice, good work!

PR #13807, opened 2023-10-18
feat(protocol-designer): correct step count in create file wizard

jerader:
closes RAUT-775

# Overview

Now that a staging area slot tile was created for the Flex, the step count at the top of the create file wizard was incorrect for the Ot-2. This PR fixes that as well as fixing the go back button from the last page of the wizard.

NOTE: the total step counts stay as 7 for Flex and 6 for OT-2 regardless of if you end up skipping steps (if you don't add a 2nd pipette or if you add the 96-channel). I opted for this because a user will be aware of the total steps possible for the certain robot. Please let me know if you have any questions/concerns about this though!

# Test Plan

- turn on the deck modification & 96-channel feature flags. Create a Flex protocol and go thorugh the file wizard, make sure the step count work as expected for a total of 7 steps. See that if you add a 96-channel or don't add a 2nd pipette that the wizard skips to the appropriate page but the total step count remains at 7. Try to go back on the `Additional Items` (last page) and see that it goes to the appropriate page depending on if you have a 2nd pipette or not.

- now follow those steps for an Ot-2 protocol. See that the total number of steps is 6 and that the `Staging Area` tile is skipped. Try the variations of adding a 2nd pipette/not adding a 2nd pipette. Make sure the step counts work as expected.

- now turn off the deck modification feature flag and create a Flex protocol, see that it skips the `Staging Area` tile. 

# Changelog

- add `goback` logic for the `ModulesAndOtherTile`
- add logic for wizard step and total step count
- add test coverage
- fix the console warning for the `flexMap` component by adding keys to the children

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
@@ -173,10 +173,23 @@ export function ModulesAndOtherTile(props: WizardTileProps): JSX.Element {
               if (!enableDeckModification || robotType === OT2_ROBOT_TYPE) {
                 if (values.pipettesByMount.left.pipetteName === 'p1000_96') {
                   goBack(4)
-                } else if (values.pipettesByMount.right.pipetteName === '') {
+                } else if (
+                  values.pipettesByMount.right.pipetteName === '' &&
+                  robotType === FLEX_ROBOT_TYPE
+                ) {
                   goBack(3)
-                } else {
+                } else if (
+                  values.pipettesByMount.right.pipetteName === '' &&
+                  robotType === OT2_ROBOT_TYPE
+                ) {
                   goBack(2)
+                } else if (
+                  values.pipettesByMount.right.pipetteName !== '' &&
+                  robotType === FLEX_ROBOT_TYPE
+                ) {
+                  goBack(2)
+                } else {
+                  goBack()
                 }
               } else {
                 goBack()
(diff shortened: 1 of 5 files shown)


protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -173,10 +173,23 @@ export function ModulesAndOtherTile(props: WizardTileProps): JSX.Element {
...
+                  robotType === OT2_ROBOT_TYPE
+                ) {
+                  goBack(2)
+                } else if (
+                  values.pipettesByMount.right.pipetteName !== '' &&
+                  robotType === FLEX_ROBOT_TYPE
+                ) {
                   goBack(2)
+                } else {
+                  goBack()
```
jerader, line 192: I know this is very confusing, I am not sure how to clean it up. When the feature flag is removed the options will become clearer I think. So maybe we can keep as is for now and refactor when the ff is removed.

koji: I think the nest isn't deep (only 1 nest) so it would be okay.
When we can remove ff, we can remove this if statement, `if (!enableDeckModification || robotType === OT2_ROBOT_TYPE) {` right?

koji:
probably use `overflow-y: auto;` would solve this.

Screenshot 2023-10-18 at 5 39 47 PM

koji:
Question about the last case in the test plan
The wizard skips staging area as expected but step jumps up from 4/7 to 7/7.
Is that an expected behavior?

koji (approved):
The changes look good to me.
Tested 3 cases and they worked as expected.
I left one question about the step count.

jerader:
> Question about the last case in the test plan The wizard skips staging area as expected but step jumps up from 4/7 to 7/7. Is that an expected behavior?

@koji ya that happens if you don't select a 2nd pipette because you skip the 2nd pipette tip tiles and the staging area tile. So it jumps from step 4 to 7. This might be confusing without the feature flag but hoping 7.1 removes the ff 😅

PR #13811, opened 2023-10-18
refactor(api): Drop tip after finish behavior removal

CaseyBatten:
# Overview
 With the addition of the tip drop wizard, we no longer should be automatically dropping tips at the end of protocols. Instead, users should explicitly state in their protocols to drop tips. Alternatively, if there is still a tip attached at the end of a protocol finish then the user should be prompted on the ODD and/or app to manually drop the tip utilizing the drop tip wizard, and be given the choice to keep the tip.

# Test Plan
The test plan will be the same for Flex and OT2 at this current state. The test plan is as follows:
1. Load a protocol on a Robot that picks up a tip without a drop tip at the end.
2. Ensure the Robot keeps the tip after the protocol finishes.
3. Load a protocol on a Robot with a drop tip command at the end.
4. Ensure the Robot discards the tip at the end of the protocol before finishing.

# Risk assessment
There is a risk that users utilizing older protocols on OT2 will expect the tip to automatically drop at the end of their protocol. We can put a statement on the latest API version that states that protocols will no longer automatically drop tips upon completion, or possibly prompt users with a warning during analysis.


CaseyBatten:
Tested on a Flex with protocols that meet both test conditions, can confirm the Flex held onto the pipette and homed. In section scenario, Flex dropped tip as desired and homed.

sanni-t (commented):
The drop tips after run removal looks good 👍

But until the app is able to handle post-run homing as well, we should implement a thoughtful homing depending on whether there's a tip attached at the end of protocol or not. i.e, do not home pipette plungers if there's a tip.

CaseyBatten:
> The drop tips after run removal looks good 👍
> 
> But until the app is able to handle post-run homing as well, we should implement a thoughtful homing depending on whether there's a tip attached at the end of protocol or not. i.e, do not home pipette plungers if there's a tip.

I agree with that notion.

On another note, just tested on an OT-2, works as expected.

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
```diff
@@ -98,14 +98,9 @@ async def do_halt(self, disengage_before_stopping: bool = False) -> None:
...
 
     async def do_stop_and_recover(
-        self,
-        post_run_hardware_state: PostRunHardwareState,
-        drop_tips_after_run: bool = False,
+        self, post_run_hardware_state: PostRunHardwareState
     ) -> None:
-        """Stop and reset the HardwareAPI, optionally dropping tips and homing."""
-        if drop_tips_after_run:
-            await self._drop_tip()
```
SyntaxColoring, line 107: This removes the only use of the private `_drop_tip()` method, right? Let's delete that method, then. ✂️

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
```diff
@@ -98,14 +98,9 @@ async def do_halt(self, disengage_before_stopping: bool = False) -> None:
         )
 
     async def do_stop_and_recover(
-        self,
-        post_run_hardware_state: PostRunHardwareState,
-        drop_tips_after_run: bool = False,
```
SyntaxColoring, line 103: Agreed with @sanni-t that we need to be careful to avoid homing the plungers unless we know there's no liquid physically in the pipette tips.

SyntaxColoring (commented):
Looks good so far! Is there a Jira ticket that this references?

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
```diff
@@ -48,45 +45,6 @@ def __init__(
...
-                    await ot3api.home_z(mount=OT3Mount.GRIPPER)
-            except HardwareNotSupportedError:
-                pass
-            await self._movement_handler.home(
-                axes=[MotorAxis.X, MotorAxis.Y, MotorAxis.LEFT_Z, MotorAxis.RIGHT_Z]
-            )
-
-        for pipette_id, tip in attached_tips:
-            try:
-                await self._tip_handler.add_tip(pipette_id=pipette_id, tip=tip)
```
SyntaxColoring, line 73: `self._tip_handler` also becomes unused with this deletion, so it can be removed from this class.

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
```diff
@@ -98,16 +56,18 @@ async def do_halt(self, disengage_before_stopping: bool = False) -> None:
...
-            PostRunHardwareState.HOME_AND_STAY_ENGAGED,
-            PostRunHardwareState.HOME_THEN_DISENGAGE,
-        )
-        await self._hardware_api.stop(home_after=home_after_stop)
+        """Stop and reset the HardwareAPI, perform homing if the pipette is empty."""
+        attached_tips = self._state_store.pipettes.get_all_attached_tips()
+        if attached_tips:
+            await self._hardware_api.stop(home_after=False)
+            await self._movement_handler.home(
+                axes=[MotorAxis.X, MotorAxis.Y, MotorAxis.LEFT_Z, MotorAxis.RIGHT_Z]
```
SyntaxColoring, line 66: The old code mentioned `OT3Mount.GRIPPER`. It had a `# todo` comment related to `MotorAxis` that I wasn't really following. @sanni-t Do you know if we need to account for the gripper here?

api/tests/opentrons/protocol_engine/execution/test_hardware_stopper.py
```diff
@@ -225,28 +197,9 @@ async def test_hardware_stopping_sequence_with_gripper(
     decoy.when(state_store.config.use_virtual_gripper).then_return(False)
     decoy.when(ot3_hardware_api.has_gripper()).then_return(True)
     await subject.do_stop_and_recover(
-        drop_tips_after_run=True,
         post_run_hardware_state=PostRunHardwareState.HOME_AND_STAY_ENGAGED,
     )
 
     decoy.verify(
         await ot3_hardware_api.stop(home_after=False),
```
SyntaxColoring, line 204: It looks like this new test is missing a verification that we homed.

SyntaxColoring (commented):
A few remaining code comments before merging, but this mostly looks good. Thanks!

You mentioned that there's an open product question of whether we should actually do this, so we'll also wait for a decision on that before merging. I'll label this with `DO NOT MERGE` until we get a decision.

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
sanni-t: Yes, the gripper Z also needs to be retracted but maybe we can do it in a better way than what that code was doing. We can now check for the robot type via `config.robot_type` then if it's a flex, and not using virtual gripper, we can check for attached instruments and if there's a gripper in it then use `_movement.retract_axis`. 

Or if we anticipate this to be temporary code anyway, we can just copy paste the old code for gripper retraction.

PR #13808, opened 2023-10-18
feat(app): add Deck configuration screen to ProtocolSetup

koji:
# Overview
- Add Deck configuration screen to ProtocolSetup.
- Export CloseButton and PlayButton from ProtocolSetup component to reduce the lines.
- Rename `AddDeckConfigurationModal` to `AddFixtureModal`
- Add a new prop fixtureLocation to FixtureTable and ProtocolSetupModuleAndDeck to pass the slotName to AddFixtureModal.

For `Confirm` button, I will wire up in another PR (RAUT-804)

design
https://www.figma.com/file/eMAHM8g07dtTKr2RsFjkzB/OCT-Release%3A-Touchscreen-%5BDeck-configuration-%26-Drop-tips%5D?type=design&node-id=1-386&mode=design&t=rydOLM2zP7FPNv6j-0

close RAUT-672

# Test Plan

# Changelog
- Add ProtocolSetupDeckConfiguration
- update FixtureTable and ProtocolSetupModulesAndDeck to add props to pass slotName to AddFixtureModal
- export CloseButton and PlayButton (Currently Buttons.tsx under ProtocolSetup folder)
They can be a part of `atomes/buttons` in the future.

# Review requests

# Risk assessment
low

PR #13817, opened 2023-10-19
feat(app, react-api-client): wire up deck configuration editor and add useCreateDeckConfigurationMutation

koji:
# Overview
update deck configuration editor, add fixture modal, and deck config discard modal to align with expected behavior.

close RAUT-804

# Test Plan

# Changelog
- add useCreateDeckConfigurationMutation and test (test will be added later)
- update DeckConfiguration page component, DeckConfiguration for ProtocolSetup and AddFixtureModal to wire up Confirm button

# Review requests
- I added DeckConfigData to detect data changes for confirm button, but I feel that there might be more better way for doing that.
- I change DeckConfiguration page component to `DeckConfigurationEditor` to avoid a name conflict but I didn't change the folder name since Rob is working on re-organizing Figma and at some point we will need to align components' names with design. 
[related thread]
https://opentrons.slack.com/archives/CSCLVUW3C/p1698246368962059?thread_ts=1698246124.637669&cid=CSCLVUW3C 
- I created handleTapAdd in AddFixtureModal to avoid using if-statement since the behavior of Desktop and the behavior of ODD are slightly different. But at the same time maybe it would be good to keep using handleClickAdd for both cases?

# Risk assessment
low


koji:
- [x] Deck configuration page component
  - [x] update ODD part
  - [x] update Desktop part
  - [x] update tests
    - [x] AddFixtureModal
    - [x] DeckConfigurationDiscardChangesModal
    - [x] DeckConfiguration page component
- [x] Deck configuration in ProtocolSetup (need to merge https://github.com/Opentrons/opentrons/pull/13808 into `edge`)
  - [x] update ODD part 
- [x] useCreateDeckConfigurationMutation
  - [x] update test

app/src/organisms/DeviceDetailsDeckConfiguration/AddFixtureModal.tsx
```diff
@@ -73,6 +78,25 @@ export function AddDeckConfigurationModal({
...
+      })
+    }
+
+    setShowAddFixtureModal(false)
+  }
+
+  // For Desktop app
   const handleClickAdd = (fixtureLoadName: FixtureLoadName): void => {
     updateDeckConfiguration({
       fixtureLocation,
```
koji: Maybe use if statement with isOnDevice inside handleClickAdd instead of creating handleTapAdd?
Basically I do this to avoid if statement nest.

app/src/pages/DeckConfiguration/index.tsx
```diff
@@ -51,6 +57,11 @@ export function DeckConfiguration(): JSX.Element {
   const deckConfig = useDeckConfigurationQuery().data ?? []
   const { updateDeckConfiguration } = useUpdateDeckConfigurationMutation()
 
+  const [deckConfigData, setDeckConfigData] = React.useState<DeckConfigData>({
+    addedFixture: null,
+    currentDeckConfig: deckConfig,
+  })
```
brenthagen: this local state should just be a `DeckConfiguration` - we don't need to pass an added fixture into state.

then, probably the way we should handle setting state (in add fixture modal) is to use an [updater function](https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state) to set state based on previous state. something like (psuedo):
```
previousState => {
// filter previous state by fixtureLocation
// append new fixture
// return new state
}
```

another option is useReducer but we probably don't need to go there for this

brenthagen: and `handleClickRemove` needs to be updated to use the local state config in a similar way, appending a standard slot to the local state

app/src/pages/DeckConfiguration/index.tsx
```diff
@@ -118,7 +145,7 @@ export function DeckConfiguration(): JSX.Element {
           justifyContent={JUSTIFY_CENTER}
         >
           <DeckConfigurator
-            deckConfig={deckConfig}
+            deckConfig={deckConfigData.currentDeckConfig}
```
brenthagen: this data initializes to `[]` because the query is async - we need to:
1) add a `useEffect` to update local state when the deck config query resolves
2) probably block rendering the configurator (render null) when deck config length == 0 because current it renders as an empty deck map on load

brenthagen: Screen Shot 2023-10-26 at 10 45 30 AM

brenthagen (commented):
some initial comments on handling deck config local state

app/src/organisms/DeviceDetailsDeckConfiguration/AddFixtureModal.tsx
```diff
@@ -28,23 +28,30 @@ import { TertiaryButton } from '../../atoms/buttons'
...
+  Fixture,
+  FixtureLoadName,
+} from '@opentrons/shared-data'
 import type { ModalHeaderBaseProps } from '../../molecules/Modal/types'
 import type { LegacyModalProps } from '../../molecules/LegacyModal'
 
 interface AddFixtureModalProps {
   fixtureLocation: Cutout
   setShowAddFixtureModal: (showAddFixtureModal: boolean) => void
+  setCurrentDeckConfig?: React.Dispatch<React.SetStateAction<DeckConfiguration>>
```
koji, line 43: we need this change because the following gets typescript type error

```ts
setCurrentDeckConfig: (currentDeckConfig: DeckConfiguration) => void
```

app/src/pages/DeckConfiguration/index.tsx
```diff
@@ -49,28 +50,41 @@ export function DeckConfiguration(): JSX.Element {
...
+    )
+    createDeckConfiguration(currentDeckConfig)
   }
 
   const handleClickConfirm = (): void => {
-    // ToDo (kk:10/13/2023) add a function for the confirmation
+    if (!isEqual(deckConfig, currentDeckConfig)) {
+      createDeckConfiguration(currentDeckConfig)
+    }
   }
```
brenthagen, line 81: need a `history.goBack()` on confirm in all cases

app/src/organisms/ProtocolSetupDeckConfiguration/index.tsx
```diff
@@ -90,22 +90,31 @@ export function ProtocolSetupDeckConfiguration({
...
+          : fixture
+      )
+    )
+    createDeckConfiguration(currentDeckConfig)
   }
 
   const handleClickConfirm = (): void => {
-    // ToDo (kk:10/17/2023) add a function for the confirmation in a following PR for RAUT-804
+    createDeckConfiguration(currentDeckConfig)
   }
```
brenthagen, line 95: similar, need to `setSetupScreen('modules')` on confirm

app/src/organisms/ProtocolSetupDeckConfiguration/index.tsx
```diff
@@ -90,22 +90,31 @@ export function ProtocolSetupDeckConfiguration({
...
+  ] = React.useState<DeckConfiguration>(deckConfig)
+
+  const { createDeckConfiguration } = useCreateDeckConfigurationMutation()
 
   const handleClickAdd = (fixtureLocation: Cutout): void => {
     setTargetFixtureLocation(fixtureLocation)
     setShowConfigurationModal(true)
   }
 
   const handleClickRemove = (fixtureLocation: Cutout): void => {
```
brenthagen: we don't need `handleClickAdd` or `handleClickRemove` in this component because we want to replace the configurator with a static view of the deck map based on the protocol (same view as in https://opentrons.atlassian.net/browse/RAUT-671 actually). even if the deck map is replaced later, probably best to remove this functionality now to avoid confusion

brenthagen (approved):
just some routing comments, good to merge after those addressed ✔️

PR #13818, opened 2023-10-20
app(feat): launch drop tip wizard logic

mjhuff:
Closes RAUT-795

# Overview

This PR adds logic for wiring up the drop tip modal in additional locations. It also refactors the nice-modal-react context providers to the appropriate app/ODD locations to take advantage of top-level context. It includes some drive by refactoring as well - primarily adding missing tests.

A few notes of interest/controversy:
1) Because the OT-2 provides no way to validate that a tip has been removed, and we probably shouldn't always assume the Flex sensor is accurate for determining whether a tip is attached, the banner/ODD modal logic assumes that once a user begins the drop tip flow that the tip will be removed or the user is no longer interested with tip removal (and therefore the alert should be removed). 
2) If a user has tips on both the left and right pipettes and chooses not to skip the drop tip wizard (clicks dismiss on the dekstop banner or presses skip on the ODD), BUT the user exists a drop tip wizard flow prematurely, the banner/modal popup will still exist for the second pipette. 
3) The one optimization for getPipettesWithTipAttached is if both pipettes are used in a run, check for tip status for both pipettes on the same command loop iteration. 

### Protocol Header Entrypoint 
Screenshot 2023-10-20 at 11 18 29 AM

### ODD InstrumentDetail Entrypoint
Screenshot 2023-10-20 at 11 18 09 AM

### ODD RunSummary Entrypoint

Screenshot 2023-10-20 at 11 17 48 AM

# Test Plan

Protocols That May Be Useful:
[pick_up_tip.py.zip](https://github.com/Opentrons/opentrons/files/13055798/pick_up_tip.py.zip)
[Just Pipettes.py.zip](https://github.com/Opentrons/opentrons/files/13055797/Just.Pipettes.py.zip)

- Place tips on one/both of the pipettes and run the just pipettes protocol. After the run the banner should appear (Desktop) or clicking return to dashboard/run again will open the modal (ODD).
- Dismissing the banner will cause the drop tip modal not to render if it is not clicked in a few seconds. Clicking "skip" on the ODD will cause the drop tip banner on the desktop app to disappear. 
- Run the pick up tip protocol and press the estop some time AFTER the tip has attached. Verify the banner/ODD modal appears. 
- Run the pick up tip protocol and press the estop BEFORE the tip has attached (but before the pickUpTip command has been performed). Verify there is no banner/ODD modal.

# Changelog
- Added ODD overflow menu on the InstrumentDetail page.
- Wired up the drop tip wizard on the Desktop & ODD apps.

# Review requests
Test on a physical robot, please! AppUI is configured for these protocols.

# Risk assessment
low


app/src/assets/localization/en/drop_tip_wizard.json
koji, line 1: nit
recommend run sort on VSCode.

app/src/assets/localization/en/run_details.json
```diff
@@ -89,33 +89,34 @@
...
   "run_failed_modal_header": "{{errorName}}: {{errorCode}} at protocol step {{count}}",
   "run_failed_modal_title": "Run failed",
   "run_failed_splash": "Run failed",
-  "run_failed": "Run failed.",
   "run_has_diverged_from_predicted": "Run has diverged from predicted state. Cannot anticipate new steps.",
   "run_preview": "Run Preview",
   "run_protocol": "Run Protocol",
   "run_status": "Status: {{status}}",
   "run_time": "Run Time",
-  "run": "Run",
```
koji, line 111: Generally this change makes sense but currently we are using sort lines asc on VSCode and it sorts like below

run_failed
run

instead of 

run
run_failed

I think we can solve this after 7.1.0 release.

app/src/atoms/MenuList/index.tsx
```diff
@@ -21,7 +21,7 @@ export const MenuList = (props: MenuListProps): JSX.Element | null => {
   return isOnDevice && onClick != null ? (
     <LegacyModalShell
       borderRadius={BORDERS.borderRadiusSize4}
-      width="21.875rem"
+      width="max-content"
```
koji, line 24: thanks!

app/src/organisms/Devices/ProtocolRun/ProtocolDropTipBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <Banner
+      type="warning"
+      onCloseClick={onCloseClick}
+      marginBottom={SPACING.spacing16}
+      paddingRight={SPACING.spacing16}
+    >
+      <Flex flexDirection={DIRECTION_COLUMN}>
+        <StyledText as="p" fontWeight={FONT_WEIGHT_SEMIBOLD}>
```
koji: we can use `TYPOGRAPHY.fontWeightSemiBold` instead of `FONT_WEIGHT_SEMIBOLD`

```suggestion
        <StyledText as="p" fontWeight={TYPOGRAPHY.fontWeightSemiBold}>
```

app/src/organisms/Devices/ProtocolRun/ProtocolDropTipBanner.tsx
```diff
@@ -0,0 +1,58 @@
...
+          {t('tips_may_be_attached')}
+        </StyledText>
+
+        <Flex flexDirection={DIRECTION_ROW}>
+          <StyledText as="p" marginRight={SPACING.spacing4}>
+            {t('remove_the_tips_from_pipette')}
+          </StyledText>
+          <Btn
+            textAlign={ALIGN_START}
+            fontSize={TYPOGRAPHY.fontSizeP}
```
koji: Do we need this?
StyledText uses `as="p"`

app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+import { useCloseCurrentRun } from '../ProtocolUpload/hooks'
+import { SmallButton } from '../../atoms/buttons'
+import { StyledText } from '../../atoms/text'
+import { Modal } from '../../molecules/Modal'
+import { DropTipWizard } from '.'
+
+import type { ModalHeaderBaseProps } from '../../molecules/Modal/types'
+import type { PipetteData } from '@opentrons/api-client'
+import type { PipetteModelSpecs, RobotType } from '@opentrons/shared-data'
```
koji: nit

```suggestion
import type { PipetteData } from '@opentrons/api-client'
import type { PipetteModelSpecs, RobotType } from '@opentrons/shared-data'
import type { ModalHeaderBaseProps } from '../../molecules/Modal/types'
```

app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+interface TipsAttachedModalProps {
+  mount: PipetteData['mount']
+  instrumentModelSpecs: PipetteModelSpecs
+  robotType: RobotType
```
koji, line 21: Question
Do we need robotType?  From import statements, this modal is for the touchscreen app which means robotType is always "OT-3 Standard".

Will this component be used for OT-2?

app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+                buttonText={t('skip')}
+                onClick={() => {
+                  onCloseClick?.([])
+                  closeCurrentRun()
+                  modal.remove()
+                }}
+              />
+              <SmallButton
+                flex="1"
+                buttonType="primary"
```
koji: nit
buttonType can be removed since the default is primary.

```suggestion
                
```

app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+                }}
+              />
+            </Flex>
+          </Flex>
+        </Modal>
+        {showWizard ? (
+          <DropTipWizard
+            {...props}
+            closeFlow={() => {
+              onCloseClick?.((pipettesWithTip: any[]) => {
```
koji: Can we avoid using `any[]`?

app/src/organisms/DropTipWizard/__tests__/TipsAttachedModal.test.tsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react'
+import NiceModal from '@ebay/nice-modal-react'
+import { fireEvent } from '@testing-library/react'
+
+import { renderWithProviders } from '@opentrons/components'
+
+import { i18n } from '../../../i18n'
+import { handleTipsAttachedModal } from '../TipsAttachedModal'
+import { LEFT } from '@opentrons/shared-data'
```
koji, line 9: this line can be right after `renderWithProviders`

koji:
For a test of time, you may need to call a convert function in test.

app/src/assets/localization/en/drop_tip_wizard.json
mjhuff: Discussed w/ Nick. His PR (which will be merged in alongside mine) will fix this!

app/src/assets/localization/en/run_details.json
mjhuff: Totally agree - IDK why my VSCode sorts opposite yours, but we should enforce some sort of standard.

app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
mjhuff: Yep! We do need it for DropTipWizard. It's a bit hard to see, but we spread in the props on L94.

mjhuff:
> For a test of time, you may need to call a convert function in test.

Thanks for noticing. I think I'll remove the exact time testing, since the clock functionality is tested elsewhere and I'm just testing for the existing of elements on a page. It seems to be way more hassle doing this kind of string manipulation then it's worth.

mjhuff:
Note: The PD test failure is unrelated to this PR.

mjhuff:
@sanni-t Here are the python-sided changes we discussed in our sync on Monday.

koji:
> Note: The PD test failure is unrelated to this PR.

Shlok's PR to fix PD and LL is in edge already. I think once this PR takes that, the CI issue will be gone.

app/src/pages/OnDeviceDisplay/RunSummary.tsx
```diff
@@ -109,17 +129,37 @@ export function RunSummary(): JSX.Element {
...
 
   const handleReturnToDash = (): void => {
-    closeCurrentRun()
-    history.push('/')
+    const { mount, specs } = pipettesWithTip[0] || {}
+    if (isRunCurrent && pipettesWithTip.length !== 0 && specs != null) {
+      handleTipsAttachedModal(
+        mount,
+        specs,
+        ROBOT_MODEL_OT3,
```
b-cooper: This should really be `FLEX_ROBOT_TYPE` from shared data, that ROBOT_MODEL_OT3 happens to be the same string, but it's a deprecated const

b-cooper (approved):
Changes look good 🪶

robot-server/robot_server/maintenance_runs/router/base_router.py
```diff
@@ -153,10 +153,10 @@ async def create_run(
         run_data_manager: Current run data management.
         run_id: Generated ID to assign to the run.
         created_at: Timestamp to attach to created run.
-        protocol_run_has_been_played: Whether a protocol run has been played yet.
+        protocol_run_is_terminal: Whether a protocol run is in a terminal state.
         check_estop: Dependency to verify the estop is in a valid state.
     """
-    if protocol_run_has_been_played:
+    if not protocol_run_is_terminal:
```
sanni-t: This will change the behavior so that you can only create maintenance runs once a protocol run is completed. You want to also allow starting maintenance runs when a protocol run has been created but not played.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -642,6 +642,17 @@ def has_been_played(self) -> bool:
...
+        """Get whether engine is in a terminal state."""
+        if (
+            self._state.run_result == RunResult.SUCCEEDED
+            or self._state.run_result == RunResult.STOPPED
+            or self._state.run_result == RunResult.FAILED
+        ):
+            return True
+        else:
+            return False
+
```
sanni-t: You can check this same condition with just a `self._state.run_result is not None`

robot-server/robot_server/runs/dependencies.py
```diff
@@ -121,15 +121,15 @@ async def get_engine_store(
...
+async def get_protocol_run_is_terminal(
     engine_store: EngineStore = Depends(get_engine_store),
 ) -> bool:
     """Whether the current protocol run, if any, has been played."""
     try:
         protocol_run_state = engine_store.engine.state_view
     except AssertionError:
         return False
-    return protocol_run_state.commands.has_been_played()
+    return protocol_run_state.commands.get_is_terminal()
```
sanni-t: You can change this to something like:

```
async def get_is_okay_to_create_maintenance_run(
    engine_store: EngineStore = Depends(get_engine_store),
) -> bool:
    """Whether a maintenance run can be created if a protocol run already exists. """
    try:
        protocol_run_state = engine_store.engine.state_view
    except AssertionError:
        return False
    return not protocol_run_state.commands.has_been_played() or protocol_run_state.commands.get_is_terminal()
```

sanni-t (changes requested):
The robot server change in this PR will block you from creating maintenance runs before a protocol is started. That means no LPC or calibrations during protocol run setup. Added suggestions for fixing this.

mjhuff:
@sanni-t I retested with these changes and things seem to be working as expected. No issue with calibration/LPC/maintenance runs now 👍 

@b-cooper Can confirm the useEffect in ProtocolRunHeader isn't firing excessive network requests or re-renders 👍

sanni-t (approved):
🙌 🧙‍♂️ Thanks for making the changes!

PR #13820, opened 2023-10-20
fix(api): repair display names in tip rack fixtures for api tests

b-cooper:
# Overview

Fix up tip rack display name checks for failing api tests.

# Risk assessment
low


shlokamin (approved):
🧪

PR #13822, opened 2023-10-20
feat(hardware): read tip update from firmware whenever a notification is received

ahiuchingau:
# Overview
FLEX pipettes have tip sensing ability (thanks to tip presence sensors!); it would be nice if we could get real-time status of the attached tip from the pipette while the hardware controller is running.

This PR does it by creating a TipPresenceManager in the hardware controller that monitors tip updates from the pipettes. The tip presence manager is in charge of building a TipDetector dedicated to an attached pipette and pass any tip notification from the CAN bus to any of its listeners. 

The tip presence manager currently serves the two following purposes:
(1) prompting the firmware to confirm tip presence after pick up and drop tip actions
(2) caching the latest tip update from the detectors, and presenting that info to the `/instruments` endpoint

# Changelog
* the `has_tip` property of the OT3 pipette does not in fact reflect whether or not a tip is currently attached, but if a tip length has been added so we're changing the logic there
* removing the `tip_presence_detection_enabled` feature flag and make it the default behavior for all FLEX pipettes 

# What's next
This opens up an opportunity for the protocol engine to listen to the unexpected tip change during a protocol and issue the appropriate proper error/warning to the user.

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -1311,7 +1311,11 @@ class OT3Controller:
 
     async def update_tip_detector(self, mount: OT3Mount, sensor_count: int) -> None:
         """Build indiviudal tip detector for a mount."""
+        await self.teardown_tip_detector(mount)
         await self._tip_presence_manager.build_detector(mount, sensor_count)
+
+    async def teardown_tip_detector(self, mount: OT3Mount) -> None:
+        await self._tip_presence_manager.clear_detector(mount)
 
     async def get_tip_status(self, mount: OT3Mount) -> TipStateType:
         return await self.tip_presence_manager.get_tip_status(mount)
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -47,6 +47,7 @@ from .ot3utils import (
     moving_axes_in_move_group,
     gripper_jaw_state_from_fw,
 )
+from .tip_presence_manager import TipPresenceManager
 
 try:
     import aionotify  # type: ignore[import]
@@ -86,7 +87,6 @@ from opentrons_hardware.hardware_control.motor_position_status import (
     update_motor_position_estimation,
 )
 from opentrons_hardware.hardware_control.limit_switches import get_limit_switches
-from opentrons_hardware.hardware_control.tip_presence import get_tip_ejector_state
 from opentrons_hardware.hardware_control.current_settings import (
     set_run_current,
     set_hold_current,
@@ -127,7 +127,6 @@ from opentrons.hardware_control.types import (
     SubSystemState,
     SubSystem,
     TipStateType,
-    FailedTipStateCheck,
     EstopState,
     GripperJawState,
 )
@@ -172,7 +171,6 @@ from opentrons_shared_data.gripper.gripper_definition import GripForceProfile
 from opentrons_shared_data.errors.exceptions import (
     EStopActivatedError,
     EStopNotPresentError,
-    UnmatchedTipPresenceStates,
     PipetteOverpressureError,
     FirmwareUpdateRequiredError,
 )
@@ -316,6 +314,7 @@ class OT3Controller:
                 "or door, likely because not running on linux"
             )
         self._current_settings: Optional[OT3AxisMap[CurrentConfig]] = None
+        self._tip_presence_manager = TipPresenceManager(self._messenger)
 
     async def get_serial_number(self) -> Optional[str]:
         if not self.initialized:
@@ -867,34 +866,6 @@ class OT3Controller:
         res = await get_limit_switches(self._messenger, motor_nodes)
         return {node_to_axis(node): bool(val) for node, val in res.items()}
 
-    async def check_for_tip_presence(
-        self,
-        mount: OT3Mount,
-        tip_state: TipStateType,
-        expect_multiple_responses: bool = False,
-    ) -> None:
-        """Raise an error if the expected tip state does not match the current state."""
-        res = await self.get_tip_present_state(mount, expect_multiple_responses)
-        if res != tip_state.value:
-            raise FailedTipStateCheck(tip_state, res)
-
-    async def get_tip_present_state(
-        self,
-        mount: OT3Mount,
-        expect_multiple_responses: bool = False,
-    ) -> bool:
-        """Get the state of the tip ejector flag for a given mount."""
-        expected_responses = 2 if expect_multiple_responses else 1
-        node = sensor_node_for_mount(OT3Mount(mount.value))
-        assert node != NodeId.gripper
-        res = await get_tip_ejector_state(self._messenger, node, expected_responses)  # type: ignore[arg-type]
-        vals = list(res.values())
-        if not all([r == vals[0] for r in vals]):
-            states = {int(sensor): res[sensor] for sensor in res}
-            raise UnmatchedTipPresenceStates(states)
-        tip_present_state = bool(vals[0])
-        return tip_present_state
-
     @staticmethod
     def _tip_motor_nodes(axis_current_keys: KeysView[Axis]) -> List[NodeId]:
         return [axis_to_node(Axis.Q)] if Axis.Q in axis_current_keys else []
@@ -1333,3 +1304,17 @@ class OT3Controller:
     def estop_state_machine(self) -> EstopStateMachine:
         """Accessor for the API to get the state machine, if it exists."""
         return self._estop_state_machine
+
+    @property
+    def tip_presence_manager(self) -> TipPresenceManager:
+        return self._tip_presence_manager
+
+    async def update_tip_detector(self, mount: OT3Mount, sensor_count: int) -> None:
+        """Build indiviudal tip detector for a mount."""
+        await self._tip_presence_manager.build_detector(mount, sensor_count)
+
+    async def get_tip_status(self, mount: OT3Mount) -> TipStateType:
+        return await self.tip_presence_manager.get_tip_status(mount)
+
+    def current_tip_state(self, mount: OT3Mount) -> Optional[bool]:
+        return self.tip_presence_manager.current_tip_state(mount)
--- a/api/src/opentrons/hardware_control/backends/tip_presence_manager.py
+++ b/api/src/opentrons/hardware_control/backends/tip_presence_manager.py
@@ -1,0 +1,95 @@
+import logging
+from functools import partial
+from typing import Callable, Optional, List, Dict
+
+
+from opentrons.hardware_control.types import TipStateType, OT3Mount
+
+from opentrons_hardware.drivers.can_bus import CanMessenger
+from opentrons_hardware.firmware_bindings.constants import NodeId
+from opentrons_hardware.hardware_control.tip_presence import (
+    TipDetector,
+    types as tip_types,
+)
+from opentrons_shared_data.errors.exceptions import UnmatchedTipPresenceStates
+
+log = logging.getLogger(__name__)
+
+TipListener = Callable[[OT3Mount, bool], None]
+TipDetectorByMount = Dict[OT3Mount, Optional[TipDetector]]
+TipUpdateByMount = Dict[OT3Mount, Optional[bool]]
+
+
+class TipDetectorNotFoundError(Exception):
+    pass
+
+
+def _mount_to_node(mount: OT3Mount) -> NodeId:
+    return {
+        OT3Mount.LEFT: NodeId.pipette_left,
+        OT3Mount.RIGHT: NodeId.pipette_right,
+    }[mount]
+
+
+class TipPresenceManager:
+    """Handle tip change notification coming from CAN."""
+
+    _listeners: List[TipListener]
+    _detectors: TipDetectorByMount
+    _last_state: TipUpdateByMount
+
+    def __init__(self, can_messenger: CanMessenger) -> None:
+        self._messenger = can_messenger
+        self._listeners = []
+        self._detectors = {m: None for m in OT3Mount}
+        self._last_state = {m: None for m in OT3Mount}
+
+    async def build_detector(self, mount: OT3Mount, sensor_count: int) -> None:
+        # clear detector if pipette does not exist
+        if not sensor_count > 0:
+            self._detectors[mount] = None
+        else:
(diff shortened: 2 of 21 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1333,3 +1304,17 @@ async def build_estop_detector(self) -> bool:
     def estop_state_machine(self) -> EstopStateMachine:
         """Accessor for the API to get the state machine, if it exists."""
         return self._estop_state_machine
+
+    @property
+    def tip_presence_manager(self) -> TipPresenceManager:
+        return self._tip_presence_manager
+
+    async def update_tip_detector(self, mount: OT3Mount, sensor_count: int) -> None:
```
sfoster1, line 1312: probably want a way to remove one when a pipette is removed, right?

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -0,0 +1,95 @@
...
+from opentrons_shared_data.errors.exceptions import UnmatchedTipPresenceStates
+
+log = logging.getLogger(__name__)
+
+TipListener = Callable[[OT3Mount, bool], None]
+TipDetectorByMount = Dict[OT3Mount, Optional[TipDetector]]
+TipUpdateByMount = Dict[OT3Mount, Optional[bool]]
+
+
+class TipDetectorNotFoundError(Exception):
```
sfoster1: make it enumerated probably

sfoster1: some kidn of itnernal error code

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def __init__(self, can_messenger: CanMessenger) -> None:
+        self._messenger = can_messenger
+        self._listeners = []
+        self._detectors = {m: None for m in OT3Mount}
+        self._last_state = {m: None for m in OT3Mount}
+
+    async def build_detector(self, mount: OT3Mount, sensor_count: int) -> None:
+        # clear detector if pipette does not exist
+        if not sensor_count > 0:
```
sfoster1:
```suggestion
        if sensor_count <= 0:
```
nit, makes it easier to read for me

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -0,0 +1,95 @@
...
+from opentrons_hardware.hardware_control.tip_presence import (
+    TipDetector,
+    types as tip_types,
+)
+from opentrons_shared_data.errors.exceptions import UnmatchedTipPresenceStates
+
+log = logging.getLogger(__name__)
+
+TipListener = Callable[[OT3Mount, bool], None]
+TipDetectorByMount = Dict[OT3Mount, Optional[TipDetector]]
```
sfoster1: we've done this a couple other places and it's been sort of normalized, but for my money 
- if you have a dynamic store of data with a consistent key type and a consistent value type, _but no restriction on what value of key may be present_, that should be a `Dict`
- if you have a store of data with a consistent key type and a consistent value type, and you want to look up keys programmatically with strings, and some keys always need to be there, that should be a `TypedDict` with keys as literals
- If you have a store of data where you don't need to access things programatically, that's a dataclass

So for both `_detectors` and `_last_state`, where there has to be keys for all mounts - though really it's just the pipette mounts - at all times but the value can vary, to me that should be a `TypedDict` with key literals

api/src/opentrons/hardware_control/instruments/ot3/pipette.py
```diff
@@ -232,7 +231,6 @@ def reset_state(self) -> None:
         self._working_volume = float(self.liquid_class.max_volume)
         self._current_tip_length = 0.0
```
sfoster1, line 233: nit: if we're going to do "has_tip_length" let's make the internal variable an optional rather than checking for value == 0.0

api/src/opentrons/hardware_control/instruments/ot3/pipette_handler.py
```diff
@@ -273,6 +273,7 @@ def get_attached_instrument(self, mount: OT3Mount) -> PipetteDict:
             result[
                 "default_push_out_volume"
             ] = instr.active_tip_settings.default_push_out_volume
+            result["tip_presence_responses"] = instr.tip_presence_responses
```
sfoster1: don't love exposing this here - can we avoid it?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1913,41 +1928,54 @@ async def blow_out(
...
-        """
-        Check tip presence status. If a high throughput pipette is present,
-        move the tip motors down before checking the sensor status.
-        """
-        checked_mount = OT3Mount.from_mount(mount)
-        high_throughput = self._gantry_load == GantryLoad.HIGH_THROUGHPUT
-        if high_throughput:
-            # check if we're already at the tip_presence_check position, if we are dont move
+    @contextlib.asynccontextmanager
+    async def _high_throughput_tip_action(self) -> AsyncIterator[None]:
```
```suggestion
    async def _high_throughput_check_tipsself) -> AsyncIterator[None]:
```
maybe? `tip_action` is sort of nonspecific

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2040,12 +2068,7 @@ def add_tip_to_instr() -> None:
 
         # TODO: implement tip-detection sequence during pick-up-tip for 96ch,
         #       but not with DVT pipettes because those can only detect drops
-
-        if (
-            self.gantry_load != GantryLoad.HIGH_THROUGHPUT
-            and ff.tip_presence_detection_enabled()
-        ):
-            await self._backend.check_for_tip_presence(realmount, TipStateType.PRESENT)
+        await self.verify_tip_presence(realmount, TipStateType.PRESENT)
```
sfoster1: Honestly I think we should remove this call and the one in drop tip for low-throughput pipettes, and maybe for high-throughput pipettes, where we can run the commanded action regardless of whether things are present. Let the engine worry about it

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2104,12 +2127,8 @@ def _remove_tips() -> None:
...
         for shake in spec.shake_off_moves:
             await self.move_rel(mount, shake[0], speed=shake[1])
-        # TODO: implement tip-detection sequence during drop-tip for 96ch
-        if (
-            self.gantry_load != GantryLoad.HIGH_THROUGHPUT
-            and ff.tip_presence_detection_enabled()
-        ):
-            await self._backend.check_for_tip_presence(realmount, TipStateType.ABSENT)
+
+        await self.verify_tip_presence(realmount, TipStateType.ABSENT)
```
sfoster1: see the note above - I don't think we need to be doing this, since the engine does

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -0,0 +1,95 @@
...
+        detector = self.get_detector(mount)
+        return self._get_tip_presence(await detector.request_tip_status())
+
+    def get_detector(self, mount: OT3Mount) -> TipDetector:
+        detector = self._detectors[mount]
+        if not detector:
+            raise TipDetectorNotFoundError(f"Tip detector not set up for {mount} mount")
+        return detector
+
+    def add_listener(self, listener: TipListener) -> None:
```
sfoster1: do we call this anywhere?

sfoster1: also we typically implement these functions like `add_listener()` to return a function instance you can call that will remove the listener:
```
def add_listener(self, listener: TipListener) -> Callable[[],None]:
    self._listeners.append(listener)
    def remove():
        self._listeners.remove(listener)
    return remove
```

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class TipPresenceManager:
+    """Handle tip change notification coming from CAN."""
+
+    _listeners: List[TipListener]
+    _detectors: TipDetectorByMount
+    _last_state: TipUpdateByMount
+
+    def __init__(self, can_messenger: CanMessenger) -> None:
```
sfoster1: maybe take an initial set of listeners?

hardware/opentrons_hardware/hardware_control/tip_presence/detector.py
```diff
@@ -0,0 +1,153 @@
...
+        """Remove listener to tip change notification."""
+        if subscriber in self._subscribers:
+            self._subscribers.remove(subscriber)
+
+    def _filter(self, arb_id: ArbitrationId) -> bool:
+        return (arb_id.parts.message_id == PushTipPresenceNotification.message_id) and (
+            NodeId(arb_id.parts.originating_node_id) == self._node
+        )
+
+    def __del__(self) -> None:
```
sfoster1: I think we probably do not want this to be in a `__del__` magicmethod. The `__del__` magic method is called [only when an object's reference counts hit 0 or it's garbage collected](https://docs.python.org/3/reference/datamodel.html#object.__del__) making it really only useful for cleaning up resources the interpreter doesn't know about. 

Trying to use `__del__` to remove _other python objects_ will probably just result in the object never being cleaned up. The tasks in `self._tasks` hold references to the object because they're bound methods that hold references to `self`. Unless the objects in tasks are explicitly destroyed, those references will remain live.

So I'm pretty sure that `__del__` will never be called as-implemented, because even if this object is popped out of the `OT3Controller` and gets decref'd there, the tasks will still hold refs, and the object won't be gc'd.

I think the right thing to do is have a `def cleanup():` method that `OT3Controller` calls explicitly when we remove detectors, which cancels the tasks, which removes the references from the tasks, which will result (implicitly) in the object getting gc'd.

hardware/opentrons_hardware/hardware_control/tip_presence/detector.py
```diff
@@ -0,0 +1,153 @@
...
+        if isinstance(response, PushTipPresenceNotification):
+            tip_change = _parse_tip_status(response)
+
+        # pass new update to active waiters
+        for waiter in self._waiters:
+            waiter(tip_change)
+        # create a task that ends up broadcasting this update
+        self._create_reject_debounce_task(tip_change)
+
+    async def _broadcast(
```
sfoster1: I don't love `_broadcast` as a name because broadcast is also a thing you do to can messages, I had to read this a couple times before I realized it was "broadcast to subscribers above me". can we call this like `_update_subscribers()` or something

hardware/opentrons_hardware/hardware_control/tip_presence/detector.py
```diff
@@ -0,0 +1,153 @@
...
+        self._messenger.add_listener(self._receive_message, self._filter)
+        self._waiters: Set[Callable[[TipNotification], None]] = set()
+        self._tasks: Set[asyncio.Task[Any]] = set()
+
+    def add_subscriber(self, subscriber: TipChangeListener) -> None:
+        """Add listener to tip change notification."""
+        if subscriber not in self._subscribers:
+            self._subscribers.append(subscriber)
+
+    def remove_subscriber(self, subscriber: TipChangeListener) -> None:
```
sfoster1, line 61: can we implement this as a function object returned from `add_subscriber()`

hardware/opentrons_hardware/hardware_control/tip_presence/detector.py
```diff
@@ -0,0 +1,153 @@
...
+            waiter(tip_change)
+        # create a task that ends up broadcasting this update
+        self._create_reject_debounce_task(tip_change)
+
+    async def _broadcast(
+        self, fut: asyncio.Future[None], tip_change: TipNotification
+    ) -> None:
+        # wait for the future to time out before broadcasting the update
+        try:
+            asyncio.wait_for(fut, timeout=1.0)
```
sfoster1: the `1.0` probably wants to be an instance variable called like `debounce_time`

hardware/opentrons_hardware/hardware_control/tip_presence/detector.py
```diff
@@ -0,0 +1,153 @@
...
+            # this tip change is rejected, end the task here
```
sfoster1: it's less that we want to _reject_ the change, and more that we want the final result to be _the last thing we get_ during the debounce time. The debounce can never really end early.

sfoster1 (changes requested):
This is awesome. I think there's two bigger changes we should make, and then I have some inline comments.

First, I think the debounce logic is slightly wrong. The thing we want to do is ignore bouncing around within a timeout. When we see the first tip change happen, we start a timer; we keep collecting tip state changes; and the end result is the _last tip state value when the timer expires_, everything _before_ that is ignored. What we have here is that any tip state change is ignored if it changes after the first time.

Consider: the goal here is that if you take your finger and like flick the tip ejector shroud upwards and then it falls down, we don't tell the engine that a tip was removed and replaced within like 800 ms. We'd want to
- see the "tip removed" event and start the timer
- keep capturing events during the timer; the last one will be the "tip added" event as the shroud falls down, maybe bounces a couple times, and then settles
- and so the result, a second later, will be "no change" because the sensor went from present to present.

2. While we don't have to add the engine subscribing to the tip state in this PR, I think we _should_ add the framework to route events up out of the hardware controller - add it to the hardware events type, have `ot3api` add a listener to the tip state detector, pass it through ot3controller, etc

api/src/opentrons/hardware_control/backends/ot3controller.py
ahiuchingau: When it pipette is removed, we'll pass sensor_count as 0, which then removes the attached detector

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
ahiuchingau: Yea that makes sense

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
ahiuchingau: Not currently being called, but i imagine the protocol engine can make use of this to listen to the tip change event

sfoster1 (approved):
Looks really good to me!

PR #13823, opened 2023-10-20
fix(app): fix attachment flow firmware step filter

mjhuff:
Closes RQA-1827

# Overview

Fixes the logic for determining whether or not the firmware step is necessary during an attachment flow. Because it's not possible to check whether an instrument or module needs a firmware update until it is attached, do not filter out the step entirely. Instead, if the instrument is known to have the latest firmware, gracefully proceed to the next step. 

Before we know whether an update is required (or not), let's use a dummy loading spinner and then render the appropriate "updated" or "update in progress" text.

https://github.com/Opentrons/opentrons/assets/64858653/74dff5d2-1ecd-4632-b00d-bd771f576380

# Test Plan

- Attach a pipette

# Changelog
- Fix issue with firmware update screen not displaying.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/InstrumentsAndModules.tsx
+++ b/app/src/organisms/Devices/InstrumentsAndModules.tsx
@@ -154,6 +154,7 @@ export function InstrumentsAndModules({
             proceed={() => setSubsystemToUpdate(null)}
             description={t('updating_firmware')}
             proceedDescription={t('firmware_up_to_date')}
+            isODD={isFlex}
           />
         </ModalShell>
       )}
--- a/app/src/organisms/Devices/InstrumentsAndModules.tsx
+++ b/app/src/organisms/Devices/InstrumentsAndModules.tsx
@@ -154,7 +154,7 @@ export function InstrumentsAndModules({
             proceed={() => setSubsystemToUpdate(null)}
             description={t('updating_firmware')}
             proceedDescription={t('firmware_up_to_date')}
-            isODD={false}
+            isOnDevice={false}
           />
         </ModalShell>
       )}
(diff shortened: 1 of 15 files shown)


app/src/organisms/Devices/InstrumentsAndModules.tsx
```diff
@@ -153,6 +153,8 @@ export function InstrumentsAndModules({
             subsystem={subsystemToUpdate}
             proceed={() => setSubsystemToUpdate(null)}
             description={t('updating_firmware')}
+            proceedDescription={t('firmware_up_to_date')}
+            isODD={isFlex}
```
smb2268: This `InstrumentsAndModules` file is only rendered on desktop and the `FirmwareUpdateModal` is only rendered on the Flex - what is this `isODD` prop used for? Maybe we can make it optional and add a default value since `isODD={isFlex}` isn't accurate?

mjhuff: Do you mean FirmwareUpdateModal is only rendered on the ODD? 

This is used to adjust the size of the spinner. It should be checking if we're on the ODD/Desktop app. Yeah, isFlex is wrong here.

smb2268: This component `InstrumentsAndModules` is only rendered on desktop, the ODD equivalent is `InstrumentsDashboard`. Then within `InstrumentsAndModules` we only render the `FirmwareUpdateModal` if a Flex instrument needs to be updated. So for this case, I think you can say `isODD="false"`

koji: I'd recommend to use `isOnDevice` instead of `isODD`. We used to use both but now we use `isOnDevice` (maybe still isODD is exiting in our code base)

smb2268 (approved):
changes look good to me. thanks for fixing!

PR #13824, opened 2023-10-20
fix(app): fix pipette image size issue

koji:
# Overview
The issue was `height` and the image size scale didn't match the design (60x54). 
Modify transform to align the pipette image size with the design.
I think this modification isn't an ideal fix since we will need to Desktop app's since it's not fully responsive right now.
I guess after 7.1.0 we will be able to dig into this type of layout issue.

design
https://www.figma.com/file/0hYQ4lFJbAAI33jQMRt1Di/Release%3A-Desktop-June-Flex-launch?node-id=17265%3A286113&mode=dev

### before
10_20 instrument images

### after Flex
Screenshot 2023-10-20 at 5 05 11 PM

### after OT-2
Screenshot 2023-10-20 at 5 03 11 PM

close RQA-1825

# Test Plan

# Changelog
- modify PipetteCard component

# Review requests
This issue only happens to `edge` but I couldn't find out anything that causes this issue in [this PR](https://github.com/Opentrons/opentrons/pull/13813/files).
If there is a better way to fix this issue, please let me know.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/PipetteCard/index.tsx
+++ b/app/src/organisms/Devices/PipetteCard/index.tsx
@@ -236,11 +236,8 @@ export const PipetteCard = (props: PipetteCardProps): JSX.Element => {
                     pipetteSpecs={pipetteModelSpecs}
                     mount={mount}
                     //  pipette images for Flex are slightly smaller so need to be scaled accordingly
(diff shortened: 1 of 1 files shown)


mjhuff (approved):
Super strange. I couldn't find the immediate cause of this either. If this occurs again it might be worth digging more into it. Thanks for fixing this!

PR #13825, opened 2023-10-20
docs(api): additional low-volume pipetting information

ecormany:
# Overview

Adding some more information about low-volume mode:
- It's best practice to set the volume mode before picking up a tip.
- List concrete numbers for how `configure_for_volume` values affect pipette min and max volumes.

# Test Plan

- Checked local docs build and the [sandbox](http://sandbox.docs.opentrons.com/docs-low-volume-followup/v2/new_pipette.html#volume-modes).
- Confirmed bounds of `configure_for_volume()` values and min/max in simulation, e.g.:
``` python
    pipette50.configure_for_volume(4.999)
    pipette50.pick_up_tip()
    pipette50.aspirate(1, plate["A1"])
    protocol.comment(str([pipette50.min_volume, pipette50.max_volume]))  # [1.0, 30]
```

# Changelog

- Additions and line edits, all within the Volume Modes section of the Pipettes page.

# Review requests

Good and accurate advice for users?

# Risk assessment

nil, docs

### DIFF
--- a/api/docs/v2/new_pipette.rst
+++ b/api/docs/v2/new_pipette.rst
@@ -318,12 +318,13 @@ The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a l
     
 .. versionadded:: 2.15
 
-Setting the volume mode changes the minimum and maximum volume of Flex 50 µL pipettes as follows:
+Passing different values to ``configure_for_volume()`` changes the minimum and maximum volume of Flex 50 µL pipettes as follows:
 
 .. list-table::
     :header-rows: 1
+    :widths: 2 3 3
     
-    * - ``configure_for_volume()`` Value
+    * - Value
       - Minimum Volume (µL)
       - Maximum Volume (µL)
     * - 1–4.9
--- a/api/docs/v2/new_pipette.rst
+++ b/api/docs/v2/new_pipette.rst
@@ -321,7 +321,7 @@ The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a l
 .. note::
     The pipette must not contain liquid when you call ``configure_for_volume()``, or the API will raise an error.
     
-    Also, if the pipette is in a well location that may contain liquid, it will move upward to ensure it is not immersed in liquid before changing its mode.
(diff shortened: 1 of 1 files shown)


api/docs/v2/new_pipette.rst
```diff
@@ -310,34 +310,49 @@ The single- and multi-channel P50 GEN1 pipettes are the exceptions here. If your
 Volume Modes
 ============
 
-The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a low-volume mode to accurately dispense 1 µL of liquid. Set low-volume mode by calling :py:meth:`.InstrumentContext.configure_for_volume` with the amount of liquid you plan to aspirate, in µL::
+The Flex 1-Channel 50 µL and Flex 8-Channel 50 µL pipettes must operate in a low-volume mode to accurately dispense very small volumes of liquid. Set the volume mode by calling :py:meth:`.InstrumentContext.configure_for_volume` with the amount of liquid you plan to aspirate, in µL::
```
jwwojak, line 313: This is a nice change.

api/docs/v2/new_pipette.rst
```diff
@@ -310,34 +310,49 @@ The single- and multi-channel P50 GEN1 pipettes are the exceptions here. If your
...
 
     pipette50.configure_for_volume(1)
     pipette50.pick_up_tip()
     pipette50.aspirate(1, plate["A1"])
     
 .. versionadded:: 2.15
 
+Setting the volume mode changes the minimum and maximum volume of Flex 50 µL pipettes as follows:
```
jwwojak: One thought about this table: the monospace code font in column one makes it very wide. I get that including `configure_for_volume()` in the header tells the reader what the column contains, but I wonder if the method could go in the preceding sentence (to reduce column width)  and then just use a single word or 2-3 in the column header?

Here's an idea:

(Table intro sentence)
Assigning an **input** value to `configure_for_volume()`  changes the minimum and maximum volume of Flex 50 µL pipettes as follows:

1st column header:  "**Input**"

Just using bold here to highlight the connection between the text and col header.

❓

jwwojak (approved):
Added my .10 ¢ 

Could make a change to that table, but wholly optional if that makes things complex.

api/docs/v2/new_pipette.rst
ecormany: I agree with the structure. I think I have to avoid `assign` as the verb or I'll get put in Python jail. But I'll rework it so the header is **Value**.

sfoster1 (approved):
Yeah, this looks good to me!

PR #13826, opened 2023-10-23
fix(app): fix Flex protocol app analysis failing with ModuleNotFoundError

fsinapi:
# Overview

When #13770 changed the `opentrons_hardware` package to allow installation without the `python-can` module, I neglected to update the app-shell packaging script to include that optional dependency. As a result, app builds were not including all of the required packages and analyzing any OT Flex protocols will fail erroneously. 

# Test Plan

Once the app builds for this PR, try to analyze a Flex protocol.

# Changelog

- Specify the `flex-hardware` optional dependency when importing `api` during app packaging

# Review requests

Anyplace else that uses `setup.py` that I should update while I'm at it?

# Risk assessment

Low, fixes broken builds

### DIFF
--- a/app-shell/scripts/before-pack.js
+++ b/app-shell/scripts/before-pack.js
@@ -124,8 +124,7 @@ module.exports = function beforeBuild(context) {
         'install',
         `--target=${path.join(PYTHON_DESTINATION, sitePackages)}`,
         path.join(__dirname, '../../shared-data/python'),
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
oooohh yeah

koji (approved):
Nice! 
I just tested the branch with the json Rob tried to import and it was imported without any errors.

PR #13827, opened 2023-10-23
feat(api): Add prepare to aspirate to PAPI and engine

sfoster1:
Add the ability to explicitly call prepare_for_aspirate() to the python protocol API.

We make sure the pipette is prepared to aspirate before any aspiration, which is good, but the way we make this absolute is that if the pipette isn't ready for aspirate when you call aspirate(), we move it to the top of the current well (if necessary) to move the plunger up. This is a problem for users who are, for instance, building explicit prewetting behavior out of protocol API calls. It's common to move the pipette into the liquid and then delay to let the liquid settle before pipetting; if the pipette then moves up to prepare for aspirate before coming back down, that settling is undone. By adding the ability to explicitly prepare_for_aspirate(), the user can call it while they know the pipette is above the well.

This is added in a new 2.16 API level, which will ship in 7.1.

The command needs to be added to the API, but also to the engine (the `aspirate` command was, and is, the place where the fallback occurred rather than being a separately-issued command) and the sync client.

## Review requests

- This PR does the sort of minimum-viable change of adding the command, adding the API, and making the API call the command. It does _not_ touch any of the built in make-sure-we're-ready behaviors in either the PAPI implementation or the protocol engine. Should it? Should we make the engine emit a second command here? I don't think so - I think that's a better thing for a hypothetical resolver layer to do - but I'm interested to hear if you disagree
- Did I mess something up about the new api level? It's been a while

## Testing
- [x] Write a protocol that would incur the jump-out-of-the-well prepare behavior and
   - [x] on 2.16, add a prepare_for_aspirate() ahead of time, and make sure it prepares when the command says and doesn't come out of the well before aspirate
   - [x] on 2.16, _don't_ add a prepare_for_aspirate(), and make sure it still does the prep behavior


ecormany:
A terminology question while this is still in progress…
Currently the function is named `prepare_for_aspirate`, but much of our natural writing (and speaking) about it uses the phrase "prepare to aspirate". The _to_-construction means that we are treating _aspirate_ as a verb. The _for_-construction means that we are treating _aspirate_ as a noun. I think the noun form is more jargon-y either way you slice it:
- It could refer to the API method `aspirate()`.
- It could be a different noun _aspirate_. You won't find a dictionary entry that defines _aspirate_ (n.) as the act of sucking something in with air, even though we may use it that way inside Opentrons, or some scientists may use it that way. We have a natural tendency to use _aspiration_ as the noun for the action (you'll find it throughout this very PR).

tl;dr: Is it more natural to name this function `prepare_to_aspirate`?

andySigler (approved):

shared-data/command/schemas/7.json
SyntaxColoring, line 1: JSON protocols v7 are released into the wild now, right? I think these changes need to be in a new `8-draft.json` instead of `7.json`. We want `7.json` to be a snapshot of all the commands that are guaranteed to be usable on any robot that accepts a JSONv7 protocol.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1624,3 +1624,46 @@ def configure_for_volume(self, volume: float) -> None:
         if last_location and isinstance(last_location.labware, labware.Well):
             self.move_to(last_location.labware.top())
         self._core.configure_for_volume(volume)
+
+    @requires_version(2, 16)
+    def prepare_for_aspirate(self) -> None:
```
SyntaxColoring: Thank you for the thoughtful docstring here. 🙏

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1624,3 +1624,46 @@ def configure_for_volume(self, volume: float) -> None:
...
+        Use ``prepare_to_aspirate`` when you need to control exactly when the plunger
+        motion will happen. A common use case is a pre-wetting routine, which requires
+        preparing for aspiration, moving into a well, and then aspirating *without
+        leaving the well*::
+
+             pipette.move_to(well.top(z=-5))
+             pipette.delay(5)
+             pipette.mix(10, 10)
+             pipette.blow_out()
+             pipette.prepare_for_aspirate()
```
SyntaxColoring: Does this intentionally blow out and then `prepare_for_aspirate()` inside the well, while the tip is still submerged? I haven't tested this, but that's my reading based on the `blow_out()` reference:

>  location: The location to blow out into. If not specified, defaults to the current location of the pipette

And I'm assuming that the "current location" is submerged after the preceding `mix()`, but maybe that's wrong.

SyntaxColoring (commented):
Yay, love it!

> This PR does the sort of minimum-viable change of adding the command, adding the API, and making the API call the command. It does not touch any of the built in make-sure-we're-ready behaviors in either the PAPI implementation or the protocol engine. Should it? Should we make the engine emit a second command here? I don't think so - I think that's a better thing for a hypothetical resolver layer to do - but I'm interested to hear if you disagree

I think you're right not to touch the built-in make-sure-we're-ready behaviors in PAPI or Protocol Engine. That sounds like a breaking behavioral change in Protocol Engine, especially.

I think ideally, for composability and ease of reasoning, these always would have been two separate Protocol Engine commands, and the `aspirate` Protocol Engine command never would have prepared implicitly. But we're committed to that interface now. 🤷‍♂️

> Did I mess something up about the new api level? It's been a while

This looks right to me.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: ooh, i presumed that `blow_out()` would put the tip above the well. we should either test this on hardware or specify a location compatible with `prepare_for_aspirate()`, like `blow_out(well.top())`.

sfoster1: Yeah I think blow_out is implicitly in-place, I'll change the docstring

sfoster1:
behavior is correct after https://github.com/Opentrons/opentrons/pull/13827/commits/a653ad161f5ca15011f4831b298641207d6a43f3

shared-data/command/schemas/7.json
sfoster1: @SyntaxColoring This now has an `8.json`. Making it `8-draft.json` breaks automatic discovery of latest-version in tests, and I think we generally don't support `-draft` type things. We can add that support but I think this is better.

sfoster1:
Called `prepare_to_aspirate`/`PrepareToAspirate` as of https://github.com/Opentrons/opentrons/pull/13827/commits/1a28dc204268ba35707aa36b2fe7f971de1a5a8d

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1624,3 +1624,47 @@ def configure_for_volume(self, volume: float) -> None:
...
+
+             pipette.move_to(well.top(z=-5))
+             pipette.delay(5)
+             pipette.mix(10, 10)
+             pipette.move_to(well.top(z=5))
+             pipette.blow_out()
+             pipette.prepare_to_aspirate()
+             pipette.move_to(well.top(z=-5))
+             pipette.delay(5)
+             pipette.aspirate(10, well.top(z=-5))
```
SyntaxColoring: The negative sign in `well.top(z=-5)` vs `well.top(z=5)` seems easy to miss. Maybe let's use `well.bottom` for the submerged parts, to draw a clearer distinction.

```suggestion
             pipette.move_to(well.bottom(z=2))
             pipette.delay(5)
             pipette.mix(10, 10)
             pipette.move_to(well.top(z=5))
             pipette.blow_out()
             pipette.prepare_to_aspirate()
             pipette.move_to(well.bottom(z=2))
             pipette.delay(5)
             pipette.aspirate(10, well.bottom(z=2))
```

SyntaxColoring (approved):
Changes since last review look great, thanks. One optional docstring suggestion.

PR #13828, opened 2023-10-23
feat(api): Experimental, barebones support for dropping stuff in the waste chute

SyntaxColoring:
# Overview

This adds *very experimental, unstable, internal* methods to the Python Protocol API to let you eject tips and drop labware into the Opentrons Flex Waste Disposal Chute™.

This isn't specifically ticketed. It goes towards RSS-333 in general.

# Changelog

* Add `ProtocolContext.load_waste_chute()`, which returns a new `WasteChute` object.
* You can now pass that `WasteChute` object to `ProtocolContext.move_labware()` as a destination.
* You can now pass that `WasteChute` object to `InstrumentContext.drop_tip()` as a destination.

The implementation is hacky. For example, dropping a labware in a waste chute actually moves the labware "to slot D3" with a giant offset. We intend to change this when Protocol Engine learns about the waste chute. Hacky Python Protocol API support is beneficial for unblocking internal testers and sussing out API usability concerns.

# Known limitations

* **Motion planning is not smart enough to move the pipettes around the waste chute.** For example, it will happily jam your pipette into it if you move from slot D2 to slot C3. Plan your deck layouts carefully, or manually retract the pipettes before dangerous movements. And be ready on the e-stop.
* The API is likely to change in unpredictable ways as we figure out how to implement all of this.
* You can't dispense liquid to the waste chute. i.e. we support `drop_tip()` but not `dispense()` or `blow_out()`.
* You can't `move_to()` the waste chute.
* You can't apply offsets for dropping tips. (You *can* provide offsets for dropping labware.)
* Nothing will prevent you from doing impossible loads, like loading a labware in slot D3 and then using it as a waste chute, or loading a waste chute on an OT-2.

# Test Plan

[Here are a couple of test protocols](https://gist.github.com/SyntaxColoring/b8da3186d9b8d8f348967025a28fad49). We need to run both of them on a Flex. 

* [x] `drop_labware_in_waste_chute.py`
* [x] `drop_tips_in_waste_chute.py`
    * [x] With an 8-channel 1000µL pipette
    * [x] With a 96-channel 1000µL pipette

# Review requests

Anything too hacky to tolerate?

# Risk assessment


api/drop_labware_in_waste_chute.py
SyntaxColoring, line 1: To do before merge: Figure out how to automate these test cases. RSS discussed continuing our current pattern of testing them through robot-server, but that really seems to me like it's going to be so boilerplate-heavy that it will discourage thorough testing.

api/drop_tips_in_waste_chute.py
SyntaxColoring, line 1: Edit: See https://github.com/Opentrons/opentrons/pull/13828#discussion_r1370766599.

api/drop_labware_in_waste_chute.py
```diff
@@ -0,0 +1,17 @@
+from opentrons import protocol_api
```
sfoster1: I get what you're going for here i'm gonna get a twitch every time i see these in the root of the api subdir. can we put them somewhere, anywhere, in the test hierarchy

api/src/opentrons/protocol_api/_waste_chute_dimensions.py
```diff
@@ -0,0 +1,20 @@
+"""Constants for the dimensions of the Flex waste chute.
+
+https://docs.google.com/spreadsheets/d/1yGRN-5QOa45p-jLsi4tL_PbHmKSK-QsXqHws6oBK7do/edit#gid=258725851
```
sfoster1: let's not have this in source - it won't be accessible to anybody browsing the code. assert that it's correct to the hardware and that's good enough, we can keep it offline

sfoster1 (approved):
Looks really good to me but can we move the test protos so i don't twitch

api/drop_labware_in_waste_chute.py
SyntaxColoring: This PR was tested manually. Leaving the automated tests out, for expedience.

PR #13829, opened 2023-10-24
feat(app,components): render modules and labware in BaseDeck

brenthagen:
# Overview

refactors the `BaseDeck` component to render module and labware SVGs and info overlays. replaces two instance in the desktop app: the protocol run setup modules and labware tabs. also refactors how the render data is obtained, replacing hooks like `useModuleRenderInfoForProtocolById` that make unnecessary api calls with non-hook helper functions.

relies on a variety of stubbed data to be replaced when we can read from a protocol's deck configuration - for now, if the deck config feature flag is on these deck maps render a deck map with staging area slots.

Deck styling design updates are in progress, so details like colors are subject to change

re RAUT-793

Screen Shot 2023-10-24 at 11 19 41 AM

# Test Plan

 - manually verified accuracy of labware/modules on deck
 - updated unit tests

# Changelog

 - Renders modules and labware in BaseDeck
 - Uses instances of BaseDeck in protocol run setup modules and labware

# Review requests

review approach and props

# Risk assessment

i guess medium, changes to the deck map not behind feature flag


yarn.lock
```diff
@@ -1425,7 +1425,7 @@
   resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz#1d572bfbbe14b7704e0ba0f39b74815b84870d70"
   integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
 
-"@ebay/nice-modal-react@^1.2.13":
+"@ebay/nice-modal-react@1.2.13":
```
b-cooper, line 1428: this looks like it snuck in?

b-cooper: nevermind, this is intentional, talked it through in person

components/src/hardware-sim/Deck/FlexTrash.tsx
```diff
@@ -96,7 +96,7 @@ export const FlexTrash = ({
               transformOrigin="center"
               css={TYPOGRAPHY.bodyTextSemiBold}
             >
-              Trash
+              Trash bin
```
jerader, line 99: 🖌️

app/src/organisms/Devices/ProtocolRun/utils/getLabwareRenderInfo.ts
```diff
@@ -77,7 +78,9 @@ export const getLabwareRenderInfo = (
           )} but could not`
         )
       }
+      // TODO(bh, 2023-10-19): convert this to deck definition v4 addressableAreas
       const slotName = location.slotName.toString()
+      // TODO(bh, 2023-10-19): remove slotPosition when render info no longer relies on directly
```
jerader, line 83: 👍

app/src/organisms/Devices/ProtocolRun/SetupLabware/__tests__/SetupLabwareMap.test.tsx
```diff
@@ -277,16 +228,16 @@ describe('SetupLabwareMap', () => {
...
       .mockReturnValue(<div>mock module viz {mockTCModule.type} </div>)
 
     const { getByText } = render({
-      robotName: ROBOT_NAME,
       runId: RUN_ID,
-      commands: [],
+      protocolAnalysis: ({
+        commands: [],
+        labware: [],
+      } as unknown) as CompletedProtocolAnalysis,
```
jerader, line 240: curious why you've been typing these mocks as unknown instead of as any?

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
```diff
@@ -1,163 +1,163 @@
...
-                  }
-                )}
-
-                <SlotLabels robotType={robotType} />
-              </>
-            )}
-          </RobotWorkSpace>
+            robotType={robotType}
+            labwareLocations={labwareLocations}
+            moduleLocations={moduleLocations}
```
jerader, line 159: I wonder if `moduleLocations` and `labwareLocations` can be split up into their own utils or something since I think its similar logic between this component and setup module Map and setup liquid map. I think it would increase readability in these component as well

api-client/src/runs/types.ts
```diff
@@ -52,6 +52,7 @@ export interface VectorOffset {
   y: number
   z: number
 }
+
```
jerader: maybe you can remove this additional line since your PR doesn't change anything else in the file.

jerader (approved):
🤩 looking great! just left a few comments.

app/src/organisms/Devices/ProtocolRun/SetupLabware/__tests__/SetupLabwareMap.test.tsx
brenthagen: practically not much difference here but i guess i'm in the habit of using unknown as an intermediate:

[`Anything is assignable to unknown, but unknown isn’t assignable to anything but itself ...`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type)

asserting as unknown forces us to type assert to something after, which is safer than `any` which allows any assignment to or from.

here's the more thorough explanation of the assignability logic game: https://www.typescriptlang.org/docs/handbook/type-compatibility.html#subtype-vs-assignment

jerader: oh ya that makes a lot of sense. Thanks! I usually type mocks as any so I'll probably switch to this.

app/src/organisms/Devices/ProtocolRun/SetupLabware/SetupLabwareMap.tsx
brenthagen: there is some repeated logic here but also enough difference (and i think will be different again for liquids) that on balance i think it's worth keeping separate rather than sharing a util

PR #13830, opened 2023-10-24
feat(hardware_testing): start the plot process automatically 

ryanthecoder:
when starting a gravimetric/photometric test

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


andySigler (changes requested):
can we move this to be started not by the gravimetric script, but instead by the GravimetricRecorder? That's the thing that is doing the recording, and it used by other scripts not just the gravimetric QC script

PR #13833, opened 2023-10-24
refactor(api): remove feature-flag dependency from calibration_storage

fsinapi:
# Overview

As part of the migration of hardware-related code out of the `opentrons` package and into the `opentrons_hardware` package, we need to move dependencies on the advanced settings module out of any hardware code to avoid having multiple processes access & possibly modify the same settings file simultaneously.  

The `calibration_storage` module decides what implementation (OT2 vs OT3) to expose by checking the feature flag `enable_ot3_hardware_controller`. This poses an issue because the module is imported by a few files inside of `hardware_control`, so this PR takes the platform logic out of `calibration_storage` and offloads that logic to packages importing the module.

In most of the places that the `opentrons` package imports `calibration_storage`, the code is either for an OT2 or OT3 so it's easy to hard code the imports. In `robot-server`, we just utilize the feature-flag logic that had been done internally by `calibration_storage`.

# Test Plan

Test calibration flows on ot2/flex and make sure stuff is still working correctly!!! Pushing from this branch should still detect any calibration files already on a robot, and new files should be fully compatible with a downgrade because the code should be the same.

- [x] Test on OT-2
- [x] Test on Flex

# Changelog

# Review requests

Some tests ended up getting duplicated into `ot2` and `ot3` versions. I tried to get clever about `importlib` to parameterize based on the robot model but, in the end, the `ot2` and `ot3` subpackages are two different chunks of (very similar) code and I personally think it's okay to test them separately. Any objections?

A lot of robot-server tests were already only using the OT-2 code so I didn't bother trying to restructure them to test the OT-3 files too. Does someone think it's important to expand the tests as a part of this PR? 

# Risk assessment

High if it accidentally imports OT2 code on a Flex (or vice versa), but should be pretty easy to verify it doesn't.


api/src/opentrons/config/reset.py
```diff
@@ -5,14 +5,27 @@
...
+from opentrons.config import IS_ROBOT, feature_flags as ff
 from opentrons.calibration_storage import (
     delete_robot_deck_attitude,
-    clear_tip_length_calibration,
-    clear_pipette_offset_calibrations,
-    ot3_gripper_offset,
+    gripper_offset,
 )
 
+if ff.enable_ot3_hardware_controller():
```
sfoster1: can we also do this in the calling code? maybe even at the expense of demoting `reset` to have a version in `ot2` and a version in `ot3`?

api/tests/opentrons/calibration_storage/test_tip_length_ot3.py
```diff
@@ -0,0 +1,100 @@
+import pytest
```
sfoster1: this shouldn't exist, right? we don't do tip length calibration on ot3. I think "can we delete the tip length stuff from ot3" is a good bellwether of whether the refactor is complete

robot-server/robot_server/robot/calibration/check/user_flow.py
```diff
@@ -2,19 +2,37 @@
...
     get_custom_tiprack_definition_for_tlc,
-    load_tip_length_calibration,
-    save_tip_length_calibration,
-    create_tip_length_data,
-    get_pipette_offset,
-    save_pipette_calibration,
     mark_bad_calibration,
 )
+
+if is_ot3():
```
sfoster1: similarly, this stuff shouldn't run on the flex at all, honestly

robot-server/robot_server/robot/calibration/deck/user_flow.py
```diff
@@ -12,13 +12,28 @@
...
+from opentrons.config.feature_flags import enable_ot3_hardware_controller as is_ot3
 
 from opentrons.calibration_storage import (
     helpers,
     types as cal_types,
-    clear_pipette_offset_calibrations,
-    load_tip_length_calibration,
 )
+
+if is_ot3():
```
sfoster1: same

robot-server/robot_server/robot/calibration/pipette_offset/user_flow.py
```diff
@@ -12,11 +12,29 @@
...
     helpers,
-    get_pipette_offset,
-    save_pipette_calibration,
-    delete_pipette_offset_file,
-    load_tip_length_calibration,
 )
+
+from opentrons.config.feature_flags import enable_ot3_hardware_controller as is_ot3
+
+if is_ot3():
```
sfoster1: sam``

sfoster1 (changes requested):
This is good but I think we can take it slightly farther - once we can remove conditions from the ot2 calibration stuff (have it import from ot2 directly) and from reset I think we're done

fsinapi:
Tested on OT-2 & Flex. Tested that
* Can go through calibration flows via the app
* Old calibration data is still valid
* Calibration data created with this branch is compatible with a downgrade to `edge`

Flex is having some bugs with creating maintenance runs outside of protocol setup, but that also seems present on edge so I think it's unrelated to these changes.

sfoster1 (approved):
This looks awesome! Nice!

PR #13834, opened 2023-10-24
chore(monorepo): run jest on all frontend projects by default

shlokamin:
# Overview

So it turns out that jest's `watch` cli flag only works as expected (only rereuns tests that your code change affects using a dependency graph) if you don't provide specific test file paths. 

i.e. `yarn jest --watch` will only run tests that your src code change affects, while `yarn jest step-generation --watch` will rerun ALL step generation tests when a file change is made even if the file you modified is inside of the `app` directory.

To fix this, rather than providing paths to all JS projects by default, we run `yarn` on everything unless file paths are provided.

Good catch, @brenthagen!

# Risk assessment

Low

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ include ./scripts/python.mk
 
 API_DIR := api
 API_CLIENT_DIR := api-client
+APP_DIR := app
 APP_SHELL_DIR := app-shell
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
ah, nice!

brenthagen (approved):

PR #13837, opened 2023-10-24
feat(shared-data, api): Support for Fixture validations and low level infastructure

CaseyBatten:
# Overview
Creation of deck definitions for the new v4 Deck Fixtures schema. Designed to mimic behavior of labware definitions. Validations added to allow for checking of various requirements/conditions among the set of Fixtures to be expected.

# Risk assessment
Low level risk as this does not interface with existing software at the moment.


sfoster1 (approved):
Looks good! We might want tests in here but the validations are pretty straightforward and we can add tests as we need, could merge as is IMO

shared-data/python/opentrons_shared_data/deck/deck_definitions.py
```diff
@@ -0,0 +1,206 @@
...
+        ...,
+        description="Offset added to calculate drop coordinates of a labware placed on this deck.",
+    )
+
+
+class GripperOffsets(BaseModel):
+    default: Default
+
+
+class DeckDefinition(BaseModel):
```
SyntaxColoring: Let's call this `DeckDefinitionV4` to disambiguate from the other `TypedDict`s that we have floating around.

SyntaxColoring (approved):
Agreed with Seth—I think this is good to merge now if it helps unblock you, @jbleon95, or @sanni-t. There are some things like tests and organization that we can continue working on in follow-up PRs.

Thank you!

PR #13838, opened 2023-10-24
feat(api): Add protocol api support for partial tip configurations

Laura-Danielle:
# Overview

This pull request adds support for partial tip configuration in the python protocol api. It handles and closes three separate tickets (RSS-302, RSS-354, RSS-355). After this work is complete, a user should be able to configure their pipette in a partial tip mode. **The TODO list needs to be addressed before this PR can be tested, but the general structure is ready for review.**

# TODO
- [x]  Finish cleaning up/adding tests
- [x] Add better handling around column/row use-cases in choosing the ending pipette
- [x] Expose `NozzleMap` better in the hardware API

# Test Plan

It's another big 'un but we can strategize and choose to only test the MVP use-cases for single tip pick up and full column pick up. When this PR is finished, it should be able to

- [ ] Perform single tip pick up on 8 channel using 'A1'
- [ ] Perform single tip pick up on 8 channel using 'H1'
- [ ] Perform column pick up on 96 channel using 'A1'
- [ ] Perform column pick up on 96 channel using 'A12' (or 'H12'?)
- [ ] Perform single tip pick up on 96 channel using 'A1'
- [ ] Perform single tip pick up on 96 channel using 'H12'

All of the above configurations should also be run in a small protocol re-verify everything is still moving to the right place.
We should also test the 8 channel on the OT-2.

# Changelog
- Add a new protocol engine command to configure the pipette nozzle layout
- Add a new state field which saves the current nozzle map configuration by pipette id
- Add a new protocol api function which takes in dataclasses described in `_nozzle_map.py` to configure a nozzle state

# Review requests

Originally wanted to separate out the commits better, but it ended up being kind of a mess. Devs please review all. You can split it up into `protocol_engine` files and `protocol_api` files to more easily review. 

@ecormany you only need to review `instrument_context.py` and the docstrings there. I am definitely sure there are a bunch of edits we'll need to make. Ping me if you have questions.

# Risk assessment

Low. Adding a new feature (yay).


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1626,3 +1633,100 @@ def configure_for_volume(self, volume: float) -> None:
         if last_location and isinstance(last_location.labware, labware.Well):
             self.move_to(last_location.labware.top())
         self._core.configure_for_volume(volume)
+
+    @requires_version(2, 16)
+    def configure_nozzle_layout(
+        self, requested_nozzle_layout: Optional[NozzleLayoutBase]
```
sfoster1: I think that using a dataclass like this is kind of a lot for our intended users. I think using keywords is a lot better, something like
```
def configure_nozzle_layout(self, style='column'|'row'|'quadrant'|'full', starting_nozzle: str)
```

is a lot easier for a user to understand rather tha having to build an intermediate object. These aren't programmers.

We could make some of those magic strings imported constants like `OFF_DECK` if we wanted, I guess, but I think even that's pushing it.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1626,3 +1633,100 @@ def configure_for_volume(self, volume: float) -> None:
...
+            # Sets a pipette to single tip pick up mode using "A1" as the primary nozzle.
+            instr.configure_nozzle_layout(SingleNozzleLayout(primary_nozzle="A1")
+
+            # Sets a pipette to a full column pickup using "A1" as the primary nozzle.
+            instr.configure_nozzle_layout(ColumnNozzleLayout(primary_nozzle="A1"))
+
+            # Resets the pipette configuration to default
+            instr.configure_nozzle_layout()
+        """
+        if self._core.has_tip():
```
sfoster1: Organizationally I'm pretty sure this stuff should be in the engine if it's this important to check. I don't think we can enforce it with schemas or typechecking so the engine will have to make sure json-formatted protocols get it right also

sfoster1: and by "this stuff" i mean
- checking if we already have a tip, definitely something the engine has to do
- checking that we're only requesting columns on 8s and 96s
- checking that we're only requesting quadrants or rows on 96s

api/src/opentrons/protocol_engine/commands/configure_nozzle_layout.py
```diff
@@ -0,0 +1,101 @@
...
+    """Configure for volume command implementation."""
+
+    def __init__(self, equipment: EquipmentHandler, **kwargs: object) -> None:
+        self._equipment = equipment
+
+    async def execute(
+        self, params: ConfigureNozzleLayoutParams
+    ) -> Tuple[ConfigureNozzleLayoutResult, ConfigureNozzleLayoutPrivateResult]:
+        """Check that requested pipette can support the requested nozzle layout."""
+
```
sfoster1, line 79: like, here is where we should do the validation probably

api/src/opentrons/protocol_engine/commands/configure_nozzle_layout.py
```diff
@@ -0,0 +1,101 @@
...
+        nozzle_map = await self._equipment.configure_nozzle_layout(
+            pipette_id=params.pipetteId,
+            *params.configuration_params,
```
sfoster1: let's do kwarg expansion with `**` if possible IMO, varargs expansion with * can lead to surprises if things change

api/src/opentrons/protocol_engine/commands/configuring_common.py
```diff
@@ -1,8 +1,14 @@
 """Common configuration command base models."""
 
+from pydantic import BaseModel, Field
 from dataclasses import dataclass
+from opentrons.hardware_control.instruments.nozzle_manager import (
```
sfoster1: If `nozzle_manager` is going to be an element of hardware control's upward interface rather than just an implementation detail, it needs to live somewhere a lot less deep in the tree.

sfoster1 (changes requested):
I really think we should change the protocol API interface, but the big thing for me is that we gotta think about where those `NozzleMap` and `NozzleConfigurationType` types live if they're going to be part of the hardware controller's external interface. They shouldn't be defined so deep in the tree, they should be at the top level. That's true just for good practice reasons, but it's also true because we want to move a lot of hardware controller to a different package that the api can run without, so we'll need it to be in a typing stub kind of place, and we should put it there now. Even just `hardware_control.types` would be a good place.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: I agree, it's very easy to document "This parameter is a string. There are four good ones. If you pass a different string, the API raises an error."

Also, if we can be judicious in the use of the word `nozzle` it will make protocol code much more readable. I ceased to perceive it as an English word by the time I finished reading this line 😅 
```
instr.configure_nozzle_layout(SingleNozzleLayout(primary_nozzle="A1")
```

sanni-t: Ya I agree with @sfoster1 's suggestion. I do prefer imported constants like `OFF_DECK`, rather than magic strings as a developer, but I can understand the pains that could cause for a protocol writer so fully magical strings are fine too.

sanni-t: Also, imported constants allow us to selectively provide public interface to the configurations so that in 7.1.0, we can only provide interface for 'column'. But internally we can still use the api for row/ single tip etc by using private accessors.

api/src/opentrons/protocol_engine/commands/configure_nozzle_layout.py
```diff
@@ -0,0 +1,101 @@
...
+        )
+
+        return ConfigureNozzleLayoutResult(), ConfigureNozzleLayoutPrivateResult(
+            pipette_id=params.pipetteId,
```
sanni-t, line 86: I don't think pipette_id is needed in the result

Laura-Danielle: We need it to save the layout so that it is keyed by pipette id and so that we can look it up later.

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -143,6 +152,10 @@ def _handle_command(  # noqa: C901
                 nozzle_offset_z=config.nozzle_offset_z,
             )
             self._state.flow_rates_by_id[private_result.pipette_id] = config.flow_rates
+        elif isinstance(private_result, PipetteNozzleLayoutResultMixin):
+            self._state.nozzle_configuration_by_id[
+                private_result.pipette_id
```
sanni-t, line 157: You can use `command.params.pipetteId` to get the pipette ID for this command. So you won't need to include `pipettte_id` in the results

CaseyBatten (approved):
Ran this on RSS flex with most recent changes and was able to successfully do partial tip pickup, perform aspirations and dispenses, drop tips, reconfigure for whole 96 channel pickup and repeat.
**However important note:**
Under the shared-data pipette json definitions for the 96 channel, the "presses" parameter under the "pickUpTipConfigurations" has to be changed from a 0.0 to a 1.0 in order to successfully pick up the tips, otherwise the 96-channel would not make contact with the tips.

sfoster1 (approved):
Sure, sounds good to merge I guess. We really need to fix the gcode tests.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1626,3 +1627,58 @@ def configure_for_volume(self, volume: float) -> None:
...
+    def configure_nozzle_layout(
+        self,
+        style: NozzleLayout,
+        start: Optional[str] = None,
+        front_right: Optional[str] = None,
+    ) -> None:
+        """Configure a pipette to pick up less than the maximum tip capacity. The pipette
+        will remain in its partial state until this function is called again without any inputs. All subsequent
+        pipetting calls will execute with the new nozzle layout meaning that the pipette will perform
+        robot moves in the set nozzle layout.
```
jwwojak, line 1641: Here's a suggestion for the start of the description:

> Configure a pipette to pick up less than the tip's maximum rated capacity. The pipette will continue to aspirate in this reduced capacity until ``configure_nozzle_layout`` is called again with no arguments.

I don't know enough about what this method does to understand part of the last sentence in _italics:_

> All subsequent pipetting calls will execute with the new nozzle layout _meaning that the pipette will perform robot moves in the set nozzle layout_.

Just a reminder, even though this is a python file with a .py file extension, our docs and docstrings are rendered by Sphinx. This means to get `code` type font, you need to put two single quotes `` around the code snippet. For example:

> Certain pipettes are restricted to a subset of ``` ``NozzleLayout`` ``` types.

jwwojak (commented):
@Laura-Danielle   Added a small suggestion for your consideration. Don't let me be a blocker to moving forward. We can always edit docstrings at a later date.

PR #13843, opened 2023-10-25
fix(shared-data): reduce 96chan pickup distance

sfoster1:
Keeps the lifetime nice and long.

Numbers from hardware.


Laura-Danielle (approved):
We don't need to update the drop tip value as well?

sfoster1:
> We don't need to update the drop tip value as well?

Nope, different stresses I guess

ahiuchingau (approved):
⚙️

caila-marashaj (approved):
long live the gears

PR #13844, opened 2023-10-25
fix(api): don't move in x direction during tip pickup

caila-marashaj:
Due to a testing oversight, when `_motor_pick_up_tip` is called, if there is an ending retract distance specified, the gantry will move the retract distance in the x direction instead of the z


sfoster1 (approved):
Nice! add a test please

PR #13845, opened 2023-10-25
feat(step-generation): wire up missingAdapter() error creator

jerader:
closes RAUT-798

# Overview

This Pr wires up an error creator in `_pickUpTip` so that if you try to pick up tips from the full 96-channel pipette and an adapter is not underneath, it errors.

# Test Plan

Create a Flex protocol with the 96-channel. Add a labware and then one transfer or mix step. See that NO timeline error occurs. Now, go to the starting deck state and move the tiprack to be on top of the deck and not on top of the adapter.See that the timeline error occurs now. Should match the screenshot below.

Finally, go back and create a new protocol with no 96-channel. See that the timeline error does not occur during transfer/mix steps.

Screen Shot 2023-10-25 at 1 14 32 PM

# Changelog

add `missingAdapter()` error creator to `_pickUpTip` and move the i18n text to the correct location for timeline errors instead of warnings

# Review requests

For partial tip, we probably need to extend the args to have knowledge of if the full 96 tips are being used, but that shouldn't be too bad since `transfer`/`mix`/`consolidate`/etc will have that arg extended as well. So we can just pass the arg down from those components to `replaceTip` and then to `pickUpTip`.

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/localization/en/alert.json
+++ b/protocol-designer/src/localization/en/alert.json
@@ -157,8 +157,8 @@
         "body": "The Waste Chute or Trash Bin to drop tip in does not exist."
       },
       "MISSING_96_CHANNEL_TIPRACK_ADAPTER": {
-        "title": "Missing tiprack adapter for 96-channel",
-        "body": "A 96-channel cannot pick up a full tiprack without an adapter underneath."
+        "title": "Missing 96-channel tip rack adapter",
+        "body": "The tip rack must be placed in an adapter when picking up 96 tips simultaneously."
       }
     },
     "warning": {
--- a/protocol-designer/src/localization/en/alert.json
+++ b/protocol-designer/src/localization/en/alert.json
@@ -155,6 +155,10 @@
       "DROP_TIP_LOCATION_DOES_NOT_EXIST": {
         "title": "Attempting to drop tip in an unknown location",
         "body": "The Waste Chute or Trash Bin to drop tip in does not exist."
+      },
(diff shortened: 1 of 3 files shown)


protocol-designer/src/localization/en/alert.json
```diff
@@ -155,6 +155,10 @@
       "DROP_TIP_LOCATION_DOES_NOT_EXIST": {
         "title": "Attempting to drop tip in an unknown location",
         "body": "The Waste Chute or Trash Bin to drop tip in does not exist."
+      },
+      "MISSING_96_CHANNEL_TIPRACK_ADAPTER": {
+        "title": "Missing tiprack adapter for 96-channel",
+        "body": "A 96-channel cannot pick up a full tiprack without an adapter underneath."
```
jerader: @ecormany, thoughts on what this error text should say? It is triggered when you try to pipette from a tiprack with all 96 channels from a 96-channel pipette but the tiprack is missing the adapter underneath.

ecormany: This mirrors what we've written in documentation:
```suggestion
      "MISSING_96_CHANNEL_TIPRACK_ADAPTER": {
        "title": "Missing 96-channel tip rack adapter",
        "body": "The tip rack must be placed in an adapter when picking up 96 tips simultaneously."
```

jerader: thank you!

koji (approved):
lgtm

ecormany (approved):
Error text is good! 👍

PR #13847, opened 2023-10-25
fix(api): allow module poller to persist if it tries to complete a cancelled future

fsinapi:
# Overview

There have been a few cases where the module poller for a Heater Shaker or Thermocycler stops polling until the module is power cycled or the server is restarted. The poller is _supposed_ to ignore any exceptions (and just mark them on the waiter), so it seems weird that this happens.

In a recent ABR failure, we noticed that the poller stopped running at about the same time when a run was cancelled. The poller should never be getting cancelled by the protocol engine, but what _does_ get cancelled is the protocol engine command implementation that's running `wait_next_poll`. When that task gets cancelled, so does the associated future used to wait for a poll. Completing a cancelled future results in an exception that we were not catching.

Addresses RQA-1780

# Test Plan

Need to test on a robot still, I'm gonna run a protocol that just opens & closes the heater-shaker latch a bunch of times and cancel the protocol and make sure I can get the new log entry to show up.

The new unit test makes me pretty confident that this was an issue, it will fail if you run without the changes to the poller.

# Changelog

Added try/except to module poller Future completion to account for cancelled waiter tasks.

# Review requests

# Risk assessment

Not particularly high, this is just affecting error handling.

### DIFF
--- a/api/src/opentrons/hardware_control/poller.py
+++ b/api/src/opentrons/hardware_control/poller.py
@@ -80,7 +80,7 @@ class Poller:
         try:
             waiter.set_result(None) if e is None else waiter.set_exception(e)
         except asyncio.InvalidStateError:
-            log.warn("Poller waiter was already cancelled")
+            log.warning("Poller waiter was already cancelled")
 
     async def _poll_once(self) -> None:
         """Trigger a single read, notifying listeners of success or error."""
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
lgtm!

sanni-t (approved):
Great debugging!

fsinapi:
Tested on an OT-2 and confirmed that I was able to get the new log entry to show up. Without this PR, the poller would have died.

PR #13848, opened 2023-10-25
feat(shared-data): json protocol schema 8

sfoster1:
Add JSON protocol schema 8, which is similar in semantic concept to json protocol schema 7 but defines new keys for many large components (commands, annotations, liquids, labware) that set the expected schema for the content for that component.

For instance, now you specify a `commandSchemaId` and the elements of the `commands` array have to match that. You specify a `liquidSchemaId` and the elements of the `liquid` object have to match that.

This hopefully will let us be more flexible with upgrading things like the kinds of commands a protocol can run without having to change the protocol schema itself. Consumers will have to make sure that they can handle the schema for the components, but there's a new provided typescript validator (in `shared-data/js/protocols.ts`) that will validate any version of the protocol we support and return it with appropriate types; that's a bit more spread around in python but it similarly handles everything.

Also, you specify the deck schema version now.

This should be integrated into the robot for execution and we should make sure the app doesn't choke on it, but really it's a good way to be able to add new commands without making other protocols accidentally invalid.

This should be a big step forward in flexibility for our protocol schemas.

## testing
- [x] odd can render protocols tagged jp8
- [x] app can render protocols tagged jp8
- [x] offline analysis can handle protocols tagged jp8
- [x] robot analysis can handle protocols tagged jp8
- [x] robot can execute protocols tagged jp8
- [x] if you have a jp8 protocol run, analysis, and protocol saved on the robot and then you downgrade, the server is still operational


shared-data/command-annotation/schemas/1.json
b-cooper, line 1: This directory should be camelCase like the reference in the v8 schema `commandAnnotation/schemas`

shared-data/protocol/schemas/8.json
```diff
@@ -0,0 +1,161 @@
...
+    "metadata",
+    "robot",
+    "labwareDefinitions",
+    "labwareDefinitionSchemaId",
+    "commands",
+    "commandsSchemaId",
+    "commandAnnotations",
+    "commandAnnotationsSchemaId",
+    "liquids",
+    "liquidSchemaId"
```
SyntaxColoring: Hmm, I'm not sure about decoupling the `liquids` and `labwareDefinitions` schemas like this.

The argument for doing this for `commands` and `commandAnnotations` is that we expect them to change much more frequently than the other things in these files. Do we expect `liquids` and `labware` to change that frequently?

The downside that I have in mind is basically just complexity from the indirection.

sfoster1: labware, fair enough, but yeah I expect liquids to change pretty frequently as we start to actually like do stuff with them

jbleon95:
As discussed in the comments on #13854, we should add some sort of `deckVersion` field along with `deckId` to allow robot server/protocol runner to know whether to load fixed trash or not, rather than just relying on schema version moving forward.

shared-data/protocol/schemas/8.json
```diff
@@ -0,0 +1,161 @@
...
+    "commandAnnotationSchemaId": {
+      "description": "The schema to which the command annotations adhere.",
+      "type": "string"
+    },
+
+    "commandAnnotations": {
+      "type": "array",
+      "items": { "$ref": "#/properties/commandAnnotationSchemaRef" }
+    }
+  }
```
jerader: what is an example of `commandAnnotations`?

sfoster1: Something like grouping commands into the higher-level (i.e. PD, or python api call) command that created them, for instance. Or into user-specified arbitrary groups. Or time estimates, or really anything

shared-data/command/types/calibration.ts
```diff
@@ -0,0 +1,84 @@
+import type { CommonCommandRunTimeInfo, CommonCommandCreateInfo } from '.'
+import type {
+  PipetteMount,
+  GantryMount,
+  LabwareOffset,
+  Coordinates,
+} from '../../js/types'
+// TODO (sb 10/26/22): Separate out calibration commands from protocol schema in RAUT-272
```
jerader, line 8: I think we can delete this TODO, cc @smb2268

shared-data/js/protocols.ts
```diff
@@ -0,0 +1,287 @@
+// Helper functions for validating protocols
+
+import type * as ProtocolSchemas from '../protocol'
+import type { CreateCommand } from '../command/types'
+import type { CommandAnnotation } from '../commandAnnotation/types'
+import Ajv from 'ajv'
```
jerader, line 5: can we move the Ajv import to be at the top of the imports? And then move the `type` imports to be at the bottom of the imports before line 22?

shared-data/js/protocols.ts
```diff
@@ -0,0 +1,287 @@
...
+    const ok = validateProtocol(toValidate)
+    if (!ok) {
+      // eslint-disable-next-line prefer-promise-reject-errors
+      reject(validateProtocol.errors)
+    }
+    resolve(toValidate as ProtocolSchemas.ProtocolFileV7)
+  })
+
+// note: rejects with an array of ajv errors
+export function validate(
```
jerader, line 257: Maybe we can leave a note in here saying this is used in the `protocolValidation.test` since other files in the `shared-data/js` directory contain utils used outside of tests.

jerader (approved):
I reviewed the js stuff and sprinkled in a few clean up suggestions & notes. But lgtm! 🎉

shared-data/js/protocols.ts
sfoster1: Oh this definitely will be used outside of tests, I just don't want to gate merging this pr on "switching the entire js ecosystem over to json protocol v8". The idea with this function is that you'll use it to validate protocols in js anywhere they appear since it will do the typecasting for you.

robot-server/tests/integration/conftest.py
```diff
@@ -91,6 +91,14 @@ def _ot3_session_server(server_temp_directory: str) -> Generator[str, None, None
...
         f"{_SESSION_SERVER_SCHEME}{_SESSION_SERVER_HOST}:{_OT3_SESSION_SERVER_PORT}"
     )
+
+    with _requests_session() as requests_session:
+        has_dev_server = _check_is_ready(requests_session, base_url)
+    if has_dev_server:
+        print("Using pre-existing flex dev server")
+        yield base_url
+        return
+
```
SyntaxColoring: Is this intended to let you have `make -C robot-server dev` open in a second terminal, and if you run `make -C robot-server test`, it re-uses it? 

If so, that seems surprising. I expect tests to be isolated. If the tests can't run isolated because the port is already taken by a dev server, I expect them to error out. Could we leave this out?

Separately: if we did want to keep this in, we should do it consistently between the OT-2 and Flex dev servers.

sfoster1: yeah i can remove it

shared-data/errors/definitions/1/errors.json
```diff
@@ -205,6 +205,10 @@
     "4005": {
       "detail": "Command Parameter Limit Violated",
       "category": "generalError"
+    },
+    "4006": {
+      "detail": "Invalid Protocol Data",
+      "category": "generalError"
```
SyntaxColoring, line 211: Nice, thank you.

SyntaxColoring (approved):
Python changes look great. Thank you!

b-cooper (approved):
🏈

PR #13849, opened 2023-10-25
fix(app): improve robot update via usb

mjhuff:
Closes RQA-1844

# Overview

After making the initial query to the robot server to create the robot update session, Robot updates utilize the top-level ok property of the response body to determine whether the robot-server is ready to begin the update. The ok property is special cased for USB and expects a status text to exist in the response body, however this status text does not match the usual expectation when uploading a file to initiate the update. 

FIX: The ok property for USB is now determined by a 2XX status code.

This PR also handles cases where the robot-server has an existing session and a new USB robot update is initiated (such as during connectivity loss). The session is cleared on the robot-server before a new session is initiated.

# Test Plan
- Unplug laptop & robot from wifi/ethernet and plug in USB.
- Test that updating via zip-file works.
- Restart the app and initiate another update. Unplug the USB sometime during the update.
- See the error occur, replug the USB, click exit (try again is not working with zip-files, this is a separate, known issue), and initiate the update with a _different_ zip file. Alternatively, unplug the USB during the update and close/reopen the app. 

# Changelog
- Improve robot updates via USB.

# Risk assessment
problem = high
solution = low

### DIFF
--- a/app/src/redux/robot-api/http.ts
+++ b/app/src/redux/robot-api/http.ts
@@ -4,6 +4,7 @@ import { map, switchMap, catchError } from 'rxjs/operators'
 import mapValues from 'lodash/mapValues'
 import toString from 'lodash/toString'
 import omitBy from 'lodash/omitBy'
+import inRange from 'lodash/inRange'
 
 import { OPENTRONS_USB } from '../../redux/discovery'
 import { appShellRequestor } from '../../redux/shell/remote'
@@ -53,11 +54,6 @@ export function fetchRobotApi(
     options.body = reqForm
   }
 
-  const isFirstDigitTwo = function (statusCode: number): boolean {
-    const len = Math.floor(Math.log(statusCode) / Math.LN10)
-    return ((statusCode / Math.pow(10, len)) % 10 | 0) === 2
-  }
-
   return host.ip === OPENTRONS_USB
     ? from(
         appShellRequestor({
@@ -73,7 +69,7 @@ export function fetchRobotApi(
           body: response?.data,
           status: response?.status,
           // appShellRequestor eventually calls axios.request, which doesn't provide an ok boolean in the response
-          ok: isFirstDigitTwo(response?.status),
+          ok: inRange(response?.status, 200, 300),
         }))
       )
     : from(fetch(url, options)).pipe(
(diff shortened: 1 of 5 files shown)


app/src/redux/robot-update/__tests__/epic.test.ts
```diff
@@ -286,22 +286,27 @@ describe('robot update epics', () => {
       })
     })
 
-    it('issues error if begin request fails without 409', () => {
+    it('sends request to cancel URL if a non 409 occurs and reissues CREATE_SESSION', () => {
```
shlokamin, line 289: prolly a good idea to also test that if the cancel response is not okay we issue a `unexpectedRobotUpdateError` action

shlokamin (approved):

app/src/redux/robot-api/http.ts
```diff
@@ -68,7 +68,10 @@ export function fetchRobotApi(
           body: response?.data,
           status: response?.status,
           // appShellRequestor eventually calls axios.request, which doesn't provide an ok boolean in the response
-          ok: response?.statusText === 'OK',
+          ok:
+            response?.statusText === 'OK' ||
```
sfoster1: IMO we should just rely on the status code, or at least the status code preferentially

mjhuff: @shlokamin @brenthagen Was there a specific reason we didn't do this? I'm fine making the change if we think it won't break anything. I do think the OK check is the safer option if we're uncertain, though.

mjhuff: @sfoster1 I believe at least part of the thinking was a 3xx status code does not necessarily indicate an issue, so we open ourselves up to needing to cover every possible code. I suppose we could do some sort of generic 2XX/3XX check.

brenthagen: reading statusText was just an initial attempt to patch the lack of a response boolean in axios - based on this i think we're "ok" doing a 2xx status check here https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

seems anything else would go beyond what node-fetch provides

shlokamin: fine with me, lets go with the spec above and give an ok status for any 2xx status

mjhuff: This makes sense as long as we've been using "ok" as a 2XX substitute wherever else it is used...which I think is just here...so yeah!

PR #13852, opened 2023-10-26
fix(app): remove redundant check isOnDevice ternary

ncdiehl11:
# Overview

There is an isOnDevice ternary for an icon size in the JogToPosition tile for DropTipWizard. This check is within a return block that only executes if onDevice, so I remove the ternary here.

# Review requests

@koji spotted the redundant check

# Risk assessment

low

### DIFF
--- a/app/src/organisms/DropTipWizard/JogToPosition.tsx
+++ b/app/src/organisms/DropTipWizard/JogToPosition.tsx
@@ -70,7 +70,7 @@ const ConfirmPosition = (props: ConfirmPositionProps): JSX.Element | null => {
           <Flex gridGap={SPACING.spacing24}>
             <Icon
               name="ot-alert"
-              size={'3.75rem'}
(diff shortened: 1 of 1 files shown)


app/src/organisms/DropTipWizard/JogToPosition.tsx
```diff
@@ -70,7 +70,7 @@ const ConfirmPosition = (props: ConfirmPositionProps): JSX.Element | null => {
           <Flex gridGap={SPACING.spacing24}>
             <Icon
               name="ot-alert"
-              size={isOnDevice ? '3.75rem' : '2.5rem'}
+              size={'3.75rem'}
```
koji: nit

```suggestion
              size="3.75rem"
```

koji (approved):
LGTM
Thank you for the quick fix!

PR #13854, opened 2023-10-26
refactor(api, robot-server, shared-data): replace v3 deck definition with v4 in protocol api and engine

jbleon95:
# Overview

This PR switches the majority of references of `DeckDefinitionV3` to `DeckDefinitionV4`, besides a few legacy places in the code and tests. Additionally, the old fixed trash definitions were re-added to the new deck definition under a new name to reflect that it's a legacy feature. This has been hooked up to protocol engine to only load them if a flag is set in the `create_protocol_engine` function, with a helper `should_load_fixed_trash` function used to determine if it should. The criteria for that is being associated with a Python protocol of API level 2.15 and below, and a JSON protocol with schema version 7 and below.

A new basic `DeckConfigurationProvider` class was also added to protocol engine, with some first starts at some functionality to read in a deck configuration and provide the allowed addressable areas.

Some temporary workaround code has been used to transform the data in the new deck definition to `SlotDefV3` which mirrors the old `orderedSlots` data. The same has been used in the `Deck` class, which will need further refactor or replacement to work with the new deck configuration paradigm where slots are not guaranteed.

# Test Plan

Ran through a simple protocol on OT-2 (both 2.13 and 2.15) and on Flex to ensure that protocol does not error out unexpectedly and that geometry and movement is unchanged. Additionally went through LPC on both robots (both tip and probe based LPC) to ensure that they did not raise any errors.

# Changelog

- Switched most references of `DeckDefinitionV3` to `DeckDefinitionV4`
- Added a `DeckConfigurationProvider` class to protocol engine that uses the new deck definition
- Added in some temporary shim code to provide slot data and definitions using the new definition
- Put fixed trash data back into deck definition under the renamed property `legacyFixtures`
- ProtocolEngine will no longer automatically load fixed trash unless a flag is set in `create_protocol_engine`
- Added helper function `should_load_fixed_trash` which if associated with a protocol, will set the above flag to True if the API level is 2.15 and below, or schema version 7 and below

# Review requests

Does the logic and assumptions with legacy fixed trash loading make sense, are we okay with some of the temporary code added, and does the shape of `DeckConfigurationProvider` mesh with our expectations.

# Risk assessment

Medium. Most functionality has been refactored so that old protocols will run with no issues. The one big change this introduces is that empty runs (not associated with a protocol) and maintenance runs will no longer automatically load a fixed trash. Anything that runs on the assumption that there will be a fixed trash in those instances will no longer function as expected unless they explicitly load that trash.


api/src/opentrons/protocols/api_support/deck_type.py
```diff
@@ -10,6 +17,26 @@
 STANDARD_OT3_DECK = "ot3_standard"
 
 
+LOAD_FIXED_TRASH_GATE_VERSION_PYTHON = APIVersion(2, 15)
+# TODO(jbl 2023-10-26) potentially replace using schema version in JSON protocols with another/new field
```
sfoster1: json protocols actually have a [`deckId` field in the robot-identification section](https://github.com/Opentrons/opentrons/blob/086b9c7eb144f643553bdd81109a558bec16e154/shared-data/protocol/schemas/7.json#L217-L221) that we can use for exactly this (and we'll preserve it in jp8 #13848 )

sfoster1 (commented):
Looking really good, just letting you know json protocols have a deck id entry

api/src/opentrons/protocols/api_support/deck_type.py
jbleon95: The current `deckId` field does not have a deck version property. As we discussed, that can be added to v8 and this logic can be updated then, and a relevant TODO comment has been added.

api/src/opentrons/execute.py
```diff
@@ -536,6 +538,7 @@ def _create_live_context_pe(
             config=_get_protocol_engine_config(),
             drop_tips_after_run=False,
             post_run_hardware_state=PostRunHardwareState.STAY_ENGAGED_IN_PLACE,
+            load_fixed_trash=api_version <= LOAD_FIXED_TRASH_GATE_VERSION_PYTHON,
```
SyntaxColoring: Let's move this `<=` check to be defined in the same place as your `should_load_fixed_trash()` function. You could either add a separate function that takes an `APIVersion` instead of a `ProtocolSource`, or you could have that one function accept a `Union`.

shared-data/python/opentrons_shared_data/deck/__init__.py
```diff
@@ -17,7 +17,8 @@
         DeckSchemaVersion4,
     )
 
-DEFAULT_DECK_DEFINITION_VERSION: Final = 3
+DEFAULT_DECK_DEFINITION_VERSION: Final = 4
+DEFAULT_LEGACY_DECK_DEFINITION_VERSION: Final = 3
```
SyntaxColoring: I think we should keep `DEFAULT_LEGACY_DECK_DEFINITION_VERSION` as a constant in `opentrons.protocol_api.core.legacy`.

Just for separation-of-concerns reasons, basically. "Legacy" here means "JSONv≤5 and PAPIv≤2.13," which are higher-level concepts that deck definitions are unaware of.

And now that I'm thinking about it, `DEFAULT_DECK_DEFINITION_VERSION` should probably be spelled `LATEST_DECK_DEFINITION_VERSION`. Otherwise, it's like, for whom is this the correct default? But that's a separate thing.

api/src/opentrons/protocol_api/core/legacy/legacy_protocol_core.py
```diff
@@ -452,10 +452,14 @@ def get_labware_on_labware(
...
         assert False, "get_labware_on_labware only supported on engine core"
 
-    def get_deck_definition(self) -> DeckDefinitionV3:
+    def get_deck_definition(self) -> DeckDefinitionV4:
         """Get the geometry definition of the robot's deck."""
         assert False, "get_deck_definition only supported on engine core"
 
+    def get_slot_definition(self, slot: DeckSlotName) -> SlotDefV3:
+        """Get the slot definition from the robot's deck."""
+        assert False, "get_deck_definition only supported on engine core"
```
SyntaxColoring:
```suggestion
        assert False, "get_slot_definition only supported on engine core"
```

api/src/opentrons/protocols/api_support/deck_type.py
```diff
@@ -10,6 +17,26 @@
...
+
+
+def should_load_fixed_trash(protocol_config: ProtocolConfig) -> bool:
+    """Decide whether to automatically load fixed trash on the deck based on version."""
+    load_fixed_trash = False
+    if isinstance(protocol_config, PythonProtocolConfig):
+        load_fixed_trash = (
+            protocol_config.api_version <= LOAD_FIXED_TRASH_GATE_VERSION_PYTHON
+        )
+    # TODO(jbl 2023-10-27), when schema v8 is out, use a new deck version field to support fixed trash protocols
```
SyntaxColoring, line 57: Hmm, I'm not sure about this. Let me think things through and I'll open a thread in Slack.

api/src/opentrons/protocol_engine/resources/deck_data_provider.py
```diff
@@ -51,12 +51,12 @@ def sync() -> DeckDefinitionV3:
...
     async def get_deck_fixed_labware(
         self,
-        deck_definition: DeckDefinitionV3,
+        deck_definition: DeckDefinitionV4,
     ) -> List[DeckFixedLabware]:
         """Get a list of all labware fixtures from a given deck definition."""
         labware: List[DeckFixedLabware] = []
 
-        for fixture in deck_definition["locations"]["fixtures"]:
+        for fixture in deck_definition["locations"]["legacyFixtures"]:
```
SyntaxColoring, line 59: Good call renaming this key to anything but `fixtures`, lol.

api/src/opentrons/protocol_engine/resources/deck_configuration_provider.py
```diff
@@ -0,0 +1,69 @@
...
+    """Basic cutout fixture data class."""
+
+    name: str
+    slot_location: DeckSlotName
+
+
+class DeckConfigurationProvider:
+    """Provider class to ingest deck configuration data and retrieve relevant deck definition data."""
+
+    # TODO maybe make the key a DeckSlot enum?
```
SyntaxColoring: Nah, I think we should be careful to draw a distinction between deck slots and cutouts. They're named identically ("A3", "C2", ...), but they're different namespaces.

Staging area slots like C4 will *potentially* be in the `DeckSlotName` enum, and if they are, those values certainly won't make sense as keys for this dict.

api/src/opentrons/protocol_engine/resources/deck_configuration_provider.py
```diff
@@ -0,0 +1,69 @@
...
+
+from ..errors import FixtureDoesNotExistError
+
+
+@dataclass(frozen=True)
+class DeckCutoutFixture:
+    """Basic cutout fixture data class."""
+
+    name: str
+    slot_location: DeckSlotName
```
SyntaxColoring: Is this the slot location that this cutout fixture is providing, or the cutout that it's mounted onto? If it's the latter, see my other comment about drawing a distinction between `DeckSlotName`s and cutout names.

api/src/opentrons/protocol_engine/resources/deck_configuration_provider.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_addressable_areas_for_cutout_fixture(
+        self, cutout_fixture_id: str, deck_slot_id: str
+    ) -> List[str]:
+        """Get the allowable addressable areas for a cutout fixture loaded on a specific slot."""
+        for cutout_fixture in self._deck_definition["cutoutFixtures"]:
+            if cutout_fixture_id == cutout_fixture["id"]:
+                return cutout_fixture["providesAddressableAreas"].get(deck_slot_id, [])
+
+        # TODO need to decide whether an invalid combo should raise an error or just return empty list
```
SyntaxColoring: Raise an error, in my opinion. "You've placed a fixture here and the fixture has provided nothing" seems semantically distinct from "you cannot place that fixture there."

But the way we expect this to be used, I think we can probably get it to work either way, so it's a matter of style.

api/src/opentrons/protocol_engine/resources/deck_configuration_provider.py
```diff
@@ -0,0 +1,69 @@
...
+        """Initialize a DeckDataProvider."""
+        self._deck_definition = deck_definition
+        self._configuration = {
+            cutout_fixture.slot_location.id: cutout_fixture
+            for cutout_fixture in deck_configuration
+        }
+
+    def get_addressable_areas_for_cutout_fixture(
+        self, cutout_fixture_id: str, deck_slot_id: str
+    ) -> List[str]:
```
SyntaxColoring: Also purely a matter of style, but I would consider having all of this stuff operate on `Set` inputs and outputs instead of `List`. There isn't an order to a deck configuration, and a deck configuration never has duplicate elements.

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -304,16 +304,62 @@ def get_display_name(self, labware_id: str) -> Optional[str]:
...
+                base_position = cutout["position"]
+                break
+        else:
+            raise errors.SlotDoesNotExistError(
+                f"Slot ID {slot.id} does not exist in deck {deck_def['otId']}"
+            )
+
+        slot_def: SlotDefV3
+        # Slot 12/fixed trash for ot2 is a little weird so if its that just return some hardcoded stuff
+        if slot.id == "12":
```
SyntaxColoring, line 330: Hmm, what's the weirdness that this hard-coding is working around?

api/src/opentrons/simulate.py
```diff
@@ -792,6 +794,7 @@ def _create_live_context_pe(
             config=_get_protocol_engine_config(robot_type),
             drop_tips_after_run=False,
             post_run_hardware_state=PostRunHardwareState.STAY_ENGAGED_IN_PLACE,
+            load_fixed_trash=api_version <= LOAD_FIXED_TRASH_GATE_VERSION_PYTHON,
```
SyntaxColoring: Ditto my comment in `execute.py`.

SyntaxColoring (commented):
This great. Thanks for slogging through this. I think this takes care of one of the major scary parts of this release.

My comments so far are mostly minor:

shared-data/deck/definitions/4/ot2_standard.json
```diff
@@ -349,6 +349,14 @@
...
         "displayName": "Slot 11 Top Right Dot"
       }
+    ],
+    "legacyFixtures": [
+      {
+        "id": "fixedTrash",
+        "slot": "12",
+        "labware": "opentrons_1_trash_1100ml_fixed",
+        "displayName": "Fixed Trash"
+      }
```
SyntaxColoring, line 359: For OT-2s, how does this interact with the "new" `fixedTrash` fixture in the `fixtures` key?

api/src/opentrons/protocol_engine/state/labware.py
jbleon95: A combination of two things. One is that there is no addressable area named `12` for the new OT2 deck definitions, only `fixedTrash`, reflecting that there is no actual slot 12. The other related thing is that `fixedTrash` bounding box now reflects the dimension of the trash, not the generic ANSI slot dimensions. To preserve existing behavior, I added this hard-coding. When we fully move to new slot and deck logic, this can all go away.

robot-server/tests/integration/http_api/runs/test_run_queued_protocol_commands.tavern.yaml
```diff
@@ -89,12 +89,7 @@ stages:
           current: True
           errors: []
           id: '{run_id}'
-          labware:
-            - id: 'fixedTrash'
-              loadName: 'opentrons_1_trash_1100ml_fixed'
-              definitionUri: 'opentrons/opentrons_1_trash_1100ml_fixed/1'
-              location:
-                slotName: '12'
+          labware: []
```
SyntaxColoring, line 92: Hmm. For existing protocols, I don't think we want this change.

These existing protocols are going to continue interacting with the fixed trash as a labware, not a fixture. They have `dropTip` commands that say `labwareId: "fixedTrash"`, for example. So it's correct for this field to say that there is a `"fixedTrash"` labware loaded in the run and referenceable by its commands.

If anything, I think we want to move in the opposite direction. These protocols have had `"fixedTrash"` entries in the `labware` list, but no corresponding `loadLabware` commands. To be internally consistent, they should have corresponding `loadLabware` commands. RSS-250

@sanni-t Does this match your expectations?

api/tests/opentrons/conftest.py
```diff
@@ -300,6 +310,7 @@ def _make_ot3_pe_ctx(
         ),
         drop_tips_after_run=False,
         post_run_hardware_state=PostRunHardwareState.STAY_ENGAGED_IN_PLACE,
+        load_fixed_trash=True,  # TODO change how this works, make a new one?
```
SyntaxColoring: Yeah...I suspect things would get easier if, instead of a global `ctx` fixture that has to get configured through sub-fixtures, we just gave tests a helper function to let them directly construct the kind of `ProtocolContext` that they actually need.

Hard-coding `load_fixed_trash=True` seems fine for now, until something needs a setup where it's `False`. And then we'll have to figure it out there.

shared-data/deck/definitions/4/ot2_standard.json
jbleon95: The `legacyFixtures` fixed trash will only be loaded for protocols 2.15 and below. For anything higher, we will use the new fixed trash addressable area in code. Even though OT-2 will have only one valid deck configuration (at least for this first iteration of deck config), it'll use the same addressable area/cutout fixture system as Flex for code simplicity.

robot-server/tests/integration/http_api/runs/test_run_queued_protocol_commands.tavern.yaml
jbleon95: This specific test has an empty run created, which is one of the cases where we shouldn't be automatically loading fixed trash. When a run is created with a protocol and that is below schema v8 or API level 2.16, fixed trash will be loaded automatically as a labware like it used to be.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1454,7 +1461,7 @@ def tip_racks(self, racks: List[labware.Labware]) -> None:
 
     @property  # type: ignore
     @requires_version(2, 0)
-    def trash_container(self) -> labware.Labware:
+    def trash_container(self) -> Optional[labware.Labware]:
```
sfoster1: These are API changes, and I don't think we should make them. Instead, I think we should have this raise `NoTrashDefinedError` if there is no trash defined and keep the return type not optional. This should result in no change in behavior in API versions too low to alter the trash, which is what we want, and if we want to provide a way for someone to get the trash if present and None if not we should add a new call with an appropriate added-in version.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -979,7 +979,7 @@ def deck(self) -> Deck:
 
     @property  # type: ignore
     @requires_version(2, 0)
-    def fixed_trash(self) -> Labware:
+    def fixed_trash(self) -> Optional[Labware]:
```
sfoster1: see the note on `InstrumentContext.trash_container`

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1464,7 +1471,7 @@ def trash_container(self) -> labware.Labware:
         return self._trash
 
     @trash_container.setter
-    def trash_container(self, trash: labware.Labware) -> None:
+    def trash_container(self, trash: Optional[labware.Labware]) -> None:
```
sfoster1: This is an API change also. It's probably fine because it's more permissive rather than less, but it probably should raise an exception if the api version is below that version in which we added the possibility for this to be none. I think it's probably also a fine idea to not make this change, since I don't quite get why someone would want to explicitly remove a trash from a pipette, or have a different API for that.

api/src/opentrons/protocols/api_support/deck_type.py
```diff
@@ -10,6 +17,36 @@
 STANDARD_OT3_DECK = "ot3_standard"
 
 
+LOAD_FIXED_TRASH_GATE_VERSION_PYTHON = APIVersion(2, 15)
+LOAD_FIXED_TRASH_GATE_VERSION_JSON = 7
+
+
+class NoTrashDefinedError(Exception):
```
sfoster1: make this an enumerated error

sfoster1 (changes requested):
Couple little things about changes to the python protocol API.

Also, how downgrade-safe are stored runs and analyses with this change?

sfoster1 (approved):
Looks good to me!

SyntaxColoring (approved):
yea boiii

SyntaxColoring:
> Also, how downgrade-safe are stored runs and analyses with this change?

There's always going to be a hazard with RSS-98, but if we had problems in this particular PR, I'm like 95% sure our integration persistence tests would detect them.

PR #13855, opened 2023-10-26
feat(app): ODD protocol detail and protocol setup status for deck configuration

ncdiehl11:
closes [RAUT-667](https://opentrons.atlassian.net/browse/RAUT-667)

# Overview

The ODD ProtocolSetup page should render the correct module and deck status taking into account module calibration and connection, and now deck configuration. We need to check for the current deck configuration state to ensure the protocol doesn't require hardware where a fixture currently exists, and alert the user of a location conflict if necessary. The status will also render modules and/or fixtures are missing. This hardware status will also be rendered as the chipText on the ProtocolDetails page.

# Test Plan

Testing all cases below on a physical Flex

### Protocol requires neither modules nor fixtures

- [x] render disabled setup step with text that no hardware is needed

Screen Shot 2023-11-01 at 1 53 33 PM

### Protocol requires only modules
- [x] connected, and calibrated: setup step is ready and renders number of modules attached
Screen Shot 2023-11-01 at 2 13 45 PM
- [x] connected and not calibrated: setup step is not ready and renders 'Calibration required'
Screen Shot 2023-11-01 at 2 00 08 PM
- [x] not connected: setup step is not ready and renders "Missing <moduleName>" or "Missing <#> modules"
Screen Shot 2023-11-01 at 2 03 18 PM
Screen Shot 2023-11-01 at 2 06 26 PM
- [x] location conflict with deck configuration: setup step is not ready and renders location conflict
Screen Shot 2023-11-01 at 2 07 51 PM

### Protocol requires only fixtures
- [x] properly configured: setup step is ready and renders number of fixtures attached
Screen Shot 2023-11-01 at 2 22 43 PM
- [x] location conflict with deck configuration: setup step is not ready and renders location conflict
Screen Shot 2023-11-01 at 2 25 03 PM
- [x] not configured: setup step is not ready and renders "Missing <fixtureName> or "Missing <#> fixtures"
Screen Shot 2023-11-01 at 2 27 16 PM
Screen Shot 2023-11-01 at 2 29 55 PM

### Protocol requires both modules and fixtures
- [x] location conflict for any modules or fixtures with deck configuration: setup step is not ready and renders location conflict
- [x] fixtures configured but modules missing: setup step is not ready and renders "Missing <moduleName>" or "<#> modules missing"
Screen Shot 2023-11-01 at 2 35 03 PM
- [x] modules connected but fixtures missing: setup step is not ready and renders "Missing <fixtureName>" or "<#> fixtures missing"
Screen Shot 2023-11-01 at 2 49 28 PM
- [x] fixtures configured and modules connected but module calibrations missing: setup step is not ready and renders "Calibration required"
Screen Shot 2023-11-01 at 2 52 02 PM
- [x] fixtures not configured and modules not connected: setup step is not ready and renders "Missing hardware"
Screen Shot 2023-11-01 at 2 37 05 PM
- [x] fixtures configured and modules connected and calibrated: setup step is ready and renders number of fixtures and modules attached
Screen Shot 2023-11-01 at 2 51 10 PM

# Changelog

- add hooks for missing and required protocol hardware to optionally accept a protocol analysis object instead of a protocol ID (when called from ProtocolSetup page rather than ProtocolDetails page)
- change rendered text in Modules and deck setup step depending on conflicting, missing, or un-calibrated hardware
- add a hook for getting the number of unready instruments to render on the instruments setup step
- add keys to translation files
- update tests to look for updated for setup step text

# Review requests

@brenthagen as discussed in person

# Risk assessment
medium. Main risk lies in the enabling/disabling of the protocol Play button based on new deck configuration considerations

[RAUT-667]: https://opentrons.atlassian.net/browse/RAUT-667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/pages/OnDeviceDisplay/ProtocolSetup/__tests__/ProtocolSetup.test.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react'
 import { Route } from 'react-router'
 import { MemoryRouter } from 'react-router-dom'
 import { when, resetAllWhenMocks } from 'jest-when'
+import { UseQueryResult } from 'react-query'
```
koji, line 5:
```suggestion
import type { UseQueryResult } from 'react-query'
```

koji: this can be before `import type { DeckConfiguration,`
```suggestion
import type { UseQueryResult } from 'react-query'
```

app/src/assets/localization/en/protocol_setup.json
```diff
@@ -62,14 +62,19 @@
   "feedback_form_link": "Let us know!",
   "fixture_name": "fixture",
   "fixture": "Fixture",
+  "fixtures_connected_plural": "{{count}} fixtures attached",
+  "fixture_connected": "{{count}} fixture attached",
```
jerader: should you have a `count` const in the nonplural i18n key? Since you aren't counting it?

app/src/assets/localization/en/protocol_setup.json
```diff
@@ -62,14 +62,19 @@
...
   "get_labware_offset_data": "Get Labware Offset Data",
+  "hardware_missing": "Hardware missing",
   "heater_shaker_extra_attention": "Use latch controls for easy placement of labware.",
   "heater_shaker_labware_list_view": "To add labware, use the toggle to control the latch",
   "how_offset_data_works": "How labware offsets work",
   "initial_liquids_num_plural": "{{count}} initial liquids",
   "initial_liquids_num": "{{count}} initial liquid",
   "initial_location": "Initial Location",
   "install_modules_and_fixtures": "Install the required modules and power them on. Install the required fixtures and review the deck configuration.",
+  "instrument_calibrations_missing_plural": "Missing {{num}} calibrations",
```
jerader: this seems like it should be "Missing {{count}} calibrations"

app/src/assets/localization/en/protocol_setup.json
```diff
@@ -114,6 +119,10 @@
   "magnetic_module_extra_attention": "Opentrons recommends securing labware with the module’s bracket",
   "map_view": "Map View",
   "missing": "Missing",
+  "missing_gripper": "Missing gripper",
+  "missing_instruments_plural": "Missing {{num}}",
```
jerader: same here for "Missing {{count}}"

app/src/assets/localization/en/protocol_setup.json
```diff
@@ -135,11 +144,12 @@
   "modules": "Modules",
   "mount_title": "{{mount}} MOUNT:",
   "mount": "{{mount}} mount",
+  "multiple_fixtures_missing": "{{count}} fixtures missing",
   "multiple_modules_example": "Your protocol has two Temperature Modules. The Temperature Module attached to the first port starting from the left will be related to the first Temperature Module in your protocol while the second Temperature Module loaded would be related to the Temperature Module connected to the next port to the right. If using a hub, follow the same logic with the port ordering.",
   "multiple_modules_explanation": "To use more than one of the same module in a protocol, you first need to plug in the module that’s called first in your protocol to the lowest numbered USB port on the robot. Continue in the same manner with additional modules.",
   "multiple_modules_help_link_title": "See How To Set Up Modules of the Same Type",
   "multiple_modules_learn_more": "Learn more about using multiple modules of the same type",
-  "multiple_modules_missing": "Multiple modules missing",
+  "multiple_modules_missing": "{{count}} modules missing",
```
jerader: this isn't using the plural i18n key so should be {{num}}

app/src/organisms/ProtocolSetupInstruments/utils.ts
```diff
@@ -62,3 +62,25 @@ export function getAreInstrumentsReady(
 
   return allSpeccedPipettesReady && isExtensionMountReady
 }
+
+export function getNumUnreadyInstruments(
```
jerader: I would rename this, maybe `getNumIncompleteInstruments`, I mistook `Unready` as `ready` when i first read this.

app/src/organisms/ProtocolSetupInstruments/utils.ts
```diff
@@ -62,3 +62,25 @@ export function getAreInstrumentsReady(
...
+  analysis: CompletedProtocolAnalysis,
+  attachedInstruments: Instruments
+): number {
+  const speccedPipettes = analysis?.pipettes ?? []
+
+  const numUnreadySpeccedPipettes = speccedPipettes.filter(loadedPipette => {
```
jerader: same here

app/src/pages/OnDeviceDisplay/ProtocolSetup/__tests__/ProtocolSetup.test.tsx
```diff
@@ -305,6 +326,12 @@ describe('ProtocolSetup', () => {
       <div>mock ConfirmAttachedModal</div>
     )
     mockUseDoorQuery.mockReturnValue({ data: mockDoorStatus } as any)
+    mockUseModulesQuery.mockReturnValue({
+      data: { data: [mockHeaterShaker] },
+    } as any)
+    when(mockUseDeckConfigurationQuery).mockReturnValue({
+      data: [mockFixture],
+    } as UseQueryResult<DeckConfiguration>)
```
jerader: nit, since you're not specifying `calledWith()` in here, you can simplify this:

```suggestion
    mockUseDeckConfigurationQuery.mockReturnValue({
      data: [mockFixture],
    } as UseQueryResult<DeckConfiguration>)
```

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -47,6 +48,7 @@ import {
 } from '../../../organisms/Devices/hooks'
 import { useMostRecentCompletedAnalysis } from '../../../organisms/LabwarePositionCheck/useMostRecentCompletedAnalysis'
 import { getProtocolModulesInfo } from '../../../organisms/Devices/ProtocolRun/utils/getProtocolModulesInfo'
+import type { ProtocolModuleInfo } from '../../../organisms/Devices/ProtocolRun/utils/getProtocolModulesInfo'
```
jerader: this type import can go lower

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -342,25 +416,91 @@ function PrepareToRun({
...
+      : t('multiple_fixtures_missing', { count: missingFixtures.length })
+
+  const missingMultipleHardwareTypes =
+    [missingModules, missingFixtures].filter(
+      missingHardwareArr => missingHardwareArr.length > 0
+    ).length > 1
+
+  const modulesDetail = (): { detail: string; subdetail?: string } => {
+    if (isLocationConflict) {
+      return { detail: t('location_conflict') }
```
jerader: it would be good to try to test all these different return options. it might be easier to split the `modulesDetail` function up into its own util and then just test that util.

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -471,9 +611,12 @@ function PrepareToRun({
             <ProtocolSetupStep
               onClickSetupStep={() => setSetupScreen('modules')}
               title={enableDeckConfig ? t('modules_and_deck') : t('modules')}
-              detail={modulesDetail()}
+              detail={modulesDetail().detail}
+              subDetail={modulesDetail().subdetail}
```
jerader: nit: instead of calling this function twice, I'd deconstruct it up above like `const {detail, subdetail} = modulesDetail()`

jerader (commented):
looking good to far! The main thing I noticed in your i18n strings is you're inconsistent when you use `{{count}}` vs `{{num}}`. Since `count` is associated with the plural usage, it would be good to only use `{{count}}` when using `plural`.

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
ncdiehl11: completely agree, my bad

app/src/assets/localization/en/protocol_setup.json
ncdiehl11: I saw this precedent in other cases where the count is necessarily equal to 1, but I can remove

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -81,6 +83,11 @@ import { CloseButton, PlayButton } from './Buttons'
 
 import type { Cutout, FixtureLoadName } from '@opentrons/shared-data'
 import type { OnDeviceRouteParams } from '../../../App/types'
+import {
+  useRequiredProtocolHardwareFromAnalysis,
+  useMissingProtocolHardwareFromAnalysis,
+} from '../../Protocols/hooks'
```
brenthagen: this import can be grouped further up near the other non-type imports

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -268,16 +288,16 @@ function PrepareToRun({
     attachedModules,
     protocolModulesInfo
   )
-  const areInstrumentsReady =
+  const numUnreadyInstruments =
```
brenthagen: i'd suggest explicitly giving `numUnreadyInstruments` type `number | null` rather than the implicit `number | boolean` here. what we're saying is that if we don't have analysis or instruments data we can't tell how many unready instruments we have - null is the more typical value for the absence of information.

app/src/organisms/ProtocolSetupInstruments/utils.ts
brenthagen: i'd suggest using count rather than num,  `getIncompleteInstrumentCount`

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
brenthagen: i don't think `modulesDetail` should be a function - we can directly assign the detail/subdetail values in the body of the component

app/src/organisms/OnDeviceDisplay/RobotDashboard/hooks/useHardwareStatusText.ts
```diff
@@ -1,51 +1,72 @@
 import { useTranslation } from 'react-i18next'
 
-import { useFeatureFlag } from '../../../../redux/config'
-
 import type { ProtocolHardware } from '../../../../pages/Protocols/hooks'
 
 export function useHardwareStatusText(
```
brenthagen: unless you've heard otherwise from Rob/design, based on my read of figma I think we should leave the changes to `useHardwareStatusText` (and therefore Protocol Details and the recently run protocol card) out of this PR. there are different sets of statuses and text to be displayed in each of the 3 situations, and we should limit the changes in this PR to the protocol setup page

brenthagen (commented):
left a few comments - main one is to make sure we're only making changes to protocol setup status here

app/src/organisms/OnDeviceDisplay/RobotDashboard/hooks/useHardwareStatusText.ts
ncdiehl11: sounds good— I stepped outside of the scope of this ticket to address the Protocols Details status, but I will separate out those changes into another PR which we can decide whether or not to merge now.

brenthagen: that makes sense to me, thanks

app/src/assets/localization/en/protocol_setup.json
jerader: Ah yes you're right! My bad. I think though you can simplify your usages of plural more:

`t('missing_fixture', { count: countMissingFixtures }}`

it should automatically take into account the plural version.

ncdiehl11: Just dug into this for a sec with Shlok, I think the convention is to use i18n key suffixes like `_one` and `_other` ([ref](https://www.i18next.com/translation-function/plurals#singular-plural)), and then leave the suffixes out when calling the keys in the translation

app/src/pages/Protocols/hooks/index.ts
```diff
@@ -214,26 +226,80 @@ export const useRequiredProtocolLabware = (
  * @returns {ProtocolHardware[]} An array of ProtocolHardware objects that are required by the given protocol ID,
  * but not currently connected.
  */
```
brenthagen: comment no longer accurate to function params and return value, either update or remove

app/src/pages/Protocols/hooks/index.ts
```diff
@@ -214,26 +226,80 @@ export const useRequiredProtocolLabware = (
...
-export const useMissingProtocolHardware = (
-  protocolId: string
+
+const useMissingProtocolHardwareFromRequiredProtocolHardare = (
```
brenthagen: typo `useMissingProtocolHardwareFromRequiredProtocolHardware`

app/src/pages/OnDeviceDisplay/ProtocolSetup/index.tsx
```diff
@@ -295,17 +316,68 @@ function PrepareToRun({
...
-    isMissingModules || !moduleCalibrationStatus.complete
-      ? 'not ready'
-      : 'ready'
+    areModulesReady && areFixturesReady && !isLocationConflict
+      ? 'ready'
+      : 'not ready'
 
-  const isReadyToRun = areInstrumentsReady && !isMissingModules
+  const isReadyToRun =
+    incompleteInstrumentCount === 0 && areModulesReady && areFixturesReady
```
brenthagen: until we can remove the stubbed fixtures, let's comment out this line and leave a TODO, otherwise it will block playing all runs based on the stubbed fixtures that don't reflect actual protocols.

and leave in code just `const isReadyToRun = incompleteInstrumentCount === 0 && areModulesReady`

brenthagen (commented):
code makes sense to me, checked a few different states but not all - let's be sure to comment out the `isReadyToRun` condition before merging

brenthagen (approved):
LGTM !

PR #13861, opened 2023-10-27
feat(protocol-engine): Add skeleton for `loadAddressableArea` command

SyntaxColoring:
# Overview
Set up the `loadAddressableArea` interface described in https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3853615149/Deck+configuration+Protocol+Engine+architecture. Closes RSS-338.

# Test Plan

None applicable. This just adds shapes, no logic.

# Changelog

* Do the usual boilerplate for new Protocol Engine commands.
* When you execute the new command, it will raise a `NotImplementedError`.
* Treat this new command as a member of a new command schema, v8.
* Deliberately *do not* update the JSON protocol parsing code to understand command schema v8 instead of v7. That's already happening in #13848.

# Review requests

Take a look at the `loadAddressableArea` docstrings. Any surprises there? Any questions, comments, or concerns about this interface before we increase our commitment to it?

RSS noticed that just loading an addressable area by name leaves the protocol no way to distinguish a movable trash caddy slot A1 from a movable trash caddy in slot  A3 (for example). Our solution to that is to have 8 different movable trash addressable areas. This keeps the approach consistent; we have 4 different addressable areas for the staging area slots, too.

# Risk assessment

Low. Although, we might have to be careful about how this merges with #13848.


SyntaxColoring:
Reminder to self: depending on merge timing, leave a ticket to make sure we account for this new command in JSON protocols.

shared-data/command/schemas/8.json
```diff
@@ -992,6 +995,49 @@
...
+      "type": "object",
+      "properties": {
+        "addressableAreaName": {
+          "title": "Addressableareaname",
```
jerader, line 1004: maybe: 
```suggestion
          "title": "AddressableAreaName",
```

jerader (commented):
I think this is okay, it would be helpful to have a write up of all the `loadAddressableAreaName`s somewhere though. It would help me wrap my head around why  the command doesn't need a location param.

shared-data/command/schemas/8.json
SyntaxColoring: This is unfortunately bad behavior on Pydantic's part, and it happens for every field, everywhere in our APIs. e.g. `pipetteId` becomes `Pipetteid`, and `wellName` becomes `Wellname`. Drives me crazy. :(

We can fix this at the level of individual fields, if we really want. But given how pervasive it is, I'd rather hold out for some kind of global solution.

SyntaxColoring:
This is under question given this morning's meeting between RSS and App+UI. Some people want it so that if you create a maintenance run, you have all of the robot's addressable areas available implicitly without needing to call `loadAddressableArea`. If that's the case, then the whole premise of the `loadAddressableArea` command falls apart, and it probably doesn't make sense to have it at all.

SyntaxColoring:
We're going with a Protocol Engine interface that loads addressable areas implicitly, so this `loadAddressableArea` command is obsolete.

PR #13862, opened 2023-10-27
fix(app): fix selected mount for detach instrument 96 flow

smb2268:
fix: RQA-1833, RQA-1834, RQA-1836, RQA-940, RQA-1262, RQA-1835, RAUT-732, RAUT-733, RAUT-731
# Overview
This PR fixes and updates various 96-channel bugs and enhances error handling for them

# Test Plan
Tests I have run (corresponding to steps of changelog):
1. Launch 96 attach flow from device card with one other pipette on gantry and see that the mount moves down to detach z-axis after the other pipette is successfully removed
2. With one pipette on gantry, launch 96 attach flow from ODD and see that the correct mount moves down for detaching single channel pipette
3. Launch calibration flow for any pipette and try to proceed to calibration without attaching the probe, see that you are prompted to attach the calibration probe
4. Try to exit from `MountingPlate` step of 96 channel process and see that you are now able to exit from the critical step screen
5. Go through 96-channel detach flow and leave the pipette attached - see that you get the "pipette still attached" error on the first step instead of the last step
6. Successfully remove single channel pipettes in 96 attach flow and see that the button no longer overlaps with the text

# Changelog
1. Add `nextMount` to results step in `getPipetteWizardSteps` so mount moves down after single channel detach before 96-channel attach
2. Stop overriding selected mount when selecting attach 96 channel from ODD - we need this info to determine what mount we need to detach a single channel from
3. Start checking `tipDetected` field on `PipetteData` to see whether calibration probe is attached before proceeding to calibration. Build in "try again" error handling loop for this.
4. Allow exiting from any step of the 96-channel process, but show a more serious warning message for steps that will leave the gantry in a weird state (mounting plate attached or z-axis detached)
5. Move logic to check for successful pipette detachment to the `DetachPipette` component so that you see the "pipette still detected" on this screen instead of once you get to results
6. Add margin to subheader in SimpleWizardBody so there isn't overlap with the button when text goes over one line
7. Remove z-axis still attached error loop since it is no longer necessary

# Review requests
Look through code, I've tested pretty robustly

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
+++ b/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
@@ -25,8 +25,8 @@ import pipetteProbe8 from '../../assets/videos/pipette-wizard-flows/Pipette_Prob
 import probing96 from '../../assets/videos/pipette-wizard-flows/Pipette_Probing_96.webm'
 import { BODY_STYLE, SECTIONS, FLOWS } from './constants'
 import { getPipetteAnimations } from './utils'
+import type { PipetteData } from '@opentrons/api-client'
 import type { PipetteWizardStepProps } from './types'
-import { PipetteData } from '@opentrons/api-client'
 
 interface AttachProbeProps extends PipetteWizardStepProps {
   isExiting: boolean
@@ -113,7 +113,7 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
     setIsPending(true)
     refetch()
       .then(() => {
-        if (attachedPipette?.state?.tipDetected === true) {
+        if (attachedPipette?.state?.tipDetected) {
           chainRunCommands?.(
             [
               {
--- a/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
+++ b/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
@@ -213,6 +213,9 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
     return (
       <SimpleWizardBody
         header={t('unable_to_detect_probe')}
+        subHeader={
+          numberOfTryAgains > 2 ? t('something_seems_wrong') : undefined
+        }
         iconColor={COLORS.errorEnabled}
         isSuccess={false}
       >
--- a/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
+++ b/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
@@ -3,12 +3,19 @@ import { css } from 'styled-components'
 import { Trans, useTranslation } from 'react-i18next'
 import {
   Flex,
+  Btn,
+  PrimaryButton,
   TYPOGRAPHY,
   COLORS,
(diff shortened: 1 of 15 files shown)


app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
```diff
@@ -19,6 +26,7 @@ import probing96 from '../../assets/videos/pipette-wizard-flows/Pipette_Probing_
 import { BODY_STYLE, SECTIONS, FLOWS } from './constants'
 import { getPipetteAnimations } from './utils'
 import type { PipetteWizardStepProps } from './types'
+import { PipetteData } from '@opentrons/api-client'
```
jerader: import can go higher

app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
```diff
@@ -50,46 +85,73 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
...
-          },
-        },
-      ],
-      false
-    )
+    setIsPending(true)
+    refetch()
       .then(() => {
-        proceed()
+        if (attachedPipette?.state?.tipDetected === true) {
```
jerader: shouldn't this work?

```suggestion
        if (attachedPipette?.state?.tipDetected) {
```

app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -248,7 +248,20 @@ export const PipetteWizardFlows = (
...
     isOnDevice,
   }
-  const exitModal = (
+  const is96ChannelUnskippableStep =
+    currentStep.section === SECTIONS.CARRIAGE ||
+    currentStep.section === SECTIONS.MOUNTING_PLATE ||
+    (selectedPipette === NINETY_SIX_CHANNEL &&
+      currentStep.section === SECTIONS.DETACH_PIPETTE)
+
+  const exitModal = is96ChannelUnskippableStep ? (
```
jerader, line 257: this refactor reads a lot easier to me than before! 👍

app/src/organisms/PipetteWizardFlows/DetachPipette.tsx
```diff
@@ -90,24 +157,47 @@ export const DetachPipette = (props: DetachPipetteProps): JSX.Element => {
   }
 
   if (isRobotMoving) return <InProgressModal description={t('stand_back')} />
+  if (showPipetteStillAttached)
+    return (
```
jerader: nit: totally optional but maybe the return can be wrapped in curly braces like `if () {return ()}`, would increase readability imo since there is another if statement right above it.

jerader (approved):
lgtm just left a few clean up comments/suggestions!

app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
smb2268: yeah it does! I think I was trying to get rid of a lint warning but it isn't necessary

PR #13863, opened 2023-10-27
feat(protocol-engine): Move above the waste chute if it's mentioned in the protocol

SyntaxColoring:
# Overview

This makes it so that if you run a protocol that mentions the waste chute, the pipettes will physically move high enough to clear the waste chute when moving around the deck.

This is a temporary hack to unblock hardware testing. The hacky part is that we want this obstacle avoidance to happen if the *robot is configured with the waste chute,* but this PR only implements obstacle avoidance if *the protocol mentions a waste chute.* Doing this properly is ticketed as RSS-390, which this goes towards but does not close.

# Test Plan

[Done by @SyntaxColoring]

Run this on a Flex:

```python

    protocol.load_waste_chute(with_lid=True)

    tip_rack = protocol.load_labware(
        "opentrons_flex_96_tiprack_1000ul",
        "D1",
        adapter="opentrons_flex_96_tiprack_adapter",
    )
    well_plate_1 = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "D2"
    )
    well_plate_2 = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "C3"
    )

        "flex_96channel_1000", mount="left", tip_racks=[tip_rack]
    )

    pipette.pick_up_tip()
    protocol.comment("Hold on to your hats.")
    for _ in range(10):
        pipette.move_to(well_plate_1["A1"].top())
        pipette.move_to(well_plate_2["A1"].top())
    pipette.return_tip()
```

The pipette should move high enough to pass above the waste chute.

Feel free to replace the 96-channel with a different pipette.

Also try commenting out the `protocol.load_waste_chute()` line. For this PR, we're expecting the pipette to be too low to clear the waste chute. (You should probably physically remove the waste chute before trying this.)

# Changelog

When a Python protocol calls `load_waste_chute()`, emit a Protocol Engine `loadAddressableArea` command with `"addressableAreaName": "stagingAreaSlotWithWasteChuteRightAdapter{Covered,NoCover}"`, [per our latest deck definition](https://github.com/Opentrons/opentrons/blob/8db511b60602d8b72eeaeb0725878b927d4430db/shared-data/deck/definitions/4/ot3_standard.json#L488).

I'm deliberately omitting unit tests for all of this because it's going to be replaced soon with code that looks at the *robot's* deck configuration instead of the *protocol's.*

# Review requests

None in particular.

# Risk assessment

Low.


jbleon95 (approved):
These changes look good for a better intermediate way of dodging the waste chute. The GeometryView changes are pretty much in line with how it'll be ultimately done with some changes to what Store we will be referencing

SyntaxColoring:
This is currently blocked behind #13861. This should merge into `edge` after #13861 does.

SyntaxColoring:
This approach is obsolete since we're not merging #13861.

PR #13864, opened 2023-10-27
fix(app): Fix missed i18n key

smb2268:
While looking through https://opentrons.atlassian.net/browse/RQA-1834 I noticed a button that reads `detach_next_pip` - the i18n key needed to be updated to `detach_next_pipette`


koji (approved):
lgtm

Issue #11905, opened 2022-12-19
feat: Unpin requirements


ecederstrand:
### Overview

https://github.com/Opentrons/opentrons/blob/edge/api/setup.py pins a handful of Python package dependencies, e.g. anyio==3.3.0. I cannot decide from the commit history that this is due to specific requirements of opentrons.

Pinning packages makes it difficult to combine opentrons with other packages that also depend on e.g. anyio. Instead, it would be helpful to specify dependencies either open, anyio>=3.3.0, or limited to the next major version (anyio>=3.3.0,<4), assuming dependency maintainers adhere to semantic versioning and don't introduce breaking changes between major versions.

### Implementation details

Dependencies that need more specific versions could be pinned with a commit that references the issue describing the reasoning for pinning to exact versions. Then it's easier to go back later and check if the pinning is still necessary.

SyntaxColoring:
See https://github.com/Opentrons/opentrons/pull/11912#issuecomment-1932990863.

PR #13865, opened 2023-10-27
chore: bump python deps for 3.10 and robot system updates

caila-marashaj:
This PR contains updates required to handle https://github.com/Opentrons/oe-core/pull/110 and https://github.com/Opentrons/buildroot/pull/213 which update the flex operating system and ot-2 operating system respectively to more recent versions of, well everything. Perhaps most significantly to us, both PRs update the python that runs on the machine to 3.10, which requires a lot of changes to a lot of our code.

This PR makes those changes by updating the python versions used in our test harnesses to 3.10; setting those dependencies explicitly captured by the system builders to their new versions from the new versions of the system builders; and bumping other dependencies the _minimum required_ to make it all work. Further work may increase dependency versions further.

The big ones are updating pydantic to 1.9, which is something we've been putting off for a while and probably unlocks further updates. This update is because of the following, which is typical of the process:
- system builders updated numpy, which is locked in because it must be cross-compiled, to 1.17
- numpy 1.17 includes type annotations via .pyi
- mypy must be updated to 0.981 because previous versions interpreted .pyi files as-if the python version was 3.7, but the .pyis had some >3.7-only features
- pydantic must be updated to 1.9 because 1.8 doesn't work with that updated mypy because of a change to internal mypy hooks
- a lot of server code and tests have to be slightly altered because pydantic 1.9 doesn't elide None keys, which is (IMO) good anyway

This should hopefully free us to update mypy to the latest again, and hopefully the changes to explicit handling of none keys will let us update pydantic further.

Anyway, there's a fair amount of "just run the robot" style testing to do here, and we have to decide what to do about loading pickles; handling user-installed dependencies; and further support of 3.7. But all the tests should now pass.


.github/actions/python/setup/action.yaml
```diff
@@ -27,6 +27,6 @@ runs:
     - shell: bash
       run: |
         npm install --global shx@0.3.3
-        $OT_PYTHON -m pip install pipenv==2021.5.29
```
sfoster1, line 30: among other things! note that we'll need to downgrade this to [v2023.10.13](https://github.com/pypa/pipenv/releases/tag/v2023.10.3) if we want to a) keep 3.7 support and b) ensure that by having i guess separate pipfiles (we shouldn't do this, if we want to do something like that we should set up a tox environment just for the api stack)

.github/workflows/api-test-lint-deploy.yaml
```diff
@@ -75,17 +75,13 @@ jobs:
         os: ['windows-2022', 'ubuntu-22.04', 'macos-latest']
         # TODO(mc, 2022-02-24): expand this matrix to 3.8 and 3.9,
         # preferably in a nightly cronjob on edge or something
-        python: ['3.7', '3.10']
+        python: ['3.10']
```
sfoster1, line 78: see above re: keeping 3.7 support and testing for it

.github/workflows/shared-data-test-lint-deploy.yaml
```diff
@@ -71,10 +71,8 @@ jobs:
         os: ['windows-2022', 'ubuntu-22.04', 'macos-latest']
         # TODO(mc, 2022-02-24): expand this matrix to 3.8 and 3.9,
         # preferably in a nightly cronjob on edge or something
-        python: ['3.7', '3.10']
-        exclude:
-          - os: 'macos-latest'
-            python: '3.10'
+        python: ['3.10']
```
sfoster1, line 74: see above re: python 3.7 support

Makefile
```diff
@@ -66,7 +66,7 @@ PYTHON_SETUP_TARGETS := $(addsuffix -py-setup, $(PYTHON_DIRS))
 
 .PHONY: setup-py
 setup-py:
-	$(OT_PYTHON) -m pip install pipenv==2021.5.29
+	$(OT_PYTHON) -m pip install pipenv==2023.11.15
```
sfoster1, line 69: see above re: tested versions and pipenv versions

api/Pipfile
```diff
@@ -7,24 +7,24 @@ name = "pypi"
 # atomicwrites and colorama are pytest dependencies on windows,
 # spec'd here to force lockfile inclusion
 # https://github.com/pypa/pipenv/issues/4408#issuecomment-668324177
-atomicwrites = { version = "==1.4.0", sys_platform = "== 'win32'" }
-colorama = { version = "==0.4.4", sys_platform = "== 'win32'" }
+atomicwrites = { version = "==1.4.0", markers="sys_platform=='win32'" }
```
sfoster1, line 10: one annoying subtle pipenv change was they altered how you spell markers in pipfiles

api/Pipfile
```diff
@@ -7,24 +7,24 @@ name = "pypi"
...
+colorama = { version = "==0.4.4", markers="sys_platform=='win32'" }
 coverage = "==5.1"
-mypy = "==0.910"
+mypy = "==0.981"
```
sfoster1, line 13: needed for numpy support

api/Pipfile
```diff
@@ -7,24 +7,24 @@ name = "pypi"
...
+pytest-asyncio = "~=0.16"
 pytest-cov = "==2.10.1"
 pytest-lazy-fixture = "==0.6.3"
-pytest-xdist = "~=2.2.1"
-sphinx = "==5.0.1"
-twine = "==4.0.2"
-wheel = "==0.30.0"
-typeguard = "==2.12.1"
+pytest-xdist = "~=2.5.0"
+sphinx = "==4.4.0"
```
sfoster1: @ecormany did you bump sphinx recently? if so, i'll bump it again here

api/setup.py
```diff
@@ -61,10 +61,10 @@ def get_version():
 INSTALL_REQUIRES = [
     f"opentrons-shared-data=={VERSION}",
     "aionotify==0.2.0",
-    "anyio==3.3.0",
+    "anyio==3.6.1",
     "jsonschema==3.0.2",
     "numpy>=1.15.1,<2",
-    "pydantic==1.8.2",
+    "pydantic==1.9.2",
```
sfoster1, line 66: needed for mypy support

api/setup.py
```diff
@@ -61,10 +61,10 @@ def get_version():
 INSTALL_REQUIRES = [
     f"opentrons-shared-data=={VERSION}",
     "aionotify==0.2.0",
-    "anyio==3.3.0",
+    "anyio==3.6.1",
```
sfoster1, line 63: matches system builder version

api/src/opentrons/hardware_control/adapters.py
```diff
@@ -54,7 +53,7 @@ def __repr__(self) -> str:
     @staticmethod
     def call_coroutine_sync(
         loop: asyncio.AbstractEventLoop,
-        to_call: WrappedFunc,
+        to_call: Callable[..., Awaitable[WrappedReturn]],
```
sfoster1, line 56: this actually properly typecheckes now which is sick it _might_ forward types appropriately

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -971,8 +971,8 @@ async def set_hold_current(self, axis_currents: OT3AxisMap[float]) -> None:
     @asynccontextmanager
     async def motor_current(
         self,
-        run_currents: OT3AxisMap[float] = {},
-        hold_currents: OT3AxisMap[float] = {},
+        run_currents: Optional[OT3AxisMap[float]] = None,
```
sfoster1, line 974: do not know why mypy didn't used to complain about this!

api/src/opentrons/hardware_control/instruments/ot2/instrument_calibration.py
```diff
@@ -17,6 +17,11 @@
         LabwareDefinition as TypeDictLabwareDef,
     )
 
+# These type aliases aid typechecking in tests that work the same on this and
+# the hardware_control.instruments.ot3 variant
+SourceType = types.SourceType
```
sfoster1, line 22: this was a source of much screaming around explicit re-exports, which we should stop doing

api/src/opentrons/hardware_control/instruments/ot3/instrument_calibration.py
```diff
@@ -23,6 +23,11 @@
 
 PIPETTE_OFFSET_CONSISTENCY_LIMIT: Final = 1.5
 
+# These type aliases aid typechecking in tests that work the same on this and
+# the hardware_control.instruments.ot2 variant
+SourceType = cal_top_types.SourceType
```
sfoster1, line 28: see above

api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -477,7 +477,7 @@ def _edges_from_data(
     # an N-sample rolling average. by inverting the sign of half the kernel, which is
     # why we need it to be even, we do the same thing but while also taking a finite
     # difference.
-    average_difference_kernel = np.concatenate(  # type: ignore
+    average_difference_kernel = np.concatenate(
```
sfoster1, line 480: typechecked numpy... the future is now

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -107,12 +107,12 @@ def __init__(
         ] = trash
         self.requested_as = requested_as
 
-    @property  # type: ignore
+    @property
```
sfoster1, line 110: hooray this works correctly now

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import logging
-from contextlib import nullcontext
+from contextlib import ExitStack
```
sfoster1, line 4: this is for optional publishing, the nullcontext version didn't work right

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1429,12 +1428,15 @@ def move_to(
             # TODO handle publish
             return self
 
-        if publish:
-            publish_ctx = publisher.publish_context(
-                broker=self.broker,
-                command=cmds.move_to(instrument=self, location=location),
-            )
-        with publish_ctx:
+        with ExitStack() as contexts:
```
sfoster1, line 1431: here's the exitstack version. it typechecks more nicely

api/src/opentrons/system/wifi.py
```diff
@@ -15,7 +15,17 @@ class ConfigureArgsError(Exception):
     pass
 
 
-EAP_CONFIG_SHAPE = {
+class EapConfigItem(TypedDict):
```
sfoster1, line 18: it wasn't inferring these right, and didn't like the type ignore that was preventing the issue below

api/tests/opentrons/hardware_control/instruments/test_instrument_calibration.py
```diff
@@ -82,10 +82,10 @@ def test_load_tip_length(
         tipLength=1.23,
         lastModified=datetime(year=2023, month=1, day=1),
         uri=LabwareUri("def456"),
-        source=subject.types.SourceType.factory,
+        source=subject.SourceType.factory,
```
sfoster1, line 85: here's the other side of that explicit reexport thing - we were poking in to the implicit reexports ("reexports", really just the module namespace) of a module because we were genericizing over the module as an object

api/tests/opentrons/protocol_api/core/engine/test_labware_core.py
```diff
@@ -80,6 +80,10 @@ def test_set_calibration(subject: LabwareCore) -> None:
             namespace="hello",
             parameters=LabwareDefinitionParameters.construct(loadName="world"),  # type: ignore[call-arg]
             ordering=[],
+            allowedRoles=[],
```
sfoster1, line 83: first sight of a theme around lots of the robot-server changes: pydantic no longer elides values with none keys, and _does_ elide values with default keys, so we need these to be explicitly set when we build them in python

app-testing/Pipfile
```diff
@@ -4,21 +4,21 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [packages]
-pytest = "==7.2.0"
+pytest = "==7.1.1"
```
sfoster1: @y3rsh you bumped these explicitly, right? i can undo these changes

g-code-testing/Pipfile
```diff
@@ -11,20 +11,20 @@ server-utils = { editable = true, path = "./../server-utils" }
 opentrons-shared-data = { editable = true, path = "../shared-data/python" }
 opentrons_hardware = { editable = true, path = "../hardware" }
 g-code-testing = { editable = true, path = "." }
-anyio = "==3.3.0"
-pydantic = "==1.8.2"
+anyio = "==3.6.1"
```
sfoster1, line 14: matches the api environments

hardware-testing/Pipfile
```diff
@@ -4,25 +4,25 @@ verify_ssl = true
...
 flake8-annotations = "~=2.6.2"
 flake8-docstrings = "~=1.6.0"
 flake8-noqa = "~=1.2.1"
-requests = "==2.26.0"
+requests = "==2.27.1"
 types-requests = "==2.25.6"
 
 [requires]
-python_version = "3.7"
+python_version = "3.10"
```
sfoster1, line 28: may present a problem with SZ

hardware/Pipfile
```diff
@@ -4,28 +4,28 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-python-can = "==3.3.4"
+python-can = "==4.2.2"
```
sfoster1, line 7: required for 3.10 compat and bumped in system builder

hardware/opentrons_hardware/drivers/can_bus/can_messenger.py
```diff
@@ -395,7 +395,6 @@ async def _read_task(self) -> None:
                     log.exception(f"Failed to build from {message}")
             else:
                 log.error(f"Message {message} is not recognized.")
-        raise StopAsyncIteration
```
sfoster1, line 398: there were some test failures around this code and this raise isn't necessary

hardware/opentrons_hardware/drivers/can_bus/driver.py
```diff
@@ -17,25 +16,6 @@
 log = logging.getLogger(__name__)
 
 
-if platform.system() == "Darwin":
```
sfoster1, line 20: we no longer have to deal with this. use can_control on a serial/ssh session.

hardware/opentrons_hardware/firmware_bindings/utils/binary_serializable.py
```diff
@@ -229,8 +229,7 @@ def build(cls, data: bytes) -> BinarySerializable:
                 ),
                 None,
             )
-            # Mypy is not liking constructing the derived types.
-            ret_instance = cls(**args)  # type: ignore[call-arg]
+            ret_instance = cls(**args)
```
sfoster1, line 232: it's so beautiful...

hardware/opentrons_hardware/scripts/can_log_motion_analyzer.py
```diff
@@ -25,7 +25,8 @@
 from opentrons_hardware.hardware_control.constants import interrupts_per_sec
 
 if TYPE_CHECKING:
-    import matplotlib.pyplot as pp  # type: ignore
+    import matplotlib.pyplot as pp  # noqa: F401
```
sfoster1, line 28: typechecked matplotlib!

hardware/tests/opentrons_hardware/drivers/binary_usb/test_driver.py
```diff
@@ -56,6 +56,20 @@ def __init__(
         self,
     ) -> None:
         """Open some serial ports to simulate the usb connection."""
+        # Note: fork will work on linux. it will definitely not work on windows, and probably
```
sfoster1, line 59: this was an unwelcome surprise! this test will no longer work on windows fyi

hardware/tests/opentrons_hardware/drivers/can_bus/test_can_messenger.py
```diff
@@ -106,7 +106,8 @@ async def test_send(
     message: MessageDefinition,
 ) -> None:
     """It should create a can message and use the driver to send the message."""
-    await subject.send(node_id, message)
+    async with subject:
```
sfoster1, line 109: all these tests stopped working for some reason but making them properly use the messenger context manager fixed it

notify-server/Pipfile
```diff
@@ -4,29 +4,27 @@ url = "https://pypi.org/simple/"
 verify_ssl = true
 
 [requires]
-python_version = "3.7"
+python_version = "3.10"
```
sfoster1, line 7: matches the other environments

notify-server/setup.py
```diff
@@ -50,7 +50,7 @@ def get_version():
     f"opentrons=={VERSION}",
     f"opentrons-shared-data=={VERSION}",
     "pyzmq==19.0.2",
-    "pydantic==1.8.2",
+    "pydantic==1.9.2",
```
sfoster1, line 53: matches the other environments

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
 # https://github.com/pypa/pipenv/issues/4408#issuecomment-668324177
-atomicwrites = { version = "==1.4.0", sys_platform = "== 'win32'" }
-colorama = { version = "==0.4.4", sys_platform = "== 'win32'" }
-pytest = "~=6.1"
-pytest-asyncio = "~=0.18"
+atomicwrites = { version = "==1.4.0", markers="sys_platform=='win32'" }
+colorama = { version = "==0.4.4", markers="sys_platform=='win32'" }
+# pytest 7.3.0 dropped attrs, and tavern 1.x at least has some pytest hooks
+# that implicitly require the use of attrs in pytest internals
+pytest = "<7.3.0,>=7.1.0"
```
sfoster1, line 16: pytest needs an upper limit because in 7.3.0 they dropped the use of `attrs`, but tavern actually relies on it in some hooks code

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
+requests = "==2.27.1"
+graphviz = "==0.19"
+mock = "~=4.0.3"
+mypy = "==0.981"
 flake8 = "~=3.9.0"
 flake8-annotations = "~=2.6.2"
 flake8-docstrings = "~=1.6.0"
 flake8-noqa = "~=1.2.1"
-decoy = "~=1.10"
+decoy = "~=2.1"
```
sfoster1, line 30: needed because of mcous/decoy#201 which is now a problem because our hardware controller protocols are based on typing_extensions protocols, but protocols are now in core typing, and there's a bunch of interworking in typing_extensions to make that work which includes overridden __eq__ dunders that now trigger that issue

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
 httpx = "==0.18.*"
 black = "==22.3.0"
 types-requests = "==2.25.6"
 types-mock = "==4.0.1"
 sqlalchemy2-stubs = "==0.0.2a21"
 python-box = "==5.4.1"
+opentrons-hardware = {editable = true, path='../hardware', extras=['FLEX']}
```
sfoster1, line 37: putting all these in dev packages prevents pipenv from failing for some reason. as soon as it started working i committed it so i could stop thinking about it. there may be a better way to do this

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
+notify-server = { editable = true, path = "../notify-server"}
+server-utils = {editable = true, path = "./../server-utils"}
+robot-server = { editable = true, path = "."}
 
 [packages]
-opentrons = { editable = true, path = "../api" }
-opentrons-shared-data = { editable = true, path = "../shared-data/python" }
-notify-server = { editable = true, path = "../notify-server" }
-anyio = "==3.3.0"
+anyio = "==3.6.1"
```
sfoster1, line 45: system builder version

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
 fastapi = "==0.68.1"
 python-dotenv = "==0.19.0"
 python-multipart = "==0.0.5"
-pydantic = "==1.8.2"
+pydantic = "==1.9.2"
```
sfoster1, line 49: required to work with more recent mypy

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
-requests = "==2.26.0"
-tavern = "~=1.6"
-graphviz = "==0.17"
-mock = "~=4.0.2"
-mypy = "==0.910"
+pytest-xdist = "~=2.5.0"
+requests = "==2.27.1"
+graphviz = "==0.19"
+mock = "~=4.0.3"
+mypy = "==0.981"
```
sfoster1, line 25: required to work with the version of numpy the system builders now use

robot-server/Pipfile
```diff
@@ -4,57 +4,56 @@ verify_ssl = true
...
-idna = "==3.2"
-click = "==8.0.1"
-# TODO(mc, 2022-04-04): this dep is newer than the OT-2 for dev environment reasons
-# remove this TODO when Opentrons/buildroot has updated to newer numpy
-# https://github.com/Opentrons/buildroot/pull/148
-numpy = "==1.21.2"
+systemd-python = { version = "==234", markers="sys_platform == 'linux'" }
+idna = "==3.3"
+click = "==8.1.2"
+numpy = "==1.23.3"
```
sfoster1, line 56: from the system builders

robot-server/robot_server/maintenance_runs/router/commands_router.py
```diff
@@ -182,7 +182,7 @@ async def create_run_command(
     if waitUntilComplete:
         timeout_sec = None if timeout is None else timeout / 1000.0
         with move_on_after(timeout_sec):
-            await protocol_engine.wait_for_command(command.id),
+            await protocol_engine.wait_for_command(command.id)
```
sfoster1, line 185: these trailing commas here and in other places `wait_for_command` was called were causing Significant Linting Problems for reasons I don't understand, but also either these are doing nothing or they are breaking things so removing them seems like the right call

robot-server/robot_server/protocols/analysis_models.py
```diff
@@ -82,7 +82,7 @@ class CompletedAnalysis(BaseModel):
     # Fields that are currently unique to robot-server, missing from local analysis:
     id: str = Field(..., description="Unique identifier of this analysis resource")
     status: Literal[AnalysisStatus.COMPLETED] = Field(
-        AnalysisStatus.COMPLETED,
+        ...,
```
sfoster1, line 85: a downstream effect of setting `ignore_unset` - with this defaulted, most `CompletedAnalyses` would not have a `status` key. removing it fixes that by making all the callers specify it

robot-server/robot_server/protocols/completed_analysis_store.py
```diff
@@ -301,5 +301,5 @@ def _serialize_completed_analysis_to_json(completed_analysis: CompletedAnalysis)
         # FastAPI + Pydantic + our customizations serialize these objects
         # over the `GET /protocols/:id/analyses/:id` endpoint.
         by_alias=True,
-        exclude_none=True,
+        exclude_none=False,
```
sfoster1: formalize this setting, since it wasn't actually happening even when it was set to True and this is less confusing

robot-server/robot_server/service/legacy/models/settings.py
```diff
@@ -170,10 +170,10 @@ class BasePipetteSettingFields(BaseModel):
 # A dynamic model of the possible fields in pipette configuration. It's
 # generated from pipette_config module. It's derived from an object with the
 # 'quirks` member.
-PipetteSettingsFields = create_model(
+PipetteSettingsFields = create_model(  # type: ignore[call-overload]
```
sfoster1, line 173: same effect, just new mypy checks it in a different place

robot-server/robot_server/subsystems/firmware_update_manager.py
```diff
@@ -148,9 +148,7 @@ def __init__(
         self._status_cache_lock = Lock()
         self._created_at = created_at
         self._update_id = update_id
-        # mypy thinks this is assignment to a method, which it is not. that means
-        # it's ok to assign this and it's also okay to not take a self argument
-        self._complete_callback = complete_callback  # type: ignore[assignment,misc]
+        self._complete_callback = complete_callback
```
sfoster1, line 151: mypy can handle function properties now yay

robot-server/robot_server/system/time_utils.py
```diff
@@ -44,7 +44,6 @@ async def _time_status() -> Dict[str, Union[str, bool]]:
         "timedatectl show",
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE,
-        loop=asyncio.get_event_loop(),
```
sfoster1, line 47: many things now do not let you pass a loop arg. this may present... challenges... for 3.7 compat if we want that

robot-server/setup.py
```diff
@@ -56,11 +56,11 @@ def get_version():
     f"opentrons-shared-data=={VERSION}",
     f"notify-server=={VERSION}",
     f"server-utils=={VERSION}",
-    "anyio==3.3.0",
+    "anyio==3.6.1",
```
sfoster1, line 59: from system builders

robot-server/setup.py
```diff
@@ -56,11 +56,11 @@ def get_version():
...
     "fastapi==0.68.1",
     "python-dotenv==0.19.0",
     "python-multipart==0.0.5",
-    "pydantic==1.8.2",
+    "pydantic==1.9.2",
```
sfoster1, line 63: to work with recent mypy

robot-server/tests/integration/http_api/protocols/test_deck_coordinate_load.tavern.yaml
```diff
@@ -20,11 +20,12 @@ stages:
       json:
```
sfoster1: all these test changes were about handling keys that are now present even if they're null. so many keys

robot-server/tests/service/legacy/routers/test_settings.py
```diff
@@ -119,7 +119,9 @@ def mock_pipette_data():
 def mock_known_pipettes(decoy: Decoy) -> Decoy:
     with patch(
         "opentrons_shared_data.pipette.mutable_configurations.known_pipettes",
-        new=decoy.mock(),
+        new=decoy.mock(
```
sfoster1, line 122: this being required is why decoy is 2.x now

server-utils/Pipfile
```diff
@@ -4,22 +4,22 @@ verify_ssl = true
 name = "pypi"
 
 [requires]
-python_version = "3.7"
+python_version = "3.10"
```
sfoster1, line 7: matches other packages

server-utils/setup.py
```diff
@@ -51,11 +51,11 @@ def get_version():
 DESCRIPTION = "Common utilities for various Opentrons Python servers."
 PACKAGES = find_packages(where=".", exclude=["tests.*", "tests"])
 INSTALL_REQUIRES = [
-    "anyio==3.3.0",
+    "anyio==3.6.1",
```
sfoster1, line 54: matches other packages

shared-data/python/Pipfile
```diff
@@ -4,23 +4,23 @@ verify_ssl = true
 name = "pypi"
 
 [dev-packages]
-mypy = "==0.800"
+mypy = "==0.981"
```
sfoster1, line 7: matches other packages

shared-data/python/opentrons_shared_data/labware/labware_definition.py
```diff
@@ -326,3 +326,11 @@ class LabwareDefinition(BaseModel):
         default_factory=None,
         description="Force, in Newtons, with which the gripper should grip the labware.",
     )
+
+    def dict(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
```
sfoster1: PRETTY IMPORTANT: the whole thing with exclude_unset and exclude_none - there's some things in labware definitions that are schema'd in json as properties that are non-nullable but do not appear in `required_properties`. these were and are mapped in pydantic as `Optional[...] = None` attributes. This is not the right way to model this because you should either not have the key, or have a value that's not null/none. 

It worked before because _for these keys_, being present with a value of `None` in a pydantic model instance is exactly equivalent to being not-present in json, and so `exclude_none` does exactly what you want. but it doesn't do what you want if you have something where being absent and being present with null are different things.

A solution to this, which we do here, is to make it explicit that if the value _was not set when the model was created_ then it should not be emitted. this still isn't safe because pydantic will let you (invalidly, per the jsonschema) specify the value as null when creating the model, and happily deserialize json where the value is present and null, but it means that you don't output nulls anymore so that's good at least.

The proper fix for this is probably to change the way the models work and the way the schemas work to have jsonschema `$oneOf` elements and unions of pydantic models rather than non-required properties and optionals. I'd like to do that in a different PR - i'm pretty comfortable we'll detect the correctness in tests, since the test that caught this in the first place was about roundtripping labware defs - but we can do it now if you really want. I don't consider this any _more_ broken than it used to be.

system-server/Pipfile
```diff
@@ -6,43 +6,40 @@ name = "pypi"
 [packages]
 fastapi = "==0.68.1"
 uvicorn = "==0.14.0"
-anyio = "==3.3.0"
+anyio = "==3.6.1"
```
sfoster1, line 9: from system builder

system-server/Pipfile
```diff
@@ -6,43 +6,40 @@ name = "pypi"
...
 typing-extensions = ">=4.0.0,<5"
-pyjwt = "==2.6.0"
 python-dotenv = "==0.19.0"
 python-multipart = "==0.0.5"
-pydantic = "==1.8.2"
-sqlalchemy = "==1.4.32"
-systemd-python = { version = "==234", sys_platform = "== 'linux'" }
+pydantic = "==1.9.2"
```
sfoster1, line 13: for mypy

update-server/otupdate/common/name_management/__init__.py
```diff
@@ -75,9 +75,7 @@ async def set_name_endpoint(request: web.Request) -> web.Response:
     """
 
     def build_400(msg: str) -> web.Response:
-        return web.json_response(  # type: ignore[no-untyped-call,no-any-return]
-            data={"message": msg}, status=400
-        )
+        return web.json_response(data={"message": msg}, status=400)
```
sfoster1, line 78: typed aiohttp? the future is now!

update-server/otupdate/common/update.py
```diff
@@ -97,6 +97,8 @@ async def status(request: web.Request, session: UpdateSession) -> web.Response:
 async def _save_file(part: BodyPartReader, path: str) -> None:
```
sfoster1, line 97: all this stuff wasn't type safe at all and now it should be

app-testing/Pipfile
y3rsh: We need not touch app-testing at all.  It does not plug into global make and does not need to.
In the PR for the analyses snapshots, that I will put up in January, I have everything upgraded and pinned using python 3.12 😊

api/Pipfile
ecormany: We've been sitting at 5.0.1. Far from the latest, but gets the job done.

robot-server/robot_server/protocols/analysis_models.py
SyntaxColoring: Hmm, does the same problem happen with Protocol Engine `commandType`s?

shared-data/python/opentrons_shared_data/labware/labware_definition.py
SyntaxColoring: **I think we should merge this as you have it and sort it out in follow-up PRs,** but I'm skeptical of this.

We had a general solution for this. The solution was `exclude_none` (instead of `exclude_unset`), as described [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/1086030289/Pydantic#typing.Optional-and-Pydantic%E2%80%99s-inconsistent-treatment-of-None) and implemented, among other places, [here](https://github.com/Opentrons/opentrons/blob/024984a9db2d79656e93cb8e1aceefeb5576a2b8/robot-server/robot_server/service/json_api/response.py#L39). If that solution has stopped working with a Pydantic upgrade, we should at least figure out why.

Also, `exclude_unset` doesn't seem like the theoretically better solution; it seems roughly equally quasi-correct as `exclude_none`, just in different ways. You're gaining the ability to distinguish between JSON-`null` and JSON-omitted, but in exchange, other basic expectations, like `a.field = b.field` being a safe and lossless thing to do, are no longer true.

> [`exclude_none`] doesn't do what you want if you have something where being absent and being present with null are different things.

My experience is that, in general, Pydantic is not appropriate for implementing APIs that deal with JSON-`null` *and* JSON-omitted *and* treat them as semantically distinct from each other. So the workaround has been to just not have APIs like that.

So in other words, I worry that switching from `exclude_none` to `exclude_unset` is going to give us the ability to do something that we've already given up on, at the cost of introducing additional gotchas.

> i'm pretty comfortable we'll detect the correctness in tests, since the test that caught this in the first place was about roundtripping labware def

[This one](https://github.com/Opentrons/opentrons/blob/0c2484261920a074ba06c5a0698df94ae8a04bcf/robot-server/tests/integration/sessions/test_analysis_round_trip.tavern.yaml)? I'm delighted that that was useful. But case in point: for the bug that originally prompted that test, there was a 1.5 month lag between [the breaking change](https://github.com/Opentrons/opentrons/pull/13186) and [us noticing](https://opentrons.atlassian.net/browse/RSS-349). Maybe this is pessimistic, but model bugs can be subtle, so I don't want us to get complacent and overestimate our ability to deal with more Pydantic gotchas.

> The proper fix for this is probably to change the way the models work and the way the schemas work to have jsonschema `$oneOf` elements and unions of pydantic models rather than non-required properties and optionals.

Maybe. I'd rather put that effort into switching to a better JSON library, like msgspec.

SyntaxColoring (commented):
Hallelujah.

Here are some preliminary comments and questions. I'll take a closer look when we return in January, but I don't anticipate anything that would block this from merging.

shared-data/python/opentrons_shared_data/labware/labware_definition.py
sfoster1: > **I think we should merge this as you have it and sort it out in follow-up PRs,** but I'm skeptical of this.

I think if we want to go back to `exclude_none` we should do it here so we don't have to redo a bunch of tavern work.

> We had a general solution for this. The solution was `exclude_none` (instead of `exclude_unset`), as described [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/1086030289/Pydantic#typing.Optional-and-Pydantic%E2%80%99s-inconsistent-treatment-of-None) and implemented, among other places, [here](https://github.com/Opentrons/opentrons/blob/024984a9db2d79656e93cb8e1aceefeb5576a2b8/robot-server/robot_server/service/json_api/response.py#L39). If that solution has stopped working with a Pydantic upgrade, we should at least figure out why.
> 
> Also, `exclude_unset` doesn't seem like the theoretically better solution; it seems roughly equally quasi-correct as `exclude_none`, just in different ways. You're gaining the ability to distinguish between JSON-`null` and JSON-omitted, but in exchange, other basic expectations, like `a.field = b.field` being a safe and lossless thing to do, are no longer true.

True. Don't like that at all.

> > [`exclude_none`] doesn't do what you want if you have something where being absent and being present with null are different things.
> 
> My experience is that, in general, Pydantic is not appropriate for implementing APIs that deal with JSON-`null` _and_ JSON-omitted _and_ treat them as semantically distinct from each other. So the workaround has been to just not have APIs like that.
> 
> So in other words, I worry that switching from `exclude_none` to `exclude_unset` is going to give us the ability to do something that we've already given up on, at the cost of introducing additional gotchas.

I mean maybe now we can un-give-up on it. If a middleware categorically excludes some kind of useful distinction we should figure out a way around that limitation.

> > The proper fix for this is probably to change the way the models work and the way the schemas work to have jsonschema `$oneOf` elements and unions of pydantic models rather than non-required properties and optionals.
>
> Maybe. I'd rather put that effort into switching to a better JSON library, like msgspec.

I think we should definitely switch. I also think that the way we specify these details in the jsonschema is actually wrong because `gripForce` and `gripperHeightFromBottom` are independent and I don't think they should be. We can write the schema in a way that doesn't change valid definitions with optional subschemas pretty easily and the pydantic version of that is a union of models.

I really really hate `exclude_none` because I do not think it's a safe way to represent data but it's probably better to avoid changing it if we can. I think it's a fastapi thing that breaks it, specifically [this](https://github.com/tiangolo/fastapi/issues/3314#issuecomment-1003083891) though I'm not sure why it would have changed.

robot-server/robot_server/protocols/analysis_models.py
sfoster1: No, which is weird now that I think of it

shared-data/python/opentrons_shared_data/labware/labware_definition.py
SyntaxColoring: > I think if we want to go back to exclude_none we should do it here so we don't have to redo a bunch of tavern work.

OK, I'll take a look and see what it would take to get that to keep working.

> I mean maybe now we can un-give-up on [distinguishing null from omitted]. If a middleware categorically excludes some kind of useful distinction we should figure out a way around that limitation.

That's a good point, yeah.

> I think it's a fastapi thing that breaks it, specifically [this](https://github.com/tiangolo/fastapi/issues/3314#issuecomment-1003083891) though I'm not sure why it would have changed.

Wow, great find. I bet that is it. I'll investigate.

robot-server/tests/integration/http_api/test_deck_configuration.tavern.yaml
```diff
@@ -5,21 +5,25 @@ marks:
...
       json:
         data:
-          # lastModifiedAt is deliberately omitted from this expected object.
-          # A lastModifiedAt that's omitted or null means the deck configuration has never been set.
-          #
-          # Unfortunately, this makes this test order-dependent with any other tests
-          # that modify the deck configuration, even if they try to restore the original value
-          # after they're done. We probably need some kind of deck configuration factory-reset.
-          #
-          # lastModifiedAt: null
```
SyntaxColoring: Pending the discussion in https://github.com/Opentrons/opentrons/pull/13865#discussion_r1433161062, if we go with the approach as-written, and we do want to update all of these Tavern tests to account for additional `null`s:

Instead of disabling strict mode for this whole file, let's do explicit `null` assertions for `lastModifiedAt` and `links`. And as far as I can tell, the `headers:off` isn't doing anything.

Here's a commit you can cherry-pick: https://github.com/Opentrons/opentrons/commit/afa83468f14bf824f6d98026ab7d8c17f90f77bc. `git fetch && git cherry-pick afa83468f14bf824f6d98026ab7d8c17f90f77bc`

shared-data/python/opentrons_shared_data/labware/labware_definition.py
SyntaxColoring: This breakage turned out to be from https://github.com/pydantic/pydantic/pull/2650.

We were overriding `BaseModel.dict()` to always act as if `exclude_none=True`. That used to be enough, but now, we also need to override `BaseModel.json()` in the same way.

I am coming around to your idea that `exclude_unset` is a lesser evil than `exclude_none`. We can think about swapping everything over to that in separate PRs.

robot-server/tests/integration/http_api/test_deck_configuration.tavern.yaml
SyntaxColoring: Moot.

y3rsh:
Are we addressing #11905 request here?

robot-server/robot_server/protocols/analysis_models.py
SyntaxColoring: I think this change is no longer required given the discussion in https://github.com/Opentrons/opentrons/pull/13865#discussion_r1433161062, but it's a good change for other reasons, so let's keep it. I think this makes it so that `status` is marked as a `required` field in the OpenAPI spec, which is correct.

SyntaxColoring (approved):
I've reviewed this in detail and it looks great to merge whenever you're ready with https://github.com/Opentrons/oe-core/pull/110 and https://github.com/Opentrons/buildroot/pull/213. Thank you so much, @caila-marashaj and @sfoster1, for pulling this off!

sfoster1:
> Are we addressing #11905 request here?

No, but I want to do the following:
- Merge Opentrons/oe-core#117 and Opentrons/buildroot#214 to allow fully separated internal release builds
- Merge this PR, Opentrons/oe-core#110, Opentrons/buildroot#213 ; these have updates to essentially the minimal new versions of everything 
- tag an internal release and spend a bit baselining and fixing anything that has gone wrong
- start systematically updating versions of deps as far as possible/as we want and refactor some of our dependency management by doing the following:
  - weaken library dependency requirements (i.e. the linked issue)
  - centralize pipfiles to the application entrypoints (robot-server, update-server, ot3usb) and refactor dev side linting and testing to use those centralized pipfiles
  - flesh out xplatform and python version tests for libraries and limit tests for applications

PR #13867, opened 2023-10-28
feat(hardware): handle motor driver errors

pmoegenburg:
This PR enables the reporting of Flex motor driver errors, adding the MotorDriverError exception. Specific motor driver errors are over-temperature, short circuit, and open circuit. An ErrorMessage is reported from the Flex firmware in two cases:
1. As a response to a move request when a motor driver is in error state. In this case, the error is logged.
2. If a motor driver enters error state during a move. In this case, a subsequent ReadMotorDriverErrorStatusResponse message is reported from the Flex firmware, which details the specific error. This PR logs these messages and also raises these messages as MotorDriverError exceptions.

# Overview

# Test Plan

- [ ] Test both use cases on robot. In second use case, ensure ReadMotorDriverErrorStatusResponse message is associated with requesting MoveID.
- [x] Write software tests for both use cases. In second use case, ensure each specific error is parsed.

# Changelog

# Review requests

# Risk assessment

Low. Adds MotorDriverError to existing list of exceptions. Introduces handling and parsing of ReadMotorDriverErrorStatusResponse message, using existing well-tested frameworks and concepts.


sfoster1 (approved):
This is looking really good! Let's just fix up those last lint and test problems. Good call on the error code value and location!

PR #13870, opened 2023-10-30
fix(protocol-designer): grab defUri from labwareDefinitions key

jerader:
addresses RESC-161 https://opentrons.atlassian.net/browse/RESC-161

# Overview

There was a bug reported from the Support team where a Protocol made from version 6.0.0 was imported into 7.0.0 and then reimported into 7.0.0 and white screened. This is because there was a mistake in `loadLabware` in the 7.0.0 migration where the `loadName` was accidentally set to the `definitionURI` And then resulted in very brittle splitting logic in the `labwareInvariantProperties` reducer to grab the definition URI either from the <v7 `loadName` or from splitting the =v7 `labwareId`.

# Test Plan

Upload the `Created 8 September` protocol attached to the ticket. See that it imports correctly. Export it and then change the software version to "7.1.0" and reimport that exported protocol. It should work as expected.

# Changelog

- add to te 7.1.0 migration to change the loadName param in `loadLabware` commands to the true `loadName` instead of the `definitionId` (which is actually the definition uri but in schema 6 the top level labware key refers to it was "definitionId".).
- change the logic in the reducer to not branch and convert that loadname into the definition uri. 

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/load-file/migration/7_1_0.ts
+++ b/protocol-designer/src/load-file/migration/7_1_0.ts
@@ -24,7 +24,7 @@ interface LabwareLocationUpdate {
 export const migrateFile = (
   appData: ProtocolFile<DesignerApplicationData>
 ): ProtocolFile => {
-  const { designerApplication, robot, commands } = appData
+  const { designerApplication, robot, commands, labwareDefinitions } = appData
   const labwareLocationUpdate: LabwareLocationUpdate =
     designerApplication?.data?.savedStepForms[INITIAL_DECK_SETUP_STEP_ID]
       .labwareLocationUpdate
@@ -68,23 +68,29 @@ export const migrateFile = (
     savedStepForms: Record<string, any>
   ): Record<string, any> => {
     return mapValues(savedStepForms, stepForm => {
+      const sharedParams = {
+        blowout_location:
+          stepForm.blowout_location === 'fixedTrash'
+            ? trashId
+            : stepForm.blowout_location,
+        dropTip_location: trashId,
+      }
+
       if (stepForm.stepType === 'moveLiquid') {
         return {
           ...stepForm,
-          blowout_location:
-            stepForm.blowout_location === 'fixedTrash'
+          dispense_labware:
+            stepForm.dispense_labware === 'fixedTrash'
               ? trashId
-              : stepForm.blowout_location,
-          dropTip_location: trashId,
+              : stepForm.dispense_labware,
+          ...sharedParams,
         }
       } else if (stepForm.stepType === 'mix') {
         return {
           ...stepForm,
-          blowout_location:
-            stepForm.blowout_location === 'fixedTrash'
-              ? trashId
-              : stepForm.blowout_location,
-          dropTip_location: trashId,
+          labware:
+            stepForm.labware === 'fixedTrash' ? trashId : stepForm.labware,
+          ...sharedParams,
         }
       }
 
@@ -100,6 +106,22 @@ export const migrateFile = (
   const newFilteredSavedStepForms = migrateSavedStepForms(
     filteredSavedStepForms
   )
+
+  const loadLabwareCommands: LoadLabwareCreateCommand[] = commands
+    .filter(
+      (command): command is LoadLabwareCreateCommand =>
+        command.commandType === 'loadLabware'
+    )
+    .map(command => {
+      const { parameters } = labwareDefinitions[command.params.loadName]
+      return {
+        ...command,
+        params: {
+          ...command.params,
+          loadName: parameters.loadName,
+        },
+      }
+    })
 
   return {
     ...appData,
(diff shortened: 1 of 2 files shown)


protocol-designer/src/load-file/migration/7_1_0.ts
```diff
@@ -122,7 +122,7 @@ export const migrateFile = (
       },
     },
     labwareDefinitions: {
-      ...{ [trashId]: trashDefinition },
+      ...{ [trashDefUri]: trashDefinition },
```
jerader, line 160: accidentally used `id` instead of `defUri` for the key here

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1154,12 +1154,23 @@ export const labwareInvariantProperties: Reducer<
...
+            const labwareDefintionForV7 = Object.entries(
+              file.labwareDefinitions
+            ).find(([key, value]) => value.parameters.loadName === loadName)
+            let defUri: string = ''
+            if (labwareDefintionForV7 != null) {
+              const [key] = labwareDefintionForV7
+              defUri = key
+            } else {
+              defUri = loadName
+            }
```
jerader, line 1166: if you import a protocol from v6 or lower, the `loadName` === `defUri`. But if you import a protocol from v7, the `loadName` === `labwareDefinition`'s `loadName`. That is why this logic is needed.

protocol-designer/src/load-file/migration/7_1_0.ts
```diff
@@ -68,23 +68,29 @@ export const migrateFile = (
...
         return {
           ...stepForm,
-          blowout_location:
-            stepForm.blowout_location === 'fixedTrash'
-              ? trashId
-              : stepForm.blowout_location,
-          dropTip_location: trashId,
+          labware:
+            stepForm.labware === 'fixedTrash' ? trashId : stepForm.labware,
+          ...sharedParams,
```
jerader, line 93: This is not related to the fix. This change is because when I imported the "created september 2022" protocol, lots of the transfer commands weren't being imported properly. It is because I forgot to migrate the `dispense_labware` and `labware` params in `moveLiquid` and `mix` from `fixedTrash` to the newly created trash entity's id. LMK if I should pull this out into its own PR though!

protocol-designer/src/load-file/migration/7_1_0.ts
```diff
@@ -101,6 +107,22 @@ export const migrateFile = (
     filteredSavedStepForms
   )
 
+  const loadLabwareCommands: LoadLabwareCreateCommand[] = commands
+    .filter(
+      (command): command is LoadLabwareCreateCommand =>
+        command.commandType === 'loadLabware'
+    )
+    .map(command => {
+      const { parameters } = labwareDefinitions[command.params.loadName]
```
jerader, line 116: this works because the loadName in the 7.0.0 migration is set to the definition URI.

protocol-designer/src/load-file/migration/7_1_0.ts
```diff
@@ -68,23 +68,33 @@ export const migrateFile = (
...
         return {
           ...stepForm,
-          blowout_location:
-            stepForm.blowout_location === 'fixedTrash'
-              ? trashId
-              : stepForm.blowout_location,
-          dropTip_location: trashId,
+          labware:
+            stepForm.labware === 'fixedTrash' ? trashId : stepForm.labware,
+          ...sharedParams,
```
jerader, line 97: This is not related to the fix. This change is because when I imported the "created september 2022" protocol, lots of the transfer commands weren't being imported properly. It is because I forgot to migrate the `dispense_labware`, `aspirate_labware`, and `labware` params in `moveLiquid` and `mix` from `fixedTrash` to the newly created trash entity's id (As part of the movable trash work). LMK if I should pull this out into its own PR though!

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1154,12 +1154,19 @@ export const labwareInvariantProperties: Reducer<
         ...loadLabwareCommands.reduce(
           (acc: NormalizedLabwareById, command: LoadLabwareCreateCommand) => {
             const { labwareId, loadName } = command.params
-            const defUri = labwareId?.split(':')[1]
+            const labwareDefinitionMatch = Object.entries(
+              file.labwareDefinitions
+            ).find(
+              ([definitionUri, labwareDef]) =>
+                labwareDef.parameters.loadName === loadName
```
b-cooper: can we match on the def uri here instead? That way we account for version and namespace differences

jerader: how would we get the def uri though? I thought there is no other area in a file that has the defUri by itself except for the `labwareDefinitions` keys

b-cooper: can we match on the def uri here instead? That way we account for version and namespace differences

b-cooper: You can generate the def URI using the `getLabwareDefURI` utility function. it takes the definition in and concatenates the loadname, namespace and version

b-cooper (approved):
if we can match on defURI, this looks great

protocol-designer/src/step-forms/reducers/index.ts
jerader: cool thanks, ya i can do that!

PR #13871, opened 2023-10-30
fix(api): max discontinuity speed for gripper Z was too high

ahiuchingau:
We've been seeing the gripper z stage drops unexpectedly during protocol cleanup phase in ABR testing. This happens  when the z stage is very close to the limit switch. It is because the motor is homing at a speed that's so fast that the firmware doesn't have enough time to register the limit switch has been triggered, hence not being to stop the motor in time.
 
I've tested this PR on two ABR robots that were seeing this issue - and the problem has disappeared after the fix.


sfoster1 (approved):
Nice, thanks!

PR #13872, opened 2023-10-30
refactor(shared-data): Make cutout IDs more obviously different from addressable area IDs

SyntaxColoring:
# Overview

Our latest thinking in deck definitions is that *cutouts* can have *cutout fixtures* mounted onto them, and those cutout fixtures provide *addressable areas* to protocols. Cutouts, cutout fixtures, and addressable areas are all separate things.

Our cutouts happen to have the same `id` strings as some of our addressable areas, like `"C2"`. This is confusing because it baits us into conflating the two concepts, or assuming a link where one does not necessarily exist. So this makes the `id` strings different, so the distinction is more obvious.

This goes towards RSS-333 in general.

# Test Plan

I don't think we've built anything so far that is actually looking at cutout `id`s and would actually be in danger of breaking, have we?

If that's correct, we're good with just CI testing.

# Changelog

* Rename cutout `id`s like `"C2"` to be like `"cutoutC2"`.
* Leave the addressable area `id`s like `"C2"` alone, so they will continue to generally match what protocols write.
* Rewrite a test that assumed, for example, that slot C2 was defined by addressable area C2 and cutout C2. It now traverses the deck definition's `providesAddressableAreas` fields to avoid assuming anything about the cutout IDs.

# Review requests

* Double-check my work in the JSON files to make sure I didn't miss anywhere.
* This will surface problems in any code that assumes anything about cutout fixture IDs. For example, if we improperly have something like `CutoutFixtureID = Literal["A1", "A2", ...]` anywhere, that will break. Are you aware of any code that does that?

# Risk assessment

Low.

### DIFF
--- a/shared-data/deck/definitions/4/ot2_short_trash.json
+++ b/shared-data/deck/definitions/4/ot2_short_trash.json
@@ -216,73 +216,73 @@
     ],
     "cutouts": [
       {
-        "id": "1",
+        "id": "cutout1",
         "position": [0.0, 0.0, 0.0],
         "displayName": "Cutout 1",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "2",
+        "id": "cutout2",
         "position": [132.5, 0.0, 0.0],
         "displayName": "Cutout 2",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "3",
+        "id": "cutout3",
         "position": [265.0, 0.0, 0.0],
         "displayName": "Cutout 3",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "4",
+        "id": "cutout4",
         "position": [0.0, 90.5, 0.0],
         "displayName": "Cutout 4",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "5",
+        "id": "cutout5",
         "position": [132.5, 90.5, 0.0],
         "displayName": "Cutout 5",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "6",
+        "id": "cutout6",
         "position": [265.0, 90.5, 0.0],
         "displayName": "Cutout 6",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "7",
+        "id": "cutout7",
         "position": [0.0, 181.0, 0.0],
         "displayName": "Cutout 7",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "8",
+        "id": "cutout8",
         "position": [132.5, 181.0, 0.0],
         "displayName": "Cutout 8",
         "compatibleFixtureTypes": ["singleStandardSlot"]
       },
       {
-        "id": "9",
+        "id": "cutout9",
         "position": [265.0, 181.0, 0.0],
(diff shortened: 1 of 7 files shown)


SyntaxColoring:
This might conflict with #13854's introduction of "legacy fixtures" to the v4 schema?

sfoster1 (dismissed):
- api-client tests: `api-client/src/protocols/__tests__/utils.test.ts` identify cutouts explicitly as constants
- deck configuration code in `app/src/pages/Protocols/hooks/index.ts`, `app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx`, `app/src/organisms/ProtocolSetupDeckConfiguration/index.tsx`, `app/src/organisms/ProtocolDetails/index.tsx`, `app/src/organisms/ProtocolSetupModulesAndDeck/FixtureTable.tsx` and various corresponding tests do this also
- ts types in `shared-data/js/types.ts` has an explicit list of cutout ids

brenthagen:
> * api-client tests: `api-client/src/protocols/__tests__/utils.test.ts` identify cutouts explicitly as constants
> * deck configuration code in `app/src/pages/Protocols/hooks/index.ts`, `app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx`, `app/src/organisms/ProtocolSetupDeckConfiguration/index.tsx`, `app/src/organisms/ProtocolDetails/index.tsx`, `app/src/organisms/ProtocolSetupModulesAndDeck/FixtureTable.tsx` and various corresponding tests do this also
> * ts types in `shared-data/js/types.ts` has an explicit list of cutout ids

@sfoster1 @SyntaxColoring I can make these changes in RAUT-781 - app isn't using deck definition v4 yet

SyntaxColoring:
@sfoster1 Thanks, nice catches.

@brenthagen Yeah, thanks, I think I'll leave this to you, or we can pair on it. I'm looking at these casts and I'm not sure how to deal with them on my own:

https://github.com/Opentrons/opentrons/blob/346245f891be94d5392e3d169d65aa1e397ff671/app/src/organisms/Devices/ProtocolRun/SetupModuleAndDeck/SetupModulesList.tsx#L355

https://github.com/Opentrons/opentrons/blob/346245f891be94d5392e3d169d65aa1e397ff671/protocol-designer/src/file-data/selectors/fileCreator.ts#L309

https://github.com/Opentrons/opentrons/blob/346245f891be94d5392e3d169d65aa1e397ff671/protocol-designer/src/components/DeckSetup/index.tsx#L498

https://github.com/Opentrons/opentrons/blob/346245f891be94d5392e3d169d65aa1e397ff671/protocol-designer/src/components/DeckSetup/index.tsx#L535

https://github.com/Opentrons/opentrons/blob/346245f891be94d5392e3d169d65aa1e397ff671/protocol-designer/src/components/modules/StagingAreasModal.tsx#L65

Just glancing at those, it looks like, if we're given a slot like `A3`, we sometimes want to find the cutout underneath that slot, like `cutoutA3`. That's totally possible and there are a few ways to do it, but I'm a bit scared of messing with these things myself without knowing the context.

SyntaxColoring:
Spoke with @brenthagen in-person and we agreed to deal with the TypeScript stuff described above in a separate PR.

brenthagen (approved):
change makes sense to me, app & ui is equipped to adjust the typescript as needed as part of ongoing migration to deck definition v4

SyntaxColoring:
Got verbal approval from @jbleon95 and various 👍s in Slack.

PR #13873, opened 2023-10-30
refactor(shared-data): Delete `compatibleFixtureTypes` from schema v4

SyntaxColoring:
# Overview

Right now, every cutout has a list of `compatibleFixtureTypes`. That seems redundant. Every `cutoutFixture` defines the cutouts that it's allowed to mount to (via the `mayMountTo` field). So, there's no reason for every cutout to also define the `cutoutFixtures` that are allowed to mount to it.

This field used to link with the `fixtureType` field, but that was removed in #13785.

# Test Plan

Just CI.

# Review requests

Is this architecturally correct, or am I missing a use case for these fields?

# Risk assessment

Low. As far as I can tell, nothing was using these.

### DIFF
--- a/shared-data/deck/definitions/4/ot2_short_trash.json
+++ b/shared-data/deck/definitions/4/ot2_short_trash.json
@@ -218,74 +218,62 @@
       {
         "id": "1",
         "position": [0.0, 0.0, 0.0],
-        "displayName": "Cutout 1",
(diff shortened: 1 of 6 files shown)


sfoster1 (approved):
Makes sense, that's just one more thing we'd have to change

SyntaxColoring:
Got a verbal 👍 from @jbleon95.

PR #13875, opened 2023-10-30
refactor(app): replace RobotWorkSpace in DeckThumbnail with BaseDeck

koji:
# Overview
replace RobotWorkSpace with BaseDeck
- ChooseProtocolSlideout
- ProtocolCard
- ProtocolDetails
- ProtocolDetails for ODD

For LabwareInfoOverlay, the component requires `runId` since `LabwareInfo` uses `runId` for `useLabwareOffsetForLabware`.
`useLabwareOffsetForLabware` uses `runId` to get `mostRecentAnalysis`.

close RAUT-819

# Test Plan
open the above components on Desktop app and Touchscreen app

# Changelog
- update DeckThumbnail component and its test
- add a prop(protocolAnalysis) to the above 4 components 

# Review requests

# Risk assessment
low


app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,125 @@
...
   )
-  const initialLoadedModulesBySlot = parseInitialLoadedModulesBySlot(commands)
-  const initialLoadedLabwareByModuleId = parseInitialLoadedLabwareByModuleId(
-    commands
+  const attachedModules =
+    useAttachedModules({
+      refetchInterval: ATTACHED_MODULE_POLL_MS,
+    }) ?? []
+
+  const enableDeckConfig = useFeatureFlag('enableDeckConfiguration')
```
brenthagen: no need to use the deck config feature flag, will be removed soon

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,116 @@
...
-  DeckSlot,
+  CompletedProtocolAnalysis,
   Liquid,
   LoadedLabware,
+  ProtocolAnalysisOutput,
   RunTimeCommand,
 } from '@opentrons/shared-data'
 
 interface DeckThumbnailProps extends StyleProps {
+  protocolAnalysis?: CompletedProtocolAnalysis | ProtocolAnalysisOutput | null
```
brenthagen: should make `protocolAnalysis` required

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,116 @@
...
   commands: RunTimeCommand[]
   labware: LoadedLabware[]
   liquids?: Liquid[]
```
brenthagen: think we don't need these props, we can get them from `protocolAnalysis`

brenthagen (commented):
looks good so far, but we need to add liquid info also. we don't have an example `BaseDeck` implementation of liquids yet - will be similar to RAUT-821

koji:
- [ ] add liquids

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,258 @@
...
-  showSlotLabels?: boolean
+  protocolAnalysis: CompletedProtocolAnalysis | ProtocolAnalysisOutput | null
 }
 
-export function DeckThumbnail(props: DeckThumbnailProps): JSX.Element {
-  const {
-    commands,
-    liquids,
-    labware = [],
-    showSlotLabels = false,
```
brenthagen, line 52: we want leave this prop in, and defaulted to false to hide the `SlotLabels`. only the ODD protocol details map version of `DeckThumbnail` shows slot labels.

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,258 @@
...
+    const topLabwareDefinition =
+      labwareInAdapterInMod?.result?.definition ?? module.nestedLabwareDef
+    return {
+      moduleModel: module.moduleDef.model,
+      moduleLocation: { slotName: module.slotName },
+      innerProps:
+        module.moduleDef.model === THERMOCYCLER_MODULE_V1
+          ? { lidMotorState: 'open' }
+          : {},
+      nestedLabwareDef: topLabwareDefinition,
```
brenthagen, line 90: here we want to add the well fill to the nested labware - add a prop `nestedLabwareWellFill` to `moduleLocations` in `BaseDeck`

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,258 @@
...
+        labwareInAdapter?.result?.definition ?? labwareDef
+      const topLabwareId = labwareInAdapter?.result?.labwareId ?? labwareId
+      const topLabwareDisplayName =
+        labwareInAdapter?.params.displayName ?? displayName
+
+      return {
+        labwareLocation: { slotName },
+        definition: topLabwareDefinition,
+        topLabwareId,
+        topLabwareDisplayName,
```
brenthagen: simliar, we want to add the well fill to the labware - add a prop `wellFill` to `labwareLocations` in `BaseDeck`

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,258 @@
...
-            const labwareInSlot =
-              slotId in initialLoadedLabwareBySlot
-                ? initialLoadedLabwareBySlot[slotId]
-                : null
-            const labwareInModule =
-              moduleInSlot?.result?.moduleId != null &&
-              moduleInSlot.result.moduleId in initialLoadedLabwareByModuleId
-                ? initialLoadedLabwareByModuleId[moduleInSlot.result.moduleId]
-                : null
+      {map(
```
brenthagen: all of the children content here inside `BaseDeck` should not not be needed if we are processing `moduleLocations` and `labwareLocations` correctly above - we should be able to use

```
          const wellFill = getWellFillFromLabwareId(
            topLabwareId ?? '',
            liquids,
            labwareByLiquidId
          )
          const labwareHasLiquid = !isEmpty(wellFill)
```
in those two locations above with the added well fill props

brenthagen (commented):
with these changes, we should end up with a fairly simple `<BaseDeck {...someProps} />` instance (with no children) in the render of `DeckThumbnail`

brenthagen:
also, it would be a good idea to change the title of this PR to indicate it is only doing a replacement in DeckThumbnail

app/src/molecules/DeckThumbnail/index.tsx
koji: @brenthagen 
updated code based on your comment and one question about `labwareHasLiquid`.
Do we need to add interactions that only exists in DeckView of `SetupLiquidsMap`? 

Screenshot 2023-11-06 at 1 47 25 PM

brenthagen: i think no, we don't need to add liquid hover or click interactions to DeckThumbnail. so we won't need `labwareHasLiquid` in DeckThumbnail

koji: okay, thank you!

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -1,179 +1,129 @@
...
-                  </g>
-                ) : null}
-              </React.Fragment>
-            )
-          })}
-          {showSlotLabels ? <SlotLabels robotType={robotType} /> : null}
-        </>
-      )}
-    </RobotWorkSpace>
+      {showSlotLabels ? <SlotLabels robotType={robotType} /> : null}
```
brenthagen: `SlotLabels` can actually be removed in this component, `showSlotLabels` added as a prop to `BaseDeck` and passed there

components/src/hardware-sim/BaseDeck/BaseDeck.tsx
```diff
@@ -184,21 +191,30 @@ export function BaseDeck(props: BaseDeckProps): JSX.Element {
...
               labwareLocation !== 'offDeck' &&
               'slotName' in labwareLocation &&
               s.id === labwareLocation.slotName
           )
+          const labwareHasLiquid = !isEmpty(wellFill)
           return slotDef != null ? (
             <g
               key={slotDef.id}
               transform={`translate(${slotDef.position[0]},${slotDef.position[1]})`}
+              cursor={labwareHasLiquid ? 'pointer' : ''}
```
brenthagen: would be better to condition cursor on existence of `onLabwareClick`

brenthagen (approved):
looks good, comments on slot labels and hover could be addressed in a follow on PR if wanted

PR #13877, opened 2023-10-31
fix(robot-server): create maintenance run if there is no protocol run

mjhuff:
Closes RQA-1859

# Overview

The AssertionError case for protocol_run_state should return true (no protocol run exists) instead of false.

# Test Plan
- Perform any type of maintenance run. Confirm that you can move the gantry to the front.

# Changelog
- Fix bug where users can't create maintenance runs if a protocol run doesn't exist.

### DIFF
--- a/robot-server/robot_server/runs/dependencies.py
+++ b/robot-server/robot_server/runs/dependencies.py
@@ -121,6 +121,7 @@ async def get_engine_store(
     return engine_store
 
 
+# TODO (jh, 2023-10-31): Add test cases.
 async def get_is_okay_to_create_maintenance_run(
     engine_store: EngineStore = Depends(get_engine_store),
 ) -> bool:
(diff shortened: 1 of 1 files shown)


fsinapi (approved):
cool and good 👍

sanni-t (approved):
We should add a test for this.. not for this PR though. Can you add a todo in there to test this case?

sfoster1 (commented):
We sure about this? This code appears to have been this way since maintenance runs were introduced. @sanni-t can you take a look?

sanni-t:
> We sure about this? This code appears to have been this way since maintenance runs were introduced. @sanni-t can you take a look?

That particular line has been there, yes, but we changed the dependency from `get_protocol_run_has_benn_played` to `get_is_okay_to_create_maintenance_run` which changes the assertion.. so changes from 'is it not okay?' to 'is it okay?'

sfoster1 (approved):
Ah, nice!

PR #13878, opened 2023-10-31
test(robot-server): Add integration tests for maintenance run creation

SyntaxColoring:
# Changelog

Add Tavern integration tests for the following cases:

* You should be able to create a maintenance run while there is a normal run, but the normal run hasn't been started yet. (This has always worked.)
* You should be able to create a maintenance run while there is a normal run, but the normal run has completed. (This was added in #13818.)
* You should be able to create a maintenance run while there is no normal run. (This was accidentally broken in #13818 and fixed in #13877.)

This resolves a todo comment added in #13877.

Also, our production code was catching `AssertionError`s. This is generally a bug because it [relies on](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) `__debug__ == True`, and it was unclear what was raising it. So, replace it with an explicit exception type defined by `EngineStore`.

# Test plan

None needed.

# Review requests

Any other juicy cases we want to add while we're at it?

# Risk assessment

Low.

### DIFF
--- a/robot-server/robot_server/maintenance_runs/maintenance_engine_store.py
+++ b/robot-server/robot_server/maintenance_runs/maintenance_engine_store.py
@@ -33,7 +33,7 @@ class EngineConflictError(RuntimeError):
     """
 
 
-class NoEngineError(RuntimeError):
+class NoRunnerEnginePairError(RuntimeError):
     """Raised if you try to get the current engine or runner while there is none."""
 
 
@@ -95,14 +95,14 @@ class MaintenanceEngineStore:
     def engine(self) -> ProtocolEngine:
         """Get the "current" ProtocolEngine."""
         if self._runner_engine_pair is None:
-            raise NoEngineError()
+            raise NoRunnerEnginePairError()
         return self._runner_engine_pair.engine
 
     @property
     def runner(self) -> LiveRunner:
         """Get the "current" ProtocolRunner."""
         if self._runner_engine_pair is None:
-            raise NoEngineError()
+            raise NoRunnerEnginePairError()
         return self._runner_engine_pair.runner
 
     @property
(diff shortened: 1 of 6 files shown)


robot-server/robot_server/maintenance_runs/maintenance_engine_store.py
```diff
@@ -33,6 +33,10 @@ class EngineConflictError(RuntimeError):
     """
 
 
+class NoEngineError(RuntimeError):
```
sanni-t: I think we should call this something like `NoRunnerEnginePair` to indicate that there's neither a runner nor an engine present. This makes the below errors more relevant- both the `engine` and `runner` properties can raise this error, but getting a `NoEngineError` when asked for a `runner` could cause confusion.

robot-server/robot_server/runs/engine_store.py
```diff
@@ -41,6 +41,10 @@ class EngineConflictError(RuntimeError):
     """
 
 
+class NoEngineError(RuntimeError):
```
sanni-t: Same comment as for maintenance_engine_store

robot-server/tests/integration/http_api/runs/test_maintenance_run_creation.tavern.yaml
```diff
@@ -0,0 +1,117 @@
...
+  - usefixtures:
+      - ot2_server_base_url
+
+stages:
+  - *assert_no_runs
+  - *assert_no_maintenance_runs
+
+  - name: Create a normal run
+    request:
+      url: '{ot2_server_base_url}/runs'
```
sanni-t: Maybe create a protocol run to be more accurate about the scenario? This will create a 'setup run' whose behavior is similar to maintenance runs (and we should really just remove it at this point).

sanni-t (commented):
Nice integration tests! Just some nitpicks

sanni-t (approved):
🏎️

PR #13880, opened 2023-10-31
ci(app-shell): use notarytool for OSX signing

b-cooper:
# Overview

The legacy tool will no longer be supported, move to use `notarytool`

# Risk assessment

low

### DIFF
--- a/app-shell/electron-builder.config.js
+++ b/app-shell/electron-builder.config.js
@@ -2,7 +2,7 @@
 const path = require('path')
 const { versionForProject } = require('../scripts/git-version')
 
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good. To verify that it worked properly we'll want to get a full app build going (which means I think we need to add `-app-build` on the end of the branch name) and check notarization by running it on a machine

sfoster1:
Pushed https://github.com/Opentrons/opentrons/tree/app-shell_chore-use-notarytool-for-signing-app-build-both to force the build, more changes might want to go there for automatic build starting @b-cooper

b-cooper:
Moved to https://github.com/Opentrons/opentrons/pull/13883 to include branch name with app build trigger

PR #13881, opened 2023-10-31
chore(robot-server): Fix robot integration tests

sfoster1:
196fdc002c added some integration tests for json protocol v8 on flex but didn't mark them as ot3_only. This does that.

Also, add a 30 second timeout for starting the background server for tavern tests so that if there's an error that causes the server to fail to start, instead of waiting infinitely we'll actually raise an error.

### DIFF
--- a/robot-server/tests/integration/conftest.py
+++ b/robot-server/tests/integration/conftest.py
@@ -4,6 +4,7 @@ import json
 import time
 from pathlib import Path
 from typing import Any, Dict, Generator
+from datetime import datetime
 
 import pytest
 import requests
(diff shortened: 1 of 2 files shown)


robot-server/tests/integration/conftest.py
```diff
@@ -110,7 +111,11 @@ def _requests_session() -> Generator[requests.Session, None, None]:
 
 def _wait_until_ready(base_url: str) -> None:
     with _requests_session() as requests_session:
+        started = datetime.now()
         while True:
+            now = datetime.now()
+            if (now - started).total_seconds() > 30:
```
sanni-t: I feel like we have a dev server initialization timeout declared somewhere that we can use here. If not, might not be bad to create one instead of the magic number. Just nitpicking though.

sanni-t (approved):

PR #13882, opened 2023-10-31
fix(api): add unit test for z axis moves in tip pickup

caila-marashaj:
## Overview
Just adds a unit test for non-tip motor moves in tip pickup. This wasn't done for dropoff  because we don't expect any such moves before and after drop-off.

## Review Request
Please let me know if there are any more tests I should add !


sfoster1 (approved):
thank ye

PR #13883, opened 2023-10-31
ci(app-shell): use notarytool for OSX signing

b-cooper:
# Overview

The legacy tool will no longer be supported, move to use notarytool

# Risk assessment
low


sfoster1 (approved):
Runs and passes gatekeeper on my machine, looks good!

PR #13888, opened 2023-11-01
feat(protocol-designer): use correct types in 8_0_0 migration

jerader:
closes RAUT-759

# Overview

Followup to the schemav8 migration. I accidentally forgot to update the `ProtocolFile` types in the `8_0_0` migration to be from V7 -> V8. This is to ensure protocols imported into PD follow the proper schema.

# Test Plan

Import a protocol to PD and make sure it imports as expected.

# Changelog

- update the types in `8_0_0` migration and fix the test
- add test cases to the cypress migrations test

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/cypress/integration/migrations.spec.js
+++ b/protocol-designer/cypress/integration/migrations.spec.js
@@ -167,7 +167,8 @@ describe('Protocol fixtures migrate and match snapshots', () => {
                       labwareId.includes(OT_2_TRASH_DEF_URI) ||
                       labwareId.includes(FLEX_TRASH_DEF_URI)
                     ) {
-                      labwareLocationUpdate['trashId'] = slot
+                      const trashId = 'trashId'
+                      labwareLocationUpdate[trashId] = slot
                       delete labwareLocationUpdate[labwareId]
                     }
                   }
--- a/protocol-designer/cypress/integration/migrations.spec.js
+++ b/protocol-designer/cypress/integration/migrations.spec.js
@@ -12,64 +12,56 @@ describe('Protocol fixtures migrate and match snapshots', () => {
   })
 
   const testCases = [
-    //  TODO(jr, 10/31/23): when 8.0 is more developed, lets go back and fix these
-    //  {
-    //   title: 'example_1_1_0 (schema 1, PD version 1.1.1) -> PD 7.0.x, schema 7',
-    //   importFixture: '../../fixtures/protocol/1/example_1_1_0.json',
-    //   expectedExportFixture:
-    //     '../../fixtures/protocol/7/example_1_1_0MigratedFromV1_0_0.json',
-    //   unusedPipettes: true,
-    //   migrationModal: 'newLabwareDefs',
-    // },
-    // {
-    //   title: 'doItAllV3 (schema 3, PD version 4.0.0) -> PD 7.0.x, schema 7',
-    //   importFixture: '../../fixtures/protocol/4/doItAllV3.json',
-    //   expectedExportFixture:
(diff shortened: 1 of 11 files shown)


protocol-designer/cypress/integration/migrations.spec.js
```diff
@@ -161,10 +153,48 @@ describe('Protocol fixtures migrate and match snapshots', () => {
                 f.metadata.lastModified = 123
                 f.designerApplication.data._internalAppBuildDate = 'Foo Date'
                 f.designerApplication.version = 'x.x.x'
+
+                //  NOTE: labwareLocationUpdates, trash stubs can be removed for the release after 8.0.0
+                //  currently stubbed because of the newly created trash id for movable trash support
+                f.designerApplication.data.savedStepForms.__INITIAL_DECK_SETUP_STEP__.labwareLocationUpdate =
+                  'trash'
+
+                Object.values(f.designerApplication.data.savedStepForms).map(
```
shlokamin: nit, this `map` is misleading because its return value is not being used. if we're just mutating object properties in place (not returning a new cloned object with changed properties), we should use `forEach` for clarity

protocol-designer/cypress/integration/migrations.spec.js
```diff
@@ -161,10 +153,48 @@ describe('Protocol fixtures migrate and match snapshots', () => {
...
+                  stepForm => {
+                    if (stepForm.stepType === 'moveLiquid') {
+                      stepForm.dropTip_location = 'trash drop tip location'
+                      stepForm.blowout_location = 'trash blowout location'
```
shlokamin: how come we have to generify the drop tip and blowout locations? this seems like important information to verify

protocol-designer/cypress/integration/migrations.spec.js
```diff
@@ -161,10 +153,48 @@ describe('Protocol fixtures migrate and match snapshots', () => {
...
+                    }
+                    if (stepForm.stepType === 'mix') {
+                      if (stepForm.labware.includes('trash')) {
```
shlokamin: hmm doing a string.includes feels brittle — how come we cant check for an explicit id?

protocol-designer/cypress/integration/migrations.spec.js
```diff
@@ -161,10 +153,48 @@ describe('Protocol fixtures migrate and match snapshots', () => {
                 f.metadata.lastModified = 123
                 f.designerApplication.data._internalAppBuildDate = 'Foo Date'
                 f.designerApplication.version = 'x.x.x'
+
+                //  NOTE: labwareLocationUpdates, trash stubs can be removed for the release after 8.0.0
+                //  currently stubbed because of the newly created trash id for movable trash support
+                f.designerApplication.data.savedStepForms.__INITIAL_DECK_SETUP_STEP__.labwareLocationUpdate =
+                  'trash'
+
```
shlokamin: lets just mask the trash, and lets do some sort of check to make sure the trash is in the right location

protocol-designer/src/load-file/migration/utils/getLoadLiquidCommands.ts
```diff
@@ -3,21 +3,21 @@ import { uuid } from '../../../utils'
...
   >
-  ingredLocations: {
+  ingredLocations?: {
     [labwareId: string]: {
       [wellName: string]: { [liquidId: string]: { volume: number } }
     }
   }
   savedStepForms: Record<string, any>
-  orderedStepIds: string[]
+  orderedStepIds?: string[]
```
shlokamin: i think these properties are still required, lets make sure about this

shlokamin (approved):

PR #13889, opened 2023-11-01
feat(app): prepareToAspirate support in command text

jerader:
closes RAUT-830

# Overview

This PR adds the newly created `prepare_to_aspirate` to the command text. 

# Test Plan

Just review the code - the test case is already testing what it would read. If you want, you can create a python protocol that uses `prepare_to_aspirate` See PR that adds it here: https://github.com/Opentrons/opentrons/pull/13827

# Changelog

- add i18n string to `prepareToAspirate` and add test 

# Review requests

see test plan

# Risk assessment

low


brenthagen (approved):
makes sense to me !

PR #13891, opened 2023-11-01
fix(app, robot_server): allow non-standard tip racks in tip length calibration from dashboard

b-cooper:
# Overview

Mirror our approach for loading an alternate tip rack into a legacy calibration session from deck/pipette-offset to tip length calibration

Closes RAUT-809

# Review requests

- starting with no tip length calibration values on an OT-2 proceed to the calibration dashboard
- you should be able to select an alternate tip rack for performing TLC on a pipette. 
- that alternate tip rack should be used when calibrating the pipette offset in the next step

# Risk assessment
high, this touches both FE and BE and the code is quite old and brittle


mjhuff (approved):
Working as expected and code makes sense. Thanks for walking me through the backend!

PR #13894, opened 2023-11-01
feat(robot-server): Add HTTP endpoints for getting and setting the robot's deck configuration

SyntaxColoring:
# Overview

Closes RSS-334 and RSS-337.

# Test Plan

No further manual testing needed. I've tested this in Postman and added an integration test.

# Changelog

Adds placeholder implementations of two new HTTP endpoints, `PUT /deck_configuration` and `GET /deck_configuration`.

```http
PUT /deck_configuration
{
  "data": {
    "cutoutFixtures": [
      {
        "cutoutFixtureId": "singleLeftSlot",
        "cutoutId": "cutoutA1"
      },
      {
        "cutoutFixtureId": "singleLeftSlot",
        "cutoutId": "cutoutB1"
      }
      <...>
    ],
  }
}
```

```http
GET /deck_configuration
{
  "data": {
    "lastUpdatedAt": "<timestamp>"
    "cutoutFixtures": [
      {
        "cutoutFixtureId": "singleLeftSlot",
        "cutoutId": "cutoutA1"
      },
      {
        "cutoutFixtureId": "singleLeftSlot",
        "cutoutId": "cutoutB1"
      }
      <...>
    ],
  }
}
```

Current limitations (to be addressed in other PRs):

* The new endpoints don't actually do anything yet, in terms of motion planning or other run behavior.
* The data gets wiped away when the server restarts.
* The default deck configuration is, erroneously, empty.
* When you set a new deck configuration, it's not validated in any way.

# Review requests

Run `make -C robot-server dev` and peruse the public HTTP API documentation. http://localhost:31950/redoc#tag/Deck-Configuration

Any surprises in the specification? Anything that's unclear?

# Risk assessment

No risk. This is all new code; nothing is relying on it yet.

I do suggest we merge #13917 immediately after this. If we go for too long without this endpoint having any server-side validation, the client-side code might have a chance to develop bugs that would cause undue churn.


robot-server/robot_server/deck_configuration/__init__.py
```diff
@@ -0,0 +1 @@
+"""The HTTP API, and supporting code, for getting and setting the robot's deck configuration."""
```
SyntaxColoring, line 1: Most of our `robot_server` subdirectories have an `__init__.py` file that re-exports the stuff that's meant to be public outside that subdirectory.

I'm deliberately *not* adding any re-exports here, because:

1. I think it's a poor way to indicate submodule visibility, given that Python already has the underscore-prefix convention.
2. In practice, it causes circular import problems for us. 

Happy to clarify either of these if anybody cares.

SyntaxColoring:
I'll be out until 2023-11-13. Feel free to take this over or merge it in my absence.

brenthagen (approved):
from app & ui perspective this makes sense and generally matches our expectations.

robot-server/robot_server/deck_configuration/models.py
```diff
@@ -0,0 +1,64 @@
...
+class CutoutFixture(pydantic.BaseModel):
+    """A single element of the robot's deck configuration."""
+
+    # These are deliberately typed as plain strs, instead of Enums or Literals,
+    # because we want shared-data to be the source of truth.
+    #
+    # The downside of this is that to use this HTTP interface, you need to be familiar with deck
+    # definitions. To make this better, we could perhaps autogenerate OpenAPI / JSON Schema spec
+    # fragments from shared-data and inject them here.
+    cutoutFixtureId: str = pydantic.Field(
```
sanni-t, line 22: How about creating the types in shared-data with tests validating them against the definitions (quite like schema validation tests in js)?

robot-server/robot_server/deck_configuration/models.py
```diff
@@ -0,0 +1,64 @@
...
+            " [deck definition](https://github.com/Opentrons/opentrons/tree/edge/shared-data/deck)."
+        )
+    )
+
+
+class DeckConfigurationRequest(pydantic.BaseModel):
+    """A request to set the robot's deck configuration."""
+
+    cutoutFixtures: List[CutoutFixture] = pydantic.Field(
+        description="A full list of all the cutout fixtures that are mounted onto the deck."
```
sanni-t, line 42: What's the reason for disallowing updating partial deck config? Is it mainly because the app will never need to send a partial deck config? Just curious

robot-server/robot_server/deck_configuration/validation.py
```diff
@@ -0,0 +1,114 @@
...
+    deck_definition: deck_types.DeckDefinitionV4, cutout_fixture_id: str
+) -> Union[deck_types.CutoutFixture, UnrecognizedCutoutFixtureError]:
+    try:
+        return next(
+            cutout_fixture
+            for cutout_fixture in deck_definition["cutoutFixtures"]
+            if cutout_fixture["id"] == cutout_fixture_id
+        )
+    except StopIteration:
+        return UnrecognizedCutoutFixtureError(cutout_fixture_id=cutout_fixture_id)
```
sanni-t, line 114: Interesting. Why return the error instead of raising and catching it when called?

sanni-t (approved):
Great testing on the config validation! 
LGTM! Just some thoughts and questions

robot-server/robot_server/deck_configuration/models.py
CaseyBatten: The idea is that the requestor for reconfiguration is responsible for acknowledging the layout of their entire deck, rather than either side assuming the layout based on previous layout information.

robot-server/robot_server/deck_configuration/validation.py
SyntaxColoring: Oh this came from #13917.

Yeah, raising and catching also would have worked here.

I returned this error as a value instead because it felt more harmonious with this module's public `get_configuration_errors()` API. *That* chooses to return errors instead of raising them because: 

* It lets us use the static type-checker to keep things in sync, which is especially meaningful here because of how many possible errors there are.
* It avoids having to wrap a collection of multiple errors in some kind of `MultiValidationError` wrapper exception.

PR #13897, opened 2023-11-01
fix(app): fix choose robot slideout for usb

shlokamin:
# Overview

This PR fixes the shoose a robot slideout when sending a protocol to a Flex and starting a protocol run on a Flex over usb.

# Test Plan

- Verified that I could send a protocol/start a protocol run on a Flex over USB
- Verified that I could send a protocol/start a protocol run on a Flex over the network

# Changelog

- Fix choose robot slideout for usb

Risk assessment

Low/med


brenthagen (approved):
✔️

PR #13900, opened 2023-11-02
feat(protocol-designer): modal updates and deck map design tweaks

jerader:
closes RAUT-807 RAUT-816 RAUT-817

# Overview

Updates the announcement and export modals for PD 8.0 release and a few design tweaks to the deck map

Screen Shot 2023-11-01 at 9 24 27 PM
Screen Shot 2023-11-01 at 9 23 58 PM

# Test Plan

Open protocol designer and look at the new announcement modal. if the modal doesn't pop up then type `localStorage.clear()` in dev tools and refresh. Then go through the new protocol file wizard and export. See that the export modal says that app version 7.1 is required.

# Changelog

- update export modal
- update announcement array with a new image and text
- tweak deck setup so its centered and add a border on the module tags so it doesnt blend into the trash bin or waste chute (if a module is on the slot above the trash)
- fix tests

# Review requests

see test plan

# Risk assessment

low


mjhuff (approved):
LGTM! Always appreciate the clear test plans.

Just my two cents, but the deck config image seems like it should have a slightly higher top margin value than it does. Maybe this is a design decision? Not a big deal either way 🚀

jerader:
> Just my two cents, but the deck config image seems like it should have a slightly higher top margin value than it does. Maybe this is a design decision? Not a big deal either way 🚀

thank you @mjhuff! Nah there are no designs, looking at it now i agree that the top margin should be higher, thank you!

PR #13901, opened 2023-11-02
fix(app-shell)Fix python env contamination app build

sfoster1:
This prevents python from picking up versions of dependencies installed on the system, which will break offline analysis if the dependencies installed globally are old.

# testing
- [x] on a compiled windows and compiled mac build, check that you can run offline analysis


sfoster1:
This works on the original issue also!

shlokamin (approved):

PR #13903, opened 2023-11-02
fix(protocol-designer): loadName in loadLabware is now true loadName

jerader:
hot fix addressing https://opentrons.atlassian.net/browse/RESC-161

# Overview

I mistakenly defined the `loadLabware`'s loadname as `definitionUri` instead of the true `loadName` and it caused a white screen bug related to duplicating labware and then importing the protocol into PD and then reimporting. (because duplicate labware from <v7 don't have a `definitionURI` as part of the `labwareId`)

# Test Plan

Upload the Created 8 September protocol attached to the ticket. See that it imports correctly. Export it and then reimport that exported protocol. It should work as expected.

# Changelog

- fix the 7.0.0 migration and then remove branching logic from the reducer.

# Review requests

see test plan

# Risk assessment

low but will turn into PD 7.0.1 hot fix

### DIFF
--- a/protocol-designer/src/step-forms/reducers/index.ts
+++ b/protocol-designer/src/step-forms/reducers/index.ts
@@ -1157,6 +1157,11 @@ export const labwareInvariantProperties: Reducer<
                 labwareDef.namespace === namespace &&
                 labwareDef.version === version
             )
+            if (labwareDefinitionMatch == null) {
+              console.error(
+                `expected to find labware definition match with loadname ${loadName} but could not`
+              )
+            }
             const labwareDefURI =
               labwareDefinitionMatch != null ? labwareDefinitionMatch[0] : ''
             const id = labwareId ?? ''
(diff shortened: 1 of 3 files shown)


protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1148,13 +1148,22 @@ export const labwareInvariantProperties: Reducer<
...
+            const labwareDefinitionMatch = Object.entries(
+              file.labwareDefinitions
+            ).find(
+              ([definitionUri, labwareDef]) =>
+                labwareDef.parameters.loadName === loadName &&
+                labwareDef.namespace === namespace &&
+                labwareDef.version === version
+            )
+            const labwareDefURI =
+              labwareDefinitionMatch != null ? labwareDefinitionMatch[0] : ''
```
shlokamin, line 1166: if we cant find a labware definition match that means the user's PD protocol must be corrupted. we should console error or something here when that happens

shlokamin (approved):

PR #13904, opened 2023-11-02
fix(protocol-designer): separate well order field tooltip and modal

b-cooper:
# Overview

Prevent ghost tooltip from appearing after well order modal is opened

Closes RAUT-266

# Risk assessment

low


jerader (approved):
lgtm!

PR #13905, opened 2023-11-02
fix(app): fix last run protocol sorting on protocol dashboard

b-cooper:
# Overview

We were incorrectly calculating the last run timestamp of a given protocol, this fixes the logic to report the correct value

Closes RAUT-843

# Review Requests

- re-run a protocol on a Flex.  From the Protocol Dashboard the timestamp of the newest run of that protocol should be visible in the table. 

# Risk assessment
low


mjhuff (approved):
lgtm!

PR #13915, opened 2023-11-03
feat(app, app-shell): add download progress to robot update flows

mjhuff:
Closes RAUT-789, RAUT-827

# Overview

If the robot update cache files are still downloading, no longer is an error thrown. Instead, the update modal renders and a downloading stage of the update flow begins. Additionally, release files are downloaded and returned to the browser layer separately from system files, so users may see release notes as soon as they are downloaded.

### What's Changed

The substantive changes are in app-shell/src/robot-update/index & release-files and app/src/redux/robot-update/epic & reducer. I've left comments to help explain what's happening.

### Before

https://github.com/Opentrons/opentrons/assets/64858653/5719968a-97b6-427e-9f58-5636aa1f3706

### After

https://github.com/Opentrons/opentrons/assets/64858653/c3c9d455-b821-4e9a-b142-d8a47325f04c

# Test Plan
- Build the app with make -C app-shell package. Failing the signing step is expected (and fine for testing purposes).
- Before opening the app, clear out the robot-update-cache files (located in Users/username/.Library/Application Support/opentrons/robot-update-cache).
- Perform the next step quickly if you have a good downlink. If you want to throttle your bandwidth, you can use the Network Link Conditioner offered by Mac. 
- Follow the video flow, testing on a Flex. You may need to downgrade a robot first via zip-file (via USB if you're feeling bold).
- Repeat testing flow on an OT-2 (although there should be no functional difference).

# Changelog
- Initiating a robot update while the files are still downloading no longer throws an error. 

# Review requests
- Let me know if you need help testing!

# Risk assessment
Medium - Getting the logic right is important, but it's relatively (keyword) straightforward as far as update flows are concerned...


app-shell/src/robot-update/index.ts
```diff
@@ -137,20 +137,28 @@ export function registerRobotUpdate(dispatch: Dispatch): Dispatch {
...
-          return dispatch({
-            type: 'robotUpdate:UNEXPECTED_ERROR',
-            payload: { message: 'Robot update file not downloaded' },
-          })
+          if (checkingForUpdates) {
+            dispatch({
+              type: 'robotUpdate:CHECKING_FOR_UPDATE',
+              payload: target,
+            })
+          } else {
```
mjhuff, line 153: This condition is hit if a user actively starts a robot update, but the file has not yet been downloaded. The CHECKING_FOR_UPDATE action creates an update session for the robot with a downloadFile stage. The UI layer uses this session to alert the user that the file is downloading.

app-shell/src/robot-update/release-files.ts
```diff
@@ -44,41 +51,65 @@ export function getReleaseFiles(
...
+        downloadAndNotify(true, urls.releaseNotes, directory, dispatch, target),
+        downloadAndNotify(
+          false,
+          urls.system,
+          directory,
+          dispatch,
+          target,
+          onProgress
+        ),
+      ]).then(([releaseNotes, system]) => ({ releaseNotes, system }))
```
mjhuff, line 64: If a user clicks an the robot update banner while the file is still downloading, there will not be release notes (as seen in the before vide). Let's separate the release notes from the system file download.

If a user somehow navigates to the robot update banner before 20kb have been downloaded, they won't see the notes, but this won't prevent the download.

app-shell/src/robot-update/release-files.ts
```diff
@@ -44,41 +51,65 @@ export function getReleaseFiles(
...
+          return readFile(path, 'utf8').then(releaseNotes =>
+            dispatch({
+              type: 'robotUpdate:UPDATE_INFO',
+              payload: { releaseNotes, target, version: CURRENT_VERSION },
+            })
+          )
+        }
+        // This action will only have an effect if the user is actively waiting for the download to complete.
+        else {
+          return dispatch({
```
mjhuff, line 106: This dispatch occurs after the system files are downloaded but not yet cleaned up. The redux store will update the robot's session from "waiting on download to finish" to done if the robot is of the same version as the file that finished downloading.

All the file cleanup can happen asynchronously without any problem.

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/useRobotUpdateInfo.ts
mjhuff, line 1: All the changes here are a refactor. I'll revisit this again in the future as it's still obtuse. 

The guarded transitions can be safely removed now, since the app-shell guards the transitions safely.

app/src/redux/robot-update/epic.ts
```diff
@@ -144,6 +145,19 @@ export const startUpdateEpic: Epic = (action$, state$) =>
     })
   )
 
+export const startUpdateAfterFileDownload: Epic = (_, state$) => {
```
mjhuff, line 148: After the robot update file download completes, this pipe sends us back to the app-shell to kick off the installation process. 

This time on READ_SYSTEM_FILE in app-shell, the filename will not be null. Note that the process isn't recursive (checkingForUpdates can never occur more than 1x unless the app is hard restarted. If the file doesn't exist the second time we get here, we'll throw the UNEXPECTED_ERROR.

app/src/redux/robot-update/reducer.ts
```diff
@@ -76,6 +91,24 @@ export const robotUpdateReducer: Reducer<RobotUpdateState, Action> = (
       }
     }
 
+    case Constants.ROBOTUPDATE_DOWNLOAD_DONE: {
+      if (!state.session) return state
+
+      const { target, ...session } = state.session
+      const isThisRobotDownloadDone =
+        session?.step === Constants.DOWNLOAD_FILE && target === action.payload
```
mjhuff, line 102: This check prevents us initiating an update session when the user isn't actively engaged in the robot update flows, and it makes sure the correct system file update has completed (ex, the OT-2 file completing download won't trigger a Flex waiting for download to register as "done").

app-shell/src/http.ts
```diff
@@ -125,10 +125,10 @@ function createReadStreamWithSize(
     }
 
     readStream.once('error', handleError)
-    readStream.on('data', onData)
 
     function handleSuccess(): void {
       resolve(readStream)
+      readStream.removeListener('error', handleError)
```
sfoster1: this technically has the same pitfall that https://github.com/Opentrons/opentrons/pull/13923 fixes on the odd side (errors from the read stream will be lost) but here I think those errors would only happen if the user deleted the cached file while the update was running

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/RobotUpdateProgressModal.tsx
```diff
@@ -105,7 +98,9 @@ export function RobotUpdateProgressModal({
     progressPercent
   )
 
-  let modalBodyText = t('installing_update')
+  let modalBodyText = ''
```
sfoster1, line 101: unless we can prove exhaustion in the if cases after this I think we should give this a meaningful default value
```
let modalBodyText = t('downloading_update')
if (updateStep === 'install') modalBodyText = t('downloading_update')
```
or even `const` it and do a chained ternay assign, or have a function that takes the step and returns the text

sfoster1 (approved):
Looks good to me, some small nits

PR #13920, opened 2023-11-06
fix(app): fix step meter tearing during wizard flows

mjhuff:
Closes RAUT-627, RQA-1341, RQA-1342 

# Overview

Add ODD optimizations to prevent step meter tearing that occurs when pressing "Go Back" during various flows. This seems to solve the issue where it was occurring on the physical ODD. 

### Prior Behavior

a7a4658d-3bea-4fa3-ae70-b60e709a8503

# Test Plan

This issue is intermittent, so it may be difficult to repro while testing. 

- Test the "Go Back" button within any wizard flow, preferably the instrument attach/cal flows.
- Confirm that StepProgress behavior is consistent with the old behavior and that tearing does not occur.

# Changelog
- Fixed step meter tearing in wizard flows.

# Review requests
- Test on the physical Flex, please!

# Risk assessment
low

### DIFF
--- a/app/src/atoms/StepMeter/index.tsx
+++ b/app/src/atoms/StepMeter/index.tsx
@@ -32,15 +32,21 @@ export const StepMeter = (props: StepMeterProps): JSX.Element => {
       height: ${SPACING.spacing12};
     }
   `
+
+  const ODD_ANIMATION_OPTIMIZATIONS = `
+  backface-visibility: hidden;
+  perspective: 1000;
+  will-change: transform;
+  `
+
   const StepMeterBar = css`
+    ${ODD_ANIMATION_OPTIMIZATIONS}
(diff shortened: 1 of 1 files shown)


app/src/atoms/StepMeter/index.tsx
```diff
@@ -32,15 +32,21 @@ export const StepMeter = (props: StepMeterProps): JSX.Element => {
       height: ${SPACING.spacing12};
     }
   `
+
+  const ODD_ANIMATION_OPTIMIZATIONS = `
+  backface-visibility: hidden;
+  perspective: 1000;
+  will-change: transform;
```
jerader, line 39: wow i'm always learning new css properties from you! 🤩

jerader (approved):
🖌️

PR #13921, opened 2023-11-06
fix(api): teardown non-existing tip detector shouldn't raise an error

ahiuchingau:
# Overview
This PR fixes a few things that block a new tip detector from being built. 

1. we shouldn't raise a tip detector not found error when we're trying to tear down one that doesn't exist
2. `get_detector()` method doesn't return None actually
3. f-String log statement missing an "f"
4. rename test file so it's only being tested for ot3
5. delete obsolete files in `hardware/`


api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -78,14 +78,16 @@ def _unsubscribe() -> None:
...
                 self.set_unsub(mount, None)
 
-        detector = self.get_detector(mount)
-        if detector:
+        try:
+            detector = self.get_detector(mount)
             _unsubscribe()
             detector.cleanup()
             self.set_detector(mount, None)
+        except TipDetectorNotFound:
```
sfoster1, line 84: Wait, we need a bit more than this - if there's an exception early in this process, code that runs after it won't actually run. We should do as much as possibly in a `finally`

sfoster1 (changes requested):

sfoster1 (approved):
looks good to me!

PR #13923, opened 2023-11-06
feat(app): Update robots from USB flash drive

sfoster1:
This PR adds the ability to update a Flex's software from a USB flash drive inserted in the front (or I guess rear) port containing an update file. This closes a functionality gap where Flexes can be really hard to update with absolutely no app connectivity, given that the OT-2's fallback of "pull and reflash the SD card" isn't available.

It also fixes a bug where updates downloaded from the internet on the ODD wouldn't work because of old code we'd added when 

These are a lot of changes, so I separated them by commit:
- a235409961a2dd2b4d63019ec4ea11f4b9c34e2b introduces the filesystem watch for USB mass storage devices
- b976819aed57ae04fbc23e023d2782e74e8a50a1 adds error handling for the readstream we use to post files from the app shell
- 7981c068bd44419f6909a2103c79b68f73282ada is a minor app-side fix to actually add update polling to the ODD
- 87fd827e748b522ee19b6b15ab6af5edec14eee9 is the node side of providing system updates from USB, with a lot of changes to `app-shell-odd/system-updates` (and a bonus bugfix for an issue that caused the ODD to never in practice prompt for an update ([RQA-1602](https://github.com/Opentrons/opentrons/pull/13923))
- 574a1d90ae7407af9297639afc905f978f20cbc3 is the react side and is actually mostly about disentangling the robot update code from the onboarding flow

## Risk assessment
Medium; this is a little-used flow on the ODD but it is pretty annoying to test, and it did touch onboarding

## Testing
- [x] Updates during onboarding should still work
   - [x] You should still go to the check-update screen from both the wifi connection successful screen and the ethernet connection successful screen
   - [x] If no update is available (do this by running onboarding on the most recent version) then that screen should move you to estop connection
   - [x] If the update succeeds, you should come up after reboot to the estop screen
   - [x] If you cancel the update or there's an error you should go to the estop screen
- [x] Updates not during onboarding should never accidentally drop you into onboarding
   - [x] If you update, when the robot boots you should not be in the estop screen
   - [x] If you cancel the update, you should end up back at the robot settings screen
- [x] Updates should still, like, function
- [x] If no update is available from the internet, no prompt should be shown
   - [x] If you then insert a USB drive, you should get a prompt for the update on the drive
      - [x] even if it's for a lower version than you're currently running
   - [x] If you then remove the USB drive the prompt should go away
      - [x] If you remove the USB drive while viewing the update details screen the prompt should go away
- [x] If an update is available from the internet, a prompt should be shown (this is the little bonus bugfix)
   - [x] The update's details screen should now actually have release notes
   - [x] If you plug in a USB drive you should still have a notification, and the details screen should say it's from USB
   - [x] If you remove the USB drive you should be back at the notification and release notes for the internet update
   - [x] If you do an update with a USB drive you should get the update from the USB drive
- [x] If you start an update from a USB drive and then pull it out while the update is "downloading", you should more or less immediately get an error that says something about os errors (this is the http.ts thing)
- [x] If you start an update from a USB drive and then pull it out while the update is "validating" or later it should be fine

Closes RAUT-829
Closes RQA-1602

[RQA-1602]: https://opentrons.atlassian.net/browse/RQA-1602?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mjhuff (approved):
This is awesome! Thanks for fixing the update check logic, too!

PR #13927, opened 2023-11-06
fix(app): fix calibration copy, 96 wizard attach title

smb2268:
fix RQA-1866, RQA-1767

# Overview

Fix two 96-channel bugs:
1. Title of 96-channel attach wizard changes to "Detach and attach 96-channel" after attachment
2. Robot settings calibration item mount listed as "left" instead of "both"

# Test Plan

1. Attach a 96-channel pipette with an empty gantry, see that for the duration of the wizard flow the title remains "Attach 96-Channel Pipette"
2. Once the pipette is attached and calibrated, go to robot settings/calibration on desktop and see that under "Mount" you now see "Both"

# Changelog
Memoize "isGantryEmpty" variable so it doesn't become `false` after 96-pipette is attached and update mount to say "both" when 96 channel is attached

# Review requests
Look over code and test plan

# Risk assessment
Low


koji (approved):
The changes look good to me.

PR #13933, opened 2023-11-06
fix(api): only send moves to moving axes, unless specified otherwise

ahiuchingau:
# Overview
This allows us to keep certain axes disengaged while we command other axes to move.

Instead of sending moves to all present motor nodes all the time, regardless of
whether they actually need to move or not, we are only sending the move requests to 
the axes that are defined in the moves.

Most of the time we're ordering a mount to move (see various move methods in ot3api).
This guarantees that the axes pertaining to that mount (X, Y, and Z) will be receiving the moves
together but the other idle mounts will be left alone.


sfoster1 (approved):
Ah, this is really nice, thanks!

PR #13934, opened 2023-11-07
fix(api): refresh motor & encoder positions after fast home failure

ahiuchingau:
# Overview
After we encounter and ignore a Stall/Collision error during fast home, we need to manually refresh motor positions so we can properly compute the distance/duration for the following moves.


sfoster1 (approved):
Makes sense!

PR #13935, opened 2023-11-07
fix(app): Fix desktop app update modals

mjhuff:
Closes RQA-1877, RQA-1643

# Overview

This PR fixes a bug in the desktop robot update flow in which a user downgrades their app then tries to downgrade the robot. Upon clicking the robot update banner, a user first sees an "update your app modal", which is expected. However, if the user clicks X or "Remind me later", the app crashes. 

### Before 

https://github.com/Opentrons/opentrons/assets/64858653/2d96ce7b-b097-47ea-a79b-cffdf065dc52

### After

https://github.com/Opentrons/opentrons/assets/64858653/1b0e5e69-9265-4772-895b-8da9f0f5a5cd

This PR also adds a new link to the GitHub release page.

Screenshot 2023-11-07 at 1 06 41 PM

# Test Plan

Update flows are always a bit fun to test.

### Release Notes button

- Checkout the TEST_BRANCH_RELEASE_NOTES branch.
- Run the app in dev mode and open the software updates modal.

### Robot Updates
- [Download a built version of the app here. ](https://opentrons.slack.com/archives/C81CR4VCZ/p1699301353028729)
- Follow the flow in the video. 

# Changelog
- Fixed a bug in which the app crashes when attempting to downgrade a robot and declining a desktop app update.
- Add a Release Notes link in desktop update modals.

# Risk assessment
low - very shallow update flow changes

### DIFF
--- a/app/src/assets/localization/en/app_settings.json
+++ b/app/src/assets/localization/en/app_settings.json
@@ -70,6 +70,7 @@
   "problem_during_update": "This update is taking longer than usual.",
   "prompt": "Always show the prompt to choose calibration block or trash bin",
   "receive_alert": "Receive an alert when an Opentrons software update is available.",
+  "release_notes": "Release notes",
   "remind_later": "Remind me later",
   "reset_to_default": "Reset to default",
   "restart_touchscreen": "Restart touchscreen",
--- a/app/src/organisms/Devices/RobotSettings/UpdateBuildroot/UpdateRobotModal.tsx
+++ b/app/src/organisms/Devices/RobotSettings/UpdateBuildroot/UpdateRobotModal.tsx
@@ -23,6 +23,7 @@ import {
   UPGRADE,
   REINSTALL,
   DOWNGRADE,
+  getRobotUpdateVersion,
 } from '../../../../redux/robot-update'
 import { ExternalLink } from '../../../../atoms/Link/ExternalLink'
 import { ReleaseNotes } from '../../../../molecules/ReleaseNotes'
@@ -35,8 +36,8 @@ import { useDispatchStartRobotUpdate } from '../../../../redux/robot-update/hook
 import type { State, Dispatch } from '../../../../redux/types'
 import type { RobotSystemType } from '../../../../redux/robot-update/types'
 
-export const RELEASE_NOTES_URL =
-  'https://github.com/Opentrons/opentrons/releases'
+export const RELEASE_NOTES_URL_BASE =
+  'https://github.com/Opentrons/opentrons/releases/tag/v'
 
 const UpdateAppBanner = styled(Banner)`
   border: none;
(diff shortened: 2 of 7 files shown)


app/src/assets/localization/en/app_settings.json
```diff
@@ -70,6 +70,7 @@
   "problem_during_update": "This update is taking longer than usual.",
   "prompt": "Always show the prompt to choose calibration block or trash bin",
   "receive_alert": "Receive an alert when an Opentrons software update is available.",
+  "release_notes": "Release notes",
```
koji, line 73: nit
probably add this to `shared.json` would be better since the same text is used in `device_settings.json` too.

mjhuff: Good idea, but I think we might end up changing the exact text that goes down here at some point in the near future depending on the update type. 

If this turns out not to be the case, we can refactor it later 👍

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/UpdateRobotModal.tsx
```diff
@@ -33,6 +35,9 @@ import { useDispatchStartRobotUpdate } from '../../../../redux/robot-update/hook
 import type { State, Dispatch } from '../../../../redux/types'
 import type { RobotSystemType } from '../../../../redux/robot-update/types'
 
+export const RELEASE_NOTES_URL =
+  'https://github.com/Opentrons/opentrons/releases'
+
```
shlokamin: this works and is what we've done up until now, but it's not a great user experience because its really hard to find the correct release from this url. i think it would be worth dynamically rendering a link depending on the update type. 

i.e. for v7.0.2, we could construct out own https://github.com/Opentrons/opentrons/releases/tag/v7.0.2 link from the `getRobotUpdateVersion` selector or something

app/src/organisms/UpdateAppModal/index.tsx
```diff
@@ -56,7 +58,8 @@ const PlaceholderError = ({
     </>
   )
 }
-
+export const RELEASE_NOTES_URL =
+  'https://github.com/Opentrons/opentrons/releases'
```
shlokamin: same thing here about a dynamically generated release notes url for a specific release

shlokamin (approved):
lgtm, i think we should auto generate release note links to specific release pages tho

mjhuff:
> lgtm, i think we should auto generate release note links to specific release pages tho

Would this work for users on alpha/beta channels though? Ex, I noticed https://github.com/Opentrons/opentrons/releases/tag/7.0.2-alpha.1 404s. 

Yeah, agreed that there's probably a more helpful way of handling release notes. 

Something Rob & I discussed yesterday was that any sort of release notes page is largely redundant compared to what is already displayed here.

Just some context - the initial designs for the updated modals had several links - release notes, issue tracking, and changelog links. We got rid of them because it was occupying too much of the screen and the changelog doesn't exist anymore AFAIK. This release notes link is supposed to serve as a catch all, so a savvy user can just navigate around our github and report issues, etc. The reason we called the link "Release Notes" was due to space concerns and starting users off on the release notes page maybe makes some kind of sense.

EDIT: After in person convo, we decided the 404s are fine for alpha/beta builds.

PR #13937, opened 2023-11-07
fix(hardware): only listen to the correct acknowledge messages from CAN

ahiuchingau:
# Overview
We should only add the message to the ack queue if the message index matches, so we don't accidentally report the wrong error, like this:
```
opentrons-api[187176]: Setting gantry load to GantryLoad.HIGH_THROUGHPUT
.... stop request errors .....
opentrons-api[187176]: received error ErrorCode.stop_requested trying to set currents for NodeId.gantry_x
```


sfoster1 (approved):
Lol whoopsie!

PR #13941, opened 2023-11-07
feat(app): final copy and animations for golden tip LPC probe screens

smb2268:
fix RAUT-810, RAUT-811

# Overview
Bring golden tip LPC wizard in line with designs and functionality requirements:

Screen Shot 2023-11-07 at 6 01 36 PM
Screen Shot 2023-11-07 at 5 43 59 PM
Screen Shot 2023-11-07 at 5 41 54 PM
Screen Shot 2023-11-07 at 5 33 00 PM
Screen Shot 2023-11-07 at 5 32 49 PM

# Test Plan
Run through LPC on both OT2 and Flex and ensure copy and assets are correct for golden tip and old LPC

# Changelog
1. Bring copy and assets in line with designs for golden tip LPC
2. Add move to maintenance position command to beginning of AttachProbe and DetachProbe, ensure probe is attached before proceeding with LPC

# Review requests
Sanity check code

# Risk assessment
Low

Screen Shot 2023-11-07 at 5 32 35 PM


app/src/organisms/LabwarePositionCheck/AttachProbe.tsx
```diff
@@ -41,55 +46,116 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
...
+  )
 
   const pipette = protocolData.pipettes.find(p => p.id === pipetteId)
   const pipetteName = pipette?.pipetteName
   const pipetteChannels =
     pipetteName != null ? getPipetteNameSpecs(pipetteName)?.channels ?? 1 : 1
-  const pipetteMount = pipette?.mount
-  if (pipetteName == null || pipetteMount == null) return null
+  let probeVideoSrc = attachProbe1
+  let probeLocation = ''
```
mjhuff, line 61: Maybe we should include a meaningful default value for probeLocation?

mjhuff (commented):
Code looks great! Just one nit. Assuming no one else does, I'll test it out tomorrow AM so we can get it merged 👍

smb2268:
> Code looks great! Just one nit. Assuming no one else does, I'll test it out tomorrow AM so we can get it merged 👍

@mjhuff thank you! I tested this out on a Flex but haven't yet tested on OT2

app/src/organisms/LabwarePositionCheck/AttachProbe.tsx
smb2268: We actually want this to be an empty string for a single channel - the i18n string is "Push the pipette ejector up and press the probe firmly onto the <bold>{{location}}</bold> pipette nozzle as far as it can go." and if the pipette is a single channel we don't specify a location since there is only one nozzle.

mjhuff (approved):
Definitely not blocking and probably nothing you can do about it, but the cal probe on step 5/7 with the align to the top of labware looks pretty low res... I don't know if we have anything better, but we should swap that out at some point.

Looks good though! 🚀 

EDIT: Talked to Sue, the new asset should be up soon!

PR #13942, opened 2023-11-07
fix(api, shared-data, robot-server): stringify all error info in run log

fsinapi:
# Overview

The `GET /runs` endpoint was returning a 500 error in some cases as a result of invalid entries in the robot run log. In the specific case of RQA-1872, the cause is a non-string entry in the `errorInfo` dictionary for an error that failed a run. To get _more_ specific, the issue was that the PipetteOverpressureError exception was being created with an info field with a value field that was an instance of `OT3Mount`.

Before this PR, our enumerated errors would accept a dictionary with string keys and _any_ value type. This PR changes the classes to force the dictionary values to be strings. The `ErrorOccurrence` constructor was not performing any validation on its inputs because it was using the `construct` method, and now it doesn't have to worry about that because any exceptions _have_ to use strings.

In addition, this PR adds a catch for `ValidationErrors` when getting the State Summary from the Run Store. It will return None, which is already tolerated by the higher calling functions in the `GET /runs` chain.

# Test Plan

Fail a run with an overpressure error on a robot with current `edge`, then push this branch, and then make sure `GET /runs` doesn't return a 500 error (there should be an empty state summary for the run that failed). 

# Changelog

# Review requests

# Risk assessment

Pretty low, touches error handling that was already broken and stringifies a bunch of error messages.

### DIFF
--- a/robot-server/robot_server/runs/run_store.py
+++ b/robot-server/robot_server/runs/run_store.py
@@ -1,4 +1,5 @@
 """Runs' on-db store."""
+import logging
 from collections import defaultdict
 from dataclasses import dataclass
 from datetime import datetime
@@ -17,6 +18,8 @@ from robot_server.protocols import ProtocolNotFoundError
 
 from .action_models import RunAction, RunActionType
 from .run_models import RunNotFoundError
+
+log = logging.getLogger(__name__)
 
 _CACHE_ENTRIES = 32
 
@@ -263,6 +266,7 @@ class RunStore:
                 else None
             )
         except ValidationError:
+            log.warn(f"invalid state summary for run ID: {run_id}")
             return None
 
     @lru_cache(maxsize=_CACHE_ENTRIES)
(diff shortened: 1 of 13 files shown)


robot-server/robot_server/runs/run_store.py
```diff
@@ -256,11 +256,14 @@ def get_state_summary(self, run_id: str) -> Optional[StateSummary]:
...
-            if row.state_summary is not None
-            else None
-        )
+        try:
+            return (
+                StateSummary.parse_obj(row.state_summary)
+                if row.state_summary is not None
+                else None
+            )
+        except ValidationError:
```
sfoster1: Let's log something when this happens please

sfoster1 (approved):
Perfect, thanks! Let's log in that except tho

robot-server/robot_server/runs/run_store.py
fsinapi: 🪵

sfoster1 (approved):
Awesome, thanks!

fsinapi:
Tested on Flexington. Confirmed that the route no longer returns an error, and the app is able to get past the initializing screen. The summary of the run _is_ misleading since it returns a fake empty state summary, we should be sure to follow up on that.

PR #13943, opened 2023-11-07
feat(api): enforce only one item in any location

sanni-t:
# Overview

Previously, we had to allow loading multiple items (hopefully of the same item, just multiple times) in any location to accommodate for LPC happening in the same run. Now with LPC happening in separate maintenance runs, we do not have to support this anymore. This PR removes all allowances of loading multiple items in a given location. Note, a 'location' means a very specific location like a particular slot or a module or a labware.

This is required in order to perform robust deck conflict checks without having to account for the impractical case of multiple items loaded in any given location (also makes the code cleaner).

# Test Plan

- Test protocols that load multiple items in the same location and verify that an error is raised
- Start an http run and try loading multiple items in the same location and verify that an error response is received
- Verify that labware stacking or loading labware on modules still completes successfully

# Changelog

- added checks in `loadLabware` and `loadModule` to verify that the specified load location is not occupied
- removed the labware & module ids parsing filter logic that was used to disregard equipment added to the engine during setup
- updated the existing JSON test protocols by removing the load trash commands in protocols. I tested the actual protocols exported by PD from 6.0.0 until the latest one and verified that none of the protocols _load_ the fixed trash explicitly. So unless someone manually edits a JSON protocol to add a load trash command, no protocols in the field should have this.
- updated the tests to reflect the updated the JSON protocols

# Review requests

- Check if I have missed any deck conflict check. I haven't added any checks for addressable areas but should I? They are a bit complicated as in they don't always reflect an equipment in a location..
- any tests I've missed?

# Risk assessment

- This PR adds a small risk of breaking protocols written outside our API & PD recommendations. Both PD & Python API have had checks for preventing loading of multiple equipment in one location, except for a bug that allowed loading multiple labware on a module. So unless a user has hand-edited JSON protocols or wrote python protocols that exploited the mentioned bug in their protocols, protocols in the field should not break. Also for protocols that did exploit this loophole, their runs had chances of getting into unexpected states so this change makes sure that doesn't happen from now on.


api/src/opentrons/protocol_engine/state/modules.py
```diff
@@ -932,7 +928,8 @@ def get_heater_shaker_movement_restrictors(
         return hs_restrictors
 
     def raise_if_module_in_location(
-        self, location: Union[DeckSlotLocation, ModuleLocation]
+        self,
+        location: DeckSlotLocation,
```
sanni-t, line 932: @jbleon95 we're still loading modules in deck slots only, ya?

robot-server/tests/integration/protocols/simple_v6.json
```diff
@@ -1284,15 +1284,6 @@
...
     },
-    {
-      "commandType": "loadLabware",
-      "params": {
-        "labwareId": "fixedTrash",
-        "location": {
-          "slotName": "12"
-        }
-      }
-    },
```
SyntaxColoring, line 1295: Thanks for fixing these fixture inconsistencies.

After these command removals, we're still left with the `"fixedTrash"` entries in the top-level `"labware"` key. Is that intentional? It would not surprise me if that is how JSON protocols actually look, but just making sure.

robot-server/tests/integration/protocols/simpleV8.json
```diff
@@ -1276,20 +1276,6 @@
...
-        "labwareId": "fixedTrash",
-        "location": {
-          "slotName": "12"
-        },
-        "loadName": "opentrons_1_trash_1100ml_fixed",
-        "namespace": "opentrons",
-        "version": 1,
-        "displayName": "Trash"
-      }
-    },
```
SyntaxColoring: Hmm are we intentionally removing the explicit load from schema v8? I thought that because the trash isn't fixed in the general case anymore, protocols would always have to load it explicitly.

api/src/opentrons/protocol_api/core/engine/protocol.py
SyntaxColoring, line 1: This file has a few comments around its calls to `deck_conflict.check()` that we can update now.

```
            # It's important that we don't fetch these IDs from Protocol Engine, and
            # use our own bookkeeping instead. If we fetched these IDs from Protocol
            # Engine, it would have leaked state from Labware Position Check in the
            # same HTTP run.
```

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -486,15 +489,22 @@ def get_ancestor_slot_name(self, labware_id: str) -> DeckSlotName:
...
+                ModuleLocation,
+                OnLabwareLocation,
+                AddressableAreaLocation,
+            ),
+        ):
             self._labware.raise_if_labware_in_location(location)
+        if isinstance(location, DeckSlotLocation):
             self._modules.raise_if_module_in_location(location)
-        return location
+        return cast(_LabwareLocation, location)
```
sanni-t, line 511: Slot locations can also be addressable areas so we will need to cross-check against items loaded in both location types. Something like 'check if an item is in both list of labware on addressable areas as well as labware on slots'. Same for modules.

api/src/opentrons/protocol_engine/state/modules.py
sanni-t: yes that's correct - @jbleon95

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -584,17 +594,13 @@ def get_extra_waypoints(
     def get_slot_item(
```
SyntaxColoring, line 595: Is it now appropriate to delete the `# TODO` above?

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -284,11 +283,9 @@ def raise_if_labware_has_labware_on_top(self, labware_id: str) -> None:
...
-    # TODO(mc, 2022-12-09): enforce data integrity (e.g. one labware per slot)
-    # rather than shunting this work to callers via `allowed_ids`.
-    # This has larger implications and is tied up in splitting LPC out of the protocol run
     def get_by_slot(
-        self, slot_name: DeckSlotName, allowed_ids: Set[str]
+        self,
+        slot_name: DeckSlotName,
     ) -> Optional[LoadedLabware]:
         """Get the labware located in a given slot, if any."""
         loaded_labware = reversed(list(self._state.labware_by_id.values()))
```
SyntaxColoring: Does it now make sense to remove the `reversed(...)` thing?

SyntaxColoring (approved):
Yay! This is a tremendous simplification.

> This PR adds a small risk of breaking protocols written outside our API & PD recommendations. Both PD & Python API have had checks for preventing loading of multiple equipment in one location, except for a bug that allowed loading multiple labware on a module. So unless a user has hand-edited JSON protocols or wrote python protocols that exploited the mentioned bug in their protocols, protocols in the field should not break. Also for protocols that did exploit this loophole, their runs had chances of getting into unexpected states so this change makes sure that doesn't happen from now on.

Yeah. Fortunately, I think we're in a good position now to recommend the real solutions, which are `moveLabware` commands and maintenance runs. We should call this out in the release notes, though.

api/src/opentrons/protocol_engine/state/geometry.py
sanni-t: added as todo for addressing in follow-up

robot-server/tests/integration/protocols/simple_v6.json
sanni-t: Yep, that's intentional and that's how the JSON protocols look like.

robot-server/tests/integration/protocols/simpleV8.json
sanni-t: Good catch! Restored this loadLabware

SyntaxColoring (approved):
Changes since my last review look good.

PR #13944, opened 2023-11-08
fix(shared-data): remove requirement for current entry to TipHandlingConfigurations

caila-marashaj:
The schema has been updated to no longer need a `current` from shared data's `TipHandlingConfigurations`, but the schema and `pipette_definitions` needed a couple of small changes to also allow for that.

I'm working on updating the tests for the schema, but I figure we should get the fix in as soon as possible.

### DIFF
--- a/shared-data/pipette/schemas/2/pipettePropertiesSchema.json
+++ b/shared-data/pipette/schemas/2/pipettePropertiesSchema.json
@@ -61,7 +61,7 @@
     "tipConfigurations": {
       "type": "object",
       "description": "Object containing configurations specific to tip handling",
-      "required": ["current", "speed"],
+      "required": ["speed"],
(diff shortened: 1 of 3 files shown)


shared-data/pipette/schemas/2/pipettePropertiesSchema.json
```diff
@@ -63,7 +63,6 @@
       "description": "Object containing configurations specific to tip handling",
       "required": ["current", "speed"],
```
ahiuchingau: Current needs to be removed from the required list as well

ahiuchingau (approved):
To unblock testing, we should get this merge in asap. But we need to a followup PR that makes sure the definitions are being validated against the schema properly

PR #13945, opened 2023-11-08
refactor(protocol-designer): remove unneeded loadLabware commands fro…

jerader:
…m 8_0_0 migration

# Overview

Now that the PD 7.0.1 hot fix is in and merged back into edge, this PR removes the unneeded labwareId changes to `8_0_0` migration because those changes are happening in the `7_0_0` migration

# Test Plan

just review the code 😄 you can test importing a protocol too to see if it works correctly but we have tje cypress migration test that does that as well.

# Changelog

remove the labwareId migration

# Review requests

n/a

# Risk assessment

low


mjhuff (approved):
🚀

PR #13946, opened 2023-11-08
refactor(components): update deck map and deck configurator styling

koji:
# Overview
update deck map and deck configurator styling
Deck map part will be followed by another pr.

design
https://www.figma.com/file/1ot18My22DALIcjdLl5LJh/Helix-Design-System?node-id=358%3A76193&mode=dev
~~I'm asking about the border-radius size to Mel since the size in the design is 3.46px but we don't have it as a radius size const.~~

There are some same styles in waste chute, trash bin, and staging area. I think we can create a common style and share it with them, but we can do it after 7.1.0 and we can have a rule for this kind of case.

close RAUT-773 partially 

# Test Plan
Go to Device Details page in Desktop app

# Changelog
- add css to each config fixture component 

# Review requests

# Risk assessment
low

### DIFF
--- a/components/src/hardware-sim/DeckConfigurator/EmptyConfigFixture.tsx
+++ b/components/src/hardware-sim/DeckConfigurator/EmptyConfigFixture.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { css } from 'styled-components'
 
 import {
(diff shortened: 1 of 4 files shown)


mjhuff (approved):
Good thinking! lgtm.

PR #13949, opened 2023-11-08
fix(shared-data): Fix the pipette schemas, tests

sfoster1:
The tests that were supposed to ajv the pipette definitions weren't set up properly, and the tests that were supposed to test that the tests were set up properly weren't set up properly, so we weren't testing the schemas so the schemas were wrong. Fix the schema-test setup, the test-setup-schema-test setup, and the schema.

### DIFF
--- a/shared-data/python/tests/pipette/test_validate_schema.py
+++ b/shared-data/python/tests/pipette/test_validate_schema.py
@@ -18,6 +18,7 @@ def test_check_all_models_are_valid() -> None:
         "ninety_six_channel": "96",
         "eight_channel": "multi",
     }
+    assert os.listdir(paths_to_validate), "You have a path wrong"
     for channel_dir in os.listdir(paths_to_validate):
         for model_dir in os.listdir(paths_to_validate / channel_dir):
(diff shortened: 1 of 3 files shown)


ahiuchingau (approved):
Thank you!

ryanthecoder (approved):
lgtm

shared-data/python/tests/pipette/test_validate_schema.py
```diff
@@ -18,6 +18,7 @@ def test_check_all_models_are_valid() -> None:
         "ninety_six_channel": "96",
         "eight_channel": "multi",
     }
+    assert os.listdir(paths_to_validate), "You have a path wrong"
```
Laura-Danielle, line 21: Is this necessary?

Laura-Danielle (approved):
oopsie and lgtm

shared-data/python/tests/pipette/test_validate_schema.py
sfoster1: IMO yes, we do it in the js tests too

PR #13950, opened 2023-11-08
feat(step-generation): wire up moveLabware into waste chute

jerader:
closes RAUT-778

# Overview

Wires up `moveLabware` into waste chute in step-generation and adds a bit of logic for labware listed in the dropdowns and adds an error creator for if there is no gripper.

# Test Plan

Turn on the deck modification feature flag. Create a flex protocol and add a trash bin, a gripper and a waste chute. 

In the deck setup, add a Move Labware step. select the tiprack to move and select the gripper checkbox. Select "Waste Chute in D3" for the new location. Create the step. See that nothing errors. 

Now, go to the file setup (click on the file tab on the left) and delete the gripper. Go back to the deck setup and see that there is a timeline error saying that a gripper is required. 

Go back to file setup and re-add the gripper and see that the timeline error is gone.

now export the protocol. Look at the json file and examine that the `moveLabware` command's `newLocation` says "addressableAreaName: gripperWasteChute"

# Changelog

- add `addressableAreaName` location logic to the protocol-designer & step-generation components
- add logic for `getLabwareLocation` selector to filter out labware that is in the waste chute
- add a new error creator for checking if gripper is required and plug into `moveLabware`
- add test coverage

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/steplist/fieldLevel/index.ts
```diff
@@ -77,6 +86,15 @@ const getLabwareLocation = (
...
   ) {
     return { labwareId: newLocationString }
+  } else if (
+    getIsAddressableAreaLocation(
+      newLocationString,
+      state.additionalEquipmentEntities
+    )
+  ) {
+    return {
+      addressableAreaName: 'gripperWasteChute',
```
jerader, line 96: the `getLabwareLocation` util that this is a part of is specifically for the `newLocation` field in `moveLabware`. That is why the addressableAreaName is always going to be `gripperWasteChute`, since you can't move labware with a pipette or move labware into a trash bin.

protocol-designer/src/steplist/fieldLevel/index.ts
```diff
@@ -77,6 +86,15 @@ const getLabwareLocation = (
     getIsAdapterLocation(newLocationString, state.labwareEntities)
   ) {
     return { labwareId: newLocationString }
+  } else if (
+    getIsAddressableAreaLocation(
+      newLocationString,
+      state.additionalEquipmentEntities
```
shlokamin: `getIsAddressableAreaLocation` isnt just checking if this is an addressable area (a generic rectangular plane), its specifically checking if the labware is an "additional equipment entity" right? it looks like an `AdditionalEquipmentEntity` is typed to only be a `'gripper' | 'wasteChute' | 'stagingArea'`

just wanna make sure we're using the terminology consistently cc @brenthagen

shlokamin (approved):
makes sense to me! just left a naming nit (which i think is important thing to clear up since all of the new deck config lingo is confusing as it is)

protocol-designer/src/steplist/fieldLevel/index.ts
jerader: So the only time the `moveLabware`'s new location is going to be an `addressableAreaName` is if its moving labware into the waste chute, right? Because you can only move labware into a slot, an adapter, on top of a module, off deck, and into a waste chute. 

Maybe I should rename the util to `getIsWasteChuteLocation`. the `additionalEquipmentEntities` type is `AdditionalEquipmentEntities` but should technically be something more specific for only waste chute.

jerader: does that make sense? cc @brenthagen @shlokamin

shlokamin: if we're only checking to see if the labware is being moved into the waste chute then yeah lets rename the function to be that specfic.

more broadly though, maybe this is pointing to an interface issue with `AdditionalEquipmentEntity`, which can be a gripper, waste chute, or staging area. it reasonable to think that you can move labware into a staging area (which has a location), but what does it mean to move a labware into a gripper? what is the gripper's location? 

```js
export interface NormalizedAdditionalEquipmentById {
  [additionalEquipmentId: string]: {
    name: 'gripper' | 'wasteChute' | 'stagingArea'
    id: string
    location?: string
  }
}
```

PR #13951, opened 2023-11-08
refactor(app): remove console log on release notes component

shlokamin:
# Overview

I noticed a bunch of console logs when viewing app + robot release notes in the app. This PR just removes a leftover console log.


mjhuff (approved):
I could never tell if these were important for debugging or not 😅 ...thank you!

PR #13952, opened 2023-11-08
fix(hardware): let can messenger known what nodes are present

ahiuchingau:
# Overview
When we broadcast a CAN message to all the nodes, we are only expecting the basic nodes (X,Y and head) to respond. But now with network info, we do know what devices are present. So we can send over the known (found) nodes to the can messenger whenever we probe the network so our AckListener can actually wait for all present nodes to respond.


sfoster1 (approved):
Looks good to me

PR #13955, opened 2023-11-08
fix(api): fix origin & target for fast home should not include other axes

ahiuchingau:
# Overview
Forgot to include this in my previous PR #13933, without this, we'd accidentally engage other axes while we're homing one.


sfoster1 (approved):
Nice

PR #13956, opened 2023-11-08
fix(app): module calibration should only use calibrated pipette

smb2268:
# Overview
We should not allow users to calibrate modules unless an attached pipette is calibrated. We should ensure a pipette has calibration data before selecting that as the pipette to use for module calibration.

# Test Plan
1. From both run setup and the module card, ensure the calibrate module CTA is disabled if no attached pipette is calibrated
2. Check that if two pipettes are attached but only the right pipette is calibrated, the right pipette is be used for module calibration.

# Changelog
1. Check for pipette calibration data before selecting a pipette to use for module calibration
2. Disable module card overflow menu calibrate button if no pipette has calibration data

# Review requests
Run through the test plan

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/ModuleWizardFlows/index.tsx
+++ b/app/src/organisms/ModuleWizardFlows/index.tsx
@@ -60,7 +60,11 @@ export const ModuleWizardFlows = (
   const isOnDevice = useSelector(getIsOnDevice)
   const { t } = useTranslation('module_wizard_flows')
   const attachedPipettes = useAttachedPipettesFromInstrumentsQuery()
-  const attachedPipette = attachedPipettes.left ?? attachedPipettes.right
+  const attachedPipette =
+    attachedPipettes.left?.data.calibratedOffset?.last_modified != null
+      ? attachedPipettes.left
+      : attachedPipettes.right
+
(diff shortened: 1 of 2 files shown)


app/src/organisms/ModuleWizardFlows/index.tsx
```diff
@@ -193,7 +197,11 @@ export const ModuleWizardFlows = (
         continuePastCommandFailure
       )
   }
-  if (currentStep == null || attachedPipette == null) return null
+  if (
+    currentStep == null ||
+    attachedPipette?.data.calibratedOffset?.last_modified == null
+  )
+    return null
```
koji, line 204: I think the changes make sense, but these lines could be higher like right after `currentStep`(L76) 🤔

smb2268: I can't really move it much higher because you can't have an early return statement before all of the hooks are called and there are quite a few in the first two hundred lines

koji: oh, that totally makes sense. thanks!

koji (approved):
LGTM

PR #13958, opened 2023-11-09
feat(app): add inPlace and moveToAddressableArea command text

ncdiehl11:
closes [RAUT-851](https://opentrons.atlassian.net/browse/RAUT-851)

# Overview

Adds new command text for in place commands and moveToAddressableArea

# Test Plan

- write tests to verify that all above commands return the correct text

# Changelog

- add command text to translation file
- expand PipettingCommandText to include dropTipInPlace, blowOutInPlace, and dispenseInPlace commands
- refactor moveToWell command text to index file (not an instance of PipettingRunTimeCommand)
- add moveToAddressableArea text handling to index file
- write tests

# Risk assessment

low

[RAUT-851]: https://opentrons.atlassian.net/browse/RAUT-851?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shared-data/command/types/pipetting.ts
```diff
@@ -1,17 +1,23 @@
...
-  | PrepareToAspirateRunTimeCommand
   | TouchTipRunTimeCommand
 
+export type PipettingInPlaceRunTimeCommand =
+  | BlowoutInPlaceRunTimeCommand
+  | ConfigureForVolumeRunTimeCommand
+  | DispenseInPlaceRunTimeCommand
+  | DropTipInPlaceRunTimeCommand
+  | PrepareToAspirateRunTimeCommand
+
```
jerader: is splitting this up necessary? Seems like we can just keep it as how we add it before.

app/src/organisms/CommandText/PipettingCommandText.tsx
```diff
@@ -1,15 +1,10 @@
...
-  AspirateRunTimeCommand,
-  DispenseRunTimeCommand,
-  BlowoutRunTimeCommand,
-  MoveToWellRunTimeCommand,
-  DropTipRunTimeCommand,
-  PickUpTipRunTimeCommand,
   getLabwareDefURI,
   RobotType,
 } from '@opentrons/shared-data'
+import type { PipettingRunTimeCommand } from '@opentrons/shared-data'
```
jerader: since this is a type import, it can go last, before `getFinalLabwareLocation`

app/src/organisms/CommandText/PipettingCommandText.tsx
```diff
@@ -38,7 +26,15 @@ export const PipettingCommandText = ({
...
-  const { labwareId, wellName } = command.params
+  let labwareId = ''
+  let wellName = ''
+  if ('labwareId' in command.params) {
+    labwareId = command.params.labwareId
+  }
+  if ('wellName' in command.params) {
+    wellName = command.params.wellName
+  }
+
```
jerader: Since there are only 2 options, these can be replaced with ternarys:

`const labwareId =
    'labwareId' in command.params ? command.params.labwareId : ''`
    
`const wellName = 'wellName' in command.params ? command.params.wellName : ''`

app/src/organisms/CommandText/PipettingCommandText.tsx
```diff
@@ -61,7 +57,7 @@ export const PipettingCommandText = ({
       const { volume, flowRate } = command.params
       return t('aspirate', {
         well_name: wellName,
-        labware: getLabwareName(robotSideAnalysis, labwareId),
+        labware: getLabwareName(robotSideAnalysis, labwareId ?? ''),
```
jerader: I think this extra protection isn't needed since labwareId is already specified to be a string up above 

```suggestion
        labware: getLabwareName(robotSideAnalysis, labwareId),
```

jerader (commented):
looks good! tests look great. Just left a few clean up comments.

shared-data/command/types/pipetting.ts
ncdiehl11: I was talking through this with Jamey. Initially, we were type checking inPlace vs atLocation command types to determine the existence of `wellName` and `labwareId` params. Though we landed on explicitly checking the existence of these params on the command, we felt it was appropriate to leave the 2 pipetting command types separate. I don't feel strongly either way, however.

jerader: okay, that makes sense. I think if we keep it like this, we should change the `PipettingCreateCommand` type to follow the same pattern. Either way works for me though. Lets just make sure its consistent between `PipettingRunTimeCommand` and `PipettingCreateCommand`

jerader (approved):
Lets either undo the change to `PipettingRunTimeCommand` or expand the pattern to `PipettingCreateCommand`. After that, this lgtm!

PR #13959, opened 2023-11-09
feat(shared-data): add 8 moveable trashes!

shlokamin:
# Overview
This PR adds 8 unique addressable areas for the moveable trash so we can satisfy the lemma below:

> Each AA is provided by a fixture mounted to only one cutout id

Closes RSS-397

# Changelog

- Add 8 different movable trash addressable areas

# Risk assessment

Low


SyntaxColoring (approved):
I haven't checked the offset math yet, but the structure of this looks correct. Thanks!

PR #13960, opened 2023-11-09
fix(app): fix step meter tearing by using no animation

mjhuff:
# Overview

ODD optimizations do not prevent step meter tearing on the physical ODD when step progress regresses. After talking with Design, we will play no animation solves the problem until we optimize ODD animations.

### No Progress Animation When Pressing "Go Back"

https://github.com/Opentrons/opentrons/assets/64858653/b00926b5-18c8-440b-bd38-2e71de4d59a1

# Test Plan
- Launch any flow and confirm the behavior shown in the video.

# Risk assessment
low

### DIFF
--- a/app/src/atoms/StepMeter/index.tsx
+++ b/app/src/atoms/StepMeter/index.tsx
@@ -16,13 +16,13 @@ interface StepMeterProps {
 
 export const StepMeter = (props: StepMeterProps): JSX.Element => {
   const { totalSteps, currentStep } = props
+  const prevPercentComplete = React.useRef(0)
   const progress = currentStep != null ? currentStep : 0
-  const percentComplete = `${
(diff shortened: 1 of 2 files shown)


app/src/atoms/StepMeter/index.tsx
```diff
@@ -45,11 +45,15 @@ export const StepMeter = (props: StepMeterProps): JSX.Element => {
     top: 0;
     height: 100%;
     background-color: ${COLORS.blueEnabled};
-    width: ${percentComplete};
+    width: ${percentComplete}%;
```
koji, line 48: 👍

koji (approved):
LGTM
I think we can improve touchscreen's UI stuff after 7.1.0 which will be fun & painful 😁

PR #13961, opened 2023-11-09
feat(shared-data, app): add configureNozzleLayout command type in run log

smb2268:
# Overview
This PR adds support for the `configureNozzleLayout` command in run log

# Test Plan
1. Upload a partial-tip pickup protocol (you can find some here: https://github.com/Opentrons/test-protocols/tree/main/abr_testing/96_channel_partial_tip_protocols)
2. Start setup for the protocol and see that there will be an item in Run Log for the `configureNozzleLayout` command and that pickup tip commands now print a range of wells dependent on how many nozzles are being used at that time

# Changelog
1. Add `configureNozzleLayout` command type and supporting types 
2. Add a string for this command in `CommandText`
3. Update `pickUpTip` command text to use new util `getWellRange` that will return the range of wells that a pipette is going to pick up tips from based on the number of nozzles the pipette is currently using

# Review requests
Look through util and rest of code for sanity and test this out if you can!

# Risk assessment
Low


shared-data/command/types/setup.ts
```diff
@@ -71,13 +71,26 @@ export interface LoadFixtureRunTimeCommand
...
+export interface ConfigureNozzleLayoutCreateCommand
+  extends CommonCommandCreateInfo {
+  commandType: 'configureNozzleLayout'
+  params: ConfigureNozzleLayoutParams
+}
+
+export interface ConfigureNozzleLayoutRunTimeCommand
+  extends CommonCommandRunTimeInfo,
+    ConfigureNozzleLayoutCreateCommand {
+  result?: {}
```
jerader, line 83: maybe add a TODO here to remind us to add a result type?

shared-data/command/types/setup.ts
```diff
@@ -71,13 +71,26 @@ export interface LoadFixtureRunTimeCommand
...
+}
+
 export type SetupRunTimeCommand =
   | LoadPipetteRunTimeCommand
   | LoadLabwareRunTimeCommand
   | LoadFixtureRunTimeCommand
   | LoadModuleRunTimeCommand
   | LoadLiquidRunTimeCommand
   | MoveLabwareRunTimeCommand
+  | ConfigureNozzleLayoutRunTimeCommand
```
jerader: nit, no big deal but can we alphabetize this 😅

shared-data/command/types/setup.ts
```diff
@@ -161,3 +175,26 @@ interface LoadFixtureParams {
...
+
+interface NozzleConfigurationParams {
+  primary_nozzle: string
+  style: NozzleConfigurationStyle
+}
+
+interface ConfigureNozzleLayoutParams {
+  pipetteId: string
+  configuration_params: NozzleConfigurationParams
+}
```
jerader, line 200: Oh wait i didn't realize a `pipetteId` is needed in the params, I was looking at the implementation section on the proposal: https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3732602905?atlOrigin=eyJpIjoiMTQxNjBkYTBkYzhjNDM0YzgzMGY2YmEwMWZlOGY4NTMiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ

smb2268: I'm pretty sure it's necessary based on this https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/protocol_engine/commands/configure_nozzle_layout.py#L33 - @sanni-t can you confirm whether the types I have here are correct?

shared-data/command/types/setup.ts
smb2268: I was looking at the analysis output and the result was just an empty object - @sanni-t is that correct?

shared-data/command/types/setup.ts
```diff
@@ -161,3 +175,26 @@ interface LoadFixtureParams {
...
+
+export type NozzleConfigurationStyle =
+  | typeof COLUMN
+  | typeof SINGLE
+  | typeof ROW
+  | typeof QUADRANT
+  | typeof EMPTY
+
+interface NozzleConfigurationParams {
+  primary_nozzle: string
```
jerader, line 193: we probably want both `primary_nozzle` and `configuration_params` to be camelCase, like what we do for `pipetteId`

jerader (approved):
lgtm, just pending the type questions for Sanniti. I left a comment about camelCasing a few things.

shared-data/command/types/setup.ts
sanni-t: Yes, that's correct. The result is an empty object.

shared-data/command/types/setup.ts
sanni-t: Yep, `pipette_id` is required for knowing which pipette we are setting to partial nozzle configuration.

shared-data/command/types/setup.ts
smb2268: I think this would have to be a BE change, wouldn't it?

PR #13962, opened 2023-11-09
chore: Separate lt and ht pickup and dropoff

sfoster1:
This PR looks gigantic but most of it is little fiddly fixups to pipette config files that are schema checked and don't change data.

Our 96 channel pipettes pick up full racks of tips using a cam grabber, but they're also capable of picking up less tips than that through the normal press-fit routine that the low throughput pipettes use. Those routines need different configuration values that do not overlap, and we previously had them overlapping.

This commit
- Changes the pipette config general schema (this is a good place to start when reviewing) to fully separate the two styles of pickup (press fit and cam action) and the two styles of drop tip (plunger eject and cam action). Low throughput pipettes do press fit and plunger eject; high throughput pipettes do both kinds of pickup, and cam action drop. Current configuration is folded in to the two new locations.
- Changes all the other code to adapt to that
- Changes the pipette geometry schema to add the rows and columns in which the nozzles are arranged so we can do partial tip pickup calculations based on it
- Changes the partial tip pickup stuff to work based on symbolic geometry instead of points

## Review requests
- Check that no data changed except in https://github.com/Opentrons/opentrons/commit/a9ec6395e4fa259180d40863c56514625aab37da
- Do the new configs make a little more sense

## Testing
- [x] Pick up some tips!


shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -150,31 +150,89 @@ class PlungerHomingConfigurations(BaseModel):
     )
 
 
-class TipHandlingConfigurations(BaseModel):
+class PressFitPickUpTipConfiguration(BaseModel):
     presses: int = Field(
-        default=0.0, description="The number of tries required to force pick up a tip."
```
Laura-Danielle, line 155: Can we keep the previous descriptions for these fields? They make more sense imo.

shared-data/pipette/schemas/2/pipettePropertiesSchema.json
```diff
@@ -162,42 +158,124 @@
...
-      "type": "object",
       "description": "Object containing configurations for picking up tips common to all partial configurations",
-      "required": ["speed"],
-      "properties": {
-        "presses": { "$ref": "#/definitions/positiveNumber" },
-        "speed": { "$ref": "#/definitions/positiveNumber" },
-        "increment": { "$ref": "#/definitions/positiveNumber" },
-        "distance": { "$ref": "#/definitions/positiveNumber" }
-      }
+      "anyOf": [
```
sanni-t, line 162: Why `anyof` and not `oneof`?

Laura-Danielle (approved):
Code in general looks good, just had a comment about some updated field descriptions.

I think also we'll want to do some extra testing on mutable configurations since there are a lot of changes there.

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_5.json
```diff
@@ -44,22 +70,7 @@
...
-      "6": 0.09,
-      "7": 0.11,
-      "8": 0.13,
-      "12": 0.19,
-      "16": 0.25,
-      "24": 0.38,
-      "48": 0.75,
-      "96": 1.5
-    }
+    "availableConfigurations": [1, 8, 12, 16, 24, 48, 96]
```
sanni-t, line 73: Interesting, why are the other versions limited to [1, 8, 12, 96] configurations while this one has all available? Is there something in the old versions that limits their use of other partial configurations?

ahiuchingau (approved):
THIS IS FANTASTIC! THANK YOU!!

What does cam stand for?

sanni-t (approved):
Changes make sense to me. Just a couple of questions.
Sounds like this will need thorough testing of all pipettes? (even OT2) Let me know if you or @caila-marashaj need any help with that. The 96 channel partial tip protocols we've tested so far are [here](https://github.com/Opentrons/test-protocols/tree/main/abr_testing/96_channel_partial_tip_protocols)

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_5.json
Laura-Danielle: Human error 😆

shared-data/pipette/schemas/2/pipettePropertiesSchema.json
sfoster1: Because you're allowed to have both! The 96, for instance, has both: it needs to be able to do press-fit tip pickup for partial tip configurations, and needs to be able to do cam-action tip pickup for full configurations.

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_5.json
sfoster1: @Laura-Danielle Human error like "3.5 should also only have [1, 8, 12, 96] available" or human error like "all versions should have 24 and 48"?

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
sfoster1: That's not what that is, though. You're going incrementally deeper every time, it's not just a retry

shared-data/pipette/schemas/2/pipettePropertiesSchema.json
sanni-t: Yea, I totally had a brain fart moment

api/src/opentrons/hardware_control/nozzle_manager.py
```diff
@@ -26,60 +48,81 @@ class NozzleConfigurationType(Enum):
...
+                next(iter(current_cols.values()))[-1],
+            )
+        )
+        common_corners = physical_corner_nozzles.intersection(current_corner_nozzles)
+        if len(common_corners) == 1:
+            if len(current_rows) == len(physical_rows) // 2:
+                if len(current_cols) == len(physical_cols) // 2:
+                    return NozzleConfigurationType.QUADRANT
+
+        return NozzleConfigurationType.SUBRECT
```
Laura-Danielle, line 75: Can we just remove `QUADRANT` in favor of `SUBRECT` since a quadrant is a a subrectangle anyways? I can't see a need to distinguish between the two.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -359,6 +359,16 @@ class Config:
         }
 
 
+class PipetteRowDefinition(BaseModel):
```
Laura-Danielle, line 367: Can we just call this `PipetteRowColumnDefinition` or something to that effect and get rid of the separate definitions for the two? They have the exact same fields.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -369,6 +379,8 @@ class PipetteGeometryDefinition(BaseModel):
         alias="pathTo3D",
     )
     nozzle_map: Dict[str, List[float]] = Field(..., alias="nozzleMap")
+    ordered_columns: List[PipetteColumnDefinition] = Field(..., alias="orderedColumns")
+    ordered_rows: List[PipetteRowDefinition] = Field(..., alias="orderedRows")
```
Laura-Danielle, line 400: We should probably add validation on these two fields as well similar to `nozzle_map`.

api/src/opentrons/hardware_control/nozzle_manager.py
```diff
@@ -95,13 +138,8 @@ def xy_center_offset(self) -> Point:
...
         # front left-most nozzle of the 96 channel in a given configuration
         # and front nozzle of the 8 channel
-        if self.starting_nozzle == self.front_right:
-            return self.map_store[self.front_right]
-        map_store_list = list(self.map_store.values())
-        starting_idx = map_store_list.index(self.map_store[self.back_left])
-        difference = self.map_store[self.back_left] - self.map_store[self.front_right]
-        y_rows_length = int(difference[1] // INTERNOZZLE_SPACING)
-        return map_store_list[starting_idx + y_rows_length]
+        front_left = next(iter(self.columns.values()))[-1]
```
Laura-Danielle, line 148: This technically is only the 'front left' of the 8 channel. It would give an incorrect value if the 96 channel was in any configuration that did not include the leftmost nozzle. Should maybe just determine this in the `build` function and save.

api/src/opentrons/hardware_control/nozzle_manager.py
```diff
@@ -110,87 +148,123 @@ def tip_count(self) -> int:
...
-                    "requested_front_right_nozzle": front_right_nozzle,
-                },
-            )
-
-        if not default_configuration.map_store.get(front_right_nozzle):
+                message=f"No entry for front right nozzle {e} in pipette",
+                wrapping=[PythonException(e)],
+            ) from e
+
+        def _rows_in_map() -> Iterator[Tuple[str, List[str]]]:
```
Laura-Danielle: imo I think this would be cleaner if we just took the first/last index from front/back row/column and then used list splicing.

Laura-Danielle: less lines of code too :D

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -150,31 +150,89 @@ class PlungerHomingConfigurations(BaseModel):
...
     presses: int = Field(
-        default=0.0, description="The number of tries required to force pick up a tip."
+        ...,
+        description="The number of times to press (incrementally more each time by increment)",
+    )
+    increment: float = Field(
+        ..., description="The incremental amount to press more by each press"
+    )
+    distance: float = Field(
+        ..., description="The distance to press on the first stroke"
```
Laura-Danielle:
```suggestion
        ..., description="The starting distance to begin a pick up tip from"
```

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -150,31 +150,89 @@ class PlungerHomingConfigurations(BaseModel):
...
 
-class TipHandlingConfigurations(BaseModel):
+class PressFitPickUpTipConfiguration(BaseModel):
     presses: int = Field(
-        default=0.0, description="The number of tries required to force pick up a tip."
+        ...,
+        description="The number of times to press (incrementally more each time by increment)",
+    )
+    increment: float = Field(
+        ..., description="The incremental amount to press more by each press"
```
Laura-Danielle:
```suggestion
        ..., description="The increment to move the pipette down on each force tip pickup press"
```

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -150,31 +150,89 @@ class PlungerHomingConfigurations(BaseModel):
     )
 
 
-class TipHandlingConfigurations(BaseModel):
+class PressFitPickUpTipConfiguration(BaseModel):
     presses: int = Field(
-        default=0.0, description="The number of tries required to force pick up a tip."
+        ...,
+        description="The number of times to press (incrementally more each time by increment)",
```
Laura-Danielle:
```suggestion
        description="The number of times to force pickup (incrementally more each time by increment)",
```

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -150,31 +150,89 @@ class PlungerHomingConfigurations(BaseModel):
...
+    )
+    increment: float = Field(
+        ..., description="The incremental amount to press more by each press"
+    )
+    distance: float = Field(
+        ..., description="The distance to press on the first stroke"
+    )
+    speed: float = Field(..., description="How fast to press each stroke")
```
Laura-Danielle:
```suggestion
    speed: float = Field(..., description="The speed to move the Z axis for each force pickup")
```

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
Laura-Danielle: Added suggestions below.

Laura-Danielle (commented):
Will not request changes as to ensure I don't block this PR from being merged. However, I do have a few comments that I think we should address before merging.

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_5.json
Laura-Danielle: Human error as in I forgot about 16, 24, 48 as tip options. To be fair though, we don't really use this config as much as I thought we would.

sfoster1: will add them

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: We allow actual quadrants but not arbitrary subrects though, right? I feel like it's worth being able to distinguish for that reason

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
sfoster1: They have the same fields but those fields do not have the same types in the json, and the only reason they have the same type here is because our version of pydantic isn't capable of doing fine enough divisions. I actually wrote a whole validator for this before realizing I did it based on future pydantic docs that we can't use, and as soon as we update I'll be adding them again. Specifically, the row names and column names I'd like to be separate and verified independently.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
sfoster1: I want to but I want to do it inside the submodules and I want to wait until we're on a higher version of pydantic because it will be 99 times easier

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: So actually, and I think that maybe I should clarify this more, in an individual `NozzleMap` instance the `map_store`, `rows`, and `columns` actually only refer to the tips that are valid in the map. The first element in columns will _always_ be the left-most column of this specific partial tip configuration, therefore the last element of that first column will always be the frontest, leftest channel. This is verified in the `test_*_pipette_map_entries` tests in `test_nozzle_manager`.

api/src/opentrons/hardware_control/nozzle_manager.py
Laura-Danielle: No I've been referring to quadrant as representing arbitrary sub-rectangles. We're never actually verifying (until here I guess) that it is truly a quadrant, but I don't think we really _need_ to do that. I'm currently working on boundary box checks again -- finally -- and I can't see any need to distinguish between the two.

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: Yes, but that would require (if I'm understanding right) parsing the row and column from the name of the well, which is something I tried to avoid as much as I could in favor of keeping the well names opaque atomic artifacts, which is an approach that feels more robust. For instance, if we wanted to, there would be nothing preventing us from naming a tip location just `1` or for that matter `steve` in this implementation - they're fully arbitrary. 

The iteration process also keeps a robustness element of handling arbitrarily and possibly irregular pipettes - because we're iterating linearly, we don't actually care how many things we emit, and that's another thing that isn't really necessary but that I like a lot about this.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
Laura-Danielle: I mean they have the same basic type with some pattern validation in JSON. If you want to split them off here already, then I'd add the validation checks in these pydantic models.

api/src/opentrons/hardware_control/nozzle_manager.py
Laura-Danielle: yeah the frontest/leftest channel of _that_ configuration, not of the whole pipette -- whereas previously it would only ever refer to the front left-most? (H1) of the pipette.

I can't think of an exact use-case where it could be a problem to change that behavior, but something to look out for/be aware of in case there is weird behavior somewhere. I would also change the comment inside the property to be a little more explicit about what's happening.

api/src/opentrons/hardware_control/nozzle_manager.py
Laura-Danielle: No you shouldn't need to.

You could just return the back_left_row/back_left_column as a tuple of (name, contents) rather than just name. Then you could do:

```
list(ordered_dict.items()).index((name, contents))
```

As long as that matches the internals then you will get the correct index. I think this should also take care of irregular pipette patterns, unless I don't understand what types of irregularities you have in mind.

shared-data/pipette/definitions/2/general/eight_channel/p1000/3_5.json
```diff
@@ -4,14 +4,28 @@
...
-    "increment": 0.0,
-    "distance": 13.0
+    "pressFit": {
+      "presses": 1,
+      "speed": 10,
+      "increment": 0.0,
+      "distance": 13.0,
+      "currentByTipCount": {
+        "1": 0.2,
+        "2": 0.14,
```
ryanthecoder, line 14: Is this one meant to be lower than everything else

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: Okay!

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
sfoster1: They do not have the same pattern validation. The column names look for `/[0-9]+/` and the row names look for `/[A-Z]+/`. I'll add the validation

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: Sure. I think the idea is that if you want the front-left most of the pipette you would get it from the physical nozzlemap instance that the manager keeps around.

shared-data/pipette/definitions/2/general/eight_channel/p1000/3_5.json
sfoster1: Yeah, these are sort of known-to-be-messed-up first pass numbers

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
sfoster1: I think we should bias to the side of keeping more complex constraints in the jsonschema, though. It's just so much better at it.

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: Okay. It wouldn't handle the kind of irregularity where you have rows with different numbers of columns though. Which we don't have.

sfoster1: Oh ok I get it. Yeah I can do that. It won't support irregular format pipettes, like if rows had different number of columns, but we won't have those anyway probably.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
sfoster1: ended up doing it

api/src/opentrons/hardware_control/nozzle_manager.py
sfoster1: @Laura-Danielle how does https://github.com/Opentrons/opentrons/pull/13962/commits/39fe1515e6bbf119d520d88efc93b26f1ab3e66a look

Laura-Danielle (approved):
nice, lgtm!

PR #13963, opened 2023-11-09
refactor(api, robot-server): move feature flag file access outside of hardware_control

fsinapi:
# Overview

As a part of the long-term goal of the hardware controller process split, we need to address simultaneous access & modification of the feature flags file on a robot. This PR takes out any parts of the `hardware_control` submodule in the `opentrons` package that directly access the feature flags file, replacing them with a dataclass that just gets populated by the caller of the factory functions. This gets used to dictate whether the pipette handlers use the new or old aspiration functions, as well as a few Flex-specific hardware functions.

# Test Plan

To verify that the settings are being updated in the API correctly, I pushed on a robot, used HTTP to change the advanced setting requiring the estop, and disconnected the estop to ensure that no error popped up.

CI tests should cover that the settings are still being set/read the right way.

Ran the REPL on a robot and changed some settings in `feature_flags.json` to ensure they are reflected in the constructed API.

# Changelog

- Updated OT2 and Flex hardware feature flags to be passed into the API instance on init
- Modified OT2 and Flex pipette handler to not read from feature flag file
- Added function to API protocol to allow reconfiguration of feature flags

# Review requests

I don't like this interface but I don't see an elegant alternative. Injecting into the robot configs seems inappropriate imo. Any strong opinion on an alternative approach is appreciated.

# Risk assessment

Medium, mostly if I missed any hardware API initializers somewhere in the repo.


api/src/opentrons/hardware_control/api.py
```diff
@@ -111,6 +112,7 @@ def __init__(
         backend: Union[Controller, Simulator],
         loop: asyncio.AbstractEventLoop,
         config: RobotConfig,
+        feature_flags: HardwareFeatureFlags = HardwareFeatureFlags(),
```
sfoster1: this probably wants to be `Optional[HardwareFeatureFlags] = None` because this code will create _one_ default `HardwareFeatureFlags()` at function parse time that will be reused lol

api/src/opentrons/hardware_control/api.py
```diff
@@ -170,6 +173,7 @@ async def build_hardware_controller(
         port: Optional[str] = None,
         loop: Optional[asyncio.AbstractEventLoop] = None,
         firmware: Optional[Tuple[pathlib.Path, str]] = None,
+        feature_flags: HardwareFeatureFlags = HardwareFeatureFlags(),
```
sfoster1: ditto here

api/src/opentrons/hardware_control/api.py
```diff
@@ -249,6 +258,7 @@ async def build_hardware_simulator(
         config: Optional[Union[RobotConfig, OT3Config]] = None,
         loop: Optional[asyncio.AbstractEventLoop] = None,
         strict_attached_instruments: bool = True,
+        feature_flags: HardwareFeatureFlags = HardwareFeatureFlags(),
```
sfoster1: ditto here

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -241,7 +242,11 @@ class OT3Controller:
 
     @classmethod
     async def build(
-        cls, config: OT3Config, use_usb_bus: bool = False, check_updates: bool = True
+        cls,
+        config: OT3Config,
+        use_usb_bus: bool = False,
+        check_updates: bool = True,
+        feature_flags: HardwareFeatureFlags = HardwareFeatureFlags(),
```
sfoster1: ditto here

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -274,6 +283,7 @@ def __init__(
         usb_driver: Optional[SerialUsbDriver] = None,
         eeprom_driver: Optional[EEPROMDriver] = None,
         check_updates: bool = True,
+        feature_flags: HardwareFeatureFlags = HardwareFeatureFlags(),
```
sfoster1: ditto here

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1613,6 +1633,16 @@ async def update_config(self, **kwargs: Any) -> None:
...
         self._config = replace(self._config, **kwargs)
 
+    @property
+    def hardware_feature_flags(self) -> HardwareFeatureFlags:
+        return self._feature_flags
+
+    @hardware_feature_flags.setter
+    def hardware_feature_flags(self, feature_flags: HardwareFeatureFlags) -> None:
+        self._feature_flags = feature_flags
+        if isinstance(self._backend, OT3Controller):
```
sfoster1: let's just do it for the simulator too

sfoster1 (changes requested):
I think unfortunately we need to not build function argument defaults like this but other than that looks great!

sfoster1 (approved):
Looks good though I feel like we might want a note about the no-arguments constructor wanting to be where default-true arguments are or something

PR #13965, opened 2023-11-13
feat(protocol-designer): deck v4 touchups and staging area support

jerader:
closes RAUT-787, RAUT-805, RAUT-806, RAUT-850

# Overview

This PR wires up a few v4 deck migration things and then wires up the staging area slot support (4th column is now selectable on the deck!)

# Test Plan

Make sure your deck modification FF is turned on! 

- create a Flex protocol and select to add a staging area slot
- in the `additional items` section, try to add modules/staging area slots/waste chute and see that things look correct and the slot conflict banner properly shows up 
- move to the deck map and see that the staging area slot renders as expected on the deck map. Hover your mouse over the 4th column where there is a slot and see that the hover works as expected, allowing you to add labware on that slot
- add a labware to the deck
- create a `move labware` form step and move the labware on the deck to the 4th column slot
- now create a `transfer` form step and see that the labware moved to the 4th column is on the aspirate/dispense dropdown. when you create the step, see that the transfer errors because you cannot pipette into a labware on the 4th column
- delete the transfer step and export the protocol
- open the json file and find the `moveLabware` `commandType` command. See that the labware location says that the "slotName" is a column 4 slot
- now import that protocol back into Pd. See that the staging area slot is added to the correct location on the deck map

# Changelog

- various v4 touchups among the edit modules/waste chute/staging area slot components
- fix logic in `getSlotEmpty` util 
- in `deckSetup`, add slot id for labware on top of a module and change type
- fix other utils that rely on slots such as `getUnusedStagingAreaSlot`, `getTrashSlot`, `getSlotEmpty`, `getUnoccupiedSlotOptions`
- extend logic in new location and labware dropdowns to support staging area slots
- remove the `loadFixture` command information from `fileCreator` and the `load_file` reducer and replace with logic to look for the waste chute and staging area slot commands
- create a new `errorCreator` for if a user is trying to pipette into a 4th column slot in step-generation and plug into pipetting atomic commands + add a bit of test coverage

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/DeckSetup/LabwareOverlays/SlotControls.tsx
```diff
@@ -42,7 +42,7 @@ interface DNDP {
 interface OP {
   slotPosition: CoordinateTuple | null
   slotBoundingBox: Dimensions
-  slotId: AddressableArea['id']
+  slotId: string
```
jerader, line 45: slotId can either be an `AddresableAreaName` or a `moduleId` (for labware nested on the modules)

b-cooper: oh good catch, could you add that to the docstring or a comment?

shared-data/js/helpers/index.ts
```diff
@@ -286,6 +282,9 @@ export const getAreSlotsVerticallyAdjacent = (
 
   return areSlotsVerticallyAdjacent
 }
+//  TODO(jr, 11/12/23): rename this utility to mention that it
+//  is only used in the OT-2, same with getAreSlotsHorizontallyAdjacent
+//  and getAreSlotsVerticallyAdjacent
```
jerader, line 287: made a ticket for this: https://opentrons.atlassian.net/browse/RAUT-858

components/src/hardware-sim/BaseDeck/StagingAreaFixture.tsx
```diff
@@ -5,7 +5,11 @@ import { SlotClip } from './SlotClip'
 
 import type { DeckDefinition, ModuleType } from '@opentrons/shared-data'
 
-export type StagingAreaLocation = 'A3' | 'B3' | 'C3' | 'D3'
+export type StagingAreaLocation =
+  | 'cutoutA3'
+  | 'cutoutB3'
+  | 'cutoutC3'
+  | 'cutoutD3'
```
brenthagen, line 12: 💯

protocol-designer/src/components/DeckSetup/FlexModuleTag.tsx
```diff
@@ -18,7 +18,7 @@ export function FlexModuleTag(props: FlexModuleTagProps): JSX.Element {
   return (
     <RobotCoordsForeignDiv
       width={dimensions.labwareInterfaceXDimension}
-      height={16}
+      height={20}
```
b-cooper, line 21: just making sure this was an intentional change

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -613,6 +630,9 @@ export const DeckSetup = (): JSX.Element => {
                 robotType={robotType}
                 activeDeckSetup={activeDeckSetup}
                 selectedTerminalItemId={selectedTerminalItemId}
+                stagingAddressableAreas={stagingAreaFixtures.map(
+                  areas => areas.location as AddressableAreaName
```
b-cooper: is there a reason we need to cast the type here?

protocol-designer/src/components/DeckSetup/FlexModuleTag.tsx
jerader: the tags were too low after the v4 change and were interfering with the entity in the slot below it. So I changed the height slightly

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -83,9 +83,8 @@ export const getTrashBinOptionDisabled = (values: FormState): boolean => {
 export const getTrashSlot = (values: FormState): string => {
   const stagingAreaLocations = values.additionalEquipment
     .filter(equipment => equipment.includes('stagingArea'))
-    .map(stagingArea => stagingArea.split('_')[1])
-
-  //   return default trash slot A3 if staging area is not on slot
+    .map(stagingArea => stagingArea.split('_cutout')[1])
```
b-cooper: it would be great to avoid string splitting here. We should be using the deck def to relate addressable areas with cutouts

step-generation/src/constants.ts
```diff
@@ -20,3 +20,5 @@ export const FIXED_TRASH_ID: 'fixedTrash' = 'fixedTrash'
 
 export const OT_2_TRASH_DEF_URI = 'opentrons/opentrons_1_trash_1100ml_fixed/1'
 export const FLEX_TRASH_DEF_URI = 'opentrons/opentrons_1_trash_3200ml_fixed/1'
+
+export const COLUMN_4_SLOTS = ['A4', 'B4', 'C4', 'D4']
```
brenthagen, line 24: `FLEX_STAGING_AREA_SLOT_ADDRESSABLE_AREAS` is defined in shared-data constants, could be used in place of this

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1328,58 +1327,91 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
...
-          command.commandType === 'loadFixture'
+      const hasWasteChuteCommands = Object.values(file.commands).some(
+        command =>
+          (command.commandType === 'moveToAddressableArea' &&
+            command.params.addressableAreaName.includes(
+              WASTE_CHUTE_ADDRESSABLE_AREAS
+            )) ||
+          (command.commandType === 'moveLabware' &&
+            command.params.newLocation !== 'offDeck' &&
+            'addressableAreaName' in command.params.newLocation)
```
b-cooper: shouldn't this right side of the `||` include the waste chute addressable area somehow?

protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1328,58 +1327,91 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
...
+              command
+            ): command is MoveLabwareCreateCommand | LoadLabwareCreateCommand =>
+              command.commandType === commandType &&
+              command.params[locationKey] !== 'offDeck' &&
+              'slotName' in command.params[locationKey] &&
+              COLUMN_4_SLOTS.includes(command.params[locationKey].slotName)
+          )
+          .map(
+            command =>
+              //  @ts-expect-error: ts can't determine the type for location and newLocation
```
b-cooper: :/ is there a way to avoid opting out of type checking here?

protocol-designer/src/step-forms/reducers/index.ts
jerader: good call, I didn't future proof this. i was thinking that currently, you can only move labware into a waste chute and not a trash bin.

protocol-designer/src/step-forms/reducers/index.ts
jerader: yeah i'm trying to figure out how to write this to not get the TS error 🙃 . The `filter()` prior to the `map()` makes sure that `location` or `newLocation` exist in the command, so i'm not sure why ts can't tell that

step-generation/src/constants.ts
jerader: Thanks Brent, I am having some issues with TS when i use `FLEX_STAGING_AREA_SLOT_ADDRESSABLE_AREAS`, I'm going to leave updating this in a followup.

jerader:
For a follow up, I will:
1. use `FLEX_STAGING_AREA_SLOT_ADDRESSABLE_AREAS` instead of `COLUMN_4_SLOTS`
2. pull out shared logic in `getStagingAreaSlotsCutouts` and `getUnocuppiedLabwareLocationOptions` into a util

https://opentrons.atlassian.net/browse/RAUT-860

b-cooper (approved):
🌳

PR #13966, opened 2023-11-13
chore(app-shell): Fix internal release updates

sfoster1:
Since ac0304a5c61462c2c31174308f88eed5404cfda8 we were not properly setting the folder-part of the app update flow in internal-releases which meant the app couldn't update.

## Testing
- Download the built app and see if the logs say it's looking at the right place

Closes RQA-1882


vegano1 (approved):

mjhuff (approved):
Thanks for fixing this!

PR #13969, opened 2023-11-13
feat(components): render 4th column slot label for staging area slots

ncdiehl11:
closes [RAUT-782](https://opentrons.atlassian.net/browse/RAUT-782)

# Overview

Creates a 4th column in SlotLabels component to be rendered if used for coordinates for slot positions AND at least one staging area fixture exists in deck config. Note that SlotLabels do not render 4th column when rendered as coordinates for fixture positions.

# Test Plan

- Verify that 4th column renders in BaseDeck instances where DeckConfiguration includes >1 staging area slot
- NOTE: further testing blocked by inability to test a protocol including a staging area slot
Screen Shot 2023-11-16 at 1 13 23 PM
Screen Shot 2023-11-16 at 1 15 10 PM

# Changelog

- add 4th column label to SlotLabels (renders conditionally on cases specified above)
- determine presence of staging areas in BaseDeck component from deckConfig, and render 4th column slot label based on this condition being true
- ensure slot labels only render columns 1-3 for DeckLocationSelect

# Review requests
@brenthagen 

# Risk assessment
low

[RAUT-782]: https://opentrons.atlassian.net/browse/RAUT-782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jerader (approved):
nice this makes sense to me!

components/src/hardware-sim/Deck/SlotLabels.tsx
```diff
@@ -19,7 +20,10 @@ interface SlotLabelsProps {
 export const SlotLabels = ({
   robotType,
   color,
+  show4thColumn = false,
 }: SlotLabelsProps): JSX.Element | null => {
+  const widthColumnLabelsRem = 30.375
```
jerader: where did this number come from?

components/src/hardware-sim/Deck/SlotLabels.tsx
```diff
@@ -71,14 +75,20 @@ export const SlotLabels = ({
       </RobotCoordsForeignObject>
       <RobotCoordsForeignObject
         height="2.5rem"
-        width="30.375rem"
+        width={`${
+          show4thColumn ? (widthColumnLabelsRem * 4) / 3 : widthColumnLabelsRem
```
jerader: i'm having a hard time following the math 😬 so `widthColumn` * `4` because there are 4 columns. Why do you have to divide by 3?

components/src/hardware-sim/Deck/SlotLabels.tsx
jerader: oh nevermind, that was the original width.

components/src/hardware-sim/Deck/SlotLabels.tsx
ncdiehl11: We're scaling by 4/3 if there's a 4th column label. widthColumnLabelsRem refers to the default width for 3 columns (I can rename the constant to make it clearer?)

ncdiehl11: or something more intuitive like `widthPerColumnLabel`, and scale that unit width by 3 or 4 depending on how many slot labels we should render

jerader: yeah lets rename it! `widthPerColumnLabel` makes sense to me. thanks for the explanation.

jerader (approved):
Let's rename the const but then this lgtm!

components/src/hardware-sim/Deck/SlotLabels.tsx
```diff
@@ -19,7 +20,10 @@ interface SlotLabelsProps {
 export const SlotLabels = ({
   robotType,
   color,
+  show4thColumn = false,
 }: SlotLabelsProps): JSX.Element | null => {
+  const widthPerColumnLabelRem = 10.125
```
brenthagen: instead of having equal column widths for flex children, i think we want to define individual label column widths that match the width of the fixture/slot provided.

it should work to use two different widths, a wider "fixture width" and narrower "slot width". column 1 will always be "fixture width" and column 2 wlll always be "slot width". column 4, when it exists, will be "slot width". and column 3 will be one or the other based on the existence of column 4.

see design system screen shot of the label positioning:

Screen Shot 2023-11-15 at 11 54 10 AM
https://www.figma.com/file/1ot18My22DALIcjdLl5LJh/Helix-Design-System?type=design&node-id=358-31899&mode=design&t=uhpVrtS4ZnlyGJuH-4

components/src/hardware-sim/Deck/SlotLabels.tsx
```diff
@@ -71,14 +75,22 @@ export const SlotLabels = ({
       </RobotCoordsForeignObject>
       <RobotCoordsForeignObject
         height="2.5rem"
-        width="30.375rem"
+        width={`${
+          show4thColumn
+            ? widthPerColumnLabelRem * 4
+            : widthPerColumnLabelRem * 3
+        }rem`}
         x="-15"
```
brenthagen: and this x position will need to adjust left to align with the left edge of the deck map SVG, i think `x="-100"` works

components/src/hardware-sim/Deck/SlotLabels.tsx
ncdiehl11: this makes sense. I mistakenly thought the widths were fixed but will add individual widths now

components/src/hardware-sim/BaseDeck/BaseDeck.tsx
```diff
@@ -260,6 +260,13 @@ export function BaseDeck(props: BaseDeckProps): JSX.Element {
             ) : null
           }
         )}
+        {showSlotLabels ? (
+          <SlotLabels
+            robotType={robotType}
+            color={darkFill}
+            show4thColumn={stagingAreaFixtures.length > 0}
```
brenthagen: `SlotLabels` was moved to line 123 in above in edge - this prop needs to be added there and this instance of `SlotLabels` removed. also, with changes to edge adding waste chute staging area fixtures, the 4th column condition should be 
```              
              show4thColumn={
                stagingAreaFixtures.length > 0 ||
                wasteChuteStagingAreaFixtures.length > 0
              }
```

brenthagen (approved):
4️⃣

PR #13970, opened 2023-11-13
ci(app): deploy app artifacts to build number prefixed folder

shlokamin:
# Overview

**DO NOT MERGE UNTIL AFTER 7.1 IS RELEASED**

This PR deploys app builds to a folder prefixed by build number. This is so we don't overwrite historical build artifacts so we can route electron updater to previous non revoked builds if the latest build has been revoked.

Related to https://github.com/Opentrons/robot-stack-infra/pull/34
 
# Changelog

- Deploy app artifacts to build number prefixed folder 

# Review requests
This one is hard to test without pushing actual internal release/release tags, we'll have to battle test this during the next alpha process after 7.1 gets released to production

# Risk assessment

High, this touches critical release artifact infrastructure

### DIFF
--- a/.github/workflows/app-test-build-deploy.yaml
+++ b/.github/workflows/app-test-build-deploy.yaml
@@ -290,7 +290,7 @@ jobs:
           HOST_PYTHON: python
           OPENTRONS_PROJECT: ${{ steps.project.outputs.project }}
           OT_APP_DEPLOY_BUCKET: ${{ steps.project.outputs.bucket }}
-          OT_APP_DEPLOY_FOLDER: ${{ steps.project.outputs.folder }}/${{OT_BUILD}}
+          OT_APP_DEPLOY_FOLDER: ${{ steps.project.outputs.folder }}
 
         run: |
           make -C app-shell dist-${{ matrix.os }}
@@ -343,13 +343,13 @@ jobs:
           aws configure set source_profile identity --profile deploy
         shell: bash
       - name: 'deploy release builds to s3'
-        if: contains(fromJSON(needs.determine-build-type.outputs.variants), 'release')
+        if: needs.determine-build-type.outputs.type == 'release'
         run: |
-          aws --profile=deploy s3 sync --acl=public-read to_upload_release/ s3://${{ env.OT_APP_DEPLOY_BUCKET }}/${{ env.OT_APP_DEPLOY_FOLDER }}
+          aws --profile=deploy s3 sync --acl=public-read to_upload_release/ s3://${{ env.OT_APP_DEPLOY_BUCKET }}/${{ env.OT_APP_DEPLOY_FOLDER }}/${{OT_BUILD}}
       - name: 'deploy internal-release release builds to s3'
-        if: contains(fromJSON(needs.determine-build-type.outputs.variants), 'internal-release')
+        if: needs.determine-build-type.outputs.type == 'internal-release'
         run: |
-          aws s3 --profile=deploy sync --acl=public-read to_upload_internal-release/ s3://${{ env.OT_APP_DEPLOY_BUCKET }}/${{ env.OT_APP_DEPLOY_FOLDER }}
+          aws s3 --profile=deploy sync --acl=public-read to_upload_internal-release/ s3://${{ env.OT_APP_DEPLOY_BUCKET }}/${{ env.OT_APP_DEPLOY_FOLDER }}/${{OT_BUILD}}
 
       - name: 'upload windows artifacts to GH release'
         uses: 'ncipollo/release-action@v1.12.0'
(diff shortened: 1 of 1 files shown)


.github/workflows/app-test-build-deploy.yaml
```diff
@@ -230,7 +230,7 @@ jobs:
              echo "Configuring project, bucket, and folder for ot3"
              echo "project=ot3" >> $GITHUB_OUTPUT
              echo "bucket=${{env._APP_DEPLOY_BUCKET_OT3}}" >> $GITHUB_OUTPUT
-             echo "folder=${{env._APP_DEPLOY_BUCKET_OT3}}" >> $GITHUB_OUTPUT
+             echo "folder=${{env._APP_DEPLOY_FOLDER_OT3}}" >> $GITHUB_OUTPUT
```
shlokamin, line 233: i think this was just straight up wrong before and we never noticed because we were using top level env vars to deploy below.

sfoster1: yup https://github.com/Opentrons/opentrons/pull/13966

.github/workflows/app-test-build-deploy.yaml
```diff
@@ -290,7 +290,7 @@ jobs:
           HOST_PYTHON: python
           OPENTRONS_PROJECT: ${{ steps.project.outputs.project }}
           OT_APP_DEPLOY_BUCKET: ${{ steps.project.outputs.bucket }}
-          OT_APP_DEPLOY_FOLDER: ${{ steps.project.outputs.folder }}
+          OT_APP_DEPLOY_FOLDER: ${{ steps.project.outputs.folder }}/${{OT_BUILD}}
```
sfoster1: I don't think this is right because there's two places that need the `OT_BUILD` to not be included:
1. Below (see comment) we upload `releases.json` to `OT_APP_DEPLOY_FOLDER`, and `releases.json` needs to be in `/app` not `/app/$OT_BUILD`
2. this environment variable configures the electron-updater's target URL, and we want electron-updater to continue to look at `/app/`

.github/workflows/app-test-build-deploy.yaml
```diff
@@ -343,11 +343,13 @@ jobs:
           aws configure set source_profile identity --profile deploy
         shell: bash
       - name: 'deploy release builds to s3'
+        if: contains(fromJSON(needs.determine-build-type.outputs.variants), 'release')
```
sfoster1: this definitely is wrong - `internal-release` contains `release` so it will fire on both. I think this won't do anything though because [the code above](https://github.com/Opentrons/opentrons/pull/13970/files#diff-61f587522c94a184b8acd038347c1bec0809206cdab7756d0d6655329b36ce7aR326-R333) creates one subdir for release artifacts and one subdir for internal-release artifacts and sorts them, so `to_upload_release/` would be empty if this was an internal release (this is also why these steps were not `if`-gated before)

.github/workflows/app-test-build-deploy.yaml
```diff
@@ -453,13 +455,13 @@ jobs:
       - name: 'update internal-releases releases.json'
         if: needs.determine-build-type.outputs.type == 'release' && contains(fromJSON(needs.determine-build-type.outputs.variants), 'internal-release')
         run: |
-          aws --profile=deploy s3 cp s3://${{ env._APP_DEPLOY_BUCKET_OT3 }}/${{ env._APP_DEPLOY_FOLDER_OT3 }}/releases.json ./to_upload_internal-release/releases.json
+          aws --profile=deploy s3 cp s3://${{ env._APP_DEPLOY_BUCKET_OT3 }}/${{ env.OT_APP_DEPLOY_FOLDER }}/releases.json ./to_upload_internal-release/releases.json
```
sfoster1, line 458: `releases.json` still wants to stay in `builds.opentrons.com/app/`, not `/app/$BUILDNUMBER` - this is one of the reasons I don't think we can just append the build id to the environment variable above

sfoster1 (changes requested):
inline notes

.github/workflows/app-test-build-deploy.yaml
shlokamin: lol yes that if statement is useless, good catch. i still think its worth correcting the if statement and including it though, even though one of the subdirs would be empty i think it makes things easier to follow

.github/workflows/app-test-build-deploy.yaml
shlokamin: beep boop beep yes releases.json needs to be at the top level. ill remove the build number from the env variable and only use it for app artifact deploys

.github/workflows/app-test-build-deploy.yaml
shlokamin: oopsy didnt realize that env var is used by the app shell and app shell odd, thanks!

shlokamin:
oops i let this get very stale, going to close this for now. we can reopen later if we want to pursue this

PR #13971, opened 2023-11-13
feat(robot-server): Load endpoint supplied deck configuration into protocol engine

CaseyBatten:
# Overview
Robot server side of the changes resolving deck configuration support between server and PE.

The process of putting a deck configuration list of Cutout fixtures now includes injecting the list of fixtures into the protocol engine. This occurs twice, first when the protocol engine is created and again bounded to the play action to ensure the most up to deck configuration is utilized. 

# Test Plan
Execute the following series of actions on a Flex:

- PUT a /deck_configuration to the robot, with a full list of deck fixtures including a "stagingAreaRightSlot" in B3 and a "stagingAreaSlotWithWasteChuteRightAdapterNoCover" in D3.

- Load the team internally built addressable areas JSON protocol which will load an armadillo plate to C1, have the gripper move it to B4, then into the Waste Chute.

- Ensure the JSON Protocol passes analysis.

- Run the protocol, ensure all events occur as described end-to-end and the gantry moves to appropriate areas of the deck as expected, and that the protocol ends gracefully.


sfoster1 (commented):
If it's the `PlayAction` that's carrying the information (rather than it being imbued into an engine instance) does that mean that you can change the deck configuration during a pause?

sanni-t:
> If it's the `PlayAction` that's carrying the information (rather than it being imbued into an engine instance) does that mean that you can change the deck configuration during a pause?

Nope. The `PlayAction` only gets the deck configuration when the engine has not been played yet (see `RunController.create_action` for how we pass the deck config). 
It is a bit of a confusing way to implement this. We are probably going to refactor this to use an explicit `updateDeckConfig` action. This PR is mainly to unblock the app team.

robot-server/robot_server/deck_configuration/store.py
```diff
@@ -26,3 +27,10 @@ async def get(self) -> models.DeckConfigurationResponse:
         return models.DeckConfigurationResponse.construct(
             cutoutFixtures=self._cutoutFixtures, lastUpdatedAt=self._last_updated_at
         )
+
+    def get_cutoutFixtures(self) -> DeckConfigurationType:
```
b-cooper: is this meant to be half camel half snake case?

CaseyBatten: I also thought that was a bit odd while writing it, it is querying for the variable literally named "cutoutFixtures" (hence the camel casing) and the "get_" naming was in line with other 'get' methods elsewhere, but it would probably feel more natural to change it to all snake casing such as "get_cutout_fixtures" or "get_deck_configuration_type" or just "get_deck_configuration".

api/src/opentrons/cli/analyze.py
```diff
@@ -79,7 +79,7 @@ async def _analyze(
     runner = await create_simulating_runner(
         robot_type=protocol_source.robot_type, protocol_config=protocol_source.config
     )
-    analysis = await runner.run(protocol_source)
+    analysis = await runner.run(deck_configuration=[], protocol_source=protocol_source)
```
SyntaxColoring, line 82: Am I correct in reading this `[]` as basically, "the `protocol_config` says to use a simulated deck, so the `deck_configuration` that we provide here doesn't matter"?

If so, and that's actually how it works, let's pass `None` instead of `[]`. `[]` reads to me as an *actual* deck configuration that happens to be empty (which is not valid), whereas `None` is closer to not mentioning any deck configuration.

api/src/opentrons/execute.py
```diff
@@ -627,7 +627,9 @@ async def run(protocol_source: ProtocolSource) -> None:
         try:
             # TODO(mm, 2023-06-30): This will home and drop tips at the end, which is not how
             # things have historically behaved with PAPIv2.13 and older or JSONv5 and older.
-            result = await protocol_runner.run(protocol_source)
+            result = await protocol_runner.run(
+                deck_configuration=[], protocol_source=protocol_source
+            )
```
SyntaxColoring, line 632:
```suggestion
            # TODO: `[]` needs to be replaced with the robot's actual current deck configuration.
            # https://opentrons.atlassian.net/browse/RSS-400
            result = await protocol_runner.run(
                deck_configuration=[], protocol_source=protocol_source
            )
```

api/src/opentrons/protocol_engine/types.py
```diff
@@ -774,3 +774,6 @@ class QuadrantNozzleLayoutConfiguration(BaseModel):
     RowNozzleLayoutConfiguration,
     QuadrantNozzleLayoutConfiguration,
 ]
+
+# TODO make the below some sort of better type
+DeckConfigurationType = List[Tuple[str, str]]
```
SyntaxColoring, line 779: 👍 Yeah I have some ideas for this. Making it a shared named type is a good move.

api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -138,6 +139,12 @@ class AddLabwareOffsetAction:
     request: LabwareOffsetCreate
 
 
+# class UpdateDeckConfigurationAction:
+#     """Update the existing deck configuration."""
+#     deck_configuration: DeckConfigurationType
+#     created_at: datetime
+
+
```
SyntaxColoring: Let's remove this commented code unless there's an imminent plan to uncomment it, in which case we should mark it with an explanatory comment.

(This is an Opentrons style guideline in general. If we need to resurrect dormant code, we tend to prefer digging through PRs and Git history, instead of keeping the dormant code in the source but commented.)

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -428,6 +429,16 @@ async def finish(
...
+    # def update_deck_configuration(self, request: DeckConfigurationType) -> None:
+    #     """Update the existing robot deck configuration."""
+    #     created_at = self._model_utils.get_timestamp()
+    #     self._action_dispatcher.dispatch(
+    #         UpdateDeckConfigurationAction(
+    #             deck_configuration=request,
+    #             created_at=created_at,
+    #         )
+    #     )
+
```
SyntaxColoring: Ditto my other comment about removing commented code.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -133,13 +134,13 @@ def add_plugin(self, plugin: AbstractPlugin) -> None:
         """Add a plugin to the engine to customize behavior."""
         self._plugin_starter.start(plugin)
 
-    def play(self) -> None:
+    def play(self, deck_configuration: Optional[DeckConfigurationType] = None) -> None:
```
SyntaxColoring, line 137: A couple things.

First, a request. Can we document what `deck_configuration=None` means here, and when the caller should do that?

-----

Second, a Python API design suggestion. You're doing two things here:

1. `Optional` is saying that `deck_configuration` can take the value of an actual list, or the value `None`.
2. `= None` is saying that if the caller doesn't mention `deck_configuration` at all, it should automatically take the value `None`.

The first thing is fine, but the second thing seems dangerous. `None` is not, in general, a safe default when calling this function. If I understand correctly, it's only appropriate when resuming an already-played run, or when analyzing a protocol. The caller really does need to understand what it's doing and choose what's appropriate for that context, and forcing the caller to specify an explicit value for `deck_configuration` is good because it prompts them to do that.

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -30,7 +30,9 @@ async def analyze(
             robot_type=protocol_resource.source.robot_type,
             protocol_config=protocol_resource.source.config,
         )
-        result = await runner.run(protocol_resource.source)
+        result = await runner.run(
+            protocol_source=protocol_resource.source, deck_configuration=[]
+        )
```
SyntaxColoring, line 35: Ditto my other comment about not passing `[]` as a placeholder value for simulation. We'd want this to be something like `None`, so it's explicitly no deck configuration instead of an invalid empty deck configuration.

g-code-testing/g_code_parsing/g_code_engine.py
```diff
@@ -176,7 +176,9 @@ async def run_protocol(
                     hardware_api=hardware,  # type: ignore
                 )
                 with GCodeWatcher(emulator_settings=self._config) as watcher:
-                    await protocol_runner.run(protocol_source=protocol_source)
+                    await protocol_runner.run(
+                        deck_configuration=[], protocol_source=protocol_source
+                    )
```
SyntaxColoring, line 181: Okay, this one is interesting.

This is `g-code-testing`, where we're running protocols with a more-deeply-emulated-than-usual hardware environment and making sure that the G-code stream is correct. In other words, we want this to act as if we're actually running on a real robot. It's not like other kinds of simulation, where we're half-assed executing a protocol basically just to scrape its equipment requirements and run log.

All this to say, we want the `deck_configuration` here to actually be full, concrete, and correct. Not `[]` and not `None`. You can leave this as a `# TODO`, but we should have a ticket for it.

robot-server/robot_server/deck_configuration/store.py
```diff
@@ -26,3 +27,10 @@ async def get(self) -> models.DeckConfigurationResponse:
         return models.DeckConfigurationResponse.construct(
             cutoutFixtures=self._cutoutFixtures, lastUpdatedAt=self._last_updated_at
         )
+
+    def get_deck_configuration(self) -> DeckConfigurationType:
+        """Get the robot's current deck configuration in an expected typing."""
+        deck_configuration: DeckConfigurationType = []
+        for item in self._cutoutFixtures:
+            deck_configuration.append((item.cutoutId, item.cutoutFixtureId))
+        return deck_configuration
```
SyntaxColoring, line 36: Python style thing: this imperative loop can be simplified into a list comprehension.

```suggestion
        return [(item.cutoutId, item.cutoutFixtureId) for item in self._cutoutFixtures]
```

robot-server/robot_server/maintenance_runs/maintenance_engine_store.py
```diff
@@ -125,6 +127,7 @@ async def create(
         run_id: str,
         created_at: datetime,
         labware_offsets: List[LabwareOffsetCreate],
+        deck_configuration: Optional[DeckConfigurationType] = [],
```
SyntaxColoring, line 130: Ditto my other comment about not having default values (`= []`) for these arguments, because in general, it's not safe for the caller to disregard them.

Actually, in this case, I think we can simplify the whole thing by removing the `= []` *and* the `Optional`. The `MaintenanceRunDataManager` always provides an explicit value, anyway.

robot-server/robot_server/runs/router/base_router.py
```diff
@@ -170,6 +181,7 @@ async def create_run(
             run_id=run_id,
             created_at=created_at,
             labware_offsets=offsets,
+            deck_configuration=deck_configuration,
```
SyntaxColoring, line 184: Does this router need to pass down an initial deck configuration? It will always be overwritten by the eventual `play` action, won't it? Removing this would let you simplify `RunDataManager`, etc.

Or maybe we do need to do this, because of `setup` commands? Do we still care about those?

robot-server/robot_server/runs/engine_store.py
```diff
@@ -171,6 +175,7 @@ async def create(
         else:
             load_fixed_trash = False
 
+        # convert the robot server deck config type to a list of tupled strings and pass to create protocol engine
```
SyntaxColoring: I think this comment is outdated.

robot-server/robot_server/runs/engine_store.py
```diff
@@ -33,6 +33,9 @@
 
 from robot_server.protocols import ProtocolResource
 
+# tbd cleanup this import
+from opentrons.protocol_engine.types import DeckConfigurationType
```
SyntaxColoring: Seems fine to me. What was on your mind?

robot-server/robot_server/runs/run_controller.py
```diff
@@ -69,7 +71,13 @@ def create_action(
                     # TODO(mc, 2022-05-13): engine_store.runner.run could raise
                     # the same errors as runner.play, but we are unable to catch them.
                     # This unlikely to occur in production, but should be addressed.
-                    self._task_runner.run(self._run_protocol_and_insert_result)
+                    # assert isinstance(
+                    #     action_payload, DeckConfigurationType
+                    # ), "Received invalid deck configuration"
```
SyntaxColoring: Looks like an `assert` left over from debugging that we can remove?

```suggestion
```

robot-server/tests/runs/router/conftest.py
```diff
@@ -52,3 +53,9 @@ def mock_run_auto_deleter(decoy: Decoy) -> RunAutoDeleter:
 def mock_maintenance_engine_store(decoy: Decoy) -> MaintenanceEngineStore:
     """Get a mock MaintenanceEngineStore interface."""
     return decoy.mock(cls=MaintenanceEngineStore)
+
+
+@pytest.fixture
+def mock_deck_configuration_store(decoy: Decoy) -> DeckConfigurationStore:
+    """Get a mock RunDataManager."""
```
SyntaxColoring:
```suggestion
    """Get a mock DeckConfigurationStore."""
```

robot-server/tests/maintenance_runs/router/conftest.py
```diff
@@ -27,3 +28,9 @@ def mock_protocol_engine(decoy: Decoy) -> ProtocolEngine:
 def mock_maintenance_run_data_manager(decoy: Decoy) -> MaintenanceRunDataManager:
     """Get a mock RunDataManager."""
     return decoy.mock(cls=MaintenanceRunDataManager)
+
+
+@pytest.fixture
+def mock_deck_configuration_store(decoy: Decoy) -> DeckConfigurationStore:
+    """Get a mock RunDataManager."""
```
SyntaxColoring:
```suggestion
    """Get a mock DeckConfigurationStore."""
```

api/src/opentrons/simulate.py
```diff
@@ -900,7 +900,9 @@ async def run(protocol_source: ProtocolSource) -> _SimulateResult:
 
         scraper = _CommandScraper(stack_logger, log_level, protocol_runner.broker)
         with scraper.scrape():
-            result = await protocol_runner.run(protocol_source)
+            result = await protocol_runner.run(
+                deck_configuration=[], protocol_source=protocol_source
+            )
```
SyntaxColoring, line 905: Ditto my other comment about using something more explicit, like `None`, as a placeholder when simulating.

SyntaxColoring (approved):
Discussed in-person. This looks like it'll do the trick, thanks!

Various low-level Python style things, and some medium-sized Protocol Engine architectural things, but overall this looks great to merge for unblocking further integration and QA.

robot-server/robot_server/runs/engine_store.py
CaseyBatten: Old comment, this was the cleaned version.

robot-server/robot_server/runs/run_controller.py
CaseyBatten: Originally the plan was to have the deck configuration be part of a greater class, but as it is now a list of tuples asserting isinstance is invalid here so it was commented out/removed.

PR #13972, opened 2023-11-13
feat(api): disable return tip for partially configured pipette

Laura-Danielle:
## Overview

It is unsafe for us to return tips to a tiprack while in partial tip configuration because we risk the potential of picking up used tips. Also added a few more tests for the virtual configured nozzle state to make sure it's functioning as expected.

Closes RSS-360

## Test Plan

- [ ] Try to return tip when pipette is in partially configured state and see that it fails
- [ ] Try to return tip in default state and see that it succeeds.


sfoster1 (approved):
Looks correct to me

PR #13974, opened 2023-11-13
chore(hardware-testing): Mergeback hw testing tags into edge

ryanthecoder:
# Overview

This mergebacks all of the various tags/branches used in the factor onto edge

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (changes requested):
Some of this is a little eyebrow raising and some of this we just cannot do.
- Are those 96 channel functions right? The ones with way fewer entries?
- P50 single v3.5 has already shipped. Changing the default liquid handling functions would result in different behavior of the robot when running protocols the customer has been running for a while. If we want to change the behavior of the pipette, we need to have a 3.6.
- I'm a bit uncomfortable with the `move_utils` change _unless_ these devices have been running _with stall detection on_. If they haven't been, we can't take that change for granted.

hardware/opentrons_hardware/hardware_control/motion_planning/move_utils.py
ryanthecoder, line 1: get rid of this

sfoster1 (approved):
Ah sweet thanks!

PR #13975, opened 2023-11-13
feat(hardware-testing): Autofind hw testing devices

ryanthecoder:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/hardware-testing/hardware_testing/drivers/asair_sensor.py
+++ b/hardware-testing/hardware_testing/drivers/asair_sensor.py
@@ -80,12 +80,9 @@ def BuildAsairSensor(simulate: bool, autosearch: bool = True) -> AsairSensorBase
     if not simulate:
         if not autosearch:
             port = list_ports_and_select(device_name="Asair environmental sensor")
-            try:
-                sensor = AsairSensor.connect(port)
-                ui.print_info(f"Found sensor on port {port}")
-                return sensor
-            except SerialException:
-                pass
+            sensor = AsairSensor.connect(port)
+            ui.print_info(f"Found sensor on port {port}")
+            return sensor
         else:
             ports = comports()
             assert ports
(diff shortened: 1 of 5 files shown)


hardware-testing/hardware_testing/drivers/pressure_fixture.py
```diff
@@ -98,6 +103,40 @@ def read_all_pressure_channel(self) -> List[float]:
         return pressure
 
 
+def connect_to_fixture(
```
andySigler, line 106: if something unexpected happened, raise an error, don't pass

hardware-testing/hardware_testing/drivers/asair_sensor.py
```diff
@@ -72,17 +74,31 @@ def get_serial(self) -> str:
         ...
 
 
-def BuildAsairSensor(simulate: bool) -> AsairSensorBase:
+def BuildAsairSensor(simulate: bool, autosearch: bool = True) -> AsairSensorBase:
```
andySigler, line 77: if something unexpected happened, raise an error, don't pass

andySigler (changes requested):

sfoster1 (changes requested):
This doesn't look like the right approach. It looks like what you're trying to do is
for each sensor,
- iterate through comports
- try to talk to that sensor
- if it doesn't respond right, it's not that sensor

But what if some sensor has coincidentally the same feedback? What if there's a different kind of error than "didn't respond right?"

If these sensors are running their own usb interfaces, we should filter on their vid/pid. If they're not, if all they have is identical-looking serial connections and you cannot tell from the host side which is which, I guess this is ok but I'd be much more strongly looking at instead having a guided flow where you start with everything disconnected and connect them one by one or something.

ryanthecoder:
The issue with using VID/PID is that these devices use serial adapters, and the adapters are the VID/PID that show up. The factory/Dumbo do not use consistent adapters either, so they aren't consistent and/or multiple devices show up as the same vid/pid.
There is currently a flow where the operator gets a list of the connected devices on the terminal and just has to remember which one is which and they absolutely hate this and if they select the wrong one it causes the script to fail and they need to restart the process.

We can't have them connect/disconnect because in the gravimetric/photometric setups they cannot reach the USB ports when the enclosure is over the robot.

I went through the datasheets for the sensors to ensure that these discovery messages won't interfere with each other. I tested this code on a bot that had all of them connected and tried connected to them in several orders to ensure that they don't block up or anything. Not Ideal but kind of the only option.

sfoster1 (approved):
ok, looks good if we raise an error when we don't find anything

andySigler (approved):

PR #13976, opened 2023-11-13
refactor(shared-data): update movable trash offset from cutout fixture

shlokamin:
# Overview

Fixes the offset from cutout fixture for movable trash addressable areas in the left most slots

Changes to the deck schema will follow soon which will necessitate updating these offset vectors again


SyntaxColoring (approved):
🆒

PR #13978, opened 2023-11-14
ci(labware-library): add build timeout of 30 minutes

shlokamin:
# Overview

LL builds sometimes hang and cause CI to hold onto runners for hours for no reason. builds should not take longer than 30 mins, and if they do, something bad probably happened and the action would need to be re triggered anyways.


b-cooper (approved):
🕛 🕧

PR #13982, opened 2023-11-14
refactor(components): update deck map styling

koji:
# Overview
Update the map view styling for BaseDeck.
For box-shadow, the design is using it for the labware in ODD but actually we cannot use `box-shadow` since `box-shadow` does not support svg. We need to use filters to draw lines on labware svg images.

design
https://www.figma.com/file/1ot18My22DALIcjdLl5LJh/Helix-Design-System?node-id=361%3A64749&mode=dev

close RAUT-773

The test errors came from a component that this PR didn't touch.

# Test Plan

# Changelog

# Review requests
`app/src/organisms/Devices/ProtocolRun/SetupLabware/__tests__/SetupLabwareMap.test.tsx` got the following error but seems that fixture hasn't updated for a while. Any suggestions are appreciated.

```
TestingLibraryElementError: Unable to find an element with the text: mock labware render of 300ul Tiprack FIXTURE. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```
 

# Risk assessment
low


koji:
- [x] resolve conflicts

brenthagen:
changing `componentPropsMatcher` to `partialComponentPropsMatcher` for the mock labware render in `SetupLabwareMap.test.tsx` should fix those tests

koji:
> changing `componentPropsMatcher` to `partialComponentPropsMatcher` for the mock labware render in `SetupLabwareMap.test.tsx` should fix those tests

Thanks you!

app/src/molecules/DeckThumbnail/index.tsx
```diff
@@ -116,6 +116,7 @@ export function DeckThumbnail(props: DeckThumbnailProps): JSX.Element | null {
       labwareLocations={labwareLocations}
       moduleLocations={moduleLocations}
       showSlotLabels={showSlotLabels}
+      isOnDevice={showSlotLabels}
```
brenthagen: instead of reusing `showSlotLabels` we should pass a separate optional `isOnDevice` prop to the deck thumbnail, defaulted to false

brenthagen (approved):
looks good - thanks for figuring out the svg outline render

PR #13984, opened 2023-11-15
chore: update update-from-usb fake release notes

sfoster1:
Add a note to not remove the USB drive while an update from it is in progress.


app-shell-odd/src/system-update/index.ts
```diff
@@ -213,6 +213,8 @@ const fakeReleaseNotesForMassStorage = (version: string): string => `
 # Opentrons Robot Software Version ${version}
 
 This update is from a USB mass storage device connected to your flex, and release notes cannot be shown.
```
ecormany: capitalize `Flex` 😇

sfoster1: Fixed @ecormany

smb2268 (approved):

mjhuff (approved):
lgtm!

PR #13988, opened 2023-11-15
app(fix): Connect to Wifi Network via USB

mjhuff:
Closes RQA-1777, RQA-1781

# Overview

This PR fixes a bug when attempting to connect to a wifi network via USB connection only - a success modal would immediately appear upon entering any password. Wifi form data was not properly passed to the appshellrequestor, yielding a POST request with a null body. This fix solves a few minor undocumented USB bugs as well concerning system time syncing and deck configuration sessions. 

Additionally, the Wifi error modal error handling did not accurately account for all instances that could throw an error (such as this bug). In addition to checking if the response contains an error object, let's also check the error response yielding "failure." This PR also adds more meaningful default error handling text for Wifi connectivity issues.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/99505e41-e045-4015-9439-3d5f9059fe12

### Fixed Behavior

https://github.com/Opentrons/opentrons/assets/64858653/60414a54-ba73-4d35-8fae-325986e01794

# Test Plan

- Connect to a robot that is only connected to a network via USB.
- Connect to a wifi network.

# Changelog
- Fix connecting to a Wifi network via USB.
- Add more robust error handling for Wifi network connectivity. 

# Risk assessment
low - there are only a few minor places that make use of fetchRobotApi and send data along with the request. These should be the only places affected.

### DIFF
--- a/app/src/molecules/modals/ErrorModal.tsx
+++ b/app/src/molecules/modals/ErrorModal.tsx
@@ -11,7 +11,7 @@ interface Props {
   description: string
   close?: () => unknown
   closeUrl?: string
-  error: { message?: string; [key: string]: unknown }
+  error: { message?: string; [key: string]: unknown } | null
 }
 
 const DEFAULT_HEADING = 'Unexpected Error'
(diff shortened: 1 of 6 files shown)


sfoster1 (approved):
Looks good to me!

PR #13989, opened 2023-11-15
fix(api): Aspirate, dispense, and mix can perform at 0ul

andySigler:
# Overview

I would expect `.aspirate(0)` and `.dispense(0)` to aspirate/dispense 0 ul.

Currently `.aspirate(0)` will aspirate the tip's max-volume, and `.dispense(0)` will dispense all volume currently in the tip.

Looking in git history and this line was added in instrument_context.py in 2020, where the `if not volume` is used to determine if the volume should be defaulted:

```python
c_vol = self._core.get_available_volume() if not volume else volume
```

I'm assuming this behavior wasn't intentional.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -165,7 +165,7 @@ class InstrumentContext(publisher.CommandPublisher):
 
         See :ref:`new-aspirate` for more details and examples.
 
-        :param volume: The volume to aspirate, measured in µL. If 0 or unspecified,
+        :param volume: The volume to aspirate, measured in µL. If unspecified,
                     defaults to the maximum volume for the pipette and its currently
                     attached tip.
         :type volume: int or float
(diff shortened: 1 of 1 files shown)


sfoster1 (changes requested):
Right or wrong, this is established behavior in currently-running protocol api versions and we cannot change it retroactively. Please gate this change behind being on api version 2.16 or above.

sfoster1 (changes requested):
Oh one more thing besides fixing tests - the docstrings for all three methods (which generate our docs) explicitly call out this behavior (see [aspirate](https://github.com/Opentrons/opentrons/pull/13989/files#diff-6811e2f4d45b8d159abed031a1b00c50e69ecc5aa7d8e19a674527fb4be3298cR168-R170) , [dispense](https://github.com/Opentrons/opentrons/pull/13989/files#diff-6811e2f4d45b8d159abed031a1b00c50e69ecc5aa7d8e19a674527fb4be3298cR280-R282), [mix](https://github.com/Opentrons/opentrons/pull/13989/files#diff-6811e2f4d45b8d159abed031a1b00c50e69ecc5aa7d8e19a674527fb4be3298cR414-R415) ). We also might call it out in other places and need to search through them.

sfoster1 (approved):
Happy with functionality and docs if it passes tests but @ecormany would you mind taking a look at the docstrings?

ecormany (approved):
LGTM. Thanks for writing up the docstrings to go with the new behavior!

PR #13990, opened 2023-11-15
feat(robot-server): Define default deck configurations

SyntaxColoring:
# Overview

Before this PR, if you do `GET /deck_configuration` on a fresh robot, it returns an empty list. This is confusing and wrong because an empty list is, by our own rules, invalid data for this endpoint.

We need to either:

* Return `null`, explicitly indicating that there is no deck configuration.
* Return some kind of default.

We [discussed this](https://opentrons.slack.com/archives/CSCLVUW3C/p1700065977943189) and decided to go with returning a default.

# Test Plan

Basic functionality is covered by integration tests.

End-to-end functionality is not covered. By "end-to-end functionality," I mean things like "these defaults let you run all the old protocols that you were running before when you upgrade from v7.0 to v7.1." I'm not sure how to cover that.

# Changelog

* Hard-code some default deck configurations for the `ot2_standard`, `ot2_short_trash`, and `ot3_standard` deck definitions.
    * Per [discussion](https://opentrons.slack.com/archives/CSCLVUW3C/p1700065977943189), the defaults have a trash in the back-right corner.
    * For immediate convenience, I've defined these in `robot-server`, in terms of its HTTP-facing Pydantic models. This is definitely wrong: we will need this stuff to be in `api` instead, in order to support `opentrons_execute` and friends. Moving everything over is for another PR.
* Make `GET /deck_configuration` return those defaults if the client has not yet issued a `PUT /deck_configuration` request.

# Review requests

* Does this make sense for the OT-2, given its "legacy fixtures" mechanism?

# Risk assessment

Low.


CaseyBatten (approved):
Changes all sensible - ran end to end on RSS Flex and successfully returned a default Deck value on GET all, was overwritten on a PUT call, and maintained through subsequent JSON Protocol runs during that session referencing non-default areas on the deck.

rickwierenga:
noob question: how to access `/deck_configuration`?

```
curl $ROBOT_IP:31950/deck_configuration
{"errors":[{"id":"BadRequest","title":"Bad Request","detail":"Not Found","errorCode":"4000"}]}
```

SyntaxColoring:
@rickwierenga Hm, that looks correct to me. Are you on robot software v7.1.0 or newer?

rickwierenga:
I am not. smh

PR #13998, opened 2023-11-16
fix(api): correct logic for simming partial tips

sfoster1:
When we simulate partial tip pickup we do it by emulating the hardware control logic in the virtual pipette data provider and then passing the result to a nozzle manager the way the hardware controller does. Unfortunately, the logic in the pipette virtual data handler did not match the hardware controller.

Until we get to a place where we can have an actual simualting hardware controller or something, we just have to keep these in sync.

## Testing
This is an analysis-only change, and has no effect on execution. The best way to test it is to test https://github.com/Opentrons/opentrons/pull/13972 since that's where I noticed the issue (we'll rebase that PR on this).

### DIFF
--- a/api/src/opentrons/protocol_engine/execution/tip_handler.py
+++ b/api/src/opentrons/protocol_engine/execution/tip_handler.py
@@ -63,6 +63,47 @@ class TipHandler(TypingProtocol):
         """Tell the Hardware API that a tip is attached."""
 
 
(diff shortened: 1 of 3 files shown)


Laura-Danielle (approved):
lgtm

sanni-t:
Oh awesome! This was giving me trouble in deck conflict check testing too. Thanks!

PR #13999, opened 2023-11-16
feat(shared-data, app): FE support for aspirateInPlace

jerader:
# Overview

This command is used for wasteChute and movableTrash

# Test Plan

There isn't really a way to test this so just review the code. The test case checks that the command text renders properly

# Changelog

- add `aspirateInPlace` types 
- add support in command text

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #14001, opened 2023-11-16
docs(robot-server): Document deck configuration snapshotting

SyntaxColoring:
# Overview

The new `PUT /deck_configuration` endpoint has a special timing relationship with the `/runs` endpoints. The Opentrons App will rely on this heavily, so we need to document it.

# Test Plan

None needed. Docs only.

# Changelog

Edit the public description of `PUT /deck_configuration` to explain how, even though the server will accept a `PUT` request at any time, it only has an effect on the robot's movement if you do it before you play a given run for the first time.

Leave some room for us to add validation in the future, because not having any right now makes me nervous, but we might not get to it before we release v7.1.

# Review requests

Please put any feedback in the form of haiku.

# Risk assessment

None. Docs only.

### DIFF
--- a/robot-server/robot_server/deck_configuration/router.py
+++ b/robot-server/robot_server/deck_configuration/router.py
@@ -37,6 +37,13 @@ router = fastapi.APIRouter()
         " configuration, such as loading a labware into a staging area slot that this deck"
         " configuration doesn't provide, the run command will fail with an error."
         "\n\n"
+        "**Warning:**"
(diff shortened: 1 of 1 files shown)


CaseyBatten (approved):
This description change looks good to me, accurately reflects current behavior.

SyntaxColoring:
I got a 👍 from @brenthagen and @b-cooper on the future plans for making this an error.

PR #14002, opened 2023-11-16
refactor(protocol-designer): staging area slots now render properly

jerader:
# Overview

This PR fixes some errors made during the full v4 deck migration stuff in the app. There were a few errors for editing staging area slots and rendering them on the deck map.

# Test Plan

Create a Flex protocol and add a some staging area slots. After you create the file, scroll down to modify the staging area slots. try adding and removing the slots and see that the function works as expected.

add the waste chute

Go to the deck map view and see that the staging area slots and waste chute render correctly and the waste chute isn't overlapping with the slot labels.

# Changelog

- fix some logic in `deckSetup` and `StagingAreasModal`
- add logic to PD's slot labels for rendering the labels lower if the waste chute is present

# Review requests

see test plan

# Risk assessment

low


mjhuff (approved):
lgtm!

PR #14004, opened 2023-11-16
fix(api): InstrumentContext.mix should not jump up after each dispense

andySigler:
# Overview

Gated behind api >=2.16, this makes it to the pipette doesn't jump up out of the well after each dispense during a mix.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Okay, looks good if it passes tests.

PR #14005, opened 2023-11-16
docs(robot-server): Update GET /health response examples

SyntaxColoring:
# Overview

The HTTP API docs were showing an incomplete example for `GET /health`. (@y3rsh thanks for catching this.)

# Test Plan

Run `make -C robot-server dev` and view http://localhost:31950/redoc#tag/Health/operation/get_health_health_get.

# Changelog

* Fix some fields, especially `robot_type`, not being shown in our `GET /health` example response on ReDoc. [Before](https://github.com/Opentrons/opentrons/assets/3236864/1fd1501f-5a33-4123-a620-9c690e65e36f), [after](https://github.com/Opentrons/opentrons/assets/3236864/2136ebe8-2312-4be9-8c39-8b7fced7fec6).
    * ReDoc normally keeps this stuff up to date automatically. The problem was that our Python code overrode the examples for *the whole model* instead of just for specific fields. This changes it to only override the examples for specific fields, so unmentioned fields will have their examples autogenerated, like we want.
* Update the example values for `name` and `robot_serial` to show how they can be different. Otherwise, leave the existing example values in place.

# Review requests

* Are there any apparent downsides to defining the examples field-by-field like this?
* There are a lot of other old endpoints that theoretically have this same problem. Searching `robot-server` for `schema_extra` is a good starting point. Do we want to ticket that, or just let it lie?

# Risk assessment

Low.

### DIFF
--- a/robot-server/robot_server/health/models.py
+++ b/robot-server/robot_server/health/models.py
@@ -21,10 +21,15 @@ class HealthLinks(BaseModel):
     serverLog: str = Field(
         ...,
         description="The path to the HTTP server logs endpoint",
+        examples=["/logs/server.log"],
     )
     oddLog: typing.Optional[str] = Field(
         None,
-        description="The path to the ODD app logs endpoint",
+        description=(
+            "The path to the on-device display app logs endpoint"
+            " (only present on the Opentrons Flex)"
+        ),
+        examples=["/logs/touchscreen.log"],
     )
     apiSpec: str = Field(
         ...,
--- a/robot-server/robot_server/health/models.py
+++ b/robot-server/robot_server/health/models.py
@@ -11,10 +11,12 @@ class HealthLinks(BaseModel):
     apiLog: str = Field(
         ...,
         description="The path to the API logs endpoint",
+        examples=["/logs/api.log"],
     )
     serialLog: str = Field(
         ...,
         description="The path to the motor control serial communication logs endpoint",
+        examples=["/logs/serial.log"],
     )
     serverLog: str = Field(
         ...,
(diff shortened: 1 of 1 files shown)


ecormany (commented):
Tested [in the sandbox for this branch](http://sandbox.docs.opentrons.com/field_level_examples/http/api_reference.html#tag/Health/operation/get_health_health_get). Generally looks good, although I see this in the links entry at the bottom of the example:
```
  "links": {
    "apiLog": "/logs/api.log",
    "serialLog": "/logs/serial.log",
    "serverLog": "string",
    "oddLog": "string",
    "apiSpec": "/openapi.json",
    "systemTime": "/system/time"
  }
```
1. Should there be an actual value for `serverLog`?
2. Should `oddLog` be hidden or have a null value because this example is for an OT-2?

SyntaxColoring:
> Should there be an actual value for `serverLog`?

Yes, thank you. Fixed for `serverLog` and `oddLog`.

> Should `oddLog` be hidden or have a null value because this example is for an OT-2?

Yeahhh maybe. A downside of doing field-level examples like this is that the example sort of becomes an amalgamation. I think that's okay-ish in principle because it demonstrates all the fields that could exist, even if they won't necessarily exist together, but I acknowledge it's potentially confusing.

I'm going to leave it non-null for now.

ecormany (approved):
Looks good with the new additions.

We can clear up the `oddLog` on OT-2 confusion later with either:
- better documentation
- rational behavior if you try to query the file at that path (valid blank file or a simple JSON object that says "OT-2 doesn't have a touchscreen" or what have you)

PR #14007, opened 2023-11-16
fix(app, component): dqa device detail deck configurator

koji:
# Overview
DQA for Device Detail Deck Configurator

Update slot label text 
[Address the comment](https://github.com/Opentrons/opentrons/pull/14007#issuecomment-1815319223)
Add No deck fixtures

- When flexing down, deck map could go center aligned and the list goes underneath the deck config as full width 
I will work on ↑ in a following PR. Seems that `SlotLabel` size is varied by Deck Configurator size so if we use the size in Figma the label is hard to read.
Additionally, if the app window width exceeds 1130px, the huge white space shows up which doesn't look good. 
Asking about these to Rob.

The interaction part was already implemented.

close RAUT-831

# Test Plan
- Go to Device Detail page

# Changelog
- update label font size
- add No deck fixtures and its test

# Review requests

# Risk assessment
low


koji:
- [ ] make the section [responsive](https://www.figma.com/file/nJBVk8a26wdls3ZnkgRmns/OCT-Release%3A-Desktop-%5BDeck-configuration-%26-Drop-tips%5D?type=design&node-id=1-459&mode=design&t=m7W6uuXHGfmMkQDy-0)

brenthagen:
let's remove the `wasteChuteDef` in `WasteChuteConfigFixture` as well as `trashBinDef`, `stagingAreaDef`, and `standardSlotDef` in the other config fixture components and instead define shared constants in a `constants.ts` file in the deck configurator directory: `SINGLE_SLOT_FIXTURE_WIDTH`, `STAGING_AREA_FIXTURE_WIDTH`, and `FIXTURE_HEIGHT` to pass to all of the config fixture foreign object components.

as far as i'm aware, we will not be defining fixture dimensions by JSON file or otherwise in `shared-data`, so let's just do it here for these components

brenthagen:
font size and line height should change for waste chute config fixture also - can we define a const `smallBodyTextSemiBold` in typography?
Screen Shot 2023-11-16 at 5 54 10 PM

components/src/hardware-sim/DeckConfigurator/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  parameters: {
+    loadName: 'trash_chute',
+  },
+  boundingBox: {
+    xDimension: 286.5,
+    yDimension: FIXTURE_HEIGHT,
+    zDimension: 0,
+  },
+}
```
brenthagen: should delete this

components/src/hardware-sim/DeckConfigurator/index.tsx
```diff
@@ -66,6 +67,11 @@ export function DeckConfigurator(props: DeckConfiguratorProps): JSX.Element {
     ({ cutoutFixtureId }) =>
       cutoutFixtureId != null && WASTE_CHUTE_FIXTURES.includes(cutoutFixtureId)
```
brenthagen: this needs to now check array `WASTE_CHUTE_ONLY_FIXTURES`

app.py
```diff
@@ -0,0 +1,104 @@
+import gradio as gr
```
brenthagen: looks like this file got in this PR by mistake

components/src/hardware-sim/DeckConfigurator/index.tsx
```diff
@@ -64,7 +66,14 @@ export function DeckConfigurator(props: DeckConfiguratorProps): JSX.Element {
   )
   const wasteChuteFixtures = configurableDeckConfig.filter(
     ({ cutoutFixtureId }) =>
-      cutoutFixtureId != null && WASTE_CHUTE_FIXTURES.includes(cutoutFixtureId)
+      cutoutFixtureId != null &&
+      (WASTE_CHUTE_FIXTURES.includes(cutoutFixtureId) ||
```
brenthagen: needs to be just `WASTE_CHUTE_ONLY_FIXTURES` not `WASTE_CHUTE_FIXTURES`

brenthagen (approved):
LGTM !

PR #14008, opened 2023-11-16
fix(app): render compatible fixture in protocol setup fixture list when conflicted

brenthagen:
# Overview

picks the fist compatible fixture to render name/image when existing configured fixture is not compatible with what is required by the protocol. tweaks trash bin text color because design mismatch was noticed.

closes RAUT-863

# Test Plan

 - manually verified correct fixture rendering in list

# Changelog

 - Renders compatible fixture in protocol setup fixture list when conflicted

# Review requests

# Risk assessment

low


b-cooper (approved):
🗑️

PR #14011, opened 2023-11-16
fix(app): fix double rendering labware in basedeck component

mjhuff:
Closes RQA-1899

# Overview

Adds logic to prevent duplicate rendering of labware if it has already been rendered on top of a module. This bug is only noticeable on the thermocycler, since labware is rendered on the TC at an offset.

### Current Behavior

Screenshot 2023-11-16 at 1 34 49 PM

### With Fix

Screenshot 2023-11-16 at 2 45 45 PM

# Test Plan
- AppUI is set up to test! Here's a protocol. 
[LPC Test (2).json](https://github.com/Opentrons/opentrons/files/13382982/LPC.Test.2.json)
- Launch LPC and proceed to a step with labware. Confirm in devtools there's no double rendering of labware.
- Perform LPC until step 4. Confirm the TC labware does not render directly on slot B1.

# Changelog
- Fix double rendering of labware while performing LPC.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/LabwarePositionCheck/PrepareSpace.tsx
+++ b/app/src/organisms/LabwarePositionCheck/PrepareSpace.tsx
@@ -112,7 +112,9 @@ export const PrepareSpace = (props: PrepareSpaceProps): JSX.Element | null => {
                 labwareLocation: location,
                 definition: labwareDef,
               },
-            ]}
+            ].filter(
+              () => !('moduleModel' in location && location.moduleModel != null)
+            )}
             deckConfig={deckConfig}
           />
         </Flex>


mjhuff:
Because BaseDeck and related components are in active development, I didn't want to refactor this too much as things will likely change. I think utilizing the filter method directly on the array is fine for now, but I'm open to feedback!

jerader (approved):
nice i'm glad this ended up being an easy fix and not some weird TC labware positioning issue. There are other components that renders both modules and labware like `DeckThumbnail` and `SetupLabwareMap` - i wonder if they also are double rendering labware on a module or not. If not, maybe we should follow the pattern that they have? This way is fine though but would defer to what @brenthagen thinks!

mjhuff:
> nice i'm glad this ended up being an easy fix and not some weird TC labware positioning issue. There are other components that renders both modules and labware like `DeckThumbnail` and `SetupLabwareMap` - i wonder if they also are double rendering labware on a module or not. If not, maybe we should follow the pattern that they have? This way is fine though but would defer to what @brenthagen thinks!

Good call - @jerader ! I double checked these other spots, and they are fortunately working as expected. The exact code shape in these components is relatively different than PrepareSpace. I do agree that we should keep things consistent, but I'd rather not refactor all of PrepareSpace at the moment given how quickly we are changing this part of the codebase. Totally down to circle back on this once the 7.1 work slows down!

PR #14015, opened 2023-11-16
fix(api): MoveToAddressableArea noop not raise

sfoster1:
This lets us test the parts of the frontend that require the command.


SyntaxColoring (approved):
yay

PR #14018, opened 2023-11-17
feat(api): load and move labware to and from staging slots

jbleon95:
# Overview

This PR hooks up the public Python API to allow `load_labware`, `load_adapter` and `move_labware` to use the A4, B4, C4, and D4 staging slots.

This was accomplished via adding a new `StagingSlotName` enum type that can be returned from the validation function `ensure_and_convert_deck_slot`. The new type is then allowed to be passed to the three methods mentioned above. For trying to load a module into a staging slot, an error will be raised. For now, attempting to access a staging slot via the `Deck` class will also raise an error.

# Test Plan

Unit test coverage and analyzed and ran the following protocol on hardware.

```
requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16",
}


def run(protocol_context):

    tiprack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_1000ul", "B4")
    
        "flex_1channel_1000", mount="left", tip_racks=[tiprack1]
    )

    protocol_context.move_labware(tiprack1, "B2", use_gripper=True)
```

# Changelog

- Added `StagingSlotName` type
- Modified `ensure_and_convert_deck_slot` to return `StagingSlotName` in addition to `DeckSlotName`
- Allow engine based protocol core to use `StagingSlotName` for `load_labware`, `load_adapter` and `move_labware`

# Review requests

The general location and shape of `StagingSlotName` and if this is a good direction to move forward with this.

# Risk assessment

Low, it's mostly hooking up PAPI to send the correct PE type when making the engine client call, some simple error checking and no other unnecessary changes.

### DIFF
--- a/api/src/opentrons/protocol_api/_types.py
+++ b/api/src/opentrons/protocol_api/_types.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from typing_extensions import Final
 import enum
 
@@ -16,3 +17,34 @@ A special location value, indicating that a labware is not currently on the robo
 
 See :ref:`off-deck-location` for details on using ``OFF_DECK`` with :py:obj:`ProtocolContext.move_labware()`.
 """
+
+
+class StagingSlotName(enum.Enum):
+    """Staging slot identifiers."""
+
+    SLOT_A4 = "A4"
+    SLOT_B4 = "B4"
+    SLOT_C4 = "C4"
+    SLOT_D4 = "D4"
+
+    @classmethod
+    def from_primitive(cls, value: str) -> StagingSlotName:
+        str_val = str(value).upper()
+        return cls(str_val)
+
+    @property
+    def id(self) -> str:
+        """This slot's unique ID, as it appears in the deck definition.
+
+        This can be used to look up slot details in the deck definition.
+
+        This is preferred over `.value` or `.__str__()` for explicitness.
+        """
+        return self.value
+
+    def __str__(self) -> str:
+        """Stringify to the unique ID.
+
+        For explicitness, prefer using `.id` instead.
+        """
+        return self.id
--- a/api/src/opentrons/protocol_api/_types.py
+++ b/api/src/opentrons/protocol_api/_types.py
@@ -19,6 +19,7 @@ See :ref:`off-deck-location` for details on using ``OFF_DECK`` with :py:obj:`Pro
 """
 
 
+# TODO(jbl 11-17-2023) move this away from being an Enum and make this a NewType or something similar
 class StagingSlotName(enum.Enum):
     """Staging slot identifiers."""
 
@@ -29,7 +30,7 @@ class StagingSlotName(enum.Enum):
 
     @classmethod
     def from_primitive(cls, value: str) -> StagingSlotName:
-        str_val = str(value).upper()
+        str_val = value.upper()
         return cls(str_val)
 
     @property
--- a/api/src/opentrons/protocol_api/validation.py
+++ b/api/src/opentrons/protocol_api/validation.py
@@ -158,10 +158,8 @@ def ensure_and_convert_deck_slot(
             raise APIVersionError(
                 f"Using a staging deck slot requires apiLevel {_STAGING_DECK_SLOT_VERSION_GATE}."
             )
-        try:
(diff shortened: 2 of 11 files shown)


api/src/opentrons/protocol_api/_types.py
```diff
@@ -16,3 +17,34 @@ class OffDeckType(enum.Enum):
...
+    """Staging slot identifiers."""
+
+    SLOT_A4 = "A4"
+    SLOT_B4 = "B4"
+    SLOT_C4 = "C4"
+    SLOT_D4 = "D4"
+
+    @classmethod
+    def from_primitive(cls, value: str) -> StagingSlotName:
+        str_val = str(value).upper()
```
SyntaxColoring: Looks like this can just be

```suggestion
        str_val = value.upper()
```

Since these are never `int`s.

api/src/opentrons/protocol_api/_types.py
```diff
@@ -16,3 +17,34 @@ class OffDeckType(enum.Enum):
...
 """
+
+
+class StagingSlotName(enum.Enum):
+    """Staging slot identifiers."""
+
+    SLOT_A4 = "A4"
+    SLOT_B4 = "B4"
+    SLOT_C4 = "C4"
+    SLOT_D4 = "D4"
```
SyntaxColoring, line 29: Recap from in-person discussion:

This is good for this PR. Ideally, instead of an `Enum`, we'll make this a `NewType` or something so we don't have to duplicate these names between JSON and Python files.

Would you mind making a `# TODO` comment for that?

api/src/opentrons/protocol_api/validation.py
```diff
@@ -127,7 +131,7 @@ def ensure_pipette_name(pipette_name: str) -> PipetteNameType:
 
 def ensure_and_convert_deck_slot(
     deck_slot: Union[int, str], api_version: APIVersion, robot_type: RobotType
-) -> DeckSlotName:
+) -> Union[DeckSlotName, StagingSlotName]:
```
SyntaxColoring, line 137: Okay, so right now, this function is ensuring that the primitive represents a valid deck slot for the given *robot type.* We'll want to change its interface so it instead ensures that the primitive represents a valid deck slot for the given *deck configuration.*

Would you mind making a `# TODO` comment for that?

api/tests/opentrons/protocol_api/test_validation.py
```diff
@@ -131,6 +132,11 @@ def test_ensure_pipette_input_invalid(input_value: str) -> None:
         ("a3", APIVersion(2, 15), "OT-3 Standard", DeckSlotName.SLOT_A3),
         ("A3", APIVersion(2, 15), "OT-2 Standard", DeckSlotName.FIXED_TRASH),
         ("A3", APIVersion(2, 15), "OT-3 Standard", DeckSlotName.SLOT_A3),
+        # Staging slots:
+        ("A4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_A4),
+        ("B4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_B4),
+        ("C4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_C4),
+        ("D4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_D4),
```
SyntaxColoring: Let's get a lowercase test case in here, since the staging area slots are sometimes "special" and there's a chance we could accidentally drop our autocapitalization logic from them.

```suggestion
        ("A4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_A4),
        ("a4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_A4),
        ("B4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_B4),
        ("C4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_C4),
        ("D4", APIVersion(2, 16), "OT-3 Standard", StagingSlotName.SLOT_D4),
```

SyntaxColoring (approved):
Looks great. I just have minor nitpicks. Thank you!

PR #14022, opened 2023-11-17
feat(api,robot-server): Persist deck configurations across boots

SyntaxColoring:
# Overview

Closes RSS-396.

# Test Plan

This should be tested on an OT-2 and a Flex.

1. Push this branch to a fresh robot.
2. `GET /deck_configuration` and make sure it's the default.
3. `PUT /deck_configuration`, with new data, then `GET /deck_configuration` again and make sure it's changed.
4. On the robot's filesystem, make sure the file is where we expect it to be. (Should be `/data/robot/deck_configuration.json` on an OT-2. I'm not sure where these things end up on a Flex.)
5. Power-cycle the robot.
6. `GET /deck_configuration` and make sure it's the last thing that you set.
7. `POST /settings/reset {"deckConfiguration": true}`, then `GET /deck_configuration` and make sure it's back to the default.

# Changelog

* Persist the deck configuration across reboots, and wire this up to `robot-server`'s `PUT /deck_configuration` and `GET /deck_configuration` endpoints.
  * To store it, we extend `opentrons.calibration_storage` to support deck configuration. It's saved in a new `deck_configuration` file, adjacent to the deck *calibration.* On an OT-2, for example, it will be `/data/robot/deck_configuration.json`.
  * This implementation is a change from the [original plan](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3825041467/Deck+configuration+database+architecture), which would have used `robot-server`'s SQL database. We're not doing that here because it turns out `opentrons_execute` and friends need to be able to access this data, which means it needs to be owned by `api` and not `robot-server`.
* Add a new `"deckCalibration"` option to `POST /settings/reset`.

# Review requests

* I'm pretty unfamiliar with `opentrons.config` and `opentrons.calibration_storage`, so I'd appreciate general feedback in that area.
* Do we like the path `/data/robot/deck_configuration.json` for this new file? It will be hard to change in the future.
* Do we want to rename `opentrons.calibration_storage` to just `opentrons.storage` now?

# Risk assessment

Low. This mostly builds out new code. Changes to existing code are minor.


api/src/opentrons/config/__init__.py
```diff
@@ -184,7 +184,7 @@ class ConfigElement(NamedTuple):
         "Deck Calibration",
         Path("deck_calibration.json"),
         ConfigElementType.FILE,
-        "The file storing the deck calibration",
+        "The file storing the deck calibration. Superseded in v4 by robot_calibration_dir.",
```
SyntaxColoring, line 187: Is this correct? I'm going by this code, which calls this the "legacy deck calibration file."

https://github.com/Opentrons/opentrons/blob/a98c6ee558f4c6b86fdfc895303fa1eab15d632d/api/src/opentrons/calibration_storage/ot2/deck_attitude.py#L24

api/src/opentrons/calibration_storage/file_operators.py
```diff
@@ -63,7 +69,7 @@ def read_cal_file(
     # This can be done when the labware endpoints
     # are refactored to grab tip length calibration
     # from the correct locations.
-    with open(filepath, "r") as f:
+    with open(file_path, "r", encoding="utf-8") as f:
```
SyntaxColoring, line 72: The omission of `encoding="utf-8"` here looked like a bug.

If you omit `encoding`, the default is platform-dependent. That's fine, but `save_to_file()` explicitly specifies `encoding="utf-8"`. They need to be consistent.

I think this has been dormant because we've happened to never store any characters outside the common subset of UTF-8 and Windows's default codec (cp1252?).

api/src/opentrons/calibration_storage/file_operators.py
```diff
@@ -76,22 +82,68 @@ def read_cal_file(
...
     """
     Function used to save data to a file
 
-    :param filepath: path to save data at
-    :param data: data to save
+    :param directory_path: path to the directory in which to save the data
+    :param file_name: name of the file within the directory, *without the extension*.
+    :param data: The data to save. Either a Pydantic model, or a JSON-like dict to pass to
+        `json.dumps()`. If you're storing a Pydantic model, prefer `save_pydantic_model_to_file()`
+        and `read_pydantic_model_from_file()` for new code.
```
SyntaxColoring, line 99: To elaborate shortly on why I made a new pair of I/O functions instead of reusing these existing ones, the big things on my mind are:

* I wanted a simple interface for the various errors; I didn't want the caller to be burdened by the specifics of the file-missing, invalid-as-JSON, and invalid-as-model cases.
* The older functions go JSON↔dict↔Pydantic. In my experience, you ideally want to go straight JSON↔Pydantic. Putting the dict in the middle [isn't really supported in our Pydantic version](https://github.com/pydantic/pydantic/issues/1409), even though it often works in practice.

Those are fixable in the older I/O functions, but since they're on such a critical path, I didn't want to mess with them until I have time to test changes to them properly.

jbleon95:
Went through test plan successfully with expected results on Flex and OT-2.

api/src/opentrons/config/__init__.py
sfoster1: yeah

sfoster1 (approved):
The code here looks good to me. I don't love storing this in the api but we can for now.

I think that if we want to tie this in to the simulate and execute cli when run on the robot we would be better off even with 
- using requests to fetch the deck configuration from the running server
- storing it in json in `robot_server.deck_configuration.store` but in the server's persistence directory and using out-of-band knowledge to load it in a one-off place in the script entrypoints
something like that. I don't think it belongs next to calibration storage and I don't think the api directory should grow a general storage class because I'd like to reduce the ones it does have.

CaseyBatten (approved):
This looks good, and given the reported passing tests on Flex/OT2 hardware I think we're safe to move forward. I do think that there isn't necessarily an issue with having this stored among the other calibrations under API for now.

Theres been a bit of back and forth in person as well about where the "storage responsibility" should fall, but we don't really know how the items included in a deck configuration will grow in the future (if it ever does) so the appropriate location for this might change over time. I'm not too picky either way for the time being, as this is functionally capable of handling what we want right now.

SyntaxColoring:
> I think that if we want to tie this in to the simulate and execute cli when run on the robot we would be better off even with
> 
> * using requests to fetch the deck configuration from the running server

Unfortunately I don't think we have a path to do this now, because we instruct users to `systemctl stop opentrons-robot-server` before they use `opentrons.execute.execute()` and friends.

> * storing it in json in `robot_server.deck_configuration.store` but in the server's persistence directory and using out-of-band knowledge to load it in a one-off place in the script entrypoints

@sfoster1 and I discussed this in person and agreed to try this. The out-of-band magic file path is a hack, but in The Grand Vision, that hack is architecturally better than the `opentrons` library having to provide even more of a mutable storage system.

So that's implemented in this PR now. I've tested it on a Flex and confirmed that the storage still works. The wire-up to `opentrons.execute` can't be tested until after other work (e.g. RSS-400), but it's easy to see how we'd do it.

A known bug in this implementation is that factory-resetting "runs and protocols" will inadvertently *also* factory-reset the deck configuration. That's because it deletes `robot-server`'s entire persistence directory, which used to just be for runs and protocols. I think that's fixable, it'll just take some annoying renaming and reorganization. RSS-403

sfoster1 (approved):
I'm really happy with this approach, I think it's the appropriate separation of concerns. Thank you!

CaseyBatten (approved):
All changes look good, based on discussions in person have a good idea of the remaining issues that need to be ticketed out, but this looks great for current needs.

api/src/opentrons/calibration_storage/deck_configuration.py
```diff
@@ -0,0 +1,57 @@
...
+            )
+            for e in cutout_fixture_placements
+        ],
+        lastModified=last_modified,
+    )
+    return io.serialize_pydantic_model(data)
+
+
+# TODO(mm, 2023-11-21): If the data is corrupt, we should propagate the underlying error.
+# And there should be an enumerated "corrupt storage" error in shared-data.
```
sanni-t, line 39: +1

sanni-t (approved):

PR #14023, opened 2023-11-20
feat(protocol-designer, step-generation): trash is no longer a labware entity

jerader:
closes RAUT-864 RAUT-865

# Overview

Now that trash bin is no longer a labware entity, I've started the refactor to remove it from the PD's `labwareEntities` object and instead add it to `additionalEquipmentEntities`. Also, dropping tip is supported but aspirate/dispense/blowout/airgap is NOT yet.

Additionally, since aspirate/dispense/blowout/airgap isn't fully wired up yet, this breaks the E2E experience (for now) where you can't import old protocols 
 
# Test Plan

Create a Flex protocol, make sure your deck configuration ff is turned on. Go through the file wizard and just make sure your trash bin is added.

Go to the deck map and add a transfer form step and make sure aspirate/dispense is NOT in the trash bin. But make sure the drop tip is.

Export the protocol and look at the JSON file. You should see a `moveToAddressableArea` with the `addressableAreaName` param matching the correct addressable area name for the trash bin's location, followed by a `dropTipInPlace`

Bonus: if you move a labware into the waste chute (you need to use a gripper in order to do this), and then don't add any pipetting commands, the protocol should successfully upload to the app!

# Changelog

- a bunch of refactoring to move trash bin out of labware entities and into additonalEquipmentEntities
- some slot fixing so the trash's location is a cutoutId and fix up the logic throughout the components
- creation of `movableTrashCommandsUtil` to be used to help wire up all the commands in step-generation
- wire up dropping tip into the trash bin in step-generation

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/cypress/integration/migrations.spec.js
+++ b/protocol-designer/cypress/integration/migrations.spec.js
@@ -13,31 +13,32 @@ describe('Protocol fixtures migrate and match snapshots', () => {
   })
 
   const testCases = [
-    {
-      title: 'example_1_1_0 (schema 1, PD version 1.1.1) -> PD 8.0.x, schema 8',
-      importFixture: '../../fixtures/protocol/1/example_1_1_0.json',
-      expectedExportFixture:
-        '../../fixtures/protocol/8/example_1_1_0MigratedToV8.json',
-      unusedPipettes: true,
-      migrationModal: 'newLabwareDefs',
-    },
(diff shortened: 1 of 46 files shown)


mjhuff (approved):
Passes testing and the code makes enough sense. LGTM! 🗑️

PR #14025, opened 2023-11-20
refactor(app): make deck config section responsive

koji:
# Overview
Add media query to make deck config section responsive.
This PR does not align with the design completely responsive since deck configurator(svg) has own paddings.
Brian and Brent talked about making cartoon version of deck view for this in the future.

The key for this pr is to avoid cutting off the table when the windows width is less than 1024px.

I think we can make the entire device detail page responsive after 7.1.0.
 
design
https://www.figma.com/file/nJBVk8a26wdls3ZnkgRmns/OCT-Release%3A-Desktop-%5BDeck-configuration-%26-Drop-tips%5D?node-id=1620%3A14486&mode=dev

Screenshot 2023-11-20 at 12 05 30 PM

Screenshot 2023-11-20 at 12 07 42 PM

close RAUT-872

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/app/src/organisms/DeviceDetailsDeckConfiguration/index.tsx
+++ b/app/src/organisms/DeviceDetailsDeckConfiguration/index.tsx
@@ -11,10 +11,8 @@ import {
   DIRECTION_COLUMN,
   DIRECTION_ROW,
   Flex,
-  JUSTIFY_CENTER,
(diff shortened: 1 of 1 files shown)


mjhuff (approved):
LGTM. Yes, making this page responsive at some point would be great. Thanks for adding this!

PR #14028, opened 2023-11-20
refactor(protocol-designer): change tiprack default slot and copy upd…

jerader:
…ates

closes RAUT-875

# Overview

This PR does a few things as discussed in a slack thread. They are all quality of life improvements from Anurag:
1. updates the copy in the pipette type tiles to specify "left" and "right" instead of "first" and "second"
2. adds a default slot for the generated pipettes. instead of being generated in the next available slot, they are generated in the center column for both the Ot-2 and Flex. This means that you can modify the modules without having to worry about the slot being occupied by the default tiprack
3. updates the copy for an occupied slot warning when editing the modules to something more descriptive

# Test Plan

Create a Flex protocol and notice that the pipette type tiles include which mount you are selecting a pipetting for. Make sure you add a pipette to both mounts. Once you've created the protocol, navigate to the design tab and see that the tipracks are generated in slots B2 and C2. Move one of the tipracks into slot A2. Then add a magnetic block from the file tab and see that the error says that the slot is occupied.

Now create a new protocol for the OT-2. Add both pipettes. Go to the design tab and see that the tipracks are generated in slots 2 and 5. 

# Changelog

- add a slot for the generated tips
- change the copy for the alert text and the modal tile headers

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/index.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/index.tsx
@@ -263,9 +263,15 @@ export function CreateFileWizard(): JSX.Element | null {
       const newTiprackModels: string[] = uniq(
         pipettes.map(pipette => pipette.tiprackDefURI)
       )
-      newTiprackModels.forEach(tiprackDefURI => {
+      newTiprackModels.forEach((tiprackDefURI, index) => {
+        const ot2Slots = index === 0 ? '2' : '5'
+        const flexSlots = index === 0 ? 'C2' : 'B2'
         dispatch(
           labwareIngredActions.createContainer({
+            slot:
+              values.fields.robotType === FLEX_ROBOT_TYPE
+                ? flexSlots
+                : ot2Slots,
             labwareDefURI: tiprackDefURI,
             adapterUnderLabwareDefURI:
               values.pipettesByMount.left.pipetteName === 'p1000_96'
--- a/protocol-designer/src/components/modals/CreateFileWizard/PipetteTypeTile.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/PipetteTypeTile.tsx
@@ -46,7 +46,7 @@ export function FirstPipetteTypeTile(
       mount={mount}
       allowNoPipette={false}
       display96Channel={allow96Channel}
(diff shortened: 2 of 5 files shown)


protocol-designer/src/components/modals/CreateFileWizard/index.tsx
```diff
@@ -263,9 +263,15 @@ export function CreateFileWizard(): JSX.Element | null {
       const newTiprackModels: string[] = uniq(
         pipettes.map(pipette => pipette.tiprackDefURI)
       )
-      newTiprackModels.forEach(tiprackDefURI => {
+      newTiprackModels.forEach((tiprackDefURI, index) => {
+        const ot2Slots = index === 0 ? '2' : '5'
+        const flexSlots = index === 0 ? 'C2' : 'B2'
```
jerader, line 268: any suggestions for a more elegant way to accomplish this? It works but idk if it could be cleaner.

mjhuff: Off the top of my head, I can't think of anything clearer than this. 

The only thing slightly confusing IMO is variable assignment above the dispatch and then property assignment within the dispatch. Maybe something like this? Again, just my opinion - this is definitely nit territory.
```
const ot2Slots = index === 0 ? '2' : '5'
        const flexSlots = index === 0 ? 'C2' : 'B2'
        const slot =
          values.fields.robotType === FLEX_ROBOT_TYPE ? flexSlots : ot2Slots
        const adapterUnderLabwareDefURI =
          values.pipettesByMount.left.pipetteName === 'p1000_96'
            ? adapter96ChannelDefUri
            : undefined
```

        dispatch(
          labwareIngredActions.createContainer({
            slot,
            labwareDefURI: tiprackDefURI,
            adapterUnderLabwareDefURI,
          })
        )`

mjhuff (approved):
🚀

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
koji: I think L267-268 is clear enough.
I think Jamey's suggestion makes the code more readable and easier to debug.

koji (approved):
The changes look good to me.

PR #14029, opened 2023-11-20
refactor(robot-server,api-client): Rename `GET /deck_configuration` field `lastUpdatedAt` to `lastModifiedAt`

SyntaxColoring:
# Overview

When I added the `GET /deck_configuration` endpoint, I gave it a `lastUpdatedAt` field.

That name is accidentally inconsistent with our calibration endpoints (deck, pipette, etc.), which instead use `lastModified`.

This PR changes `lastUpdatedAt` to `lastModifiedAt`.

I'm deliberately keeping the `At` suffix because that has been an intentional naming convention of our HTTP API going forward. (Protocol Engine has `createdAt`, `completedAt`, etc.) So that's an intentional inconsistency.

# Test Plan

None applicable. As far as I can tell, there's not yet any app code that reads this field.

# Changelog

* Rename the `lastUpdatedAt` field in `GET /deck_configuration` to `lastModifiedAt`.
* Update the TypeScript bindings. 

# Review requests

None in particular.

# Risk assessment

Low for now, since nothing uses this yet.

### DIFF
--- a/api-client/src/deck_configuration/types.ts
+++ b/api-client/src/deck_configuration/types.ts
@@ -9,6 +9,6 @@ export interface UpdateDeckConfigurationRequest {
 export interface DeckConfigurationResponse {
   data: {
     cutoutFixtures: DeckConfiguration
-    lastUpdatedAt: string
+    lastModifiedAt: string
   }
 }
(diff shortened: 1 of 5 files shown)


mjhuff (approved):
Yep, doesn't look like we're using this anywhere app side, and I can't imagine this could be an issue. LGTM!

jerader (approved):
makes sense to me! thank you!

SyntaxColoring:
~~Will merge after #14022 to avoid a merge conflict.~~

PR #14031, opened 2023-11-20
feat(app): Add warning for 96-channel calibration with waste chute

smb2268:
fix RQA-1883, RQA-1897, RAUT-848
# Overview
This PR adds a warning to calibration flows that use a 96 channel telling users to detach the waste chute before calibrating. It also updates some styling and copy of banners in these flows to match designs.

# Test Plan
With a 96-channel attached, initiate both calibration and module calibration flows. Ensure the following is true:
1. For pipette cal, if no waste chute is in the deck config you should see a yellow warning on both the before beginning and attach probe screens
2. If the waste chute is in the deck config, this should be a red banner instead of yellow
3. For module calibration, the same logic should apply to the attach probe screen only if there is a module in slot C3 being calibrated

# Changelog
1. Add banner for waste chute collision to pipette cal before beginning and attach probe and module cal attach probe screen
2. Move all banners in pipette flows to below the body text instead of above it
3. Update copy and padding to match designs
4. Fix styling bug on LPC early exit screen

# Review requests
Test this out if you can! Should be doable with a simulator

# Risk assessment
Low

### DIFF
--- a/app/src/assets/localization/en/pipette_wizard_flows.json
+++ b/app/src/assets/localization/en/pipette_wizard_flows.json
@@ -46,7 +46,7 @@
   "hold_and_loosen": "Hold the pipette in place and loosen the pipette screws. (The screws are captive and will not come apart from the pipette.) Then carefully remove the pipette.",
   "hold_pipette_carefully": "Hold onto the pipette so it does not fall. Connect the pipette by aligning the two protruding rods on the mounting plate. Ensure a secure attachment by screwing in the four front screws with the provided screwdriver.",
   "how_to_reattach": "<block>Push the right pipette mount up to the top of the z-axis. Then tighten the captive screw at the top right of the gantry carriage.</block><block>When reattached, the right mount should no longer freely move up and down.</block>",
-  "install_probe": "Take the calibration probe from its storage location. Ensure its collar is fully unlocked. Push the pipette ejector up and press the probe firmly onto the <bold>{{location}}</bold> pipette nozzle as far as it can go. Twist the collar to lock the probe. Test that the probe is secure by gently pulling it back and forth.",
+  "install_probe": "Take the calibration probe from its storage location. Ensure its collar is unlocked. Push the pipette ejector up and press the probe firmly onto the <bold>{{location}}</bold> pipette nozzle. Twist the collar to lock the probe. Test that the probe is secure by gently pulling it back and forth.",
   "loose_detach": "Loosen screws and detach ",
   "move_gantry_to_front": "Move gantry to front",
   "must_detach_mounting_plate": "You must detach the mounting plate and reattach the z-axis carraige before using other pipettes. We do not recommend exiting this process before completion.",
@@ -84,6 +84,8 @@
   "unscrew_and_detach": "Loosen Screws and Detach Mounting Plate",
   "unscrew_at_top": "<block>Loosen the captive screw on the top right of the  carriage. This releases the right pipette mount, which should then freely move up and down.</block>",
   "unscrew_carriage": "unscrew z-axis carriage",
+  "waste_chute_error": "Remove the Waste Chute from the Deck Plate Adaptor before proceeding.",
+  "waste_chute_warning": "If the Waste Chute is installed, remove it from the Deck Plate Adaptor before proceeding.",
   "wrong_pip": "wrong instrument installed",
   "z_axis_still_attached": "z-axis screw still secure"
 }
--- a/app/src/assets/localization/en/pipette_wizard_flows.json
+++ b/app/src/assets/localization/en/pipette_wizard_flows.json
@@ -84,8 +84,8 @@
   "unscrew_and_detach": "Loosen Screws and Detach Mounting Plate",
   "unscrew_at_top": "<block>Loosen the captive screw on the top right of the  carriage. This releases the right pipette mount, which should then freely move up and down.</block>",
   "unscrew_carriage": "unscrew z-axis carriage",
-  "waste_chute_error": "Remove the Waste Chute from the Deck Plate Adaptor before proceeding.",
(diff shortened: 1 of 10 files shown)


app/src/organisms/ModuleWizardFlows/AttachProbe.tsx
```diff
@@ -59,26 +62,34 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
...
       pipetteAttachProbeVideoSource = attachProbe96
-      i18nKey = 'install_probe_96_channel'
+      probeLocation = t('pipette_wizard_flows:ninety_six_probe_location')
       break
   }
+  const wasteChuteConflict =
+    slotName === 'C3' && attachedPipette.data.channels === 96
+  const deckConfig = useDeckConfigurationQuery().data
+  const isWasteChuteOnDeck =
+    deckConfig?.find(fixture => fixture.loadName === 'wasteChute') ?? false
```
koji:
```ts
export interface CutoutConfig {
  cutoutId: CutoutId
  cutoutFixtureId: CutoutFixtureId | null
}

export type DeckConfiguration = CutoutConfig[]
```
probably need to change something like this.
```suggestion
    import { WasteChuteCutoutFixtureId } from '@opentrons/shared-data'
    deckConfig?.find(fixture => fixture.cutoutFixtureId=== WasteChuteCutoutFixtureId) ?? false
```

app/src/organisms/ModuleWizardFlows/AttachProbe.tsx
```diff
@@ -135,19 +146,30 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
...
           block: <StyledText css={BODY_STYLE} />,
         }}
       />
-    ) : (
-      <StyledText css={BODY_STYLE}>{t('install_probe')}</StyledText>
-    )
+      {wasteChuteConflict && (
+        <Banner
+          type={isWasteChuteOnDeck ? 'error' : 'warning'}
+          size={isOnDevice ? '1.5rem' : SIZE_1}
```
koji: nit
We stopped using `SIZE_number` for the future Design system unification.

```suggestion
          size={isOnDevice ? '1.5rem' : '1rem'}
```

app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
```diff
@@ -75,6 +80,9 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
     (instrument): instrument is PipetteData =>
       instrument.ok && instrument.mount === mount
   )
+  const deckConfig = useDeckConfigurationQuery().data
+  const isWasteChuteOnDeck =
+    deckConfig?.find(fixture => fixture.loadName === 'wasteChute') ?? false
```
koji: same as the above

app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
```diff
@@ -214,16 +222,29 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
...
+              values={{ location: probeLocation }}
+              components={{
+                bold: <strong />,
+              }}
+            />
+          </StyledText>
+          {is96Channel && (
+            <Banner
+              type={isWasteChuteOnDeck ? 'error' : 'warning'}
+              size={isOnDevice ? '1.5rem' : SIZE_1}
```
koji: nit
same

```suggestion
              size={isOnDevice ? '1.5rem' : '1rem'}
```

app/src/organisms/PipetteWizardFlows/BeforeBeginning.tsx
```diff
@@ -15,6 +15,7 @@ import {
   LoadedPipette,
   getPipetteNameSpecs,
 } from '@opentrons/shared-data'
+import { useDeckConfigurationQuery } from '@opentrons/react-api-client'
 import { Trans, useTranslation } from 'react-i18next'
```
koji: nit
we can move up `react-i18next`

app/src/organisms/ModuleWizardFlows/AttachProbe.tsx
smb2268: just noticed this - about to push an update!

app/src/assets/localization/en/pipette_wizard_flows.json
```diff
@@ -84,6 +84,8 @@
   "unscrew_and_detach": "Loosen Screws and Detach Mounting Plate",
   "unscrew_at_top": "<block>Loosen the captive screw on the top right of the  carriage. This releases the right pipette mount, which should then freely move up and down.</block>",
   "unscrew_carriage": "unscrew z-axis carriage",
+  "waste_chute_error": "Remove the Waste Chute from the Deck Plate Adaptor before proceeding.",
+  "waste_chute_warning": "If the Waste Chute is installed, remove it from the Deck Plate Adaptor before proceeding.",
```
ecormany: We've gone with `waste chute` (lowercase) as the short name for the `Opentrons Flex Waste Chute`. 
And all Opentrons products use the `-er` spelling of `adapter`. 

```suggestion
  "waste_chute_error": "Remove the waste chute from the deck plate adapter before proceeding.",
  "waste_chute_warning": "If the waste chute is installed, remove it from the deck plate adapter before proceeding.",
```

Also, if there's space here, we may want to add something like `You don't need to remove the deck plate adapter from slot D3.` Skip it if this is already in crowded/wordy UI.

smb2268: I'll update the casing of these! We're short on space on these screens so I'll leave out the second string.

mjhuff (approved):
lgtm!

PR #14035, opened 2023-11-21
refactor(protocol-engine): Delete duplicate DeckConfiguration type

SyntaxColoring:
# Overview

`opentrons.protocol_engine` had two identical type aliases, `DeckConfiguration` and `DeckConfigurationType`. I think this was just an accidental thing left over from #13971.

This PR deletes the older `addressable_areas.DeckConfiguration` in favor of the newer `types.DeckConfigurationType`.

The `# todo` to make this type more semantic (not a `Tuple` alias) still stands. We also still intend to deduplicate this further with other deck configuration types outside of `opentrons.protocol_engine`.

# Test Plan

None needed.

# Review requests

Did I resolve this duplication in the correct direction?

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/addressable_areas.py
+++ b/api/src/opentrons/protocol_engine/state/addressable_areas.py
@@ -1,6 +1,6 @@
 """Basic addressable area data state and store."""
 from dataclasses import dataclass
-from typing import Dict, Set, Union, List, Tuple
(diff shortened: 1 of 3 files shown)


jbleon95 (approved):
Good catch, fixed this locally myself but good to get it into edge.

PR #14036, opened 2023-11-22
fix(hardware-testing): remove partial for gravi

sfoster1:
If we want to use the nozzle configurations in the gravimetric testing, in addition to setting the configuration we need to change the protocols to specify
- pick_up_tip in the right well rather than A1 for alignment
- liquid handling in the right well rather than A1 for alignment
- which nozzle should be active every time we change which tip we're testing


ryanthecoder (approved):
lgtm

PR #14039, opened 2023-11-27
feat(engine): moveToAddressableArea implementation

jbleon95:
# Overview

This PR adds the execution implementation of `moveToAddressableArea` to the Protocol Engine.

The command takes a pipette id, an addressable area name, an optional offset vector, and the standard optional move parameters (force direct, minimum z height, and speed). For every addressable area, it will move to the XY critical point of the pipette to the top center of the addressable area.

This means that it is up to the client (the Python API, the app, step generation) to provide the necessary offsets for dropping tips, alternating tip location, or anywhere else where the default location is not accurate.

In order to accomplish dodging and location tracking for the pipettes, now that addressable areas themselves can be moved to, the `PipetteView` was refactored to make the formerly named `current_well` keep track of addressable areas too, and renamed as `current_location`.

# Test Plan

Tested on Flex using postman to ensure that all types of pipettes (single, multi and 96 channel) move to the correct location for the different types of addressable areas.

# Changelog

- implement `moveToAddressableArea`
- add `base_slot` to PE `AddressableArea` object to enable module dodging and proper behavior
- refactor `PipetteView.state.current_well` to `PipetteView.state.current_location` and make it a union of `CurrentWell` and `CurrentAddressableArea`

# Review requests

Nothing specific, just general organization and naming of parameters.

# Risk assessment

Medium, mostly adding a new command but some geometry and move to well related functions were refactored.


api/src/opentrons/protocol_engine/types.py
```diff
@@ -671,16 +671,27 @@ class PotentialCutoutFixture:
     cutout_fixture_id: str
 
 
+class AreaType(str, Enum):
+    """Strategy to use for labware movement."""
```
SyntaxColoring: Copypasta'd docstring?

api/src/opentrons/protocol_engine/types.py
```diff
@@ -671,16 +671,27 @@ class PotentialCutoutFixture:
     cutout_fixture_id: str
 
 
+class AreaType(str, Enum):
```
SyntaxColoring: Can we avoid subclassing this from `str`? [Rationale here.](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3479896162/Don+t+subclass+Python+enums+from+str+proposed)

```suggestion
class AreaType(Enum):
```

SyntaxColoring: This is a 1:1 binding to the `areaType` field in deck definitions, right? If so, can we move this to `shared-data` to make it easier to keep them in sync?

shared-data/deck/definitions/4/ot3_standard.json
```diff
@@ -203,7 +203,7 @@
       },
       {
         "id": "A4",
-        "areaType": "slot",
+        "areaType": "stagingSlot",
```
SyntaxColoring, line 206: Looks like we can straightforwardly update `is_staging_slot()` now to avoid hard-coding these:

https://github.com/Opentrons/opentrons/blob/d0cfa1a23ccf8267139e15236d10374c4f5afb42/api/src/opentrons/protocol_engine/resources/fixture_validation.py#L30-L32

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -59,6 +59,36 @@ def _get_conflicting_addressable_areas(
     return loaded_areas_on_cutout
 
 
+CUTOUT_TO_DECK_SLOT_MAP: Dict[str, DeckSlotName] = {
```
SyntaxColoring, line 66: What do you think of adding this comment?

```suggestion
# This is a temporary shim while Protocol Engine's conflict-checking code
# can only take deck slots as input.
# Long-term solution: Check for conflicts based on bounding boxes, not slot adjacencies.
# Shorter-term: Change the conflict-checking code to take cutouts instead of deck slots.
CUTOUT_TO_DECK_SLOT_MAP: Dict[str, DeckSlotName] = {
```

api/src/opentrons/protocol_engine/commands/move_to_addressable_area.py
```diff
@@ -21,11 +22,16 @@ class MoveToAddressableAreaParams(PipetteIdMixin, MovementMixin):
     """Payload required to move a pipette to a specific addressable area."""
```
SyntaxColoring: How about this, adapting from #13861:

```suggestion
    """Payload required to move a pipette to a specific addressable area.

    An *addressable area* is a space in the robot that may or may not be usable depending on how
    the robot's deck is configured. For example, if a Flex is configured with a waste chute, it will
    have additional addressable areas representing the opening of the waste chute, where tips and
    labware can be dropped.

    This moves the pipette so all of its nozzles are centered over the addressable area.
    If the pipette is currently configured with a partial tip layout, this centering is over all of
    the pipette's physical nozzles, not just the nozzles that are active.

    The z-position will be chosen to put the bottom of the tips---or the bottom of the nozzles,
    if there are no tips---level with the top of the addressable area.

    When this command is executed, Protocol Engine will make sure the robot's deck is configured
    such that the requested addressable area actually exists. For example, if you request
    the addressable area B4, it will make sure the robot is set up with a B3/B4 staging area slot.
    If that's not the case, the command will fail.
    """
```

* Provide background on addressable areas, since we don't do that elsewhere.
* Describe centering behavior.
* Specify some behavior for partially-configured pipettes. I wrote that the centering should apply over "all physical nozzles" instead of "all active nozzles." This is heavy guesswork on my part for what will be easier for clients, so we should feel free to change it in the future, but I think it's important to have something documented in this merge.

(I'm never quite sure where to put detailed Protocol Engine command docstrings like this. Putting it on the `Params` model makes it show up in the command schema, at least, so that's good.)

api/src/opentrons/protocol_engine/state/motion.py
```diff
@@ -105,9 +109,98 @@ def get_movement_waypoints_to_well(
...
+            addressable_area_name
+        )
+        area_height = self._addressable_areas.get_addressable_area_height(
+            addressable_area_name
+        )
+        destination = base_destination + Point(
+            x=offset.x, y=offset.y, z=offset.z + area_height
+        )
+
+        # TODO I think this is what we always want but honestly unsure
```
SyntaxColoring: Yeah, this looks right for now, given your testing. I'm not sure what should happen for partial tip configuration. I would either remove this `# TODO`, or call out partial tip configuration specifically as a thing to worry about.

```suggestion
```

```suggestion
# TODO Is this correct for partial tip configurations on a 96-channel?
```

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -97,6 +106,7 @@ class PipetteState:
     aspirated_volume_by_id: Dict[str, Optional[float]]
     current_well: Optional[CurrentWell]
     current_deck_point: CurrentDeckPoint
+    current_addressable_area: CurrentAddressableArea
```
SyntaxColoring: Yeah, we started to discuss this in person, but I'm worried about having `current_well` *and* `current_addressable_area`. Especially since `current_addressable_area` is implemented in this PR as being `None` most of the time, which is okay only because `current_well` takes precedence over it.

It seems like we want a single piece of state for the logical location, structured as a 3-way union:

```
      None
    | CurrentWell
          pipette_id: str
          labware_id: str
          well_name: str
    | CurrentAddressableArea
          pipette_id: str
          addressable_area_name: str

File: api/src/opentrons/protocol_engine/state/motion.py [hunk 8]
```diff
@@ -87,6 +90,7 @@ def get_movement_waypoints_to_well(
     ) -> List[motion_planning.Waypoint]:
         """Calculate waypoints to a destination that's specified as a well."""
         location = current_well or self._pipettes.get_current_well()
+        addressable_area_name = self._pipettes.get_current_addressable_area(pipette_id)
```
SyntaxColoring: Could you explain why, when we get the current addressable area, we need to match `pipette_id`, but when getting the current well, we don't?

I think the reasoning behind *not* doing it when we get the current well is that even if it's the *other* pipette that was in that well, that's still a usable heuristic for where *this* pipette is, since they're mounted right next to each other.

For example, on an OT-2, if I aspirate with the left pipette in slot 4, and then I aspirate with the right pipette in slot 9, I expect the robot to dodge around the Thermocycler, taking an L-shaped path between slots 4 and 9, even though I used two different pipettes in those two slots.

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -71,6 +72,14 @@ class CurrentDeckPoint:
...
 
 
+@dataclass(frozen=True)
+class CurrentAddressableArea:
+    """The latest addressable area name and mount the robot has accessed."""
+
+    mount: Optional[MountType]
+    addressable_area_name: Optional[str]
+
+
```
SyntaxColoring: Just as a state representation minimalism thing, I suspect we want this to be an optional dataclass with mandatory fields, instead of a mandatory dataclass with optional fields. Like, there's no case where we have an `addressable_area_name` but not a `mount`, or vice-versa, is there?

SyntaxColoring (commented):
Addressable area accomplishments appealing; acknowledgements, admirations abound.

api/src/opentrons/protocol_engine/commands/move_to_addressable_area.py
sfoster1: I can't imagine why you'd want to center physical nozzles rather than active nozzles, honestly. When you're in a partial tip configuration, those are presumably the only tips you care about.

sfoster1: that said, if we center active nozzles (which I think we should) we need some logic to limit movement to motion bounds so we don't collide with the sides of the robot. That wouldn't be an issue with centering physical nozzles. So I guess that's a pretty good reason.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -302,6 +349,19 @@ def get_addressable_area_position(self, addressable_area_name: str) -> Point:
         position = addressable_area.position
         return Point(x=position.x, y=position.y, z=position.z)
 
+    def get_addressable_area_move_to_location(
+        self, addressable_area_name: str
+    ) -> Point:
+        """Get the move to position (top center) for an addressable area."""
```
sfoster1:
```suggestion
        """Get the move-to position (top center) for an addressable area."""
```

api/src/opentrons/protocol_engine/state/pipettes.py
sfoster1: I agree with this but I also think the other option is to not have it be `None` most of the time. We presumably can get the addressable area that another piece of geometry is loaded on, so we can update this piece of state from the result of the other movement commands.

I think making this a union is only useful if we're saying that it's either None or "some symbolic geometry thing that can be rendered down to an addressable area" - aka, the labware half of the Location type from the python api.

In the long term, now that we have addressable areas loaded so nicely, we can probably abandon this state and compute it from the deck position as "closest addressable area".

jbleon95: I agree in practice with Seth's proposal to continuously track the closest addressable area. But for this PR i've gone with Max's approach of making the two a union for now.

api/src/opentrons/protocol_engine/state/motion.py
jbleon95: You're correct, this is now no longer taken into account with the recent refactoring to `current_well` and `current_addressable_area` I have done

sfoster1 (commented):
This looks good but you should be aware that this will use the current partial nozzle configuration as far as I'm aware - hardware will interpret the critical point as the critical point of whatever the current nozzle configuration is. We may need to add a PHYSICAL_XY_CENTER critical point for this

SyntaxColoring (approved):
Hardcore to the mega!

PR #14040, opened 2023-11-27
chore(api): Update overpressure text

sfoster1:
The overpressure disable button is going to be useful for when people want to purposefully put high pressures inside the pipette air columns. Let's make the name and description a bit more explicit.

Here's what it looks like now:
Screenshot 2023-11-27 at 4 06 01 PM

### DIFF
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -232,8 +232,8 @@ settings = [
     ),
     SettingDefinition(
         _id="disableOverpressureDetection",
-        title="Disable overpressure detection on pipettes.",
-        description="This setting disables overpressure detection on pipettes, do not turn this feature off unless recommended.",
+        title="Disable pipette pressure sensing.",
+        description="This setting disables pressure sensing capability on pipettes. This will prevent errors when overpressuring the pipette. Do not set this setting unless you are intentionally causing >8kPa inside the pipette air channel.",
         robot_type=[RobotTypeEnum.FLEX],
     ),
     SettingDefinition(
--- a/robot-server/Makefile
+++ b/robot-server/Makefile
@@ -129,10 +129,21 @@ lint:
 format:
 	$(python) -m black .
 
-.PHONY: dev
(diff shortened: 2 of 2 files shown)


mikecardarelli (approved):

api/src/opentrons/config/advanced_settings.py
```diff
@@ -232,8 +232,8 @@ class Setting(NamedTuple):
     ),
     SettingDefinition(
         _id="disableOverpressureDetection",
-        title="Disable overpressure detection on pipettes.",
-        description="This setting disables overpressure detection on pipettes, do not turn this feature off unless recommended.",
+        title="Disable pipette pressure sensing.",
+        description="This setting disables pressure sensing capability on pipettes. This will prevent errors when overpressuring the pipette. Do not set this setting unless you are intentionally causing >8kPa inside the pipette air channel.",
```
ecormany, line 236: Because we have puny little monkey brains that are bad at negation, "turn this switch **on** to **disable** the thing" can be confusing. Suggested text for flipping the polarity of the switch:
```suggestion
        title="Pipette pressure sensing",
        description="By default, Flex stops and displays an error when it detects high pressure inside a pipette. Only turn this setting off if you are intentionally causing pressures over 8 kPa inside the pipette air channel.",
```

This parallels the structure for explaining the Home Gantry on Restart setting that we have in the Flex manual (which I think also went through a similar polarity flip during its development):
> By default ,the gantry moves to its home position any time you turn on Flex. Only disable this behavior if you have a reason that the gantry must remain stationary after powering on.

ecormany (commented):
Suggesting that we flip this setting around so when its switch is on, then pressure sensing is on; when it's off, pressure sensing is off. See inline comment for more explanation and proposed new text.

If you don't want to reverse the behavior in code, re-request review and I can do a lighter editing pass on the title and description.

ecormany:
Ope, too slow I guess! 🐌 

As merged, this is fine for a feature flag. If this ever gets promoted to an advanced setting, we should revisit my comments above.

PR #14041, opened 2023-11-27
chore(shared-data): Update the waste chute dimensions for the new lowered lid

SyntaxColoring:
# Overview

Closes RSS-394.

# Test Plan

* [x] Drop some tips with an 8-channel pipette. It should be ~~10 mm below~~ level with the new, shorter lid.
    * The new lid doesn't physically exist at the time of writing, so you'll have to use your 🌈imagination🌈. The lid would be roughly flush with the top of the lid-less waste chute.

# Changelog

Shorten the waste chute's enclosing bounding boxes. They now enclose just the lid-less part, since the new lid won't extend any higher than that.

Lower the 1- and 8-channel drop offsets to be 10 mm below the new lid instead of 10 mm below the old lid.

# Review requests

~~This is pending [one clarification](https://opentrons.slack.com/archives/C049WNGU22H/p1701121268815999?thread_ts=1698264698.866189&cid=C049WNGU22H) from HW on the "max envelope of chute" numbers.~~

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/_waste_chute_dimensions.py
+++ b/api/src/opentrons/protocol_api/_waste_chute_dimensions.py
@@ -7,7 +7,7 @@ TODO: These should be moved into shared-data and interpreted by Protocol Engine.
 from opentrons.types import Point
 
 
-SLOT_ORIGIN_TO_1_OR_8_TIP_A1 = Point(64, 21.91, 144)
+SLOT_ORIGIN_TO_1_OR_8_TIP_A1 = Point(64, 21.91, 115)
 SLOT_ORIGIN_TO_96_TIP_A1 = Point(14.445, 42.085, 115)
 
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
🔢 ⬇️

SyntaxColoring:
~~If I'm reading the `edge` code correctly, I think this is blocked behind us having a `height` field in *cutout fixtures,* separate from the heights of *addressable areas.* The pipette needs to fly above the rim of the waste chute, and then descend into it to meet the addressable area.~~

We do need to solve that problem but it's not a blocker for this PR. With this PR, Protocol Engine will safely clear the rim of the waste chute. It will incorrectly drop tips at the exact height of the rim instead of 10 mm below it, though. Fixing that needs RSS-409.

PR #14045, opened 2023-11-28
refactor(protocol-designer): do not render staging area fixture under trash bin

jerader:
# Overview

Just adding some more logic to when the staging area fixtures are rendered

# Test Plan

Make sure your deck config ff is turned on and create a flex protocol. Add a waste chute and trash bin. Go to the deck map and see that the staging areas are not rendered.

# Changelog

- only create staging area fixture if the additional equipment name is staging area

# Review requests

see test plan

# Risk assessment

low


mjhuff (approved):
lgtm!

PR #14049, opened 2023-11-28
fix(app): protect against null access of gripper data when fw update is ongoing

b-cooper:
# Overview

Prevent attempted access to gripper instrument data in cases where the gripper may be undergoing a fw update

Closes RAUT-882

# Risk assessment
low


smb2268 (approved):
looks good, thanks for fixing!

PR #14051, opened 2023-11-28
fix(app): search for correct subsystem update for strung together flows

smb2268:
Fix RQA-1925

# Overview
If you launch a 96-channel attachment flow via the right pipette card, the `mount` value in the `PipetteWizardFlows` component will be `right`. In `getPipetteWizardSteps` we handle the logic of determining which mount each step in a flow corresponds to, so in the child components of these flows we should always use `currentStep.mount` to determine what pipette we're referring to. We missed this in one instance where we use `mount` to determine which `subsystem` we should check for a firmware update. Because of this, 96-channel pipette attachments initiated via the right pipette card were skipping firmware updates.

# Test Plan
Initiate a 96-channel attachment from the right pipette card and attach a pipette that needs a firmware update. See that the firmware update is initiated after attachment.

# Changelog
Use `currentStep.mount` to determine which subsystem we need to check for a firmware update

# Review requests
Sanity check

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/PipetteWizardFlows/index.tsx
+++ b/app/src/organisms/PipetteWizardFlows/index.tsx
@@ -347,7 +347,9 @@ export const PipetteWizardFlows = (
     modalContent = (
       <FirmwareUpdateModal
         proceed={proceed}
-        subsystem={mount === LEFT ? 'pipette_left' : 'pipette_right'}
+        subsystem={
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
looks good to me!

mjhuff (approved):
Tested and can confirm this fixes the issue! 🚀

PR #14052, opened 2023-11-28
feat(protocol-designer, step-generation): wire up trash bin commands

jerader:
closes RAUT-866 RAUT-883, partially addresses RAUT-876

# Overview

This PR wires up `dispense`, `air_gap`, `blow_out` into the trash bin as well as a bit of clean up followup work.

# Test Plan

Create a flex protocol and make sure the trash bin is added. Add some Transfer steps where you dispense and drop tip into the trash bin. see that nothing errors.

then export the protocol and examine the commands, you should see the `moveToAddressableArea` commands followed by the correct `inPlace` command for the action. You can even import the protocol in the app and it should pass analysis! 

Then reimport the protocol back to PD and see that it imports correctly. (note: a timeline error might appear saying that the drop tip entity is missing, a fix for that will be in a follow up)

# Changelog

- wiring up PD components to have trash bin option in the transfer form
- default aspirate and mix trash steps to null in the migration
- wire up in the compound commands for dispensing, air gapping , or blowing out over a waste chute
- some follow up clean up turning the low volume pipettes into a constant
- refactor `wasteChuteCommandsUtil` and `movableTrashCommandsUtil` to be of return type `CurriedCommandCreator[]` since before, they were treated like compound commands but they are actually more like utils. So it is refactored to be similar to `blowoutUtil`
- fix the broken cypess tests (Migrations, MixSettings, TransferSettings) and fix the bugs found when fixing the cypress test

# Review requests

see test plan

# Risk assessment

low


jerader:
Note: refactoring `wasteChuteCommandsUtil` and `movableTrashCommandsUtil` to return `CurriedCommandCreator[]` keeps being a bigger task since it breaks a bunch of things. So I'm gonna do that in a follow up again 😬 https://opentrons.atlassian.net/browse/RAUT-883

protocol-designer/src/load-file/migration/8_0_0.ts
```diff
@@ -92,7 +92,7 @@ export const migrateFile = (
           ...stepForm,
           aspirate_labware:
             stepForm.aspirate_labware === 'fixedTrash'
-              ? trashId
+              ? null
```
jerader, line 95: after the `moveToAddressableArea` command is finalized, I'll work on adding back the migration cypress tests! So there will be a test case to test this. The reason why we are defaulting to null here is because we will no longer support aspirating or mixing into the trash bin.

step-generation/src/commandCreators/compound/consolidate.ts
```diff
@@ -61,6 +63,12 @@ export const consolidate: CommandCreator<ConsolidateArgs> = (
     }
   }
 
+  //  TODO(jr, 11/28/23): wire this up fully, need to add additionalEquipmentEntities
+  //  to RobotState
+  if (!args.destLabware) {
+    return { errors: [errorCreators.equipmentDoesNotExist()] }
+  }
```
jerader, line 70: I have a ticket for this to do in a follow up https://opentrons.atlassian.net/browse/RAUT-884

step-generation/src/commandCreators/compound/transfer.ts
```diff
@@ -102,13 +103,21 @@ export const transfer: CommandCreator<TransferArgs> = (
     )
   }
 
+  if (
+    !args.destLabware ||
+    (!invariantContext.labwareEntities[args.destLabware] &&
+      !invariantContext.additionalEquipmentEntities[args.destLabware])
```
jerader, line 108: this should work right? For some reason it only works if you import a protocol that is missing the dest labware (rather than deleting the dest labware mid protocol). I also noticed `labwareDoesNotExist` up above on line 99 has the same behavior. Maybe that can be investigated later.

protocol-designer/src/components/StepEditForm/fields/StepFormDropdownField.tsx
```diff
@@ -24,11 +24,9 @@ export const StepFormDropdown = (props: StepFormDropdownProps): JSX.Element => {
     updateValue,
     errorToShow,
   } = props
-  const wasteChuteOption = useSelector(getWasteChuteOption)
+  const disposalOptions = useSelector(getDisposalOptions)
```
shlokamin: it feels weird that the generic component (`StepFormDropDownField`) is injecting options. how come its not the responsibility of its parent component(s) to provide the disposal options? it looks like all of the other options are pass in as props

protocol-designer/src/components/StepEditForm/fields/StepFormDropdownField.tsx
```diff
@@ -24,11 +24,9 @@ export const StepFormDropdown = (props: StepFormDropdownProps): JSX.Element => {
...
   const fullOptions =
-    wasteChuteOption != null && name === 'dispense_labware'
-      ? [...options, wasteChuteOption]
-      : options
+    name === 'dispense_labware' ? [...options, ...disposalOptions] : options
```
shlokamin: related to my comment above, we're polluting this generic component with specific form field concerns. it should be the responsibility of the parent to provide these options

protocol-designer/src/components/StepEditForm/fields/StepFormDropdownField.tsx
jerader: cool, i have been messing around with this the past few days lol and kept running into issues with the fact that the parent component is a redux connect function 🤮 , i think i need to refactor it to not use a connect function at all.

jerader: Okay, refactoring the component worked and i was able to remove that branching logic for the StepFormDropdownField!

protocol-designer/src/load-file/migration/8_0_0.ts
```diff
@@ -121,6 +120,8 @@ export const migrateFile = (
   const newFilteredSavedStepForms = migrateSavedStepForms(
     filteredSavedStepForms
   )
+  console.log('savedStepForms', filteredSavedStepForms)
+  console.log('newFilteredSavedStepForms', newFilteredSavedStepForms)
```
shlokamin: leftover console logs

protocol-designer/src/components/FileSidebar/utils/getUnusedTrash.ts
```diff
@@ -23,9 +23,10 @@ export const getUnusedTrash = (
       ? commands?.some(
           command =>
             command.commandType === 'moveToAddressableArea' &&
-            MOVABLE_TRASH_ADDRESSABLE_AREAS.includes(
+            (MOVABLE_TRASH_ADDRESSABLE_AREAS.includes(
               command.params.addressableAreaName as AddressableAreaName
-            )
+            ) ||
+              command.params.addressableAreaName === 'fixedTrash')
```
shlokamin: nit: can we import this as `FIXED_TRASH_ID` from either shared-data or pd constants?

protocol-designer/src/components/StepEditForm/forms/MoveLiquidForm/SourceDestHeaders.tsx
```diff
@@ -38,26 +38,27 @@ export const SourceDestHeaders = (props: Props): JSX.Element => {
...
+    additionalEquipmentEntities[trashOrLabwareId]?.name === 'wasteChute' ||
+    additionalEquipmentEntities[trashOrLabwareId]?.name === 'trashBin'
 
   React.useEffect(() => {
-    if (isWasteChute) {
+    if (isTrash) {
       propsForFields.dispense_wells.updateValue(['A1'])
     }
-  }, [isWasteChute])
+  }, [isTrash])
```
shlokamin: i thought we dont need this use effect anymore since the move liquid form dependent field updater takes care of this

protocol-designer/src/components/StepEditForm/forms/MoveLiquidForm/SourceDestHeaders.tsx
```diff
@@ -38,26 +38,27 @@ export const SourceDestHeaders = (props: Props): JSX.Element => {
     getAdditionalEquipmentEntities
   )
   const labwareLabel = i18n.t(`form.step_edit_form.labwareLabel.${prefix}`)
-  const wasteChuteOrLabwareId = formData[addFieldNamePrefix('labware')]
-  const isWasteChute =
-    additionalEquipmentEntities[wasteChuteOrLabwareId]?.name === 'wasteChute'
+  const trashOrLabwareId = formData[addFieldNamePrefix('labware')]
+  const isTrash =
```
shlokamin: nit: could we call this `disposalLocation` or something? cc @ncdiehl11 and @SyntaxColoring

protocol-designer/src/steplist/formLevel/stepFormToArgs/moveLiquidFormToArgs.ts
```diff
@@ -85,8 +85,9 @@ export const moveLiquidFormToArgs = (
     fields.aspirate_wellOrder_second
   )
 
-  const dispenseInWasteChute =
-    'name' in destLabware && destLabware.name === 'wasteChute'
+  const dispenseInTrash =
```
shlokamin: same thing here, can we name this something else. maybe `isDispensingIntoWasteChute` or something

step-generation/src/commandCreators/compound/distribute.ts
```diff
@@ -356,17 +356,16 @@ export const distribute: CommandCreator<DistributeArgs> = (
...
-        invariantContext.pipetteEntities[args.pipette].name === 'p50_multi_flex'
-          ? [
-              curryCommandCreator(configureForVolume, {
-                pipetteId: args.pipette,
-                volume: args.volume * destWellChunk.length + disposalVolume,
-              }),
-            ]
-          : []
+      const configureForVolumeCommand: CurriedCommandCreator[] = LOW_VOLUME_PIPETTES.includes(
+        invariantContext.pipetteEntities[args.pipette].name
```
shlokamin, line 360: 😄

protocol-designer/src/components/StepEditForm/forms/MoveLiquidForm/SourceDestHeaders.tsx
jerader: `disposalLocation` makes sense to me. I wasn't sure the best word to encompass both trash bin and waste chute.

protocol-designer/src/components/StepEditForm/forms/MoveLiquidForm/SourceDestHeaders.tsx
jerader: good call, you're right it shouldn't be required. Turns out the issue is when getting fieldLevelErrors, the `dispense_wells` is required to have 1 well. Fixing it now

step-generation/src/commandCreators/compound/transfer.ts
shlokamin: it depends where — step generation shorts circuits as soon as it sees an error so if there was another error before this in another form we wouldnt see anything

shlokamin (approved):
lgtm!

PR #14053, opened 2023-11-29
fix(protocol-designer): fix lint error in protocol-designer

koji:
# Overview
Noticed an linting error while I was trying to update Electron version. 

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


jerader (approved):
thanks koji!

PR #14054, opened 2023-11-29
fix(shared-data): add pickUpOffset, dropOffset params to schema v8 models

sanni-t:
# Overview

JSON Schema v8 includes the `pickUpOffset` and `dropOffset` params to `moveLabware` command. But probably since PD doesn't use these params, we never added them to the json protocol pydantic models. This was blocking ABR testing since we need to use JSON protocols to test using engine commands without the python API.

# Changelog

- just addition of the two params

# Review requests

- Do we need to add these to the v7 the model too? I don't think it'd cause any harm but I don't think it'll be useful either.

# Risk assessment

- This is mainly being added for internal testing but what happens if I hand-edit a PD protocol to add these params to a `moveLabware` command and reupload in PD? Does PD know to ignore it or will it raise errors?

### DIFF
--- a/shared-data/python/opentrons_shared_data/protocol/models/protocol_schema_v8.py
+++ b/shared-data/python/opentrons_shared_data/protocol/models/protocol_schema_v8.py
@@ -65,6 +65,8 @@ class Params(BaseModel):
     pushOut: Optional[float]
     # schema v8 add-ons
     addressableAreaName: Optional[str]
+    pickUpOffset: Optional[OffsetVector]
+    dropOffset: Optional[OffsetVector]
 
 
 class Command(BaseModel):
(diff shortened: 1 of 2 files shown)


jerader (approved):
lgtm! PD should ignore those params upon import. We might need to add these params to the command text in the app though, otherwise the user won't have knowledge of it in the run log, i'd say its low priority though.

shared-data/python/opentrons_shared_data/protocol/models/protocol_schema_v8.py
```diff
@@ -65,6 +65,8 @@ class Params(BaseModel):
     pushOut: Optional[float]
     # schema v8 add-ons
     addressableAreaName: Optional[str]
+    pickUpOffset: Optional[OffsetVector]
+    dropOffset: Optional[OffsetVector]
```
SyntaxColoring, line 67: FYI this is a slight mismatch against the actual command params schema. This `OffsetVector` lets you e.g. omit `x` or make it `null`, but in the actual params, `x`, `y`, and `z` are all required to be `float`s.

I think that's fine because this entire amalgamation is already a mismatch against the actual schema.

SyntaxColoring (approved):
Looks good to merge. Thanks!

> Do we need to add these to the v7 the model too? I don't think it'd cause any harm but I don't think it'll be useful either.

Yes please. I agree that it's not useful, but it seems more conceptually correct.

PR #14056, opened 2023-11-29
refactor(protocol-designer): deck map renders wasteChuteStagingAreaFi…

jerader:
…xture now

closes RAUT-886

# Overview

the deck map component now renders the waste chute staging area fixture instead of rendering the staging area + waste chute fixtures on top of each other.

# Test Plan

Create a flex protocol and make sure the deck config ff is turned on. Look at the deck map and mess around with adding/deleting staging areas and waste chute and trash bin. see that things render as expected.

# Changelog

- add `wasteChuteStagingAreaFixture` and some extra logic to know when to render the different fixtures
- 
# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/DeckSetup/index.tsx
+++ b/protocol-designer/src/components/DeckSetup/index.tsx
@@ -15,6 +15,7 @@ import {
   TrashCutoutId,
   useOnClickOutside,
   WasteChuteFixture,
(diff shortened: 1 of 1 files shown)


mjhuff (approved):
👍

PR #14057, opened 2023-11-29
fix(api): resolve errors when dropping tip racks in trash chute

fsinapi:
# Overview

Dropping tip racks in the trash chute can be problematic. The two main issues that have been happening recently are:
- The notch on the side of the tip rack can catch the gripper paddles when the rack is dropped. The force of this seems to be enough to knock the gripper Z axis out of position and cause it to fall down until it hits something. This is bad.
- Sometimes (possibly depending on the drop height offset?) the tip rack will get momentarily stuck on one of the gripper paddles when it first hits the slope of the chute. Before the tip rack has a chance to slide off and fall down the chute, the gripper will close its paddles and re-grab the tiprack. This is also bad.

This PR addresses those issues with two fixes:
- The gripper Z axis is disengaged before letting go of a labware. This implicitly engages the brake, which is better at holding the gripper in place than the hold current on the motor driver. We have to explicitly re-home the gripper afterwards but that's ok because it's always going that direction anyways.
- We add a 1 second delay before re-closing the gripper jaws after dropping off **only in the trash chute.** From doing a bunch of testing today, that seems like a very safe amount of time to let the labware slide away.

# Test Plan

We dropped a ton of tip racks down trash chutes today. With both fixes, there are no failures

# Changelog

- added a flag to the GripperMovementWaypointsWithJawStatus type to mark if a step is dropping a labware
- when dropping labware, disable the Z_G axis first and then home it to avoid any errors
- delay 1 second after dropping labware in the trash chute to avoid closing the jaws on the labware
- updated tests

# Review requests

Should there be some more specific handling of the last step in a labware movement? The movement after homing the Z axis is technically superfluous but it should just get filtered out by the API for being so short anyways. 

# Risk assessment

Not especially high now that we have the fix for engaging the brake on Z axes.

### DIFF
--- a/api/src/opentrons/motion_planning/types.py
+++ b/api/src/opentrons/motion_planning/types.py
@@ -39,3 +39,4 @@ class GripperMovementWaypointsWithJawStatus:
     position: Point
     jaw_open: bool
     dropping: bool
+    """This flag should only be set to True if this waypoint involves dropping a piece of labware."""
--- a/api/src/opentrons/motion_planning/types.py
+++ b/api/src/opentrons/motion_planning/types.py
@@ -38,3 +38,4 @@ class GripperMovementWaypointsWithJawStatus:
 
     position: Point
     jaw_open: bool
+    dropping: bool
--- a/api/src/opentrons/protocol_engine/execution/labware_movement.py
+++ b/api/src/opentrons/protocol_engine/execution/labware_movement.py
@@ -83,6 +83,7 @@ class LabwareMovementHandler:
         current_location: OnDeckLabwareLocation,
         new_location: OnDeckLabwareLocation,
         user_offset_data: LabwareMovementOffsetData,
+        delay_after_drop: Optional[float] = None,
     ) -> None:
         """Move a loaded labware from one location to another using gripper."""
         use_virtual_gripper = self._state_store.config.use_virtual_gripper
@@ -139,6 +140,8 @@ class LabwareMovementHandler:
                     await ot3api.ungrip()
                     if waypoint_data.dropping:
                         await ot3api.home_z(OT3Mount.GRIPPER)
+                        if delay_after_drop is not None:
+                            await ot3api.do_delay(delay_after_drop)
                 else:
                     await ot3api.grip(force_newtons=labware_grip_force)
                 await ot3api.move_to(
--- a/api/src/opentrons/protocol_engine/execution/labware_movement.py
+++ b/api/src/opentrons/protocol_engine/execution/labware_movement.py
@@ -136,6 +136,11 @@ class LabwareMovementHandler:
             for waypoint_data in movement_waypoints:
                 if waypoint_data.jaw_open:
                     if waypoint_data.dropping:
+                        # This `disengage_axes` step is important in order to engage
+                        # the electronic brake on the Z axis of the gripper. The brake
+                        # has a stronger holding force on the axis than the hold current,
+                        # and prevents the axis from spuriously dropping when  e.g. the notch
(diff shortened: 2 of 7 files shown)


api/src/opentrons/motion_planning/types.py
```diff
@@ -38,3 +38,4 @@ class GripperMovementWaypointsWithJawStatus:
 
     position: Point
     jaw_open: bool
+    dropping: bool
```
SyntaxColoring, line 41: Would you mind adding a comment explaining what this new attribute means? I think it's `True` if this waypoint is the one to open the jaws and release the labware, and `False` for all other waypoints. Is that correct?

No preference from me either way, but does this get conceptually easier if the pause is its own kind of waypoint? So a gripper movement plan would be a sequence of `List[Union[GripperPauseWaypoint, GripperMotionWaypoint]]` or something.

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -134,7 +135,13 @@ async def move_labware_with_gripper(
 
             for waypoint_data in movement_waypoints:
                 if waypoint_data.jaw_open:
+                    if waypoint_data.dropping:
+                        await ot3api.disengage_axes([Axis.Z_G])
```
SyntaxColoring, line 144: Mind adding a comment here explaining why `disengage_axes()` is important? (That the brake is stronger than the hold current, and we need that extra strength to resist getting dragged down by jut-outs in the tip racks.)

api/tests/opentrons/protocol_engine/execution/test_labware_movement_handler.py
```diff
@@ -229,9 +234,13 @@ async def test_move_labware_with_gripper(
         current_location=from_location,
         new_location=to_location,
         user_offset_data=user_offset_data,
+        delay_after_drop=delay,
     )
 
     gripper = OT3Mount.GRIPPER
+    decoy.verify(
+        await ot3_hardware_api.do_delay(delay or 0), times=0 if delay is None else 1
+    )
```
SyntaxColoring: I don't think the `times=0` here is doing quite what you want.

If the subject code has a bug where it does `.do_delay(1)` unconditionally, this assertion will pass because it *did* call `.do_delay(0)` zero times, which I don't think is your intent.

We want to assert that it called `do_delay(<anything>)` zero times. I would write this like this:

```suggestion

    if delay is None:
        decoy.verify(
            await ot3_hardware_api.do_delay(),  # type:ignore[call-arg]
            times=0,
            ignore_extra_args=True
        )
    else:
        decoy.verify(await ot3_hardware_api.do_delay(delay))

```

SyntaxColoring (approved):
Makes sense. My favorite problems are the ones that go away when you do nothing.

Looks great, just a few small requests. Thank you!

api/tests/opentrons/protocol_engine/execution/test_labware_movement_handler.py
fsinapi: thx thx thx

api/src/opentrons/protocol_engine/execution/labware_movement.py
fsinapi: 👍

api/src/opentrons/motion_planning/types.py
fsinapi: 1. Added
2. I think I prefer keeping this as-is because I personally see the delay as an extension of "drop handling" rather than a separate concern. Also there's a really high chance it goes away in favor of a sideways move when dropping off in the trash chute (TBD)

PR #14059, opened 2023-11-29
refactor(app, components): factor out DeckThumbnail from app into ProtocolDeck in components

b-cooper:
# Overview

In order to provide a consistent abstraction for showing the initial state of modules, labware, and
liquids on the BaseDeck according to the contents of an analysis, add ProtocolDeck to the Components
Library to be consumed in app and eventually protocol library

re: RPLC-818

# Changelog

- remove `DeckThumbnail` component from `app` and replace all instances with the new `ProtocolDeck`
- update `BaseDeck` props to reflect current purpose (`moduleLocations` => `modulesOnDeck`, `labwareLocations` => `labwareOnDeck`)
- move utils required for `ProtocolDeck` to appropriate new destinations (e.g. shared-data, components library)

# Review requests

- Sniff test the 4 instances of what was `DeckThumbnail` but is now `ProtocolDeck` in the app. 
  - `ChooseProtocolSlideout`
  - `ProtocolDetail`
  - `ProtocolCard`
  - ODD's `Deck` from `ProtocolDetails` 

# Risk assessment
med, this refactor is far reaching across the app


api-client/src/protocols/utils.ts
```diff
@@ -246,63 +245,6 @@ export function parseInitialLoadedFixturesByCutout(
   )
 }
 
-export function parseAllAddressableAreas(
```
b-cooper, line 249: moved to shared-data js helpers as `getAddressableAreasFromProtocol`

shlokamin (approved):
works as expected! not introduced by this PR but i noticed the trash is not showing up anymore and the background color is darker than it is in 7.0.2 which makes labware harder to see cc @brenthagen

brenthagen (approved):

brenthagen:
> works as expected! not introduced by this PR but i noticed the trash is not showing up anymore and the background color is darker than it is in 7.0.2 which makes labware harder to see cc @brenthagen

There's another chunk of work to do to account for the legacy load labware fixed trash both client-side and robot-side. Client-side, the plan is to special case that command in `getAddressableAreasInProtocol`.

PR #14060, opened 2023-11-29
feat(api): add TrashBin and WasteChute API support

ncdiehl11:
closes [RSS-404](https://opentrons.atlassian.net/browse/RSS-404)
closes [RSS-405](https://opentrons.atlassian.net/browse/RSS-405)
closes [RSS-406](https://opentrons.atlassian.net/browse/RSS-406)

# Overview

Dropping tips previously required a reference _well_, either explicitly or implicitly. In the implicit case, the protocol referenced a single-well "fixed trash" labware object that was loaded automatically upon protocol instantiation. Here, I add/refine new classes for TrashBin and WasteChute "disposal locations," add a protocol context method for loading each of these entities, and add instrument support for dropping tips in these non-labware entities. The protocol core tracks the presence of loaded TrashBins and WasteChutes to accommodate tip drops in the event that the user did not specify a drop tip location or tie a trash_container to a their loaded pipette.

TODO: utilize the new MoveToAddressableArea API command to avoid calculating and moving to hard-coded offsets

# Test Plan

- TODO:  resolve failing API tests that are failing because of the unexpected absense of a fixed trash on a protocol's labware core for Flex protocols on API version > 2.16. It sounds like this is a known issue that @jbleon95 and @SyntaxColoring have also run into.

# Changelog

# Risk assessment

Medium. Notably, this PR modifies the InstrumentContext's drop_tip method and its implicated lower level engine methods by searching for a valid drop tip location in the event that no fixed trash is loaded upon protocol initialization.

[RSS-404]: https://opentrons.atlassian.net/browse/RSS-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSS-405]: https://opentrons.atlassian.net/browse/RSS-405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSS-406]: https://opentrons.atlassian.net/browse/RSS-406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1447,22 +1461,32 @@ def tip_racks(self, racks: List[labware.Labware]) -> None:
 
     @property  # type: ignore
     @requires_version(2, 0)
-    def trash_container(self) -> labware.Labware:
-        """The trash container associated with this pipette.
+    def trash_container(
+        self
+    ) -> Union[labware.Labware, TrashBin, WasteChute]:
+        """The trash container(s) associated with this pipette.
```
ecormany:
```suggestion
        """The trash containers associated with this pipette.
```
Always favor the plain plural over `(s)`.

If we want to be even more explicit about this object, and further moot the plural issue: `The list of any trash containers...`

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1447,22 +1461,32 @@ def tip_racks(self, racks: List[labware.Labware]) -> None:
...
 
         This is the property used to determine where to drop tips and blow out liquids
         when calling :py:meth:`drop_tip` or :py:meth:`blow_out` without arguments.
 
-        By default, the trash container is in slot A3 on Flex and in slot 12 on OT-2.
+        By default, the trash container is in slot 12 on OT-2 and in slot A3 on Flex
+        for API Versions < 2.16. For API Versions >= 2.16, we search the protocol
+        core's set of disposal locations for the first valid location if it exists.
```
ecormany: If at all possible, we should avoid mentioning the API core at all in docs. Users shouldn't really have to worry about how we've implemented the API.

As for the new behavior, how does the search proceed? Does it find all trash containers? This description makes it sound like it stops after it finds one. But what's the slot order, and does it differ between OT-2 and Flex?

Finally, we decided to list Flex first in docs whenever applicable, so I'd go back to that order in the first sentence.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -438,6 +439,20 @@ def load_adapter_from_definition(
             location=location,
         )
 
+    @requires_version(2, 16)
+    # TODO: Confirm official naming of "trash bin".
```
ecormany: Confirming: it is _trash bin_.

api/src/opentrons/protocol_api/validation.py
```diff
@@ -261,6 +265,27 @@ def ensure_module_model(load_name: str) -> ModuleModel:
...
+    if robot_type == 'OT-2 Standard':
+        raise InvalidTrashBinLocationError('Cannot load trash on OT-2.')
+
+    validated_slot = ensure_and_convert_deck_slot(deck_slot)
+    slot_name_ot3 = validated_slot.to_ot3_equivalent()
+    valid_trash_bin_slots = {'A1', 'B1', 'C1', 'D1', 'A3', 'B3', 'C3', 'D3'}
+    if slot_name_ot3 not in valid_trash_bin_slots:
+        raise InvalidTrashBinLocationError(
+            f'Invalid location for trash bin: {slot_name_ot3}.\n'
+            f'Valid slots: {",".join(valid_trash_bin_slots)}')
```
ecormany: Instead of listing these, sometimes we say "The trash bin can go in any slot in column 1 or 3."

ecormany (commented):
Some docstring comments.

I think the spot that will need the most editing/explanation is the new behavior of `trash_container`, how it's determined, and whether its return value could be a breaking change when updating `apiLevel` on existing protocols.

api/src/opentrons/protocol_api/instrument_context.py
ncdiehl11: I think it's a good call to avoid mentioning the core. 

As far as hierarchy, we have decided for the time being on accessing the first _loaded_ disposal location (trash bin or waste chute), regardless of slot position. If a user implicitly calls `drop_tip` on an OT-2 OR on a Flex running API version < 2.16, we auto drop in the presumed fixed trash.

I'll fix the order of Flex vs OT-2!

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1447,22 +1467,38 @@ def tip_racks(self, racks: List[labware.Labware]) -> None:
...
-        By default, the trash container is in slot A3 on Flex and in slot 12 on OT-2.
+        On a Flex running a protocol with API Version >= 2.16, `trash_container` returns
+        the first TrashBin or WateChute object loaded in the protocol.
+        On a Flex running a protocol with API Version < 2.16, `trash_container` returns
+        a single-well fixed trash labware in slot D3.
+        On a an OT-2, `trash_container` always returns a single-well fixed trash labware
+        in slot 12.
+
+        .. versionchanged:: 2.16
+            Adds the support for TrashBin and WasteChute objects.
```
ecormany: Great! This suggestion is mostly formatting cleanup. (RST double backticks, sigh…) Also clarifying that it doesn't return anything because it's a property, not a method.

```suggestion
        On a Flex running a protocol with API version 2.16 or higher, ``trash_container`` is
        the first ``TrashBin`` or ``WasteChute`` object loaded in the protocol.
        On a Flex running a protocol with API version 2.15, ``trash_container`` is
        a single-well fixed trash labware in slot D3.
        On a an OT-2, ``trash_container`` is always a single-well fixed trash labware
        in slot 12.

        .. versionchanged:: 2.16
            Added support for ``TrashBin`` and ``WasteChute`` objects.
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -922,6 +924,12 @@ def drop_tip(
             - As a :py:class:`~.types.Location`. For example, to drop a tip from an
               unusually large height above the tip rack, you could call
               ``pipette.drop_tip(tip_rack["A1"].top(z=10))``.
+            - As a :py:class:`.TrashBin`. This uses a default location relative to the
```
ecormany, line 930: Will we be making these classes public? If not, let's use descriptive terms instead of a cross-reference. And I think we can drop the examples:
`- As a trash bin or waste chute object. This uses a default location relative to the object.`

And add the same ``..versionchanged`` statement from `trash_container`

ncdiehl11: To my knowledge, both `TrashBin` and `WasteChute` will be private classes (cc: @SyntaxColoring). I'll make this change.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -394,48 +395,78 @@ def _drop_tip_in_place(self, home_after: Optional[bool]) -> None:
...
 
-    def drop_tip_in_waste_chute(
-        self, waste_chute: WasteChute, home_after: Optional[bool]
+    def drop_tip_in_disposal_location(
+        self, disposal_location: Union[TrashBin, WasteChute], home_after: Optional[bool]
     ) -> None:
         # TODO: Can we get away with implementing this in two steps like this,
         # or does drop_tip() need to take the waste chute location because the z-height
         # depends on the intent of dropping tip? How would Protocol Designer want to implement
         # this?
```
SyntaxColoring: [We can delete this `# TODO`](https://opentrons.slack.com/archives/CSPMY9048/p1701197302581729?thread_ts=1701194571.926949&cid=CSPMY9048) while we're here.

```suggestion
```

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -394,48 +395,78 @@ def _drop_tip_in_place(self, home_after: Optional[bool]) -> None:
...
         self,
-        waste_chute: WasteChute,
+        disposal_location: Union[TrashBin, WasteChute],
         force_direct: bool,
         speed: Optional[float],
     ) -> None:
-        if self.get_channels() == 96:
-            slot_origin_to_tip_a1 = _waste_chute_dimensions.SLOT_ORIGIN_TO_96_TIP_A1
+        # NOTE: only invoked by `drop_tip_in_disposal_location`, so does not currently accommodate gripper
+
```
SyntaxColoring:
```suggestion
```

Your call, but I think it's okay to leave this note out because we're in `InstrumentContext`, which isn't expected to accommodate the gripper in general. (PAPI's `InstrumentContext` is just for pipettes, despite the gripper also being called an "instrument" elsewhere in our product.)

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -394,48 +395,78 @@ def _drop_tip_in_place(self, home_after: Optional[bool]) -> None:
...
+        map_trash_bin_addressable_area = {
+            'A1': 'movableTrashA1',
+            'B1': 'movableTrashB1',
+            'C1': 'movableTrashC1',
+            'D1': 'movableTrashD1',
+            'A3': 'movableTrashA3',
+            'B3': 'movableTrashB3',
+            'C3': 'movableTrashC3',
+            'D3': 'movableTrashD3',
+        }
```
SyntaxColoring: Let's deduplicate this with `ensure_trash_bin_location()` to avoid the hazard of a `KeyError` if the sets of allowed locations ever get out of sync with each other.

`ensure_trash_bin_location()` would either:

* Raise, if the location is invalid.
* Return a string like `movableTrashA1` if the location is valid. The returned string would correspond to the input location.

Then I guess `load_trash_bin()` would store the returned string in the new `TrashBin()` object. By the time we reach here, that lets this method simply do `disposal_location._addressable_area_name` instead of having to do its own mapping.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -394,48 +395,78 @@ def _drop_tip_in_place(self, home_after: Optional[bool]) -> None:
...
+
+        # map pipette channel to waste chute addressable area
+        map_waste_chute_addressable_area: {
+            1: {
+                'addressableAreaName': '1and8ChannelWasteChute',
+                'offset': AddressableOffsetVector(64.0, 21.91, 144.0)
```
SyntaxColoring: Let's delete the `offset` keys from `map_waste_chute_addressable_areas`. These numbers (which I think you got from [here](https://github.com/Opentrons/opentrons/blob/b540785a43f7c1ed4a8e3ea67c9852e4645d29c9/shared-data/deck/definitions/4/ot3_standard.json#L371)) are already known to Protocol Engine and built in to the `moveToAddressableArea` command.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -394,48 +395,78 @@ def _drop_tip_in_place(self, home_after: Optional[bool]) -> None:
...
-        # Normally, we use a 10 mm margin. (DEFAULT_GENERAL_ARC_Z_MARGIN.) Unfortunately, with
-        # 1000µL tips, we have slightly not enough room to meet that margin. We can make the margin
-        # as big as 7.5 mm before the motion planner raises an error. So, use that reduced margin,
-        # with a little more subtracted in order to leave wiggle room for pipette calibration.
-        minimum_z = _waste_chute_dimensions.ENVELOPE_HEIGHT + 5.0
-
-        self.move_to(
-            Location(destination_point, labware=None),
-            well_core=None,
+            slot_name = DeckSlotName.SLOT_D3
```
SyntaxColoring: This `slot_name` variable looks unused.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -394,48 +395,78 @@ def _drop_tip_in_place(self, home_after: Optional[bool]) -> None:
...
+            miminum_z_height = _waste_chute_dimensions.ENVELOPE_HEIGHT + 5.0
```
SyntaxColoring: Let's delete this `minimum_z_height` override now.

From the comment in the old code, this override *reduced* the margin by which we overfly obstacles on the deck. But:

* That was only necessary because of inconvenient hardware dimensions, which we are changing (RSS-394).
* According to [this description](https://github.com/Opentrons/opentrons/blob/b540785a43f7c1ed4a8e3ea67c9852e4645d29c9/api/src/opentrons/protocol_engine/commands/pipetting_common.py#L56-L63) of Protocol Engine's `minimumZHeight` field, it can only increase the margin, not reduce it.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -939,9 +947,18 @@ def drop_tip(
...
                 alternate_drop_location = True
+            if isinstance(trash_container, labware.Labware):
+                well = self.trash_container.wells()[0]
+            else:  # implicit drop tip in disposal location, not well
+                self._core.drop_tip_in_disposal_location(
+                    trash_container,
+                    home_after=home_after
+                )
+                self._last_tip_picked_up_from = None
+                return self
```
SyntaxColoring: Looks like this code is conflicting with the code on line 981?

api/src/opentrons/protocol_api/instrument_context.py
SyntaxColoring: I've been going back and forth and I can see it going either way.

In favor of public:

* Eventually we may need to support custom offsets like `drop_tip(waste_chute.top(z=10))`. (RSS-391.) Then, the `.top()` method needs to be documented somewhere.
* These classes will show up in various method signatures.
* For user code to be able to annotate the type of `TrashBin` and `WasteChute` variables, the `TrashBin` and `WasteChute` types need to be public.

In favor of private:

* General uncertainty and YAGNI.

I lean towards making them public and I'm comfortable making that executive decision if nobody objects.

SyntaxColoring: (Note: If we do make them private, we need to remove the re-exports from `protocol_api/__init__.py`.)

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1447,22 +1467,38 @@ def tip_racks(self, racks: List[labware.Labware]) -> None:
 
     @property  # type: ignore
     @requires_version(2, 0)
-    def trash_container(self) -> labware.Labware:
-        """The trash container associated with this pipette.
+    def trash_container(
+        self
+    ) -> Union[labware.Labware, TrashBin, WasteChute]:
+        """The trash containers associated with this pipette.
```
SyntaxColoring: Sorry for the back-and-forth: this should just say "The trash container", singular. It only returns one.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1447,22 +1467,38 @@ def tip_racks(self, racks: List[labware.Labware]) -> None:
...
+        On a Flex running a protocol with API version 2.15, ``trash_container`` is
+        a single-well fixed trash labware in slot D3.
+        On a an OT-2, ``trash_container`` is always a single-well fixed trash labware
+        in slot 12.
+
+        .. versionchanged:: 2.16
+            Added support for ``TrashBin`` and ``WasteChute`` objects.
         """
+
         if self._trash is None:
```
SyntaxColoring: Can we rename `self._trash` to `self._user_specified_trash`?

SyntaxColoring (commented):
This is coming along great. Thank you for your help!

Various comments here and there. I know this is still a draft so you might already be addressing some of these.

api/src/opentrons/protocol_api/instrument_context.py
ncdiehl11: How so? We enter the block beginning line 956 if `drop_tip` is called without a location argument, and we check to see if there's a valid loaded disposal_location on the protocol core. The block beginning line 981 executes if the explicitly passed argument for `drop_tip` `location` is an instance of `TrashBin` or `WasteChute`

api/src/opentrons/protocol_api/core/engine/instrument.py
ncdiehl11: great point. making that change.

api/src/opentrons/protocol_api/instrument_context.py
ncdiehl11: @ecormany if we settle on public (seems like we will), should we restore the examples in the docstring?

ecormany: I think we can leave the examples out of the reference entry here. "Specify location…as a `TrashBin`" gives all the information contained in `location=trash_bin` — plus that variable name could be anything depending on how the user loads it.

The "as a Well" and "as a Location" examples are more useful, because they show how to reference a well or location inline with a method, not premade and saved in a variable. And `reservoir` and `tip_rack` are defined in our [Protocol Template](https://docs.opentrons.com/v2/new_examples.html#protocol-template).

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -450,7 +463,9 @@ def load_waste_chute(
             raise NotImplementedError(
                 "The waste chute staging area slot is not currently implemented."
             )
-        return WasteChute(with_staging_area_slot_d4=with_staging_area_slot_d4)
+        waste_chute = WasteChute(with_staging_area_slot_d4=with_staging_area_slot_d4)
```
CaseyBatten, line 473: Waste Chute Object probably needs modifying (not in this PR but before release) to no longer utilize this staging area slot parameter. This information is inferred from any referential call to the staging area slots elsewhere.

api/src/opentrons/protocol_api/validation.py
```diff
@@ -261,6 +265,27 @@ def ensure_module_model(load_name: str) -> ModuleModel:
...
 
 
+def ensure_trash_bin_location(
+    deck_slot: Union[int, str],
+    robot_type: RobotType
+) -> DeckSlotName:
+
+    """Ensure trash bin load location is valid."""
+
+    if robot_type == 'OT-2 Standard':
```
CaseyBatten: Are we sure this is how we want to do this? Currently the thinking is that OT-2 will have two deck configurations, an OT-2 deck with all slots and an OT-2 deck with a trash. How do we differentiate the deck config types for a v2.16 OT-2? Would doing something like ctx.load_trash_bin('ot2_short_trash', 12) not be a valid way to let us differentiate the decks?

api/src/opentrons/protocol_api/instrument_context.py
CaseyBatten: > * Eventually we may need to support custom offsets like `drop_tip(waste_chute.top(z=10))`. ([RSS-391](https://opentrons.atlassian.net/browse/RSS-391).) Then, the `.top()` method needs to be documented somewhere.

This is also in line with what is discussed int the [RSS-404](https://opentrons.atlassian.net/browse/RSS-404) requirements and the initial [PAPI documentation regarding deck configuration](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3887759414/Deck+configuration+Python+Protocol+API), so the public facing methodology would be most in-line with our planned behavior so far.

SyntaxColoring: Discussed in-person. We'll make them public.

api/src/opentrons/protocol_api/waste_chute.py
SyntaxColoring, line 1: Sorry for the back-and-forth: this should still be named `_waste_chute.py` because we only want people to be able to do `from opentrons.protocol_api import WasteChute`, not `from opentrons.protocol_api.waste_chute import WasteChute`.

api/src/opentrons/protocol_api/trash_bin.py
SyntaxColoring, line 1: Ditto my other comment about keeping this file named `_trash_bin.py` instead of `trash_bin.py`.

api/src/opentrons/protocols/advanced_control/transfers.py
```diff
@@ -800,9 +800,10 @@ def _after_dispense(self, dest, src, is_disp_next=False):  # noqa: C901
                 self._strategy.blow_out_strategy == BlowOutStrategy.TRASH
                 or self._strategy.disposal_volume
             ):
-                yield self._format_dict(
-                    "blow_out", [self._instr.trash_container.wells()[0]]
-                )
+                if isinstance(self._instr.trash_container, Labware):
+                    yield self._format_dict(
+                        "blow_out", [self._instr.trash_container.wells()[0]]
+                    )
```
SyntaxColoring, line 806: This change looks a bit dangerous. If the trash isn't a labware, we're skipping the blowout.

We should either:

* Get this code to support non-labware trashes in this PR.
* Leave a `# TODO`, raise a `NotImplementedError`, and come back to this in another PR before launch.

SyntaxColoring (changes requested):

api/tests/opentrons/protocol_api/core/engine/test_protocol_core.py
```diff
@@ -1223,6 +1225,7 @@ def test_load_module_raises_wrong_location(
         )
 
 
+@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```
SyntaxColoring, line 1224:
```suggestion
# APIv2.15 because we're expecting a fixed trash.
@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```

api/tests/opentrons/protocol_api/core/engine/test_protocol_core.py
```diff
@@ -1049,6 +1050,7 @@ def test_add_labware_definition(
         ),
     ],
 )
+@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```
SyntaxColoring, line 1050:
```suggestion
# APIv2.15 because we're expecting a fixed trash.
@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```

api/tests/opentrons/protocol_api/core/engine/test_protocol_core.py
```diff
@@ -186,6 +187,7 @@ def test_api_version(
 #     assert result == ot2_standard_deck_def["locations"]["orderedSlots"][5]
 
 
+@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```
SyntaxColoring, line 192:
```suggestion
# APIv2.15 because we're expecting a fixed trash.
@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```

api/tests/opentrons/protocol_api/core/engine/test_protocol_core.py
```diff
@@ -146,6 +146,7 @@ def mock_sync_hardware_api(decoy: Decoy) -> SyncHardwareAPI:
 
 
 @pytest.fixture
+@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```
SyntaxColoring, line 150:
```suggestion
# APIv2.15 because we're expecting a fixed trash.
@pytest.mark.parametrize("api_version", [APIVersion(2, 15)])
```

SyntaxColoring (commented):
Pinning the tests to APIv2.15 seems good but let's explain what that's doing.

CaseyBatten:
Worked through various issues live and ran a protocol with loaded trash and loaded waste chute live, testing for the following cases:
Default trash drop tip, Specified trash drop tip, pipette trash container variable empty drop tip, transfer command default trash drop tip, multiple loaded movable trashes.

api/src/opentrons/protocols/advanced_control/transfers.py
SyntaxColoring: https://opentrons.atlassian.net/browse/RSS-414

SyntaxColoring (approved):
Good to merge if we need to unblock v7.1.0 alpha testing. We will need to come back to this and add tests covering PAPI. ~~The timing-out G-code tests are also concerning; we need to confirm whether those are the fault of this PR.~~ [Edit: G-code tests are passing now, yay.]

PR #14061, opened 2023-11-30
feat(protocol-designer): no longer put off-deck button behind ff

jerader:
closes RAUT-887

# Overview

To prepare for the 1st 8.0 release candidate, this is to no longer put the off-deck button behind the ff and then rename the ff to only encompass multi-tip support

# Test Plan

Create a flex protocol and go to deck setup and see that the off deck button is visible and works correctly!

Create a protocol for the ot-2 and notice the off-deck button there as well.

# Changelog

- remove off-deck button from being behind a feature flag and rename the ff to something that doesn't include the off-deck button

# Review requests

see test plan

# Risk assessment

low


mjhuff (approved):
Looks great! 🥳

PR #14062, opened 2023-11-30
feat(api): raise error when loading pipette incompatible for robot type

sanni-t:
Closes RQA-1917 and RSS-341

# Overview

Adds pipette w/ robot type validation during pipette loading. This mainly affects analysis since loading the wrong pipette will always fail during actual run since incompatible pipettes can't be physically connected to the robot anyway.

# Test Plan

Write protocols for OT2 & Flex and have them load incompatible pipettes and verify that the analyses fail.

# Changelog

- added validation in `loadPipette`

# Review requests

- while testing I realized that `p1000_96` pipette types didn't have the correct generation associated with it because it doesn't follow the expected naming. How has this not caused a problem in hardware controller?

# Risk assessment

Low. Only adds param validation


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -791,7 +791,6 @@ def load_instrument(
                              `mount` (if such an instrument exists) should be
                              replaced by `instrument_name`.
         """
-        # TODO (spp: 2023-08-30): disallow loading Flex pipettes on OT-2 by checking robotType
```
SyntaxColoring, line 794: Hmm, it looks like if we only implement this validation in Protocol Engine, PAPIv≤2.13 protocols can erroneously still load Flex pipettes.

```python
requirements = {"apiLevel": "2.13"}

def run(protocol):
    protocol.load_instrument("flex_1channel_1000", mount="left")
```

api/src/opentrons/protocol_engine/commands/load_pipette.py
```diff
@@ -61,13 +69,43 @@ class LoadPipetteImplementation(
...
+            )
+        ) or (
+            (
+                robot_type == RobotTypeEnum.OT2
+                and pipette_generation
+                not in [PipetteGenerationType.GEN1, PipetteGenerationType.GEN2]
+            )
+        ):
+            raise InvalidSpecificationForRobotTypeError(
+                f"Cannot load a {pipette_generation.value} pipette on {robot_type.name} robot."
```
SyntaxColoring: Heads up that this will capitalize and punctuate the robot names weirdly. `FLEX` instead of `Flex`, and `OT2` instead of `OT-2`.

SyntaxColoring (approved):
TY!

sfoster1 (approved):
Looks good to me!

SyntaxColoring:
Oh heads up that you have some CI errors, though.

api/src/opentrons/protocol_engine/commands/load_pipette.py
ecormany: I'd love to get an indefinite article in there, although I get that English makes the code more complicated. i.e., "Cannot load **a Flex** pipette on **an OT-2** robot." and "Cannot load **an OT-2** pipette **a Flex** robot."

api/src/opentrons/protocol_api/protocol_context.py
sanni-t: 🤦🏻‍♀️

shared-data/python/opentrons_shared_data/robot/__init__.py
```diff
@@ -23,3 +23,11 @@ def load(
...
             return cast(RobotDefinition, defn)
     raise KeyError(robot_type)
+
+
+def user_facing_robot_type(robot_type: RobotType, include_article: bool = False) -> str:
+    """Appropriately formatted robot type string for use in user-facing messages."""
+    if robot_type == "OT-2 Standard":
+        return "an OT-2" if include_article else "OT-2"
+    elif robot_type == "OT-3 Standard":
+        return "a Flex" if include_article else "Flex"
```
ecormany, line 33: Awesome, thank you! I'm sure this is only the first time of many that this code will get used.

ecormany (approved):
Words are good! :shipit:

shared-data/python/opentrons_shared_data/robot/__init__.py
sanni-t: Thanks to @SyntaxColoring !

PR #14063, opened 2023-11-30
Remove redundant pd e2e tests

DerekMaggio:
# Overview

- Removes all Cypress e2e tests that are redundantly testing logic covered by the Jest unit tests 
- Combines Cypress visual inspection e2e tests into a single test. This takes advantage of Cypress's asynchronous testing strategy, to reduce setup and actual test run time.
  - migration tests are turned off in this PR
  - This PR e2e test run time = 7 seconds
  - [Another PR without redundant tests removed](https://github.com/Opentrons/opentrons/actions/runs/7052114022/job/19196600405) e2e test run time - migration run time = 89 seconds
  - **Time % decrease = 92%** 

# E2E Test Auditing Process

- audit Jest test cases to find Cypress cases that already have existing coverage
    - attempt to find test cases that cover each assertion of the Cypress tests
    - in this PR, provide Cypress assertion explanation and direct link to pre-existing Jest test coverage
     - if unable to confirm coverage, note in this PR, and ask for help
     - if there is a Cypress test that provides coverage for a hole in Jest coverage, note in this PR. Evaluate whether it makes sense to keep coverage in Cypress or convert to Jest

# Audit Results

## Summary 
- The only especially useful Cypress spec is the migration test
- The batchEdit, mixSettings, and transferSettings tests are all completely redundant and should be removed
- The home, settings, and sidebar specs have some possible use, but should be refactored to:
  - Take advantage of Cypress being able to perform multiple asynchronous asserts at the same time

### batchEdit.spec.js

- Using Batch Edit to select a range with Shift Click -> [ConnectedStepItem.test.tsx](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/containers/__tests__/ConnectedStepItem.test.tsx#L364-L387)
- Changing properties with Batch Edit -> [makeBatchEditFieldProps.test.ts](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/components/BatchEditForm/__tests__/makeBatchEditFieldProps.test.ts#L24-L61)
- Should only allow multistep setting editing for transfer or mix steps -> [ui/steps/test/selectors.test.ts](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/ui/steps/test/selectors.test.ts#L378-L393)
- Expanding / Selecting / Deselecting / Deleting / Duplicating with Batch Editing Toolbar -> [MultiSelectToolbar.test.tsx](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/components/steplist/__tests__/MultiSelectToolbar.test.tsx)

### mixSettings.spec.js

- Correct parameters are disabled when editing 2 mix steps that have different labware (aspirate & dispense) -> [ui/steps/test/selectors.test.ts](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/ui/steps/test/selectors.test.ts#L1046-L1108)
- No parameters disabled when editing 2 mix steps that have the same labware and same pipettes -> [ui/steps/test/selectors.test.ts](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/ui/steps/test/selectors.test.ts#L1247-L1269)
- Invalid input into delay field -> [getDelayData.test.ts](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/steplist/formLevel/stepFormToArgs/test/getDelayData.test.ts)
- Verify indeterminate values -> all over [ui/steps/test/selectors.test.ts](https://github.com/Opentrons/opentrons/blob/edge/protocol-designer/src/ui/steps/test/selectors.test.ts)

### transferSettings.spec.js

- same as mixSettings.spec.js

### home.spec.js

Not sure how useful this is as an e2e test. Might be useful for a simple smoke test to run against the live website to ensure it is up and not utterly broken. Marking to add to a sanity test

### settings.spec.js

This is one that I think might be missing test coverage in Jest, at least I couldn't find it. **BUT**, I don't think a Cypress test is the way to add test coverage. I think it should be a Jest test. 
As far as all the validation of page content. Might be in the same vein as home.spec.js and be useful to run against the live website. 

### sidebar.spec.js

Marking to add to a sanity test

# Test Plan

- [x] go through audit process
- [x] compile results
- [x] verify any changes to test cases run both locally and in CI
- [x] discuss with @y3rsh any updated Cypress test cases and get input
- [ ] share out any questions about coverage or testing strategy

# Changelog

- Remove e2e-test dependency on js unit test in Github Action
  - Not gaining anything by doing this. Only delaying the start of e2e tests. I think job concurrency is the solution here
- Add `open-cypress` Makefile command to open interactive test editor
- Completely remove batchEdit.spec.js, mixSettings.spec.js, transferSettings.spec.js
- Combine home.spec.js, settings.spec.js, and sidebar.spec.js into visual_smoke_test.spec.js 

# Review requests

So I started learning, Typescript, React, Redux, Cypress, and Jest like 3 days ago. 0 prior experience with any of this. I did my best to ensure that all my summations of the e2e test were correct, but having a 2nd opinion would be helpful.

# Risk assessment

Low, with the caeveat that someone with experience in Protocol Designer's Jest testing validates my conclusions

# Proposed Followups

- Add concurrency between unit and e2e tests so if unit tests fail e2e tests get canceled.


shlokamin (changes requested):
i think there was a miscommunication when talking about "reducing redundant tests". my intention in that statement was to reduce redundant test coverage _across existing end to end tests_. 

it is very normal (and good) to have your end to end tests cover some of what your unit tests cover. it is bad, however, to have your end to end tests cover what your existing end to end tests cover.

this is because unit tests are cheap, reliable, and have additional benefits to system reliability like encouraging good api design.

end to end tests, on the other hand, are expensive, flakey, and harder to maintain. we want them to test the most critical user flows, but we don't want them testing the same user flows multiple times. 

that being said, the idea here is to audit the existing end to end tests, figure out where assertions are being duplicated, and merge the existing tests down to fewer tests. for example, if you look at `transferSettings.spec` and `mixSettings.spec`, you'll see that they are doing mostly the same things. these should really be combined into one test that tests both kinds of steps. i'd take this idea even further and say we should have one test like this that uses all step types.

i think we should pump the breaks on this and come up with clear guidelines before preceding so we don't waste time cc @nusrat813

.github/workflows/pd-test-build-deploy.yaml
```diff
@@ -80,7 +80,6 @@ jobs:
 
   e2e-test:
     name: 'pd e2e tests'
-    needs: ['js-unit-test']
```
shlokamin, line 83: we want this to depend on unit tests, its expensive to kick of an e2e testing process and i dont think we should unless unit tests pass

protocol-designer/cypress/integration/migrations.spec.js
```diff
@@ -6,7 +6,7 @@ const semver = require('semver')
 
 // TODO: (sa 2022-03-31: change these migration fixtures to v6 protocols once the liquids key is added to PD protocols
 // https://github.com/Opentrons/opentrons/issues/9852
-describe('Protocol fixtures migrate and match snapshots', () => {
+describe.skip('Protocol fixtures migrate and match snapshots', () => {
```
shlokamin, line 9: this is the most important PD e2e test, we definitely do not want to turn this off

DerekMaggio: Sorry, I didn't mean keep it turned off. I saw Jethary had them commented out. I was going to merge the corrected branch into mine once they were merged. My bad on this miscommunication

.github/workflows/pd-test-build-deploy.yaml
DerekMaggio: Got it

DerekMaggio:
@shlokamin, thank you for the insight. I appreciate you taking the time to look at it.
When we come back around to it, the next iteration will include all this.

PR #14065, opened 2023-11-30
fix(api, shared-data): avoid failures when dropping tip rack in trash chute

fsinapi:
# Overview

PR #14057 helps with most tip rack drop issues but there are still a few cases where the tip rack falls in _just_ the perfect position to be resting against both the bottom of the trash chute and one of the gripper paddles. To fully eliminate any issues when dropping a tip rack, we are adding a horizontal movement "post-drop slide" whenever something gets dropped in the trash chute. 

# Test Plan

Ran lots of iterations of a protocol that just drops tip racks in the chute. Saw at least one error that would have caused a stall/failure to drop without the slide, but which worked fine with this PR.

# Changelog

- Added a shared-data constant for the horizontal width of the gripper paddles (I just measured with a set of calipers)
- Removed the post-drop delay from labware movements in the trash chute
- Added a post-drop slide offset parameter to allow the gripper to slide away after dropping a piece of labware in the trash chute
- Updated tests

# Review requests

# Risk assessment

Pretty low. The biggest concern that we (me + @CaseyBatten + hardware) had is about the clearance to the right of the trash chute, but we collectively agreed that there _shouldn't_ be anything installed over that slot that would prevent the gripper (when fully homed) from moving over it.


CaseyBatten (approved):
These changes look good, and in testing on RSS Flex I've had excellent results with the "slide" maneuver so far. I think for 99% of 'stuck labware' cases this will probably suffice to free us.

sfoster1 (approved):
Code looks good to me, but let's add a block comment about why we're doing this to the command implementation please

PR #14066, opened 2023-11-30
feat(api): add deck conflict checks for pipetting with partial tip configuration

sanni-t:
Closes RSS-378

# Overview

Adds conflicts that could arise by moving a partial configuration pipette to the specified location.
This PR adds full conflict check for 96-channel when configured with a rightmost column nozzle layout which includes:
 - [x] conflict check between destination location and labware in the west slot based on final pipette position and west labware height
 - [x] conflict check based on the physical extents of the 96 channel within the Flex deck space 

This PR also adds partial conflict check for an 8-channel when configured with single nozzle layout (`H1` nozzle) which includes:
- [x] conflict check between destination location and labware in the north slot based on final pipette position and north labware height. 
- ❌ conflict checks based on physical extents of 8-channel pipettes within OT2 or Flex deck space 

If he pipette is configured with any other configuration, there will not be any deck conflict checking and all checks will simply no-op so use other configurations with caution.

NOTE: Conflict checking with addressable areas of trash bin and waste chute will be in a follow-up ticket.

# Test Plan

- Here's a sample protocol to test the deck conflicts of the 96-channel. The protocol should fail analysis if there's deck conflicts. If there aren't, then you should be able to run the full protocol on a robot.

```


requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(protocol_context):

	#############  TALL LABWARE  #################
	# These labware should be to the east of short labware to avoid partial tip deck conflicts
	trash_labware = protocol_context.load_labware("opentrons_1_trash_3200ml_fixed", "A3")

	# Will result in errors in this placement-  <---------------------
	tip_rack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C1")
	tip_rack2 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "B1")
	
	# Will NOT raise errors in this placement-
	# tip_rack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C3")
	# tip_rack2 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "B3")


	instrument = protocol_context.load_instrument('flex_96channel_1000', mount="left", tip_racks=[tip_rack1, tip_rack2])
	
	#############  SHORT LABWARE  ################
	# These labware should be to the west of tall labware to avoid any partial tip deck conflicts
	my_pcr_plate = protocol_context.load_labware('nest_96_wellplate_200ul_flat', "C2")
	my_other_pcr_plate = protocol_context.load_labware('nest_96_wellplate_200ul_flat', "D2")


	############# MEDIUM LABWARE ###############
	# These should be in the middle, but might be fine to the west of some labware 
	# if the pipette doesn't lower enough to collide with it
	my_tuberack = protocol_context.load_labware('opentrons_10_tuberack_falcon_4x50ml_6x15ml_conical', "D1")

	instrument.configure_nozzle_layout(style=COLUMN, start="A12")

	# No error cuz short labware to left of tiprack
	instrument.pick_up_tip(tip_rack1.wells()[0])

	# No error cuz no labware on left of plate

	# No error cuz dispensing from high above plate, so it clears reservoir on the left
	instrument.dispense(20, my_other_pcr_plate.wells_by_name()["A1"].top(150))

	# Will raise error   <---------------------
	instrument.dispense(20, my_other_pcr_plate.wells_by_name()["A1"].bottom())

	# No error cuz A1 is within 96 channel right column bounds
	instrument.dispense(10, my_tuberack.wells_by_name()["A1"])

	instrument.drop_tip()

	######### CHANGE CONFIG TO ALL #########
	instrument.configure_nozzle_layout(style=ALL)

	# No error cuz short labware to left of tiprack
	instrument.pick_up_tip(tip_rack2.wells()[0])

	# No error cuz no labware on left of plate

	# No error NOW because of full config
	instrument.dispense(20, my_other_pcr_plate.wells_by_name()["A1"].bottom())
```

- Also test a regular 8- and single channel protocol to make sure its behavior is unchanged.

# Changelog

- added deck conflict check logic to all pipette movement methods, based on pipette configuration
- added deck item geometry getters to calculate heights of items in slots

# Review requests

- Test and make sure the checks work correctly

# Risk assessment

Low for existing API

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
+++ b/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
@@ -1,10 +1,16 @@
 """A Protocol-Engine-friendly wrapper for opentrons.motion_planning.deck_conflict."""
 
 import itertools
-from typing import Collection, Dict, Optional, Tuple, overload
+import logging
+from typing import Collection, Dict, Optional, Tuple, overload, Union
 
+from opentrons.hardware_control.nozzle_manager import NozzleConfigurationType
 from opentrons.hardware_control.modules.types import ModuleType
 from opentrons.motion_planning import deck_conflict as wrapped_deck_conflict
+from opentrons.motion_planning.adjacent_slots_getters import (
+    get_north_slot,
+    get_west_slot,
+)
 from opentrons.protocol_engine import (
     StateView,
     DeckSlotLocation,
@@ -12,9 +18,25 @@ from opentrons.protocol_engine import (
     OnLabwareLocation,
     AddressableAreaLocation,
     OFF_DECK_LOCATION,
+    WellLocation,
+    DropTipWellLocation,
 )
 from opentrons.protocol_engine.errors.exceptions import LabwareNotLoadedOnModuleError
-from opentrons.types import DeckSlotName
+from opentrons.types import DeckSlotName, Point
+
+
+class PartialTipMovementNotAllowedError(ValueError):
+    """Error raised when trying to perform a partial tip movement to an illegal location."""
+
+
+_log = logging.getLogger(__name__)
+
+# Arbitrary safety margin in z-direction
+Z_SAFETY_MARGIN = 10
+
+# Bounding box measurements
+A12_column_front_left_bound = Point(x=-1.8, y=2)
+A12_column_back_right_bound = Point(x=592, y=506.2)
 
 
 @overload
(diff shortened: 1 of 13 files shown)


api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -1,20 +1,42 @@
...
     OFF_DECK_LOCATION,
+    WellLocation,
+    DropTipWellLocation,
 )
 from opentrons.protocol_engine.errors.exceptions import LabwareNotLoadedOnModuleError
-from opentrons.types import DeckSlotName
+from opentrons.types import DeckSlotName, Point
+
+
+class PartialTipMovementNotAllowedError(ValueError):
```
sfoster1: We should make this an enumerated error

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -1,20 +1,42 @@
...
+    """Error raised when trying to perform a partial tip movement to an illegal location."""
+
+
+_log = logging.getLogger(__name__)
+
+# Arbitrary safety margin in z-direction
+Z_SAFETY_MARGIN = 10
+
+# Bounding box measurements
+A12_column_front_left_bound = Point(x=-1.8, y=2)
```
sfoster1, line 44: Let's add a todo to push this somewhere else because I would like to derive this from geometry definitions

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -106,6 +128,203 @@ def check(
...
+
+    well_location_point = engine_state.geometry.get_well_position(
+        labware_id=labware_id, well_name=well_name, well_location=well_location
+    )
+
+    if not _is_within_pipette_extents(
+        engine_state=engine_state, pipette_id=pipette_id, location=well_location_point
+    ):
+        # WARNING: (spp, 2023-11-30: this needs to be wired up to check for
+        # 8-channel pipette extents on both OT2 & Flex!!)
```
sfoster1, line 275: is it wired up for flex only right now? what's the failure case?

shared-data/module/definitions/3/thermocyclerModuleV2.json
```diff
@@ -8,7 +8,7 @@
     "z": 108.96
   },
   "dimensions": {
-    "bareOverallHeight": 98.0,
+    "bareOverallHeight": 108.96,
```
sfoster1, line 11: where'd this come from?

sfoster1 (approved):
Looks good to me. I think the warnings about ot2 vs flex are because we don't do extent checking for ot2 yet, is that right?

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: yep yep

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -376,7 +407,13 @@ def drop_tip(
...
         else:
             well_location = DropTipWellLocation()
-
+        deck_conflict.check_safe_for_pipette_movement(
+            engine_state=self._engine_client.state,
+            pipette_id=self._pipette_id,
+            labware_id=labware_id,
+            well_name=well_name,
+            well_location=WellLocation(),
+        )
```
sanni-t, line 416: ~I think I missed adding the check to `move_to`. Do we want to allow `move_to` to bypass this check?~
Nevermind. `move_to` is for moving to non-well locations, which makes it checking for collisions with surrounding deck items non-trivial.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: It's wired up to check the extents of only the 96-channel in rightmost column configuration. I don't have the 8-channel movement bounds for both OT2 & Flex. 
So failure case is if you configure an 8-channel to use the frontmost nozzle and you made it move to well A1 of a labware in slot 11/ A1, then the pipette might crash against the back of the robot instead of raising an error that it's an unreachable location for the 8-channel in this configuration.

sanni-t:
> Looks good to me. I think the warnings about ot2 vs flex are because we don't do extent checking for ot2 yet, is that right?

Yes, but also we don't do all kinds of extents checking for Flex either. It is usually not an issue because the pipettes are designed to reach all deck locations when used normally, but because a partially configured pipette might not be able to reach locations that it can reach under normal config, it becomes more important to do the checks.

shared-data/module/definitions/3/thermocyclerModuleV2.json
sanni-t: So [this PR](https://github.com/Opentrons/opentrons/pull/13567) updated the `labwareOffset` z to be the topmost point of the labware-holding metal cylinders. Previously this z was 98.26, and I believe the bare overall height was the bottom plane of those cylinders. When I updated the z to be the topmost point, I forgot to update the overall height. Also it makes more sense to have the topmost point be the height, especially now that some pipette movements will be allowed or not depending on the heights of deck items.

sanni-t:
Successfully tested on a Flex to make sure a protocol that passes deck conflict checks is able to run correctly without issues.

PR #14067, opened 2023-11-30
fix(engine): deck configuration pipette movement related errors

jbleon95:
# Overview

This PR addresses RSS-410 and fixes a few bugs in Protocol Engine related to the recent deck configuration work.

If labware was on a staging slot, this was raising an error when trying to find the tallest z height of the labware, it will now find the height of labware on staging slots.

Related to that, the engine will now raise a readable error if you try to move to a labware on a staging slot, an off-deck labware, or the staging slot itself.

Another bug was found where anything loaded on slot A3/12 was treated as a "fixed trash". This check has been improved to instead check the labware defintition for the `"fixedTrash"` quirk.

# Test Plan

[Tip Drop Tip Rack Drop Well Plate Drop Waste chute no cover.json](https://github.com/Opentrons/opentrons/files/13517061/Tip.Drop.Tip.Rack.Drop.Well.Plate.Drop.Waste.chute.no.cover.json)

This handwritten JSON protocol was causing the errors listed above, and the protocol now analyzes and runs properly with these fixes in place.

# Changelog

- Take into account labware on staging slots when calling `get_all_labware_highest_z`
- Raise a friendly error when trying to move to labware on a staging slot or off-deck, as well as when moving to a staging slot addressable area directly
- Check if a labware is fixed trash by checking the relevant `"fixedTrash"` quirk, not its location.

# Review requests

# Risk assessment

Low.


sfoster1 (changes requested):
I'm a little dubious about outright not counting labware in staging slots in highest-z checks.

The reason we call staging slots pipette-inaccessible is the left pipette can't access _all_ the wells. It's still possible to get the left pipette over some of them, and the right pipette over most of them, and so we probably want to take them into account when motion planning. 

For instance, if you drop a labware in D4 and then move to well H12 of a plate in slot A3 to pipette, we'd go over D4, C4, and B4 right?

sfoster1 (approved):
Makes sense to me, thanks for the change!

PR #14068, opened 2023-11-30
fix(app): only show 96 channel weight warning for 96 detach

smb2268:
fix RQA-1960, RQA-1962

# Overview
Fix ternary to show warnings on the Before you Begin screen _only_ for 96 channel pipettes. For the detach flow we should show the heavy pipette warning and for the attach and calibrate flows we should show the waste chute warning.

# Test Plan
Launch a single or 8 channel attach and/or calibrate flow and see that the 96-channel warning is no longer present

# Changelog
Fix parentheses location in ternary

# Review requests
Sanity check

# Risk assessment
Low


mjhuff (approved):
)

PR #14070, opened 2023-11-30
feat(app): Add addressable area command display names

mjhuff:
Closes RAUT-878

# Overview

This PR adds command text support for addressable areas (via a utility function). When moving to labware within an addressable area, the slot name is properly displayed. Additionally, appropriate, translated text is shown when moving to a waste chute or fixed trash. 

### Current Behavior (Step 8)

Screenshot 2023-11-30 at 9 20 21 AM

### Fixed Behavior (Step 8)

Screenshot 2023-11-30 at 1 58 06 PM

# Test Plan

- Run a protocol that makes use of a trash bin, a waste chute, and contains labware on an addressable area.
- Confirm on the desktop app & ODD that appropriate text is now displayed. 

# Changelog
- Added appropriate command display text during protocol runs for addressable areas. 

# Risk assessment
low

### DIFF
--- a/app/src/assets/localization/en/protocol_command_text.json
+++ b/app/src/assets/localization/en/protocol_command_text.json
@@ -36,7 +36,7 @@
   "move_to_coordinates": "Moving to (X: {{x}}, Y: {{y}}, Z: {{z}})",
   "move_to_slot": "Moving to Slot {{slot_name}}",
   "move_to_well": "Moving to well {{well_name}} of {{labware}} in {{labware_location}}",
-  "move_to_addressable_area": "Moving to {{addressable_area}} at {{speed}} mm/s at {{height}} mm high",
+  "move_to_addressable_area": "Moving to {{addressable_area}}",
   "notes": "notes",
   "off_deck": "off deck",
   "offdeck": "offdeck",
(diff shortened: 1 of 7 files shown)


app/src/organisms/CommandText/index.tsx
```diff
@@ -235,13 +236,16 @@ export function CommandText(props: Props): JSX.Element | null {
       )
     }
     case 'moveToAddressableArea': {
-      const { addressableAreaName, speed, minimumZHeight } = command.params
```
mjhuff, line 238: Talked w/ Brian - we don't really need to provide speed and height offsets to the user as command text. Additionally, these values aren't always present, so special casing them whenever available is superfluous work.

app/src/assets/localization/en/protocol_command_text.json
mjhuff, line 1: Added the "Waste Chute" and "Trash Bin" as proper nouns here. This is what's on our sales page and documentation...I think this makes sense but open to discuss!

app/src/organisms/CommandText/utils/getAddressableAreaDisplayName.ts
```diff
@@ -0,0 +1,34 @@
...
+  ) {
+    return ''
+  }
+
+  const addressableAreaName: MoveToAddressableAreaParams['addressableAreaName'] =
+    addressableAreaCommand.params.addressableAreaName
+  const movableTrashSubstr = 'movableTrash'
+
+  if (addressableAreaName.includes(movableTrashSubstr)) {
+    const slotName = addressableAreaName.split(movableTrashSubstr)[1]
```
jerader: hmm is there a more elegant way to accomplish this without using `split`? Maybe we can make a const that maps the addressable area name with the slot name? Do you have thoughts @b-cooper?

jerader (approved):
left a suggestion for removing the usage of `split()` but otherwise lgtm!

app/src/organisms/CommandText/utils/getAddressableAreaDisplayName.ts
mjhuff: Thanks for bring this to my attention! Talked w/ Brian - I wasn't aware of the pattern we had concerning handling consts such as these. I've updated it!

PR #14072, opened 2023-11-30
fix(app): add final gripper animations to all flows

smb2268:
fix RQA-1931

# Overview
Added animations from google sheet for gripper flows

# Test Plan
Go through gripper detach, attach and calibrate flows and ensure animations are correct

# Changelog
Replaced every animation file, a few were already correct so only 4 ended up having changes

# Review requests
Review video, ensure animations match those in ticket - I ran into some errors during calibration so wasn't able to capture the last few pin animations

https://github.com/Opentrons/opentrons/assets/14302493/9355cf6e-c5c0-4247-9241-4c49f6afbc18

# Risk assessment
Low


mjhuff (approved):
LGTM - verified pin animations as well.

PR #14073, opened 2023-11-30
docs(api): Versioning page updates for Python API 2.16

ecormany:
# Overview

Updates the Versioning page for new and improved features in Python API version 2.16.

# Test Plan

Review [sandbox](http://sandbox.docs.opentrons.com/docs-versioning-2.16/v2/versioning.html)

# Changelog

- updated version table
- list of new features in 2.16, with cross-references
- `|apiLevel|` substitution -> `2.16`

# Review requests

Any important features missing? Anything included that was actually in 2.15?

Note that the cross-reference to "Partial Tip Pickup" depends on #14189 and will only display as plain text here.

# Risk assessment

nil, docs only.

### DIFF
--- a/api/docs/v2/versioning.rst
+++ b/api/docs/v2/versioning.rst
@@ -127,12 +127,12 @@ Changes in API Versions
 Version 2.16
 ------------
 
-This version introduces new features for Flex and adds and improves methods for aspirating and dispensing.
+This version introduces new features for Flex and adds and improves methods for aspirating and dispensing. Note that when updating Flex protocols to version 2.16, you *must* load a trash container before dropping tips.
 
 - New features
 
   - Use :py:meth:`.configure_nozzle_layout` to pick up a single column of tips with the 96-channel pipette. See :ref:`partial-tip-pickup <Partial Tip Pickup>`.
-  - Specify the deck configuration of your Flex with :py:meth:`.load_waste_chute` and :py:meth:`.load_trash_bin`.
+  - Specify the trash containers attached to your Flex with :py:meth:`.load_waste_chute` and :py:meth:`.load_trash_bin`.
   - Dispense, blow out, drop tips, and dispose labware in the waste chute. Disposing labware requires the gripper and calling :py:meth:`.move_labware` with ``use_gripper=True``.
   - Perform actions in staging area slots by referencing slots A4 through D4. See :ref:`deck-slots`.
   - Explicitly command a pipette to :py:meth:`.prepare_to_aspirate`. The API usually prepares pipettes to aspirate automatically, but this is useful for certain applications, like pre-wetting routines.
--- a/api/docs/v2/versioning.rst
+++ b/api/docs/v2/versioning.rst
@@ -140,7 +140,6 @@ This version introduces new features for Flex and adds and improves methods for 
 - Improved features
 
   - :py:meth:`.aspirate`, :py:meth:`.dispense`, and :py:meth:`.mix` will not move any liquid when called with ``volume=0``.
-  - ``opentrons_simulate`` now works for all API versions.
 
 - Other changes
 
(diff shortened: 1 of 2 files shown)


api/docs/v2/versioning.rst
```diff
@@ -122,6 +124,28 @@ This table lists the correspondence between Protocol API versions and robot soft
...
+  - Use :py:meth:`.configure_nozzle_layout` to pick up a single column of tips with the 96-channel pipette. See :ref:`Partial Tip Pickup <partial-tip-pickup>`.
+  - Specify the trash containers attached to your Flex with :py:meth:`.load_waste_chute` and :py:meth:`.load_trash_bin`.
+  - Dispense, blow out, drop tips, and dispose labware in the waste chute. Disposing labware requires the gripper and calling :py:meth:`.move_labware` with ``use_gripper=True``.
+  - Perform actions in staging area slots by referencing slots A4 through D4. See :ref:`deck-slots`.
+  - Explicitly command a pipette to :py:meth:`.prepare_to_aspirate`. The API usually prepares pipettes to aspirate automatically, but this is useful for certain applications, like pre-wetting routines.
+
+- Improved features
+
+  - :py:meth:`.aspirate`, :py:meth:`.dispense`, and :py:meth:`.mix` will not move any liquid when called with ``volume=0``.
+  - ``opentrons_simulate`` now works for all API versions.
```
SyntaxColoring: The `opentrons_simulate` fix is technically an `opentrons` package version thing, not PAPI version thing.

SyntaxColoring (approved):

jwwojak (commented):
Nice! LGTM. :shipit:

api/docs/v2/versioning.rst
```diff
@@ -145,6 +145,10 @@ This version introduces new features for Flex and adds and improves methods for
 - Other changes
 
   - :py:obj:`.ProtocolContext.fixed_trash` and :py:obj:`.InstrumentContext.trash_container` now return :py:class:`.TrashBin` objects instead of :py:class:`.Labware` objects.
+  
+- Known issues
+
+  - It's possible to load a Thermocycler and then load another item in slot A1. It's also possible to load any labware or adapter on in the same slot as a trash bin. Don't do either of these things, as it could lead to unexpected pipetting behavior and crashes.
```
SyntaxColoring: > It's also possible to load any labware or adapter on in the same slot as a trash bin

I don't think it's possible to do this, is it? Here's my test case:

```python


def run(protocol):
    trash = protocol.load_trash_bin("A1")
    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_200ul", "A1")

    )

    pipette.pick_up_tip()

    # ProtocolCommandFailedError:
    # Error 4000 GENERAL_ERROR (ProtocolCommandFailedError):
    # IncompatibleAddressableAreaError: Cannot use Trash Bin in A1,
    # not compatible with one or more of the following fixtures: Slot A1
    pipette.drop_tip()
```

There *is* an implementation quirk where the error will only happen when you *use* the trash bin. If you *only* `load_labware()` and `load_trash_bin()`, you won't get an error. But that's not dangerous; it's just a no-op.

```suggestion
  - It's possible to load a Thermocycler and then load another item in slot A1. Don't do this, as it could lead to unexpected pipetting behavior and crashes.
```

-----

But if we end up keeping this, there's an "on in" typo.

```suggestion
  - It's possible to load a Thermocycler and then load another item in slot A1. It's also possible to load any labware or adapter in the same slot as a trash bin. Don't do either of these things, as it could lead to unexpected pipetting behavior and crashes.
```

ecormany: You're right, the error only happens when `drop_tip()` is called, but it does raise the error. I think I was expecting a `DeckConflictError`, which would happen regardless of pipetting commands.

PR #14076, opened 2023-11-30
fix(app): Pipette and Gripper flow design fixes

smb2268:
fix RQA-1963, RQA-1946
# Overview
This resolves a few styling related bugs:
1. Use "Pipette" if pipette name is missing on still attached error screen so copy will read "Pipette still attached" instead of "still attached"
2. Add margin to subheader on simple wizard body screen so that it doesn't overlap with the buttons if the copy is more than one line - see screenshot
3. Capitalize exit CTA on gripper success screen
Screen Shot 2023-11-30 at 5 36 38 PM

# Test Plan

# Changelog

# Review requests
Sanity check code

# Risk assessment
Low


mjhuff (approved):
Thanks for fixing! LGTM after test fixes ✅

PR #14077, opened 2023-11-30
ci(shared-data, components): add github action for publishing packages to npm

b-cooper:
# Overview

In order to support a single source of truth for applications outside of the monorepo (read:
"Protocol Library") that rely on fundamental components and geometries within the mono repo, add tag
based deploy actions that publish bundled artifacts to NPM.

# Changelog

# Review requests

# Risk assessment


.github/workflows/components-test-build-deploy.yaml
```diff
@@ -137,3 +158,41 @@ jobs:
...
         run: |
           aws s3 sync ./dist s3://opentrons-components/${{ env.OT_BRANCH}} --acl public-read
+
+  publish-components:
+    name: 'publish components package to npm'
+    runs-on: 'ubuntu-latest'
+    needs: ['js-unit-test', 'determine-build-type']
+    if: needs.determine-build-type.outputs.type == 'publish'
+    env: 
+      VERSION_STRING: echo github.ref | sed 's/refs\/tags\/components@//'
```
sfoster1: I don't think you can drop shell commands here, I think it's [expressions](https://docs.github.com/en/actions/learn-github-actions/expressions#functions) only. i'd recommend adding the tag value as an output from the `determine-build-type` job, where you can use shell

.github/workflows/components-test-build-deploy.yaml
```diff
@@ -137,3 +158,41 @@ jobs:
...
+            ${{ github.workspace }}/.yarn-cache
+            ${{ github.workspace }}/.npm-cache
+          key: js-${{ secrets.GH_CACHE_VERSION }}-${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            js-${{ secrets.GH_CACHE_VERSION }}-${{ runner.os }}-yarn-
+      - name: 'set version number'
+        run: |
+          npm install -g json
+          json -I -f ./components/package.json -e "this.version=\"$VERSION_STRING\""
+      - run: cd ./components && npm publish --provenace --access public
```
sfoster1:
```suggestion
      - run: cd ./components && npm publish --provenance --access public
```

Makefile
```diff
@@ -61,7 +61,6 @@ setup-js:
 	yarn
 	$(MAKE) -C $(APP_SHELL_DIR) setup
 	$(MAKE) -C $(APP_SHELL_ODD_DIR) setup
-	$(MAKE) -C $(SHARED_DATA_DIR) setup-js
```
b-cooper, line 64: this call is no longer necessary.

shared-data/Makefile
```diff
@@ -28,17 +25,12 @@ clean: clean-js clean-py
...
 # JavaScript targets
 
-.PHONY: setup-js
-setup-js: dist-js
+.PHONY: lib-js
+lib-js: export NODE_ENV := production
+lib-js:
+	yarn webpack
 
-.PHONY: dist-js
```
b-cooper, line 34: these were purely used for bundling v1 labware defs for old versions of monorepo SPA's

protocol-library-kludge/src/URLDeck.tsx
```diff
@@ -115,14 +106,7 @@ export class URLDeck extends React.Component<{}> {
                   >
                     {labwareDefV2 ? (
                       <LabwareRender definition={labwareDefV2} />
-                    ) : (
```
b-cooper, line 118: this whole directory will never be deployed again. It can be removed when legacy PL is unlisted

sfoster1: We eagerly await the day

sfoster1 (approved):
Looks good to me!

.github/workflows/components-test-build-deploy.yaml
```diff
@@ -102,11 +102,32 @@ jobs:
...
+              echo 'type=storybook' >> $GITHUB_OUTPUT
+            elif [ "${{ format('{0}', startsWith(github.ref, 'refs/tags/components')) }}" = "true" ] ; then
+              echo "publish builds for components tags"
+              echo 'type=publish' >> $GITHUB_OUTPUT
+            else
+              echo "No build for ref ${{github.ref}} and event ${{github.event_type}}"
+              echo 'type=none' >> $GITHUB_OUTPUT
+            fi
+
   deploy-components:
```
shlokamin, line 126:
```suggestion
  deploy-components-storybook:
```

shlokamin (approved):
🥳

PR #14079, opened 2023-12-01
fix(app): do not redundantly check labware in LPC

shlokamin:
# Overview

This PR dedupes labware def uri <> location combos in tip based LPC so we do not redundantly check labware def uri <> location combos.

closes RAUT-889

# Test Plan
Tested on an OT-2 (tip based LPC) just to make sure nothing broke
Tested on a Flex (probe based LPC) just to make sure nothing broke

# Changelog

- Do not redundantly check labware in LPC

# Risk assessment

Low./med

### DIFF
--- a/app/src/organisms/LabwarePositionCheck/utils/getTipBasedLPCSteps.ts
+++ b/app/src/organisms/LabwarePositionCheck/utils/getTipBasedLPCSteps.ts
@@ -148,6 +148,11 @@ function getCheckTipRackSectionSteps(args: LPCArgs): CheckTipRacksStep[] {
 function getCheckLabwareSectionSteps(args: LPCArgs): CheckLabwareStep[] {
   const { labware, modules, commands, primaryPipetteId } = args
   const labwareDefinitions = getLabwareDefinitionsFromCommands(commands)
+  const labwareLocationCombos = getLabwareLocationCombos(
+    commands,
+    labware,
+    modules
+  )
 
   return labware.reduce<CheckLabwareStep[]>((acc, currentLabware) => {
     const labwareDef = labwareDefinitions.find(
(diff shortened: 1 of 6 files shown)


app/src/organisms/LabwarePositionCheck/utils/getTipBasedLPCSteps.ts
```diff
@@ -163,14 +163,26 @@ function getCheckLabwareSectionSteps(args: LPCArgs): CheckLabwareStep[] {
     const adapter = (labwareDef?.allowedRoles ?? []).includes('adapter')
     if (isTiprack || adapter) return acc // skip any labware that is a tiprack or adapter
 
-    const labwareLocationCombos = getLabwareLocationCombos(
+    const deDuplicatedLabwareLocationCombos = getLabwareLocationCombos(
       commands,
       labware,
       modules
+    ).filter(
+      (outerLabwareLocationCombo, _index, nonDupedLabwareLocationCombos) => {
```
sfoster1:
```suggestion
      (outerLabwareLocationCombo, _index, nonDeDupedLabwareLocationCombos) => {
```
?

app/src/organisms/LabwarePositionCheck/utils/getTipBasedLPCSteps.ts
```diff
@@ -163,14 +163,26 @@ function getCheckLabwareSectionSteps(args: LPCArgs): CheckLabwareStep[] {
...
+        return nonDupedLabwareLocationCombos.some(
```
sfoster1: talked in person but i don't think is right for the following reasons:
- this returns `true` if the labware+position combo _is_ duplicated when I think it should be the reverse
- this will always detect the labware+position combo as duplicated because it's checking the whole of the list we're iterating through, so the element we're currently checking will always be in that list

sfoster1 (changes requested):

sfoster1 (approved):
Looks good to me, way simpler than I was thinking

b-cooper (approved):
🪑

PR #14081, opened 2023-12-01
refactor(step-generation): remove fixedTrash from labwareEntities in test fixtures

jerader:
closes RAUT-884

# Overview

This is the final portion of fixed trash being a part of `labwareEntities`. This PR does 2 things primarily:

1. fix when the `dropTipLocationDoesNotExist()` error creator occurs - since PD only supports dropping tip into the waste chute and trash bin right now, I updated the logic to search for the `dropTipLocation` only in the `additionalEquipmentEntities`. 
2. Due to item 1, many tests broke so this PR also updates all of step-generation tests to not have `fixedTrash` in the labware entities fixture (both in invariant context and labware location), and instead have the `fixedTrash` in the additional equipment entities fixture.

# Test Plan

Create a Flex protocol. Add a transfer or mix step that aspriates and dispenses into a labware but then drops tip in to the waste chute or trash bin. Go back to the file tab and delete the drop tip entity. See that the step now has an error and when you click to expand it, see that the error is the correct timeline error.

Review my changes to the tests! A lot had to be updated because almost every step-generation test had fixedTrash listed in its labware entities 😅 

# Changelog

- update when `dropTipLocationDoesNotExist()` error creator occurs in `mix`, `distribute`, `transfer`, and `consolidate`
- delete `dropAllTips` and test since it has not been in use
- extend `dropTip` logic in `distribute` compound command
- delete `fixedTrash` from `labwareLocationsUpdate` and `labwareEntities` in the fixtures
- update all affected tests

# Review requests

see test plan

# Risk assessment

low


step-generation/src/commandCreators/compound/distribute.ts
```diff
@@ -283,16 +301,33 @@ export const distribute: CommandCreator<DistributeArgs> = (
...
+        })
+      }
+      if (isTrashBin) {
+        dropTipCommand = movableTrashCommandsUtil({
+          type: 'dropTip',
+          pipetteId: args.pipette,
+          prevRobotState,
+          invariantContext,
+        })
+      }
```
jerader, line 322: I forgot to extend this logic to this component! This is basically what is being done in Transfer and Consolidate already.

step-generation/src/commandCreators/atomic/dropAllTips.ts
```diff
@@ -1,51 +0,0 @@
-import { FLEX_TRASH_DEF_URI, OT_2_TRASH_DEF_URI } from '../../constants'
```
jerader, line 1: this atomic command component isn't in use and i don't forsee it being in use so i figured i'd just delete it

step-generation/src/__tests__/consolidate.test.ts
```diff
@@ -102,7 +99,6 @@ beforeEach(() => {
   }
   invariantContext = {
```
mjhuff: Do we need L100-102 now?

step-generation/src/__tests__/distribute.test.ts
mjhuff, line 1: Nice, I like the new helper functions!

step-generation/src/commandCreators/compound/distribute.ts
```diff
@@ -108,6 +110,22 @@ export const distribute: CommandCreator<DistributeArgs> = (
   )
   const { pipette } = args
 
+  const isWasteChute =
+    invariantContext.additionalEquipmentEntities[args.dropTipLocation] !=
+      null &&
```
mjhuff:
```suggestion
    invariantContext.additionalEquipmentEntities[args.dropTipLocation] !=
      null &&
```
I could be missing something here, but do we need these nullish checks for isWasteChute and isTrashBin?

mjhuff (approved):
Just a couple very minor comments, but this makes enough sense to me!

step-generation/src/__tests__/consolidate.test.ts
jerader: oh ya good catch haha thanks!

step-generation/src/commandCreators/compound/distribute.ts
jerader: good call, i think this is redundant. Thanks!

PR #14082, opened 2023-12-02
feat(shared-data,protocol-engine): Use the tallest fixture, not the tallest addressable area, for motion planning

SyntaxColoring:
# Overview

Closes RSS-409.

# Test Plan

I've tested this through this procedure:

1. Hand-edit the deck definition to have a higher `height` for one of the waste chute fixtures.
2. Use the Opentrons App to configure the deck with that waste chute fixture.
3. Run a protocol and note the height that the pipette climbs to when it moves from one labware to another.
4. Use the Opentrons App to configure the deck with a different waste chute fixture, one that doesn't have the hand-edited `height`.
5. Run the same protocol and note the height that the pipette climbs to when it moves from one labware to another. It should be different.

We can test this further when we do RSS-408.

# Changelog

* Give every `cutoutFixture` a `height`. This new field represents the overall height of the physical thing mounted to the deck. This is different from the existing heights of the `addressableArea`s—those are logical "interaction points" that aren't necessarily tied to the physical geometry.
* During a protocol run, make sure to move the pipette above the highest `cutoutFixture` `height`. Formerly, we were moving the pipette above the highest `addressableArea` height, which was dangerous because the physical thing could be taller than that.
* Various small refactors.

# Review requests

See my inline notes.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/addressable_areas.py
+++ b/api/src/opentrons/protocol_engine/state/addressable_areas.py
@@ -1,6 +1,6 @@
 """Basic addressable area data state and store."""
 from dataclasses import dataclass
-from typing import Dict, Set, Union, List
+from typing import Dict, List, Optional, Set, Union
 
 from opentrons_shared_data.deck.dev_types import DeckDefinitionV4, SlotDefV3
 
@@ -36,9 +36,30 @@ class AddressableAreaState:
     """State of all loaded addressable area resources."""
 
     loaded_addressable_areas_by_name: Dict[str, AddressableArea]
+    """The addressable areas that have been loaded so far.
+
+    When `use_simulated_deck_config` is `False`, these are the addressable areas that the
+    deck configuration provided.
+
+    When `use_simulated_deck_config` is `True`, these are the addressable areas that have been
+    referenced by the protocol so far.
+    """
+
     potential_cutout_fixtures_by_cutout_id: Dict[str, Set[PotentialCutoutFixture]]
+
     deck_definition: DeckDefinitionV4
+
+    deck_configuration: Optional[DeckConfigurationType]
+    """The host robot's full deck configuration.
+
+    If `use_simulated_deck_config` is `True`, this is meaningless and this value is undefined.
+    In practice it will probably be `None` or `[]`.
+
+    If `use_simulated_deck_config` is `False`, this will be non-`None`.
+    """
+
     use_simulated_deck_config: bool
+    """See `Config.use_simulated_deck_config`."""
 
 
 def _get_conflicting_addressable_areas(
@@ -115,6 +136,7 @@ class AddressableAreaStore(HasState[AddressableAreaState], HandlesActions):
                 )
             )
         self._state = AddressableAreaState(
+            deck_configuration=deck_configuration,
             loaded_addressable_areas_by_name=loaded_addressable_areas_by_name,
             potential_cutout_fixtures_by_cutout_id={},
             deck_definition=deck_definition,
@@ -127,7 +149,11 @@ class AddressableAreaStore(HasState[AddressableAreaState], HandlesActions):
             self._handle_command(action.command)
         if isinstance(action, PlayAction):
             current_state = self._state
-            if action.deck_configuration is not None:
+            if (
+                action.deck_configuration is not None
+                and not self._state.use_simulated_deck_config
+            ):
+                self._state.deck_configuration = action.deck_configuration
                 self._state.loaded_addressable_areas_by_name = (
                     self._get_addressable_areas_from_deck_configuration(
                         deck_config=action.deck_configuration,
--- a/api/src/opentrons/protocol_engine/state/addressable_areas.py
+++ b/api/src/opentrons/protocol_engine/state/addressable_areas.py
@@ -310,6 +310,21 @@ class AddressableAreaView(HasState[AddressableAreaState]):
         """Get a list of all loaded addressable area names."""
         return list(self._state.loaded_addressable_areas_by_name)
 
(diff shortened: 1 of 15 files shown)


shared-data/deck/definitions/4/ot3_standard.json
```diff
@@ -479,7 +479,8 @@
         "cutoutC1": ["C1"],
         "cutoutB1": ["B1"],
         "cutoutA1": ["A1"]
-      }
+      },
+      "height": 0
```
SyntaxColoring, line 483: The standard slots have L-shaped corners that jut up by a couple cm, and it would be appropriate to reflect their height here. I couldn't find their exact height anywhere so I'm leaving them as 0 for now, but if anyone can point to it, we should add it in this PR.

shared-data/deck/definitions/4/ot3_standard.json
```diff
@@ -536,15 +540,17 @@
...
     },
     {
       "id": "wasteChuteRightAdapterCovered",
       "mayMountTo": ["cutoutD3"],
       "displayName": "Waste Chute Adapter for 1 or 8 Channel Pipettes",
       "providesAddressableAreas": {
         "cutoutD3": ["1and8ChannelWasteChute"]
-      }
+      },
+      "height": 124.5
```
SyntaxColoring, line 553: 124.5 mm as the waste chute heights comes from the "Max envelope of chute" value in the robot extents spreadsheet.

shared-data/deck/definitions/4/ot3_standard.json
```diff
@@ -536,15 +540,17 @@
         "cutoutC3": ["movableTrashC3"],
         "cutoutB3": ["movableTrashB3"],
         "cutoutA3": ["movableTrashA3"]
-      }
+      },
+      "height": 40
```
SyntaxColoring, line 544: This 40 mm comes from the "Outer dims of trash bin" value in the robot extents spreadsheet, and it matches the old `opentrons_1_trash_3200ml_fixed` definition.

shared-data/deck/schemas/4.json
```diff
@@ -292,6 +293,10 @@
                 "type": "string"
               }
             }
+          },
+          "height": {
```
SyntaxColoring, line 297: Do we like the name `height` for this, or do we want to use `zDimension` like we do in labware definitions?

shared-data/deck/definitions/4/ot2_short_trash.json
```diff
@@ -389,15 +389,17 @@
...
     },
     {
       "id": "fixedTrashSlot",
       "mayMountTo": ["cutout12"],
       "displayName": "Fixed Trash",
       "providesAddressableAreas": {
         "cutout12": ["shortFixedTrash"]
-      }
+      },
+      "height": 58
```
SyntaxColoring, line 402: I'm getting this value of 58 from the `opentrons_1_trash_850ml_fixed` definition.

shared-data/deck/definitions/4/ot2_standard.json
```diff
@@ -389,15 +389,17 @@
...
     },
     {
       "id": "fixedTrashSlot",
       "mayMountTo": ["cutout12"],
       "displayName": "Fixed Trash",
       "providesAddressableAreas": {
         "cutout12": ["fixedTrash"]
-      }
+      },
+      "height": 82
```
SyntaxColoring, line 402: I'm getting this value of 82 from the `opentrons_1_trash_1100ml_fixed` definition.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -127,7 +149,11 @@ def handle_action(self, action: Action) -> None:
             self._handle_command(action.command)
         if isinstance(action, PlayAction):
             current_state = self._state
-            if action.deck_configuration is not None:
+            if (
+                action.deck_configuration is not None
+                and not self._state.use_simulated_deck_config
+            ):
```
SyntaxColoring, line 155: I changed this check from `if action.deck_configuration is not None` to *also* check for `and not self._state.use_simulated.deck_config`. If you configured the engine to use a simulating deck configuration, it doesn't make sense to give it a real deck configuration later on.

Does this seem correct? This matches what this class's `__init__()` does.

shared-data/deck/schemas/4.json
sfoster1: I like zdimension better, and I think in the future we should probably move to a full boundingbox, but I'm fundamentally ok with what we've got for now

sfoster1 (approved):
Looks and sounds good to me! I think in the future we should move to `boundingBoxes` and `zDimensions` and so on but this is certainly good to merge and we can change it later.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -36,9 +36,30 @@ class AddressableAreaState:
...
 
     loaded_addressable_areas_by_name: Dict[str, AddressableArea]
+    """The addressable areas that have been loaded so far.
+
+    When `use_simulated_deck_config` is `False`, these are the addressable areas that the
+    deck configuration provided.
+
+    When `use_simulated_deck_config` is `True`, these are the addressable areas that have been
+    referenced by the protocol so far.
+    """
```
SyntaxColoring, line 46: Is this correct? @jbleon95

api/src/opentrons/protocol_engine/state/addressable_areas.py
jbleon95: Yes, we should never be reloading the deck configuration during analysis in the first place, but this is a correct check.

PR #14083, opened 2023-12-04
refactor(protocol-designer): reimporting protocol with deck config fixes

jerader:
closes RAUT-890

# Overview

When you reimport a protocol with staging area slots and waste chute, they don't populate correctly. But this PR fixes that. 

# Test Plan

Create a flex protocol and add a staging area and waste chute. Delete the trash bin. Add a labware and then add a transfer step! Also add a misc labware to be on top of the staging areas. (otherwise the staging areas aren't needed and won't repopulate during reimport)

Export the protocol. Open it up and change the version from `7.0.2` to `8.0.0` (this version change happens during the release process so we have to manually change it whenever we want to reimport a protocol)

Reimport the protocol. See that the staging area slots and waste chute are added as expected. Then go to the deck view and see that none of the timeline steps have an error.

I also attached this protocol you can use! It has 2 staging area slots in A4 and B4 and a waste chute (plus some modules):
[doItAllV8.json](https://github.com/Opentrons/opentrons/files/13546687/doItAllV8.json)

# Changelog

- fix the reducer logic for staging areas to look for `addressableAreaName` instead of `slotName`
- fix the reducer logic for waste chute by not generating a new Id and instead, grabbing the Id from the step form steps (autogenerating was causing 2 different waste chute Ids to appear in the protocol!)
- fix the `getUnusedStagingArea` logic by making it opposite and cleaning it up and also looking for `addressableAreaName`. Then fix the test

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/step-forms/reducers/index.ts
+++ b/protocol-designer/src/step-forms/reducers/index.ts
@@ -1350,16 +1350,6 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
               command.params.addressableAreaName
             ))
       )
-      const wasteChuteId = `${uuid()}:wasteChute`
-      const wasteChute = hasWasteChuteCommands
-        ? {
-            [wasteChuteId]: {
-              name: 'wasteChute' as const,
-              id: wasteChuteId,
-              location: 'cutoutD3',
-            },
-          }
-        : {}
 
       const getStagingAreaSlotNames = (
         commandType: 'moveLabware' | 'loadLabware',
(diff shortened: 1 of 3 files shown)


protocol-designer/src/step-forms/reducers/index.ts
```diff
@@ -1461,6 +1460,48 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
...
+          moveLiquidStepWasteChute.blowOut_location?.includes('wasteChute')
+        ) {
+          wasteChuteId = moveLiquidStepWasteChute.blowOut_location
+        }
+      }
+
+      const wasteChute =
+        hasWasteChuteCommands && wasteChuteId != null
+          ? {
+              [wasteChuteId]: {
```
jerader, line 1497: before, we were regenerating a new waste chute Id in this reducer. This was causing the protocol to have 2 different waste chute ids 😬

mjhuff (approved):
lgtm! 🗑️ 🚀

PR #14086, opened 2023-12-04
docs(api): refactor docstrings for labware.py

jwwojak:
# Overview

The [Labware and Wells section](https://docs.opentrons.com/v2/new_protocol_api.html#labware-and-wells) (labware.py) of the API docs has some mis-formatted docstrings. It looks like this is because authors are using a single back-tick instead of 2 to identify text that should be rendered in `monospace` code font. Instead of that, some text is either _italics_ or just has a single quote around it like this 'A1', 'A2', etc.

This PR goes after that text formatting issue, and a few other small but noticeable errors. Not making big changes to the docstrings themselves.

Also fixes a broken anchor in the `robot_position.rst` file.

# Test Plan

Make changes, build locally, make sure `text formatted as code` looks like `text formatted as code`.

# Changelog

Change improperly formatted docstring text to use the right code-style formatting.

# Review requests

Not looking at prose in this PR, just the text style. There may be minor changes to some text.

# Risk assessment

It's pretty low. Docstring changes only and 1 anchor.

### DIFF
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -643,8 +643,8 @@ class Labware:
         Accessor function used to create a look-up table of wells by name.
 
         With indexing one can treat it as a typical Python
-        dictionary whose keys are well names. To access well A1 for example,
-        write ``labware.wells_by_name()['A1']``.
+        dictionary whose keys are well names. For example, access well A1
+        with ``labware.wells_by_name()['A1']``.
 
         :return: Dictionary of well objects keyed by well name.
         """
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -611,7 +611,7 @@ class Labware:
         """
         Accessor function that generates a list of wells in a top down,
         left to right order. This is representative of moving down rows and
-        across columns (i.e., A1, B1, C1 . . . A2, B2, C2 . . . ).
+        across columns (i.e., A1, B1, C1…A2, B2, C2…).
 
         With indexing one can treat it as a typical python
         list. For example, access well A1 with ``labware.wells()[0]``.
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -611,7 +611,7 @@ class Labware:
         """
         Accessor function that generates a list of wells in a top down,
         left to right order. This is representative of moving down rows and
-        across columns (e.g., A1, B1, C1…A2, B2, C2…)
+        across columns (i.e., A1, B1, C1 . . . A2, B2, C2 . . . ).
 
         With indexing one can treat it as a typical python
         list. For example, access well A1 with ``labware.wells()[0]``.
@@ -646,7 +646,7 @@ class Labware:
         dictionary whose keys are well names. For example, access well A1
         with ``labware.wells_by_name()['A1']``.
 
-        :return: Dictionary of well objects keyed by well name.
+        :return: Dictionary of :py:class:`.Well` objects keyed by well name.
         """
         return dict(self._wells_by_name)
 
@@ -667,12 +667,12 @@ class Labware:
         Accessor function used to navigate through a labware by row.
 
         With indexing one can treat it as a typical python nested list.
-        To access row A for example, write ``labware.rows()[0]``. This
+        For example, access row A with ``labware.rows()[0]``. This
         will output ``['A1', 'A2', 'A3', 'A4'...]``.
 
         Note that this method takes args for backward-compatibility. But using args is deprecated and will be removed in future versions. Args can be either strings or integers, but must all be the same type. For example, ``self.columns(1, 4, 8)`` or ``self.columns('1', '2')`` are valid, but ``self.columns('1', 4)`` is not.
 
-        :return: A list of row lists
+        :return: A list of row lists.
         """
         if not args:
             return [
@@ -700,10 +700,10 @@ class Labware:
         Accessor function used to navigate through a labware by row name.
 
         With indexing one can treat it as a typical python dictionary.
-        To access row A for example, write ``labware.rows_by_name()['A']``.
+        For example, access row A with ``labware.rows_by_name()['A']``.
         This will output ``['A1', 'A2', 'A3', 'A4'...]``.
 
-        :return: Dictionary of Well lists keyed by row name.
+        :return: Dictionary of :py:class:`.Well` lists keyed by row name.
         """
         return {
             row_name: [self._wells_by_name[well_name] for well_name in row]
@@ -725,7 +725,7 @@ class Labware:
         Accessor function used to navigate through a labware by column.
 
         With indexing one can treat it as a typical python nested list.
-        To access row A for example, write ``labware.columns()[0]``.
+        For example, access row A with ``labware.columns()[0]``.
         This will output ``['A1', 'B1', 'C1', 'D1'...]``.
 
         Note that this method takes args for backward-compatibility. But using args is deprecated and will be removed in future versions. Args
@@ -761,11 +761,10 @@ class Labware:
         Accessor function used to navigate through a labware by column name.
 
         With indexing one can treat it as a typical python dictionary.
-        To access row A for example,
-        write ``labware.columns_by_name()['1']``.
+        For example, access row A with ``labware.columns_by_name()['1']``.
         This will output ``['A1', 'B1', 'C1', 'D1'...]``.
 
-        :return: Dictionary of Well lists keyed by column name.
+        :return: Dictionary of :py:class:`.Well` lists keyed by column name.
         """
         return {
             column_name: [self._wells_by_name[well_name] for well_name in column]
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -611,7 +611,7 @@ class Labware:
         """
         Accessor function that generates a list of wells in a top down,
         left to right order. This is representative of moving down rows and
-        across columns (e.g. ``'A1'``, ``'B1'``, ``'C1'...'A2'``, ``'B2'``, ``'C2'``)
+        across columns (i.e., A1, B1, C1…A2, B2, C2…)
 
         With indexing one can treat it as a typical python
         list. To access well A1 for example, write ``labware.wells()[0]``.
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -596,7 +596,7 @@ class Labware:
 
     @requires_version(2, 0)
     def well(self, idx: Union[int, str]) -> Well:
-        """Deprecated---use result of ``wells`` or ``wells_by_name``."""
+        """Deprecated. Use result of :py:meth:`wells` or :py:meth:`wells_by_name`."""
         if isinstance(idx, int):
             return self.wells()[idx]
         elif isinstance(idx, str):
@@ -611,7 +611,7 @@ class Labware:
         """
         Accessor function that generates a list of wells in a top down,
         left to right order. This is representative of moving down rows and
-        across columns (i.e., A1, B1, C1…A2, B2, C2…)
+        across columns (e.g., A1, B1, C1…A2, B2, C2…)
 
         With indexing one can treat it as a typical python
         list. For example, access well A1 with ``labware.wells()[0]``.
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -642,7 +642,7 @@ class Labware:
         """
         Accessor function used to create a look-up table of wells by name.
 
-        With indexing one can treat it as a typical python
+        With indexing one can treat it as a typical Python
         dictionary whose keys are well names. To access well A1 for example,
         write ``labware.wells_by_name()['A1']``.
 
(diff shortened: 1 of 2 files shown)


jwwojak:
@ecormany 

Would you also take a look at lines 374-382. This is for the ``parent`` method. I think what it's trying to do is show some options that get returned as strings, but I'm not sure and these might be better in a bullet list.

```
Returns:
            If the labware is directly on the robot's deck, the ``str`` name of the deck slot,
            like ``"D1"`` (Flex) or ``"1"`` (OT-2). See :ref:`deck-slots`.

            If the labware is on a module, a :py:class:`ModuleContext`.

            If the labware is on a labware or adapter, a :py:class:`Labware`.

            If the labware is off-deck, :py:obj:`OFF_DECK`.
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -596,7 +596,7 @@ def calibrated_offset(self) -> Point:
 
     @requires_version(2, 0)
     def well(self, idx: Union[int, str]) -> Well:
-        """Deprecated---use result of `wells` or `wells_by_name`"""
+        """Deprecated---use result of ``wells`` or ``wells_by_name``."""
```
ecormany: Would be even better to turn these into `:py:meth:` links to those methods (even though they're very nearby on the page).

api/src/opentrons/protocol_api/labware.py
```diff
@@ -609,20 +609,16 @@ def well(self, idx: Union[int, str]) -> Well:
     @requires_version(2, 0)
     def wells(self, *args: Union[str, int]) -> List[Well]:
         """
-        Accessor function used to generate a list of wells in top -> down,
-        left -> right order. This is representative of moving down `rows` and
-        across `columns` (e.g. 'A1', 'B1', 'C1'...'A2', 'B2', 'C2')
+        Accessor function that generates a list of wells in a top down,
+        left to right order. This is representative of moving down rows and
+        across columns (e.g. ``'A1'``, ``'B1'``, ``'C1'...'A2'``, ``'B2'``, ``'C2'``)
```
ecormany: This list is caught somewhere between being a code representation and an English description of well names. I favor the latter in terms of its readability — plus there's no way to format the code-like version, with the ellipsis, as valid code. So: 
```suggestion
        across columns (i.e., A1, B1, C1…A2, B2, C2…)
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -609,20 +609,16 @@ def well(self, idx: Union[int, str]) -> Well:
...
 
         With indexing one can treat it as a typical python
-        list. To access well A1, for example, write: labware.wells()[0]
+        list. To access well A1 for example, write ``labware.wells()[0]``.
```
ecormany: I agree on eliminating the colon.
I like keeping the first comma. But better yet, move `For example, ` to the start of the sentence.
```suggestion
        list. For example, access well A1 with ``labware.wells()[0]``.
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -609,20 +609,16 @@ def well(self, idx: Union[int, str]) -> Well:
...
 
-        Note that this method takes args for backward-compatibility, but use
-        of args is deprecated and will be removed in future versions. Args
-        can be either strings or integers, but must all be the same type (e.g.:
-        `self.wells(1, 4, 8)` or `self.wells('A1', 'B2')`, but
-        `self.wells('A1', 4)` is invalid.
+        Note that this method takes args for backward-compatibility. But using args is deprecated and will be removed in future versions. Args can be either strings or integers, but must all be the same type. For example, ``self.columns(1, 4, 8)`` or ``self.columns('1', '2')`` are valid, but ``self.columns('1', 4)`` is not.
```
ecormany, line 619: I have Thoughts on this note but we should save them for another day. 😅

api/src/opentrons/protocol_api/labware.py
```diff
@@ -644,13 +640,13 @@ def wells(self, *args: Union[str, int]) -> List[Well]:
     @requires_version(2, 0)
     def wells_by_name(self) -> Dict[str, Well]:
         """
-        Accessor function used to create a look-up table of Wells by name.
+        Accessor function used to create a look-up table of wells by name.
 
         With indexing one can treat it as a typical python
```
ecormany:
```suggestion
        With indexing one can treat it as a typical Python
```
🐍

api/src/opentrons/protocol_api/labware.py
```diff
@@ -644,13 +640,13 @@ def wells(self, *args: Union[str, int]) -> List[Well]:
...
-        dictionary whose keys are well names. To access well A1, for example,
-        write: labware.wells_by_name()['A1']
+        dictionary whose keys are well names. To access well A1 for example,
+        write ``labware.wells_by_name()['A1']``.
```
ecormany: Suggesting same change as for `wells()` above.
```suggestion
        dictionary whose keys are well names. For example, access well A1
        with ``labware.wells_by_name()['A1']``.
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -671,14 +667,10 @@ def rows(self, *args: Union[int, str]) -> List[List[Well]]:
         Accessor function used to navigate through a labware by row.
 
         With indexing one can treat it as a typical python nested list.
-        To access row A for example, write: labware.rows()[0]. This
-        will output ['A1', 'A2', 'A3', 'A4'...]
+        To access row A for example, write ``labware.rows()[0]``. This
```
ecormany: Once again…
```suggestion
        For example, access row A with ``labware.rows()[0]``. This
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -671,14 +667,10 @@ def rows(self, *args: Union[int, str]) -> List[List[Well]]:
...
+        will output ``['A1', 'A2', 'A3', 'A4'...]``.
```
ecormany, line 671: This one is also quasi-code (the ellipsis wouldn't parse as a Python list), but this formatting seems right. We're visually truncating actual valid output. The ellipsis comes along for the ride.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -708,10 +700,10 @@ def rows_by_name(self) -> Dict[str, List[Well]]:
         Accessor function used to navigate through a labware by row name.
 
         With indexing one can treat it as a typical python dictionary.
-        To access row A for example, write: labware.rows_by_name()['A']
-        This will output ['A1', 'A2', 'A3', 'A4'...].
+        To access row A for example, write ``labware.rows_by_name()['A']``.
```
ecormany: Wow, there's a lot of these, huh? Same as above.

And the same below, a couple more times.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -708,10 +700,10 @@ def rows_by_name(self) -> Dict[str, List[Well]]:
...
+        This will output ``['A1', 'A2', 'A3', 'A4'...]``.
 
-        :return: Dictionary of Well lists keyed by row name
+        :return: Dictionary of Well lists keyed by row name.
```
ecormany: Probably follow the change you made on line 647 above, lowercasing `well`. The alternative is to make it a reference to the (capitalized) class with 
```
:py:class:`.Well`
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -835,7 +825,7 @@ def tip_length(self, length: float) -> None:
             raise APIVersionError("Labware.tip_length setter has been deprecated")
 
         # TODO(mc, 2023-02-06): this assert should be enough for mypy
-        # invvestigate if upgrading mypy allows the `cast` to be removed
+        # investigate if upgrading mypy allows the `cast` to be removed
```
ecormany, line 827: 😁 gotta fix even the invisible typos

api/src/opentrons/protocol_api/labware.py
```diff
@@ -671,14 +667,10 @@ def rows(self, *args: Union[int, str]) -> List[List[Well]]:
...
+        will output ``['A1', 'A2', 'A3', 'A4'...]``.
 
-        Note that this method takes args for backward-compatibility, but use
-        of args is deprecated and will be removed in future versions. Args
-        can be either strings or integers, but must all be the same type (e.g.:
-        `self.rows(1, 4, 8)` or `self.rows('A', 'B')`, but  `self.rows('A', 4)`
-        is invalid.
+        Note that this method takes args for backward-compatibility. But using args is deprecated and will be removed in future versions. Args can be either strings or integers, but must all be the same type. For example, ``self.columns(1, 4, 8)`` or ``self.columns('1', '2')`` are valid, but ``self.columns('1', 4)`` is not.
 
         :return: A list of row lists
```
ecormany:
```suggestion
        :return: A list of row lists.
```

api/src/opentrons/protocol_api/labware.py
```diff
@@ -772,10 +762,10 @@ def columns_by_name(self) -> Dict[str, List[Well]]:
 
         With indexing one can treat it as a typical python dictionary.
         To access row A for example,
-        write: labware.columns_by_name()['1']
-        This will output ['A1', 'B1', 'C1', 'D1'...].
+        write ``labware.columns_by_name()['1']``.
+        This will output ``['A1', 'B1', 'C1', 'D1'...]``.
 
-        :return: Dictionary of Well lists keyed by column name
+        :return: Dictionary of Well lists keyed by column name.
```
ecormany: `well` or 
```
:py:class:`.Well`
```
…to match what you choose above.

ecormany (approved):
Everything here improves the formatting of these docstrings. My inline comments are only further improvements at a similar level.

These docstrings could definitely stand additional improvement at a conceptual level, but that's certainly a matter for another PR.

api/src/opentrons/protocol_api/labware.py
jwwojak: And now I'm only just seeing that colon in the parenthesis.  `(e.g.:   )`

It's a fever swamp to be sure.

api/src/opentrons/protocol_api/labware.py
jwwojak: I will do that, change these to links to the methods.

api/src/opentrons/protocol_api/labware.py
jwwojak: I found 4. Will fix as you recommend throughout.

api/src/opentrons/protocol_api/labware.py
jwwojak: I think linking to resources is great and will do as you recommend.

api/src/opentrons/protocol_api/labware.py
jwwojak: Will turn that into a link.

api/src/opentrons/protocol_api/labware.py
jwwojak: Fixing separately with a push of other changes.

api/src/opentrons/protocol_api/labware.py
jwwojak: Will leave in place.

api/src/opentrons/protocol_api/labware.py
jwwojak: Fixing in a separate push of changes.

jwwojak: I've found at least 4 instances of these.

api/src/opentrons/protocol_api/labware.py
jwwojak: Lost my comment from unresolving the change.

Was saying, if we're treating this more like English, the ellipses ought to be formatted properly with a space before and after and between the periods:  " . . . "

jwwojak: Oh, but it looks like Sphinx changes that to " ... "

api/src/opentrons/protocol_api/labware.py
jwwojak: Will fix in separate push.

api/src/opentrons/protocol_api/labware.py
jwwojak: I've found 4 so far. Making changes to all.

api/src/opentrons/protocol_api/labware.py
jwwojak: Oops. 🐍

api/src/opentrons/protocol_api/labware.py
jwwojak: Will fix in separate push.

api/src/opentrons/protocol_api/labware.py
jwwojak: changing to a link, and the other instance you note on line 647 (now 649 w/ recent changes).

api/src/opentrons/protocol_api/labware.py
ecormany: Proper English ellipses should be single characters: … (option-; on Mac)

api/src/opentrons/protocol_api/labware.py
jwwojak: 🍰   Taking a cue from Black.

api/src/opentrons/protocol_api/labware.py
jwwojak: ✅

api/src/opentrons/protocol_api/labware.py
jwwojak: Going with a link to the class instead of plain text.

api/src/opentrons/protocol_api/labware.py
jwwojak: 🏴󠁧󠁢󠁥󠁮󠁧󠁿  Fixing this. I really always thought there's a space between the periods. My life now feels like a fraud.

PR #14087, opened 2023-12-04
refactor(api): update PAPI Deck class to work with deck configuration changes

jbleon95:
# Overview

Addresses RSS-413.

This PR updates the `ProtocolContext.Deck` class and underlying engine functions to support deck configuration work.

The `Deck` class, when using `len` on it, iterating through it, or calling `Deck.slots` will now use all 16 possible slots for a Flex (the 12 standard slots and the 4 staging slots) regardless of the actual deck configuration, both in analysis and on the robot. These are purely references to slot information and do not signify the existence of the slot. OT-2 and 2.15 protocols on Flex will return 12 slots with the same behavior as before.

Additionally, all methods in the `Deck` class that took a deck slot as an argument now also take staging slots, including referencing the deck with dictionary-esque access and deleting of deck items.

# Test Plan

Tested the following protocols to make sure they passed analysis.

```
from opentrons.protocol_api import OFF_DECK

metadata = {
    'protocolName': 'New Deck Test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}

ORDERED_SLOTS = ["D1", "D2", "D3", "C1", "C2", "C3", "B1", "B2", "B3", "A1", "A2", "A3", "D4", "C4", "B4", "A4"]


def run(context):
    # Deck returns 16 standard slots always on Flex 2.16 and above
    deck = context.deck
    assert len(deck) == 16
    for i, slot in enumerate(deck):
        assert slot == ORDERED_SLOTS[i]

    # Loading a labware in a standard slot doesn't change anything, since that is already assumed
    labware_a1 = context.load_labware('opentrons_flex_96_tiprack_50ul', 'A1')
    assert deck['A1'] is labware_a1
    assert labware_a1.parent == 'A1'

    # Deleting the labware also doesn't change anything, deck slot is still there
    del deck['A1']
    assert labware_a1.parent == OFF_DECK

    # Checking if there is a deck item on a staging slot does NOT add it to the deck configuration assumption
    assert deck['A4'] is None

    # But loading a labware does
    labware_a4 = context.load_labware('opentrons_flex_96_tiprack_50ul', 'A4')
    assert deck['A4'] is labware_a4
    assert labware_a4.parent == 'A4'

    # Deleting a labware from a staging slot does not get rid of it either
    del deck['A4']
    assert labware_a4.parent == OFF_DECK
```

```
from opentrons.protocol_api import OFF_DECK

metadata = {
    'protocolName': 'New Deck Test Flex 2.15',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.15"
}

ORDERED_SLOTS = ["D1", "D2", "D3", "C1", "C2", "C3", "B1", "B2", "B3", "A1", "A2", "A3"]


def run(context):
    # Deck returns 12 standard slots always for 2.15 on Flex
    deck = context.deck
    assert len(deck) == 12
    for i, slot in enumerate(deck):
        assert slot == ORDERED_SLOTS[i]
```

and

```
from opentrons.protocol_api import OFF_DECK

metadata = {
    'protocolName': 'New Deck Test OT-2',
}

requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.16"
}

ORDERED_SLOTS = ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12"]


def run(context):
    # Deck returns 12 standard slots always on OT-2
    deck = context.deck
    assert len(deck) == 12
    for i, slot in enumerate(deck):
        assert slot == ORDERED_SLOTS[i]

    # Loading a labware in a standard slot doesn't change anything, since that is already assumed
    labware_10 = context.load_labware('opentrons_flex_96_tiprack_50ul', '10')
    assert deck['10'] is labware_10
    assert labware_10.parent == '10'

    # Deleting the labware also doesn't change anything, deck slot is still there
    del deck['10']
    assert labware_10.parent == OFF_DECK
```

# Changelog

- Refactored `Deck` class so that `__len__`, `__iter__` and `slots` return 16/16 slots for a Flex on 2.16 and above and 12 slots otherwise
- Allow using staging slots in many `Deck` functions including `Deck[slot_name]` access and `del Deck[slot_name]` functionality
- Correctly return slot name if the protocol engine location of a labware is an `AddressableAreaLocation`

# Review requests

Does the new logic for assumed slots make sense, and are we ok with allowing and OT-2 protocol to reference staging slots even if nothing can ever be functionally loaded on them.

# Risk assessment

Medium. Slightly changing some lesser known public API, most changes are straightforward.


sfoster1 (approved):
This looks really good, thanks!

api/src/opentrons/protocol_api/deck.py
```diff
@@ -194,7 +156,16 @@ def left_of(self, slot: DeckLocation) -> Optional[DeckItem]:
...
+        west_slot: Optional[DeckLocation]
+        if isinstance(slot_name, DeckSlotName):
+            west_slot = adjacent_slots_getters.get_west_slot(slot_name.as_int())
+        else:
+            west_slot = {
+                StagingSlotName.SLOT_A4: "A3",
+                StagingSlotName.SLOT_B4: "B3",
+                StagingSlotName.SLOT_C4: "C3",
+                StagingSlotName.SLOT_D4: "D4",
+            }[slot_name]
```
SyntaxColoring: Any chance we can move this mapping to `adjacent_slots_getters`?

I see there's a mismatch in type conventions, where all of the functions in `adjacent_slots_getters()` are taking plain `int`s. We don't have to resolve that for all of them. We can just add a new `get_west_of_staging_slot(StagingSlotName)` function.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -248,15 +263,26 @@ def _validate_addressable_area_for_simulation(
...
 
+    def _update_assumed_slots_for_deck(self, cutout_id: str) -> None:
+        """Update assumed deck slot for analysis run and remove if no longer possible in a deck configuration."""
+        deck_slot = CUTOUT_TO_DECK_SLOT_MAP[cutout_id]
+        potential_addressable_areas: Set[str] = set()
+        for fixtures in self._state.potential_cutout_fixtures_by_cutout_id[cutout_id]:
+            potential_addressable_areas.update(fixtures.provided_addressable_areas)
+        if deck_slot.id not in potential_addressable_areas:
+            self._state.assumed_slots_for_deck.discard(deck_slot.id)
+
```
SyntaxColoring: I'm not following what this is doing, sorry—could you clarify?

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -37,6 +39,7 @@ class AddressableAreaState:
 
     loaded_addressable_areas_by_name: Dict[str, AddressableArea]
     potential_cutout_fixtures_by_cutout_id: Dict[str, Set[PotentialCutoutFixture]]
+    assumed_slots_for_deck: Set[str]
```
SyntaxColoring: I'd find this confusing in Protocol Engine state. This seems like something very specific to a very small corner of the *Python Protocol API.* Nothing *in Protocol Engine* is "assuming" any slots, right?

Is it possible to redefine this without introducing new concepts to Protocol Engine? For example, is there a way to compute the same value given the information that's already in state? And/or tweak the state to only deal with things that are really intrinsic to Protocol Engine, like "the addressable areas that have been referenced so far," or "all possible addressable areas"?

SyntaxColoring: #14082 gives us the full deck configuration here in `AddressableAreaState`, if that makes any of that easier.

SyntaxColoring: If we do stick with this exact piece of state, could we have a comment of some kind to indicate what it means and why it's necessary?

SyntaxColoring (commented):
TY! Some questions:

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -427,14 +454,14 @@ def get_fixture_height(self, cutout_fixture_name: str) -> float:
...
-    def get_slot_definition(self, slot: DeckSlotName) -> SlotDefV3:
+    @lru_cache(maxsize=16)
+    def get_slot_definition(self, slot_id: str) -> SlotDefV3:
         """Get the definition of a slot in the deck."""
         try:
-            # TODO This should be the regular `get_addressable_area` once Robot Server deck config and tests is hooked up
-            addressable_area = self._get_addressable_area_for_simulation(slot.id)
-        except (AreaNotInDeckConfigurationError, IncompatibleAddressableAreaError):
+            addressable_area = self._get_addressable_area_from_deck_data(slot_id)
+        except AssertionError:
```
SyntaxColoring: Is this raising from the `deck_configuration_provider.get_potential_cutout_fixtures()` call inside of `_get_addressable_area_from_deck_state()`? Let's make it a real exception type instead of `AssertionError`. Partially for traceability but also because [`assert` statements can be disabled](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -449,3 +459,13 @@ def get_slot_definition(self, slot: DeckSlotName) -> SlotDefV3:
             "displayName": addressable_area.display_name,
             "compatibleModuleTypes": addressable_area.compatible_module_types,
         }
+
+    def get_deck_slot_definitions(self) -> Dict[str, SlotDefV3]:
+        """Get all slot definitions either configured for robot or assumed for analysis run so far."""
```
SyntaxColoring: Docstring update

api/src/opentrons/protocol_api/core/protocol.py
```diff
@@ -179,14 +179,19 @@ def set_last_location(
...
     @abstractmethod
-    def get_slot_definition(self, slot: DeckSlotName) -> SlotDefV3:
+    def get_slot_definition(
+        self, slot: Union[DeckSlotName, StagingSlotName]
+    ) -> SlotDefV3:
         """Get the slot definition from the robot's deck."""
 
+    @abstractmethod
+    def get_slot_definitions(self) -> Dict[str, SlotDefV3]:
+        """Get the configured or assumed slot definitions from the robot's deck."""
```
SyntaxColoring: Docstring update

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -570,9 +577,15 @@ def get_deck_definition(self) -> DeckDefinitionV4:
...
-    def get_slot_definition(self, slot: DeckSlotName) -> SlotDefV3:
+    def get_slot_definition(
+        self, slot: Union[DeckSlotName, StagingSlotName]
+    ) -> SlotDefV3:
         """Get the slot definition from the robot's deck."""
-        return self._engine_client.state.addressable_areas.get_slot_definition(slot)
+        return self._engine_client.state.addressable_areas.get_slot_definition(slot.id)
+
+    def get_slot_definitions(self) -> Dict[str, SlotDefV3]:
+        """Get the configured or assumed slot definitions from the robot's deck."""
```
SyntaxColoring: Docstring update

api/src/opentrons/protocol_api/core/legacy/legacy_protocol_core.py
```diff
@@ -480,17 +480,23 @@ def get_deck_definition(self) -> DeckDefinitionV4:
...
 
-    def get_slot_definition(self, slot: DeckSlotName) -> SlotDefV3:
+    def get_slot_definition(
+        self, slot: Union[DeckSlotName, StagingSlotName]
+    ) -> SlotDefV3:
         """Get the slot definition from the robot's deck."""
         assert False, "get_slot_definition only supported on engine core"
 
+    def get_slot_definitions(self) -> Dict[str, SlotDefV3]:
+        """Get the configured or assumed slot definitions from the robot's deck."""
```
SyntaxColoring: Docstring update

api/src/opentrons/protocol_api/validation.py
```diff
@@ -186,15 +185,18 @@ def ensure_and_convert_deck_slot(
...
 
     This normalizes the string to the robot type's native format, like "5" for OT-2s or "C2" for
     Flexes. This probably won't change anything because the internal `DeckSlotName` should already
     match the robot's native format, but it's nice to have an explicit interface barrier.
     """
-    return slot_name.to_equivalent_for_robot_type(robot_type).id
+    if isinstance(slot_name, DeckSlotName):
+        return slot_name.to_equivalent_for_robot_type(robot_type).id
+    else:
+        return slot_name.id
```
SyntaxColoring, line 200: Possibly helpful comment. Up to you.

```suggestion
        # No need to convert staging slot names per robot type, since they only exist on Flex.
        return slot_name.id
```

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -336,7 +347,7 @@ def _get_loaded_addressable_area(
                 f"{addressable_area_name} not provided by deck configuration."
             )
 
-    def _get_addressable_area_for_simulation(
+    def _get_addressable_area_from_deck_data(
```
SyntaxColoring, line 368: This shouldn't block this PR, but I don't know if this name is jiving with me. I think the phrase "deck data" is confusing to me because I don't know if it's talking about the *definition* or the *configuration* or both or what.

SyntaxColoring (commented):
Minor suggestions if we go with the latest ("eager") approach.

sfoster1 (approved):
Okay, sounds good to me.

PR #14089, opened 2023-12-04
chore(api): fix protocol engine tests to test both OT3 API and OT2 API when needed

ahiuchingau:
# Overview
There are some tests in `api/tests/opentrons/protocol_engine` were using only the OT2 API when the same behavior should be observed for the OT3 as well. This PR updates the pytest fixture so it conditionally yield both OT3 and OT2 API. 

So I ended up having to use `Any` as the OT2 or OT3 API type for some tests because mypy didn't like the union type that is HardwareControlAPI. Is there a way to use the handy `@pytest.mark.ot3_only/ot2_only` marker to tell mypy to only check for the proper type so we can keep using the union type?

### DIFF
--- a/api/tests/opentrons/protocol_engine/commands/calibration/test_calibrate_gripper.py
+++ b/api/tests/opentrons/protocol_engine/commands/calibration/test_calibrate_gripper.py
@@ -9,7 +9,7 @@ from decoy import Decoy
 from typing import TYPE_CHECKING
 
 from opentrons.hardware_control import ot3_calibration
-from opentrons.hardware_control.api import API as OT2API
+from opentrons.hardware_control import HardwareControlAPI
 from opentrons.hardware_control.types import GripperProbe, OT3Mount
 from opentrons.hardware_control.instruments.ot3.instrument_calibration import (
     GripperCalibrationOffset,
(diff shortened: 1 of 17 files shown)


api/tests/opentrons/protocol_engine/commands/calibration/test_calibrate_gripper.py
```diff
@@ -53,18 +48,18 @@ def _mock_ot3_calibration(decoy: Decoy, monkeypatch: pytest.MonkeyPatch) -> None
 )
 async def test_calibrate_gripper(
     decoy: Decoy,
-    ot3_hardware_api: OT3API,
+    hardware_api: Any,
```
SyntaxColoring, line 51: Yeah I think we need to figure this `Any` thing out before merging this. Do you know what mypy error you were getting?

SyntaxColoring (changes requested):
Thanks for the cleanup work. Would you mind providing some more background on this change?

I understand the distinction between "OT-2 API" or "OT-3 API" or "either API," but in all of these tests, we're mocking the API out anyway. There's very little distinction between `decoy.mock(cls=API)` and `decoy.mock(cls=OT3API)`—as far as I know, it just affects `isinstance()` checks.

Is this just a clarity thing? Like, the subject in `test_move_relative()` is supposed to be totally agnostic to what kind of robot it's running on, so it's confusing if the test specifically requests a hardware API for one robot?

Or is there an actual functional difference that you're worried about? Is something doing `isinstance()`?

PR #14090, opened 2023-12-05
feat(app): add recalibrate button in ProtocolSetupInstruments

mjhuff:
Closes RQA-1949

# Overview

This PR adds a recalibrate button for instruments within the ProtocolSetup > Instruments page. 

### Current Behavior
Screenshot 2023-12-05 at 8 50 46 AM

### New Behavior

Screenshot 2023-11-29 at 9 56 24 AM

# Test Plan
- Navigate to ProtocolSetup for any protocol that has an instrument (already calibrated). The recalibrate button should now be visible, and it should properly launch the recalibration flows.

# Changelog
- Added new recalibrate button in ProtocolSetupInstruments.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/InstrumentMountItem/ProtocolInstrumentMountItem.tsx
+++ b/app/src/organisms/InstrumentMountItem/ProtocolInstrumentMountItem.tsx
@@ -14,7 +14,6 @@ import {
   JUSTIFY_FLEX_START,
 } from '@opentrons/components'
 import {
-  CompletedProtocolAnalysis,
   getGripperDisplayName,
   getPipetteNameSpecs,
   NINETY_SIX_CHANNEL,
@@ -47,7 +46,6 @@ export const MountItem = styled.div<{ isReady: boolean }>`
 `
 interface ProtocolInstrumentMountItemProps {
   mount: Mount | 'extension'
-  mostRecentAnalysis?: CompletedProtocolAnalysis | null
   attachedInstrument: InstrumentData | null
   speccedName: PipetteName | GripperModel
   instrumentsRefetch?: () => void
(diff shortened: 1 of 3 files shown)


app/src/organisms/InstrumentMountItem/__tests__/ProtocolInstrumentMountItem.test.tsx
```diff
@@ -139,7 +139,7 @@ describe('ProtocolInstrumentMountItem', () => {
     fireEvent.click(button)
     getByText('pipette wizard flow')
   })
-  it('renders the correct information when gripper needs to be atached', () => {
+  it('renders the correct information when gripper needs to be attached', () => {
```
jerader, line 142: 🖊️

app/src/organisms/InstrumentMountItem/ProtocolInstrumentMountItem.tsx
```diff
@@ -165,7 +172,6 @@ export function ProtocolInstrumentMountItem(
           closeFlow={() => setShowPipetteWizardFlow(false)}
           selectedPipette={selectedPipette}
           mount={mount as Mount}
-          pipetteInfo={mostRecentAnalysis?.pipettes}
```
jerader, line 168: why isn't `pipetteInfo` needed anymore?

jerader (commented):
looks good, I left a question about the `pipetteInfo` prop

app/src/organisms/InstrumentMountItem/ProtocolInstrumentMountItem.tsx
mjhuff: Good question! So this prop actually blocks the recalibration flow from occurring if we keep it. If you look in getPipetteWizardSteps on L19-20, this pipetteInfo is used to return an empty array....

So after doing some digging, it seems that the entire reason we sometimes pass in pipetteInfo (at times) is to prevent unnecessary calibration if the pipette is already calibrated. I think we do this because we don't actually have FLOWS.RECALIBRATE for pipettes like we do the gripper, and this is some attempt to work around that. In all spots where we allow pipette recalibration, we don't pass this prop in. 

It's confusing for sure, and I think we could probably clean up the pipetteFlows at some point. I did thoroughly test the attach/calibrate/recalibrate flows without the prop here, and it didn't seem to be a problem.

jerader: Cool thank you for the explanation!

jerader (approved):
lgtm!

PR #14091, opened 2023-12-05
refactor(api): add explicit Flex hardware protocol

fsinapi:
# Overview

We currently differentiate between the OT3 and OT2 API classes with a simple `isinstance` check. This works okay when the simulator for each class is an instance of the normal hardware API, but there are a few reasons that we want to move over to having a separate simulator class for each API that _does not depend on the hardware-specific code_. In that case, checking the types to differentiate the robot model doesn't work.

This PR establishes extensions to the Hardware API MyPy protocols in order to explicitly differentiate an OT-2 from an OT-3. It also adds a function that returns the enumerated model of the robot (OT2 or FLEX). This is preferable to type-checking the class because `isinstance` is slow/not recommended on Protocol classes.

# Test Plan

Check that
- Calibration still works
- Model-specific endpoints are filtered properly (can't send `GET /instruments` on OT-2)
- Protocol engine can run a protocol fine

# Changelog

# Review requests

# Risk assessment

Medium. Touches some core stuff but doesn't really change any behavior.


api/src/opentrons/hardware_control/api.py
```diff
@@ -302,6 +302,9 @@ async def build_hardware_simulator(
     def __repr__(self) -> str:
         return "<{} using backend {}>".format(type(self), type(self._backend))
 
+    def get_robot_type(self) -> RobotTypeEnum:
```
sfoster1: If you define this to return an enum literal `Literal[RobotTypeEnum.OT2]`, and define the Flex protocol's version of the method to return `Literal[RobotTypeEnum.FLEX]` (or however we spelled it), then in theory that should let the union of the two protocols get narrowed by the result of calling `get_robot_type()` which should remove some casts

api/src/opentrons/protocol_engine/resources/ot3_validation.py
```diff
@@ -1,28 +1,22 @@
...
-        ) from exception
-
-    if not isinstance(hardware_api, OT3API):
+    if hardware_api.get_robot_type() != RobotTypeEnum.FLEX:
         raise HardwareNotSupportedError(
             error_msg or "This command is supported by OT-3 only."
         )
 
-    return hardware_api
+    return hardware_api  # type: ignore
```
sfoster1: here this cast should be removable if you do the enum-literal thing above

sfoster1 (approved):
Love it, big step forward. Can we do enum literal returns for the robot_type and get mypy automatically narrowing union types?

api/src/opentrons/hardware_control/api.py
fsinapi: 629ac675efa272abf75ccc795086fa69dc06dfc1

api/src/opentrons/protocol_engine/resources/ot3_validation.py
fsinapi: MyPy disagrees :(

api/src/opentrons/hardware_control/ot3_calibration.py
```diff
@@ -47,7 +47,7 @@
 from .util import DeckTransformState
 
 if TYPE_CHECKING:
```
sfoster1, line 49: we can take this out of `TYPE_CHECKING`, right? it's just contained there to prevent transitive includes of canbus and now that won't happen

api/src/opentrons/protocol_engine/resources/ot3_validation.py
```diff
@@ -1,28 +1,22 @@
...
-from typing import TYPE_CHECKING, Optional
+from typing import Optional, TYPE_CHECKING
 
 from opentrons.protocol_engine.errors import HardwareNotSupportedError
+from opentrons.hardware_control.protocols.types import FlexRobotType
 
 if TYPE_CHECKING:
-    from opentrons.hardware_control.ot3api import OT3API
-    from opentrons.hardware_control import HardwareControlAPI
+    from opentrons.hardware_control import HardwareControlAPI, OT3HardwareControlAPI
```
sfoster1: here too - shouldn't need TYPE_CHECKING?

robot-server/robot_server/instruments/router.py
```diff
@@ -49,7 +49,7 @@
 from robot_server.subsystems.router import status_route_for, update_route_for
 
 if TYPE_CHECKING:
-    from opentrons.hardware_control.ot3api import OT3API
+    from opentrons.hardware_control import OT3HardwareControlAPI
```
sfoster1: no `TYPE_CHECKING`?

sfoster1 (approved):
Can we remove TYPE_CHECKING? otherwise looks good

api/src/opentrons/hardware_control/ot3_calibration.py
fsinapi: There's a circular dependency for this one actually, can either leave in `TYPE_CHECKING` or move the `OT3Transforms` class somewhere else

sfoster1 (approved):
Awesome, let's do it!

PR #14092, opened 2023-12-05
style(app): reformat TC run profile command text on ODD

mjhuff:
Closes RQA-1937

# Overview

This PR special cases the ODD TC runProfile command text to match designs. We introduce isOnDevice as a prop passed to CommandText, which doesn't feel great, but this feels like the best way to handle it. 

### Current

Protocol running step copy not truncating and indentation bug

### Restyled

Screenshot 2023-12-05 at 10 45 11 AM

# Test Plan
- Run any protocol that has a TC run profile (here's one: [tc_temp_cycle_test.py.zip](https://github.com/Opentrons/opentrons/files/13562598/tc_temp_cycle_test.py.zip) )
- Confirm ODD matches designs (and nothing on the desktop version of runProfile has changed). 

# Changelog
- Truncated ODD thermocycler run profile display text.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/CommandText/index.tsx
+++ b/app/src/organisms/CommandText/index.tsx
@@ -1,7 +1,12 @@
 import * as React from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { Flex, DIRECTION_COLUMN, SPACING } from '@opentrons/components'
+import {
+  Flex,
+  DIRECTION_COLUMN,
+  SPACING,
+  ALIGN_CENTER,
+} from '@opentrons/components'
 import { getPipetteNameSpecs, RunTimeCommand } from '@opentrons/shared-data'
 import {
   getAddressableAreaDisplayName,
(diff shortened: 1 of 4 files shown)


app/src/organisms/CommandText/index.tsx
```diff
@@ -96,20 +108,31 @@ export function CommandText(props: Props): JSX.Element | null {
       const { profile } = command.params
       const steps = profile.map(
         ({ holdSeconds, celsius }: { holdSeconds: number; celsius: number }) =>
-          t('tc_run_profile_steps', { celsius: celsius, seconds: holdSeconds })
+          t('tc_run_profile_steps', {
+            celsius: celsius,
+            seconds: holdSeconds,
+          }).trim()
```
jerader, line 114: nice i've never used trim before! I think i've always removed with a css property

jerader (approved):
lgtm! this is optional but you could add a test case to the CommandText component to test this.

koji (approved):
The changes look good to me.
Thank you for fixing this!

PR #14095, opened 2023-12-05
feat(api): add get tip presence command in protocol engine

ahiuchingau:
# Overview
This PR adds the `get_tip_presence` command in protocol engine so we can actually get the realtime tip presence response from the pipette, instead of having to rely on the instrument endpoint which only returns the cached tip detected value.
Closes RQA-1992

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/verify_tip_presence.py
+++ b/api/src/opentrons/protocol_engine/commands/verify_tip_presence.py
@@ -1,0 +1,77 @@
+"""Verify tip presence command request, result and implementation models."""
+from __future__ import annotations
+
+from pydantic import Field, BaseModel
+from typing import TYPE_CHECKING, Optional, Type
+from typing_extensions import Literal
+
+from .pipetting_common import PipetteIdMixin
+from .command import AbstractCommandImpl, BaseCommand, BaseCommandCreate
+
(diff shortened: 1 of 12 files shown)


api/src/opentrons/protocol_engine/commands/verify_tip_presence.py
```diff
@@ -0,0 +1,79 @@
...
+        **kwargs: object,
+    ) -> None:
+        self._tip_handler = tip_handler
+
+    async def execute(self, params: VerifyTipPresenceParams) -> VerifyTipPresenceResult:
+        """Verify if tip presence is as expected for the requested pipette."""
+        pipette_id = params.pipetteId
+        expected_state = params.expectedState
+
+        result = await self._tip_handler.verify_tip_presence(
```
sfoster1: I think I'd expect this to raise, because the only reason that I would call this instead of get tip presence is to simplify a status check that should fail.

sfoster1 (changes requested):
I think it actually makes more sense for the verify to raise, and for verify to be something that's really very rarely called and only intentionally when you want it to raise.

sfoster1 (approved):
Okay, looks good to me! Should fix the tip presence error too!

sfoster1 (approved):

PR #14100, opened 2023-12-05
fix(api,shared-data): Correct addressable area positions

SyntaxColoring:
# Overview

This closes RSS-408.

# Test Plan

* [x] 8-Channel drops tips in a waste chute in the right place
* [x] 1-Channel drops tips in a waste chute in the right place
* [x] 96-Channel drops tips in a waste chute in the right place
* [x] Gripper drops labware in the waste chute in the right place
* [x] All pipettes drop tips in movable trash in the right place
* [x] The gripper and all pipettes safely clear the rim of the waste chute

I've used these protocols to test: [Addressable area tests.zip](https://github.com/Opentrons/opentrons/files/13591162/Addressable.area.tests.zip)

I only have an old waste chute to test with, so I couldn't directly test the y-position for 1- and 8-channel tip drops. (I cannot use the new waste chute cover.) So I held up the cover while the protocol was paused and eyeballed it.

# Changelog

* Update the positions of all waste chute addressable areas to match the new hardware, which has a lower cover with a different hole position.
* ⚠️ **Breaking:** Correct the positioning of the 1-channel pipette by splitting the `1and8ChannelWasteChute` addressable area into `1ChannelWasteChute` and `8ChannelWasteChute`.
    * This is necessary because we want the drop-tip position to be same for their *rear channels,* not their *centers.*
    * This will cause our demo JSON protocols to fail with an "unknown addressable area" error. Python protocols will be fine.
    * The JavaScript half of this is done in #14109.
* Fix a Python Protocol API bug that had the 8-channel and 96-channel positions swapped with each other.
* Refactor the deck definition to remove some redundant, unused, and confusing fields.
    * Delete `dropTipsOffset` from the addressable area definitions. This was unused.
    * Delete `dropLabwareOffset` from the addressable area definitions. This was used by Protocol Engine.
    * To replace both of those, move the addressable area `offsetFromCutoutFixture`s and `boundingBox`es so the bounding box's top-center is at the expected interaction point.
        * For motion planning, this was not safe to do before, but it is now that #14082 has merged.
        * For display in the Opentrons App, this is safe because it's only looking at these positions for *deck slots,* which are not affected here. Only the waste chute and trash bins are affected.

# Review requests

None in particular.

# Risk assessment

Medium. This affects motion planning, so we need to manually test a bunch of pipette positions and make sure there are no collisions.


jerader (approved):
thank you! This makes sense to me. I tested uploading a JSON file where i manually changed the `1and8ChannelWasteChute` addressable area name to `1ChannelWasteChute` and it passed analysis 🥳

sanni-t (approved):
Reviewed in person. I'll go through the calculations thoroughly for my understanding later but the logic makes sense  and looks good to merge!

PR #14102, opened 2023-12-06
fix(app): Improve ODD "run again" behavior

mjhuff:
Closes RQA-1972, RQA-1973

# Overview

This PR adds a loading state to the "Run Again" button on the ODD. 

It also fixes a routing bug in which pressing "run again" would send the user to the protocol setup page, back to the /protocols page, then back to the protocol setup page. Changes to TopLevelRedirect did not account for ODD "run again" routing logic. By removing the onResetSuccess callback, we avoid the superfluous redirect.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/2b5fcbf6-d6ab-43e7-b1be-9cd59889e832

### Fixed Behavior

https://github.com/Opentrons/opentrons/assets/64858653/92bcfa5e-6564-4792-9272-6229a959bf6d

# Test Plan
- Follow the video flow above. Just verify that there's loading state and there is no redirection to the /protocols route after pressing "run again".

# Changelog
- Added loading state to the "Run again" button the ODD.
- Fixed a routing issue when pressing "Run again" on the ODD.

# Risk assessment
low

### DIFF
--- a/app/src/pages/OnDeviceDisplay/RunSummary.tsx
+++ b/app/src/pages/OnDeviceDisplay/RunSummary.tsx
@@ -64,7 +64,6 @@ import { handleTipsAttachedModal } from '../../organisms/DropTipWizard/TipsAttac
 import { getPipettesWithTipAttached } from '../../organisms/DropTipWizard/getPipettesWithTipAttached'
 import { getPipetteModelSpecs, FLEX_ROBOT_TYPE } from '@opentrons/shared-data'
 
-import type { Run } from '@opentrons/api-client'
 import type { OnDeviceRouteParams } from '../../App/types'
 import type { PipetteModelSpecs } from '@opentrons/shared-data'
 
@@ -106,9 +105,7 @@ export function RunSummary(): JSX.Element {
     runStatus === RUN_STATUS_FAILED || runStatus === RUN_STATUS_SUCCEEDED
   )
   const { trackProtocolRunEvent } = useTrackProtocolRunEvent(runId)
-  const onResetSuccess = (_createRunResponse: Run): void =>
-    history.push(`/runs/${runId}/setup`)
-  const { reset } = useRunControls(runId, onResetSuccess)
+  const { reset } = useRunControls(runId)
   const trackEvent = useTrackEvent()
   const { closeCurrentRun, isClosingCurrentRun } = useCloseCurrentRun()
   const localRobot = useSelector(getLocalRobot)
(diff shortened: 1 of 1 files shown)


app/src/pages/OnDeviceDisplay/RunSummary.tsx
```diff
@@ -197,6 +198,19 @@ export function RunSummary(): JSX.Element {
       })
   }, [])
 
+  const RUN_AGAIN_SPINNER_TEXT = (
```
mjhuff, line 201: This isn't as clean as I'd like. If we end up using the spinner icon in the upper-right corner as a pattern on the ODD, we should extend the large button component to support a spinner.

The only other place we use a spinner on click on the ODD is the recently run protocols page, but we don't use a button there. I think this is another spot that we could refactor to use a new large button w/ loading spinner component. 

But yeah, out of scope for this PR. I'll probably ticket this for future work.

app/src/pages/OnDeviceDisplay/RunSummary.tsx
```diff
@@ -399,9 +416,24 @@ const SummaryDatum = styled.div`
...
   width: max-content;
 `
-
 const DURATION_TEXT_STYLE = css`
   font-size: ${TYPOGRAPHY.fontSize22};
   line-height: ${TYPOGRAPHY.lineHeight28};
   font-weight: ${TYPOGRAPHY.fontWeightRegular};
 `
+
+const RUN_AGAIN_CLICKED_STYLE = css`
```
jerader, line 425: it feels a bit weird to me that `LargeButton` would special case a css prop for when the spinner is showing, given that `LargeButton` is meant to be a design component that encompasses all the stylings. Looks like `LargeButton` is only used in this component so I think its safe to move this to be directly in `LargeButton`.

app/src/pages/OnDeviceDisplay/RunSummary.tsx
jerader: I agree we should refactor large button to support a spinner!

jerader (approved):
lgtm, I think we should definitely refactor `LargeButton` to support a spinner and the spinner css state. Looks like `LargeButton` is only used in this component so I _think_ it'd be a quick refactor. But feel free to make a follow up ticket for that instead to do post-launch!

mjhuff:
> lgtm, I think we should definitely refactor `LargeButton` to support a spinner and the spinner css state. Looks like `LargeButton` is only used in this component so I _think_ it'd be a quick refactor. But feel free to make a follow up ticket for that instead to do post-launch!

Thanks for your feedback! I didn't notice that we are using LargeButton only here... I'm going to ticket this out separately, because the more I think through it, the more it seems like this should involve some architecting. 🎨

PR #14103, opened 2023-12-06
fix(app): always enable BackToTop button

ncdiehl11:
closes [RQA-1954](https://opentrons.atlassian.net/browse/RQA-1954)

# Overview

Remove checks for disabling `BackToTop` button at the last setup step and remove hover tooltips for disabled reason. The user should always be enabled to click `BackToTop` in `ProtocolRun` even if proceeding to run is disabled.

Note: I set the disabled prop for the button to always `false` because the only time component is rendered is within `ProtocolRun`. If we utilize the button elsewhere, we can add a prop for disabling.

# Test Plan

- Proceed to protocol setup that should not be ready to run (missing calibrations, incorrect pipettes, etc.)
- Observe that the `BackToTop` button renders after the last setup step and is enabled.

# Changelog

- remove disabling `BackToTop` inside component
- update tests to always expect the button to be enabled

# Risk assessment

low

[RQA-1954]: https://opentrons.atlassian.net/browse/RQA-1954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
@@ -9,11 +9,6 @@ import {
   useTrackEvent,
   ANALYTICS_PROTOCOL_PROCEED_TO_RUN,
 } from '../../../../redux/analytics'
-import {
-  useRunCalibrationStatus,
-  useRunHasStarted,
-  useUnmatchedModulesForProtocol,
-} from '../../hooks'
 import { BackToTopButton } from '../BackToTopButton'
 
 jest.mock('@opentrons/components', () => {
@@ -24,17 +19,7 @@ jest.mock('@opentrons/components', () => {
   }
 })
 jest.mock('../../../../redux/analytics')
-jest.mock('../../hooks')
 
-const mockUseUnmatchedModulesForProtocol = useUnmatchedModulesForProtocol as jest.MockedFunction<
-  typeof useUnmatchedModulesForProtocol
->
-const mockUseRunCalibrationStatus = useRunCalibrationStatus as jest.MockedFunction<
-  typeof useRunCalibrationStatus
->
-const mockUseRunHasStarted = useRunHasStarted as jest.MockedFunction<
-  typeof useRunHasStarted
->
 const mockUseTrackEvent = useTrackEvent as jest.MockedFunction<
   typeof useTrackEvent
 >
@@ -62,20 +47,6 @@ let mockTrackEvent: jest.Mock
 
 describe('BackToTopButton', () => {
   beforeEach(() => {
-    when(mockUseUnmatchedModulesForProtocol)
-      .calledWith(ROBOT_NAME, RUN_ID)
-      .mockReturnValue({
-        missingModuleIds: [],
-        remainingAttachedModules: [],
-      })
-
-    when(mockUseRunCalibrationStatus)
-      .calledWith(ROBOT_NAME, RUN_ID)
-      .mockReturnValue({
-        complete: true,
-      })
-    when(mockUseRunHasStarted).calledWith(RUN_ID).mockReturnValue(false)
-
     mockTrackEvent = jest.fn()
     when(mockUseTrackEvent).calledWith().mockReturnValue(mockTrackEvent)
   })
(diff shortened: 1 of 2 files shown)


app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
```diff
@@ -102,19 +102,18 @@ describe('BackToTopButton', () => {
     })
   })
 
-  it('should be disabled with modules not connected tooltip when there are missing moduleIds', () => {
+  it('should not disabled with modules not connected tooltip when there are missing moduleIds', () => {
```
jerader:
```suggestion
  it('should not disable with modules not connected tooltip when there are missing moduleIds', () => {
```

jerader (commented):
looks good, `BackToTopButton` test suite can be cleaned up a little bit. I think lines 28-37 can be deleted since we no longer need those mocks.

app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
ncdiehl11: my bad, went through the test descriptions too quickly. fixing now

ncdiehl11: Correct me if I'm wrong, but don't we need those mocks? For example, the test "should not be disabled with modules not connected tooltip when there are missing moduleIds" requires `mockUseUnmatchedModulesForProtocol`. The difference is that we are expecting the button to be enabled under this set of conditions now.

jerader: Nah, you deleted the usages of the hooks in the component, so every instance of  `useRunCalibrationStatus`, `useRunHasStarted`, and `useUnmatchedModulesForProtocol` can be deleted, including in the individual test cases

app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
```diff
@@ -102,19 +102,18 @@ describe('BackToTopButton', () => {
...
       })
-    const { getByRole, getByText } = render()
+    const { getByRole } = render()
     const button = getByRole('button', { name: 'Back to top' })
-    expect(button).toBeDisabled()
-    getByText('Make sure all modules are connected before proceeding to run')
+    expect(button).not.toBeDisabled()
   })
-  it('should be disabled with modules not connected and calibration not completed tooltip if missing cal and moduleIds', async () => {
+  it('should not be disabled with modules not connected and calibration not completed tooltip if missing cal and moduleIds', async () => {
```
jerader: oh also lets delete the mention of tooltip in the test case names since we no longer render a tooltip
```suggestion
  it('should not be disabled if missing cal and moduleIds', async () => {
```

app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
```diff
@@ -126,32 +125,25 @@ describe('BackToTopButton', () => {
...
+    const { getByRole } = render()
     const button = getByRole('button', { name: 'Back to top' })
-    expect(button).toBeDisabled()
-    getByText(
-      'Make sure robot calibration is complete and all modules are connected before proceeding to run'
-    )
+    expect(button).not.toBeDisabled()
   })
-  it('should be disabled with calibration not complete tooltip when calibration not complete', async () => {
+  it('should not be disabled with calibration not complete tooltip when calibration not complete', async () => {
```
jerader:
```suggestion
  it('should not be disabled when calibration not complete', async () => {
```

app/src/organisms/Devices/ProtocolRun/__tests__/BackToTopButton.test.tsx
```diff
@@ -126,32 +125,25 @@ describe('BackToTopButton', () => {
...
+    const { getByRole } = render()
     const button = getByRole('button', { name: 'Back to top' })
-    expect(button).toBeDisabled()
-    getByText(
-      'Make sure robot calibration is complete before proceeding to run'
-    )
+    expect(button).not.toBeDisabled()
   })
-  it('should be disabled with protocol run started tooltip when run has started', async () => {
+  it('should not be disabled with protocol run started tooltip when run has started', async () => {
```
jerader:
```suggestion
  it('should not be disabled when run has started', async () => {
```

jerader (approved):
lgtm! thanks for cleaning up the test 🦕

PR #14105, opened 2023-12-06
fix(app): detach gripper go back button alignment

ncdiehl11:
closes [RQA-1969](https://opentrons.atlassian.net/browse/RQA-1969)

# Overview

When detaching gripper, if you confirm detachment without physically detaching, you are prompted
with a message stating that the gripper is still attached. There is a Go Back button on this screen
that needs left padding.

Note— should probably add padding to the SimpleWizardBody css for desktop, but holding off for now because lots of components implicated.

# Test Plan

Screen Shot 2023-12-06 at 11 30 17 AM

# Risk assessment

low

[RQA-1969]: https://opentrons.atlassian.net/browse/RQA-1969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/GripperWizardFlows/UnmountGripper.tsx
+++ b/app/src/organisms/GripperWizardFlows/UnmountGripper.tsx
@@ -108,6 +108,7 @@ export const UnmountGripper = (
         justifyContent={JUSTIFY_SPACE_BETWEEN}
         alignItems={isOnDevice ? ALIGN_CENTER : ALIGN_FLEX_END}
         gridGap={SPACING.spacing8}
+        paddingLeft={isOnDevice ? 0 : SPACING.spacing32}
       >
         <Btn onClick={() => setShowGripperStillDetected(false)}>
           <StyledText css={GO_BACK_BUTTON_TEXT_STYLE}>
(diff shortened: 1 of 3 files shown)


app/src/organisms/GripperWizardFlows/UnmountGripper.tsx
```diff
@@ -109,7 +109,10 @@ export const UnmountGripper = (
         alignItems={isOnDevice ? ALIGN_CENTER : ALIGN_FLEX_END}
         gridGap={SPACING.spacing8}
       >
-        <Btn onClick={() => setShowGripperStillDetected(false)}>
+        <Btn
+          paddingLeft={isOnDevice ? 0 : SPACING.spacing32}
```
jerader, line 113: Maybe we could use the `justifyContentForOddButton` prop in `SimpleWizardBody` and then delete the Flex 🤔 it just feels a bit weird to special case this here when many components follow the same pattern/designs. Here is an example where this pattern is being used: `app/src/organisms/PipetteWizardFlows/ExitModal.tsx`

ncdiehl11: Screen Shot 2023-12-06 at 2 38 26 PM

Following that pattern looks like the above because of different button styles passed as children between this case and `ExitModal`

jerader: okay thanks for this, lets leave it the way you had it!

jerader (approved):
lgtm!

PR #14116, opened 2023-12-06
fix(app): fix styling of read-only deck configurator

ncdiehl11:
closes [RQA-1987](https://opentrons.atlassian.net/browse/RQA-1987)

# Overview

When deck configurator is read only (during protocol/maintenance run), any loaded deck fixtures
should show default cursor on hover, not show remove icon, and not be clickable

# Test Plan

- Configure Flex deck with various fixtures on DeviceDetails
- Begin a protocol or maintenance run
- Move mouse over loaded fixtures and observe default cursor, no remove icon, and unselectable fixtures

https://github.com/Opentrons/opentrons/assets/47604184/e5ebaaf2-08c8-4283-9141-668d0e5206d1

# Risk assessment

low

[RQA-1987]: https://opentrons.atlassian.net/browse/RQA-1987?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/src/hardware-sim/DeckConfigurator/StagingAreaConfigFixture.tsx
+++ b/components/src/hardware-sim/DeckConfigurator/StagingAreaConfigFixture.tsx
@@ -47,7 +47,7 @@ export function StagingAreaConfigFixture(
       <Btn
         css={
           handleClickRemove != null
-            ? STAGING_AREA_CONFIG_STYLE
+            ? STAGING_AREA_CONFIG_STYLE_EDITABLE
             : STAGING_AREA_CONFIG_STYLE_READ_ONLY
         }
         cursor={handleClickRemove != null ? 'pointer' : 'default'}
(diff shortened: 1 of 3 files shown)


components/src/hardware-sim/DeckConfigurator/StagingAreaConfigFixture.tsx
```diff
@@ -84,3 +90,18 @@ const STAGING_AREA_CONFIG_STYLE = css`
     border: 3px solid ${COLORS.fundamentalsFocus};
   }
 `
+
```
brenthagen: not necessary, but since the read-only styles are a subset of the "editable" styles you should be able to reference one in the other to avoid duplicating code:

```
const STAGING_AREA_CONFIG_STYLE_EDITABLE = css`
  ${STAGING_AREA_CONFIG_STYLE_READ_ONLY}

  &:active {
    background-color: ${COLORS.darkBlack90};
  }

  &:hover {
    background-color: ${COLORS.grey1};
  }
`
```

components/src/hardware-sim/DeckConfigurator/StagingAreaConfigFixture.tsx
```diff
@@ -84,3 +90,18 @@ const STAGING_AREA_CONFIG_STYLE = css`
...
+  display: ${DISPLAY_FLEX};
+  align-items: ${ALIGN_CENTER};
+  background-color: ${COLORS.grey2};
+  border-radius: ${BORDERS.borderRadiusSize1};
+  color: ${COLORS.white};
+  grid-gap: ${SPACING.spacing8};
+  justify-content: ${JUSTIFY_CENTER};
+  width: 100%;
+
+  &:focus-visible {
```
brenthagen: do we want focus-visible in the read-only state?

brenthagen (approved):
looks good, a couple small comments

components/src/hardware-sim/DeckConfigurator/StagingAreaConfigFixture.tsx
ncdiehl11: good call, I wasn't aware of this. Thanks

components/src/hardware-sim/DeckConfigurator/StagingAreaConfigFixture.tsx
ncdiehl11: Likely not. I will change this.

PR #14118, opened 2023-12-06
fix(app): surface errors between attach pipette and attach probe

smb2268:
fix RQA-1968

# Overview

We should surface any errors that come up on the Results screen, which could happen if we issue commands between strung together flows (ie between attach & calibrate)
# Test Plan

1. Attach a pipette
2. Before proceeding from the successful attachment screen, attach the calibration probe
3. See that the overpressure error from probe being attached during plunger home is now surfaced

# Changelog
Surface error message

# Review requests

# Risk assessment
Low


sfoster1 (approved):
It's actually really hard to get an error to happen here and I can't manage to do it but it does look correct, so I think we should merge this and move forward.

PR #14119, opened 2023-12-06
feat(hardware-control): disengage 96-channel and gripper z after retract

ahiuchingau:
# Overview
We should disengage the 96-channel pipette mount and the gripper mount motor whenever they're not active. This helps reduce current consumption and lower the chances of them dropping due to noise.

Note: This only affect the `_cache_and_maybe_retract` func in the hardware control API, this means you would only see this change if `ot3api.move_to()/move_rel()` is used.

# Test Plan
Make sure protocols with a 96-channel pipette and a gripper still work like normal

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1305,9 +1305,25 @@ class OT3API(
         (and :py:attr:`_last_moved_mount` exists) then retract the mount
         in :py:attr:`_last_moved_mount`. Also unconditionally update
         :py:attr:`_last_moved_mount` to contain `mount`.
+
+        Disengage the 96-channel and gripper mount if retracted.
         """
         if mount != self._last_moved_mount and self._last_moved_mount:
             await self.retract(self._last_moved_mount, 10)
+
+            # disengage Axis.Q motor and engage the brake to lower power
+            # consumption and reduce the chance of the 96-channel pipette dropping
+            if (
+                self.gantry_load == GantryLoad.HIGH_THROUGHPUT
+                and self._last_moved_mount == OT3Mount.LEFT
+            ):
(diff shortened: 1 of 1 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1293,9 +1293,25 @@ async def _cache_and_maybe_retract_mount(self, mount: OT3Mount) -> None:
         (and :py:attr:`_last_moved_mount` exists) then retract the mount
         in :py:attr:`_last_moved_mount`. Also unconditionally update
         :py:attr:`_last_moved_mount` to contain `mount`.
+
+        Disengage the 96-channel and gripper mount if retracted.
         """
         if mount != self._last_moved_mount and self._last_moved_mount:
             await self.retract(self._last_moved_mount, 10)
+
+            # disengage Axis.Q motor and engage the brake to lower power
```
Laura-Danielle:
```suggestion
            # disengage Axis.Z_L motor and engage the brake to lower power
```

Laura-Danielle (approved):
lgtm aside from small comment typo

ahiuchingau:
After talking with @andySigler, we've decided to not merge this until some more hardware testing is done on the 96-channel pipette; the reason being that the changes here might have an effect on the overall temperature of the pipette and affect the pipetting volume.

andySigler:
sry for the delay, we'll be doing this testing first thing next week. The test is to leave the 96ch disengaged for several hours, then engage it and immediately run a gravimetric test. Worst case scenario, the pipette will start heating up enough that we'll see a change in the average volume during the first maybe 10x trials. That is the concern, that the change in this PR would create a scenario where one of our pipettes was not able to reach a stable temperature before a protocol is ran.

andySigler (approved):
does not cause changes in volume aspirated from heating up (my only concern)

PR #14120, opened 2023-12-06
fix(app): filter lpc labware combos by single slot

brenthagen:
# Overview

filters the labware location combos generated by `getLabwareLocationCombos` by Flex or OT-2 single slot addressable area name, to exclude labware in staging area slots as LPC options

closes RQA-1989

# Test Plan

# Changelog

 -  Filters lpc labware combos by single slot

# Review requests

check LPC on Flex and OT-2 - labware in staging area slots should not be available. confirm nothing else is broken

# Risk assessment

low


mjhuff (approved):
lgtm!

PR #14121, opened 2023-12-06
fix(app): make button cta text the same across odd and app

jerader:
closes RQA-2010

# Overview

Fixes the button for updating a deck fixture so the text is the same between ODD and app

# Test Plan

Look at the location conflict modal when there is a fixture conflict on ODD and app and see that the confirm button text says "update deck" for both.

# Changelog

- update the button text to say `update deck` only for both cases, and remove the branching logic for the app button
- fix test
- remove now unused i18n string

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/assets/localization/en/protocol_setup.json
+++ b/app/src/assets/localization/en/protocol_setup.json
@@ -42,7 +42,6 @@
   "configured": "configured",
   "confirm_heater_shaker_module_modal_description": "Before the run begins, module should have both anchors fully extended for a firm attachment. The thermal adapter should be attached to the module. ",
   "confirm_heater_shaker_module_modal_title": "Confirm Heater-Shaker Module is attached",
-  "confirm_removal": "Confirm removal",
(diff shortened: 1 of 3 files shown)


koji:
seems that both of them using `confirm removal`.

desktop
https://www.figma.com/file/nJBVk8a26wdls3ZnkgRmns/OCT-Release%3A-Desktop-%5BDeck-configuration-%26-Drop-tips%5D?node-id=1%3A384&mode=dev

odd
https://www.figma.com/file/eMAHM8g07dtTKr2RsFjkzB/OCT-Release%3A-Touchscreen-%5BDeck-configuration-%26-Drop-tips%5D?node-id=1%3A386&mode=dev

jerader:
> seems that both of them using `confirm removal`.
> 

@koji thanks for this! @ecormany mentioned in the ticket that `Update deck` makes more sense.

koji (approved):
lgtm

PR #14122, opened 2023-12-06
fix(app): add deck config query refetch interval

brenthagen:
# Overview

refetches the deck config on a 5 second interval on the device details page and in hooks that determine location conflicts

closes RQA-1993

# Test Plan

# Changelog

 - Adds deck config query polling

# Review requests

confirm that the deck config updates within 5 seconds when updated by another client

# Risk assessment

low


koji (approved):
lgtm

I think we need to update `app/src/pages/DeckConfiguration/index.tsx` too.

brenthagen:
> lgtm
> 
> I think we need to update `app/src/pages/DeckConfiguration/index.tsx` too.

that page is different because we are initializing based on the query then modifying the deck. i think we may want `staleTime: Infinity` there, but should be ticketed separately.

PR #14125, opened 2023-12-06
fix(api): Use enumerated error for tips

sfoster1:
Raises a proper enumerated error for the tip state checking failures in protocol engine.

Also fixes an issue where error chains were stringified by PE when they happened inside a python protocol (we were checking a top level thing when we needed to be looking one level lower)

Fixes RQA-2012


Laura-Danielle (approved):

sanni-t (approved):
🙌

PR #14126, opened 2023-12-06
refactor(app): verify probe presence before pipette calibration and labware position check

ahiuchingau:
# Overview
The app now checks for probe presence for pipette before performing pipette calibration and labware position check.

### DIFF
--- a/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
+++ b/app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
@@ -8,7 +8,7 @@ import {
   SPACING,
   RESPONSIVENESS,
 } from '@opentrons/components'
-import { LEFT, MotorAxes, WASTE_CHUTE_CUTOUT } from '@opentrons/shared-data'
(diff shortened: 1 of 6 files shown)


app/src/organisms/PipetteWizardFlows/AttachProbe.tsx
```diff
@@ -69,65 +73,64 @@ export const AttachProbe = (props: AttachProbeProps): JSX.Element | null => {
...
+      },
+      {
+        commandType: 'calibration/moveToMaintenancePosition' as const,
+        params: {
+          mount: mount,
+        },
+      },
+    ]
     setIsPending(true)
     refetch()
```
smb2268: I added this refetch in so that we would have the most up to date `attachedInstrument` state to check for the tip. Is that still necessary? Looks like it could be removed.

sfoster1:
The nesting definitely ended up being needed but @smb2268 you're right the instruments query was not, so ditched that. Tested on a robot.

smb2268 (approved):
Changes look good to me! Thanks for updating 🚀

PR #14127, opened 2023-12-06
fix(app): fix deck view diff between deck and labware deck map modal

koji:
# Overview
Fix deck map with staging area labware different for protocol details vs protocol setup.
The issue is the current implementation does not consider `addressableAreaName`.
close RQA-2008

https://github.com/Opentrons/opentrons/assets/474225/b93cfbff-885c-4634-9471-4915f28417de

# Test Plan
- compare Deck in Protocol Details screen and Deck map in the modal from Labware - Protocol Setup screen
- The deck map in the modal allows us to check a labware details when tapping it

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ProtocolSetupLabware/LabwareMapViewModal.tsx
+++ b/app/src/organisms/ProtocolSetupLabware/LabwareMapViewModal.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
-import map from 'lodash/map'
 import { useTranslation } from 'react-i18next'
+
 import { BaseDeck } from '@opentrons/components'
 import {
   FLEX_ROBOT_TYPE,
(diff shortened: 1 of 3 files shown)


app/src/organisms/ProtocolSetupLabware/LabwareMapViewModal.tsx
```diff
@@ -85,34 +79,68 @@ export function LabwareMapViewModal(
     }
   })
 
-  const labwareLocations = map(
-    labwareRenderInfo,
-    ({ labwareDef, slotName }, labwareId) => {
+  const labwareOnDeck = mostRecentAnalysis?.commands
```
brenthagen, line 82: instead of processing commands here, is it possible to use the new `ProtocolDeck` component like protocol details does?

koji: I think it's possible but, for this modal, we need to allow users to tap each labware to display the labware info modal and this functionality is for this screen.
I avoid using `ProtocolDeck` for this PR.

brenthagen:
i did some investigation, and I think the reason this bug is happening is because the deck definition changed the `areaType` of staging area slots from `slot` to `stagingSlot`. so the fix we want is:

 - update the `AreaType` union type in shared-data/js/types to include `'stagingSlot'`
 - update the `isSlot` boolean in `getLabwareRenderInfo` to check for `aa.areaType === 'stagingSlot'` in addition to `'slot'`

i think those two changes should be all that is needed, and should also fix rendering in `SetupLabwareMap`, etc

koji:
> i did some investigation, and I think the reason this bug is happening is because the deck definition changed the `areaType` of staging area slots from `slot` to `stagingSlot`. so the fix we want is:
> 
> * update the `AreaType` union type in shared-data/js/types to include `'stagingSlot'`
> * update the `isSlot` boolean in `getLabwareRenderInfo` to check for `aa.areaType === 'stagingSlot'` in addition to `'slot'`
> 
> i think those two changes should be all that is needed, and should also fix rendering in `SetupLabwareMap`, etc

@brenthagen thanks, I just followed your suggestion, and it worked as expected.
I'm going to close this and open another PR since it is easy to change this PR.

PR #14130, opened 2023-12-07
feat(api): raise an error when the gripper pickup fails

caila-marashaj:
## Overview
During and after we use the gripper to move labware inside of a protocol, before we move forward with the rest of the protocol, we should make sure that none of the following have happened:
- we've grabbed empty air
- we've grabbed the labware so that it's crooked
- the labware has fallen out of the gripper jaws somewhere along its journey between deck slots

This code raises an error if we think we should, during a call to `move_labware_with_gripper()`, have a piece of labware inside the gripper jaws, but the position of the gripper jaws can't affirm that that's actually the case.

## Changelog
- add `FailedGripperPickupError` inside `shared-data`, as a `RoboticsControlError`
- add a `max_allowed_grip_error` to `shared-data`. The value is negotiable but I started with 6 just because that seemed reasonable from hardware testing.
- create a function inside the `Gripper` class to check the gripper position against what we think it should be and potentially throw an error
- call said function inside `move_labware_with_gripper()`, only when we should have labware in the gripper's clutches
- some unit tests

## Review Requests
I want to see what we think of the logic I'm using to determine when to call the "check" function, `check_labware_pickup()`. The call happens inside `move_labware_with_gripper()`, in the loop where the gripper's movement waypoints are processed. After the first call to `ungrip()`, a boolean gets set to `True`. Only when this is `True`( i.e., in the following iterations of the loop), and `jaw_open` is `False`, do we call `check_labware_pickup()`.

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/gripper.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/gripper.py
@@ -22,7 +22,10 @@ from .instrument_calibration import (
 )
 from ..instrument_abc import AbstractInstrument
 from opentrons.hardware_control.dev_types import AttachedGripper, GripperDict
-from opentrons_shared_data.errors.exceptions import CommandPreconditionViolated
+from opentrons_shared_data.errors.exceptions import (
+    CommandPreconditionViolated,
+    FailedGripperPickupError,
+)
 
 from opentrons_shared_data.gripper import (
     GripperDefinition,
@@ -100,6 +103,10 @@ class Gripper(AbstractInstrument[GripperDefinition]):
     def remove_probe(self) -> None:
         assert self.attached_probe
         self._attached_probe = None
+
+    @property
+    def max_allowed_grip_error(self) -> float:
+        return self._geometry.max_allowed_grip_error
 
     @property
     def jaw_width(self) -> float:
@@ -193,6 +200,23 @@ class Gripper(AbstractInstrument[GripperDefinition]):
                 detail={
                     "probe": str(self._attached_probe),
                     "jaw_state": str(self.state),
+                },
+            )
+
+    def check_labware_pickup(self, labware_width: float) -> None:
+        """Ensure that a gripper pickup succeeded."""
+        # check if the gripper is at an acceptable position after attempting to
+        #  pick up labware
+        expected_gripper_position = labware_width
+        current_gripper_position = self.jaw_width
+        if (
+            abs(current_gripper_position - expected_gripper_position)
+            > self.max_allowed_grip_error
+        ):
+            raise FailedGripperPickupError(
+                details={
+                    "expected jaw width": expected_gripper_position,
+                    "actual jaw width": current_gripper_position,
                 },
             )
 
--- a/api/src/opentrons/protocol_engine/execution/labware_movement.py
+++ b/api/src/opentrons/protocol_engine/execution/labware_movement.py
@@ -22,7 +22,6 @@ from ..errors import (
     ThermocyclerNotOpenError,
     HeaterShakerLabwareLatchNotOpenError,
     CannotPerformGripperAction,
-    FailedGripperPickupError,
 )
 
 from ..types import (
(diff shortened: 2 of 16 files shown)


sfoster1 (approved):
This looks really good to me and looks like appropriate logic. I agree with it going into `edge` rather than this release so we can get it into ABR since it has the potential to add a bunch of surprising errors. Nice work!

api/src/opentrons/hardware_control/instruments/ot3/gripper.py
```diff
@@ -196,6 +203,23 @@ def check_calibration_pin_location_is_accurate(self) -> None:
                 },
             )
 
+    def check_labware_pickup(self, labware_width: float) -> None:
```
sanni-t, line 206: naming this something like `raise_if_labware_not_picked_up` makes it easier to quickly tell what it's doing.

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -146,12 +146,22 @@ async def move_labware_with_gripper(
...
                         # We lost the position estimation after disengaging the axis, so
                         # it is necessary to home it next
                         await ot3api.home_z(OT3Mount.GRIPPER)
                 else:
                     await ot3api.grip(force_newtons=labware_grip_force)
+                    # we only want to check position after the gripper has opened and
+                    # should be holding labware
+                    if gripper_opened:
+                        assert ot3api.hardware_gripper
+                        ot3api.hardware_gripper.check_labware_pickup(
```
sanni-t, line 160: Maybe `check_labware_pickup` should be a part of the ot3api since it's only going to be called from outside hardware controller. That'll also get rid of the assert above.

api/tests/opentrons/protocol_engine/execution/test_labware_movement_handler.py
```diff
@@ -116,6 +178,83 @@ def subject(
...
+    hardware_gripper_offset_data: Tuple[
+        LabwareMovementOffsetData, LabwareMovementOffsetData
+    ],
+) -> None:
+    """Test that the gripper position check is called at the right time."""
+    #  This function should only be called when after the gripper opens,
+    #  and then closes again. This is when we expect the labware to be
+    #  in the gripper jaws.
+    await set_up_decoy_hardware_gripper(decoy, ot3_hardware_api, state_store)
+    assert ot3_hardware_api.hardware_gripper
```
sanni-t, line 197: Why is this necessary?

api/tests/opentrons/hardware_control/test_gripper.py
```diff
@@ -67,6 +88,32 @@ def test_load_gripper_cal_offset(fake_offset: "GripperCalibrationOffset") -> Non
...
+    error_context: Any,
+) -> None:
+    """Test that FailedGripperPickupError is raised correctly."""
+    #  This should only be triggered when the difference between the
+    #  gripper jaw and labware widths is greater than the max allowed error.
+    gripr = gripper.Gripper(fake_gripper_conf, fake_offset, "fakeid123")
+    mock_jaw_width.return_value = jaw_width_val
+    mock_max_grip_error.return_value = 6
+    with error_context:
+        gripr.check_labware_pickup(85)
```
sanni-t, line 114: Nice test!

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -146,12 +146,22 @@ async def move_labware_with_gripper(
...
+                    gripper_opened = True
                     if waypoint_data.dropping:
                         # We lost the position estimation after disengaging the axis, so
                         # it is necessary to home it next
                         await ot3api.home_z(OT3Mount.GRIPPER)
                 else:
                     await ot3api.grip(force_newtons=labware_grip_force)
+                    # we only want to check position after the gripper has opened and
+                    # should be holding labware
+                    if gripper_opened:
```
sanni-t, line 158: So this will check whether the labware is being held at the end of each movement made while the gripper is holding the labware, ya? That's quite neat!
Can we call this var `gripper_holding_labware` or something? Makes it easier to understand.

sanni-t (approved):
LGTM! Just some nitpicks but not necessary to address in this PR.

api/tests/opentrons/protocol_engine/execution/test_labware_movement_handler.py
caila-marashaj: the linter throws an error about trying to access attributes of `NoneType` if I don't have that there

api/tests/opentrons/hardware_control/test_gripper.py
caila-marashaj: thank you!!!

api/src/opentrons/protocol_engine/execution/labware_movement.py
caila-marashaj: yes! ty!

PR #14104, opened 2023-12-06
feat(api): add tip tracking for partial tip configuration

sanni-t:
Closes RSS-299

# Overview

Adds automatic tip selection for pipettes in partial tip configuration. Also updates signature of `configure_nozzle_layout` to include list of tip racks to use during the configuration.

# Test Plan

Sample protocol for testing. 
- [ ] should raise errors where indicated
- [ ] once erroring lines are removed/fixed, it should analyze successfully and run on the robot correctly

```
from opentrons.protocol_api import COLUMN, ALL


requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(protocol_context):
	trash_labware = protocol_context.load_labware("opentrons_1_trash_3200ml_fixed", "A3")

	tip_rack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C3")
	tip_rack2 = protocol_context.load_labware("opentrons_flex_96_tiprack_200ul", "B3")
	tip_rack3 = protocol_context.load_labware("opentrons_flex_96_tiprack_200ul", "D3")


	instrument = protocol_context.load_instrument('flex_96channel_1000', mount="left", tip_racks=[tip_rack3, tip_rack2, tip_rack1])
	instrument.trash_container = trash_labware

	# Should pick all tips up from tiprack3
	instrument.pick_up_tip()
	instrument.drop_tip()

	instrument.configure_nozzle_layout(style=COLUMN, start="A12", tip_racks=[tip_rack1])

	# Should pick tips up column-wise from tiprack1
	for _ in range(12):
		instrument.pick_up_tip()
		instrument.drop_tip()

	# Should raise error. Comment out to fix
	instrument.pick_up_tip()

	######### CHANGE CONFIG TO ALL #########
	instrument.configure_nozzle_layout(style=ALL, tip_racks=[tip_rack2])

	# Should pick all tips up from tiprack2
	instrument.pick_up_tip()
	instrument.drop_tip()

	# Should raise error. Comment out to fix
	instrument.pick_up_tip()
```

# Changelog

- added a new `active_channels` getter to get the number of pipette channels that are actually in use
- updated pick_up_tip and tip update logic to use active channels instead of pipette static channels

# Review requests

- make sure all the tip tracking logic is updated in PAPI & engine
- test on a robot

# Risk assessment

Medium. Affects existing tip tracking logic

PR #13997, opened 2023-11-16
docs(api): edit docstring for `configure_nozzle_layout`

ecormany:
# Overview

Update the docstring for `configure_nozzle_layout`, first added in #13838. Goals for this PR:
- Clearly explain the function's parameters.
- Only explain layouts that will be supported at the release of PAPI v2.16.
  - Column and full-plate layouts for 96-channel.
  - No partial layouts for 8-channel.
- Remove code sample from API Reference. A later PR will add multiple code samples and explanation on the Pipettes page of the docs.

# Test Plan

[Reference entry in sandbox](http://sandbox.docs.opentrons.com/docstring-configure_nozzle_layout/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.configure_nozzle_layout)

# Changelog

- Fix RST formatting
- Edit docstring prose
- Temporarily remove explanation of `front_right` parameter
- Remove code snippet from reference
- Couple typo fixes in related/nearby core docstrings

# Review requests

- Make sure I've characterized features properly. I haven't been able to simulate protocols using this function yet, much less test them on hardware.

- Fact check this claim: `If you try to perform partial tip pickup on a tip rack that is in an adapter, the API will raise an error.`

- Check that I didn't mess anything up when rebasing and resolving conflicts.

# Risk assessment

low, docstrings + a comment hanging off the bottom of the docstring (can't comment _within_ the string)

PR #14132, opened 2023-12-07
docs(api): more partial tip pickup docstrings

ecormany:
# Overview

Expand on the initial docstring for `configure_nozzle_layout` (#13997) to include the new tip-tracking behavior (#14104).

# Test Plan

Check the entries in the sandbox.
- [active channels](http://sandbox.docs.opentrons.com/docstring-partial-tip_racks/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.active_channels)
- [configure nozzle layout](http://sandbox.docs.opentrons.com/docstring-partial-tip_racks/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.configure_nozzle_layout)

# Changelog

- Explained behavior of `tip_rack` parameter, with reference to same parameter of `load_instrument`, based on conversation with @sanni-t.
- Expanded docstring for related `active_channels` property.

# Review requests

This is how it works, yeah?

# Risk assessment

nil, docstrings.

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1818,11 +1818,12 @@ class InstrumentContext(publisher.CommandPublisher):
             should be of the same format used when identifying wells by name. 
             Required unless setting ``style=ALL``. 
 
-            .. warning::
-                Currently, when using the ``COLUMN`` layout, you *must* set
-                ``start="A12"``. Using any other value will cause the pipette to move to
-                unexpected locations, pick up incorrect tips, or crash into items on the
-                deck.
+            .. note::
+                When using the ``COLUMN`` layout, the only fully supported value is
+                ``start="A12"``. You can use ``start="A1"``, but this will disable tip
+                tracking and you will have to specify the ``location`` every time you
+                call :py:meth:`.pick_up_tip`, such that the pipette picks up columns of
+                tips *from right to left* on the tip rack.
 
         :type start: str or ``None``
         :param tip_racks: Behaves the same as setting the ``tip_racks`` parameter of
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1816,23 +1816,13 @@ class InstrumentContext(publisher.CommandPublisher):
             to determine how it will move to different locations on the deck. The string
             should be of the same format used when identifying wells by name. Use
             ``"A1"`` to have the pipette use its leftmost nozzles to pick up tips
-            *left-to-right* from a tip rack. Use ``"A12"`` to have the pipette use its
-            rightmost nozzles to pick up tips *right-to-left* from a tip rack.
+            *right-to-left* from a tip rack. Use ``"A12"`` to have the pipette use its
+            rightmost nozzles to pick up tips *left-to-right* from a tip rack.
         :type start: str or ``None``
         :param tip_racks: Behaves the same as setting the ``tip_racks`` parameter of
-            :py:meth:`.load_instrument`. If not specified,
-            :py:obj:`.InstrumentContext.tip_racks` will be reset to ``None`` and you
-            must specify the location every time you call
-            :py:meth:`~.InstrumentContext.pick_up_tip`.
-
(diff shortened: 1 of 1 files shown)


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1815,7 +1819,21 @@ def configure_nozzle_layout(
             *left-to-right* from a tip rack. Use ``"A12"`` to have the pipette use its
             rightmost nozzles to pick up tips *right-to-left* from a tip rack.
         :type start: str or ``None``
```
sanni-t: This is the opposite actually.. the rightmost nozzle column picks up tips left-to-right from the tiprack. Or am I not understanding this sentence corectly?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1815,7 +1819,21 @@ def configure_nozzle_layout(
...
-        :param tip_racks: List of tip racks to use during this configuration.
+        :param tip_racks: Behaves the same as setting the ``tip_racks`` parameter of
+            :py:meth:`.load_instrument`. If not specified,
+            :py:obj:`.InstrumentContext.tip_racks` will be reset to ``None`` and you
```
sanni-t: `tip_racks` is reset to an empty list actually

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1815,7 +1819,21 @@ def configure_nozzle_layout(
...
+            must specify the location every time you call
+            :py:meth:`~.InstrumentContext.pick_up_tip`.
+
+            To keep the same list of tip racks when changing configurations, reference
+            the same instrument's existing ``tip_racks`` parameter::
+
+                pipette.configure_nozzle_layout(
+                    style=COLUMN,
+                    tip_racks=pipette.tip_racks
+                )
```
sanni-t: I feel like this would be better as part of the high level Opentrons docs than the API method docstrings because a) it's not critical info and b) it doesn't give you the full and probably more important details of why you actually _shouldn't_ use the same tip racks when switching between the 96 channel's partial and full configurations.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1636,7 +1636,11 @@ def channels(self) -> int:
     @property  # type: ignore
     @requires_version(2, 16)
     def active_channels(self) -> int:
-        """The number of channels configured for active use using configure_nozzle_layout()."""
+        """The number of channels the pipette will use to pick up tips.
```
sanni-t, line 1639: This makes so much more sense than my complicated sentence 😆

api/src/opentrons/protocol_api/instrument_context.py
sanni-t: Actually, I just realized the note about 96 ch partial config tipracks already exists in this docstring. Then I'm cool keeping this here.

ecormany: Nah, I think you're right. Too clever by half for docstrings.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: Yes, you're right, it's wrong. I keep having to make hand motions every time I need to figure this out 😆

sanni-t (approved):
Changes look good! 

There's just one thing to be aware of- we currently only (fully) support setting the rightmost column of the 96 channel for partial config. If you set the A1 as primary nozzle then your pipette will move correctly but there's no deck conflict checking or auto tip tracking for it yet. But it won't raise any errors either. So maybe we should put a note in there that currently only A12 as primary nozzle is fully supported and the others are use it at your own risk.

ecormany:
> If you set the A1 as primary nozzle then your pipette will move correctly but there's no deck conflict checking

Talked with @anthonywilleee and we decided this is too risky to recommend, even with warnings. Gonna push another change to reflect that. I made a note on RTC-365 to come back to the A1 configuration later.

ecormany:
OK, I think this PR has finally settled into the same place we expect the feature to be at launch!

sanni-t (approved):
Nice! Changes since my last review look good!

PR #14137, opened 2023-12-07
refactor(shared-data): update pipetteNameSpecs flowrates

jerader:
# Overview

Update the `pipetteNameSpecs.json` to match the latest flow rates for the flex pipettes

# Test Plan

Verify that it matches the version 2 definitions

# Changelog

- update flow rates to match the version 2 definition values

# Review requests

see test plan

# Risk assessment

low

PR #14138, opened 2023-12-07
feat(protocol-designer): update the defaulted flowrates for blowout

jerader:
# Overview

The flowrates in the mix and moveLiquid form for blow out were defaulted to the dispense flow rates which is incorrect since in most cases, the blowout rates are much higher.

# Test Plan

Just verify that the code looks good. Uploading an old protocol will migrate this information but I included a blurb about it in the migration modal.

FYI: i cherrypicked a few commits from this pr that was merged into the release branch. The PR was approved by Seth and Sanniti and updates the `pipetteNameSpecs` file: https://github.com/Opentrons/opentrons/pull/14137 The reason why i cherrypicked is so i don't have to merge the release branch back into edge again. But LMK if this is a bad idea and if i should do that merge back.

# Changelog

- fix logic to return blowout default instead of dispense default
- fix cypress test
- add migration text
- add on to fixture

# Review requests

see test plan

# Risk assessment

low


sfoster1 (approved):
The values look good to me! We should probably try and trickle in some refactors to get stuff using less of the old pipette name specs with time.

PR #14139, opened 2023-12-07
fix(app, api): expose failOnNotHomed parameter for save_position command

mjhuff:
# Overview

Drop tip flows require saving pipette positions without homing the plunger axis (we definitely do not want to home the plunger). The save_position command should optionally expose a failOnNotHomed parameter to prevent an error from being thrown when the plunger is not homed - the plunger position value is not stored in saved_position, anyway. 

# Test Plan

- [x]  Run the drop tip wizard flows on a pipette such as the P300 8-Channel GEN2 and confirm via Network tab that the save position response is successful. NOTE: Drop tip flows will work on the OT-2 without a deckmap, just click Move to Slot - there's a default position. 
- [x] Test the flows on a couple pipettes and Flex/OT-2 to confirm there aren't any unforeseen issues.

# Changelog
- The savePosition command now takes an optional failOnNotHomed boolean. 

# Risk assessment
low

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/save_position.py
+++ b/api/src/opentrons/protocol_engine/commands/save_position.py
@@ -26,6 +26,9 @@ class SavePositionParams(BaseModel):
         description="An optional ID to assign to this command instance. "
         "Auto-assigned if not defined.",
     )
+    failOnNotHomed: Optional[bool] = Field(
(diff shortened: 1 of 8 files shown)


sfoster1 (approved):
Nice, nice work doing both sides!

PR #14144, opened 2023-12-07
fix(app): properly display no liquids used text in protocol details/setup

ncdiehl11:
closes [RQA-2006](https://opentrons.atlassian.net/browse/RQA-2006)

# Overview

When we decide whether or not to display liquids in ProtocolDetails or ProtocolSetup, we need to determine if the liquids are actually loaded in protocol (commands) rather than accessing the liquids property of the protocol analysis.

# Test Plan

- create protocol that specifies liquids but does not load them in any wells [example](https://github.com/Opentrons/opentrons/files/13605826/96_ch.Tip.Drop.200uL_Tip.Rack.Drop.Well.Plate.Drop.Waste.chute.no.cover.1.json)
- upload protocol to App
- confirm that ProtocolDetails and ProtocolSetup properly render "No liquids are specified for this protocol."

Screen Shot 2023-12-07 at 10 41 26 AM
Screen Shot 2023-12-07 at 4 57 31 PM

# Risk assessment

low

[RQA-2006]: https://opentrons.atlassian.net/browse/RQA-2006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
@@ -144,7 +144,7 @@ export function ProtocolRunSetup({
       ? parseLiquidsInLoadOrder(liquids, protocolAnalysis.commands)
       : []
 
-  const hasLiquids = protocolAnalysis != null && liquidsInLoadOrder.length > 0
+  const hasLiquids = liquidsInLoadOrder.length > 0
 
   const hasModules = protocolAnalysis != null && modules.length > 0
 
--- a/app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
@@ -1,7 +1,10 @@
 import * as React from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { parseAllRequiredModuleModels } from '@opentrons/api-client'
+import {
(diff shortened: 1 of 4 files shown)


app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
```diff
@@ -133,8 +136,16 @@ export function ProtocolRunSetup({
...
-    protocolAnalysis != null && protocolAnalysis.liquids?.length > 0
+
+  const liquids = protocolAnalysis?.liquids ?? []
+
+  const liquidsInLoadOrder =
+    protocolAnalysis != null
+      ? parseLiquidsInLoadOrder(liquids, protocolAnalysis.commands)
+      : []
+
+  const hasLiquids = protocolAnalysis != null && liquidsInLoadOrder.length > 0
```
koji: I think hasLiquieds just needs to check liquidsInLoadOrder.length 🤔 

```suggestion
  const hasLiquids =  liquidsInLoadOrder.length > 0
```

koji (approved):
lgtm!

app/src/organisms/Devices/ProtocolRun/ProtocolRunSetup.tsx
ncdiehl11: Good catch

PR #14145, opened 2023-12-07
fix(api): OT2 fixed trash load fix and API 2,15 support for new trash container structure

CaseyBatten:
# Overview
This seeks to close: RSS-417 

Also seeks to add coverage for API 2.15 protocols expected fixed trash behavior.

# Test Plan

- Ensure no analysis failures!
- Test a 2.16 Protocol on OT2, it should execute correctly
- Test a 2,16 protocol on Flex with multiple trashes
- Test a 2.15 protocol with trashing behavior on OT2 and Flex, legacy trash behavior should be expected.


ecormany (commented):
Thanks for the quick fix!

I pulled the branch and simulated the protocol included in RSS-417. Works as expected. 🚀 

Will let others test the rest and inspect the code.

SyntaxColoring:
> This seeks to close [...] https://opentrons.atlassian.net/browse/RQA-2022

Are we sure that this closes that? That looks like something from within Protocol Engine, unrelated to PAPI. Notice how the error message complains about a missing `fixedTrash` (camelCase, like it appears in Protocol Engine commands and the deck definition), not a missing `fixed_trash` (snake_case, as it appears in the Python Protocol API).

[Update] Yeah, after digging into the error from that ticket, I'm pretty sure it's unrelated to this. I'm removing that ticket from this PR for now.

SyntaxColoring:
Unfortunately I don't think this fully resolves RSS-417. Try this:

```python
requirements = {
    # "apiLevel": "2.15",  # Works
    "apiLevel": "2.16",  # Does not work
    "robotType": "OT-2",
}


def run(protocol):
    # Note: ProtocolContext.fixed_trash, not InstrumentContext.trash_container.
    # 
    # This raises NoTrashDefinedError.
    protocol.fixed_trash
```

I caught this by running this against the tests in #14142. I'm going to pull those into this PR.

SyntaxColoring (changes requested):
Unfortunately I don't think this fully resolves RSS-417. Try this:

```python
requirements = {
    # "apiLevel": "2.15",  # Works
    "apiLevel": "2.16",  # Does not work
    "robotType": "OT-2",
}


def run(protocol):
    # Note: ProtocolContext.fixed_trash, not InstrumentContext.trash_container.
    # 
    # This raises NoTrashDefinedError.
    protocol.fixed_trash
```

I caught this by running this against the tests in #14142. I'm going to pull those into this PR.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1024,17 +1046,33 @@ def deck(self) -> Deck:
...
+    def fixed_trash(self) -> Union[Labware, TrashBin]:
         """The trash fixed to slot 12 of the robot deck.
 
-        It has one well and should be accessed like labware in your protocol.
+        In API Versions prior to 2.16 it has one well and should be accessed like labware in your protocol.
         e.g. ``protocol.fixed_trash['A1']``
+
+        In API Version 2.16 and above it returns a Trash fixture for OT-2 Protocols.
         """
+        if self._api_version >= APIVersion(2, 16):
```
CaseyBatten, line 1057: This newly added section to should provide handling for the issue Max identified and has tests implemented for. The following behavior is to be expected as a result:
v2.16 OT2 Protocol will identify protocol.fixed_trash as a TrashBin object
v2.16 Flex Protocol will raise an error on protocol.fixed_trash, notifying that for the Flex fixed trash does not exist on API version 2.16 and above.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -138,7 +141,26 @@ def __init__(
...
         }
         self._bundled_data: Dict[str, bytes] = bundled_data or {}
+
+        # With the addition of Moveable Trashes and Waste Chute support, it is not necessary
+        # to ensure that the list of "disposal locations", essentially the list of trashes,
+        # is initialized correctly on protocols utilizing former API versions prior to 2.16
+        # and also to ensure that any protocols after 2.16 intialize a Fixed Trash for OT-2
+        # protocols so that no load trash bin behavior is required within the protocol itself.
+        # Protocols prior to 2.16 expect the Fixed Trash to exist as a Labware object, while
+        # protocols after 2.16 expect trash to exist as either a TrashBin or WasteChute object.
```
SyntaxColoring, line 151: Can this comment be broken up, or shortened, for readability? I'm having trouble parsing it.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -138,7 +141,26 @@ def __init__(
...
+
         self._load_fixed_trash()
```
SyntaxColoring, line 153: It looks like some trash initialization code already existed in `self._load_fixed_trash()`. Organizationally, it seems like we should be adding our new code there.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -138,7 +141,26 @@ def __init__(
...
+        if should_load_fixed_trash_for_python_protocol(self._api_version):
+            self._core.append_disposal_location(self.fixed_trash)
+        elif (
+            self._api_version >= APIVersion(2, 16)
+            and self._core.robot_type == "OT-2 Standard"
+        ):
+            _fixed_trash_trashbin = TrashBin(
+                location=DeckSlotName.FIXED_TRASH, addressable_area_name="fixedTrash"
+            )
+            self._core.append_disposal_location(_fixed_trash_trashbin)
```
SyntaxColoring, line 163: I think this is showing us that we need to rename or rework `should_load_fixed_trash_for_python_protocol()`.

I'm not super familiar with that code, but I'm inferring from this usage that it currently works like "should we load a fixed trash *labware*," not "should we load *any form of fixed trash at all.*"

So:

* We call `should_load_fixed_trash_for_python_protocol()`.
    * If it returns yes: Load a fixed trash as a `Labware`.
    * If it returns no:
        *  Do our own check for whether we should load a fixed trash.
            * If yes: Load a fixed trash as a `TrashBin`.
                * **This is what's confusing to me.** It reads like we just asked `should_load_fixed_trash_for_python_protocol()` a question, then ignored its answer and did things our own way anyway.

@jbleon95 Any opinions?

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -138,7 +141,26 @@ def __init__(
...
+        # and also to ensure that any protocols after 2.16 intialize a Fixed Trash for OT-2
+        # protocols so that no load trash bin behavior is required within the protocol itself.
+        # Protocols prior to 2.16 expect the Fixed Trash to exist as a Labware object, while
+        # protocols after 2.16 expect trash to exist as either a TrashBin or WasteChute object.
+
         self._load_fixed_trash()
+        if should_load_fixed_trash_for_python_protocol(self._api_version):
+            self._core.append_disposal_location(self.fixed_trash)
+        elif (
+            self._api_version >= APIVersion(2, 16)
```
SyntaxColoring, line 157: As far as I can tell, this `self._api_version >= APIVersion(2, 16)` check is not doing anything. We can only reach here if `self._api_version >= APIVersion(2, 16)`, because there's a similar version check in the call to `should_load_fixed_trash_for_python_protocol()`, above.

I see where this is coming from, but I think we should leave this `_api_version` check out. I'm worried about accidentally leaving a gap between `should_load_fixed_trash_for_python_protocol()`'s maximum `APIVersion`, which defines the latest version where the fixed trash is a `Labware`, and this `APIVersion`, which defines the earliest version where the fixed trash is a `TrashBin`. If there's a gap, we'd be left with versions where OT-2 protocols accidentally have no fixed trash.

I would just write this as `elif self._core.robot_type == "OT-2 Standard"`.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1024,17 +1046,33 @@ def deck(self) -> Deck:
...
+        In API Versions prior to 2.16 it has one well and should be accessed like labware in your protocol.
         e.g. ``protocol.fixed_trash['A1']``
+
+        In API Version 2.16 and above it returns a Trash fixture for OT-2 Protocols.
         """
+        if self._api_version >= APIVersion(2, 16):
+            if self._core.robot_type == "OT-3 Standard":
+                raise APIVersionError(
+                    "Fixed Trash is not supported on Flex protocols in API Version 2.16 and above."
+                )
```
SyntaxColoring, line 1061: Let's raise `NoTrashDefinedError` instead of `APIVersionError`. Whether `protocol.fixed_trash` is available is a question of more than just the `apiLevel`; it has to do with the robot type, too.

```suggestion
                raise NoTrashDefinedError(
                    "Fixed Trash is not supported on Flex protocols in API Version 2.16 and above."
                )
```

And we'll have to update the `except` that's catching it, too.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1024,17 +1046,33 @@ def deck(self) -> Deck:
...
+            if isinstance(disposal_locations[0], TrashBin):
+                return disposal_locations[0]
+
         fixed_trash = self._core_map.get(self._core.fixed_trash)
         if fixed_trash is None:
             raise NoTrashDefinedError(
                 "No trash container has been defined in this protocol."
             )
+
```
SyntaxColoring, line 1075: This is a bit confusing to me because it looks like we now have two competing ways of keeping track of the fixed trash: `AbstractProtocol.fixed_trash`, and `AbstractProtocol.get_disposal_locations()`.

`AbstractProtocol.append_disposal_location()`/`.get_disposal_locations()` were really only added specifically to keep track of the user's explicit calls to `ProtocolContext.load_waste_chute()` and `ProtocolContext.load_trash_bin()`. They *could* also keep track of the fixed trash, like you have here, but if we do that, we need to do it consistently.

In other words, I think we need to choose between either of these two strategies:

1. `AbstractProtocol.get_disposal_locations()` returns the fixed trash *or* the user's explicitly-loaded trashes.
    * Then, we would delete `self._core.fixed_trash` from our codebase.
2. `AbstractProtocol.get_disposal_locations()` returns only the user's explicitly-loaded trashes. `AbstractProtocol.fixed_trash` returns only the Opentrons standard fixed trash, if there is one.
    * Then, when a user accesses `ProtocolContext.fixed_trash`, we would not do `self._core.get_disposal_locations()[0]`. Instead, we would do `self._core.fixed_trash`.

CaseyBatten (commented):
Changes integrated to cover requested issue regarding protocol context fixed trash references.

SyntaxColoring (approved):
This looks like it'll work and I'm satisfied with our test coverage now. Thanks for seeing this through!

Here are some code organizational suggestions. ~~I'd like us to resolve the `NoTrashDefinedError` vs. `APIVersionError` thing before merging, because it seems easy enough. We can leave the harder ones to later, if we need to move on to other things.~~ [EDIT: Ehh we can leave it all to later if we need to.]

PR #14146, opened 2023-12-07
docs(api): update Deck Slots page for new features in Python API 2.16

ecormany:
# Overview

Significantly expands the Deck Slots page of the Python docs to include the concept of deck configuration introduced in PAPI 2.16 / robot software 7.1.0.

# Test Plan

Check out the [sandbox](http://sandbox.docs.opentrons.com/docs-deck-slots-2.16/v2/deck_slots.html).

# Changelog

- Update the deck map SVG to show installed staging slots on Flex.
- Add info on working area vs. staging area
- New, large section on deck configuration
  - what it is
  - subsections for each type of fixture
  - how the app handles protocol–config conflicts

# Review requests

- Will this make sense to users who are approaching deck configuration for the first time?
- Will this page still work for OT-2 users who don't care about deck configuration at all?
- Do all my code snippets work? ⭐ 

# Risk assessment

v low, docs


api/docs/v2/deck_slots.rst
```diff
@@ -1,24 +1,19 @@
...
 Physical Deck Labels
 ====================
 
-The Opentrons Flex uses a coordinate labeling system for slots A1 (back left) through D3 (front right).
+Flex uses a coordinate labeling system for slots A1 (back left) through D4 (front right). Columns 1 through 3 are in the *working area* and are accessible by pipettes and the gripper. Column 4 is in the *staging area* and is only accessible by the gripper. For more information on staging area slots, see :ref:`deck-configuration` below.
 
-The Opentrons OT-2 uses a numeric labeling system for slots 1 (front left) through 11 (back center). The back right slot is occupied by the fixed trash.
+OT-2 uses a numeric labeling system for slots 1 (front left) through 11 (back center). The back right slot is occupied by the fixed trash.
 
 .. image:: ../img/Flex-and-OT-2-decks.svg
```
jwwojak: I would recommend two separate deck images instead of the single image that appears at the end of this section (Physical Deck Labels).

- Flex coordinate system paragraph followed by a Flex deck map image.
- OT-2 paragraph followed by the OT-2 deck map image.

Or, if getting 2 separate images is a burden, consider a caption below the image to make it absolutely clear which map belongs to which robot. e.g., "Fig 1. Flex deck map and OT-2 deck map" (or similar).

flexDeckmap1x

Deck Slots - OT2 map

api/docs/v2/deck_slots.rst
```diff
@@ -1,24 +1,19 @@
...
-
-
 .. _deck-slots:
 
 **********
 Deck Slots
 **********
 
-When you load an item onto the robot's deck, like with :py:obj:`ProtocolContext.load_labware()` or :py:obj:`ProtocolContext.load_module()`, you need to specify which slot to put it in. The API accepts values that correspond to the physical deck slot labels on an OT-2 or Flex robot.
+Deck slots are where you place hardware items on the deck surface of your Opentrons robot. In the API, you load the corresponding items into your protocol with methods like :py:obj:`.ProtocolContext.load_labware`, :py:obj:`.ProtocolContext.load_module`, or :py:obj:`.ProtocolContext.load_trash_bin`. When you call these methods, you need to specify which slot to load the item in.
```
jwwojak, line 9: Repetition: "deck" Can the 2nd instance be omitted?

- "Deck slots are where you place hardware items on your Opentrons robot."

I'd like to say, "Deck slots are where you place hardware items on the working area of your Opentrons robot." but see that the next section introduces the working area and staging area concepts together. In any case, the sentence is fine. Just twice in one sentence, something to consider.

api/docs/v2/deck_slots.rst
```diff
@@ -83,3 +78,113 @@ The correspondence between deck labels is based on the relative locations of the
...
 .. TODO staging slots and error handling of A4–D4 in OT-2 protocols
+
+Slots A4, B4, C4, and D4 on Flex have no equivalent on OT-2. 
+
+.. _deck-configuration:
+
+Deck Configuration
+==================
+
+A Flex running robot system version 7.1.0 or higher lets you specify its deck configuration on the touchscreen or in the Opentrons App. This tells the robot the positions of unpowered *deck fixtures*: items that replace standard deck slots. The following table lists currently supported deck fixtures and their allowed deck locations.
```
jwwojak, line 92: The colon here makes the sentence abrupt. What about something like this:

"This tells the robot the positions of unpowered _deck fixtures_, which are items that replace standard deck slots).

Or even:

"This tells the robot the positions of unpowered _deck fixtures_ (items that replace standard deck slots).

Or 2 sentences:

"This tells the robot the positions of unpowered _deck fixtures_. These are items that replace standard deck slots.

api/docs/v2/deck_slots.rst
```diff
@@ -83,3 +78,113 @@ The correspondence between deck labels is based on the relative locations of the
...
+    * - Fixture
+      - Slots
+    * - Staging area slots
+      - A3–D3
+    * - Trash bin
+      - A1–D1, A3-D3
+    * - Waste chute
+      - D3
+      
+Which fixtures you need to configure depend on both load methods and the effects of other methods called in your protocol. The following sections explain how to configure each type of fixture.
```
jwwojak, line 106: "both" isn't really needed because of "and". This says the same thing:

"Which fixtures you need to configure depend on the load methods and the effects of other methods called in your protocol."

You could even put _and_ in italics for emphasis that both are required.

api/docs/v2/deck_slots.rst
```diff
@@ -83,3 +78,113 @@ The correspondence between deck labels is based on the relative locations of the
...
+
+.. _configure-staging-area-slots:
+
+Staging Area Slots
+------------------
+
+Slots A4 through D4, which are not accessible by pipettes, are always available in the API for loading and moving labware. Using a slot in column 4 as the ``location`` argument of :py:meth:`.load_labware` or the ``new_location`` argument of :py:meth:`.move_labware` will require the corresponding staging area slot in the robot's deck configuration::
```
jwwojak: Perhaps the clause isn't really needed here. Instead, consider starting the sentence like this:

- "Slots A4 through D4 are not accessible by pipettes, but they're always available in the API for loading and moving labware."

Or, let's define the slots and then move on to say what they can and cannot do:

- "Slots A4 through D4 are staging area slots. Pipettes cannot reach into the staging area, but these slots are always available in the API for loading and moving labware."

You could even add "with the gripper" at the end of that sentence.

api/docs/v2/deck_slots.rst
```diff
@@ -83,3 +78,113 @@ The correspondence between deck labels is based on the relative locations of the
...
+    The :py:class:`TrashBin` class doesn't have any callable methods, so you don't have to save the result of ``load_trash_bin()`` to a variable, especially if your protocol only loads a single trash container. Being able to reference the trash bin by name is useful when dealing with multiple trash containers.
+
+Call ``load_trash_bin()`` multiple times to add more than one bin. See :ref:`Adding a Trash Container <pipette-trash-container>` for more information on using pipettes with multiple trash bins.
+
+.. _configure-waste-chute:
+
+Waste Chute
+-----------
+
+The waste chute accepts various waste materials from Flex pipettes or the Flex Gripper. Pipettes can dispense liquid or drop tips in the chute. The gripper can drop tip racks and other labware into the chute.
```
jwwojak: As with "deck" above, 2 instances of "waste." I understand this may be deliberate, but also suggest it is repetitive and makes the sentence awkward. Perhaps take something from the quick-start intro/summary?

"The waste chute is a deck fixture for the Flex. It accepts various waste materials dropped into it from Flex pipettes or the Flex Gripper.

jwwojak (commented):
Left a few optional comments for consideration.

api/docs/v2/deck_slots.rst
ecormany: Yep, the only reason to have these in a single image was to keep them to scale, which isn't mentioned and was probably only known to me. Will revisit.

api/docs/v2/deck_slots.rst
ecormany: Repetition here to avoid ambiguity. You could say that you "place" instruments on their mounts on the gantry, and they are "hardware". And future hardware could be placed in even more…places.

api/docs/v2/deck_slots.rst
ecormany: If we need the emphasis, I'd prefer "both…and" over relying on formatting alone. Perfectly cromulent English and accessible to screen readers.

api/docs/v2/deck_slots.rst
ecormany: Deliberately left out "with the gripper", because you can move labware to/from staging manually.

api/docs/v2/deck_slots.rst
```diff
@@ -83,3 +78,113 @@ The correspondence between deck labels is based on the relative locations of the
...
+In total, there are four possible deck configurations for the waste chute.
+    - Waste chute only
+    - Waste chute with cover
+    - Waste chute with staging area slot
+    - Waste chute with staging area slot and cover
+
+Deck Conflicts
+==============
+
+When a Flex analyzes a Python protocol, it will check whether there are any conflicts with its current deck configuration. You can't start the protocol run until any conflicts are resolved. You can resolve them one of two ways:
```
SyntaxColoring: I don't think this is a super-meaningful distinction for most users, but for the sake of technical precision: the *app* is the thing guarding against conflicts, not the on-robot analysis. So we could rephrase this, for example:

```suggestion
When you prepare a protocol with the Opentrons App or with Flex's touchscreen, it will check whether the protocol has any conflicts with the robot's current deck configuration. You can't start the protocol run until any conflicts are resolved. You can resolve them one of two ways:
```

api/docs/v2/deck_slots.rst
```diff
@@ -83,3 +78,113 @@ The correspondence between deck labels is based on the relative locations of the
...
+Call ``load_trash_bin()`` multiple times to add more than one bin. See :ref:`Adding a Trash Container <pipette-trash-container>` for more information on using pipettes with multiple trash bins.
+
+.. _configure-waste-chute:
+
+Waste Chute
+-----------
+
+The waste chute accepts various waste materials from Flex pipettes or the Flex Gripper. Pipettes can dispense liquid or drop tips in the chute. The gripper can drop tip racks and other labware into the chute.
+
+To use the waste chute, first use :py:meth:`.load_waste_chute` to load it in slot D3::
```
SyntaxColoring, line 164: Do we want cross-references from `load_waste_chute()` and `load_trash_bin()` to this higher-level page?

SyntaxColoring (approved):
Huzzah.

PR #14136, opened 2023-12-07
fix(api): remove fixed trash restriction from PAPI deck conflict check

jbleon95:
# Overview

Fixes RQA-2015

This PR removes the restriction from PAPI based deck conflict checking that prevented labware and modules from being loaded into slot 12/A3. With deck configuration, a trash does not necessarily need to be in that slot, and addressable area engine logic will prevent a trash and the deck slot it exists in from being used in the same protocol.

This change is safe to make for older protocols as well, as the newly added engine based conflict checking will prevent labware or modules from being loaded where the fixed trash labware has been automatically loaded.

# Test Plan

Confirmed that the first protocol passes analysis and the second one correctly fails.

```
metadata = {
    'protocolName': 'Load on A3 Test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}


def run(context):
    tiprack = context.load_labware('opentrons_flex_96_tiprack_50ul', 'A3')
    
    waste = context.load_waste_chute()
    pipette = context.load_instrument('flex_1channel_1000', 'left', tip_racks=[tiprack])

    pipette.pick_up_tip()
    pipette.drop_tip()
```

```
metadata = {
    'protocolName': 'Load on A3 Test 2.15',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.15"
}


def run(context):
    # This line will raise an error since fixed trash labware has been automatically loaded here
    tiprack = context.load_labware('opentrons_flex_96_tiprack_50ul', 'A3')
```

# Changelog

- Removed fixed trash restriction from PAPI deck conflict check

# Review requests

# Risk assessment

Low.

PR #14151, opened 2023-12-08
test(api): Add integration tests for load conflicts with the fixed trash

SyntaxColoring:
# Overview

Add tests to cover RQA-2015 / PR #14136.

# Test Plan

On a local branch, I've reverted PR #14136 and confirmed that these tests catch RQA-2015.

# Changelog

Test that you can generally load things into slot 12 when, and only when, your protocol will not have a fixed trash.

# Review requests

Take a look at the test parametrizations and make sure they make sense. The OT-2+APIv2.16 case in particular is a bit weird, arguably an architectural bug. I haven't ticketed it, but I could if you think we should actually plan to fix it.

# Risk assessment

No risk.

### DIFF
--- a/api/tests/opentrons/protocol_api_integration/test_trashes.py
+++ b/api/tests/opentrons/protocol_api_integration/test_trashes.py
@@ -3,7 +3,8 @@
 
 from opentrons import protocol_api, simulate
 
-from typing import Optional, Type
+import contextlib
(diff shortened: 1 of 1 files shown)


CaseyBatten (approved):
Looks good, appropriately tests our now hard casing of "Flex 2.15 Protocols always have a fixed trash loaded into A3".

PR #14153, opened 2023-12-08
fix(app): moveLabware command text support for waste chute

jerader:
closes RQA-2037

# Overview

Add waste chute support in run log for moving labware into the waste chute

# Test Plan

Test with the protocol attached to the ticket! Look at the run log and see that the command text for moving labware into waste chute is correct

# Changelog

- expand logic in `moveLabwareCommandText` and add test coverage

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/organisms/CommandText/MoveLabwareCommandText.tsx
+++ b/app/src/organisms/CommandText/MoveLabwareCommandText.tsx
@@ -1,14 +1,13 @@
 import { useTranslation } from 'react-i18next'
-import {
-  CompletedProtocolAnalysis,
-  MoveLabwareRunTimeCommand,
-  RobotType,
-  GRIPPER_WASTE_CHUTE_ADDRESSABLE_AREA,
-} from '@opentrons/shared-data'
+import { GRIPPER_WASTE_CHUTE_ADDRESSABLE_AREA } from '@opentrons/shared-data'
 import { getLabwareName } from './utils'
 import { getLabwareDisplayLocation } from './utils/getLabwareDisplayLocation'
 import { getFinalLabwareLocation } from './utils/getFinalLabwareLocation'
-
+import type {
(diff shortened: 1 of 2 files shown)


app/src/organisms/CommandText/MoveLabwareCommandText.tsx
```diff
@@ -1,9 +1,10 @@
 import { useTranslation } from 'react-i18next'
-import type {
+import {
   CompletedProtocolAnalysis,
   MoveLabwareRunTimeCommand,
   RobotType,
-} from '@opentrons/shared-data/'
+  GRIPPER_WASTE_CHUTE_ADDRESSABLE_AREA,
+} from '@opentrons/shared-data'
```
koji: I think these are types so probably getting back `type` and moving them down would be better.
Then import `GRIPPER_WASTE_CHUTE_ADDRESSABLE_AREA` separately.

app/src/organisms/CommandText/__tests__/CommandText.test.tsx
```diff
@@ -1,28 +1,29 @@
...
   OT2_ROBOT_TYPE,
-  PrepareToAspirateRunTimeCommand,
+  GRIPPER_WASTE_CHUTE_ADDRESSABLE_AREA,
 } from '@opentrons/shared-data'
 import { i18n } from '../../../i18n'
 import { CommandText } from '../'
 import { mockRobotSideAnalysis } from '../__fixtures__'
 
 import type {
+  AspirateInPlaceRunTimeCommand,
```
koji, line 13: 👍

koji (approved):
Unfortunately, I cannot run a protocol due to robot-server's limitations, but the changes make sense to me.
The comment is a very small thing.

PR #14154, opened 2023-12-08
fix(app): Remove probe check from module calibration

sfoster1:
Module calibration doesn't require a pipette id and thus none is available here (and in general I believe you can do it with a gripper) so this isn't really an appropriate check.

At some point we can readd it but we have to actually load a pipette in the maintenance run.


ahiuchingau (approved):
I didn't realize it was done with a gripper...

PR #14155, opened 2023-12-08
refactor(step-generation): update nozzles in robotState not from initialRobotState

jerader:
# Overview

There was a PD bug where if you have multiple column pick up steps followed by a full tip pick up step, it'd error. This is for 96-channel! But this PR fixes it and cleans up the reset command

# Test Plan

Test by adding 2 column pick up transfer steps followed by a Full tip pick up step. See that it doesn't error. Check tip state and see that it updates correctly. Download the protocol and see that there are 2 `configureNozzleLayout` commands for column and then for All

# Changelog

- remove setting nozzles from the initial robot state and set it through `configureNozzleLayout` state update.
- remove prevNozzles since it isn't needed anymore
- clean up the usage the `configureNozzleLayout` reset command, it isn't needed since the robot will reset the nozzles after each run

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/commandCreators/compound/consolidate.ts
+++ b/step-generation/src/commandCreators/compound/consolidate.ts
@@ -466,11 +466,10 @@ export const consolidate: CommandCreator<ConsolidateArgs> = (
       const dropTipAfterDispenseAirGap =
         airGapAfterDispenseCommands.length > 0 ? dropTipCommand : []
 
-      const nozzles = prevRobotState.pipettes[args.pipette].nozzles
-      const prevNozzles = prevRobotState.pipettes[args.pipette].prevNozzles
+      const stateNozzles = prevRobotState.pipettes[args.pipette].nozzles
       const configureNozzleLayoutCommand: CurriedCommandCreator[] =
         //  only emit the command if previous nozzle state is different
-        is96Channel && args.nozzles != null && nozzles !== prevNozzles
+        is96Channel && args.nozzles != null && args.nozzles !== stateNozzles
           ? [
               curryCommandCreator(configureNozzleLayout, {
                 nozzles: args.nozzles,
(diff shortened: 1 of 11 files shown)


step-generation/src/commandCreators/compound/consolidate.ts
```diff
@@ -466,22 +465,17 @@ export const consolidate: CommandCreator<ConsolidateArgs> = (
...
               curryCommandCreator(configureNozzleLayout, {
                 nozzles: args.nozzles,
                 pipetteId: args.pipette,
               }),
             ]
           : []
-      const configureNozzleLayoutCommandReset = getConfigureNozzleLayoutCommandReset(
-        args.pipette,
-        prevNozzles
-      )
```
jerader, line 484: this isn't actually needed since the robot will reset the nozzles to "All" at the end of each run. So i removed it to clean it up. Plus it should go in `generateRobotStateTimeline` instead since that's where the final drop tip commands get emitted

step-generation/src/commandCreators/compound/transfer.ts
```diff
@@ -275,22 +274,19 @@ export const transfer: CommandCreator<TransferArgs> = (
             changeTipNow =
               isInitialSubtransfer || destinationWell !== prevDestWell
           }
-          const nozzles = prevRobotState.pipettes[args.pipette].nozzles
-          const prevNozzles = prevRobotState.pipettes[args.pipette].prevNozzles
+          const stateNozzles = prevRobotState.pipettes[args.pipette].nozzles
+          console.log('args.nozzles', args.nozzles)
+          console.log('state nozzles', stateNozzles)
```
shlokamin: leftover console logs

shlokamin (approved):
much cleaner now! 😄 🧹

PR #14156, opened 2023-12-08
fix(api): clean up tip motor distance caching/usage

fsinapi:
# Overview

[RQA-2038](https://opentrons.atlassian.net/browse/RQA-2038) seems to be related to the Tip Check that occurs when cancelling a protocol. The Serialization error was a result of the software attempting to send a move group with a duration far exceeding a 32-bit integer. I was able to replicate the issue by cancelling a run during the setup stage, and then re-entering setup at a time such that a Tip Motor home action got interrupted, leaving the tip motors with a "motor_position" estimate that is orders of magnitude higher than the actual max position of the axis. 

Once the tip motor is in this situation, the software will fail a subsequent Tip Presence Check because of the invalid cached position of the motor.

To fix this, I added a few bandaids
* Wrap tip presence checks in the motion lock
* If a tip action fails for any reason, clear the cached tip motor position
* (Most importantly) if the cached tip motor position is higher than the physical limit of the axis, force the software to run a slow home.

With all of these changes, I can no longer replicate the issue.

# Test Plan

Replication is mentioned above. 

# Changelog

Listed above

# Review requests

# Risk assessment

Medium/low, doesn't touch anything too integral

[RQA-2038]: https://opentrons.atlassian.net/browse/RQA-2038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -735,13 +735,19 @@ class OT3Controller:
             if not self._feature_flags.stall_detection_enabled
             else False,
         )
-        positions = await runner.run(can_messenger=self._messenger)
-        if NodeId.pipette_left in positions:
-            self._gear_motor_position = {
-                NodeId.pipette_left: positions[NodeId.pipette_left].motor_position
-            }
-        else:
-            log.debug("no position returned from NodeId.pipette_left")
+        try:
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Ah, yeah, the downside of doing home the way we do it

PR #14157, opened 2023-12-08
feat(hardware): add can message to adjust gripper jaw holdoff ms value

ahiuchingau:
# Overview
This PR adds a new can message so that we can adjust the gripper jaw idle holdoff value from python. Requires fw changes in https://github.com/Opentrons/ot3-firmware/pull/735

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (changes requested):
can we make this value in ms? i know there's other places where we use ticks for this but i'd prefer not to add another

sfoster1 (approved):
Looks good to me!

PR #14160, opened 2023-12-11
fix(components): provide OT-2 fixed trash dimensions for deck location select

brenthagen:
# Overview

hard codes the OT-2 fixed trash dimensions for the sole purpose of legacy deck slot location select, used in drop tip wizard. needed because the OT-2 v4 deck definition no longer contains x and y bounding box dimensions

closes RQA-2042

# Test Plan

 - visually verified the dimensions of the drop tip wizard OT-2 fixed trash selection area

# Changelog

 - Provides OT-2 fixed trash dimensions for deck location select

# Review requests

check OT-2 drop tip wizard

# Risk assessment

low

### DIFF
--- a/components/src/hardware-sim/DeckSlotLocation/index.tsx
+++ b/components/src/hardware-sim/DeckSlotLocation/index.tsx
@@ -17,6 +17,10 @@ interface LegacyDeckSlotLocationProps extends React.SVGProps<SVGGElement> {
   slotClipColor?: React.SVGProps<SVGPathElement>['stroke']
 }
 
+// dimensions of the OT-2 fixed trash, not in deck definition
+export const OT2_FIXED_TRASH_X_DIMENSION = 172.86
+export const OT2_FIXED_TRASH_Y_DIMENSION = 165.86
+
 /**
  * This is a legacy component for rendering an OT-2 deck slot by reference to the V4 deck definition
  */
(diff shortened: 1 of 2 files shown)


components/src/hardware-sim/DeckSlotLocation/index.tsx
```diff
@@ -48,17 +52,23 @@ export function LegacyDeckSlotLocation(
...
+  const [xOffset, yOffset] = slotDef.offsetFromCutoutFixture
+
+  // adjust the fixed trash position and dimension to fit inside deck SVG
   const fixedTrashPositionAdjustment = isFixedTrash ? 7 : 0
   const fixedTrashDimensionAdjustment = isFixedTrash ? -9 : 0
 
-  const adjustedXPosition = xPosition + fixedTrashPositionAdjustment
-  const adjustedYPosition = yPosition + fixedTrashPositionAdjustment
+  const adjustedXPosition = xPosition + fixedTrashPositionAdjustment - xOffset
+  const adjustedYPosition = yPosition + fixedTrashPositionAdjustment - yOffset
```
SyntaxColoring, line 71: Is this deliberately undoing the `+ offsetFromCutoutFixture` that happens internally within `getPositionFromSlotId()`? It seems like that only works for the deck slots because their `offsetFromCutoutFixture` happens to be (0, 0, 0).

SyntaxColoring (approved):
Makes sense to me given the discussion in RQA-2042. Thank you!

components/src/hardware-sim/DeckSlotLocation/index.tsx
brenthagen: yeah, to undo the offset for the fixed trash. could put this in the fixed trash adjustment magic number instead. do we expect `offsetFromCutoutFixture` values in the deck definition to change?

brenthagen: for OT-2 - this component is OT-2 only

SyntaxColoring: > do we expect offsetFromCutoutFixture values in the deck definition to change?

Maybe, yeah. With the current values, one of the things that keeps confusing us is that the origin point of a cutout is where a deck slot would be if one of those fixtures were mounted to that cutout. Instead of something intrinsic to the cutout itself.

There are no plans to change this imminently, though.

PR #14161, opened 2023-12-11
refactor(protocol-designer): misc staging area slot bugs addressing

jerader:
closes RQA-2050 RQA-2051 RQA-2057

# Overview

Addresses several staging area related bugs. 

1. the staging area modal deck configurator is not too big
2. the waste chute with staging area doesn't render when there is no waste chute
3. a labware that starts off in staging area slot now renders in the dropdown menu text

# Test Plan

Create a flex protocol and add a staging area slot into slot D3. Click on editing the staging area slot and see that the deck configurator isn't too big. Do not add a waste chute.

Go to the design tab with the deck setup and there should not be waste chute rendered with the staging area. 

Move the tiprack that was autogenerated to the staging area slot. Create a move labware step and move the tiprack to another slot. See that the dropdown has "in staging area slot" text at the end. Move the labware and then create a new move labware step and see that the dropdown does not have "in staging area slot" text at the end.

# Changelog

- add more logic for when the waste chute and staging area fixture renders
- adjust sizing in staging area modal
- create util for if the labware is in column 4 for the dropdown menu

# Review requests

see test plan

# Risk assessment

low


koji (approved):

one quick question on Labware display in Design tab.
For staging area, `labware in staging area`.
For offDeck, `OffDeck - labware`

Don't we need to use the same format?

jerader:
> the sandbox worked as expected.
> 
> one quick question on Labware display in Design tab. For staging area, `labware in staging area`. For offDeck, `OffDeck - labware`
> 
> Don't we need to use the same format?

Good call! Thanks @koji!

PR #14170, opened 2023-12-12
fix(app-shell,usb-bridge): improve flex usb communication

sfoster1:
Closes RSS-420, RQA-2053, RQA-1615

This branch has fixes for a couple of the underlying issues with flex usb communication.

First, we need to limit the flex->host directional bandwidth (which we do by limiting the size of the USB packets we send on the python side) because if we send too much we will blow through the host-side serial buffers in their drivers, which are very variable. When we do that, we lose data which is part of an HTTP response; and the node http parser can't really handle that and just hangs. Even if it didn't and found the error, we need that error not to happen.

Second, we need to fix a couple socket lifetime issues on the node side. These were basically
1. By hook or by crook, close sockets less
2. When you close sockets, do it correctly so you can reopen them

(1) has some gross results because the code that uses sockets is really gross and we have not gone high when they went low, but it's a minimal changeset.
(2) is a real pain because the control flow bounces around 3 different events on like 3 different emitters and you need to trace it through; what it boils down to is noting that both the readable and writable streams AND the serialport have a "close" event they fire, and that only means the serial port is _actually closed_ if it was the serial port that did it. 

This doesn't get us all the way there almost certainly; we do occasionally still drop parts of packets on heavy usage like fetching logs sometimes, and therefore probably need to introduce more timeouts. We also haven't fully tested all the OS's and flows yet but we do get a whole lot farther with that testing than we previously did.

### Testing more
At least the following:
- Getting logs should succeed on all platforms
- Updating robots should succeed on all platforms
- Start a protocol and ideally run it on all platforms
- Do something that the robot-server uses a redirect in on all platforms (attaching a pipette is a good example)
- Other hot button flows but those above should give a broad test.

### DIFF
--- a/app-shell/src/robot-update/index.ts
+++ b/app-shell/src/robot-update/index.ts
@@ -180,13 +180,6 @@ export function getRobotSystemUpdateUrls(
     .then(manifest => {
       const urls = getReleaseSet(manifest, CURRENT_VERSION)
 
-      if (urls === null) {
-        log.warn('No release files in manifest', {
-          version: CURRENT_VERSION,
-          manifest,
-        })
-      }
-
       return urls
     })
     .catch((error: Error) => {
--- a/app-shell/src/usb.ts
+++ b/app-shell/src/usb.ts
@@ -1,5 +1,5 @@
 import { ipcMain, IpcMainInvokeEvent } from 'electron'
-import axios, { AxiosRequestConfig, AxiosError } from 'axios'
+import axios, { AxiosRequestConfig } from 'axios'
 import FormData from 'form-data'
 import fs from 'fs'
 import path from 'path'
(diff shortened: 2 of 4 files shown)


mjhuff (approved):
LGTM with CI!

Added a few tickets to the PR and wrote a follow-up ticket.

fsinapi (approved):
Oh wow nice! ot3usb changes look good to me!

SyntaxColoring:
> First, we need to limit the flex->host directional bandwidth (which we do by limiting the size of the USB packets we send on the python side) because if we send too much we will blow through the host-side serial buffers in their drivers, which are very variable.

Does the host not have a proper flow control mechanism to apply backpressure to the Flex? Some mechanism *must* exist at the USB protocol level, right? Are we accidentally subverting it because we're layering a virtual serial port (?) atop raw USB?

Solving this by metering bandwidth seems dicey. Like, it will depend on the load that the host device is under, right?

sfoster1:
> Does the host not have a proper flow control mechanism to apply backpressure to the Flex? Some mechanism _must_ exist at the USB protocol level, right? Are we accidentally subverting it because we're layering a virtual serial port (?) atop raw USB?

You know, that's a pretty good question, since flow control mechanisms do in fact exist for both actual rs-232 and usb cdc-acm, and we're just not using them. I think it's worth taking a pass at that.

app-shell/src/robot-update/index.ts
```diff
@@ -180,13 +180,6 @@ export function getRobotSystemUpdateUrls(
     .then(manifest => {
       const urls = getReleaseSet(manifest, CURRENT_VERSION)
 
-      if (urls === null) {
-        log.warn('No release files in manifest', {
-          version: CURRENT_VERSION,
-          manifest,
-        })
-      }
-
```
shlokamin, line 189: is this to cut down on noise? seems like a useful warning since we expect there to be urls in the release set

app-shell/src/usb.ts
```diff
@@ -34,19 +34,17 @@ let usbFetchInterval: NodeJS.Timeout
...
 }
-
 export function createSerialPortHttpAgent(path: string): void {
   const serialPortHttpAgent = new SerialPortHttpAgent({
     maxFreeSockets: 1,
     maxSockets: 1,
     maxTotalSockets: 1,
     keepAlive: true,
-    keepAliveMsecs: 10000,
+    keepAliveMsecs: Infinity,
```
shlokamin, line 43: lol

app-shell/src/usb.ts
```diff
@@ -63,52 +61,50 @@ function isUsbDeviceOt3(device: UsbDevice): boolean {
...
     return {
+      error: false,
       data: response.data,
       status: response.status,
       statusText: response.statusText,
     }
   } catch (e) {
-    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
-    usbLog.debug(`usbListener error ${e?.message ?? 'unknown'}`)
+    console.log(`axios request error ${e?.message ?? 'unknown'}`)
```
shlokamin, line 107: was this changed cuz reading logs from the usb logger was annoying? cuz it seems like thats the right place to report logs here

shlokamin (approved):
thanks for doing this @sfoster1 and @mjhuff!

app-shell/src/robot-update/index.ts
sfoster1: yeah the noise is just brutal

app-shell/src/usb.ts
sfoster1: yeah how does one get logs out of the usb logger?

PR #14171, opened 2023-12-12
refactor(hardware-testing): Account for InstrumentContext.trash_container not being a Labware anymore

SyntaxColoring:
# Overview

Fixes these CI errors:

```
hardware_testing/gravimetric/trial.py:244: error: Item "TrashBin" of "Union[Labware, TrashBin, WasteChute]" has no attribute "wells"  [union-attr]
hardware_testing/gravimetric/trial.py:244: error: Item "WasteChute" of "Union[Labware, TrashBin, WasteChute]" has no attribute "wells"  [union-attr]
```

# Test Plan

I don't have an easy way to test this today (I'm sick and working from home) so I'm relying on code review.

# Changelog

* Add handling for the case where `InstrumentContext.trash_container` is a `WasteChute` or `TrashBin`, and so it doesn't have a "well A1."
* Delete some dead code.

# Review requests

See my inline notes.

# Risk assessment

Low.

### DIFF
--- a/hardware-testing/hardware_testing/gravimetric/trial.py
+++ b/hardware-testing/hardware_testing/gravimetric/trial.py
@@ -241,11 +241,18 @@ def _finish_test(
         resources.ctx.home()
         if resources.pipette.current_volume > 0:
             ui.print_info("dispensing liquid to trash")
-            trash = resources.pipette.trash_container.wells()[0]
+            trash_container = resources.pipette.trash_container
+            dispense_location = (
+                trash_container.wells()[0].top()
+                if isinstance(trash_container, Labware)
+                else trash_container
+            )
             # FIXME: this should be a blow_out() at max volume,
             #        but that is not available through PyAPI yet
             #        so instead just dispensing.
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Thank you!

hardware-testing/hardware_testing/gravimetric/trial.py
```diff
@@ -241,14 +241,18 @@ def _finish_test(
         resources.ctx.home()
         if resources.pipette.current_volume > 0:
             ui.print_info("dispensing liquid to trash")
-            trash = resources.pipette.trash_container.wells()[0]
-            dispense_location = trash.top()
-            if resources.pipette.channels == 96:
-                dispense_location = dispense_location.move(Point(-36.0, -25.5, 0))
```
SyntaxColoring, line 247: This `Point(-36.0, -25.5, 0)` offset was never being used in practice because the `dispense_location` variable was unused. I'm deleting it here to preserve behavior. Is that correct? Alternatively, I can fix it  while I'm here so that the offset does get used.

PR #14172, opened 2023-12-12
fix(app): prevent terminalrunbanner from closing the current run

mjhuff:
Closes RQA-2013

# Overview

Clicking the error banner within ProtcolRunHeader has the behavior of clearing the current run, but this behavior is probably not desired. First, it actually unrenders the error banner, and second, all conditional rendering tied to the current run (such as the drop tip banner) are unrendered as well. The "Run Again" button component is the only thing that really needs to clear/reset the current run, so let's just rely on that functionality. 

### 

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/9d240261-bc4c-4960-9a13-8b4a3c14fd7d

### With PR

https://github.com/Opentrons/opentrons/assets/64858653/a9406bda-5463-4ef4-9427-cf3ba8000dfa

# Test Plan
- Trigger any sort of error during a protocol run (on the desktop app -- estop is always fun), and then click the terminal error banner. Observe that the current run no longer gets reset.
- Clicking Run Again should properly reset the run.

# Changelog
- Clicking the error banner on the desktop app during a protocol run no longer unrenders the error banner and other banners. 

# Review requests
- Just double check my logic here - I don't think we actually lose anything by not clearing the protocol run here. I think the reason this bug was in the code for so long was that it wasn't really that noticeable of a problem until we added other conditional banners, such as the drop tip banner.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/ProtocolRunHeader.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/ProtocolRunHeader.tsx
@@ -785,11 +785,6 @@ function TerminalRunBanner(props: TerminalRunProps): JSX.Element | null {
   } = props
   const { t } = useTranslation('run_details')
 
-  const handleClick = (): void => {
-    handleClearClick()
-    setShowRunFailedModal(true)
-  }
-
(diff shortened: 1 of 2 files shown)


koji (approved):
lgtm
thank you for fixing this!

PR #14173, opened 2023-12-12
feat(api): add tiprack adapter check for 96 ch tip pickup and return

sanni-t:
# Overview

Adds error raising if:
- you are using a 96 channel with all nozzles and are picking up from or returning tips to a tiprack NOT on an appropriate adapter
- you are using a 96 channel with partial nozzles and are picking up from or returning tips to a tiprack that IS on an adapter that is taller than the tiprack (like the Flex 96 tiprack adapter)

# Test Plan

- Test that analysis raises error when the above conditions are met in a protocol.
Sample protocol:
```


requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(protocol_context):
	trash_labware = protocol_context.load_labware("opentrons_1_trash_3200ml_fixed", "A3")

	tip_rack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C3")
	tip_rack2 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C3")

	tip_rack3 = protocol_context.load_labware("opentrons_flex_96_tiprack_200ul", "B3", adapter="opentrons_flex_96_tiprack_adapter")
	tip_rack4 = protocol_context.load_labware("opentrons_flex_96_tiprack_200ul", "D3", adapter="opentrons_flex_96_tiprack_adapter")


	instrument = protocol_context.load_instrument('flex_96channel_1000', mount="left", tip_racks=[tip_rack1, tip_rack2, tip_rack3, tip_rack4])

	# Should raise error because tiprack1 is not on adapter
	instrument.pick_up_tip()
	instrument.drop_tip()

	instrument.configure_nozzle_layout(style=COLUMN, start="A12")
	instrument.tip_racks = [tip_rack3]
	
	# Should pick tips up column-wise from tiprack3
	for _ in range(12):
		# Should raise error because tiprack3 is on adapter
		instrument.pick_up_tip()
		instrument.drop_tip()

	######### CHANGE CONFIG TO ALL #########
	instrument.configure_nozzle_layout(style=ALL, tip_racks=[tip_rack2])

	# Should raise error because tiprack2 is not on adapter
	instrument.pick_up_tip()
	instrument.drop_tip()
```

# Changelog

- added the above checks to `core/engine/deck_conflict.py`
- added tests for it. Also added other deck conflict checks that were missing. Found a bug in drop_tip deck conflict checking so fixed that

# Review requests

- mainly looking for opinions on the approach to the check. I am not checking if the tiprack is or isn't on the Flex 96 tiprack adapter, and instead checking if it is on an adapter that matches the height attribute of Flex tiprack (that it's taller than the tipracks). I decided on this approach after considering the number of labware definitions in use internally that are variants of the official Flex adapter definition. I can imagine that there will be customers that might write their own variants as well and the check should work for those too.

# Risk assessment

- Low-medium. Affects only 96 channel pickup & returns

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
+++ b/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
@@ -221,16 +221,16 @@ def check_safe_for_tip_pickup_and_return(
             )
         elif not is_partial_config and tiprack_height > adapter_height:
             raise PipetteMovementNotAllowed(
-                f"Tiprack {tiprack_name}  must be on a Flex Tiprack Adapter (or similar)"
-                f" in order to pick up or drop tips with full 96-channel nozzles."
+                f"Tiprack {tiprack_name} must be on a Flex Tiprack Adapter (or similar)"
+                f" in order to pick up or return tips with all 96-channel nozzles."
             )
 
     elif (
         not is_partial_config
     ):  # tiprack is not on adapter and pipette is in full config
         raise PipetteMovementNotAllowed(
-            f"Tiprack {tiprack_name}  must be on a Flex Tiprack Adapter (or similar)"
-            f" in order to pick up or drop tips with full 96-channel nozzles."
+            f"Tiprack {tiprack_name} must be on a Flex Tiprack Adapter (or similar)"
+            f" in order to pick up or return tips with all 96-channel nozzles."
         )
 
 
--- a/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
+++ b/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
@@ -36,6 +36,15 @@ class PartialTipMovementNotAllowedError(MotionPlanningFailureError):
         )
 
 
+class PipetteMovementNotAllowed(MotionPlanningFailureError):
+    """Error raised when trying to perform a pipette movement to a tiprack beased on adapter status."""
+
+    def __init__(self, message: str) -> None:
+        super().__init__(
+            message=message,
+        )
+
+
 _log = logging.getLogger(__name__)
 
 # TODO (spp, 2023-12-06): move this to a location like motion planning where we can
@@ -171,6 +180,57 @@ def check_safe_for_pipette_movement(
             labware_id=labware_id,
             well_name=well_name,
             well_location=well_location,
+        )
+
+
+def check_safe_for_tip_pickup_and_return(
+    engine_state: StateView,
+    pipette_id: str,
+    labware_id: str,
+) -> None:
+    """Check if the presence or absence of a tiprack adapter might cause any pipette movement issues.
+
+    A 96 channel pipette will pick up tips using cam action when it's configured
+    to use ALL nozzles. For this, the tiprack needs to be on the Flex 96 channel tiprack adapter
+    or similar or the tips will not be picked up.
+
+    On the other hand, if the pipette is configured with partial nozzle configuration,
+    it uses the usual pipette presses to pick the tips up, in which case, having the tiprack
+    on the Flex 96 channel tiprack adapter (or similar) will cause the pipette to
+    crash against the adapter posts.
+
+    In order to check if the 96-channel can move and pickup/drop tips safely, this method
+    checks for the height attribute of the tiprack adapter rather than checking for the
+    specific official adapter since users might create custom labware &/or definitions
+    compatible with the official adapter.
+    """
+    if not engine_state.pipettes.get_channels(pipette_id) == 96:
+        # Adapters only matter to 96 ch.
+        return
+
+    is_partial_config = engine_state.pipettes.get_is_partially_configured(pipette_id)
+    tiprack_name = engine_state.labware.get_load_name(labware_id)
+    tiprack_parent = engine_state.labware.get_location(labware_id)
+    if isinstance(tiprack_parent, OnLabwareLocation):  # tiprack is on an adapter
+        tiprack_height = engine_state.labware.get_dimensions(labware_id).z
+        adapter_height = engine_state.labware.get_dimensions(tiprack_parent.labwareId).z
+        if is_partial_config and tiprack_height < adapter_height:
+            raise PartialTipMovementNotAllowedError(
+                f"Tiprack {tiprack_name} cannot be on an adapter taller than the tiprack"
+                f"  when the 96-channel is configured for using partial nozzles."
+            )
+        elif not is_partial_config and tiprack_height > adapter_height:
+            raise PipetteMovementNotAllowed(
+                f"Tiprack {tiprack_name}  must be on a Flex Tiprack Adapter (or similar)"
+                f" in order to pick up or drop tips with full 96-channel nozzles."
+            )
+
+    elif (
+        not is_partial_config
+    ):  # tiprack is not on adapter and pipette is in full config
+        raise PipetteMovementNotAllowed(
+            f"Tiprack {tiprack_name}  must be on a Flex Tiprack Adapter (or similar)"
+            f" in order to pick up or drop tips with full 96-channel nozzles."
         )
 
 
--- a/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
+++ b/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
@@ -212,6 +212,9 @@ def check_safe_for_tip_pickup_and_return(
     tiprack_name = engine_state.labware.get_display_name(labware_id)
     tiprack_parent = engine_state.labware.get_location(labware_id)
     if isinstance(tiprack_parent, OnLabwareLocation):  # tiprack is on an adapter
+        is_96_ch_tiprack_adapter = engine_state.labware.get_has_quirk(
+            labware_id=labware_id, quirk="tiprackAdapterFor96Channel"
+        )
         tiprack_height = engine_state.labware.get_dimensions(labware_id).z
         adapter_height = engine_state.labware.get_dimensions(tiprack_parent.labwareId).z
         if is_partial_config and tiprack_height < adapter_height:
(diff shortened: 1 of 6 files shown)


api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -36,6 +36,15 @@ def __init__(self, message: str) -> None:
         )
 
 
+class PipetteMovementNotAllowed(MotionPlanningFailureError):
```
sanni-t: Looking for better names for this error. It's specifically raised when trying to pickup/ return tips with a full 96-ch, to a tiprack either not on the adapter or on an adapter that doesn't meet the height requirements.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -434,12 +439,19 @@ def drop_tip(
...
+                engine_state=self._engine_client.state,
+                pipette_id=self._pipette_id,
+                labware_id=labware_id,
+            )
         deck_conflict.check_safe_for_pipette_movement(
             engine_state=self._engine_client.state,
             pipette_id=self._pipette_id,
             labware_id=labware_id,
             well_name=well_name,
-            well_location=WellLocation(),
```
sanni-t, line 442: This was a bug 😬

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
ecormany: `TipPickupNotAllowed`?

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -36,6 +36,15 @@ def __init__(self, message: str) -> None:
...
+    """Error raised when trying to perform a pipette movement to a tiprack beased on adapter status."""
```
ecormany:
```suggestion
    """Error raised when trying to perform a pipette movement to a tip rack, based on adapter status."""
```

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -174,6 +183,57 @@ def check_safe_for_pipette_movement(
...
+    is_partial_config = engine_state.pipettes.get_is_partially_configured(pipette_id)
+    tiprack_name = engine_state.labware.get_load_name(labware_id)
+    tiprack_parent = engine_state.labware.get_location(labware_id)
+    if isinstance(tiprack_parent, OnLabwareLocation):  # tiprack is on an adapter
+        tiprack_height = engine_state.labware.get_dimensions(labware_id).z
+        adapter_height = engine_state.labware.get_dimensions(tiprack_parent.labwareId).z
+        if is_partial_config and tiprack_height < adapter_height:
+            raise PartialTipMovementNotAllowedError(
+                f"Tiprack {tiprack_name} cannot be on an adapter taller than the tiprack"
+                f"  when the 96-channel is configured for using partial nozzles."
```
ecormany:
```suggestion
                f"Tip rack {tiprack_name} cannot be on an adapter taller than the tip rack"
                f" when picking up fewer than 96 tips."
```
I think this matches language we've used elsewhere, like in the API reference. The removal of reference to the 96-channel pipette is deliberate; you should also get this error if you try to use a 1-channel or 8-channel on a tip rack in an adapter.

ecormany: And should these just be `{tiprack_name}` instead of `Tiprack {tiprack_name}`? I know that the labware name could be anything, since it's user-customizable, but defaults would say things like `Tip rack Opentrons Flex 96 Tip Rack 1000 µL` 🤪

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -174,6 +183,57 @@ def check_safe_for_pipette_movement(
...
+            )
+        elif not is_partial_config and tiprack_height > adapter_height:
+            raise PipetteMovementNotAllowed(
+                f"Tiprack {tiprack_name} must be on a Flex Tiprack Adapter (or similar)"
+                f" in order to pick up or return tips with all 96-channel nozzles."
```
ecormany:
```suggestion
                f"Tip rack {tiprack_name} must be on an Opentrons Flex 96 Tip Rack Adapter"
                f" in order to pick up or return 96 tips simultaneously."
```
Also paralleling language used elsewhere. I think we should refer to the specific labware, or we could be promising more than the hardware API can handle. For example, you have some custom adapter that's just like ours but 5 mm taller…do you crash? is the force correct? etc. It may not Just Work™.

And a fact check: can you `return_tip()` when config is set as `ALL`? Is it only disabled for partial config?

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -174,6 +183,57 @@ def check_safe_for_pipette_movement(
...
+            )
+
+    elif (
+        not is_partial_config
+    ):  # tiprack is not on adapter and pipette is in full config
```
ecormany, line 233: Maybe I'm not following the logic quite right, but this seems like a no-error case.

api/tests/opentrons/protocol_api/core/engine/test_deck_conflict.py
```diff
@@ -473,3 +476,88 @@ def test_deck_conflict_raises_for_bad_partial_8_channel_move(
...
+@pytest.mark.parametrize(
+    argnames=["tiprack_parent", "tiprack_dim", "is_partial_config", "expected_raise"],
+    argvalues=[
+        (
+            DeckSlotLocation(slotName=DeckSlotName.SLOT_5),
+            Dimensions(x=0, y=0, z=50),
+            False,
+            pytest.raises(
+                deck_conflict.PipetteMovementNotAllowed,
+                match="must be on a Flex Tiprack Adapter",
```
ecormany: update test to exactly match labware display name

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sfoster1: I think we should do this in the most straightforward possible way: call it `WrongTipRackForPartialTipPickupError` or something

sfoster1 (approved):
The approach looks good to me; I wouldn't have complained if it was super specific to the point of like checking the adapter labware name, but this is great too.

ecormany (commented):
A few comments on error wordings (and one on error logic). I like the future flexibility of comparing the heights, although I think we should specifically point to our first-party adapter if a user encounters an error.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: This is the case where the tiprack is not on an adapter and the 96 channel is using all its nozzles. We need the tiprack to be on the adapter.

ecormany: Ah, I see, no adapter as opposed to too-short adapter.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: Only disabled for partial config

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -174,6 +183,57 @@ def check_safe_for_pipette_movement(
...
 
 
+def check_safe_for_tip_pickup_and_return(
+    engine_state: StateView,
+    pipette_id: str,
+    labware_id: str,
+) -> None:
+    """Check if the presence or absence of a tiprack adapter might cause any pipette movement issues.
+
+    A 96 channel pipette will pick up tips using cam action when it's configured
```
SyntaxColoring, line 193: I see why we're adding validation for *picking up* tips, but is there physically any danger in *returning* tips the "wrong way"?

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -174,6 +183,57 @@ def check_safe_for_pipette_movement(
...
+    tiprack_parent = engine_state.labware.get_location(labware_id)
+    if isinstance(tiprack_parent, OnLabwareLocation):  # tiprack is on an adapter
+        tiprack_height = engine_state.labware.get_dimensions(labware_id).z
+        adapter_height = engine_state.labware.get_dimensions(tiprack_parent.labwareId).z
+        if is_partial_config and tiprack_height < adapter_height:
+            raise PartialTipMovementNotAllowedError(
+                f"Tiprack {tiprack_name} cannot be on an adapter taller than the tiprack"
+                f"  when the 96-channel is configured for using partial nozzles."
+            )
+        elif not is_partial_config and tiprack_height > adapter_height:
```
SyntaxColoring: This height-based check works, but in the future, should we consider a new labware quirk? `likesHugsFrom96Channel`?

SyntaxColoring (approved):
Awesome, this will prevent a lot of common user mistakes.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: haha literally asked the same question in slack a few minutes ago! Probably indicates a good idea

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: yep, the same danger there is for picking up. When using partial nozzles, even though the pipette doesn't lower as much as it does during pickup, it might still not clear the adapter posts. And I think the 96 channel does a cam action even for tip dropping when using all nozzles so it'll need to be on the tiprack (I'm not super sure of this tho).

PR #14174, opened 2023-12-12
refactor(protocol-designer, step-generation): allow aluminum blocks t…

jerader:
…o be moved, disallow adapters

closes RQA-2060

# Overview

The off-deck slideout now filters out adapters but keeps aluminum blocks. And aluminum blocks can not be moved only manually.

# Test Plan

Create either a Flex or Ot-2 protocol. Add a gripper. On the design tab, click on the edit off-deck button and in the labware selection, the adapters should NOT be available. The aluminum blocks SHOULD be available. Click to add an aluminum block off-deck. 

Create a move labware step and move the aluminum block off-deck and onto a slot. The step should be created correctly. 

Create a new move labware step and move the aluminum block to a new slot (not off-deck) and make sure the `use gripper` checkbox is selected. This step should error and show a timeline error saying that the gripper cannot move aluminum blocks. If you edit the step and uncheck the the box, the step should work as expected.

# Changelog

- clean up `LabwareSelectionModal`'s util logic and filter out adapters if the slot is off-deck
- clean up logic in the selector, allow aluminum blocks to be selectable in the dropdown menu
- create a new error creator `CANNOT_MOVE_WITH_GRIPPER` and add to i18n
- plug the error creator into `moveLabware` atomic command
- add test coverage for the error creator

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/LabwareSelectionModal/LabwareSelectionModal.tsx
```diff
@@ -207,12 +207,16 @@ export const LabwareSelectionModal = (props: Props): JSX.Element | null => {
 
   const getIsLabwareFiltered = React.useCallback(
     (labwareDef: LabwareDefinition2) => {
-      const smallXDimension = labwareDef.dimensions.xDimension < 127.75
-      const smallYDimension = labwareDef.dimensions.yDimension < 85.48
+      const { dimensions, parameters } = labwareDef
+      const { xDimension, yDimension } = dimensions
+
+      const smallXDimension = xDimension < 127.75
```
koji: nit
maybe `isSmallerXDimension` since it would be easy to understand that the const is boolean.

jerader: oh ya good call! thanks!

koji (approved):
Followed the test plan and the sandbox worked as expected.
The changes look good to me.

PR #14175, opened 2023-12-12
refactor(app): special case 96-channel pipette mount text in drop tip wizard

mjhuff:
Closes RQA-2020

# Overview

Per QA/Design feedback. 

### Current Behavior

Screenshot 2023-12-12 at 11 29 12 AM

### With Fix

Screenshot 2023-12-12 at 11 19 21 AM

# Test Plan
- Launch the drop tip flows on the 96 channel via ODD. Easiest way to currently do this is to throw an estop error after picking up a tip. 

# Changelog
- 96-channel now has special drop tip wizard launch text.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
+++ b/app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
@@ -38,7 +38,7 @@ export const handleTipsAttachedModal = (
 
 const TipsAttachedModal = NiceModal.create(
   (props: TipsAttachedModalProps): JSX.Element => {
-    const { mount, onCloseClick } = props
+    const { mount, onCloseClick, instrumentModelSpecs } = props
     const { t } = useTranslation(['drop_tip_wizard'])
     const modal = useModal()
     const [showWizard, setShowWizard] = React.useState(false)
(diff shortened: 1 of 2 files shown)


app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
```diff
@@ -51,6 +51,9 @@ const TipsAttachedModal = NiceModal.create(
       iconColor: COLORS.yellow2,
     }
 
+    const is96Channel = instrumentModelSpecs.displayName.includes('96-Channel')
```
jerader: maybe a better way to filter this is if the channels === 96

jerader (approved):
lgtm! I left a suggestion for filtering for 96-channel

app/src/organisms/DropTipWizard/TipsAttachedModal.tsx
mjhuff: Brilliant, thanks!

koji: A few existing `is96Channel` using `instrumentModelSpecs.channel === 96` so maybe good.
But' I haven't worked on 96 related components, so to me this change makes sense.

koji (approved):
lgtm

PR #14177, opened 2023-12-12
fix(protocol-designer): TC renders for OT-2 and fix isSlotEmpty logic

jerader:
closes RQA-2062

# Overview

Fixes 3 bugs: 
1. the TC wasn't rendering at all on the deck for the OT-2
2. the "getIsSlotEmpty" logic for the OT-2 didn't work completely
3. filter out slot 12 in the OT-2 new location dropdown

# Test Plan

Create an OT-2 protocol and add a thermocycler (gen1 or gen 2). Go to the deck map view and see that it renders correctly and that you can add a labware and the 4 slots are considered full. Move the tiprack into another slot and see that slot 1 and 2 are empty and available to add labware into.

Create a move labware step and see in the new location dropdown that slot 12 is not selectable

# Changelog

- pd is still special-casing GEN1 thermocyclers with the 'span7_8_10_11` slot, so i fixed the logic to change that to slot 7 in the deck setup component. we can probably stop special casing it in the future, post PD 8.0
- fix logic in `isSlotEmpty` util by mapping the slots to the cutouts and comparing that. the reason why there was an error only with the ot-2 is because the slot would be 1 or 2 and the trash bin is in cutout12, which includes 1 and 2.
- add 12 as an addressable area type for the ot-2
- filter out slot 12 in the selector for the new location dropdown

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/DeckSetup/index.tsx
+++ b/protocol-designer/src/components/DeckSetup/index.tsx
@@ -39,6 +39,7 @@ import {
   TRASH_BIN_ADAPTER_FIXTURE,
   WASTE_CHUTE_CUTOUT,
 } from '@opentrons/shared-data'
+import { SPAN7_8_10_11_SLOT } from '../../constants'
 import { selectors as labwareDefSelectors } from '../../labware-defs'
 
 import { selectors as featureFlagSelectors } from '../../feature-flags'
@@ -190,16 +191,12 @@ export const DeckSetupContents = (props: ContentsProps): JSX.Element => {
     <>
       {/* all modules */}
       {allModules.map(moduleOnDeck => {
-        // modules can be on the deck, including pseudo-slots (eg special 'spanning' slot for thermocycler position)
-        // const moduleParentSlots = [...deckSlots, ...values(PSEUDO_DECK_SLOTS)]
-        // const slot = moduleParentSlots.find(
-        //   slot => slot.id === moduleOnDeck.slot
-        // )
-        const slotPosition = getPositionFromSlotId(moduleOnDeck.slot, deckDef)
+        const slotId =
+          moduleOnDeck.slot === SPAN7_8_10_11_SLOT ? '7' : moduleOnDeck.slot
+
+        const slotPosition = getPositionFromSlotId(slotId, deckDef)
         if (slotPosition == null) {
-          console.warn(
-            `no slot ${moduleOnDeck.slot} for module ${moduleOnDeck.id}`
-          )
+          console.warn(`no slot ${slotId} for module ${moduleOnDeck.id}`)
           return null
         }
         const moduleDef = getModuleDef2(moduleOnDeck.model)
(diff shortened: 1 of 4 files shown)


shared-data/deck/types/schemaV4.ts
```diff
@@ -40,6 +40,7 @@ export type OT2AddressableAreaName =
   | '9'
   | '10'
   | '11'
+  | '12'
```
jerader, line 43: Spoke to Brent about this and he said that its fine that it'll be merged into edge and not the release branch since no place in the app lists all the addressable areas for the ot-2.

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -39,6 +39,7 @@ import {
   TRASH_BIN_ADAPTER_FIXTURE,
   WASTE_CHUTE_CUTOUT,
 } from '@opentrons/shared-data'
+import { SPAN7_8_10_11_SLOT } from '../../constants'
```
koji, line 42: nit
Maybe adding `_` between `SPAN` and `7` if the changes wouldn't be a lot.

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -128,17 +143,22 @@ export const getSlotIsEmpty = (
...
-          additionalEquipment.location?.includes(slot) &&
-          additionalEquipment.name !== 'stagingArea'
-      )
+  const filteredAdditionalEquipmentOnDeck = values(
+    initialDeckSetup.additionalEquipmentOnDeck
+  ).filter((additionalEquipment: AdditionalEquipmentOnDeck) => {
+    const cutoutForSlotOt2 = slotToCutoutOt2Map[slot]
+    const includeStaging = includeStagingAreas
+      ? true
+      : additionalEquipment.name !== 'stagingArea'
```
koji, line 152: nit

```suggestion
    const includeStaging = includeStagingAreas
      ? includeStagingAreas
      : additionalEquipment.name !== 'stagingArea'
```

koji, line 152: nit

```suggestion
const includeStaging =
      includeStagingAreas === true
        ? includeStagingAreas
        : additionalEquipment.name !== 'stagingArea'
```

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -128,17 +143,22 @@ export const getSlotIsEmpty = (
...
+    if (cutoutForSlotOt2 != null) {
+      //  for Ot-2
+      return additionalEquipment.location === cutoutForSlotOt2 && includeStaging
+    } else {
+      //  for Flex
+      return additionalEquipment.location?.includes(slot) && includeStaging
```
koji, line 158: nit 
The following might be safe.  

```suggestion
      return ((additionalEquipment.location?.includes(slot)) ?? false) && includeStaging
```

protocol-designer/src/top-selectors/labware-locations/index.ts
```diff
@@ -224,8 +224,7 @@ export const getUnoccupiedLabwareLocationOptions: Selector<
         const isTrashSlot =
           robotType === FLEX_ROBOT_TYPE
             ? MOVABLE_TRASH_ADDRESSABLE_AREAS.includes(slotId)
-            : slotId === 'fixedTrash'
-
+            : slotId === 'fixedTrash' || slotId === '12'
```
koji: completely optional

```ts
['fixedTrash', '12'].includes(slotId)
```

koji (approved):
left a couple of comments, but the changes look good to me.

protocol-designer/src/components/DeckSetup/index.tsx
jerader: this const is used all over unfortunately so i'll just keep it as is. I hope we can deprecate the usage completely at some point.

PR #14148, opened 2023-12-08
changes so far found in v7.1.0-alpha.2 vs v7.0.2

y3rsh:
# Use snapshots to compare protocol analysis across builds

## Bugs found

- <https://opentrons.atlassian.net/browse/RQA-2001> 
  - This is manual blowout? Should not throw error?
- <https://opentrons.atlassian.net/browse/RQA-2032> 
  - addressed with #14145 

## Signoff

- [ ] All these changes are expected and safe

PR #14179, opened 2023-12-12
fix(protocol-engine): Fix `blowOutInPlace` and `aspirateInPlace` not updating current volume

SyntaxColoring:
# Overview

Fixes RQA-2056.

# Test Plan

I've that the protocol from RQA-2056 and this more minimal protocol both pass analysis now, as expected:

```python


    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "A1")
    well_plate = protocol.load_labware("corning_96_wellplate_360ul_flat", "A2")
        "flex_1channel_1000", mount="right", tip_racks=[tip_rack]
    )

    trash = protocol.load_trash_bin("C3")

    pipette.pick_up_tip()

    pipette.aspirate(500, well_plate["A1"])
    pipette.air_gap()

    pipette.blow_out(trash)

    pipette.aspirate(1, well_plate["A1"])  # Should not raise.
```

And this protocol now raises an analysis error, as expected:

```python
from opentrons import types


    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "A1")
        "flex_1channel_1000", mount="right", tip_racks=[tip_rack]
    )

    pipette.pick_up_tip()
    pipette.move_to(types.Location(types.Point(10, 20, 30), labware=None))

    pipette.aspirate(900)
    pipette.aspirate(101)  # Should raise.
```

# Changelog

When Protocol Engine processes an `aspirate`, `dispense`, `blowOut`, etc., it's supposed to keep track of how much liquid is in the pipette, for the purposes of preventing mistakes like aspirating more than 1000 µL into a 1000 µL pipette.

It was simply forgetting to do this for `aspirateInPlace` and `blowOutInPlace`. So, add in that handling, and make sure it's covered by unit tests.

# Review requests

I've skimmed `PipetteStore` and it doesn't look like it's forgetting any other commands. Do you have any other commands that you're worried about, off the top of your head?

# Risk assessment

Medium. This touches a core part of protocol analysis.

This *may* cause protocols to fail analysis where they were previously succeeding. I think that's good in this case because, at least to my understanding, these protocols always would have errored out when you actually went to run them.

We should keep an eye on this PR's effects on the analysis snapshots in #14148 to make sure there are no unexpected failures.

### DIFF
--- a/api/tests/opentrons/protocol_engine/state/test_pipette_store.py
+++ b/api/tests/opentrons/protocol_engine/state/test_pipette_store.py
@@ -163,7 +163,7 @@ def test_handles_drop_tip_in_place(subject: PipetteStore) -> None:
     assert subject.state.aspirated_volume_by_id["xyz"] is None
 
 
-def test_pipette_volume_adds_aspirate(subject: PipetteStore) -> None:
+def test_aspirate_adds_volume(subject: PipetteStore) -> None:
     """It should add volume to pipette after an aspirate."""
     load_command = create_load_pipette_command(
         pipette_id="pipette-id",
@@ -225,7 +225,7 @@ def test_handles_blow_out(subject: PipetteStore) -> None:
         ),
     ],
 )
-def test_pipette_volume_subtracts_dispense(
+def test_dispense_subtracts_volume(
     subject: PipetteStore, dispense_command: cmd.Command
 ) -> None:
     """It should subtract volume from pipette after a dispense."""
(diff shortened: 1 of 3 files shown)


api/tests/opentrons/protocol_engine/state/test_pipette_store.py
```diff
@@ -192,28 +200,6 @@ def test_pipette_volume_adds_aspirate(subject: PipetteStore) -> None:
...
-
-    assert result.aspirated_volume_by_id["pipette-id"] is None
-
-    assert result.current_location == CurrentWell(
-        pipette_id="pipette-id",
-        labware_id="labware-id",
-        well_name="well-name",
-    )
-
-
```
SyntaxColoring, line 216: The `result.current_location` part of this test was already covered by `test_movement_commands_update_current_well()`. The `result.aspirated_volume_by_id` part of this test is moved to the new `test_blow_out_clears_volume()` test so it can be shared with `blowOutInPlace`.

sfoster1 (approved):
Oof, good fix.

PR #14180, opened 2023-12-12
fix(app): make sure pipette flow wizard title and results copy is correct

smb2268:
fix RQA-2018, RQA-2039

# Overview
Fixing a few bugs related to pipette flow wizard header and result screen copy

# Test Plan

- [x] With two pipettes attached to a flex, start run setup for a protocol that uses a 96 channel pipette. When going through the wizard flow, the title should be "Detach Pipettes and Attach 96-Channel Pipette" instead of referencing one specific pipette to detach: 
Screen Shot 2023-12-12 at 3 47 37 PM
- [ ] When calibrating a pipette, results screen should now say "{{pipette_name}} successfully calibrated" instead of "successfully attached and calibrated" regardless of whether the calibration followed an attachment flow since we show an "successfully attached" screen 

# Changelog
1. Change "attached and calibrated" to "calibrated"
2. Memoize wizard title so that no changing inputs like different pipettes being attached through the duration of the flow will affect the title

# Review requests
sanity check code - I've tested this

# Risk assessment
Low

### DIFF
--- a/app/src/assets/localization/en/pipette_wizard_flows.json
+++ b/app/src/assets/localization/en/pipette_wizard_flows.json
@@ -56,7 +56,7 @@
   "ninety_six_detached_success": "{{pipetteName}} pipette successfully detached",
   "ninety_six_probe_location": "A1 (back left corner)",
   "pip_cal_failed": "pipette calibration failed",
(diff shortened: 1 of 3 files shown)


koji (approved):
lgtm!
tested both cases with robots and both cases worked as expected.

PR #14182, opened 2023-12-12
refactor(api, robot-server): removal of automatic drop tip from Flex while maintaining OT2 behavior

CaseyBatten:
# Overview
Closes RQA-1990 RQA-2022 RSS-252
This will currently remove the automatic drop tip behavior from do_stop_and_recover for Flex protocols. The decision is made in robot-server under the engine store when create_protocol_run is called. For the time being, this is determined based on Robot type. If the robot is a Flex, we set drop_tips_after_run to false, otherwise it is always true. This could be changed to a settings flagged triggered boolean in the future to enable-drop tip behavior on Flex (other changes would be needed covered below related to loaded trashes).

In do_stop_and_recover, if there are still tips attached it will home the gantry and the gripper, but it will not home the plunger. The private _drop_tip function called there now has a changed behavior utilizing move_to_addressable_area. Eventually we will want to update this to pull a dynamic list of loaded trashes, for which I have added a function that returns the list of loaded trash addressable areas within protocol engine. We cannot do this currently, as a trash (and any addressable area) is not loaded to the PE addressable area store unless it is actually used. We will need to change that behavior in the future. What this will eventually allow us to do is automatically drop the tip in a trash can so long as it exists within the deck configuration, regardless of if it is used. For now, only the OT2 supports this functionality (as only the OT2 should automatically drop tips) and is simply provided a hard-coded "fixedTrash" addressable area to move to for tip drop. If a Flex protocol reaches this point (which it should not, except in analysis), for the time being it will log a debug message that states "Flex protocols do not support automatic tip dropping at this time".

# Test Plan
Run both OT2 and Flex protocols where the following happens:

- Tips are picked-up
- Tips are dropped in a trash
- Tips are picked-up

Nothing else should occur in the protocol after this.
The expected results are as follows:
**FLEX:**
The robot should pickup and throw away tips as expected, then pick up the second set of tips and end the protocol. After this, it should home the gantry and the grippers but NOT the plungers. It will then end the protocol and prompt the user, on both the App and ODD that there are still tips attached, and allow them to do a manual drop (if they want) through the tip drop wizard.

**OT2:**
The robot should pickup and throw away tips as expected, then pick up the second set of tips and go to the trash and throw away the remaining tips automatically.

# Risk assessment
On Flex protocols users are now expected to explicitly drop any remaining tips at the end of their protocols. This means that any protocol designed before this refactor will now be stuck with tips at the end if the users designed a protocol that would end with tips still on the pipette at the end.


api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -133,7 +135,20 @@ def __init__(
...
-        self._task_queue = task_queue or TaskQueue(cleanup_func=protocol_engine.finish)
+        self._task_queue = (
+            task_queue or TaskQueue()
+        )  # cleanup_func=protocol_engine.finish))
+        self._task_queue.set_cleanup_func(
+            func=protocol_engine.finish,
+            drop_tips_after_run=drop_tips_after_run,
+            post_run_hardware_state=post_run_hardware_state,
+        )
+        # self._task_queue.set_post_hardware_run_state(
```
sfoster1: commented on purpose?

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -223,7 +240,21 @@ def __init__(
...
```
sfoster1: commented on purpose?

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -312,7 +343,11 @@ def __init__(
         # TODO(mc, 2022-01-11): replace task queue with specific implementations
         # of runner interface
         self._hardware_api = hardware_api
-        self._task_queue = task_queue or TaskQueue(cleanup_func=protocol_engine.finish)
+        self._task_queue = (
+            task_queue or TaskQueue()
+        )  # cleanup_func=protocol_engine.finish)
```
sfoster1: commented out on purpose?

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -336,6 +336,21 @@ def get_all_cutout_fixtures(self) -> Optional[List[str]]:
                 for _, cutout_fixture_id in self._state.deck_configuration
             ]
 
+    def get_disposal_addressable_areas(self) -> List[str]:
```
SyntaxColoring: Is the plan for this to be used in the v7.1.0 launch? If not, I'd say leave it out of the code for now.

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
```diff
@@ -68,25 +70,43 @@ async def _drop_tip(self) -> None:
...
+                        # if API is 2.15 and below aka is should_have_fixed_trash
+
+                        await self._movement_handler.move_to_addressable_area(
+                            pipette_id=pipette_id,
+                            addressable_area_name="fixedTrash",
+                            offset=AddressableOffsetVector(x=0, y=0, z=0),
+                            force_direct=False,
+                            speed=None,
+                            minimum_z_height=None,
+                        )
```
SyntaxColoring, line 88: How does this work if the OT-2 protocol has a fixed trash *labware* (definitely PAPIv2.14 and PAPIv2.15, and who knows what happens with PAPIv≤2.13)? I would expect referencing the `fixedTrash` addressable area to raise a conflict error, because it can't exist simultaneously with the labware?

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -336,6 +336,21 @@ def get_all_cutout_fixtures(self) -> Optional[List[str]]:
...
+        """Get a list of all loaded disposal locations by addressable area name."""
+        # TODO: Fixed Trash, Trash and WasteChute addressable areas are only added to
+        # loaded_addressable_areas_by_name when used, we should add them when loaded
+        loaded_disposal_locations = []
+        for area in self._state.loaded_addressable_areas_by_name:
+            if (
+                ("fixedTrash" in area)
+                or ("movableTrash" in area)
+                or ("WasteChute" in area)
+            ):
```
SyntaxColoring: Moot if we're removing this, but the addressable areas have a semantic `ableToDropTips` field. Let's use that instead of hard-coding these name substrings.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -336,6 +336,21 @@ def get_all_cutout_fixtures(self) -> Optional[List[str]]:
...
+                # Append any and all instances/configurations of a MovableTrash or a WasteChute from this deck config
+                loaded_disposal_locations.append(area)
```
SyntaxColoring: Nitpicky Python style thing: we'd typically prefer to write this as a list comprehension instead of as an iterative loop.

```python
loaded_disposal_locations = [
    area for area in self._state.loaded_addressable_areas_by_name
    if "fixedTrash" in area or "movableTrash" in area or "WasteChute" in area
]
```

CaseyBatten:
Protocols meeting the description of the test case above were ran on both OT2 and Flex, and the desired behavior was seen. OT2 automatically dropped tips at the end then homed, and Flex simply homed (without the plungers) keeping the tips attached.

robot-server/robot_server/runs/engine_store.py
```diff
@@ -185,10 +185,20 @@ async def create(
...
+        if self._robot_type == "OT-3 Standard":
+            post_run_hardware_state = PostRunHardwareState.HOME_AND_STAY_ENGAGED
+            drop_tips_after_run = False
+
         runner = create_protocol_runner(
             protocol_engine=engine,
             hardware_api=self._hardware_api,
             protocol_config=protocol.source.config if protocol else None,
+            post_run_hardware_state=post_run_hardware_state,
+            drop_tips_after_run=drop_tips_after_run,
```
SyntaxColoring, line 200: I think this is okay for this PR if we need to, but it's really looking to me now like we should move `post_run_hardware_state` and `drop_tips_after_run` into `Config` [to avoid all of these layers having to know about them](https://www.mattduck.com/2021-04-a-philosophy-of-software-design.html#:~:text=general%2Dpurpose%20methods.%22-,8.%20Pass%2Dthrough%20variables%20add%20complexity,-Pass%2Dthrough%20variables).

api/src/opentrons/protocol_runner/task_queue.py
```diff
@@ -74,4 +92,11 @@ async def _run(self) -> None:
...
 
-        await self._cleanup_func(error=error)
+        if self._cleanup_func is not None:
+            await self._cleanup_func(error)
+
+        # await self._cleanup_func(
+        #     error=error,
+        #     drop_tips_after_run=self._drop_tips_after_run,
+        #     post_run_hardware_state=self._post_hardware_run_state,
+        # )
```
SyntaxColoring: Commented code here and elsewhere—is this left over from debugging?

api/src/opentrons/protocol_runner/task_queue.py
```diff
@@ -5,14 +5,16 @@
...
 
 
 class CleanupFunc(Callback):
     """Expected cleanup function signature."""
 
-    def __call__(self, error: Optional[Exception]) -> Any:
+    def __call__(
+        self,
+        error: Optional[Exception],
+    ) -> Any:
```
SyntaxColoring, line 17: It looks like `CleanupFunc` is unused now that `set_cleanup_func()` spells out `Callable[..., Awaitable[Any]]` itself.

Let's change `TaskQueue.set_cleanup_func()` and `TaskQueue.__init__()` to keep using `CleanupFunc` instead of spelling it out themselves. This `error` positional argument is a bit special and it's easy to accidentally omit if every place has to spell it out from scratch (`set_cleanup_func()` omitted it, for example).

sfoster1 (approved):
Looks good to me except the commented out code; we should merge it once that's addressed, but I think we should quickly follow up with going back to the original behavior for protocol API versions below 2.16.

SyntaxColoring (commented):
I think this makes sense but I have [one question](https://github.com/Opentrons/opentrons/pull/14182#discussion_r1424620583) about how this interacts with OT-2 protocols that have fixed trash labware.

api/src/opentrons/protocol_engine/state/addressable_areas.py
CaseyBatten: Not currently, I had implemented it when I was trying to dynamically choose the trash to utilize, however I ran into the issue that trash is only loaded into the PE store if it is *used* in a protocol, which means doing a protocol that only consists of a pickup tip and nothing else will result in a protocol engine addressable area list with zero trashes anywhere in it, even a fixed trash. I can remove it for now.

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
CaseyBatten: Just ran it on an OT2 with 2.16, 2.15, 2.13 and 2.11 protocols that do the test case I mentioned in the PR post. Each of them at the end did an automatic tip drop in the fixedTrash as expected, so it does work even on older API versions. Not too sure there would be a conflict between a fixedTrash addressable area existing in the deck config of an OT2 protocol while a fixedTrash labware exists as well, because the protocol end of things wouldn't really be aware of it if they are running a legacy protocol... right?

SyntaxColoring: Huh, that's surprising.

What I have in mind is: on a Flex, I can't load a labware into slot D3 *and* use the `wasteChute` addressable area. Loading the labware needs `D3`, and using the waste chute needs `wasteChute`, and those are mutually exclusive.

And I thought that was how this was working for the OT-2, too. The fixed trash labware is "in" "slot 12," which occupies cutout 12, so how is a fixed trash *also* allowed to occupy cutout 12?

CaseyBatten: I'm not entirely sure. My understanding is that move to addressable area will move to an area so long as it exists within the deck configuration. I believe the OT2 always has a fixed trash in its deck configuration, so it is a valid call. I don't think we ever explicitly do a "load fixed trash in slot 12" type of command in OT2 protocols, and the labware is added as a created entity to the disposal_locations list in the PAPI automatically, "loaded" by the protocol context. So passing in a reference to the "created" labware or the "created" addressable area both are valid? Might be a weird blindspot in our architecture. Either way as an example, a generic drop_tip call goes to the fixed trash on a 2.13 protocol (it must because nothing else exists), and in the end picking up a tip and stopping a protocol then results in us moving to the fixedTrash addressable area and dropping tips there, so both references must be valid internally. Odd.

PR #14183, opened 2023-12-12
docs(api): new section for `push_out`

ecormany:
# Overview

Explanation and example of the `push_out` parameter of `dispense()`.

Addresses RTC-373.

# Test Plan

- Checked both the [new section](http://sandbox.docs.opentrons.com/docs-push_out/v2/basic_commands/liquids.html#push-out-after-dispense) and the [API reference entry](http://sandbox.docs.opentrons.com/docs-push_out/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.dispense) in the sandbox.
- Analyzed protocol with example code.

# Changelog

- New section on Liquid Control page.
- Version changed statement and link in from API reference.

# Review requests

- Double check code
- Decision about API vs robot software as the breaking point for the behavior change. Will update note if needed.
- 
# Risk assessment

nil, docs

### DIFF
--- a/api/docs/v2/basic_commands/liquids.rst
+++ b/api/docs/v2/basic_commands/liquids.rst
@@ -115,6 +115,25 @@ Flex and OT-2 pipettes dispense at :ref:`default flow rates <new-plunger-flow-ra
 
 .. versionadded:: 2.0
 
+.. _push-out-dispense:
+
+Push Out After Dispense
+-----------------------
+
+Use the optional ``push_out`` parameter of ``dispense()`` to help guarantee all liquid leaves the tip. Pushing out is designed for applications where you need to move the pipette plunger lower than the default, without performing a full :ref:`blow out <blow-out>`.
+
+For example, this dispense action moves the plunger the equivalent of an additional 5 µL beyond where it would stop if ``push_out`` was set to zero or omitted::
+
+    pipette.pick_up_tip()
+    pipette.aspirate(100, plate['A1'])
+    pipette.dispense(100, plate['B1'], push_out=5)
+    pipette.drop_tip()
+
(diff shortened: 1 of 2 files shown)


api/docs/v2/basic_commands/liquids.rst
```diff
@@ -115,6 +115,25 @@ Flex and OT-2 pipettes dispense at :ref:`default flow rates <new-plunger-flow-ra
 
 .. versionadded:: 2.0
 
+.. _push-out-dispense:
+
+Push Out After Dispense
+-----------------------
+
+Use the optional ``push_out`` parameter of ``dispense()`` to help guarantee all liquid leaves the tip. Pushing out is designed for applications where you need to move the pipette plunger lower than the default, without performing a full :ref:`blow out <blow-out>`.
```
jwwojak: I'd recommend a word change from "guarantee." That's a pretty strong commitment here even fi modified by "to help." Instead, consider these:

- "to help ensure all liquid leaves the tip"
- "to help remove all liquid from the tip"
- "to help remove any remaining liquid from the tip"

Also this reads a little awkward to me:
> Pushing out is designed for

Instead, consider these ideas:

- This parameter is designed for ...
- Use this parameter when your applications need to ...
- Use `push_out` when you need to ...

The remaining text LGTM!

jwwojak (commented):
Added comment and LGTM. :shipit:

api/docs/v2/basic_commands/liquids.rst
ecormany: yeah, "help guarantee" is a funny hedge

jwwojak (approved):
Make it so. 🚀

PR #14184, opened 2023-12-12
feat(app, step-generation, shared-data): FE support for moveToAddressableAreaForDropTip

jerader:
closes RAUT-901

# Overview

Adds FE support for the new drop tip command which is used to drop tips in to the trash bin

# Test Plan

Currently blocked until there is API support for this. But:
1. create a PD protocol, either flex or OT-2 and create a transfer or mix step then drop tip into the trash bin, note: you can turn on the deck config ff if you want but its not necessary. 
2. open the JSON file and see that the command for drop tip includes "moveToAddressableAreaForDropTip" and does NOT include "dropTipInPlace" (since you're only dropping tip in the trash bin)
3. (This step won't work until we have backend support) upload the protocol to the app, in the command text there should be information about moving to the trash bin to drop tip.
# Changelog

- add new command to types
- add logic for command text and add to the test
- add a new atomic command in step generation
- add a robot state update fn to keep track of tip and liquid state
- plug new command into the movable trash commands util
- update all step-generation tests
- update the e2e tests

# Review requests

see test plan

# Risk assessment

low-ish


protocol-designer/src/components/FileSidebar/utils/__tests__/getUnusedTrash.test.ts
```diff
@@ -31,17 +31,15 @@ describe('getUnusedTrash', () => {
...
       {
-        labwareId: {
-          commandType: 'moveToAddressableArea',
-          params: { adressableAreaName: 'cutoutA3' },
-        },
+        commandType: 'moveToAddressableArea',
+        params: { addressableAreaName: 'movableTrashA3' },
       },
-    ] as unknown) as CreateCommand[]
+    ] as CreateCommand[]
```
jerader, line 39: cleaning up a lot of oopsies in this test suite! I mocked `mockCommand` completely wrong for all of the test cases 🤦

app/src/assets/localization/en/protocol_command_text.json
```diff
@@ -37,6 +37,7 @@
   "move_to_slot": "Moving to Slot {{slot_name}}",
   "move_to_well": "Moving to well {{well_name}} of {{labware}} in {{labware_location}}",
   "move_to_addressable_area": "Moving to {{addressable_area}}",
+  "move_to_addressable_area_drop_tip": "Moving to {{addressable_area}} to drop tip",
```
jerader: @ecormany, does this command text make sense for moving to the trash bin to drop tip? Maybe something like "Dropping tip in {{addressable_area}}" would be better?

ecormany: Current run log messages are of this form: `Dropping tip into A1 of Opentrons Fixed Trash on slot 12`

So `Dropping tip into {{addressable_area}}` would make sense.

jerader: Great, thank you! I'll update it!

step-generation/src/fixtures/commandFixtures.ts
```diff
@@ -298,6 +298,20 @@ export const delayWithOffset = (
...
+    key: expect.any(String),
+    params: {
+      pipetteId: pipette ?? DEFAULT_PIPETTE,
+      addressableAreaName: 'movableTrashA3',
+      offset: { x: 0, y: 0, z: 0 },
+      alternateDropLocation: true,
+    },
+  },
+]
+export const dropTipIntoWasteChuteHelper = (
```
brenthagen, line 312: what's this for? should it move to the waste chute?

brenthagen (approved):
changes make sense. can't test end to end but I see the `moveToAddressableAreaForDropTip` command generated.

step-generation/src/fixtures/commandFixtures.ts
jerader: its a command fixture for the step-generation compound commands, i just renamed it to specify that it is a fixture used for dropping tips into waste chute only

PR #14185, opened 2023-12-12
fix(api): display name in deck conflict error message

sanni-t:
# Overview

In #14173 I didn't realize that `labware_view.get_display_name` provided only user-specified display name. So deck conflict error messages could say "None should be on an Opentrons adapter.." if no custom display name was specified for the tiprack. 
This PR fixes that mistake.

# Test Plan

The protocol attached in #14173 should now show errors with correct display names.

# Changelog

- renamed `labware_view.get_display_name` to `get_user_specified_display_name`
- added a new `labware_view.get_display_name` that returns the user specified display name if present, else returns the display name from labware definition
- use this new getter in the error message

# Review requests

- Usual code review stuff

# Risk assessment

Low. Bug fix


sfoster1 (approved):
Looks good to me!

SyntaxColoring (approved):
TY!

PR #14186, opened 2023-12-13
fix(api): Fix RuntimeError when calling InstrumentContext.type on a 96-channel

SyntaxColoring:
# Overview

Closes RSS-342.

# Test Plan

I've made sure that this passes analysis now:

```python
def run(protocol):
    pipette = protocol.load_instrument("flex_96channel_1000", mount="left")
    protocol.comment(pipette.type)
```

# Changelog

* Make `InstrumentContext.type` return `"multi"` for 96-channels, the same way it does for 8-channels.
* In the docs, point people towards `InstrumentContext.channels` instead. It seems strictly better because it can distinguish between `1` and `8` and `96`.
* Add unit test coverage for `.channels` and `.type`.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1494,9 +1494,8 @@ class InstrumentContext(publisher.CommandPublisher):
     def type(self) -> str:
         """``'single'`` if this is a 1-channel pipette, or ``'multi'`` otherwise.
 
-        .. deprecated:: 2.16
-           Use :py:obj:`.channels` instead, which can distinguish between 8-channel and 96-channel
-           pipettes.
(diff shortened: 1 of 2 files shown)


sanni-t:
Is it really okay to deprecate `type`? @Laura-Danielle mentioned that a lot of users prefer using it over `channels` currently. But not sure what the problem could be in switching to use channels in new protocols.

SyntaxColoring:
> Is it really okay to deprecate `type`? @Laura-Danielle mentioned that a lot of users prefer using it over channels currently. But not sure what the problem could be in switching to use channels in new protocols.

🤷‍♂️ No opinion from me. I can change it to a "see also" if that's better.

sanni-t (approved):
lgtm!

PR #14187, opened 2023-12-13
feat(api): Allow omitting the mount when loading a 96-channel

SyntaxColoring:
# Overview

The 96-Channel Pipette physically occupies both mounts. But a quirk (not ticketed) of our current Python Protocol API is that you need to specify `mount="left"` or `mount="right"`. This PR fixes that quirk.

# Changelog

Before this PR:

* The Python Protocol API's `load_instrument()` method takes a `mount` argument that's required to be `"left"` or `"right"`.
* The documentation says to use `"left"` if you're loading a 96-Channel.
* The implementation ignores whichever value you pass in; `"left"` and `"right"` do the same thing in practice.

With this PR:

* The `mount` argument is allowed to be omitted or `None` if you're loading a 96-channel. For all other pipettes, it's still required to be `"left"` or `"right"`.
* The documentation says to omit `mount` if you're loading a 96-channel.
* The implementation continues to ignore whichever value you pass in. This is for compatibility with existing protocols.

This is roughly analogous to how we treat the Thermocycler Module, which can only be loaded into one location.

# Test Plan

* [x] Try out some different load syntaxes and make sure they raise errors, or don't raise them, according to your expectations.

# Review requests

Is this a good idea to do now, or were we deliberately making people do `mount="<something>"`?

# Risk assessment

Low.

### DIFF
--- a/api/docs/v2/new_pipette.rst
+++ b/api/docs/v2/new_pipette.rst
@@ -48,7 +48,7 @@ If you're writing a protocol that uses the Flex Gripper, you might think that th
 Loading a Flex 96-Channel Pipette
 ---------------------------------
 
-This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. To load the 96-Channel Pipette, omit the ``mount`` argument from ``load_instrument()`` as shown here:
+This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. When loading the 96-Channel Pipette, you can omit the ``mount`` argument from ``load_instrument()`` as shown here:
 
 .. code-block:: python
 
--- a/api/docs/v2/new_pipette.rst
+++ b/api/docs/v2/new_pipette.rst
@@ -48,13 +48,13 @@ If you're writing a protocol that uses the Flex Gripper, you might think that th
 Loading a Flex 96-Channel Pipette
 ---------------------------------
 
-This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. To load the 96-Channel Pipette, specify its position as ``mount='left'`` as shown here:
+This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. To load the 96-Channel Pipette, omit the ``mount`` argument from ``load_instrument()`` as shown here:
 
 .. code-block:: python
 
     def run(protocol: protocol_api.ProtocolContext):
-        left = protocol.load_instrument(
-            instrument_name='flex_96channel_1000', mount='left')
+        pipette = protocol.load_instrument(
+            instrument_name='flex_96channel_1000')
 
 .. versionadded:: 2.15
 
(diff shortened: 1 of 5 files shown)


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -831,15 +831,16 @@ def load_instrument(
         ensure that the correct instrument is attached in the specified
         location.
 
-        :param str instrument_name: The name of the instrument model, or a
-                                    prefix. For instance, 'p10_single' may be
```
SyntaxColoring, line 835: I'm not sure when this prefix thing was true, but [it is not true now](https://github.com/Opentrons/opentrons/blob/f98e49b0b106989b1a6bc2a0beaf7b10cb9d981d/api/src/opentrons/protocol_api/validation.py#L131) and it's weird, so I'm removing it from the docs.

api/docs/v2/new_pipette.rst
```diff
@@ -48,13 +48,13 @@ If you're writing a protocol that uses the Flex Gripper, you might think that th
 Loading a Flex 96-Channel Pipette
 ---------------------------------
 
-This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. To load the 96-Channel Pipette, specify its position as ``mount='left'`` as shown here:
+This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. To load the 96-Channel Pipette, omit the ``mount`` argument from ``load_instrument()`` as shown here:
```
ecormany: Unless this change is _enforcing_ no argument for 96-channel, let's say something like `When loading the 96-Channel Pipette, you can omit the ``mount`` argument...`

api/src/opentrons/protocol_api/protocol_context.py
ecormany: 🚮

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -831,15 +831,16 @@ def load_instrument(
...
-                                    used to request a P10 single regardless of
-                                    the version.
-        :param mount: The mount in which this instrument should be attached.
+        :param str instrument_name: Which instrument you want to load. See :ref:`new-pipette-models`
+                                    for the valid values.
+        :param mount: The mount to which this instrument should be attached.
```
ecormany: less 🧐 
```suggestion
        :param mount: The mount where this instrument should be attached.
```

ecormany (approved):
From docs side, go ahead with this. The docstrings will carry along no problem. I'm doing a huge refactor of the Pipettes page (#14189), to the point it's not compatible with rebasing, so I'll incorporate those changes by copy-paste into that branch.

api/src/opentrons/protocol_api/protocol_context.py
SyntaxColoring: mm yes indubitably

api/docs/v2/new_pipette.rst
SyntaxColoring:
```suggestion
This code sample loads the Flex 96-Channel Pipette. Because of its size, the Flex 96-Channel Pipette requires the left *and* right pipette mounts. You cannot use this pipette with 1- or 8-Channel Pipette in the same protocol or when these instruments are attached to the robot. When loading the 96-Channel Pipette, you can omit the ``mount`` argument from ``load_instrument()`` as shown here:
```

sanni-t (approved):

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -850,9 +851,18 @@ def load_instrument(
...
 
-        checked_mount = Mount.LEFT if is_96_channel else validation.ensure_mount(mount)
+        is_96_channel = checked_instrument_name == PipetteNameType.P1000_96
+        if is_96_channel:
+            checked_mount = Mount.LEFT
+        else:
+            if checked_mount is None:
+                raise ValueError(
+                    f"You must specify a mount to load {repr(instrument_name)}."
+                )
```
sanni-t: You could even move this logic to `validation.ensure_mount` so that it always returns a valid mount. Of course you'll need to pass the pipette name to check for 96 channel's mount.

SyntaxColoring: Yep, good call, this makes the tests a lot cleaner.

PR #14189, opened 2023-12-13
docs(api): refactor Pipettes page to include partial tip pickup

ecormany:
# Overview

Complete refactor of the single Pipettes page into multiple pages. It was already approaching 4000 words, and explaining partial tip pickup will take another ~1500. 

# Test Plan

- [x] Check out the four affected pages (landing page and three new subpages) in the [sandbox](http://sandbox.docs.opentrons.com/docs-refactor-pipettes-partial-pickup/v2/new_pipette.html).
- [x] Test all code snippets with `opentrons_simulate` running from `v7.1.0` tag. [protocol file](https://gist.github.com/ecormany/504963f51d5522ce0c8d73f94aadb62b)

# Changelog

- Turned `new_pipette.rst` into a landing page.
- New Loading Pipettes page. Reorganizes information and adds a couple new notes.
- New Pipette Characteristics page. Adds some info on multi-channel pipette movement.
- New Partial Tip Pickup page. All new content. The big one.
- New Volume Modes page, same as previously existing section.

# Review requests

Prose: @jwwojak 
- Does Partial Tip Pickup section make sense?
- Should Partial Tip Pickup be its own page?

Code: @sanni-t 
- I wrote a lot of snippets. Do they analyze? Does the code style look right?
- Read through the Partial Tip Pickup section and let me know if anything is wrong or missing based on final feature decisions for 7.1.0.
- Does it make sense to use `pop()` for ad hoc tip tracking?

# Risk assessment

No code risk. Big refactor for docs.


api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+
+There are no mode configurations for OT-2 pipettes.
+
+
+.. _partial-tip-pickup:
+
+Partial Tip Pickup
+==================
+
+The 96-channel pipette occupies both pipette mounts on Flex, so it's not possible to attach another pipette at the same time. Partial tip pickup lets you perform some of the same actions that you would be able to perform with a second pipette. As of version 2.16 of the API, you can configure the 96-channel pipette pick up a single column of tips, similar to the behavior of an 8-channel pipette.
```
sanni-t:
```suggestion
The 96-channel pipette occupies both pipette mounts on Flex, so it's not possible to attach another pipette at the same time. Partial tip pickup lets you perform some of the same actions that you would be able to perform with a second pipette. As of version 2.16 of the API, you can configure the 96-channel pipette to pick up a single column of tips, similar to the behavior of an 8-channel pipette.
```

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+While you can easily add tip tracking to a column 1 configuration, you will still be operating without the collision detection the API has for column 12. 
+
+.. warning::
+
+    The API *will not* raise errors for potential labware crashes when using a column 1 partial configuration. If you must use one:
+
+    - Plan your deck layout carefully. Make a diagram and visualize everywhere the pipette will travel.
+    - Simulate your protocol and compare the run preview to your expectations of where the pipette will travel.
+    - Perform a dry run with only tip racks on the deck. Have the Emergency Stop Pendant handy in case you see an impending crash.
+
```
sanni-t: Could also include the tip about the having the deck layout according to labware height; the opposite of recommended layout for column 12:
```
generally organize your deck with the shortest labware on the right side of the deck, and the tallest labware on the left side.
```

sanni-t: Another point that was found from recent testing is that the leftmost nozzle column cannot access the last 2-3 well columns of labware in column 3 of the deck. (Precisely, 28mm area from the right edge of a column 3 deck slot is inaccessible)

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+First, tip tracking is not available with column 1. You must always specify a ``location`` parameter for :py:meth:`.pick_up_tip`. This *requires careful tip tracking* so you don't place the pipette over more than a single column of unused tips at once. You can write some additional code to manage valid tip pickup locations, like this::
+
+    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "C1")
+    pipette.configure_nozzle_layout(style=COLUMN, start="A1")
+    row_a = tiprack.rows()[0]
+    pipette.pick_up_tip(row_a.pop())  # pick up A12-H12
+    pipette.drop_tip()
+    pipette.pick_up_tip(row_a.pop())  # pick up A11-H11
+    pipette.drop_tip()
```
sanni-t: smart!

sanni-t (commented):
Minor changes. Rest all looks good. Very detailed documentation!

api/docs/v2/pipettes/configuring.rst
ecormany: I was kinda choosing to focus the warning in this subsection. The best "tip" is to not use A1 for now 😅

sanni-t: That's true. You can skip this suggestion then if you think it'll make users assume it's ok to use A1.

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+*************************
+
+The API provides methods to configure certain pipettes to operate in a specific mode. These configuration methods change the default behavior of the pipette on an ongoing basis, until you call them again with different values. 
+
+This page covers the currently available modes.
+
+- :ref:`partial-tip-pickup`: The 96-channel pipette can be configured to use 8 or 96 nozzles for pipetting.
+- :ref:`pipette-volume-modes`: The Flex 50 µL pipettes can be configured to handle liquid differently for low or high volumes.
+
+There are no mode configurations for OT-2 pipettes.
```
jwwojak: Intros are always hard to write. There's something about the first sentence on line 7 and then the last sentence on line 14 that gnaws at me.

**Line 7**
The phrases "certain pipettes" and "a specific mode" are uncomfortably vague. I understand that being exact here may make the doc less future proof, but this sentence leaves me wanting more. As the child sections go on to discuss the 96 channel and low/high volumes for the 50 ul pipettes, consider adding a wee more specificity to the intro. Help the reader understand what's in here sooner and faster.

For example, consider something like this:

> The API provides methods that let you perform partial tip pickups with a 96-channel pipette and configure Flex 50 µL pipettes for special low or high volume liquid handling.

My recommendation is the intro needs to include more info:
**Q:** What topics are going to be covered here?
**A:**  Partial tip pickup and volume modes.
Make that clear in the intro.

**Line 14**
> There are no mode configurations for OT-2 pipettes.

This is a great bit of information. It would be stronger if we could make a contrast with Flex earlier in this section. For example, back to line 7 (the intro sentence), mention "Flex pipettes" somewhere in either the original or any revisions as recommended above. This does 2 things:

- Adding "Flex" to line 7 clearly indicates to the reader, right away, that the subsequent documentation applies to Flex only. Anyone with an OT-2 can stop reading.
- Line 14 becomes a strong compare/contrast because we've already mentioned Flex. Right now, the contrast is implied or assumed.

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+        column_rack = protocol.load_labware(
+            load_name="opentrons_flex_96_tiprack_1000ul",
+            location="D3"
+        )
+        pipette = protocol.load_instrument("flex_96channel_1000")
+        pipette.configure_nozzle_layout(
+            style=COLUMN,
+            start="A12",
+            tip_racks=[column_rack]
+        )
```
jwwojak: This might be a moment where we should include the import statements. I know in other code examples we don't do this to reduce repetition. However, in this case, I think it would help to show the import statements in  context within a code sample.

api/docs/v2/pipettes/configuring.rst
ecormany: I put OT-2 at the end, because it has no modes…currently. But you're right, we should make the best docs now and make the best docs later, rather than hedging against future changes.

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+-------------
+
+Use the :py:meth:`.configure_nozzle_layout` method to choose how many tips the 96-channel pipette will pick up. The method's ``style`` parameter accepts special layout constants. When using partial tip pickup, it's generally easiest to import these at the top of your protocol.
+
+.. code-block:: python
+
+    from opentrons import protocol_api
+    from opentrons.protocol_api import COLUMN, ALL
+
+Then when you call ``configure_nozzle_layout`` later in your protocol, you can set ``style=COLUMN``. If, instead, you use ``from opentrons import protocol_api``, then you can pass ``protocol_api.COLUMN``. If you don't put any ``import`` statements at the top of your protocol, you won't be able to configure the nozzle layout.
```
jwwojak: Recommend making the required nature of the import statements stronger in the text.

> If you don't put any ``import`` statements at the top of your protocol, you won't be able to configure the nozzle layout.

Rather than "If you don't", try something that makes these more explicitly mandatory. For example: 

- You must put the `import` statements at the top of your protocol to configure the nozzle layout for partial tip pickup.
- Including these import statements are required to configure a 96-channel pipette nozzles for partial tip pickup.
- These import statements are required if you want to use partial tip pickup.

I'd also encourage the use of "partial tip pickup" more here and throughout. It is clearer and more descriptive than "nozzle layout." (at least to me)

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+    )
+    pipette.pick_up_tip()  # picks up full rack in C1
+
+
+Tip Pickup and Conflicts
+------------------------
+
+The horizontally offset position of the 96-channel pipette during partial tip pickup also places restrictions on where you can put other tall labware on the deck. 
+
+Using Column 12
```
jwwojak, line 129: If the information about column 12 doesn't have greater significance than the information about column 1, I'd recommend reordering this section in numeric order:

- Using Column 1
- Using Column 12

This seems like an instance where one block of information isn't more (or less) significant than another block of information. Put 1 before 12.

🎶  [1, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12](https://youtu.be/HUL4T8WcFdA?si=_5DnODcluz29_jBD) 🎶 

1970s TV for kids was really trippy.

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+            tip_racks=[column_rack]
+        )
+
+.. versionadded:: 2.16
+
+Let's unpack some of the details of this code.
+
+First, we've given a special name to the tip rack, ``column_rack``. You can name your tip racks whatever you like, but if you're performing full pickup and partial pickup in the same protocol, you'll need to keep them separate. See :ref:`partial-tip-rack-adapters` below.
+
+Next, we load the 96-channel pipette. Note that ``load_instrument()`` only has a single argument. The 96-channel pipette occupies both mounts, so ``mount`` is omissible. The ``tip_racks`` argument is always optional. But it would have no effect to declare it here, because every call to ``configure_nozzle_layout()`` resets the pipette's :py:obj:`.InstrumentContext.tip_racks` property.
```
jwwojak: On line 61, this is the first mention of `load_instrument()`. Could we make it a link to the method? I found myself wanting to review that method and that would have been easier with a link instead of a separate search, even abbreviated:

```
:py:meth:`~.ProtocolContext.load_instrument`
```

api/docs/v2/pipettes/configuring.rst
ecormany: 12 is fully supported. 1 is not. Hence the order.

Ideally these will flip someday!

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+.. warning::
+
+    :py:meth:`.InstrumentContext.pick_up_tip` always accepts a ``location`` argument, regardless of nozzle configuration. Do not pass a value that would lead the pipette to line up over more unused tips than specified by the current layout. For example, setting ``COLUMN`` layout and then calling ``pipette.pick_up_tip(tip_rack["A2"])`` on a full tip rack will lead to unexpected pipetting behavior and potential crashes.
+
+.. _partial-tip-rack-adapters:
+
+Tip Rack Adapters
+-----------------
+
+Partial tip pickup requires a tip rack that is placed directly in a deck slot. When picking up fewer than 96 tips, the 96-channel pipette lowers onto the tip rack in a horizontally offset position. If the tip rack were in the tip rack adapter, the pipette would collide with the adapter's posts, which protrude above the top of the tip rack. If you configure a partial nozzle layout and then call ``pick_up_tip()`` on a tip rack that's loaded onto an adapter, the API will raise an error.
```
jwwojak: Here I'd recommend getting the warning information into the first sentence and making it stronger and explicit. I've also flipped the last sentence around. Might work better without the "If <comma> then" structure (does this have a name?).

Something like this:

> Do not use a tip rack adapter when your protocol specifies partial tip pickup. Instead, place the tip rack on the deck. For partial tip pickup, the 96-channel pipette lowers itself onto the tip rack in a horizontally offset position from the top of the rack. If the tip rack were in its adapter, the pipette would collide with the mounting posts, which protrude above the top of the rack. The API will raise an error if you configure a partial pickup and then call ``pick_up_tip()`` on a tip rack that's loaded onto an adapter.

- You could even put emphasis on the _do not_ in that first sentence.

- I did remove "partial nozzle layout" later in the paragraph for "if you configure a partial pickup" (could be too casual though). Varying up the word choice is important. It seems like the phrase and concept "partial tip pickup" is the way to describe it, more than "nozzle layout."

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "C1")
+    pipette.configure_nozzle_layout(style=COLUMN, start="A1")
+    row_a = tip_rack.rows()[0]
+    pipette.pick_up_tip(row_a.pop())  # pick up A12-H12
+    pipette.drop_tip()
+    pipette.pick_up_tip(row_a.pop())  # pick up A11-H11
+    pipette.drop_tip()
+
+This code first constructs a list of all the wells in row A of the tip rack. Then, when picking up a tip, instead of referencing one of those wells directly, the ``location`` is set to ``row_a.pop()``. This uses the built-in :py:meth:`pop` method to get the last item from the list and remove it from the list. If you keep using this approach to pick up tips, you'll get an error once the tip rack is empty — not from the API, but from Python itself, since you're trying to ``pop`` an item from an empty list.
```
jwwojak, line 180: In sandbox, ```:py:meth: `pop` ``` isn't linking to the python docs. I didn't think this was a link until seeing it here in the source. I did a search and found it here: https://docs.python.org/3/tutorial/datastructures.html

I tried ```:py:meth:`list.pop` ``` too and can't get it to link. Something going on, human error, or is this too obscure for .rst?

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,243 @@
...
+.. _partial-tip-rack-adapters:
+
+Tip Rack Adapters
+-----------------
+
+Partial tip pickup requires a tip rack that is placed directly in a deck slot. When picking up fewer than 96 tips, the 96-channel pipette lowers onto the tip rack in a horizontally offset position. If the tip rack were in the tip rack adapter, the pipette would collide with the adapter's posts, which protrude above the top of the tip rack. If you configure a partial nozzle layout and then call ``pick_up_tip()`` on a tip rack that's loaded onto an adapter, the API will raise an error.
+
+On the other hand, full tip pickup requires the tip rack adapter. If the 96-channel pipette is in a full layout, either by default or by configuring ``style=ALL``, and you then call ``pick_up_tip()`` on a tip rack that's not in an adapter, the API will raise an error.
+
+If your protocol switches between full and partial pickup, you may want to organize your tip racks into lists, depending on whether they're loaded on adapters or not.
```
jwwojak: From line 89 to 124, reads like a related subsection topic change. I wonder if this needs it's own header.

Also, the ability to switch between full and partial pickup is impressive. I wonder if we can make the intro a little more punchy, more marketing-style enthusiasm to highlight this capability. Revise with some language like this:

- Partial tip pickup doesn't limit you to this mode only.
- Our pipette technology is flexible. (no pun intended here, reaches for thesaurus)
- You can use full and partial tip pickup in the same protocol!
- More text introducing the "here's how!" part.

Something with enthusiasm: "it slices, it dices, it even makes julienne fries!"

For your robot to become preeminent, [it's gotta have enthusiasms.](https://youtu.be/Z5u-H9OQT2c?si=hL_37znylfXSzMI8).

api/docs/v2/pipettes/configuring.rst
ecormany: This is a weird quirk of the Python standard library docs. It _doesn't have_ an entry for `list.pop()`. Just what you linked there, but that page has no internal anchors. I'm using the `:py:meth:` syntax here purely for formatting.

api/docs/v2/pipettes/configuring.rst
jwwojak: Related to a comment I left w/ Ed in Slack. I think this advice is good:

> Make a diagram and visualize everywhere the pipette will travel.

Makes me wonder if we can upload a blank deck image and make it available for users to grab and scribble on. Similar to the Flex deck slot image shown in the [deck slots doc](http://sandbox.docs.opentrons.com/docs-refactor-pipettes-partial-pickup/v2/deck_slots.html). 

Then, add some extra text, "To help you plan your layout, download this handy Flex deck map." Might be a nice touch to add. Help folks out sort of thing. Dunno.

api/docs/v2/new_pipette.rst
```diff
@@ -2,471 +2,25 @@
 
 .. _new-pipette:
 
-########
+********
 Pipettes
-########
+********
```
jwwojak, line 8: Super nice. Breaking up a very long doc into individual sections. Partial tip is significant enough to warrant it's own standalone subsection under the Pipettes heading.

api/docs/v2/pipettes/loading.rst
```diff
@@ -0,0 +1,171 @@
...
+Loading Pipettes
+****************
+
+When writing a protocol, you must inform the Protocol API about the pipettes you will be using on your robot. The :py:meth:`.ProtocolContext.load_instrument` function provides this information and returns an :py:class:`.InstrumentContext` object.
+
+As noted above, you call the :py:meth:`~.ProtocolContext.load_instrument` method to load a pipette. This method also requires the :ref:`pipette's API load name <new-pipette-models>`, its left or right mount position, and (optionally) a list of associated tip racks. Even if you don't use the pipette anywhere else in your protocol, the Opentrons App and the robot won't let you start the protocol run until all pipettes loaded by ``load_instrument()`` are attached properly.
+
+.. _new-pipette-models:
+
+API Load Names
```
jwwojak, line 14: Nice move. This puts the load names up front. 👏

api/docs/v2/pipettes/loading.rst
```diff
@@ -0,0 +1,171 @@
...
+        right = protocol.load_instrument(
+            instrument_name='p300_multi_gen2',
+            mount='right',
+            tip_racks=[tiprack1])
+
+.. versionadded:: 2.0
+
+.. _pipette-tip-racks:
+
+Adding Tip Racks
```
jwwojak, line 131: This is a change too, moving under Loading Pipettes. 👏

jwwojak (commented):
Added my review comments. Please take a look. Partial tip pickup is an impressive feature, difficult to document.

api/docs/v2/pipettes/configuring.rst
jwwojak: 🤔  I see. Thanks for the explanation. Weird for sure.

api/docs/v2/pipettes/configuring.rst
ecormany: Integrated some of this, including putting "Do not" first. I don't think it needs further emphasis. The worst that happens here is your protocol fails to analyze.

api/docs/v2/pipettes/configuring.rst
ecormany: 🎟️ ☔

ecormany: Noted!

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,193 @@
...
+******************
+Partial Tip Pickup
+******************
+
+The 96-channel pipette occupies both pipette mounts on Flex, so it's not possible to attach another pipette at the same time. Partial tip pickup lets you perform some of the same actions that you would be able to perform with a second pipette. As of version 2.16 of the API, you can configure the 96-channel pipette to pick up a single column of tips, similar to the behavior of an 8-channel pipette.
+
+Nozzle Layout
+=============
+
+Use the :py:meth:`.configure_nozzle_layout` method to choose how many tips the 96-channel pipette will pick up. The method's ``style`` parameter accepts special layout constants. You must import these constants at the top of your protocol, or you won't be able to configure the pipette for partial tip pickup.
```
jwwojak, line 12: One comment here and it refers to the `COLUMN` argument definition for `configure_nozzle_layout`, not the text on this page. I meant to include it yesterday, but it was buried in my notes and overlooked. This part of the method definition, on first read, is confusing:

> sets the pipette to use 8 nozzles, aligned from front to back with respect to the deck. This corresponds to a column of wells on labware.

My first reaction to "aligned from front to back with respect to the deck" was a big ❓ Pipette barrels or channels don't have their own coordinates so it might help with clarity to use well coordinates in that last sentence to clarify. For example:

> This corresponds to a column of wells on labware (e.g. A1-H1, A2-H2, A3-H3, etc.).

Maybe I'm just dumb, but I had to read the argument definition a few times before coming to what I hoped is an understanding of what it means.

api/docs/v2/pipettes/configuring.rst
```diff
@@ -0,0 +1,193 @@
...
+
+Next, we load the 96-channel pipette. Note that :py:meth:`.load_instrument` only has a single argument. The 96-channel pipette occupies both mounts, so ``mount`` is omissible. The ``tip_racks`` argument is always optional. But it would have no effect to declare it here, because every call to ``configure_nozzle_layout()`` resets the pipette's :py:obj:`.InstrumentContext.tip_racks` property.
+
+Finally, we configure the nozzle layout, with three arguments.
+
+    - The ``style`` parameter directly accepts the ``COLUMN`` constant, since we imported it at the top of the protocol.
+    - The ``start`` parameter accepts a nozzle name, representing the back-left nozzle in the layout, as a string. ``"A12"`` tells the pipette to use its rightmost column of nozzles for pipetting.
+    - The ``tip_racks`` parameter tells the pipette which racks to use for tip tracking, just like :ref:`adding tip racks <pipette-tip-racks>` when loading a pipette.
```
jwwojak, line 60: Great use of bullets here.  🥇

jwwojak (approved):
This looks good. Only one comment and it relates to the description for a method's argument and not the text in this PR. 

:shipit:

api/docs/v2/pipettes/configuring.rst
ecormany: Yeah, that was a tough definition to write. "It's a column, you know, like a _column!_"

Going to leave it for now, because that text will be revised as soon as additional layouts are added.

PR #14192, opened 2023-12-13
fix(app): fix module calibration wizard copy

koji:
# Overview
The ticket mentions to h/s only but there are a couple of diffs between the current implementation and the design.
https://www.figma.com/file/0kuPeCi1t2Auu2GPMnqRb2/Primary%3A-Opentrons-Flex-Touchscreen?type=design&node-id=6549-87111&mode=design&t=3Hrx96amCPY0juuO-0

From Sue, there are a couple of variations about on this wizard step.

### For header text
module type is not heater-shaker
`Place calibration adapter in <module>`

module type is heater-shaker
`Install calibration adapter`

### For body text
module type is h/s
`Place the adapter flush on the top of the module. Secure the adapter to the module with a thermal adapter screw and T10 Torx screwdriver.`

module type is TC
`Ensure the Thermocycler lid is open and place the adapter flush on top of the module where the labware would normally go.`

module type is not h/s or TC
`Place the adapter flush on top of the module.`

close RQA-1935

# Test Plan

# Changelog
- add new text to json file
- add conditions to handle the above cases

# Review requests

# Risk assessment
low


jerader (approved):
lgtm!

PR #14193, opened 2023-12-13
docs(api): Correct and tidy up docs for ProtocolContext.fixed_trash and InstrumentContext.trash_container

SyntaxColoring:
# Overview

Corrections and clarifications in the reference documentation for `ProtocolContext.fixed_trash` and `InstrumentContext.trash_container`.

# Test Plan

Preview:

* [`ProtocolContext.fixed_trash`](http://sandbox.docs.opentrons.com/docs_cleanup_fixed_trash_and_trash_container/v2/new_protocol_api.html#opentrons.protocol_api.ProtocolContext.fixed_trash)
* [`InstrumentContext.trash_container`](http://sandbox.docs.opentrons.com/docs_cleanup_fixed_trash_and_trash_container/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.trash_container)

# Changelog

See my inline notes.

# Review requests

* Does this harmonize well with http://sandbox.docs.opentrons.com/docs_cleanup_fixed_trash_and_trash_container/v2/deck_slots.html#deck-configuration? Are we duplicating anything or spreading any information out?
* I'm adding cross-references to the `TrashBin` class, which we have prepared to be public but which is not yet actually published. My vision is that we will publish it; it is a public, stable part of the API just like `Labware`. Does that vision sound good?

# Risk assessment

No risk. Docs only.

PR #14194, opened 2023-12-13
docs(api): corrections to docstrings involving `TrashBin`s

ecormany:
# Overview

Updates the docstrings for `fixed_trash` and `trash_container` to accurately reflect how they work with different robot types and API versions.

# Test Plan

Check API reference entries.
- [`fixed_trash`](http://sandbox.docs.opentrons.com/docstring-trashbin-fixes/v2/new_protocol_api.html#opentrons.protocol_api.ProtocolContext.fixed_trash)
- [`trash_container`](http://sandbox.docs.opentrons.com/docstring-trashbin-fixes/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.trash_container)

# Changelog

- Two docstrings mentioned above.
- `drop_tip()` doesn't go to the "default" `trash_container` — it goes to the current value.

# Review requests

I love it because it's trash?

# Risk assessment

v low, docstrings only

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -945,7 +945,7 @@ class InstrumentContext(publisher.CommandPublisher):
         See :ref:`pipette-drop-tip` for examples.
 
         If no location is passed (e.g. ``pipette.drop_tip()``), the pipette will drop
-        the attached tip into its default :py:attr:`trash_container`.
+        the attached tip into its :py:attr:`trash_container`.
 
         Starting with API version 2.15, if the trash container is the default fixed
         trash, the API will instruct the pipette to drop tips in different locations
@@ -1524,12 +1524,14 @@ class InstrumentContext(publisher.CommandPublisher):
         This is the property used to determine where to drop tips and blow out liquids
         when calling :py:meth:`drop_tip` or :py:meth:`blow_out` without arguments.
 
-        On a Flex running a protocol with API version 2.16 or higher, ``trash_container`` is
-        the first ``TrashBin`` or ``WasteChute`` object loaded in the protocol.
-        On a Flex running a protocol with API version 2.15, ``trash_container`` is
-        a single-well fixed trash labware in slot D3.
-        On a an OT-2, ``trash_container`` is always a single-well fixed trash labware
-        in slot 12.
+        The default value depends on the robot type and API version:
+
+        - On a Flex running a protocol with API version 2.16 and later, the first
+          ``TrashBin`` or ``WasteChute`` object loaded in the protocol.
+        - On a Flex running a protocol with API version 2.15, a single-well fixed trash
+          ``Labware`` in slot D3.
+        - On an OT-2, it is always in slot 12. It is a ``Labware`` in API version 2.15
+          or earlier, or a ``TrashBin`` in API version 2.16 and later.
 
         .. versionchanged:: 2.16
             Added support for ``TrashBin`` and ``WasteChute`` objects.
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1524,6 +1524,8 @@ class InstrumentContext(publisher.CommandPublisher):
         This is the property used to determine where to drop tips and blow out liquids
         when calling :py:meth:`drop_tip` or :py:meth:`blow_out` without arguments.
 
+        You can set this to a :py:obj:`Labware`, :py:obj:`TrashBin`, or :py:obj:`WasteChute`.
+
         The default value depends on the robot type and API version:
 
         - On a Flex running a protocol with API version 2.16 and later, the first
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1528,12 +1528,10 @@ class InstrumentContext(publisher.CommandPublisher):
 
         The default value depends on the robot type and API version:
 
-        - On a Flex running a protocol with API version 2.16 and later, the first
-          ``TrashBin`` or ``WasteChute`` object loaded in the protocol.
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1524,12 +1524,14 @@ def trash_container(self) -> Union[labware.Labware, TrashBin, WasteChute]:
         This is the property used to determine where to drop tips and blow out liquids
         when calling :py:meth:`drop_tip` or :py:meth:`blow_out` without arguments.
 
-        On a Flex running a protocol with API version 2.16 or higher, ``trash_container`` is
-        the first ``TrashBin`` or ``WasteChute`` object loaded in the protocol.
-        On a Flex running a protocol with API version 2.15, ``trash_container`` is
-        a single-well fixed trash labware in slot D3.
-        On a an OT-2, ``trash_container`` is always a single-well fixed trash labware
-        in slot 12.
+        The default value depends on the robot type and API version:
```
SyntaxColoring, line 1529: Up to you, but we may want to explicitly say that this is overridable. I don't think Sphinx renders any distinction between read-only properties and writable ones.

```suggestion
        You can set this to a :py:obj:`Labware`, :py:obj:`TrashBin`, or :py:obj:`WasteChute`.

        The default value depends on the robot type and API version:
```

Note that I'm cross-referencing to `WasteChute` and `TrashBin` here, which aren't currently published, so the links won't go anywhere. I think we should publish them, though; they're like `Labware`.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1524,12 +1524,14 @@ def trash_container(self) -> Union[labware.Labware, TrashBin, WasteChute]:
...
+
+        - On a Flex running a protocol with API version 2.16 and later, the first
+          ``TrashBin`` or ``WasteChute`` object loaded in the protocol.
+        - On a Flex running a protocol with API version 2.15, a single-well fixed trash
+          ``Labware`` in slot D3.
+        - On an OT-2, it is always in slot 12. It is a ``Labware`` in API version 2.15
+          or earlier, or a ``TrashBin`` in API version 2.16 and later.
```
SyntaxColoring: I think we can simplify this by referring to `InstrumentContext.fixed_trash`.

Either something like this:

```suggestion
        - :py:obj:`ProtocolContext.fixed_trash`, if it exists.
        - Otherwise, the first thing that you've loaded so far with
          :py:obj:`ProtocolContext.load_trash_bin()`
          or :py:obj:`ProtocolContext.load_waste_chute()`.
```

Or, equivalently:

```suggestion
        - On a Flex running a protocol with API version 2.16 and later, the first
          :py:obj:`TrashBin` or :py:obj:`WasteChute` object loaded in the protocol.
        - Otherwise, :py:obj:`ProtocolContext.fixed_trash`.
```

Also, instead of saying "the first `TrashBin` or `WasteChute` loaded in the protocol," it might be more directly relevant to the reader if we linked to `ProtocolContext.load_trash_bin()` and `ProtocolContext.load_waste_chute()`.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1052,12 +1052,16 @@ def deck(self) -> Deck:
...
-        e.g. ``protocol.fixed_trash['A1']``
+        In API version 2.15 and earlier, the fixed trash is a :py:class:`.Labware` object with one well. Access it like labware in your protocol. For example, ``protocol.fixed_trash['A1']``.
 
-        In API Version 2.16 and above it returns a Trash fixture for OT-2 Protocols.
+        In API version 2.15 only, Flex protocols have a fixed trash in slot A3.
+
+        In API version 2.16 and later, the fixed trash only exists in OT-2 protocols. It is a :py:class:`.TrashBin` object, which doesn't have any wells. Trying to access ``fixed_trash`` in a Flex protocol will raise an error.
+
+        .. versionchanged:: 2.16
+            Returns a :py:class:`.TrashBin` object.
```
SyntaxColoring: This looks good to me, but:

* See https://github.com/Opentrons/opentrons/pull/14193/files if any of my alternative phrasing over there is helpful.
* I would consider squeezing in a reference to <code>:ref:\`configure-trash-bin\`</code>.

SyntaxColoring (approved):
LGTM, thanks!

I think we can simplify `InstrumentContext.trash_container` to take advantage of existing abstractions. Other than that, just some minor suggestions.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -945,7 +945,7 @@ def drop_tip(
         See :ref:`pipette-drop-tip` for examples.
 
         If no location is passed (e.g. ``pipette.drop_tip()``), the pipette will drop
-        the attached tip into its default :py:attr:`trash_container`.
+        the attached tip into its :py:attr:`trash_container`.
```
SyntaxColoring, line 948: Nice catch.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: I've put a few of those cross-refs in too. They get the bold monospace formatting, which is visually helpful, even though they don't link anywhere.

api/src/opentrons/protocol_api/instrument_context.py
ecormany: So like this? 
```
On a Flex running a protocol with API version 2.16 and later, the object returned 
by the first call to :py:meth:`.load_trash_bin` or :py:meth:`.load_waste_chute`.
```

SyntaxColoring: Yep that works. I had "**the first thing you loaded with** :py:meth:`.load_trash_bin` or :py:meth:`.load_waste_chute`" [emphasis added] in my draft if that reads better to you, but your call.

jwwojak (commented):
The text here all seems fine to me. No additional comments from this side of the peanut gallery. 👍

PR #14196, opened 2023-12-13
feat(engine, api): alternate tip drop location for addressable area trash bins

jbleon95:
# Overview

Addresses RSS-412

This PR adds tip drop alternation back into the Protocol Engine, for both OT-2 and Flex, for both column 1 and column 3 trashes. A new `moveToAddressableAreaForDropTip` command was added to encapsulate the logic in protocol engine for cycling between a right bias and a left bias.

In order to make the deck points the pipette critical points move to accurate, the `boundingBox` and `offsetFromCutoutFixture` values needed to be changed for the trash bin addressable areas. The previous equation for a right mount moving to the left bias for a column 1 trash needed to be adjusted since the current logic was moving it too far to the right.

# Test Plan

Extensively tested all combinations of columns, mounts, and 96 channel both with full configuration and partial tip configuration (only column in A12) to ensure that they go to the correct place, do no collide with anything, and cycle accurately through the locations. Most of this was done in postman, but this protocol was run for the partial tip drop test.

```
from opentrons.protocol_api import COLUMN


metadata = {
    'protocolName': '96 Channel Partial Tip Alternation',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}


def run(context):
    tiprack = context.load_labware('opentrons_flex_96_tiprack_1000ul', 'C2')

    pipette = context.load_instrument('flex_96channel_1000', 'left')
    pipette.configure_nozzle_layout(style=COLUMN, start="A12")

    trash_b3 = context.load_trash_bin('B3')
    trash_d1 = context.load_trash_bin('D1')

    pipette.pick_up_tip(tiprack.wells()[0])
    pipette.drop_tip(trash_b3)
    context.pause()
    pipette.reset_tipracks()

    pipette.pick_up_tip(tiprack.wells()[0])
    pipette.drop_tip(trash_b3)
    context.pause()
    pipette.reset_tipracks()

    pipette.pick_up_tip(tiprack.wells()[0])
    pipette.drop_tip(trash_b3)
    context.pause()
    pipette.reset_tipracks()

    pipette.pick_up_tip(tiprack.wells()[0])
    pipette.drop_tip(trash_d1)
    context.pause()
    pipette.reset_tipracks()

    pipette.pick_up_tip(tiprack.wells()[0])
    pipette.drop_tip(trash_d1)
    context.pause()
    pipette.reset_tipracks()

    pipette.pick_up_tip(tiprack.wells()[0])
    pipette.drop_tip(trash_d1)
    context.pause()
    pipette.reset_tipracks()
```

# Changelog

- added `moveToAddressableAreaForDropTip` PE command
- hooked up PAPI drop tip for trash bins to use the new command
- modified boundingBox` and `offsetFromCutoutFixture` values for trash bins in V4 deck definitions
- modified the equation for a right mount single/multi channel pipette accessing the left tip drop location of a column 1 trash bin to be less conservative in its leftward offset.

# Review requests

# Risk assessment

Medium.


shared-data/command/schemas/8.json
```diff
@@ -1996,6 +1999,90 @@
...
+      "description": "Payload required to move a pipette to a specific addressable area.\n\nAn *addressable area* is a space in the robot that may or may not be usable depending on how\nthe robot's deck is configured. For example, if a Flex is configured with a waste chute, it will\nhave additional addressable areas representing the opening of the waste chute, where tips and\nlabware can be dropped.\n\nThis moves the pipette so all of its nozzles are centered over the addressable area.\nIf the pipette is currently configured with a partial tip layout, this centering is over all\nthe pipette's physical nozzles, not just the nozzles that are active.\n\nThe z-position will be chosen to put the bottom of the tips---or the bottom of the nozzles,\nif there are no tips---level with the top of the addressable area.\n\nWhen this command is executed, Protocol Engine will make sure the robot's deck is configured\nsuch that the requested addressable area actually exists. For example, if you request\nthe addressable area B4, it will make sure the robot is set up with a B3/B4 staging area slot.\nIf that's not the case, the command will fail.",
```
jerader, line 2004: this example with the waste chute is a bit confusing to me since this command is for trash bins and not waste chute right?

jerader (approved):
lgtm!

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -487,6 +487,16 @@ def _move_to_disposal_location(
...
         if isinstance(disposal_location, TrashBin):
             addressable_area_name = disposal_location._addressable_area_name
+            self._engine_client.move_to_addressable_area_for_drop_tip(
+                pipette_id=self._pipette_id,
+                addressable_area_name=addressable_area_name,
+                offset=offset,
+                force_direct=force_direct,
+                speed=speed,
+                minimum_z_height=None,
+                alternate_drop_location=True,
```
SyntaxColoring: Discussed in review: let's pass `False` here when the PAPI author does `blow_out()` or `dispense()`.

shared-data/command/schemas/8.json
jbleon95: Yeah, I need to fix this up, but I can do that in a quick doc follow up tomorrow

CaseyBatten (approved):
This looks good with the noted change that _move_to_dispoal_locations needs to only alternate tip drops for a drop trip action and should ignore that for things like blowout/dispense/etc. We should take a look at how this relates to the transfer behaviors automatic drop tip behavior at the end of a given transfer. AKA:
If you have two transfers in a row, does it drop the tip in the same location twice?

Otherwise everything looks good to me.

api/src/opentrons/protocol_engine/commands/move_to_addressable_area_for_drop_tip.py
```diff
@@ -0,0 +1,149 @@
...
+
+if TYPE_CHECKING:
+    from ..execution import MovementHandler
+    from ..state import StateView
+
+MoveToAddressableAreaForDropTipCommandType = Literal["moveToAddressableAreaForDropTip"]
+
+
+class MoveToAddressableAreaForDropTipParams(PipetteIdMixin, MovementMixin):
+    """Payload required to move a pipette to a specific addressable area.
```
SyntaxColoring, line 25: We could probably nuke this docstring and say something like "this is like `moveToAddressableArea` except it has tip alternation.

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -804,7 +858,7 @@ def _get_drop_tip_well_x_offset(
             ):
                 # Pipette might not reach the default left spot so use a different left spot
                 x_well_offset = (
-                    well_x_dim / 2 - SLOT_WIDTH + drop_location_margin_from_labware_edge
+                    -well_x_dim / 2 + drop_location_margin_from_labware_edge * 2
```
SyntaxColoring, line 861: Yeah this makes sense if the intent is to double the margin from the left edge.

I think this is another really good opportunity for motion planning based on high-fidelity pipette dimensions and deck limits. ("Go as far left as you can without bumping into the frame.")

SyntaxColoring (approved):
Reviewed on a call. Looks great, thanks!

PR #14203, opened 2023-12-14
refactor(protocol-designer, step-generation): emit dropTipInPlace & m…

jerader:
…oveToAddressableAreaForDropTip

# Overview

I made a big oopsy and was only emitting `moveToAddressableAreaForDropTip` without the `dropTipInPlace` afterwards! Luckily, since those sequence of commands are consolidated in `movableTrashCommandsUtil`, it was an easy fix. I also had to update all the cypress migration tests again 🤦 

# Test Plan

Create a flex or ot-2 protocol in PD and drop tip into the trash bin. It can be a simple transfer/mix step. Download the protocol and upload it to the app. It should work in alpha.7 or the dev env built off of this branch. It should pass analysis.

# Changelog

- emit dropTipInPlace in the `movableTrashCommandsUtil` and update the test
- delete now unnecessary robot state update fn for `moveToAddressableAreaForDropTip` 
- fix cypress test fixtures

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/fixtures/protocol/8/doItAllV3MigratedToV8.json
+++ b/protocol-designer/fixtures/protocol/8/doItAllV3MigratedToV8.json
@@ -2783,6 +2783,11 @@
       }
     },
     {
+      "commandType": "dropTipInPlace",
+      "key": "814036fd-ea55-40d0-932b-d13f6113a77f",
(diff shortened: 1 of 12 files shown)


koji (approved):
Followed the test plan and it worked as expected. (OT-2/Flex both protocols passed analysis)

PR #14205, opened 2023-12-14
fix(shared-data): add moveToAddressableAreaForDropTip to getAddressableAreasInProtocol

brenthagen:
# Overview

adds the moveToAddressableAreaForDropTip command to the getAddressableAreasInProtocol helper that provides a list of addressable area names referenced in a protocol's commands

# Test Plan

# Changelog

 - Adds moveToAddressableAreaForDropTip to getAddressableAreasInProtocol

# Review requests

# Risk assessment

low

### DIFF
--- a/shared-data/js/helpers/getAddressableAreasInProtocol.ts
+++ b/shared-data/js/helpers/getAddressableAreasInProtocol.ts
@@ -75,6 +75,14 @@ export function getAddressableAreasInProtocol(
           ...acc,
           command.params.addressableAreaName as AddressableAreaName,
         ]
+      } else if (
+        command.commandType === 'moveToAddressableAreaForDropTip' &&
+        !acc.includes(command.params.addressableAreaName as AddressableAreaName)
+      ) {
+        return [
+          ...acc,
+          command.params.addressableAreaName as AddressableAreaName,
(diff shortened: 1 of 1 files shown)


shared-data/js/helpers/getAddressableAreasInProtocol.ts
```diff
@@ -75,6 +75,14 @@ export function getAddressableAreasInProtocol(
           ...acc,
           command.params.addressableAreaName as AddressableAreaName,
         ]
+      } else if (
+        command.commandType === 'moveToAddressableAreaForDropTip' &&
```
jerader, line 79: does it work to add this to line 71 and do a `command.commandType === 'moveToAddressableArea' || command.commandType === 'moveToAddressableAreaForDropTip'`?

jerader (approved):
logic make sense to me! ty for fixing this!

shared-data/js/helpers/getAddressableAreasInProtocol.ts
brenthagen: yeah that would work. i think it's a little clearer as a separate block and keeps the two commands independent if their shapes ever change

PR #14206, opened 2023-12-14
Rh 22 use partial tip pickup

ryanthecoder:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/hardware_control/backends/ot3simulator.py
```diff
@@ -476,13 +476,22 @@ def _attached_pipette_to_mount(
                     ),
                     "id": None,
                 }
-        if found_model and expected_instr or found_model:
+        if (found_model and expected_instr or found_model) and init_instr[
```
caila-marashaj: can this just be `if found_model and init_instr["id"] is not None` ?

PR #14210, opened 2023-12-15
fix(app): get addressable area from configured deck for drop tip wizard

brenthagen:
# Overview

translates a selected "slot" from the deck location select single slot representation to an addressable area compatible with the configured deck and selected pipette.

closes RQA-2099

# Test Plan

verified correct deck location select addressable area name output for a variety of flex configurations and OT-2

# Changelog

 - get addressable area from configured deck for drop tip wizard

# Review requests

check drop tip wizard for Flex trash bins and waste chutes. check OT-2 for parity.

# Risk assessment

low

### DIFF
--- a/app/src/organisms/DropTipWizard/getAddressableAreaFromConfig.ts
+++ b/app/src/organisms/DropTipWizard/getAddressableAreaFromConfig.ts
@@ -1,0 +1,87 @@
+import {
+  EIGHT_CHANNEL_WASTE_CHUTE_ADDRESSABLE_AREA,
+  getCutoutIdForAddressableArea,
+  getDeckDefFromRobotType,
+  MOVABLE_TRASH_ADDRESSABLE_AREAS,
+  NINETY_SIX_CHANNEL_WASTE_CHUTE_ADDRESSABLE_AREA,
+  ONE_CHANNEL_WASTE_CHUTE_ADDRESSABLE_AREA,
+  WASTE_CHUTE_ADDRESSABLE_AREAS,
+} from '@opentrons/shared-data'
+
+import type {
+  AddressableAreaName,
+  DeckConfiguration,
+  RobotType,
+} from '@opentrons/shared-data'
+
+export function getAddressableAreaFromConfig(
+  addressableArea: AddressableAreaName,
+  deckConfig: DeckConfiguration,
+  pipetteChannels: number,
+  robotType: RobotType
+): AddressableAreaName | null {
+  const deckDef = getDeckDefFromRobotType(robotType)
+
+  let addressableAreaFromConfig: AddressableAreaName | null = null
+
+  // match the cutout id to aa
+  const cutoutIdForAddressableArea = getCutoutIdForAddressableArea(
+    addressableArea,
+    deckDef.cutoutFixtures
+  )
+
+  // get addressable areas provided by current deck config
+  const configuredCutoutFixture =
+    deckConfig.find(fixture => fixture.cutoutId === cutoutIdForAddressableArea)
+      ?.cutoutFixtureId ?? null
+
+  const providedAddressableAreas =
+    cutoutIdForAddressableArea != null
+      ? deckDef.cutoutFixtures.find(
+          fixture => fixture.id === configuredCutoutFixture
+        )?.providesAddressableAreas[cutoutIdForAddressableArea] ?? []
+      : []
+
+  // check if configured cutout fixture id provides target addressableArea
+  if (providedAddressableAreas.includes(addressableArea)) {
+    addressableAreaFromConfig = addressableArea
+  } else if (
+    // if no, check if provides a movable trash
+    providedAddressableAreas.some(aa =>
+      MOVABLE_TRASH_ADDRESSABLE_AREAS.includes(aa)
+    )
+  ) {
+    addressableAreaFromConfig = providedAddressableAreas[0]
+  } else if (
+    // if no, check if provides waste chute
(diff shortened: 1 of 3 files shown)


app/src/organisms/DropTipWizard/getAddressableAreaFromConfig.ts
```diff
@@ -0,0 +1,87 @@
...
+      pipetteChannels === 8 &&
+      providedAddressableAreas.includes(
+        EIGHT_CHANNEL_WASTE_CHUTE_ADDRESSABLE_AREA
+      )
+    ) {
+      addressableAreaFromConfig = EIGHT_CHANNEL_WASTE_CHUTE_ADDRESSABLE_AREA
+    } else if (
+      pipetteChannels === 96 &&
+      providedAddressableAreas.includes(
+        NINETY_SIX_CHANNEL_WASTE_CHUTE_ADDRESSABLE_AREA
```
mjhuff, line 79: This condition actaully doesn't get triggered, because providedAddressableAreas isn't returning 96ChannelWasteChute - it's returning 1ChannelWasteChute and 8ChannelWasteChute when a 96ch is loaded.

I don't think this is a bug here, but elsewhere. I can take a look at it in a bit, but if you happen to find it before me, let me know!

Screenshot 2023-12-15 at 9 09 05 AM

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -347,14 +357,25 @@ export const DropTipWizardComponent = (
...
+        const addressableAreaFromConfig = getAddressableAreaFromConfig(
+          addressableArea,
+          deckConfig,
+          instrumentModelSpecs.channels,
+          robotType
+        )
+
+        if (
+          currentPosition != null &&
+          (robotType !== FLEX_ROBOT_TYPE || addressableAreaFromConfig != null)
```
mjhuff:
```suggestion
          (currentPosition != null && addressableAreaFromConfig != null)
```
I'm guessing this is just a typo? We want to run the command on flexes as well (and it does work just fine if we remove this check!)

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -347,14 +357,25 @@ export const DropTipWizardComponent = (
...
+        ) {
           return createRunCommand({
             maintenanceRunId: createdMaintenanceRunId,
             command: {
               commandType: 'moveToAddressableArea',
               params: {
                 pipetteId: MANAGED_PIPETTE_ID,
-                addressableAreaName: addressableArea,
+                addressableAreaName:
+                  addressableAreaFromConfig ?? addressableArea,
```
mjhuff:
```suggestion
                  addressableAreaFromConfig,
```

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -347,14 +357,25 @@ export const DropTipWizardComponent = (
...
     return retractAllAxesAndSavePosition()
       .then(currentPosition => {
-        if (currentPosition != null) {
+        const addressableAreaFromConfig = getAddressableAreaFromConfig(
+          addressableArea,
+          deckConfig,
+          instrumentModelSpecs.channels,
+          robotType
+        )
+
```
mjhuff, line 365: ` const zOffset =
          addressableAreaFromConfig === addressableArea
            ? (currentPosition as Coordinates).z - 10
            : 0`

After doing some testing, we need to add a zOffset conditional. This is just due to how the moveToAddressableCommand works - it lowers the Z axis way too much for drop tip comfort when moving to a slot that isn't loaded with a waste chute/trash.

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -347,14 +357,25 @@ export const DropTipWizardComponent = (
...
           return createRunCommand({
             maintenanceRunId: createdMaintenanceRunId,
             command: {
               commandType: 'moveToAddressableArea',
               params: {
                 pipetteId: MANAGED_PIPETTE_ID,
-                addressableAreaName: addressableArea,
+                addressableAreaName:
+                  addressableAreaFromConfig ?? addressableArea,
                 offset: { x: 0, y: 0, z: currentPosition.z - 10 },
```
mjhuff: ` offset: { x: 0, y: 0, z: zOffset },`

app/src/organisms/DropTipWizard/index.tsx
mjhuff, line 1: On L384, instead of returning null, let's return:
`setErrorMessage(
            `error moving to position: cannot move to addressable area.`
          )`

moveToAddressable area doesn't throw errors - it just returns error messages. We probably want to throw an error here on the frontend if we can't move.

mjhuff (commented):
Thanks for doing this so quickly! See comments. The most important thing to address is the 96ch waste chute issue.

app/src/organisms/DropTipWizard/index.tsx
brenthagen: `robotType !== FLEX_ROBOT_TYPE` bypasses the `addressableAreaFromConfig` check for OT-2. the intent is to be conservative with OT-2 since deck config is a Flex feature. the getAddressableAreaFromConfig helper should be irrelevant to OT-2 since there's a single configuration.

if it's cleaner, we can remove the robot type check. at least on 7.1 I confirmed that OT-2 returns a deck config that is compatible with the output of deck location select.

mjhuff (approved):
Looks good! Thank you again!

PR #14211, opened 2023-12-15
fix(api): allow volume 0 commands in engine

DerekMaggio:
A previous PR (12a630b / #13989 ) changed the python protocol api in version 2.16 to allow commanding 0ul liquid handling commands like aspirate, mix, and dispense. This is useful in programmatic protocols that read out volumes to handle; they can now handle 0 volume properly. In 2.15 and previous, specifying 0 would lead to those commands doing the most volume they could (i.e. aspirate the full tip volume, dispense whatever's currently in the pipette, mix at full volume) and this likely was an unintentional side effect because of python truthiness.

However, that change only touched the python protocol API; that API would emit commands to the engine that specified 0 volume, and those were not allowed: the pydantic models for the commands and responses required strictly greater than 0 volume.

This PR
- changes the pydantic models and updates the schema to allow 0ul commands
- adds a python protocol to be an integration test
- adds unit tests for the python protocol api aspirate and dispense commands

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/pipetting_common.py
+++ b/api/src/opentrons/protocol_engine/commands/pipetting_common.py
@@ -19,7 +19,7 @@ class VolumeMixin(BaseModel):
 
     volume: float = Field(
         ...,
-        description="Amount of liquid in uL. Must be greater than 0 and less "
+        description="Amount of liquid in uL. Must be at least 0 and no greater "
         "than a pipette-specific maximum volume.",
         ge=0,
     )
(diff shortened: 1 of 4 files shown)


SyntaxColoring (approved):
Tested with this protocol. Works as expected. Thank you!

```python


def run(protocol):
    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "D1")
    well_plate_1 = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "D2"
    )
    well_plate_2 = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "C2"
    )
    )

    pipette.pick_up_tip()

    protocol.comment(str(well_plate_1.wells()))

    for source, dest in zip(well_plate_1.wells(), well_plate_2.wells()):
        pipette.aspirate(0, source)
        pipette.dispense(0, dest)

    pipette.return_tip()
```

PR #14212, opened 2023-12-15
refactor(app): Remove drop tip banner after OT-2 protocol run

mjhuff:
Closes RQA-2106

# Overview

Because OT-2s implicitly drop tips at the end of the run, there's no need to render the drop tip banner now. This PR also adds more error handling to the moveToAddressableArea command during drop tip flows, and adds reasonable display text when flows start. Also fixes bugs where the drop tip banner unrenders during run cancel and the OT-2 can't move to fixedTrash. Also fixes an issue where the confirm select location step would flash after selecting a location.

# Test Plan
- Run through flows on OT-2 that pick up a tip. Cancel a run after a tip has been picked up. Note that no banner displays. 
- Confirm banner still displays properly on the Flex.
- Observe the start up text when launching a flow.

# Changelog
- OT-2 no longer displays the drop tip banner after a protocol run is in a terminal state.

# Risk assessment
low

### DIFF
--- a/app/src/assets/localization/en/drop_tip_wizard.json
+++ b/app/src/assets/localization/en/drop_tip_wizard.json
@@ -12,7 +12,7 @@
   "drop_tips": "drop tips",
   "error_dropping_tips": "Error dropping tips",
   "exit_screen_title": "Exit before completing drop tip?",
-  "getting_ready": "Getting ready...",
+  "getting_ready": "Getting ready…",
   "go_back": "go back",
   "move_to_slot": "move to slot",
   "no_proceed_to_drop_tip": "No, proceed to tip removal",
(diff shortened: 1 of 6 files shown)


app/src/assets/localization/en/drop_tip_wizard.json
```diff
@@ -12,6 +12,7 @@
   "drop_tips": "drop tips",
   "error_dropping_tips": "Error dropping tips",
   "exit_screen_title": "Exit before completing drop tip?",
+  "getting_ready": "Getting ready...",
```
ecormany: i love an ellipsis character `…` instead of `...` but :shipit: 

literally no one else will notice 😅

mjhuff: Haha, fair - I'll add this!

brenthagen (approved):
changes make sense, drop tip choose location smoke tested on Flex/OT-2

mjhuff:
> changes make sense, drop tip choose location smoke tested on Flex/OT-2

Thanks for all the help! 🔥

PR #14213, opened 2023-12-15
ci(shared-data): fix publish switch if condition

b-cooper:
Added missing spaces before brackets in condition.


sfoster1 (approved):
oh. yup yup

PR #14214, opened 2023-12-15
fix(app-shell): use a wrapping stream for usb

sfoster1:
Implement the tcp socket emulator by creating a separate stream that is piped through to the usb connection via its downward interface rather than using the same stream for both the port and the socket or pipelining the two streams together.

This implementation allows the lifecycles of the USB port connection, which we want to be the same as the physical robot connection; and the tcp socket emulators, which we want to be more like tcp sockets; to be separate, which really increases reliability because we don't have the port going up and down all the time anymore. This reduces spurious disconnects.

It will also allow us to add socket activity timeouts to help with the windows dropping data problem, though we can't really do that until the http requests that cause synchronous actions that we make get keep alive streaming responses.
## Risk
- Low (usb basically still doesn't work before this PR)
## Testing
- [x] on a windows machine disconnected from the internet
   - [x] repeated physical connections and disconnections reliably detect and remove the robot
   - [x] standard workflows (i.e. all the random polling in the device page) work and display the robot
   - [x] complex workflows (i.e. module calibration) work
   - [x] download-heavy workflows (i.e. logs) work
   - [x] upload-heavy workflows (i.e. updates) work
   - [x] the robot gets detected again when it boots after updates
## Review requests
- is there something we just can't stand for in here? it's definitely ugly but is probably going to stay that way. i'm definitely going to continue following up with improvements.

### DIFF
--- a/app-shell/src/usb.ts
+++ b/app-shell/src/usb.ts
@@ -164,7 +164,7 @@ function tryCreateAndStartUsbHttpRequests(dispatch: Dispatch): void {
       }
       if (usbHttpAgent == null) {
         createSerialPortHttpAgent(ot3UsbSerialPort.path, (err, agent?) => {
-          if (err) {
+          if (err != null) {
             const message = err?.message ?? err
             usbLog.error(`Failed to create serial port: ${message}`)
           }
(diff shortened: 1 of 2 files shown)


b-cooper (approved):
🔌

PR #14215, opened 2023-12-15
fix(protocol-engine): Fix hanging after cancellation

SyntaxColoring:
# Overview

Fixes RQA-2091

# Test Plan

* [x] Flex and OT-2 protocols can still succeed
* [x] Flex protocols, when cancelled, home the X/Y/Z axes but leave the plungers and tip ejectors alone
* [x] OT-2 protocols, when cancelled, eject the tips in the fixed trash if there are any tips attached

# Changelog

* After a protocol ends, *always* call `hardware_api.stop` before doing anything further with the hardware. This is the bug fix.
* Formerly, if a protocol ended without any tips attached, we were taking the opportunity to home the plungers. That's dropped in this PR; for simplicity, we will never ever home the plungers as part of run cleanup. Homing the plungers is taken care of at run *start.* @sfoster1 and I agree that there was no benefit to homing the plungers during run cleanup, and that it could only open up opportunities for something to go wrong.

# Review requests

There are opportunities for further cleanup here. For example, `HardwareStopper._drop_tip()` is *also* stopping and homing, which makes the overall sequence hard to follow and seems outside of its responsibility.

Since this PR has already been tested *without* that cleanup, let's not do it in this PR—but if you have any ideas, please do leave comments.

# Risk assessment

Medium. I haven't looked at how well this code is covered by unit tests, but this PR is passing CI without having had to change anything, which suggests there are gaps.

-----

Co-authored with @jbleon95 and @sfoster1

### DIFF
--- a/api/src/opentrons/protocol_engine/execution/hardware_stopper.py
+++ b/api/src/opentrons/protocol_engine/execution/hardware_stopper.py
@@ -50,25 +50,29 @@ class HardwareStopper:
             state_view=state_store,
         )
 
+    async def _home_everything_except_plungers(self) -> None:
+        # TODO: Update this once gripper MotorAxis is available in engine.
+        try:
+            ot3api = ensure_ot3_hardware(hardware_api=self._hardware_api)
+            if (
+                not self._state_store.config.use_virtual_gripper
+                and ot3api.has_gripper()
+            ):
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me if it works, I think this is the right approach for homing

jbleon95 (approved):
Fixes the cancellation bug and looks correct to me

PR #14216, opened 2023-12-15
fix(api): do not tip check during attach

sfoster1:
Since on the 96, that moves motors while someone is holding it in their hands.

### Testing
- [x] on a flex with this pr, attach a 96. the motors shouldn't move
- [x] run calibration and make sure that it still checks for a tip


fsinapi (approved):
seems fair

PR #14217, opened 2023-12-15
fix(app): e-stop modal button state issue

koji:
# Overview
The issue was caused by unexpected state change via `useEffect`.
Remove `useEffect` and update `handleClick`. 
close RQA-2111

# Test Plan
1. press e-stop button
2. twist e-stop button
3. click resume button
the button shows spinner and not back to normal state and for ODD, the modal is gone.
There will be some delay since fetching duration is 10sec. 

# Changelog

# Review requests

# Risk assessment
low


mjhuff (approved):
Works great! Thank you for fixing this!

PR #14218, opened 2023-12-15
fix(app): ODD - Remove dismisscurrentrun on protocol cancel button

mjhuff:
# Overview

Thank you for spotting this @jbleon95 ! This PR fixes a behavior mismatch issue between the desktop app & ODD. The cancel run button on the ODD removes the run context, which drop tip needs to work. This behavior doesn't need to be tied to the cancel button as far as I can tell. 

# Test Plan
- I already tested this -- just double check the logic. 
- Tested running a protocol, pushing a protocol, canceling a run, starting a new run, etc. Drop tip works as expected on ODD now and there doesn't seem to be any regression.

# Changelog
- Cancelling a run on the ODD will allow a user to access drop tip wizard.

# Review requests
- Just check my logic.

# Risk assessment
low (likely)

### DIFF
--- a/app/src/organisms/OnDeviceDisplay/RunningProtocol/ConfirmCancelRunModal.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RunningProtocol/ConfirmCancelRunModal.tsx
@@ -68,7 +68,6 @@ export function ConfirmCancelRunModal({
   React.useEffect(() => {
     if (runStatus === RUN_STATUS_STOPPED) {
       trackProtocolRunEvent({ name: ANALYTICS_PROTOCOL_RUN_CANCEL })
-      dismissCurrentRun(runId)
       if (!isActiveRun) {
         if (protocolId != null) {
           history.push(`/protocols/${protocolId}`)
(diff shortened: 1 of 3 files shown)


app/src/organisms/OnDeviceDisplay/RunningProtocol/ConfirmCancelRunModal.tsx
```diff
@@ -40,10 +40,7 @@ export function ConfirmCancelRunModal({
 }: ConfirmCancelRunModalProps): JSX.Element {
   const { t } = useTranslation(['run_details', 'shared'])
   const { stopRun } = useStopRunMutation()
-  const {
-    dismissCurrentRun,
-    isLoading: isDismissing,
-  } = useDismissCurrentRunMutation()
+  const { isLoading: isDismissing } = useDismissCurrentRunMutation()
```
brenthagen: do we need `isDismissing` now?

koji (approved):
The change makes sense to me. 
ProtocolRunHeader doesn't use `dissmissCurrentRun` anymore.
Could you remove useDismissCurrentRunMutation from `app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunHeader.test.tsx` if it doesn't take much time?

PR #14220, opened 2023-12-15
fix(app): support addressable areas in manual intervention modal locations

b-cooper:
# Overview

Fix bug in which moves to manual moves to addressable areas would show nothing in the manual intervention modal

Closes RQA-2114, Closes RQA-2115

# Risk assessment
low


brenthagen (approved):
change makes sense to me ✔️

PR #14221, opened 2023-12-15
chore(shared-data): revert PCR plate grip force back to 15N

ahiuchingau:
# Overview
We're that some grippers are not able to pick up the Armadillo PCR properly after the change in QA, so we're changing it back to 15 N.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
ok cool!

PR #14222, opened 2023-12-15
fix(app): allow users to use the same name after device reset 

koji:
# Overview
This change allows users to use the same robot's name after device reset.

In the future, device reset will support clear a robot name when doing device reset. 

close RQA-2077

# Test Plan
Need to check 2 cases.
case1
1. reset device (clear all stored data)
2. doing unboxing flow and use the same name
no error message

case2
1. reset device (clear all stored data)
2. doing unboxing flow and use a different name
no error message

# Changelog

# Review requests

# Risk assessment


mjhuff (approved):
Works great! Just remove the console log and it's good to go! 🚀

PR #14223, opened 2023-12-15
fix(app): Fix overeager drop tip errors

mjhuff:
# Overview

Woops. This PR fixes a tip check that occurs by the hardware controller after a tip has dropped. HWC throws an error here, which we are catching and displaying to the user. This results in drop tip always failing after dropping. The long term solution is to talk to embedded.

# Test Plan
- Dropping tips shouldn't fail now.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/DropTipWizard/index.tsx
+++ b/app/src/organisms/DropTipWizard/index.tsx
@@ -507,12 +507,7 @@ export const DropTipWizardComponent = (
               ],
               true
             )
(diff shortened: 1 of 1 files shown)


brenthagen (approved):
✔️

mjhuff:
Instead of not throwing errors, let's remove the assertion on the HW side. 
https://github.com/Opentrons/opentrons/pull/14224

PR #14224, opened 2023-12-15
fix(api): allow double remove tip

sfoster1:
This seems like a logic problem with how we track tip state, but removing this check (which is really a nebulous "might be using it wrong" check) we fix some incorrect errors in the drop tip wizard.


mjhuff (approved):
Thank you!

ahiuchingau (approved):
I agree with this change wholeheartedly

ahiuchingau:
Oh as a follow-up i think the workaround in protocol engine that "adds-a-tip-just-so-we-can-drop-one" can be removed!

PR #14225, opened 2023-12-15
fix(api): Flex 2 15 api fixed trash auto tipdrop support

CaseyBatten:
# Overview
Seeks to solve the API 2 15 support for RQA-1990

# Test Plan
Run a 2.15 protocol on Flex that ends with a tip attached, ensure it auto drops in the fixed trash. Also run a 2.16 protocol on flex and ensure it does not auto drop them.
Run both on OT2 and ensure that the OT2 always drops tips.

# Risk assessment
This should be fixing support for Flex, but has changed how we gate behavior exclusively for OT2, test thoroughly.

### DIFF
--- a/api/src/opentrons/protocol_engine/execution/hardware_stopper.py
+++ b/api/src/opentrons/protocol_engine/execution/hardware_stopper.py
@@ -74,14 +74,23 @@ class HardwareStopper:
 
             await self._home_everything_except_plungers()
 
-            # OT-2 Will only ever use the Fixed Trash Addressable Area
-            if self._state_store.config.robot_type == "OT-2 Standard":
-                for pipette_id, tip in attached_tips:
-                    try:
+            for pipette_id, tip in attached_tips:
+                try:
+                    if self._state_store.labware.get_fixed_trash_id() == FIXED_TRASH_ID:
+                        # OT-2 and Flex 2.15 protocols will default to the Fixed Trash Labware
                         await self._tip_handler.add_tip(pipette_id=pipette_id, tip=tip)
-                        # TODO: Add ability to drop tip onto custom trash as well.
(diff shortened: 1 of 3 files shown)


CaseyBatten:
Just ran tests with a protocol 2.15 and 2.16 on Flex and OT2, the 2.16 Flex protocol did not automatically drop tips (as expected). In all other situations it automatically dropped the tips, as expected.

api/src/opentrons/protocol_engine/execution/hardware_stopper.py
```diff
@@ -74,14 +74,23 @@ async def _drop_tip(self) -> None:
 
             await self._home_everything_except_plungers()
 
-            # OT-2 Will only ever use the Fixed Trash Addressable Area
-            if self._state_store.config.robot_type == "OT-2 Standard":
-                for pipette_id, tip in attached_tips:
-                    try:
+            for pipette_id, tip in attached_tips:
+                try:
+                    if self._state_store.labware.get_fixed_trash_id() == FIXED_TRASH_ID:
```
SyntaxColoring, line 79: I think this will work but it's pretty confusing.

For protocols that happen to load normal labware in slot A3, the way this will work now is `get_fixed_trash_id()` will return *the ID of that normal labware,* and then this code will ignore it because it doesn't match `FIXED_TRASH_ID`.

I think what we want instead is for `get_fixed_trash_id()` to either:

* Return the ID of the fixed trash labware, if there is one
    * And then this code would use that ID; this code wouldn't know about `FIXED_TRASH_ID` itself
* Otherwise, return the name of the fixed trash addressable area, if there is one
    * And then this code would use that addressable area name
* Otherwise, there must be no fixed trash, so it would return `None`

SyntaxColoring (approved):
LGTM, thank you!

PR #14226, opened 2023-12-15
fix(api): center only active column on 12-well

sfoster1:
We have a system to center pipettes appropriately on the wells of labware that span multiple "traditional" wells. For instance, in a 12-well reservoir, we need to move the center of an eight channel pipette to the center of a given well of the 12-well reservoir.

However, that code was moving the center of _any_ pipette to the center of a 12-well reservoir, inclunding the center of the 96-channel, which is definitely not right. Instead, what we want to do is move the center of the "active" column (the column that contains the active nozzle, which is the nozzle that is moved when a caller specifies a NOZZLE critical point) to the center of the well. And we only want to do this on labware that have roughly one well per column.

The fix is in two parts.

First, a Y_CENTER critical point needs to exist. This critical point is aligned in the X direction with the active nozzle of the pipette, but is halfway between the Y positions of the backmost and frontmost nozzles in the column containing the active nozzle of the currently active nozzlemap - so, on an eight channel with a full map it would be the same as XY_CENTER; on a single, an eight channel with a single nozzle map, or a 96 with a single nozzle map, it will be the same as NOZZLE (and the same as XY_CENTER); and on an eight channel or 96 with a partial-column map it will be halfway up the partial.

On a 96 with a rectangular map (i.e. either full or quadrant) it will be halfway between the back left and front left nozzles.

Next, we have to actually use that critical point. The engine would check for a labware to have the centerMultiChannelOnWells quirk and emit XY_CENTER for both source and destination if it was found; we now also consider how many wells the labware has if it has
centerMultiChannelOnWells and (somewhat hackily, sadly) emit
- Y_CENTER, if the labware has the centerMultiChannelOnWells quirk and 1 < wells < 96
- XY_CENTER, if the labware has the centerMultiChannelOnWells quirk and wells == 1, wells >= 96

This should probably change eventually into a new quirk called like centerActiveColumnOnWells or something but that's for a different time.

## Testing
- In LPC and in a protocol
   - Move an 8 channel to a 12-well reservoir
   - Move a 96 channel to a 12-well reservoir in a full config
   - Move a 96 channel to a 12-well reservoir in a single-tip config
   - Move a 96 channel to a 12-well reservoir in a column config

Closes RQA-2118

### DIFF
--- a/api/src/opentrons/hardware_control/types.py
+++ b/api/src/opentrons/hardware_control/types.py
@@ -505,6 +505,16 @@ class CriticalPoint(enum.Enum):
     back calibration pin slot.
     """
 
+    Y_CENTER = enum.auto()
+    """
+    Y_CENTER means the critical point under consideration is at the same X
+    coordinate as the default nozzle point (i.e. TIP | NOZZLE | FRONT_NOZZLE)
+    but halfway in between the Y axis bounding box of the pipette - it is the
+    XY center of the first column in the pipette. It's really only relevant for
+    the 96; it will produce the same position as XY_CENTER on an eight or one
+    channel pipette.
+    """
+
 
 class ExecutionState(enum.Enum):
     RUNNING = enum.auto()
--- a/api/src/opentrons/protocol_engine/state/labware.py
+++ b/api/src/opentrons/protocol_engine/state/labware.py
@@ -381,6 +381,30 @@ class LabwareView(HasState[LabwareState]):
         definition = self.get_definition(labware_id)
         return definition.parameters.quirks or []
 
+    def get_should_center_column_on_target_well(self, labware_id: str) -> bool:
+        """True if a pipette moving to this labware should center its active column on the target.
+
+        This is true for labware that have wells spanning entire columns.
+        """
+        has_quirk = self.get_has_quirk(labware_id, "centerMultichannelOnWells")
+        return (
+            has_quirk
+            and len(self.get_definition(labware_id).wells) > 1
+            and len(self.get_definition(labware_id).wells) < 96
+        )
+
+    def get_should_center_pipette_on_target_well(self, labware_id: str) -> bool:
+        """True if a pipette moving to a well of this labware should center its body on the target.
+
+        This is true for 1-well reservoirs no matter the pipette, and for large plates.
+        """
+        has_quirk = self.get_has_quirk(labware_id, "centerMultichannelOnWells")
+        return (
+            has_quirk
+            and len(self.get_definition(labware_id).wells) == 1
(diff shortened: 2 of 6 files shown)


api/src/opentrons/hardware_control/types.py
```diff
@@ -505,6 +505,16 @@ class CriticalPoint(enum.Enum):
     back calibration pin slot.
     """
 
+    Y_CENTER = enum.auto()
```
SyntaxColoring, line 508: I might find a name like `X_LEFT_Y_CENTER` clearer.

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -381,6 +381,30 @@ def get_quirks(self, labware_id: str) -> List[str]:
...
+    def get_should_center_pipette_on_target_well(self, labware_id: str) -> bool:
+        """True if a pipette moving to a well of this labware should center its body on the target.
+
+        This is true for 1-well reservoirs no matter the pipette, and for large plates.
+        """
+        has_quirk = self.get_has_quirk(labware_id, "centerMultichannelOnWells")
+        return (
+            has_quirk
+            and len(self.get_definition(labware_id).wells) == 1
+            or len(self.get_definition(labware_id).wells) >= 96
```
SyntaxColoring, line 403: Is this intentional?

A 384 well plate:

* `has_quirk` is `False`
* `len(self.get_definition(labware_id).wells) == 1` is `False`
* `len(self.get_definition(labware_id).wells) >= 96` is `True`

`False and False or True` is `True`

So if I do `p1000_96.move_to(plate_384["A1"].top())`, won't that center the whole pipette body over well A1?

CaseyBatten (approved):
Ran on ABR6 with protocol referencing 12 well labware at C1, Flex correctly executed commands in labware at proper position/offsets. Looks good once linting etc is good.

api/src/opentrons/hardware_control/types.py
sfoster1: I disagree because that makes me think it's on the left of the bounding box.

api/src/opentrons/protocol_engine/state/labware.py
sfoster1: It was not! It was missing parentheses. Now it is correct

SyntaxColoring (approved):
Nice

api/src/opentrons/hardware_control/types.py
SyntaxColoring: `X_LEFTCOL_Y_CENTER` maybe?

PR #14228, opened 2023-12-15
fix(app): Fix persistent cancelled state

mjhuff:
Closes RQA-2123 and RQA-2122

# Overview

We actually do want to close the current run as we were before, as this logic is used elsewhere in the app. This PR adds back this logic and only shows the drop tip banner if the most recent run is equal to the runId passed into the run details page. 

# Test Plan
- Test drop tip flows work. 
- Cancelling a run should allow you to perform actions from the devices page overflow menu such as homing the gantry or restarting the robot (these are no longer disabled). 

# Risk assessment
low


sfoster1 (approved):
Looks good to me!

PR #14230, opened 2023-12-18
feat(api): add pipette movement restrictions for 96ch col A1 and 8ch single A1

sanni-t:
Closes RSS-423

# Overview

Implements deck conflict checking for when:
- moving a 96 channel pipette configured to use the leftmost nozzle column 
- moving an 8 channel pipette configured to use the backmost nozzle

Even though 7.1.0 release doesn't officially support any partial configurations other than 96-channel rightmost column, one can write protocols that use other configurations. The conflict checking added in this PR makes these not-yet-supported configurations at least safe for use.

# Test Plan

- Below is a protocol built from the integration test for deck conflict. Make sure this analyses correctly on the robot (protocol has errors on purpose) and is actually able to run without any collisions once errors are removed and analysis passes.
- From my calculations and the integration tests, it seems that the 96-channel first column can access the full deck slot area, including the right edge of deck slot column 3. Please verify that this is actually true.

Test protocol:
```


requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}

def run(protocol):

    instrument = protocol.load_instrument("flex_96channel_1000", mount="left")
    trash_labware = protocol.load_labware("opentrons_1_trash_3200ml_fixed", "A3")

    badly_placed_tiprack = protocol.load_labware("opentrons_flex_96_tiprack_50ul", "C2")
    well_placed_tiprack = protocol.load_labware("opentrons_flex_96_tiprack_50ul", "A1")
    tiprack_on_adapter = protocol.load_labware(
        "opentrons_flex_96_tiprack_50ul",
        "C3",
        adapter="opentrons_flex_96_tiprack_adapter",
    )

    # ############  SHORT LABWARE  ################
    # These labware should be to the east of tall labware to avoid any partial tip deck conflicts
    badly_placed_plate = protocol.load_labware("nest_96_wellplate_200ul_flat", "B1")
    well_placed_plate = protocol.load_labware("nest_96_wellplate_200ul_flat", "B3")

    # ############ TALL LABWARE ###############
    my_tuberack = protocol.load_labware(
        "opentrons_10_tuberack_falcon_4x50ml_6x15ml_conical", "B2"
    )

    # ########### Use Partial Nozzles #############

    # Raises error because of conflict with tiprack on adapter in C3
    instrument.pick_up_tip(badly_placed_tiprack.wells_by_name()["H12"])

    # No error cuz within pipette extent bounds and no taller labware to right of tiprack
    instrument.pick_up_tip(well_placed_tiprack.wells_by_name()["A12"])

    # No error cuz no labware on right of plate
    instrument.aspirate(25, well_placed_plate.wells_by_name()["A4"])
    # No error cuz dispensing from high above plate, so it clears tuberack on the right
    instrument.dispense(25, badly_placed_plate.wells_by_name()["A1"].top(150))

    
    # Raises error because of conflict with tuberack
    instrument.aspirate(25, badly_placed_plate.wells_by_name()["A1"])

    # No error cuz no taller labware on the right
    instrument.aspirate(10, my_tuberack.wells_by_name()["A1"])

    instrument.drop_tip()

    # ######## CHANGE CONFIG TO ALL #########
    instrument.configure_nozzle_layout(style=ALL, tip_racks=[tiprack_on_adapter])

    # No error because of full config
    instrument.pick_up_tip()

    # No error NOW because of full config
    instrument.aspirate(50, badly_placed_plate.wells_by_name()["A1"])

    # No error NOW because of full config
    instrument.dispense(50, badly_placed_plate.wells_by_name()["A1"].bottom())
```

# Changelog

- updated the pipette movement deck conflict checks to include conflicts with east and south slots for the appropriate pipette & nozzle configuration.
- added integration tests for the 96-channel configurations. Did not add unit test for it since the existing ones cover the basic functionality testing and an integration test is more appropriate for such instances where the final result is dependent on a lot of internal state parameters.

# Review requests

- Test it on a robot to make sure that all the new as well as old conflicts with a 96-channel work as expected.
- Any code suggestions

# Risk assessment

Medium. Adds new restrictions for yet unsupported pipette configuration but also touches the logic for supported configurations.


api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -265,41 +270,50 @@ def _check_deck_conflict_for_96_channel(
...
-    )
-    if west_slot_number is None:
-        return
 
-    west_slot = DeckSlotName.from_primitive(
-        west_slot_number
-    ).to_equivalent_for_robot_type(engine_state.config.robot_type)
+    destination_slot_num = _deck_slot_to_int(DeckSlotLocation(slotName=labware_slot))
+    adjacent_slot_num = None
+    if primary_nozzle == "A12":
```
sfoster1, line 296: we probably want to change this eventually to "column 1"/"column 12" via the column mappings in the pipette geometry def.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -330,41 +343,48 @@ def _check_deck_conflict_for_8_channel(
...
-    )
-    if north_slot_number is None:
-        return
-
-    north_slot = DeckSlotName.from_primitive(
-        north_slot_number
-    ).to_equivalent_for_robot_type(engine_state.config.robot_type)
+    destination_slot = _deck_slot_to_int(DeckSlotLocation(slotName=labware_slot))
+    adjacent_slot_num = None
+    if primary_nozzle == "H1":
```
sfoster1, line 372: probably want to change this to column mapping via the geometry def at some point

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -378,20 +398,26 @@ def _is_within_pipette_extents(
...
-                A12_column_front_left_bound.x
-                < location.x
-                < A12_column_back_right_bound.x
-                and A12_column_front_left_bound.y
-                < location.y
-                < A12_column_back_right_bound.y
-            )
-    # TODO (spp, 2023-11-07): check for 8-channel nozzle H1 extents on Flex & OT2
+        if pipette_channels == 96 and nozzle_config == NozzleConfigurationType.COLUMN:
+            if primary_nozzle == "A12":
```
sfoster1, line 415: probably want to change this to column mapping in the geometry def at some point

sfoster1 (approved):
Looks good to me pending comments

CaseyBatten:
When attempting to drop tips in a trash bin in A3 with the leftmost column (A1) filled with tips, the 96 channel collides with the right side of the robot. Also revealed through this, when the 96 channel has a single column of tips in A1 and is at its rightmost position, the further it can reach is A10. The single column will never reach A11 or A12 within the tiprack/wellplate/labware it is interacting with.

sanni-t:
> When attempting to drop tips in a trash bin in A3 with the leftmost column (A1) filled with tips, the 96 channel collides with the right side of the robot. 

Hmm, the robot extents/ pipette bounds checking currently doesn't happen when moving to addressable areas so this isn't unexpected.

> Also revealed through this, when the 96 channel has a single column of tips in A1 and is at its rightmost position, the further it can reach is A10. The single column will never reach A11 or A12 within the tiprack/wellplate/labware it is interacting with.

I'm assuming you're talking about column A10 of a labware in the third deck column. This is an important thing to fix in this PR. The behavior indicates that the numbers I've used for out-of-bounds check are incorrect. I'll find out what the correct numbers are and update this PR

SyntaxColoring:
>> When attempting to drop tips in a trash bin in A3 with the leftmost column (A1) filled with tips, the 96 channel collides with the right side of the robot.
> 
> Hmm, the robot extents/ pipette bounds checking currently doesn't happen when moving to addressable areas so this isn't unexpected.

I think I told @CaseyBatten yesterday that this would be a problem for v7.1.0, but I was under the mistaken impression at the time that we were trying to support actually using column 1. After discussing with @sanni-t and Linda, I agree that it's okay for that gap in motion planning and validation to not be fixed in this PR.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -378,20 +402,28 @@ def _is_within_pipette_extents(
...
+                )
+            elif primary_nozzle == "A1":
+                return (
+                    A1_column_front_left_bound.x
+                    < location.x
+                    < A1_column_back_right_bound.x
+                    and A1_column_front_left_bound.y
+                    < location.y
+                    < A1_column_back_right_bound.y
+                )
```
SyntaxColoring: Pedantically, should these be `<=` instead of `<`? If the left boundary of a configuration is, say, 1.23, I'd personally expect to be able to move to exactly x=1.23.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -53,6 +55,10 @@ def __init__(self, message: str) -> None:
 A12_column_front_left_bound = Point(x=-1.8, y=2)
 A12_column_back_right_bound = Point(x=592, y=506.2)
 
+A1_column_front_left_bound = Point(x=-100.8, y=2)
+A1_column_back_right_bound = Point(x=493, y=506.2)
```
SyntaxColoring: It sounds like you and Seth have a vision for the proper way to compute these bounds, but until then, would this be more maintainable?

```suggestion
_NOZZLE_PITCH = 9
A1_column_front_left_bound = Point(
    x=A12_column_front_left_bound.x - _NOZZLE_PITCH * 11, y=2
)
A1_column_back_right_bound = Point(
    x=A12_column_front_left_bound.x - _NOZZLE_PITCH * 11, y=506.2
)
```

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -248,7 +254,6 @@ def _check_deck_conflict_for_96_channel(
     if not (
         engine_state.pipettes.get_nozzle_layout_type(pipette_id)
         == NozzleConfigurationType.COLUMN
-        and engine_state.pipettes.get_primary_nozzle(pipette_id) == "A12"
     ):
         # Checking deck conflicts only for 12th column config
```
SyntaxColoring: Does this `# Checking deck conflicts only for 12th column config` comment need updating?

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -330,41 +345,50 @@ def _check_deck_conflict_for_8_channel(
...
-    north_slot_number = get_north_slot(
-        _deck_slot_to_int(DeckSlotLocation(slotName=labware_slot))
-    )
-    if north_slot_number is None:
-        return
-
-    north_slot = DeckSlotName.from_primitive(
-        north_slot_number
-    ).to_equivalent_for_robot_type(engine_state.config.robot_type)
+    destination_slot = _deck_slot_to_int(DeckSlotLocation(slotName=labware_slot))
```
SyntaxColoring: Nit: I think this can be simplified to eliminate the conversion back and forth from `DeckSlotLocation`.

```suggestion
    destination_slot = labware_slot.as_int()
```

And then `_deck_slot_to_int()` can be deleted. It was helpful when the state view was returning a `DeckSlotLocation`, but it's unnecessary now that the state view is directly returning a `DeckSlotName`.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -265,41 +270,52 @@ def _check_deck_conflict_for_96_channel(
...
-    west_slot_number = get_west_slot(
-        _deck_slot_to_int(DeckSlotLocation(slotName=labware_slot))
-    )
-    if west_slot_number is None:
-        return
 
-    west_slot = DeckSlotName.from_primitive(
-        west_slot_number
-    ).to_equivalent_for_robot_type(engine_state.config.robot_type)
+    destination_slot_num = _deck_slot_to_int(DeckSlotLocation(slotName=labware_slot))
```
SyntaxColoring: same

```suggestion
    destination_slot_num = labware_slot.as_int()
```

SyntaxColoring (commented):
~~I'm still catching up on the 96-channel code so I haven't worked all the way through this, but~~ this looks good to me so far. Just some minor suggestions.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -265,41 +270,52 @@ def _check_deck_conflict_for_96_channel(
...
+    #  via the column mappings in the pipette geometry definitions.
+    if primary_nozzle == "A12":
+        adjacent_slot_num = get_west_slot(destination_slot_num)
+    elif primary_nozzle == "A1":
+        adjacent_slot_num = get_east_slot(destination_slot_num)
+
+    def _check_conflict_with_slot_item(
+        adjacent_slot: DeckSlotName,
+    ) -> None:
+        """Raises error if the pipette is ecpected to collide with adjacent slot items."""
```
SyntaxColoring:
```suggestion
        """Raises error if the pipette is expected to collide with adjacent slot items."""
```

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -313,7 +329,6 @@ def _check_deck_conflict_for_8_channel(
     if not (
         engine_state.pipettes.get_nozzle_layout_type(pipette_id)
         == NozzleConfigurationType.SINGLE
-        and engine_state.pipettes.get_primary_nozzle(pipette_id) == "H1"
     ):
         # Checking deck conflicts only for H1 single tip config
```
SyntaxColoring: Same—does this `# Checking deck conflicts only for H1 single tip config` comment need to be updated?

SyntaxColoring (approved):
Also LGTM pending comments.

sanni-t:
Tested the new robot bounds on a Flex and all accessible areas were reachable by the pipette without collision while the locations that raise an error are actually not accessible by the pipette. 
Exception is when the location we're moving to is an addressable area and doing checks for addressable areas is covered in a separate ticket.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: I had used `<` instead of `<=` to keep a little bit of margin since I don't expect all machines to have the exact same boundaries. That said, I have changed it to your suggestion since we are now aiming to fetch the boundary values from machine geometry data, which will be much more precise for all machines.

sanni-t:
@ecormany @jwwojak tagging you here to let you know about this robot bounds check change. Turns out we _can_ access the first column of thermocycler labware using A12 nozzle column of the 96 channel (although it's very close to the left bound, so big offsets to this location might still result in out of bounds error). Once this change is released the API will no longer raise an error for this condition and we will have to update the docs that mention this restriction.

api/tests/opentrons/protocol_api_integration/test_pipette_movement_deck_conflicts.py
```diff
@@ -0,0 +1,180 @@
+"""Tests for the APIs around deck conflicts during pipette movement."""
+
+import pytest
+
+from opentrons import simulate
+from opentrons.protocol_api import COLUMN, ALL
+from opentrons.protocol_api.core.engine.deck_conflict import (
+    PartialTipMovementNotAllowedError,
+)
+from opentrons.types import Point
```
SyntaxColoring:
```suggestion
```

There's a lint error from this unused import, but this otherwise looks good to me to merge.

SyntaxColoring (approved):

PR #14233, opened 2023-12-18
fix(app): Don't render ProtocolRunHeader TerminalBanner on CurrentRun

mjhuff:
Closes RQA-2013

# Overview

This PR fixes an issue where opening the error banner would de-render the error banner. Instead of conditionally rendering the error banner on whether a run is current, let's follow the new pattern we use for the drop tip banner: render it if the mostRecentRunId is the currently viewed run. 

Addresses the same issue for clicking on "remove tips" within the drop tip banner - the wizard would never properly start after viewing the error message (since it clears the run). 

PR includes some refactoring for the getMostRecentRunId hook to make it a bit safer.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/7f0d985f-104b-49a8-b518-6a8f4f414d36

### Fixed Behavior

https://github.com/Opentrons/opentrons/assets/64858653/a8220c11-940d-4764-a185-c75760581491

# Test Plan

- Follow video: Throw an error during a run, then confirm that the error banner/drop tip banner persists after opening the error modal.

# Changelog
- Fixed a bug in which viewing the error modal in a protocol run unrenders the drop tip banner.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunHeader.test.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunHeader.test.tsx
@@ -1054,26 +1054,4 @@ describe('ProtocolRunHeader', () => {
       expect(queryByText('Tips may be attached.')).not.toBeInTheDocument()
     })
   })
-
-  it('does not show the drop tip banner when the run is not over', async () => {
-    when(mockUseRunQuery)
-      .calledWith(RUN_ID)
-      .mockReturnValue({
-        data: {
-          data: {
-            ...mockIdleUnstartedRun,
-            current: false,
-            status: RUN_STATUS_SUCCEEDED,
-          },
-        },
-      } as UseQueryResult<Run>)
-    when(mockUseRunStatus)
-      .calledWith(RUN_ID)
-      .mockReturnValue(RUN_STATUS_SUCCEEDED)
-
-    const [{ queryByText }] = render()
-    await waitFor(() => {
-      expect(queryByText('Tips may be attached.')).not.toBeInTheDocument()
-    })
-  })
 })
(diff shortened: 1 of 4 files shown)


app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunHeader.test.tsx
```diff
@@ -1054,26 +1054,4 @@ describe('ProtocolRunHeader', () => {
       expect(queryByText('Tips may be attached.')).not.toBeInTheDocument()
     })
   })
-
```
mjhuff, line 1057: This test is functionally a duplicate of the one above now that we don't tie the drop tip banners to currentRunId, therefore I removed it.

app/src/organisms/ProtocolUpload/hooks/useMostRecentRunId.ts
```diff
@@ -1,8 +1,9 @@
 import { useAllRunsQuery } from '@opentrons/react-api-client'
+import { last } from 'lodash'
```
koji: Basically we import a specific one instead of importing the entire `lodash`

```suggestion
import last from 'lodash/last'
```

app/src/organisms/ProtocolUpload/hooks/useMostRecentRunId.ts
```diff
@@ -1,8 +1,9 @@
...
 
 export function useMostRecentRunId(): string | null {
   const { data: allRuns } = useAllRunsQuery()
-  return allRuns != null && allRuns.data.length > 0
-    ? allRuns.data[allRuns.data.length - 1].id
+  return allRuns != null && allRuns.data?.length > 0
+    ? last(allRuns.data)?.id ?? null
```
koji, line 7: 👍

koji (approved):
The changes look good to me.
Thank you for fixing this.

PR #14235, opened 2023-12-18
fix(shared-data): add slot transforms for A3 for heater-shaker and temp module v2

jbleon95:
# Overview

Fixes RQA-2136.

Once upon a time, modules could not go on slot A3 since that's where the fixed trash was. Then we set the trash free to roam around the deck, allowing that slot to be used for a heater-shaker or temperature module. Unfortunately, we never accounted for this in the module definition and left the A3 affine slot transformation empty. This caused large LPC offsets for labware on modules and completely broke module calibration, which can't use offsets.

Now this has been fixed by judicious copy and pasting of the column 3 transforms for slot A3 as well.

# Test Plan

Tested calibrating temperature module on A3, both without this fix and with this fix to ensure that it worked as expected (it did).

# Changelog

- added a slot A3 affine transform matrix to `slotTransforms` for `temperatureModuleV2` and `heaterShakerModuleV1`

# Review requests

# Risk assessment

Low, this fixes a clearly missing piece of supporting modules on A3. The only danger is using an old adjusted offset now that the position has been fixed, since that will cause the pipette to go too far down.


sfoster1 (approved):
hooray, thank you!

sanni-t (approved):

PR #14239, opened 2023-12-18
refactor(app): remove excessive axis homing in drop tip wizard

mjhuff:
# Overview

Now that moveToAddressableArea provides the ability to home to Z axis before moving, let's remove the homing done during the flows but keep still keep the homing behavior at the end of the flow.

# Test Plan
- Run through drop tip flows. Observe no homing of x & y axes when moving between locations. 
- Pressing "Go back" after selecting a location shouldn't home the x & y axis as well (but Z axis will home if jogging Z has occurred).

# Changelog
- Removed unnecessary x & y axis homing in drop tip wizard.

# Risk assessment
low


sfoster1 (approved):
Looks correct to me!

PR #14243, opened 2023-12-19
fix(protocol-designer): remove expansionSlot from staging area deck maps

jerader:
closes RQA-2147

# Overview

The ticket mentions that the staging area modal deck maps show the expansion slots but the module maps don't. So i removed showing the expansion slots to make it consistent. I also made the size of the staging area slot modal deck map bigger.

# Test Plan

See that in create file wizard and editing the staging area slots, the deck maps don't show the expansion slots.

# Changelog

- specify expansion slot boolean as false
- adjust size of deck map

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #14244, opened 2023-12-19
feat(hardware-testing): Improve liquid-height accuracy by using lookup tables during volumetric tests

andySigler:
# Overview

Our volumetric tests require knowing where the liquid meniscus is within a well. This means we need to somehow use the known inner-geometry of the well to convert ul to mm.

Until we have a more robust system for doing this in our software proper, then we can instead us simple lookup tables during our volumetric tests.

This PR adds lookup tables for the Nest 1-Well 195ml, and also the Corning 96-well 360ul. This PR also makes sure that these lookup tables are used during all volumetric tests, within our liquid-tracker class.

These lookup tables will, most importantly, improve our liquid-height accuracy within the 1-well nest 195ml reservoir during our 96ch photometric tests by almost 1mm, which will then help us to reduce the amount of dye waste that we see during production.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
LGTM

PR #14245, opened 2023-12-19
feat(hardware-testing): collected work for upgrading scripts to PE

ryanthecoder:
# Overview
This upgrades our pipette QC scripts to using protocol engine interface, this simplifies a lot of the code and lets us use new features:
  push out
  labware adapters
  using the normal ot3 deck definitions
  partial tip configuration

With the reduction in the amount of changes to the base repo required, we now don't need to use patches before running the tests, this means we only need the hardware-testing directory to be pushed to the bot with no changes to the base software.

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/hardware_control/backends/flex_protocol.py
```diff
@@ -433,3 +433,6 @@ async def set_hepa_uv_state(self, light_on: bool, uv_duration_s: int) -> bool:
 
     async def get_hepa_uv_state(self) -> Optional[HepaUVState]:
         ...
+
+    def _update_tip_state(self, mount: OT3Mount, status: bool) -> None:
```
sfoster1: if this is in the protocol, it's a public interface (for the backends, anyway) which means that it shouldn't be a private method.

I think it might be a good idea to not have it be
1. A public name
2. Not in the protocol or in the controller
3. Calling code should explicitly isinstance-check that this is a simulator before calling it

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -778,3 +778,7 @@ def configure_nozzle_layout(
         self._engine_client.configure_nozzle_layout(
             pipette_id=self._pipette_id, configuration_params=configuration_model
         )
+
+    def retract(self) -> None:
+        """Retract this instrument to the top of the gantry."""
+        self._sync_hardware_api.retract(self.get_mount())
```
sfoster1: code in the core here should not bypass the engine to cause motion, IMO. if we want to do this, we should implement it by calling a protocol engine command through `_engine_client` - a pretty good way to do it would be to home the mount.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1478,6 +1478,12 @@ def move_to(
 
         return self
 
+    @requires_version(2, 0)
```
sfoster1: this does not seem correct. if we haven't had this in here before, it has not been here since 2.0. This should be `@requires_version(2, 18)`

hardware-testing/Makefile
```diff
@@ -79,36 +79,36 @@ sdist:
 
 .PHONY: test
 test:
-	-$(MAKE) apply-patches-gravimetric
+#	-$(MAKE) apply-patches-gravimetric
```
sfoster1: let's delete these lines instead of commenting them

sfoster1 (changes requested):
Some stuff we should definitely change - we cannot add `retract` retroactively liks this, and I think we should implement it differently in the engine core

sfoster1 (approved):
Looks good to me, love the change to retract

PR #14249, opened 2023-12-19
Fix instrument name in firmware update results modal

smb2268:
# Overview
Stop showing instrumentName for firmware update success modal, instead show display name

# Test Plan

# Changelog
For a pipette, grab the display name from `getPipetteModelSpecs` and for gripper display "Flex Gripper"

# Review requests

# Risk assessment
Low


mjhuff (approved):
lgtm :+1: - merge target probably needs to change now, but I'll leave that to you.

PR #14250, opened 2023-12-19
fix(protocol-designer): getNextAvailableDeckSlot to filter out unneeded addressable areas

jerader:
closes RQA-2150

# Overview

Fixes 2 bugs related to `getNextAvailableDeckSlot` and filtering out certain addressable areas

# Test Plan

Create a flex protocol and go to the deck map and duplicate the tiprack as much as possible. The tiprack should not duplicate into a column 4 and should not duplicate at all. If you open dev tools, you should see a "cannot duplicate labware" error if you duplicate more. If you add a staging area slot, make sure that can be duplicated into.

Create an ot-2 protocol and do the same thing and duplicate the tiprack as much as possible, it should not duplicate into the fixed trash.

# Changelog

- add filtering for the staging area slots and trash addressable areas 
- fix a bit of stuff in the thunk to make it more readable

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/labware-ingred/utils.ts
+++ b/protocol-designer/src/labware-ingred/utils.ts
@@ -25,13 +25,12 @@ export function getNextAvailableDeckSlot(
       cutoutIds
     )
     const addressableAreaName = stagingAreaAddressableAreaNames.find(
-      aa => aa === slot?.id
+      aa => aa === slot.id
     )
     let isSlotEmpty: boolean = getSlotIsEmpty(initialDeckSetup, slot.id)
     if (addressableAreaName == null && COLUMN_4_SLOTS.includes(slot.id)) {
       isSlotEmpty = false
-    }
-    if (
+    } else if (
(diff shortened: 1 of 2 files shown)


protocol-designer/src/labware-ingred/utils.ts
```diff
@@ -1,16 +1,45 @@
...
+
+  return deckDef.locations.addressableAreas.find(slot => {
+    const cutoutIds = Object.values(initialDeckSetup.additionalEquipmentOnDeck)
+      .filter(ae => ae.name === 'stagingArea')
+      .map(ae => ae.location as CutoutId)
+    const stagingAreaAddressableAreaNames = getStagingAreaAddressableAreas(
+      cutoutIds
+    )
+    const addressableAreaName = stagingAreaAddressableAreaNames.find(
+      aa => aa === slot?.id
```
koji: nit
from the code, I think slot isn't `undefined` also because of L30 and L31
```suggestion
      aa => aa === slot.id
```

koji (approved):

PR #14252, opened 2023-12-19
fix(protocol-designer): remove fixed trash option from dropdowns

ncdiehl11:
closes [RAUT-908](https://opentrons.atlassian.net/browse/RAUT-908)

# Overview

In PD schemaV8, we use trash bin entities on OT-2 rather than implictly loaded fixed trash single-well labwares. Here, I update the v8 migration to remove loadLabware command for fixedTrash so that it is not present as an option alongside Trash Bin in dropdowns for dispense, droptip, etc.

# Test Plan

- import an old version of PD protocol ([like this v3 protocol](https://github.com/Opentrons/opentrons/files/13720048/doItAllV3.json))
- in protocol design, select a step that accepts trash locations (dispense, blowout, droptip...)
- observe that 'Fixed Trash' is no longer present. Only 'Trash Bin' should be present (refers to OT-2 slot 12 trash addressable area) 

Screen Shot 2023-12-19 at 3 27 55 PM

# Changelog

- filter loadLabware command specific to fixedTrash in v8 migration file

[RAUT-908]: https://opentrons.atlassian.net/browse/RAUT-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


protocol-designer/src/load-file/migration/8_0_0.ts
```diff
@@ -67,6 +67,20 @@ export const migrateFile = (
     },
   ]
 
+  const migrateCommands = (
+    v7Commands: ProtocolFileV7<{}>['commands']
```
jerader: typing it like this is okay but it might be better to type it as `CreateCommand[]` imported from schemaV7, that would be a bit clearer imo

protocol-designer/src/load-file/migration/8_0_0.ts
```diff
@@ -67,6 +67,20 @@ export const migrateFile = (
...
+  ): ProtocolFile['commands'] => {
```
jerader: same here, then this type can be `CreateCommand[]` from schema v8

protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -137,7 +137,6 @@ export const getLabwareOptions: Selector<Options> = createSelector(
       },
       []
     )
-
```
jerader: accidental line removal!

protocol-designer/src/load-file/migration/8_0_0.ts
```diff
@@ -67,6 +67,20 @@ export const migrateFile = (
...
+    return v7Commands.filter(
+      v7Command =>
+        !(
+          v7Command.commandType === 'loadLabware' &&
+          v7Command.params.labwareId === 'fixedTrash'
+        )
```
jerader, line 80: what you have totally works, it feels a bit cleaner imo to filter it this way as well! Either way though.

```suggestion        
          v7Command.commandType === 'loadLabware' &&
          v7Command.params.labwareId !== 'fixedTrash'       
```

jerader (approved):
thanks for tackling this! I left a few clean up suggestions.

protocol-designer/src/load-file/migration/8_0_0.ts
ncdiehl11: won't this filter out all non-loadLabware commands?

jerader: ooh good call, i forgot this is looking through all commands instead of just load commands. Ya you can keep it as is!

ncdiehl11: no prob!

ncdiehl11:
moved [here](https://github.com/Opentrons/opentrons/pull/14256) after rebase issues

PR #13546, opened 2023-09-13
fix(api): raise an error on analysis when dispensing invalid volume

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/RSS-330.
added validation for dispense in analysis - make sure there was an aspirate before dispensing. 

# Test Plan

Upload the following protocol and make sure it fails analysis. Should happen for both OT-2 and OT-3 with api version above 2.14.

```
#APIv2

requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.15",
}


def run(protocol_context):
	tiprack1 = protocol_context.load_labware("opentrons_96_tiprack_300ul", "1")
	pipette = protocol_context.load_instrument(
		"p300_single_gen2", mount="right", tip_racks=[tiprack1]
	)
	pipette.pick_up_tip(tiprack1.wells()[0])
	well_plate = protocol_context.load_labware("nest_96_wellplate_200ul_flat", "2")
	# pipette.aspirate(20, well_plate.wells()[0])
	pipette.dispense(20, well_plate.wells()[0])

```

Tested on dev server - works as expected! needs to be tested on the robot.

# Changelog

- Added validation to the dispense command. check if there was an aspiration before dispensing and that the volume is correct. 

# Review requests

are these checks sufficient? I don't think we can check the plunger position without the h/w api unless I am wrong.
Should I add this in the docs?

# Risk assessment

low. raising an exception while analyzing dispense command without aspirating before.

PR #2481, opened 2018-10-15
refactor(api): refactor aspirate/dispense

sanni-t:
Closes #2235

## overview
This PR adds aspirate and dispense functions to `hardware_control`

## changelog
Changed instruments cache from being a cache of pipette model to being a cache of all pipette config values
Added:
- `aspirate(mount: types.Mount, volume: float = None, rate: float = 1.0)`
- `dispense(mount: types.Mount, volume: float = None, rate: float = 1.0)`
- Method to retrieve pipette properties: `attached_instruments`
- backend methods: `set_active_current()`, `set_pipette_speed()`, `Controller._connect()`

PR #14253, opened 2023-12-19
fix(api): Fix floating point errors and backwards incompatibility in volume validation

SyntaxColoring:
# Overview

This fixes two bugs: RQA-2001 and RESC-182.

RQA-2001 is that if you had an incorrect protocol like this:

```python
pipette.pick_up_tip()
pipette.aspirate(50)
pipette.dispense(100)  # bad, invalid volume
```

...then it has started to raise an exception instead of silently clamping `100` to `50`. This is a good thing to do, but it's a backwards-incompatible change, and it should have been behind an `apiLevel` bump. This PR makes the change opt-in, behind `"apiLevel": "2.17"`.

RESC-182 is related. If you had a correct protocol like this:

```python
requirements = {"robotType": "OT-2", "apiLevel": "2.16"}

def run(protocol):
    tip_rack = protocol.load_labware("opentrons_96_tiprack_300ul", location="9")
    well_plate = protocol.load_labware("opentrons_24_tuberack_eppendorf_1.5ml_safelock_snapcap", location="10")

    pipette = protocol.load_instrument("p300_single_gen2", mount="left", tip_racks=[tip_rack])

    pipette.pick_up_tip()
    pipette.distribute(
        volume=[22.7, 22.7],
        source=well_plate["A1"],
        dest=[well_plate["B1"], well_plate["B2"]],
        air_gap=10,
        new_tip="never",
        disposal_volume=0,
    )
    pipette.drop_tip()
```

...then it has started to raise "invalid volume" errors stemming from floating point rounding errors. This is more straightforwardly a bug; it should just not do that.

# Test Plan

I've tested this with the protocols from the tickets.

# Background

The long-standing historical behavior (since #2481 at least) seems to be that we *clamp dispenses* but do *strict bounds-checking on aspirates.* I'm not sure why they differed from each other. I'm also not sure how the strict bounds-checking on aspirates never caused any problems with floating point rounding.

In RSS-330 / PR #13546, we added strict bounds-checking to the Protocol Engine `dispense` command. We mistakenly thought we were fixing an omission and bringing simulation in line with execution, but we were actually making a breaking change to the interface. And, separately, we forgot about floating point error.

# Changelog

* At the Protocol Engine level, lean into doing strict bounds-checking everywhere, instead of clamping.
    * Unlike the historical behavior, do it for both aspirates *and* dispenses.
    * Add a tolerance to account for floating point rounding.
    * This is a potentially breaking change for anyone directly using Protocol Engine (e.g. generating JSON protocols, or using the HTTP `/runs` API). We could perhaps add a `"clamp": true` param to `dispense` to change that, but I suspect it's not worthwhile.
    * I don't believe this is a breaking change for anyone using Protocol Designer, because Protocol Designer already does its own bookkeeping of the pipette volume.
    * This is not a breaking change  for anyone using the Python Protocol API, because of the next bullet point.
* Preserve backwards-compatible behavior *for Python Protocol API users* by adding conditional clamping based on `"apiLevel"`.

# Review requests

* Do we agree with the breaking change for the Protocol Engine `dispense` command?
* Is it correct to apply this bounds-checking to only the virtual pipette handler, or do we also want to do it on the hardware pipette handler?

# Risk assessment

Medium.


sfoster1 (approved):
Yeah, this looks good to me! I agree with the PE dispense change, nice fix.

jbleon95 (approved):
LGTM, keeps everything consistent now.

api/release-notes.md
ecormany, line 1: If everyone else is cool with this to-do hanging out in the release notes file for now, it's fine with me too. Otherwise, we can offload it to a comment on RTC-384 until the time is right.

SyntaxColoring:
I'm not sure what the deal was with the timing out G-Code-Confirm test, but it works [when I trigger it manually](https://github.com/Opentrons/opentrons/actions/runs/7495166388).

I'm unable to easily test on an actual robot at the moment because of unrelated boot problems, but I'm satisfied with our automated test coverage for now.

PR #14256, opened 2023-12-20
fix(protocol-designer): remove fixed trash option from dropdowns

ncdiehl11:
closes [RAUT-908](https://opentrons.atlassian.net/browse/RAUT-908)

# Overview

In PD schemaV8, we use trash bin entities on OT-2 rather than implictly loaded fixed trash single-well labwares. Here, I update the v8 migration to remove loadLabware command for fixedTrash so that it is not present as an option alongside Trash Bin in dropdowns for dispense, droptip, etc.

# Test Plan

- import an old version of PD protocol ([like this v3 protocol](https://github.com/Opentrons/opentrons/files/13720048/doItAllV3.json))
- in protocol design, select a step that accepts trash locations (dispense, blowout, droptip...)
- observe that 'Fixed Trash' is no longer present. Only 'Trash Bin' should be present (refers to OT-2 slot 12 trash addressable area) 

Screen Shot 2023-12-19 at 3 27 55 PM

# Changelog

- filter loadLabware command specific to fixedTrash in v8 migration file

[RAUT-908]: https://opentrons.atlassian.net/browse/RAUT-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/protocol-designer/fixtures/protocol/8/doItAllV3MigratedToV8.json
+++ b/protocol-designer/fixtures/protocol/8/doItAllV3MigratedToV8.json
@@ -1198,49 +1198,6 @@
       "schemaVersion": 2,
       "cornerOffsetFromSlot": { "x": 0, "y": 0, "z": 0 }
     },
-    "opentrons/opentrons_1_trash_1100ml_fixed/1": {
-      "ordering": [["A1"]],
-      "metadata": {
-        "displayCategory": "trash",
-        "displayVolumeUnits": "mL",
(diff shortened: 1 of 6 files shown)


jerader (approved):
wow suddenly makes a lot of sense why those trash definitions were included in those fixtures, thanks so much for fixing this!

PR #14260, opened 2023-12-21
fix(app): solve nested folder structure in pages

koji:
# Overview
solve nested folder structure in pages and add test.
remove `OnDeviceDisplay` folder
I noticed that we are missing a few page components' tests. I will follow up them in another PR.
Update a few tests that VSCode shows warnings with what Brian told us last week.

```ts
const [{ getByText }] = render()
getByText('something')

render()
screen.getByText('something') 

// click() -> fireEvent.click(targetElement)
```

# Test Plan

# Changelog
- move components under `pages/OnDeviceDisplay` to `pages`
- update paths in tests
- remove `OnDeviceDisplay` folder

# Review requests
Do we want to keep `ODDTempMenu`?

# Risk assessment
low


shlokamin (commented):
im not sure that we want to change local imports from within a module to import from the public (external) package name. and we certainly don't want to import from the public package AND reach into private directories. 

i.e. from within the app, we shouldn't be doing something like this:

`import { ConnectViaEthernet } from '@opentrons/app/src/pages/ConnectViaEthernet'`

should either be `import { ConnectViaEthernet } from '@opentrons/app'` or a relative import. and i'd prefer a relative import because, well, we're importing from within the same package!

how come we're not going with relative imports anymore?

koji:
> im not sure that we want to change local imports from within a module to import from the public (external) package name. and we certainly don't want to import from the public package AND reach into private directories.
> 
> i.e. from within the app, we shouldn't be doing something like this:
> 
> `import { ConnectViaEthernet } from '@opentrons/app/src/pages/ConnectViaEthernet'`
> 
> should either be `import { ConnectViaEthernet } from '@opentrons/app'` or a relative import. and i'd prefer a relative import because, well, we're importing from within the same package!
> 
> how come we're not going with relative imports anymore?

I think the paths were changed by VSCode, I will revert them.

app/src/pages/TempODDMenu/index.tsx
```diff
@@ -4,17 +4,17 @@ import { useDispatch, useSelector } from 'react-redux'
...
   startRobotUpdate,
-} from '../../../redux/robot-update'
-import { getLocalRobot } from '../../../redux/discovery'
+} from '../../redux/robot-update'
+import { getLocalRobot } from '../../redux/discovery'
 
-import type { RouteProps } from '../../../App/types'
+import type { RouteProps } from '../../App/types'
 
 export function TempODDMenu(): JSX.Element {
```
brenthagen: also probably time to delete the temp menu 😃

brenthagen (approved):
changes make sense to me, smoke tested the app without issue

app/src/pages/TempODDMenu/index.tsx
koji: I will delete that.

PR #14263, opened 2023-12-21
chore: release notes for 7.1.1

sfoster1:
Add release notes for 7.1.1.

PR #14271, opened 2024-01-02
chore: release notes for 7.1.1

sfoster1:
Adds quick release notes for 7.1.1.


ecormany:
This PR duplicates #14263, which has all the same info plus another commit I added. Closing in favor of that one. It's still open, so we can make any further changes there.

PR #14277, opened 2024-01-03
Analyses Snapshot Testing 

y3rsh:
# Overview

Create a test that snapshots analyses output.  This will detect expected and unexpected changes to analyses.  Run this test nightly against edge.  Once mature, consider adding it to pull requests against `api`, `hardware`, and `shared-data/python`. 

## Test Plan

This test was exercised heavily during the 7.1.0 release cycle.  @y3rsh wants to merge this to edge so that we may easily run the workflow dispatch and so the nightly runs begin. 

@y3rsh will be babysitting the nightly run for some time to discover any flakiness and/or developer pain points when change is detected.  We also desire to try and add a deduplicated change summary to the PR description that might help speed up understanding of the changes.

[Link to a successful run](https://github.com/Opentrons/opentrons/actions/runs/7402315817)
[Link to an automated PR on changes detected](https://github.com/Opentrons/opentrons/pull/14240)

## Risk assessment

Low, all testing tools, no application code changed.


nusrat813 (approved):
LGTM

app-testing/Makefile
```diff
@@ -32,24 +32,45 @@ format:
...
 	yarn prettier --ignore-path .eslintignore --write app-testing/**/*.md
 
 .PHONY: print-protocols
 print-protocols:
-	pipenv run python print_protocols.py
+	python -m pipenv run python print_protocols.py
+
+.PHONY: install-pipenv
+install-pipenv:
+	python -m pip install -U pipenv
```
SyntaxColoring, line 59: Beware that this will interfere with the top-level `make setup`, which installs a specific version of `pipenv`.

But I see that this project is excluded from the top-level `make setup`. So as long as you either only work on this project, or only work on other projects, things will be fine...it's only if you go back and forth that `pipenv` will be overwritten. Do you have any opinions on how to resolve that in the long term, if it does need to be resolved?

app-testing/Makefile
```diff
@@ -32,24 +32,45 @@ format:
...
 test-ci:
-	pipenv run python -m pytest -m "emulated_alpha"
+	python -m pipenv run python -m pytest -m "emulated_alpha"
 
 .PHONY: test-protocol-analysis
 test-protocol-analysis:
 	pipenv run python -m pytest -v tests/protocol_analyze_test.py
 
 .PHONY: setup
 setup:
```
SyntaxColoring: I don't think it actually matters in this case, but as a general best practice, if `install-pipenv` needs to happen before `setup`, the "proper way" to express that in Make is as a dependency, instead of putting `$(MAKE) install-pipenv` in the recipe.

```suggestion
setup: install-pipenv
```

https://accu.org/journals/overload/14/71/miller_2004/

app-testing/Makefile
```diff
@@ -32,24 +32,45 @@ format:
...
-	pipenv install
+	$(MAKE) install-pipenv
```
SyntaxColoring:
```suggestion
```

app-testing/citools/generate_analyses.py
```diff
@@ -0,0 +1,291 @@
...
+    # tag = args.tag
+    # console.print(f"Received tag: {tag}")
+    # tag = "v7.0.2"
+    # tag = "v7.1.0-alpha.1"
+    # tag = "v7.1.0-alpha.2"
+    # tag = "v7.1.0-alpha.3"
+    # base_tag = "v7.0.2"
+    new_release_tag = "v7.1.0-alpha.3"
+    # analyze_against_image(base_tag)
+    analyze_against_image(new_release_tag)
```
SyntaxColoring: Is this code intentionally commented out and is this intentionally using a hard-coded `v7.1.0-alpha.3` tag?

SyntaxColoring (approved):
Cool!

DerekMaggio (approved):
Everything looks great. I love the created PR. 

Are you planning an automated way to notify developers that there is a discrepancy that needs investigation? 
Are there any cases we would want to exclude a protocol from being part of the analysis snapshot testing?

app-testing/citools/generate_analyses.py
y3rsh: Yes.  This code allowed executing the file directly for development.  I will clean it up.

app-testing/Makefile
y3rsh: The development setup instructions in the README for app-testing recommend using pyenv to set Python 3.12 as the local version in the project folder. I intend to keep app-testing updated with the latest Python version, which should not conflict with the Python version used in the monorepo. So far, pyenv has seamlessly managed switching between different Python versions on my local machine.

If any issues arise, or if there's a need for closer integration with the monorepo's Python environment, we can address and resolve these challenges as they come.

PR #14280, opened 2024-01-03
feat(notify-server): Delete notify-server

SyntaxColoring:
# Overview

This deletes the `notify-server` project.

Years ago, we added `notify-server` because we thought it would be helpful for getting the robot to provide realtime notifications to clients. However, shifting priorities meant that it never developed beyond a proof-of-concept. It let you listen for door-opened/door-closed events, but that's it.

Realtime notifications are getting renewed interest, but we no longer think `notify-server` is the way to do them.

Deleting the project saves some CI time and relieves us from maintaining its dependencies.

This PR has an OT-2 Buildroot co-dependency: https://github.com/Opentrons/buildroot/pull/215

# Test Plan

Grab images from the build server (`ot2-system.zip`/`system-update.zip`) and push them to an OT-2 and Flex. Make sure robot-server still starts, and `journalctl` doesn't show any systemd errors complaining about a missing unit or anything like that.

# Changelog

The user-facing changes entailed by deleting this project are:

* Remove the `/notifications/subscribe` WebSocket endpoint. As far as I can tell, it was not publicly documented.
* Remove the `notify_server` package and script from the robot. These were described in `notify-server/README.rst`, but otherwise not advertised, as far as I can tell.

There is currently no replacement for these removals, although we are planning one. Meanwhile, I recommend simply polling `GET /robot/door/status`.

# Review requests

If anyone has reason to believe that anyone cares about these features, speak now or forever fold your pizza.

# Risk assessment

Very low, assuming no one cares about these features.


robot-server/Pipfile.lock
```diff
@@ -1446,13 +1277,6 @@
             "markers": "python_version >= '3.6'",
             "version": "==3.5.2"
         },
-        "systemd-python": {
-            "hashes": [
-                "sha256:fd0e44bf70eadae45aadc292cb0a7eb5b0b6372cd1b391228047d33895db83e7"
-            ],
-            "markers": "sys_platform == 'linux'",
-            "version": "==234"
-        },
```
sanni-t: Interesting that this can be removed too. I thought we still need this package for systemd event communication during robot-server startup..

sanni-t (approved):
This looks good and shouldn't affect robot operations but still we should check on robots (esp the systemd notifications part). I can do an OT2 test if you haven't already.

robot-server/Pipfile.lock
SyntaxColoring: Uhh you are right, good catch. I’ll see what’s going on here.

SyntaxColoring:
I’m seeing robot server startup failures. I think they’re related to the general dependency upgrades, outside of this PR, but we should hold off on merging until that’s confirmed.

robot-server/Pipfile.lock
SyntaxColoring: Despite this removal, [an identical copy remains elsewhere in the file](https://github.com/Opentrons/opentrons/blob/0bb423f845d7e7b71e7181319e7aa51d763d47f4/robot-server/Pipfile.lock#L239-L245).

I'm guessing there were duplicates in there for some reason, and Pipenv is cleaning it up now.

sanni-t: Oh that's interesting! Good find!

PR #14284, opened 2024-01-04
docs(api): add missing docstrings for Well properties

ecormany:
# Overview

A handful of Well properties that have been in the API for years never got documented.

# Test Plan

Check out the [API Reference entries in the sandbox](http://sandbox.docs.opentrons.com/docs-well-properties/v2/new_protocol_api.html#wells-and-liquids).

# Changelog

Brand new docstrings for:
- `Well.parent`
- `Well.max_volume`
- `Well.display_name`
- `Well.well_name`

# Review requests

Words good? These all _should_ be documented, yes?

# Risk assessment

none, docs only

### DIFF
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -175,9 +175,12 @@ class Well:
     @property
     @requires_version(2, 7)
     def well_name(self) -> str:
-        """A human-readable name for the well's coordinates.
+        """A string representing the well's coordinates.
 
         For example, "A1" or "H12".
+
+        The format of strings that this property returns is the same format as the key
+        for :ref:`accessing wells in a dictionary <well-dictionary-access>`.
         """
         return self._core.get_name()
 
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -93,6 +93,7 @@ class Well:
     @property
     @requires_version(2, 0)
     def parent(self) -> Labware:
+        """The :py:class:`.Labware` object that the well is a part of."""
         return self._parent
 
     @property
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_api/labware.py
```diff
@@ -159,11 +164,21 @@ def depth(self) -> float:
...
+        """
         return self._core.get_display_name()
 
     @property  # type: ignore
     @requires_version(2, 7)
     def well_name(self) -> str:
+        """A human-readable name for the well's coordinates.
+
+        For example, "A1" or "H12".
+        """
```
SyntaxColoring: Perhaps more importantly than being human-readable, this is the canonical name of the well used by `labware[well_name]` and `labware.wells_by_name()[well_name]`.

SyntaxColoring (approved):
Thanks!

api/src/opentrons/protocol_api/labware.py
ecormany: Ah yes, that's a relevant connection! I've added some two way references between this entry and the Dictionary Access sub-section of the Labware page.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -113,6 +114,10 @@ def has_tip(self, value: bool) -> None:
 
     @property
     def max_volume(self) -> float:
+        """The maximum volume, in µL, that the well can hold.
+
+        This amount is set by the JSON labware definition, specifically the ``totalLiquidVolume`` property of the particular well.
```
jwwojak, line 119: Is it worth linking to the schema for this property? 

https://github.com/Opentrons/opentrons/blob/335200315ea2b81f8c02f854891582b5b125d6c3/shared-data/labware/schemas/2.json#L253

> Total well, tube, or tip volume in microliters

The description on line 117 already covers this. Maybe redundant.

jwwojak (approved):
Left a comment. LGTM.

:shipit:

api/src/opentrons/protocol_api/labware.py
ecormany: Not going to link. Users don't really need to reference the labware schema to use this property. It's more for users who are curious how the API figures out what to return.

PR #14285, opened 2024-01-04
fix(app): remove firmware update step of module calibration flow

smb2268:
fix RQA-1933

# Overview
Module calibration flow cannot be launched unless there is a pipette attached, updated, and calibrated, ready to perform module cal. Because of this, we don't need to check for a pipette firmware update before performing module calibration. We also don't need to check for module firmware updates here since an update would not affect the calibration value. So, this PR removes the firmware update step from module calibration wizard flows.

# Test Plan
1. Attach a pipette that needs a firmware update with no other pipettes on the gantry, ensure you are unable to launch module calibration - overflow menu option should be disabled with tooltip indicating a pipette needs to be attached/calibrated  
2. Run through module cal, ensure you no longer see "firmware up to date" message

# Changelog
Remove firmware update modal from module calibration wizard flow

# Review requests
Make sure I didn't miss ripping anything out 

# Risk assessment
Low


mjhuff (approved):
looks correct to me!

PR #14286, opened 2024-01-04
perf(robot-server): Discriminate command unions

SyntaxColoring:
# Overview

Now that we have Pydantic 1.9, we can close RSS-128. This dramatically improves error reporting when you misspell something in a Protocol Engine command, and it also improves performance anywhere we're parsing Protocol Engine commands.

# Test Plan

For "this doesn't break anything," I'm mostly leaning on integration tests.

* [x] To do: Also push this to an OT-2 to make sure the new `typing.Annotated` usage hasn't broken anything across Python versions.

For improved error reporting:

* I've manually used Postman to check `POST /runs/commands`
* I've manually used Postman to check `POST /maintenance_runs/commands`
* I've manually used Postman to check `POST /commands`

For improved performance:

* I've manually used Postman on a dev server to check the speed of `POST /runs` with a JSON protocol. Seems ~20-30% better.
* The slowest test in `tests/integration/http_api/persistence/test_compatibility.py` goes from ~23s to ~16s.

# Changelog

* Annotate our command unions as being discriminated by their `commandType` field. See the Pydantic documentation: https://docs.pydantic.dev/1.10/usage/types/#discriminated-unions-aka-tagged-unions
* Unfortunately, before Pydantic v2, there appear to be bugs that prevent this from working when the union appears inside a `pydantic.Generic` model. To work around this, I unfortunately had to make some non-generic duplicates of our `RequestModel`, which is unfortunate.

# Review requests

* Are there any HTTP endpoints I missed where we're parsing Protocol Engine commands?
* Can you think of a better workaround for the `pydantic.Generic` thing?

# Risk assessment

Low—if this broke anything, I think integration tests would cover it, unless there's weirdness across Python versions.


jbleon95:
Are the changes made to the command schema compatible with the `make command-schema` target? Mostly in that it doesn't overwrite that information, but also if it would (or can) add that discriminator information automatically.

SyntaxColoring:
> Are the changes made to the command schema compatible with the make command-schema target? Mostly in that it doesn't overwrite that information, but also if it would (or can) add that discriminator information automatically.

Yep, changes to `8.json` in this PR were generated automatically via `make -C api command-schema COMMAND_SCHEMA_VERSION=8 && make format-js`.

jbleon95 (approved):
Looks great, glad we could finally get to this with the Python updates

PR #14287, opened 2024-01-04
feat(api): add deck conflict checks for staging slots and after moving labware

jbleon95:
# Overview

Addresses part of RSS-421 and all of RSS-303.

This PR adds staging slots for deck conflict checks in PAPI, which covers loading multiple labware in the same staging slot, as well as using a staging slot when a module has been loaded in the adjacent deck slot, not including the magnetic block (as well as the thermocycler which cannot be adjacent to a staging slot).

Additionally, a deck conflict check has been added to `move_labware`. Due to current limitations of how that is architected, it has to take place after execution of the command. In practice, this won't cause any additional issues due to errors being fatal, and even catching and ignoring them will have the same functionality as not-checking them right now. But this should be fixed in a future PR, whether by re-architecting it to be able to be used before command execution, or moving the logic entirely into PE.

# Test Plan

The following protocols with impossible deck conflicts will now correctly raise deck conflict errors during analysis.

```python
requirements = {
    'robotType': 'Flex',
    'apiLevel': '2.16'
}


def run(protocol):
    temp_module = protocol.load_module('temperatureModuleV2', location="D3")

    # This should raise
    pcr_plate = protocol.load_labware('opentrons_96_wellplate_200ul_pcr_full_skirt', "D4")
```

```python
requirements = {
    'robotType': 'Flex',
    'apiLevel': '2.16'
}


def run(protocol):
    pcr_plate = protocol.load_labware('opentrons_96_wellplate_200ul_pcr_full_skirt', "D4")

    # This should raise
    temp_module = protocol.load_module('temperatureModuleV2', location="D3")
```

```python
requirements = {
    'robotType': 'Flex',
    'apiLevel': '2.16'
}


def run(protocol):
    pcr_plate = protocol.load_labware('opentrons_96_wellplate_200ul_pcr_full_skirt', "D1")

    temp_module = protocol.load_module('temperatureModuleV2', location="D3")

    # This should raise
    protocol.move_labware(pcr_plate, "D4", use_gripper=True)
```

# Changelog

- Added `StagingSlotName` as a valid location in deck conflict checking
- Added checks for modules placed adjacent to labware on staging slots (and vice-versa)
- Added a `deck_conflict. MagneticBlockModule` type to deck conflict checking so that it does not get incorrectly marked as incompatible near a staging slot
- Deck conflict checking now occurs after moving a labware.
 
# Review requests

# Risk assessment

Low-medium, the change with the biggest impact is adding deck conflict checking to `move_labware`, but it's re-using the existing logic and since the Flex has very few sources of deck conflicts, it shouldn't make any valid protocol in the field suddenly fail analysis.


api/src/opentrons/motion_planning/adjacent_slots_getters.py
```diff
@@ -38,18 +38,30 @@ def get_west_slot(slot: int) -> Optional[int]:
...
+    StagingSlotName.SLOT_C4: DeckSlotName.SLOT_C3,
+    StagingSlotName.SLOT_D4: DeckSlotName.SLOT_D3,
+}
+
+_EAST_OF_FLEX_COLUMN_3_MAP = {
+    DeckSlotName.SLOT_A3: StagingSlotName.SLOT_A4,
+    DeckSlotName.SLOT_B3: StagingSlotName.SLOT_B4,
+    DeckSlotName.SLOT_C3: StagingSlotName.SLOT_C4,
+    DeckSlotName.SLOT_D3: StagingSlotName.SLOT_D4,
 }
```
SyntaxColoring: You can use [this pattern](https://github.com/Opentrons/opentrons/blob/4f8a98383fa95d116c3f97364510531c0303a9ab/api/src/opentrons/types.py#L343-L359) to reduce this duplication, if you want. Since there are only 4 elements, it's basically a matter of personal preference.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -210,10 +217,12 @@ def check(
...
 
 
-def _create_ot2_restrictions(
-    item: DeckItem, location: DeckSlotName
+def _create_ot2_restrictions(  # noqa: C901
+    item: DeckItem, location: Union[DeckSlotName, StagingSlotName]
 ) -> List[_DeckRestriction]:
     restrictions: List[_DeckRestriction] = []
+    if isinstance(location, StagingSlotName):
+        raise DeckConflictError("OT-2 does not support staging slots.")
```
SyntaxColoring: Let's include the problematic slot in the message.

```suggestion
    if isinstance(location, StagingSlotName):
        raise DeckConflictError(f"OT-2 does not support staging slots ({location.id}).")
```

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -274,20 +283,31 @@ def _create_ot2_restrictions(
...
+            location=location,
+            source_item=item,
+            source_location=location,
         )
+    ]
+
+    # Magnetic block module does not have staging slot restriction
+    if isinstance(item, (HeaterShakerModule, OtherModule)):
+        if isinstance(location, StagingSlotName):
+            raise DeckConflictError("Cannot have a module loaded on a staging slot.")
```
SyntaxColoring: Nitpick to match what I see in the manual.

```suggestion
            raise DeckConflictError("Cannot have a module loaded on a staging area slot.")
```

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -274,20 +283,31 @@ def _create_ot2_restrictions(
...
+        adjacent_staging_slot = get_adjacent_staging_slot(location)
+        if adjacent_staging_slot is not None:
+            restrictions.append(
+                _NothingAllowed(
+                    location=adjacent_staging_slot,
+                    source_item=item,
+                    source_location=location,
+                )
+            )
```
SyntaxColoring: This would be a good place for an explanatory comment. Maybe something like:

```suggestion
            # You can't have anything on a staging area slot next to a module because
            # the module caddy physically blocks you from haveing that staging area
            # slot in the first place.
            restrictions.append(
                _NothingAllowed(
                    location=adjacent_staging_slot,
                    source_item=item,
                    source_location=location,
                )
            )
```

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -398,20 +400,37 @@ def _is_within_pipette_extents(
...
+]:
     location_from_engine = engine_state.labware.get_location(labware_id=labware_id)
 
     if isinstance(location_from_engine, AddressableAreaLocation):
-        # TODO need to deal with staging slots, which will raise the value error we are returning None with below
+        # This will be guaranteed to be either deck slot name or staging slot name
+        slot: Union[DeckSlotName, StagingSlotName]
         try:
-            deck_slot = DeckSlotName.from_primitive(
+            slot = DeckSlotName.from_primitive(location_from_engine.addressableAreaName)
```
SyntaxColoring, line 412: For my edification, is this actually reachable in practice? For example, if you do `moveLabware` with `newLocation` as an `AddressableAreaLocation`, Protocol Engine doesn't normalize to a `DeckSlotLocation`?

api/src/opentrons/protocol_api/core/legacy/deck.py
```diff
@@ -175,7 +175,7 @@ def __setitem__(self, key: DeckLocation, val: DeckItem) -> None:
 
         # will raise DeckConflictError if items conflict
         deck_conflict.check(
-            existing_items=existing_items,
+            existing_items=existing_items,  # type: ignore [arg-type]
```
SyntaxColoring: Instead of a blanket `# type: ignore[arg-type]`, let's work around this [type invariance quirk](https://github.com/python/typing/issues/445) by annotating `existing_items` as a `Mapping[Union[DeckSlotName, StagingSlotName], deck_conflict.DeckItem]`. Mypy needs to know up-front that the keys can be `StagingSlotName`s (even though they never will be) because it's worried that `.check()` will call `__getitem__()` with a `StagingSlotName`.

SyntaxColoring (commented):
Looks great, just some minor suggestions.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
jbleon95: It does not, it keeps the location to what the client sent

SyntaxColoring: 👍 OK, we should probably get it to normalize it at some point.

api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -274,20 +283,35 @@ def _create_ot2_restrictions(
...
+        if adjacent_staging_slot is not None:
+            # You can't have anything on a staging area slot next to a heater-shaker or
+            # temperature module because the module caddy physically blocks you from having
+            # that staging area slot installed in the first place.
+            restrictions.append(
+                _NothingAllowed(
+                    location=adjacent_staging_slot,
+                    source_item=item,
+                    source_location=location,
+                )
```
sanni-t, line 311: We will never actually get to a state where module is loaded next to a staging slot, will we? Since the engine execution of all equipment load commands happens before the API deck conflicts are checked, this restriction is technically never going to be added, right? In which case, is it actually necessary to add?

sanni-t: Got this concern clarified on call with @jbleon95 . I was assuming that we check for conflicts between modules & fixtures when checking deck configuration conflicts in the engine. Since that is actually not the case, this check is necessary.

SyntaxColoring (approved):
I'm curious about @sanni-t's point above, but that notwithstanding, this LGTM. Thanks!

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -342,6 +339,18 @@ def move_labware(
...
             self.set_last_location(location=None, mount=Mount.EXTENSION)
 
+        # FIXME(jbl, 2024-01-04) deck conflict after execution logic issue, read notes in load_labware for more info:
+        deck_conflict.check(
+            engine_state=self._engine_client.state,
+            new_labware_id=labware_core.labware_id,
+            existing_labware_ids=[
+                labware_id
+                for labware_id in self._labware_cores_by_id
+                if labware_id != labware_core.labware_id
```
sanni-t, line 349: 👍

api/src/opentrons/motion_planning/deck_conflict.py
jbleon95: There's currently no engine restrictions on loading module near staging slot, so this logic will be hit here even after PE execution

sanni-t (approved):
✅ LGTM! Very well-thought restrictions!
For a future work- I would love to get some integration tests that check validity of different deck layouts and labware movement. They're pretty useful.

PR #14292, opened 2024-01-05
fix(shared-data): install the 'data' directory in the right place

ryanthecoder:
# Overview
after changing to the new python dist the /data folder was nested in as
/usr/lib/python3.10/site-packages/opentrons-shared-data/opentrons-shared-data/data
instead of 
/usr/lib/python3.10/site-packages/opentrons-shared-data/data

# Test Plan

# Changelog

# Review requests

# Risk assessment


vegano1 (approved):
nice

PR #14293, opened 2024-01-05
fix(app): restrict add fixture modal options when provided

brenthagen:
# Overview

when add fixture modal options are provided (during protocol setup on ODD), only show those options and do not allow the modal to close without selecting an option

closes RAUT-911

Screen Shot 2024-01-05 at 4 23 12 PM

# Test Plan

 - verified and added a unit test

# Changelog

 - Restricts add fixture modal options when provided

# Review requests

check the ODD protocol setup add fixture modal for a fixture not configured in D3

# Risk assessment

low


koji (approved):
LGTM!

PR #14294, opened 2024-01-06
docs(robot-server): Audit HTTP API documentation

SyntaxColoring:
# Overview

This closes RTC-249. Now that the HTTP API documentation is published, we want to be stricter about correctness, clarity, and currency.

# Review requests

We want to be able to advertise the HTTP API docs unabashedly, the same way we advertise [the Python Protocol API docs](https://docs.opentrons.com/v2/).

I think this will go best if we give it some time and close attention now and get it over with.

So, please closely read through [the sandbox build of this branch](http://sandbox.docs.opentrons.com/audit_openapi_docs/http/api_reference.html). Ideally, all of it. It's not as lengthy as that might sound. You can ignore the `v1` category because it's accidentally redundant; its contents already show up under other categories.

Try to counteract our collective habit of ignoring the old endpoints that we haven't touched recently. Imagine you're a 3rd-party engineer encountering this API for the first time.

The scope for this PR is that the docs *must* actively discourage you from using:

* Anything that's not tested by Opentrons
* Anything that's known by Opentrons to misbehave
* Anything that, if it turns out to be buggy, Opentrons will not prioritize a fix
* Anything that Opentrons plans to remove

And the docs *should* actively discourage you from using:

* Things that are probably a waste of time to try to get working as a 3rd-party developer, such as the OT-2's calibration sessions
* Anything else that would make Opentrons cringe

Does this accomplish that, or have I missed anything?

Style fixes are good to note, but out of scope for this PR. (I have some in the diff because I couldn't resist, sorry.)

# Changelog

* Delete the `/docs` page, which was rendered by Swagger UI. We tend to prefer the alternative UI at `/redoc`, and that's what we publish at https://docs.opentrons.com/http/api_reference.html, so let's focus our efforts on that.
* Many small deprecations and warnings.

# Risk assessment

No risk. Docs only.


robot-server/robot_server/service/legacy/routers/settings.py
```diff
@@ -229,7 +234,21 @@ async def get_settings_reset_options(
...
+    summary="Perform a factory reset",
+    description=(
+        "Perform a factory reset of some robot data."
+        "\n\n"
+        "The valid properties are given by `GET /settings/reset/options`."
+        "\n\n"
+        "You should always restart the robot after using this endpoint to"
+        " reset something."
+        "\n\n"
+        '**Warning:** Despite being called "factory reset" in Opentrons code,'
```
ecormany: Perhaps still true in code, but I think we've been pretty diligent about removing the word "factory" in UX copy. We say "Device Reset" in the app. Fortunately this endpoint name is just `/reset`.

robot-server/robot_server/docs_helpers/xrefs.py
```diff
@@ -0,0 +1,30 @@
+"""Generate cross-reference links between endpoints."""
```
SyntaxColoring: After thinking about this some more, this xref implementation is not worthwhile. We'll just live without xrefs for now.

robot-server/robot_server/service/legacy/routers/settings.py
SyntaxColoring: Thank you, that's helpful insight. I've removed mentions of "factory-reset" in the docs. It's just called "reset settings and data" now.

robot-server/robot_server/app_setup.py
SyntaxColoring, line 1: For the sake of having some organized and self-contained discussions, I'm going to open a bunch of threads on this file, even though they're talking about code that lives elsewhere.

SyntaxColoring, line 1: What should we do with `GET /motors/engaged` and `POST /motors/disengage`?

SyntaxColoring, line 1: Do we want to keep `POST /camera/picture` for the Flex? It does have a camera, right? Does that endpoint actually work on a Flex today?

SyntaxColoring, line 1: Do we want to warn people away from using `"intent": "setup"` commands, or is that cat out of the bag?

SyntaxColoring, line 1: Do we sufficiently define how `/runs` coexist with `/maintenance_runs/commands` and `/commands`? We have not always done a great job on the server of enforcing mutual exclusion. I guess I'm worried if there's a danger of somebody `POST`ing to two of these things at once, the server accidentally allowing it, and them relying on the dangerous behavior.

SyntaxColoring, line 1: Anyone know if the `GET /calibration/status` deck calibration endpoint is meaningful on a Flex?

SyntaxColoring, line 1: Anyone know what the deal is with `GET /settings/robot/`? Should people be using it? Should we deprecate it?

sfoster1: it has a camera, it should work, it may not

sfoster1: keep em?

sfoster1: i mean they're not really necessary on the flex but they don't really hurt

sfoster1: people can use it for settings, there's nothing that special about it

sfoster1 (approved):
Changes on this look good to me. I think there's more to be done, but let's do it in follow ups so we can get this merged.

PR #14300, opened 2024-01-09
(doc:) update python version from 3.7 to 3.10

koji:
# Overview
update python version from 3.7 to 3.10 in the dev setup related docs.

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/DEV_SETUP.md
+++ b/DEV_SETUP.md
@@ -124,10 +124,10 @@ Close and re-open your terminal to verify that `pyenv` is installed
 pyenv --version
 ```
 
-Now, install the required version of Python. Use the latest available version of `3.10.x`, which is `3.10.6` at the time of writing.
+Now, install the required version of Python. Use the latest available version of `3.10.x`, which is `3.10.13` at the time of writing.
 
 ```shell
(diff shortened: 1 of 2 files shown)


DEV_SETUP.md
```diff
@@ -124,10 +124,10 @@ Close and re-open your terminal to verify that `pyenv` is installed
 pyenv --version
 ```
 
-Now, install the required version of Python. Use the latest available version of `3.7.x`, which is `3.7.15` at the time of writing.
+Now, install the required version of Python. Use the latest available version of `3.10.x`, which is `3.10.6` at the time of writing.
 
 ```shell
-pyenv install 3.7.15
+pyenv install 3.10.6
```
ecormany: I've been using Python 3.10.13 with pyenv and the monorepo since we made this change.

ecormany (commented):
Changes make sense. Just check which patch version of Python 3.10 we're recommending. If it's not the latest available, we should say so.

ecormany (approved):
Looks good. Thanks, Koji!

PR #14302, opened 2024-01-09
refactor(protocol-designer): dropTipLocation defaulted through createPresavedStepForm

jerader:
addresses RAUT-921

# Overview

There was some brittleness with how the drop tip location was being auto-populated. this is because `React.useEffect` was being used. I refactored it to get populated through the `createPresavedStepForm` which is what we are currently doing to auto-populated the pipette.

Additionally, I added what the drop tip location should hydrate to

# Test Plan

Test with either ot-2 or Flex, add a transfer and mix step and make sure the drop tip location is populated correctly. If both trash bin and waste chute are attached, it should default to `wasteChute` first.

Additionally, upload the attached protocol. Add a waste chute and delete the trash bin. Go to the timeline tab and see that each step is populated correctly.

[Ribogreen assay 24 samples in 2 plates.json](https://github.com/Opentrons/opentrons/files/13878483/Ribogreen.assay.24.samples.in.2.plates.json)

# Changelog

- clean up the `dropTipField` component by removing the `React.useEffect` and `React.useState`
- create a new function `_patchDefaultDropTipLocation` to grab the default drop tip location and plug it into the `createPresavedStepForm`
- fix the `createPresavedStepForm` test
- add a `getDropTipLocation` fn for the `dropTipLocation` to hydrate to
- BONUS: there were a few errors in console regarding the substep highlights and timeline that i fixed

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/steplist/substepTimeline.ts
+++ b/protocol-designer/src/steplist/substepTimeline.ts
@@ -244,10 +244,6 @@ export const substepTimelineMultiChannel = (
           numChannels = 96
         } else if (nozzles === COLUMN) {
           numChannels = 8
-        } else {
-          console.error(
-            'we currently do not support other 96-channel configurations'
-          )
         }
         const wellsForTips =
           numChannels &&
--- a/protocol-designer/src/top-selectors/substep-highlight.ts
+++ b/protocol-designer/src/top-selectors/substep-highlight.ts
@@ -30,7 +30,7 @@ function _wellsForPipette(
     let channels: 8 | 96 = pipChannels
     if (nozzles === ALL) {
       channels = 96
-    } else if (nozzles === COLUMN) {
+    } else if (nozzles === COLUMN || pipChannels === 8) {
       channels = 8
     } else {
       console.error(`we don't support other 96-channel configurations yet`)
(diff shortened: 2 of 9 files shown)


protocol-designer/src/steplist/substepTimeline.ts
```diff
@@ -244,10 +244,6 @@ export const substepTimelineMultiChannel = (
           numChannels = 96
         } else if (nozzles === COLUMN) {
           numChannels = 8
-        } else {
-          console.error(
-            'we currently do not support other 96-channel configurations'
-          )
```
jerader, line 250: this was causing a console error since it didn't account for an 8 channel pipette, i am not sure how this bug didn't cause any issues tbh, so i'm glad it is being fixed now!

protocol-designer/src/top-selectors/substep-highlight.ts
```diff
@@ -30,7 +30,7 @@ function _wellsForPipette(
     let channels: 8 | 96 = pipChannels
     if (nozzles === ALL) {
       channels = 96
-    } else if (nozzles === COLUMN) {
+    } else if (nozzles === COLUMN || pipChannels === 8) {
```
jerader, line 33: also causing a console error here for not accounting for an 8 channel pipette

shlokamin (approved):
works as expected. question tho, when deleting a trash bin and adding a waste chute, when looking at the exported protocols saved step forms the drop tip location is this: 

`"dropTip_location": "6ddabeba-15d5-4044-959f-9d708611e4e1:trashBin",`

which is weird. the commands are right tho... do we need to clean up somewhere?

PR #14303, opened 2024-01-09
fix(protocol-designer): more space between wasteChute and slotLabel

jerader:
# Overview

A quick style change to leave more space between `wasteChute` and `slotLabel`. The `slotLabels` increased in `fontSize` last minute before 8.0 was cut, leaving barely any room between the `wasteChute` and `slotLabel`. This adds some space

Also the waste chute and trash bin have been different colors for 8.0 which is sort of ugly/annoying. So i fixed it up here. This change i think also happened last minute before 8.0 was cut.

# Test Plan

Create a flex protocol and add a waste chute. See that the slot label and waste chute space is big enough. See that the waste chute and trash bin are the same color

# Changelog

- increase the `Y` prop positioning
- change trash bin color fill to `grey.2` to match the waste chute

# Review requests

see test plan

# Risk assessment

low


koji (approved):
LGTM

PR #14304, opened 2024-01-09
fix(api): import coroutine from typing instead of collections.abc

sanni-t:
# Overview

On Python 3.10, `Coroutine` from collections.abc seem to be referring to a coroutine object instead of the Coroutine type. This was causing robot server failure on line 79 of `execution_manager.py`- 
```
DecoratedMethodReturningValue = TypeVar(
    "DecoratedMethodReturningValue",
    bound=Callable[..., Coroutine[None, None, DecoratedReturn]],
)
``` 
with the error- `TypeError: 'ABCMeta' object is not subscriptable`

This PR fixes that issue by importing the type from `typing` module instead.
Thanks to @jbleon95 for finding the solution.

# Test Plan

Push to a robot and verify that the robot server boots up.

(I've tested this on a Flex and verified it works)

# Risk assessment

None. Bug fix

### DIFF
--- a/api/src/opentrons/hardware_control/execution_manager.py
+++ b/api/src/opentrons/hardware_control/execution_manager.py
@@ -1,9 +1,8 @@
 import asyncio
 import functools
-from typing import Set, TypeVar, Type, cast, Callable, Any, overload
+from typing import Set, TypeVar, Type, cast, Callable, Any, overload, Coroutine
 from .types import ExecutionState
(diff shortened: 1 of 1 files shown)


SyntaxColoring (approved):
LGTM if it's been tested on a Flex. Thanks!

sfoster1 (approved):
Oh, nice, thank you!

PR #14305, opened 2024-01-09
feat(hardware): Add the NodeIds for the HEPA/UV module

vegano1:
# Overview

We are adding the HEPA/UV module, which talks to the Flex over CAN; this means we need two new NodeIds, one for the application and one for the bootloader. Unfortunately, the node id is 7 bits long in the arbitration bitmask that we use as the header for CAN messages, which means we can only have 7 unique bootloader_node_ids. This is because we perform a bitwise OR on the node_id with the value 0xF, which converts the node_id from something like 0x30 to 0x3e, which works well until you have node_ids that occupy the same 10s place as another node like 0x33, as this produces the same bootloader_node_id for more than one node_id, which is bad.

So instead of relying on this simple bitwise operation and treating the NodeId as just a 7-digit bitmask capable of only having 7 unique bootloaders, let's instead have a direct mapping between bootloader_node_ids and node_ids. This increases the number of unique NodeIds we can have to 125 ids ( 127d (1111111b) - 2 (broadcast, host) = 125). Since the firmware_binding/contants.py file is already the source of truth for NodeId values, and we don't actually do any multicasting using the 7-bit NodeId, we can just use these 7 bits as their decimal equivalent without causing issues as long as we keep the old NodeIds the same.

# Test Plan

- [x] Make sure that messages over CAN have the corresponding NodeId so there is no mixup.
- [x] Make sure that all axes still move without issues
- [x] Make sure that the head_l, head_r move up/down
- [x] Make sure the plunger for pipettes still works
- [x] Make sure the gripper_z and gripper_g still work

# Changelog

- Add node_id = 0x32 and bootloader_node_id = 0x3e for HEPA/UV module
- Add direct mapping between bootloader_node_ids and node_ids to move away from node_id 7bit limitation.
- Convert existing NodeId functions to use the new mapping mechanism.

# Review requests

Let's make sure this makes sense and that there aren't any deeper consequences from making this change, is there something obvious I'm missing, any other considerations?

# Risk assessment

High if the logic or assumptions are wrong

### DIFF
--- a/hardware/opentrons_hardware/firmware_bindings/constants.py
+++ b/hardware/opentrons_hardware/firmware_bindings/constants.py
@@ -24,12 +24,14 @@ class NodeId(int, Enum):
     gripper = 0x20
     gripper_z = 0x21
     gripper_g = 0x22
+    hepa_filter = 0x80
     pipette_left_bootloader = pipette_left | 0xF
     pipette_right_bootloader = pipette_right | 0xF
     gantry_x_bootloader = gantry_x | 0xF
     gantry_y_bootloader = gantry_y | 0xF
     head_bootloader = head | 0xF
     gripper_bootloader = gripper | 0xF
+    hepa_filter_bootloader = hepa_filter | 0xF
 
     def is_bootloader(self) -> bool:
         """Whether this node ID is a bootloader."""
--- a/hardware/opentrons_hardware/firmware_bindings/constants.py
+++ b/hardware/opentrons_hardware/firmware_bindings/constants.py
@@ -5,7 +5,7 @@ by default. Please do not unconditionally import things outside the python stand
 library.
 """
 from enum import Enum, unique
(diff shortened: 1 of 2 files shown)


sfoster1 (changes requested):
Unfortunately we actually only have 7 bits of node id (from [here](https://github.com/Opentrons/opentrons/blob/7a3b2da87c0899a88b823b64739db48e5a8d2248/hardware/opentrons_hardware/firmware_bindings/arbitration_id.py#L21)) so we can't set the bit for `0x80`. We may have to get a little creative and have the node id be like `0x31` and the bootloader id be like `0x3e` or something.

caila-marashaj (approved):
Nice!

sfoster1 (approved):
Great, thank you! Let's add some tests for this but otherwise looks great.

PR #14306, opened 2024-01-10
fix(app-testing): shorten shapshot filename

y3rsh:
## Overview

- On Windows the app-testing analyses snapshot files exceed the 260 character limit on filename.
- We could `git config --system core.longpaths true` but this would mean if you clone on a Windows machine without this setting it would fail.
- Instead, hash the test file name as the test id.
- In a future PR make a better mapping for troubleshooting which was why the filenames were being used in the first place.

### Failed

image

### Worked

image

## Risk assessment
None, testing code.


SyntaxColoring (approved):
TY!

PR #14313, opened 2024-01-12
fix: our packages must be in default pipfile sect

sfoster1:
Because those are the only ones installed by openembedded.


vegano1 (approved):
Nice, good catch

PR #14314, opened 2024-01-12
chore(app,app-shell,app-shell-odd, protocol-designer, labware-library, components, labware-designer): update Electron version to v27.0.0

koji:
# Overview
Update Electron version 27.0.0
https://www.electronjs.org/docs/latest/tutorial/electron-timelines#timeline

There is `loadURL` issue in the versions that include the following PR.
https://github.com/electron/electron/pull/40143

At this moment we cannot update any newer version. (I'll keep checking the repo)

This PR also includes Nodejs update because of Electron v27's requirement.
This PR needs Nodejs v21.4.0 since M1 Mac has an issue about running Desktop app. (I'll keep an eye on this to support more nodejs version). M2 Mac can use nodejs v18.

Need `NODE_OPTION` due to OpenSSL version.
I tried to export NODE_OPTION in the top of Makefile, but somehow that has an impact on other part and prevent building the app(Desktop/ODD). I use `NODE_OPTION` separately.
Another reasons for this are `labware-library` and `protocol-designer` since the option prevents Cypress.

### Running the application
node v21
- [x] macOS
- [x] Windows
- [x] Ubuntu
- [x] ODD mode
- [x] ODD

### CI
- [x] update Nodejs version (v16 -> v21)

### OT-2
- [x] USB connection

### Other applications
- [x] Protocol Designer
- [x] Labware Library
- [x] Labware Designer
- [x] Storybook

### Commands
- [x] make check-js
- [x] make lint-js
- [x] make test-js

This PR removes `usb-detection` since the repo has been archived and not compatible with node v18+.
The new package we need to use is `[node-usb](https://github.com/node-usb/node-usb)`
https://working-wormhole-d95.notion.site/usb-detection-node-usb-a6637f72f5d44f3ba9c7520b4f0da24f

# Test Plan
- Run Desktop app and check that the app works as well as the current edge
- Run ODD app and check that the app works as well as the current edge

# Changelog

# Review requests

# Risk assessment


koji:
- [x] check redux

koji:
- [x] fix cypress issue 
  - [x] labware-library
  - [x] protocol-designer

.github/workflows/api-test-lint-deploy.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
           fetch-depth: 0
       - uses: 'actions/setup-node@v3'
         with:
-          node-version: '16'
+          node-version: '21'
```
brenthagen: do we need to use 21, why not 20? Node uses an odd/even release pattern. even releases have long-term support and are recommended for production: https://nodejs.org/en/about/previous-releases

package.json
```diff
@@ -27,7 +27,7 @@
   },
   "packageManager": "yarn@1.22.19",
   "engines": {
-    "node": "^16.9.0"
+    "node": ">=18.14.0 <22.0.0"
```
brenthagen: should we pin to a version? I'm using `20.11.0`

brenthagen (commented):
I'm getting some strange behavior on this branch from locally simulated robots - reporting as unreachable but responding with 200s to various requests. I'm not sure why, digging a bit further.

Screen Shot 2024-01-26 at 11 25 47 AM

Screen Shot 2024-01-26 at 11 28 27 AM

.github/workflows/api-test-lint-deploy.yaml
koji: Ed tried to v20 and it didn't work on his end. But v21 worked on his end.

sfoster1:
Looks like flex builds will fail until we get node-native on >=18 https://github.com/Opentrons/oe-core/actions/runs/7659498340/job/20874747633

package.json
koji: yeah, fixing version would be safe.

.github/workflows/api-test-lint-deploy.yaml
mjhuff: Did we ever characterize Ed's error? I think trying to use 20 over 21 for the sake of LTS would be better if possible.

koji: There was no specific error which was the same as node v18 + M1 macOS.
logs about `shell` don't show up.

sfoster1:
Tracking the oe-core work here https://github.com/Opentrons/oe-core/pull/128

package.json
sfoster1: Can we actually not do that? For the following reasons:
- This version of node is _not_ the version of node that will be used inside electron; electron has its own node binary that runs at a specific version, which for 27 is 18.19 (see https://www.electronjs.org/docs/latest/tutorial/electron-timelines) and presumably patched to work on those osx versions
- the node that we specify here is therefore actually only run for executing dev tooling
   - we should probably run tests under the same version as electron, and should therefore prefer 18.19
   - but sometimes we can't so we allow the range

sfoster1:
That build succeeded: https://builds.opentrons.com/ot3-oe/7672229872/ot3-system.zip

sfoster1:
And it works, happy with this from the ODD side

koji:
- [x] try v18.19.0 again
 - [x] solve conflicts
 - [ ] fix CI errors

brenthagen:
> I'm getting some strange behavior on this branch from locally simulated robots - reporting as unreachable but responding with 200s to various requests. I'm not sure why, digging a bit further.

this is an issue with node-fetch and dns default result order https://opentrons.slack.com/archives/CSCLVUW3C/p1706568971026189

i've pushed a commit to fix

app-shell-odd/src/main.ts
```diff
@@ -27,6 +28,15 @@ import { registerNotify, closeAllNotifyConnections } from './notify'
 import type { BrowserWindow } from 'electron'
 import type { Dispatch, Logger } from './types'
 
+/**
+ * node 17 introduced a change to default IP resolving to prefer IPv6 which causes localhost requests to fail
+ * setting the default to IPv4 fixes the issue
+ * https://github.com/node-fetch/node-fetch/issues/1624
+ */
+// TODO(bh, 2024-1-30): @types/node needs to be updated to address this type error. updating @types/node will also require updating our typescript version
```
brenthagen, line 36: this is failing because our wildcard @types/node version is 15 in yarn.lock. i'm looking into updating our wildcard `@types/node` version and typescript in a separate commit.

labware-library/Makefile
```diff
@@ -22,17 +22,18 @@ clean:
...
-	webpack --profile
+	export NODE_OPTIONS=--openssl-legacy-provider && webpack --profile
 	node ./renderStatic.js
 
 # development assets server
 .PHONY: dev
 dev: export NODE_ENV := development
 dev:
-	webpack-dev-server --hot
+	export NODE_OPTIONS=--openssl-legacy-provider && webpack-dev-server --hot --host=::
```
sfoster1: the other side of the ipv4/v6 thing is that it also applies to cypress runs, and there we can't run our own js code, so let's just serve on v6 wildcard. you can test this by doing `curl --head '[::1]:8080/'` (note the single quotes - your shell will try and parse those brackets) when dev server is running, as well as `curl --head 127.0.0.1:8080/` and `curl --head http://localhost:8080/`

sfoster1:
- [ ] fix ci errors

labware-library/Makefile
brenthagen: think we need this in `app` also, i get same behavior locally on node 18 (but not 20 it seems).

and i guess labware-designer while we're at it

sfoster1 (approved):
Looks good to me! @koji I think you should have the honor of clicking merge!

PR #14315, opened 2024-01-12
docs(api): describe new `distribute()` volume clamping behavior

ecormany:
# Overview

Brings the docs up to date with changes in #14253.

Tagged "do not merge" because this has PAPI 2.17 features. This PR should eventually be retargeted on the feature branch for the first robot stack version that supports 2.17.

Addresses RTC-383.

# Test Plan

Check the [sandbox](http://sandbox.docs.opentrons.com/docs-dispense-note/v2/basic_commands/liquids.html#dispense).

# Changelog

- Rewrote note about dispense volumes and moved it up in the section (it's not really about `push_out` anymore)
  - Clamped down on "clamping"
- Updated bullet on Versioning page
- Cleaned up docstring for `volume` argument of `dispense()`

# Review requests

Is there anything more we'd like to say about dispense volumes vis-a-vis push out? I drafted this but I think it's not actually a good real-world example.

```
If you want to move the plunger a certain amount, regardless of how much has been
aspirated, you can calculate the values of ``volume`` and ``push_out`` based on the
:py:obj:`~InstrumentContext.current_volume` held in the pipette::

    pipette.pick_up_tip()
    plunger_amount=105
    pipette.aspirate(100, plate['A1'])
    pipette.dispense(
        volume=min(plunger_amount, pipette.current_volume), 
        location=plate['B1'], 
        push_out=max(pipette.current_volume-plunger_amount, 0)
    )   # dispense 100, push out 5
    pipette.aspirate(120, plate['A2'])
    pipette.dispense(
        volume=min(plunger_amount, pipette.current_volume), 
        location=plate['B2'], 
        push_out=max(pipette.current_volume-plunger_amount, 0)
    )   # dispense 105, push out 0, 15 left in tip
    pipette.drop_tip()
    
Note that when passing large ``push_out`` values, the plunger can't pass its physical
lower limit. For example, ``push_out=100`` will not have the effect of moving the
plunger the same distance as during a 100 µL dispense. For clearing large amounts
through the pipette, blow out instead.

```

# Risk assessment

v v low, docs and docstrings only.

### DIFF
--- a/api/docs/v2/basic_commands/liquids.rst
+++ b/api/docs/v2/basic_commands/liquids.rst
@@ -76,7 +76,9 @@ To dispense liquid from a pipette tip, call the :py:meth:`.InstrumentContext.dis
     pipette.dispense(200, plate["B1"])
 
 .. note::
-    In API version 2.16 and earlier, you could pass a ``volume`` argument to ``dispense()`` greater than what was aspirated into the pipette. The API automatically clamped this value, and the robot *would not* move the plunger lower than dispensing the exact amount. In version 2.17 and later, passing such values raises an error.
+    In API version 2.16 and earlier, you could pass a ``volume`` argument to ``dispense()`` greater than what was aspirated into the pipette. In this case, the API would ignore ``volume`` and dispense the pipette's :py:obj:`~.InstrumentContext.current_volume`. The robot *would not* move the plunger lower as a result.
+
+    In version 2.17 and later, passing such values raises an error.
 
     To move the plunger a small extra amount, add a :ref:`push out <push-out-dispense>`. Or to move it a large amount, use :ref:`blow out <blow-out>`.
 
--- a/api/docs/v2/basic_commands/liquids.rst
+++ b/api/docs/v2/basic_commands/liquids.rst
@@ -75,6 +75,11 @@ To dispense liquid from a pipette tip, call the :py:meth:`.InstrumentContext.dis
 
     pipette.dispense(200, plate["B1"])
 
+.. note::
+    In version 2.16 and earlier of the API, you could pass a ``volume`` argument to ``dispense()`` greater than what was aspirated into the pipette. The API automatically clamped this value, and the robot *would not* move the plunger lower than dispensing the exact amount. In version 2.17 and later, passing such values raises an error.
+
+    To move the plunger a small extra amount, add a :ref:`push out <push-out-dispense>`. Or to move it a large amount, use :ref:`blow out <blow-out>`.
+
 If the pipette doesn’t move, you can specify an additional dispense action without including a location. To demonstrate, this code snippet pauses the protocol, automatically resumes it, and dispense a second time from location B1.
 
 .. code-block:: python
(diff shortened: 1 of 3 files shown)


api/docs/v2/basic_commands/liquids.rst
```diff
@@ -75,6 +75,11 @@ To dispense liquid from a pipette tip, call the :py:meth:`.InstrumentContext.dis
 
     pipette.dispense(200, plate["B1"])
 
+.. note::
+    In version 2.16 and earlier of the API, you could pass a ``volume`` argument to ``dispense()`` greater than what was aspirated into the pipette. The API automatically clamped this value, and the robot *would not* move the plunger lower than dispensing the exact amount. In version 2.17 and later, passing such values raises an error.
```
jwwojak:
```suggestion
    In API version 2.16 and earlier, you could pass a ``volume`` argument to ``dispense()`` greater than what was aspirated into the pipette. The API automatically clamped this value, and the robot *would not* move the plunger lower than dispensing the exact amount. In version 2.17 and later, passing such values raises an error.
```

jwwojak: Unless "clamping" is discussed and described elsewhere, it doesn't have any meaning for the reader on a first mention like this. As a fix, there ought to be some description of "clamping" or use other words here to explain the behavior.

In fact, the text could even omit "clamping." Instead, this could be explained as the "robot ignores this value | command." The text doesn't have to explain "ignore" to the reader, but it does have to explain "clamping" because it is used as a substitute for words that don't require a definition, like "ignore," "disregard," "not pay attention to," etc.

Exegesis (indulge my moment of medieval scholasticism. Gratias!):

> In API version 2.16 and earlier,  you could pass a `volume` argument to `dispense()` greater than the amount of liquid aspirated into the pipette.

This nicely sets up the problem -- the API lets you do something contradictory, which is dispense more than was aspirated. It even sounds dangerous as the robot might try to push harder on the pipette plunger to force out liquid that's not in the tip. 

Next, the solution/resolution without "clamping:"
> In these cases, API automatically ignored the requested dispense volume, and the robot *would not* move the pipette plunger lower than the originally aspirated amount. 

Finally, tee up the change:
> In version 2.17 and later, passing such values raises an error.

ecormany: Yep, this matches usage elsewhere in the docs.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -290,15 +290,15 @@ def dispense(  # noqa: C901
 
         See :ref:`new-dispense` for more details and examples.
 
-        :param volume: The volume to dispense, measured in µL. If unspecified,
-                       defaults to :py:attr:`current_volume`. If only a volume is
+        :param volume: The volume to dispense, measured in µL. If only a volume is
                        passed, the pipette will dispense from its current position.
```
SyntaxColoring: It looks like we can remove "If only a volume is passed, the pipette will dispense from its current position" from here. It's already explained under `param location`, which is where it really belongs, anyway.

SyntaxColoring (approved):
All looks correct to me. Thanks!

PR #14319, opened 2024-01-16
fix(app): fix pause icon alignment in ODD command icon

ncdiehl11:
close [RQA-2030](https://opentrons.atlassian.net/browse/RQA-2030)

# Overview

Fix the ODD pause icon alignment in commands with text that span more than 1 line.

# Test Plan

1. Run a Flex protocol with a mult-line pause command (example)
2. Observe runlog on desktop and ODD
Screen Shot 2024-01-16 at 2 31 29 PM
Screen Shot 2024-01-16 at 2 19 47 PM

# Changelog

- add `svgWidth` prop to CommandIcon component
- pass proper `svgWidth` for desktop and ODD with respect to designs, accounting for possible multi-line CommandText 

# Risk assessment
low

[RQA-2030]: https://opentrons.atlassian.net/browse/RQA-2030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/RunPreview/CommandIcon.tsx
+++ b/app/src/organisms/RunPreview/CommandIcon.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { SPACING, Icon, IconName } from '@opentrons/components'
+import { Icon, IconName } from '@opentrons/components'
 import { RunTimeCommand } from '@opentrons/shared-data'
 import type { StyleProps } from '@opentrons/components'
 
@@ -14,7 +14,7 @@ interface CommandIconProps extends StyleProps {
   svgWidth?: string | number
 }
 export function CommandIcon(props: CommandIconProps): JSX.Element | null {
-  const { command, svgWidth, ...styleProps } = props
+  const { command, svgWidth = '1rem', ...styleProps } = props
   let iconName = null
   if (
     command.commandType === 'moveLabware' &&
--- a/app/src/organisms/RunPreview/CommandIcon.tsx
+++ b/app/src/organisms/RunPreview/CommandIcon.tsx
@@ -31,6 +31,6 @@ export function CommandIcon(props: CommandIconProps): JSX.Element | null {
   }
 
   return iconName != null ? (
-    <Icon {...styleProps} svgWidth={svgWidth} name={iconName} flex="0 0 auto" />
+    <Icon {...styleProps} size={svgWidth} name={iconName} flex="0 0 auto" />
   ) : null
 }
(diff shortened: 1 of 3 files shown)


app/src/organisms/RunPreview/CommandIcon.tsx
```diff
@@ -30,11 +31,6 @@ export function CommandIcon(props: CommandIconProps): JSX.Element | null {
   }
 
   return iconName != null ? (
-    <Icon
-      {...styleProps}
-      name={iconName}
-      size={SPACING.spacing20}
-      flex="0 0 auto"
-    />
+    <Icon {...styleProps} svgWidth={svgWidth} name={iconName} flex="0 0 auto" />
```
jerader: can you resize with the `size` attribute rather than using an `svgWidth`?

app/src/organisms/RunPreview/CommandIcon.tsx
```diff
@@ -30,11 +31,6 @@ export function CommandIcon(props: CommandIconProps): JSX.Element | null {
   }
 
   return iconName != null ? (
-    <Icon
-      {...styleProps}
-      name={iconName}
-      size={SPACING.spacing20}
-      flex="0 0 auto"
-    />
+    <Icon {...styleProps} size={svgWidth} name={iconName} flex="0 0 auto" />
```
jerader: can we rename the `svgWidth` prop to `size`? Since both `size` and `svgWidth` attributes exist, it is a bit confusing. 
```suggestion
    <Icon {...styleProps} size={size} name={iconName} flex="0 0 auto" />
```

app/src/organisms/RunPreview/index.tsx
```diff
@@ -131,7 +131,11 @@ export const RunPreviewComponent = (
                 `}
               >
                 <Flex alignItems={ALIGN_CENTER} gridGap={SPACING.spacing8}>
-                  <CommandIcon command={command} color={contentColor} />
+                  <CommandIcon
+                    command={command}
+                    color={contentColor}
+                    svgWidth="1rem"
```
jerader: can we use size here instead as well? Its just not a pattern we use currently in the repo to use `svgWidth`. So if `size` and `svgWidth` result in the same thing, lets stick with `size`!

jerader (approved):
left a few comments but once those are addressed, this lgtm!

PR #14320, opened 2024-01-16
feat(hardware-testing): Flex protocol to fill diluent for 96ch photometric test

andySigler:
# Overview

Now that we have certifiable 8ch P1000 pipettes, we can use them in production to pre-fill our photometric plates with diluent. This is very helpful for production team, because then they do not need to support having an OT2 in the volumetric testing room.

Note that it's not written as a simple protocol because we want to include some version of liquid height tracking, so that we can make sure the tips do not over-submerge into the diluent. Also, it was written to be a bit flexible in regards to the volume transferred, source/destination, and pipette used, but whatever.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
lgtm

PR #14321, opened 2024-01-16
feat(components): add special case SVG for labware adapters

smb2268:
fix RAUT-857, RQA-234

# Overview
Labware adapters don't have definitions that we can construct an SVG from. For these, we need to return a design-created SVG in the LabwareRender component
Screen Shot 2024-01-21 at 6 11 49 PM
Screen Shot 2024-01-22 at 10 36 27 AM
Screen Shot 2024-01-31 at 5 01 16 PM
Screen Shot 2024-01-31 at 4 56 44 PM

# Test Plan
1. View these SVGs in the Labware tab of the app
2. Look at protocols that use them and see how the deck map view looks - or look at them in the BaseDeck storybook

# Changelog
1. Added new `LabwareAdapter` component with a `loadName` to SVG component map. In `LabwareRender` if the labware loadName exists in that map, return `LabwareAdapter` instead of the definition constructed SVG
2. Updated the HeaterShaker SVG to more closely match the dimensions of the module
4. Added logic to rotate adapters that are screwed into a heater shaker depending on which side of the deck the module is situated
5. Added wells to the Magnetic Block svg as requested in RQA-2034

# Review requests
Look through test plan

# Risk assessment
Low


mjhuff (commented):
I'm seeing this:

Screenshot 2024-01-22 at 8 39 07 PM

When I test with this protocol:
[MagBlockWithHSAndAdapter.py.zip](https://github.com/Opentrons/opentrons/files/14017240/MagBlockWithHSAndAdapter.py.zip)

app/src/organisms/LabwareCard/index.tsx
```diff
@@ -35,6 +35,11 @@ export function LabwareCard(props: LabwareCardProps): JSX.Element {
   const displayName = definition?.metadata.displayName
   const displayCategory = startCase(definition.metadata.displayCategory)
   const isCustomDefinition = definition.namespace !== 'opentrons'
+  const xDimensionOverride =
+    definition.parameters.loadName === 'opentrons_universal_flat_adapter'
+      ? 127.4
```
koji: maybe defining a const?

components/src/hardware-sim/Labware/LabwareRender.tsx
```diff
@@ -58,6 +63,31 @@ export interface LabwareRenderProps {
 export const LabwareRender = (props: LabwareRenderProps): JSX.Element => {
   const { gRef } = props
   const cornerOffsetFromSlot = props.definition.cornerOffsetFromSlot
+  const labwareLoadName = props.definition.parameters.loadName
```
koji:
```suggestion
  const { gRef, definition } = props
  
  
  const labwareLoadName = definition.parameters.loadName
```

smb2268:
@mjhuff the issue in the screenshot you added should be fixed now! These are testable via the BaseDeck component in storybook if you change the `nestedLabwareDef.params.loadName` for the modules, I've tested with that pretty robustly but feel free to take another look with the protocol you were using!

mjhuff (approved):
Looks great! 🚀

PR #14323, opened 2024-01-16
feat(app,app-shell): create desktop app error boundary

brenthagen:
# Overview

this error boundary will capture a runtime error anywhere in the desktop app and show a general error modal with error message instead of an app whitescreen. on clicking the "Reload app" button, the app navigates to the root page and initiates an electron browser window reload via a new RELOAD_UI action that is registered in the app-shell ui module.

Confirmed modal design with Rob:

Screen Shot 2024-01-17 at 12 12 03 PM

# Test Plan

verified error boundary functionality by generating errors and reloading

# Changelog

 - Creates desktop app error boundary

# Review requests

put some errors in pages and components that would normally cause whitescreens and try out the error boundary reload.

# Risk assessment

low


app/src/App/DesktopAppFallback.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const trackEvent = useTrackEvent()
+  const dispatch = useDispatch<Dispatch>()
+  const history = useHistory()
+  const handleReloadClick = (): void => {
+    trackEvent({
+      name: ANALYTICS_DESKTOP_APP_ERROR,
+      properties: { errorMessage: error.message },
+    })
+    // route to the root page and initiate an electron browser window reload via app-shell
+    history.push('/')
```
mjhuff, line 36: Why do we need to push to root before the dispatch? Does this do something that I'm missing?

mjhuff (commented):
I like it! Any chance we can center the modal on the actual app, or is this something that has been discussed with Design already?

app/src/App/DesktopAppFallback.tsx
brenthagen: it's intended to be a bit more resilient - if we only reload, the app will reload the page that the router was on when the error occurred which may just trigger the error boundary again.

koji (approved):
LGTM
This will be helpful to work on the white screen issue. 
Thanks!

app/src/App/DesktopAppFallback.tsx
mjhuff: Gotcha, I'm surprised webContents.reload doesn't cause history to reset as well. This makes sense then 👍

PR #14324, opened 2024-01-17
fix(hardware-testing): 96ch diagnostics tip presence polarity

andySigler:
# Overview

Polarity when reading 96ch tip-presence sensor was switched at some point, and so the production script must also switch to match it.

# Test Plan

# Changelog

# Review requests

# Risk assessment


hardware-testing/hardware_testing/production_qc/ninety_six_assembly_qc_ot3/test_tip_sensor.py
```diff
@@ -82,14 +82,14 @@ async def run(api: OT3API, report: CSVReport, section: str) -> None:
     slot_5 = helpers_ot3.get_slot_calibration_square_position_ot3(5)
     home_pos = await api.gantry_position(OT3Mount.LEFT)
     await api.move_to(OT3Mount.LEFT, slot_5._replace(z=home_pos.z))
-    for expected_state in [False, True]:
+    for expected_state in [True, False]:
```
sfoster1, line 85: Wait if you switch this order but also switch the else/if order below this doesn't really do anything right? You're still expecting it to be `False` with no tips (but now that's second) and `True` with tips (but now that's first)

hardware-testing/hardware_testing/production_qc/ninety_six_assembly_qc_ot3/test_droplets.py
```diff
@@ -29,7 +29,7 @@
 TIP_RACK_96_SLOT = 4
 TIP_RACK_PARTIAL_SLOT = 5
 RESERVOIR_SLOT = 2
-TRASH_SLOT = 1
+TRASH_SLOT = 12
```
sfoster1, line 32: where's this one coming from

hardware-testing/hardware_testing/production_qc/ninety_six_assembly_qc_ot3/test_tip_sensor.py
Andiiiiiiyy: but actually, the responds of "get tip status" change, so the results will be different,  I want to do the same action to keep the SOP same.

hardware-testing/hardware_testing/production_qc/ninety_six_assembly_qc_ot3/test_droplets.py
andySigler: I had put the trash originally in slot 1 for this test, but looks like @Andiiiiiiyy prefers it in slot 12 which is fine by me

hardware-testing/hardware_testing/production_qc/ninety_six_assembly_qc_ot3/test_tip_sensor.py
andySigler: looks correct to me @sfoster1 . We are now first expecting `True`, and when it is true that means there are no tips. The ordering of the if/else is arbitrary

sfoster1: Oh. Right. Sorry

sfoster1 (approved):
Looks great to me!

PR #14325, opened 2024-01-17
feat(api): add deck conflict checks for trash bins loaded via addressable areas

jbleon95:
# Overview

Closes out RSS-421.

This PR adds the new trash bins, added in 2.16 and loaded via addressable area data, to deck conflict checking in PAPI. Previously all these checks were covered by the labware conflict checking, since the trash containers used to be labware. With the change away from that in the 7.1 release, these checks were lost and protocols that should not pass analysis were.

Now trash bins are taken into account when loading a thermocycler on the Flex (you cannot load a thermocycler when there is a trash in slot A1 and vice versa), and when loading a heater-shaker near the fixed trash on the OT-2. This also opens up potentially adding conflicts for the waste chute, but since none exist currently it is not taken into account.

# Test Plan

Verified that the following protocols now fail analysis.

```
metadata = {
    'protocolName': 'Thermocycler conflict 1',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}


def run(context):
    thermocycler = context.load_module("thermocyclerModuleV2")
    trash = context.load_trash_bin('A1')
```
```
metadata = {
    'protocolName': 'Thermocycler conflict 1',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}


def run(context):
    trash = context.load_trash_bin('A1')
    thermocycler = context.load_module("thermocyclerModuleV2")
```
```
metadata = {
    'protocolName': 'Heater-shaker conflict OT-2',
}

requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.16"
}


def run(context):
    # Both these lines should raise
    heater_shaker = context.load_module("heaterShakerModuleV1", '11')
    heater_shaker_2 = context.load_module("heaterShakerModuleV1", '9')
```

# Changelog

- add new `TrashBin` class to deck conflict check

# Review requests

# Risk assessment

Low, this introduces no new functional behavior and re-adds checks that were previously being done in lower API levels.

### DIFF
--- a/api/src/opentrons/motion_planning/deck_conflict.py
+++ b/api/src/opentrons/motion_planning/deck_conflict.py
@@ -246,7 +246,7 @@ def _create_ot2_restrictions(  # noqa: C901
             )
         )
 
-    if _is_fixed_trash(item):
+    if _is_ot2_fixed_trash(item):
         # A Heater-Shaker can't safely be placed just south of the fixed trash,
         # because the fixed trash blocks access to the screw that locks the
         # Heater-Shaker onto the deck.
@@ -391,7 +391,7 @@ def _flex_slots_covered_by_thermocycler() -> Set[DeckSlotName]:
     return {DeckSlotName.SLOT_B1, DeckSlotName.SLOT_A1}
 
 
-def _is_fixed_trash(item: DeckItem) -> bool:
+def _is_ot2_fixed_trash(item: DeckItem) -> bool:
     return (isinstance(item, Labware) and item.is_fixed_trash) or isinstance(
         item, TrashBin
     )
--- a/api/src/opentrons/motion_planning/deck_conflict.py
+++ b/api/src/opentrons/motion_planning/deck_conflict.py
@@ -61,6 +61,13 @@ class Labware:
 
 
 @dataclass
+class TrashBin:
+    """A non-labware trash bin (loaded via api level 2.16 and above)."""
+
+    name_for_errors: str
+
+
+@dataclass
 class _Module:
     name_for_errors: str
     highest_z_including_labware: float
@@ -98,6 +105,7 @@ DeckItem = Union[
     MagneticBlockModule,
     ThermocyclerModule,
     OtherModule,
+    TrashBin,
 ]
 
 
@@ -129,6 +137,10 @@ class _MaxHeight(NamedTuple):
                 return item.highest_z < self.max_height
         elif isinstance(item, _Module):
             return item.highest_z_including_labware < self.max_height
+        elif isinstance(item, TrashBin):
+            # Since this is a restriction for OT-2 only and OT-2 trashes exceeded the height limit, always return False
+            # TODO(jbl 2024-01-16) Include trash height and use that for check for more robustness
+            return False
 
 
 class _NoModule(NamedTuple):
@@ -380,4 +392,6 @@ def _flex_slots_covered_by_thermocycler() -> Set[DeckSlotName]:
 
 
 def _is_fixed_trash(item: DeckItem) -> bool:
-    return isinstance(item, Labware) and item.is_fixed_trash
+    return (isinstance(item, Labware) and item.is_fixed_trash) or isinstance(
+        item, TrashBin
+    )
--- a/api/src/opentrons/protocol_api/_trash_bin.py
+++ b/api/src/opentrons/protocol_api/_trash_bin.py
@@ -13,4 +13,10 @@ class TrashBin:
 
     @property
     def location(self) -> DeckSlotName:
+        """Location of the trash bin.
+
+        :meta private:
+
+        This is intended for Opentrons internal use only and is not a guarenteed API.
+        """
         return self._location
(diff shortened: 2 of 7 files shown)


api/src/opentrons/motion_planning/deck_conflict.py
```diff
@@ -380,4 +392,6 @@ def _flex_slots_covered_by_thermocycler() -> Set[DeckSlotName]:
 
 
 def _is_fixed_trash(item: DeckItem) -> bool:
-    return isinstance(item, Labware) and item.is_fixed_trash
+    return (isinstance(item, Labware) and item.is_fixed_trash) or isinstance(
+        item, TrashBin
+    )
```
SyntaxColoring: Does this helper function need to be renamed now? A `TrashBin` is not necessarily the *fixed* trash bin, is it?

api/src/opentrons/protocol_api/_trash_bin.py
```diff
@@ -10,3 +10,7 @@ class TrashBin:
     def __init__(self, location: DeckSlotName, addressable_area_name: str) -> None:
         self._location = location
         self._addressable_area_name = addressable_area_name
+
+    @property
+    def location(self) -> DeckSlotName:
+        return self._location
```
SyntaxColoring: Since this is a public API<sup>1</sup>, we should either:

* Officially make it public. Gate it behind a new `apiLevel`, with `@requires_version` or similar, and add a docstring.
* Keep it private. Either:
    * Spell this `_location` (with an underscore) and just deal with other Opentrons code having to access a "private" method.
    * Use a docstring, probably with `:meta private:`, to document in English that this is for Opentrons internal use only.

<sup>1</sup>It doesn't look like this `TrashBin` class is actually on docs.opentrons.com at the time of writing, but I suspect that's a bug. We should act as if it is published there.

api/tests/opentrons/motion_planning/test_deck_conflict.py
```diff
@@ -207,6 +207,41 @@ def test_flex_labware_when_thermocycler(
...
+        name_for_errors="some_thermocycler",
+        highest_z_including_labware=123,
+        is_semi_configuration=False,
+    )
+    trash = deck_conflict.TrashBin(name_for_errors="some_trash_bin")
+
+    with pytest.raises(
+        deck_conflict.DeckConflictError,
+        match=(
+            "some_trash_bin in slot A1" " prevents some_thermocycler from using slot B1"
```
SyntaxColoring:
```suggestion
            "some_trash_bin in slot A1 prevents some_thermocycler from using slot B1"
```

api/tests/opentrons/motion_planning/test_deck_conflict.py
```diff
@@ -207,6 +207,41 @@ def test_flex_labware_when_thermocycler(
...
+        deck_conflict.check(
+            existing_items={DeckSlotName.SLOT_A1: trash},
+            new_item=thermocycler,
+            new_location=DeckSlotName.SLOT_B1,
+            robot_type="OT-3 Standard",
+        )
+    with pytest.raises(
+        deck_conflict.DeckConflictError,
+        match=(
+            "some_thermocycler in slot B1" " prevents some_trash_bin from using slot A1"
```
SyntaxColoring:
```suggestion
            "some_thermocycler in slot B1 prevents some_trash_bin from using slot A1"
```

SyntaxColoring (commented):
Mostly looks good, just a pedantic thing about the public `TrashBin` class.

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -135,11 +135,24 @@ def append_disposal_location(
...
+            deck_conflict.check(
+                engine_state=self._engine_client.state,
+                new_trash_bin=disposal_location,
+                existing_disposal_locations=self._disposal_locations,
+                # It's important that we don't fetch these IDs from Protocol Engine, and
+                # use our own bookkeeping instead. If we fetched these IDs from Protocol
+                # Engine, it would have leaked state from Labware Position Check in the
+                # same HTTP run.
+                #
+                # Wrapping .keys() in list() is just to make Decoy verification easier.
```
sanni-t: Oh! We should actually remove/update this comment from this & the other `check` calls too. This is talking about labware & module ID duplication from setup-command-based LPC, which is no longer in use. In that way, we should be able to safely fetch the IDs from the engine directly, but I think quite a few entities in the engine-based API are now relying on these in-api labware and module storage so we can't just replace them easily without checking all usages. But it's also not in the scope for this ticket. So maybe just an updated comment explaining 'that this is outdated and we should fetch the IDs from engine now' should suffice.

sanni-t (approved):
LGTM! I quite like how simple this approach turned out to be. But I agree it's going to be  tough to use for any other kinds of conflict checks where we need to look at the actual dimensions of the bin, so updating this approach in a follow-up sounds good to me.

SyntaxColoring (approved):
TY!

PR #14327, opened 2024-01-17
chore(update-server): bump pyzmq to match the buildroot dependency tree requirements

ryanthecoder:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (changes requested):
We need to run `pipenv update --dev` on the notify server and on the things that depend on it (really just robot-server) and check in the resulting `Pipfile.lock` changes

sfoster1 (changes requested):
Need to relock notify-server too

PR #14328, opened 2024-01-17
refactor(api): Remove opentrons_hardware from ot3api

sfoster1:
While OT3API is only used to control hardware on a flex (where the
opentrons_hardware package is always available), it can also be used to
simulate both in the context of the robot or desktop app (similar) and
on installs from pypi, where opentrons_hardware is _not_ available
because it is not intended to work on any platform other than linux.

We can fix this by making sure that opentrons_hardware doesn't need to
be imported in any protocol simulation path, by making sure that any
imports from opentrons_hardware happen from (eventually) ot3controller,
and ot3controller is conditionally imported when you try and build a
hardware-controlling OT3API.

That is generally pretty easy but it requires some refactors. For
instance,
- Planning moves needs to happen inside ot3controller
- The backend needs to be the owner of the status bar controller and the
estop controller and offer passthrough methods to the API

These shouldn't change any functionality.

Finally, to make sure this doesn't happen again, add a test to try and
import and build a simulating OT3API in the no-hardware case.

Fixes RSS-402

### DIFF
--- a/api/mypy.ini
+++ b/api/mypy.ini
@@ -4,7 +4,7 @@ show_error_codes = True
 warn_unused_configs = True
 strict = True
 # TODO(mc, 2021-09-12): work through and remove these exclusions
-exclude = tests/opentrons/(hardware_control/test_.*py|hardware_control/integration/|hardware_control/emulation/|hardware_control/modules/|protocols/advanced_control/|protocols/api_support/|protocols/duration/|protocols/execution/|protocols/fixtures/|protocols/geometry/)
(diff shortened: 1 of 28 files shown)


andySigler (approved):
lgtm

sfoster1:
Tested it by running some calibrations and such on a flex, including with a 96 to check the tip motors

vegano1 (approved):
this is awesome, thank you.

ahiuchingau (approved):
I ~~like~~LOVE it, makes senses!

PR #14329, opened 2024-01-17
feat(robot-server): Support downgrades without clearing robot-server data

SyntaxColoring:
# Overview

This aims to solve this problem:

1. You upgrade your robot.
2. You play with it for a bit.
3. You decide you don't like the new version, and downgrade back to the original one.
4. But now you have old robot software operating on data that was stored, in step (2), by new robot software. The old software can't deal with the new data, and all kinds of things break. In the worst case, using the robot again will require resetting the stored runs and protocols.

Closes RSS-162.

# Architecture

Our solution is for `robot-server` to store an independent copy of all of its data for each version. Downgrading will then safely revert to the old data.

So, whereas now we have:

* `/var/lib/opentrons-robot-server/`
    * `robot_server.db`
    * `protocols/`

We're moving towards:

* `/var/lib/opentrons-robot-server/`
    * `robot_server.db`
    * `protocols/`
    * `7.2.0`
        * `robot_server.db`
        * `protocols/`
    * `7.3.0`
        * `robot_server.db`
        * `protocols/`

I'm using robot software versions `7.2.0`/`7.3.0` in this example, but we could also use schema versions, like `1`/`2`/`3`. This system doesn't care, and we can change our mind later.

This PR implements the infrastructure for managing these subdirectories, but it doesn't wire it up to anything yet. That's coming up in separate PRs.

# Changelog

* Set up some infrastructure to implement the architecture above.
* Set up the robot-server init sequence to use this new infrastructure to do any future migrations.
* *Do not actually implement any migrations yet.* That will happen in other PRs. So there shouldn't be any observable difference in behavior yet.

# Test Plan

This is mostly covered by existing integration and new unit tests.

The tricky part, atomicity in the face of power loss, is not tested automatically here. I don't see an easy way to do that. I plan on manually testing that on an OT-2 as part of RSS-98, when this system is actually put into use.

# Review requests

* Names and docstrings. If you were unfamiliar with this PR and its background, would the code explain itself to you? I feel like I've been staring at all of this for too long.
* The implementation of atomicity is sensitive, and deserves scrutinous human eyeballs, since we don't have a way to test it automatically.

# Risk assessment

At least medium. Anything that deals with data migrations and filesystem atomicity is going to be pretty sensitive. It's deceptively easy to leave holes that cause actual file corruption in practice. ([1](https://opentrons.atlassian.net/browse/RSS-147), [2](https://lwn.net/Articles/752063/), [etc.](https://danluu.com/file-consistency/) [etc.](https://danluu.com/deconstruct-files/))


robot-server/robot_server/persistence/folder_migrations/folder_migrator.py
```diff
@@ -0,0 +1,257 @@
...
+        root: Path,
+        legacy_uncontained_items: List[str],
+        migrations: List[Migration],
+        temp_file_prefix: str,
+    ) -> None:
+        """Sequence migrations and perform them in order."""
+        for legacy_uncontained_item in legacy_uncontained_items:
+            _validate_bare_name(legacy_uncontained_item)
+        if len(migrations) < 1:
+            raise ValueError("At least one migration is required.")
```
sfoster1: maybe you're going to change this in the future, but raising an exception like this is not a good idea because this class is instantiated at import time of `robot_server.persistence.folder_migrations`. There's no opportunity to handle an exception here without doing a dynamic import, and the entire program will crash. 

It also is the expected case for a system that has already been migrated booting, right? I think it should tolerate not having anything to do especially if it's instantiated at import.

robot-server/robot_server/persistence/folder_migrations/folder_migrator.py
```diff
@@ -0,0 +1,257 @@
...
+                migration.migrate(source_dir=previous_output, dest_dir=output_dir)
+                previous_output = output_dir
+
+        _log.info(f"All migrations complete.")
+        return previous_output
+
+    def delete_old_versions(self, num_old_versions_to_keep: int) -> None:
+        """
+        Warning:
+            Only call this *after* `migrate_to_latest()` completes. Otherwise you might
```
sfoster1: enforce this with a threading.event or something IMO

sfoster1 (commented):
Concept and the thought around FS atomicity looks good to me but we should really change that exception thing.

robot-server/robot_server/persistence/folder_migrations/folder_migrator.py
SyntaxColoring: I might need to clarify that this list of migrations is going to be a hard-coded constant of *all* the migrations up to now, not the migrations that need to be done.

So:

* If you're booting a system that's already been migrated, no, we don't expect this to `raise`.
* If you accidentally hard-code `migrations=[]`, yes, it will raise an exception at `import` time. I don't super mind that. It's akin to doing `SOME_MODULE_LEVEL_CONSTANT = 1/0`. It's a "static-ish" error.

But I might just remove this check. I don't remember why I was so worried about enforcing this in the first place.

robot-server/robot_server/persistence/folder_migrations/folder_migrator.py
SyntaxColoring: I've removed pruning of old versions from this PR, to cut scope.

robot-server/robot_server/persistence/folder_migrations/folder_migrator.py
SyntaxColoring: I've removed this check.

robot-server/robot_server/persistence/_folder_migrator.py
```diff
@@ -0,0 +1,262 @@
...
+        Args:
+            root: Where to hold all the version subdirectories.
+                For example, if you want a file tree like this:
+
+                    foo/
+                        v1/
+                        v2/
+                        ...
+
+                Then `foo/` is the root.
```
SyntaxColoring, line 54: Is `root` a decent name for this parameter? I don't like that it conflicts with the actual root directory, `/`.

robot-server/robot_server/persistence/_folder_migrator.py
```diff
@@ -0,0 +1,262 @@
...
+                return index
+        return None
+
+
+class Migration(ABC):
+    """Represents a single migration step (e.g. v1->v2, not v1->v2->v3).
+
+    Subclass this and override `migrate()` to implement the step's actual migration
+    logic.
+    """
```
SyntaxColoring, line 166: Would `Version` be a better name for this class, maybe?

robot-server/robot_server/persistence/_folder_migrator.py
SyntaxColoring: I hope so, because I've been leaning into the "root persistence directory" vs. "active persistence directory" scheme. :)

robot-server/robot_server/persistence/_persistence_directory.py
```diff
@@ -23,34 +25,45 @@
...
         await file.write_text(encoding="utf-8", data=_RESET_MARKER_FILE_CONTENTS)
 
 
-async def prepare(persistence_directory: Optional[Path]) -> Path:
-    """Create the persistence directory, if necessary, and prepare it for use.
+async def prepare_active_subdirectory(prepared_root: Path) -> Path:
+    """Return the active persistence subdirectory after preparing it, if necessary."""
+    migration_orchestrator = MigrationOrchestrator(
+        root=prepared_root,
+        migrations=[],
```
SyntaxColoring, line 47: When we actually have meaty migration code (🔜 in imminent PRs), we'll add them here.

robot-server/robot_server/persistence/_folder_migrator.py
CaseyBatten: I believe Migration is semantically acceptable here for naming purposes.

robot-server/robot_server/persistence/_folder_migrator.py
CaseyBatten: Acceptable if your point of reference is within the scope of the server directories, but definitely could be conflated with true root though for anyone unfamiliar. Probably fine?

robot-server/robot_server/persistence/_fastapi_dependencies.py
```diff
@@ -90,18 +109,29 @@ async def init_sql_engine_and_log() -> SQLEngine:
...
+    # combine it with initialization of the active subdirectory or the SQL engine.
+    # This lets PersistenceResetter remain usable even if initializing the SQL engine
+    # fails, which is important to let users recover from corrupt databases.
+
+    root_directory_init_task = asyncio.create_task(init_root_persistence_directory())
+    _root_persistence_directory_init_task_accessor.set_on(
+        app_state=app_state, value=root_directory_init_task
+    )
+
+    active_subdirectory_init_task = asyncio.create_task(
```
CaseyBatten, line 127: Assuming the raise case where we failed in the creation of root and/or active directories, what becomes of the operating state of the robot after this? Robot server is running in a degenerate state at this point, so presumably based on following code at line 214 on we will return a series of request errors from various endpoints, but what of the overall system? The ODD will presumably be unresponsive, as it is also dependent upon robot servers endpoints to interact with the system.

CaseyBatten (commented):
Overall the structure of this looks like a solid base to build upon after our various discussions and the way it is utilized here. Outstanding questions about exception cases and the state the machine is left in after those are still in the air, after we discuss some of those questions I think we can move forward with this one.

robot-server/robot_server/persistence/_fastapi_dependencies.py
SyntaxColoring: That's up to the app, but in general, it's supposed to show the robot in a degraded state, lock you out of things that won't work, and allow you to access stuff for diagnostics and recovery. The desktop app, for example, has a red banner to that says something like "the server isn't correctly responding to requests," but it will let you factory-reset, download logs, or do a manual update. (Though there have been bugs before where that stuff has been locked out.)

robot-server/robot_server/persistence/_fastapi_dependencies.py
```diff
@@ -149,23 +186,64 @@ async def get_sql_engine(
         ) from exception
 
 
-async def get_persistence_directory(
+async def get_active_persistence_directory(
```
sfoster1, line 189: Kind of a nit, but is there a reason regular application code should care about active vs root persistence directories? it seems like we should keep this the same name and have it resolve to the active one under the hood.

robot-server/robot_server/persistence/_folder_migrator.py
sfoster1: I think this is a fine name for it.

sfoster1 (approved):
This looks good to me, though I wonder if we should just keep the interface to the persistence directory the same

robot-server/robot_server/persistence/_fastapi_dependencies.py
SyntaxColoring: There isn't a reason, no. I was just unsure of naming I think.

SyntaxColoring:
Oops I just merged this into an integration branch but we want it in `edge` now. Correcting in #14383.

PR #14330, opened 2024-01-18
fix(api): Check for possible collision between lifted labware and attached tips

CaseyBatten:
Fixes RSS-425

In general, catch all cases where the z height of the bottom of attached tips would be lower than the top of gripped labware. We have special cased the left mount when single and 8 channel pipettes are attached.

Because of restrictions on the amount of information that analysis has access to, it was necessary to pass some hardcoded position information about the gantry when using the virtual gripper. This is part of a larger issue with the division between analysis and simulation. In truth, we do not have enough information during analysis to properly analyze the X axis status of the tips mid air against the incoming labware the gripper will attempt to lift. Because of this, we cannot validate X axis overlap without either passing in even more hard coded information about distance between the gripper and various pipette configurations, or by modifying analysis to query more in depth information form the hardware controller (and in turn creating a hardware controller delegate which could operate during both client and robot side analysis. 

**For the time being the solution is instituting behavior as follows:**

- If a single or eight channel pipette is installed in the left or right column, we pass validation as collision is unlikely based on in dev testing. 
- Otherwise, we check the Z height of the top of the labware from the deck once gripped and lifted to the gripper home height against the Z height of the bottom of a given tip from the deck
- If the height of the bottom of the tip from the deck is lower than the height of the top of the labware once lifted, we raise a labware movement error and return information about the labware that was lifted and the tip type it would have conflicted with

# Tests
The following protocol should be executed, with commented code enabled and disabled to cause pass and fail cases four the four test cases.
It must be CONFIRMED using Flex hardware that under no PASS case a robot actively encounters a collision mid air.
The four test cases are as follows:

1. 96 channel tip pickup and tiprack lift with 1000ul tips - Expected to FAIL
2. 1 channel left mount tip pickup and tiprack lift with 1000ul tips - Expected to PASS
3. 8 channel right mount tip pickup and tiprack lift with 1000ul tips - Expected to PASS
4. 8 channel right mount tip pickup and 384 wellplate lift with 50ul tips - Expected to PASS

```
requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(protocol_context):
    small_labware = protocol_context.load_labware("appliedbiosystemsmicroamp_384_wellplate_40ul", "D2")

    adapter = protocol_context.load_adapter("opentrons_flex_96_tiprack_adapter", "C3")
    # Test 1-3 tip racks
    tip_rack = adapter.load_labware("opentrons_flex_96_tiprack_1000ul")
    # Test 4 tiprack
    #tip_rack = adapter.load_labware("opentrons_flex_96_tiprack_50ul")
    

    #Test 1 instrument - should fail analysis
    instrument = protocol_context.load_instrument('flex_96channel_1000', mount="left", tip_racks=[tip_rack])

    #Test 2 instrument
    #instrument = protocol_context.load_instrument('flex_1channel_1000', mount="left", tip_racks=[tip_rack])

    #test 3 and 4 instrument
    #instrument = protocol_context.load_instrument('flex_8channel_1000', mount="right", tip_racks=[tip_rack])

    # Always command all instruments to pick up tips first, refresh tips between tests
    instrument.pick_up_tip(tip_rack)
    
    # Test 1-3 move labware:
    # Test 1 expected: Will fail analysis
    # Test 2 expected: Will pass analysis
    # Test 3 expected: Will pass analysis
    protocol_context.move_labware(tip_rack, new_location="C2", use_gripper=True)

    # Test 4 move labware:
    # Test 4 expected: Will pass analysis
    #protocol_context.move_labware(small_labware, new_location="C2", use_gripper=True)
```

# Risks
Low risk for OT-2 because this section of code only executes on OT-3 hardware.

Potential risk to customers who have built Protocols where they already engage in behavior where they lift labware with the gripper in potentially dangerous ways. This change would cause those protocols to now fail analysis and execution. Attempts have been made to mitigate this by restricting checks to only the most at-risk of configurations.

### DIFF
--- a/api/src/opentrons/protocol_engine/execution/labware_movement.py
+++ b/api/src/opentrons/protocol_engine/execution/labware_movement.py
@@ -98,9 +98,8 @@ class LabwareMovementHandler:
                 for pipette_id, tip in attached_tips:
                     if not self._state_store.geometry.validate_gripper_labware_tip_collision(
                         gripper_homed_position_z=166.125,
-                        pipettes_homed_position_z=248.0,
-                        pipette_id=pipette_id,
                         tip=tip,
+                        pipette_id=pipette_id,
                         labware_id=labware_id,
                         current_location=current_location,
                     ):
@@ -133,18 +132,10 @@ class LabwareMovementHandler:
         attached_tips = self._state_store.pipettes.get_all_attached_tips()
         if attached_tips:
             for pipette_id, tip in attached_tips:
-                pipette_mount = self._state_store.pipettes.get_mount(
-                    pipette_id
-                ).to_hw_mount()
-                pipetted_homed_position = await ot3api.gantry_position(
-                    mount=pipette_mount
-                )
-
                 if not self._state_store.geometry.validate_gripper_labware_tip_collision(
                     gripper_homed_position_z=gripper_homed_position.z,
-                    pipettes_homed_position_z=pipetted_homed_position.z,
-                    pipette_id=pipette_id,
                     tip=tip,
+                    pipette_id=pipette_id,
                     labware_id=labware_id,
                     current_location=current_location,
                 ):
--- a/api/src/opentrons/protocol_engine/execution/labware_movement.py
+++ b/api/src/opentrons/protocol_engine/execution/labware_movement.py
@@ -91,8 +91,8 @@ class LabwareMovementHandler:
         use_virtual_gripper = self._state_store.config.use_virtual_gripper
 
         if use_virtual_gripper:
-            # During Analysis we will pass in hard coded estimates for certain positions on accessible during execution
-            # Estimated gripper homed position Z: 253.575 + 93.85 - 94.825 - 86.475 = 166.125 mm
+            # During Analysis we will pass in hard coded estimates for certain positions only accessible during execution
+            # Estimated gripper homed Z position: 166.125 mm
             attached_tips = self._state_store.pipettes.get_all_attached_tips()
             if attached_tips:
                 for pipette_id, tip in attached_tips:
@@ -132,7 +132,12 @@ class LabwareMovementHandler:
         attached_tips = self._state_store.pipettes.get_all_attached_tips()
         if attached_tips:
             for pipette_id, tip in attached_tips:
-                pipetted_homed_position = await ot3api.gantry_position(mount=self._state_store.pipettes.get_mount(pipette_id))
+                pipette_mount = self._state_store.pipettes.get_mount(
+                    pipette_id
+                ).to_hw_mount()
+                pipetted_homed_position = await ot3api.gantry_position(
+                    mount=pipette_mount
+                )
 
                 if not self._state_store.geometry.validate_gripper_labware_tip_collision(
                     gripper_homed_position_z=gripper_homed_position.z,
--- a/api/src/opentrons/protocol_engine/execution/labware_movement.py
+++ b/api/src/opentrons/protocol_engine/execution/labware_movement.py
@@ -34,6 +34,8 @@ from ..types import (
 if TYPE_CHECKING:
     from opentrons.protocol_engine.execution import EquipmentHandler, MovementHandler
 
+_GRIPPER_HOMED_POSITION_Z = 166.125  # Height of the center of the gripper critical point from the deck when homed
+
 
 # TODO (spp, 2022-10-20): name this GripperMovementHandler if it doesn't handle
 #  any non-gripper implementations
@@ -79,7 +81,7 @@ class LabwareMovementHandler:
             )
         )
 
-    async def move_labware_with_gripper(  # noqa: C901
+    async def move_labware_with_gripper(
         self,
         labware_id: str,
         current_location: OnDeckLabwareLocation,
@@ -92,20 +94,11 @@ class LabwareMovementHandler:
 
         if use_virtual_gripper:
             # During Analysis we will pass in hard coded estimates for certain positions only accessible during execution
-            # Estimated gripper homed Z position: 166.125 mm
-            attached_tips = self._state_store.pipettes.get_all_attached_tips()
-            if attached_tips:
-                for pipette_id, tip in attached_tips:
-                    if not self._state_store.geometry.validate_gripper_labware_tip_collision(
-                        gripper_homed_position_z=166.125,
-                        tip=tip,
-                        pipette_id=pipette_id,
-                        labware_id=labware_id,
-                        current_location=current_location,
-                    ):
-                        raise LabwareMovementNotAllowedError(
-                            f"Cannot move labware '{self._state_store.labware.get(labware_id).loadName}' when {int(tip.volume)} ul tips are attached."
-                        )
+            self._state_store.geometry.check_gripper_labware_tip_collision(
+                gripper_homed_position_z=_GRIPPER_HOMED_POSITION_Z,
+                labware_id=labware_id,
+                current_location=current_location,
+            )
             return
 
         ot3api = ensure_ot3_hardware(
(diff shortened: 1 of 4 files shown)


CaseyBatten:
The protocol test listed in this PR has been tested on RSS Flex. Through this testing it was found that even 1000 ul tips to not cause collisions with standard labware when mounted to a single or 8 channel pipette on the right mount. This implies this is only an issue with columns A9 or A10 through A12 on the 96 channel pipette. Regardless, analysis and execution errors are raised at the appropriate times when a Z height conflict is caught on non-special cased configurations.

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -89,8 +89,25 @@ async def move_labware_with_gripper(
...
         use_virtual_gripper = self._state_store.config.use_virtual_gripper
+
         if use_virtual_gripper:
+            # During Analysis we will pass in hard coded estimates for certain positions only accessible during execution
+            # Estimated gripper homed Z position: 166.125 mm
+            attached_tips = self._state_store.pipettes.get_all_attached_tips()
+            if attached_tips:
+                for pipette_id, tip in attached_tips:
+                    if not self._state_store.geometry.validate_gripper_labware_tip_collision(
+                        gripper_homed_position_z=166.125,
```
SyntaxColoring: Let's pull this up to a module-level constant like `_GRIPPER_HOMED_POSITION_Z`. Mostly just so it'll stick out so that when we have a better place for it, we don't forget to move it.

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -977,6 +978,31 @@ def get_total_nominal_gripper_offset_for_move_type(
...
+        if self._pipettes.get_channels(pipette_id) in [1, 8]:
+            return True
+
+        labware_top_z_when_gripped = gripper_homed_position_z + (
+            self.get_labware_highest_z(labware_id=labware_id)
+            - self.get_labware_grip_point(
+                labware_id=labware_id, location=current_location
+            ).z
+        )
+        # TODO(mm, 2024-01-18): Utilizing the nozzle map and labware X coordinates verify if collisions will occur on the X axis (analysis will use hard coded data to measure from the gripper critical point to the pipette mount)
```
SyntaxColoring: Your initials :P

```suggestion
        # TODO(cb, 2024-01-18): Utilizing the nozzle map and labware X coordinates verify if collisions will occur on the X axis (analysis will use hard coded data to measure from the gripper critical point to the pipette mount)
```

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -977,6 +978,31 @@ def get_total_nominal_gripper_offset_for_move_type(
...
+    def validate_gripper_labware_tip_collision(
+        self,
+        gripper_homed_position_z: float,
+        tip: TipGeometry,
+        pipette_id: str,
+        labware_id: str,
+        current_location: OnDeckLabwareLocation,
+    ) -> bool:
+        """Check for potential collision of tips against labware to be lifted."""
+        pipette_home_z = 248.0
```
SyntaxColoring: Same as my other comment—let's pull this up to a module-level constant like `_PIPETTE_HOME_Z`.

Let's also leave a comment describing exactly what it points to. It's the bottom of the nozzle without a tip, right? And we can get away with this because all of the Flex pipettes have the same value for this?

SyntaxColoring (commented):
This looks good to me, thanks. Just some small comments.

@sanni-t @jbleon95 @CaseyBatten Do you think it's worth doing tests for these with the whole Decoy mock thing? How long-lived do we expect this implementation to be? I guess unless we have imminent plans to replace it in v7.2.0, we should add tests.

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -977,6 +978,31 @@ def get_total_nominal_gripper_offset_for_move_type(
...
+        labware_top_z_when_gripped = gripper_homed_position_z + (
+            self.get_labware_highest_z(labware_id=labware_id)
+            - self.get_labware_grip_point(
+                labware_id=labware_id, location=current_location
+            ).z
+        )
+        # TODO(mm, 2024-01-18): Utilizing the nozzle map and labware X coordinates verify if collisions will occur on the X axis (analysis will use hard coded data to measure from the gripper critical point to the pipette mount)
+        if (pipette_home_z - tip.length) < labware_top_z_when_gripped:
+            return False
+        return True
```
sanni-t: Instead of returning True/ False, you could have this method raise `LabwareMovementNotAllowedError` when a collision is expected. That way, you won't have to do an if/else check in `labware_movement`

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -79,7 +79,7 @@ def __init__(
             )
         )
 
-    async def move_labware_with_gripper(
+    async def move_labware_with_gripper(  # noqa: C901
```
sanni-t: Hmm, I would like to avoid making this a complex method if possible; mainly because it indicates that code is not easily readable. Leaving some refactor suggestions to see if that helps.

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -89,8 +89,25 @@ async def move_labware_with_gripper(
     ) -> None:
         """Move a loaded labware from one location to another using gripper."""
         use_virtual_gripper = self._state_store.config.use_virtual_gripper
+
         if use_virtual_gripper:
+            # During Analysis we will pass in hard coded estimates for certain positions only accessible during execution
+            # Estimated gripper homed Z position: 166.125 mm
+            attached_tips = self._state_store.pipettes.get_all_attached_tips()
+            if attached_tips:
+                for pipette_id, tip in attached_tips:
```
sanni-t: The geometry view already has access to the pipette state and can pull this info from it directly. So we can move this entire tip data fetching into `validate_gripper_labware_tip_collision`.

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -89,8 +89,25 @@ async def move_labware_with_gripper(
...
+                    if not self._state_store.geometry.validate_gripper_labware_tip_collision(
+                        gripper_homed_position_z=166.125,
+                        tip=tip,
+                        pipette_id=pipette_id,
+                        labware_id=labware_id,
+                        current_location=current_location,
+                    ):
+                        raise LabwareMovementNotAllowedError(
+                            f"Cannot move labware '{self._state_store.labware.get(labware_id).loadName}' when {int(tip.volume)} ul tips are attached."
```
sanni-t: Continuing my comment on the validation function, if that function raises error instead of returning true/ false, you won't need to raise this error twice in this method. So essentially, this new code block reduces to just a call to `validate_gripper_labware_tip_collision`.

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -111,6 +128,21 @@ async def move_labware_with_gripper(
...
+                if not self._state_store.geometry.validate_gripper_labware_tip_collision(
+                    gripper_homed_position_z=gripper_homed_position.z,
+                    tip=tip,
+                    pipette_id=pipette_id,
+                    labware_id=labware_id,
+                    current_location=current_location,
+                ):
+                    raise LabwareMovementNotAllowedError(
+                        f"Cannot move labware '{self._state_store.labware.get(labware_id).loadName}' when {int(tip.volume)} ul tips are attached."
+                    )
```
sanni-t: See above comments to reduce this entire code block to just one call to `validate_gripper_labware_tip_collision`

sanni-t (commented):
Logic looks good! I have some refactoring suggestions to make this code a bit more readable and remove some of the bloat.

And I agree with @SyntaxColoring we should add at least an integration test and a unit test for the validator.

api/src/opentrons/protocol_engine/state/geometry.py
SyntaxColoring: @CaseyBatten and I talked about this and I suggested returning a bool, because our unit testing style doesn't give us a strong way of testing that exceptions propagate correctly.

api/src/opentrons/protocol_engine/state/geometry.py
CaseyBatten: Thats what I get for copy pasting haha.

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -977,6 +985,36 @@ def get_total_nominal_gripper_offset_for_move_type(
...
+                    - self.get_labware_grip_point(
+                        labware_id=labware_id, location=current_location
+                    ).z
+                )
+                # TODO(cb, 2024-01-18): Utilizing the nozzle map and labware X coordinates verify if collisions will occur on the X axis (analysis will use hard coded data to measure from the gripper critical point to the pipette mount)
+                if (
+                    _PIPETTE_HOMED_POSITION_Z - tip.length
+                ) < labware_top_z_when_gripped:
+                    raise LabwareMovementNotAllowedError(
+                        f"Cannot move labware '{self._labware.get(labware_id).loadName}' when {int(tip.volume)} ul tips are attached."
```
ecormany: Editor brain comments on this error 🤓🧠 
- Do we want `.loadName` or `.name`? The latter falls back to the former if the protocol doesn't specify a custom name for the labware.
- We like "µL" better than "ul" or "uL" in non-code text.

SyntaxColoring (approved):
Changes look good to me. Thanks!

And good call @sanni-t on the refactoring suggestions.

sanni-t:
🙌

PR #14333, opened 2024-01-19
Abr scale

rclarke0:
# Overview

Adding script to take and analyze scale measurements.

# Test Plan

I did not take any testing steps because this script is not used by any of the robots.

# Changelog

# Review requests

Double check that changes to the drivers on the scale do not break anything else.

# Risk assessment
low risk


andySigler (changes requested):
thank you for this! let's sit down together and go through what needs change before a merge 👍

rclarke0 (commented):
LGTM

hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
```diff
@@ -0,0 +1,6 @@
+"""ABR Temperature Humidity Sensors."""
+# from hardware_testing import data
```
andySigler, line 2: You can remove commented things. If it's commented out, then it should just be deleted.

andySigler: also should this file be a part of the PR? It's empty

andySigler (approved):
Looks ready to merge, other than that one file that seems empty other than the commented out pieces

PR #14339, opened 2024-01-23
fix(app): fix Mixpanel empty string issue

koji:
# Overview
fix Mixpanel empty string issue

close RAUT-920 's `robotType` part

add trackEvent to protocol setup since there wasn't trackEvent function. (This was the reason why we couldn't see ODD when use analyze the data with `runStart`.

# Test Plan

# Changelog
- pass storedProtocol's `robotType` instead of '' when protocolAnalysis `robotType` is null/undefined

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/hooks/useProtocolRunAnalyticsData.ts
+++ b/app/src/organisms/Devices/hooks/useProtocolRunAnalyticsData.ts
@@ -46,7 +46,9 @@ export const parseProtocolRunAnalyticsData = (
       protocolAuthor: protocolAuthor !== '' ? protocolAuthor : '',
       protocolText: protocolText !== '' ? protocolText : '',
       robotType:
-        protocolAnalysis?.robotType != null ? protocolAnalysis?.robotType : '',
+        protocolAnalysis?.robotType != null
+          ? protocolAnalysis?.robotType
+          : storedProtocol?.mostRecentAnalysis?.robotType,
     },
     runTime:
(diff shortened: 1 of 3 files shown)


app/src/organisms/Devices/hooks/useProtocolRunAnalyticsData.ts
```diff
@@ -46,7 +46,9 @@ export const parseProtocolRunAnalyticsData = (
       protocolAuthor: protocolAuthor !== '' ? protocolAuthor : '',
       protocolText: protocolText !== '' ? protocolText : '',
       robotType:
-        protocolAnalysis?.robotType != null ? protocolAnalysis?.robotType : '',
+        protocolAnalysis?.robotType != null
+          ? protocolAnalysis?.robotType
+          : storedProtocol?.mostRecentAnalysis?.robotType,
```
mjhuff, line 51: Can we be certain storedProtocol and mostRecentAnalysis are defined here? Are we ok with an undefined robot type?

koji: Other items can be `undefined`. I guess it would be okay.

@shlokamin 
Should we avoid passing `not set`?
Technically, we can use `useRobotType` but for that, we will need to modify a couple of hooks and tests.

mjhuff (approved):
LGTM! I'm merging this in so we can be code complete by EOD.

PR #14341, opened 2024-01-23
fix(api): fix left mount being disengaged when moving to maintenance position

ahiuchingau:
# Overview
We now disengage the left z axis if a 96-channel pipette is attached but idle. This means that during gripper calibration, the left Z axis will not have a valid motor position (disengaged at the home position). We won't be able to move nor access the position of the Z_L until we first home it. This shouldn't take long because it should already be at its home position.


sfoster1 (approved):
Ok, looks good to me

sfoster1 (approved):
Yay much nicer

PR #14343, opened 2024-01-24
feat(app,app-shell,app-shell-odd,robot-server): Add push notification support

mjhuff:
Closes RAUT-941 and RSS-442

# Overview

This PR adds push-based notification support to the app and robot server. See [this confluence doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3928621089/Notifications) for background on the architectural decisions made and [this PR](https://github.com/Opentrons/oe-core/pull/124) for the mosquitto message broker implementation.

Two routes are refactored as proof of concept: `GET /maintenance_runs/current_run/` and `GET /runs/:runId/commands?cursor=null&pageLength=1`

This PR also fixes a tricky React Query anti-pattern that was causing unintentional infinite polling on a few routes. 

### Client Side Architecture (the parts that probably require some explanation)

Because the app-shell communicates with the browser layer via ipcRenderer, it can be a bit hard to conceptualize. Hopefully this diagram helps. Note that the diagram shows data flow from the notification client to the app only.

Screenshot 2024-01-24 at 9 05 27 AM

Communication from the app to the notification client is done via our preestablished redux-observable pattern, in which we dispatch an action to the shell. You can see this within useNotifyService, on mount and dismount.

**Connection Manager**

See the initial comment in notify.ts for a complete explanation. In order to benefit from raw TCP, the connection manager must go on the node layer, since there's no raw TCP on the browser layer. This logic existing on the node layer will come in handy when we support USB, too. A connection manager is necessary regardless of whether we choose to use MQTT or WebSockets. 

**useNotifyService**

The key browser layer utility function. This handles our subscribing/unsubscribing to topics, notification errors, tells us when to refetch data, and sets our query data if the server sends actual data. 

### Server Side Architecture (the parts that probably require some explanation)

Because we are only ever pushing a flag, not much needs to change. 

I tested running publish() synchronously and asynchronously, and there aren't noticeable performance differences either way. That being said, I feel like it's probably more correct to handle the publishing async (definitely open to talk about this!).

Having the server do the polling from PE doesn't feel great, but this is already a significant improvement over having the app do the same thing. Bubbling up the notification from PE is the better long term solution, but I think this is sufficient for now. 

**We certainly need integration testing, and I'm planning to talk to Josh about this!** 

# Test Plan

[Here is a link](https://opentrons.slack.com/archives/C81CR4VCZ/p1706228347828169) to the Flex build.

Verified that:

- [x] Robot-server and app can connect to the broker.
- [x] Robot-server pushes refetch flags when it should for both refactored routes.
- [x] Robot-server correctly connects/disconnects from the broker on startup/termination.
- [x] Robot-server continues to function if it can't establish a connection to the broker.  
- [x] App fallback logic works (easy to test on the OT-2 and USB now, since neither currently support notifications).
- [x] App network tab actually shows that refactored routes aren't polling.
- [x] Trying to unsubscribe from a topic that you aren't subscribed doesn't break the app.
- [x] Trying to disconnect from a broker you aren't connected to doesn't break the app.
- [x] The app correctly behaves when try to connect and subscribe to the same broker/topic multiple times (the subscription counter updates appropriately). 
- [x] The app correctly behaves when initiating multiple connections at once to the same broker (only one connection persists).
- [x] The app correctly unsubscribes and closes connections when they are no longer in-use.
- [x] Components that use the refactored work as expected.

# Changelog
- Added push notifications to the app and robot server.

# Risk assessment
High. If the message broker doesn't work as expected and neither does the fallback logic, that's a problem. That being said, it's held up well in testing (so far...).


app-shell-odd/src/notify.ts
mjhuff, line 1: This works as expected, so I'm fine not touching this for the initial PR. An early refactor we should make is initiating a persistent connection to the localhost robot server on startup and never terminate it. Let's keep things uniform for the moment, though.

app-shell/src/notify.ts
```diff
@@ -0,0 +1,260 @@
...
+  }
+}
+
+const connectionStore: ConnectionStore = {}
+const log = createLogger('notify')
+// MQTT is somewhat particular about the clientId format and will connect erratically if an unexpected string is supplied.
+// This clientId is derived from the mqttjs library.
+const CLIENT_ID = 'app-' + Math.random().toString(16).slice(2, 8)
+
+const connectOptions: mqtt.IClientOptions = {
```
mjhuff, line 29: These are pretty much default settings. If we ever push real data, we should revisit this, however.

app-shell/src/notify.ts
```diff
@@ -0,0 +1,260 @@
...
+            }
+          })
+        })
+      })
+      .catch((error: Error) => {
+        log.warn(
+          `Failed to connect to ${hostname} - ${error.name}: ${error.message} `
+        )
+        const failureMessage = error.message.includes('ECONNREFUSED')
+          ? 'ECONNREFUSED'
```
mjhuff, line 114: I separate out these error strings for Mixpanel reporting reasons. ECONNREFUSED is the best proxy we have for letting us know if customers firewall us. Definitely worth gathering analytics on this.

app-shell/src/robot-update/index.ts
```diff
@@ -180,6 +179,13 @@ export function getRobotSystemUpdateUrls(
     .then(manifest => {
       const urls = getReleaseSet(manifest, CURRENT_VERSION)
 
+      // if (urls === null) {
```
mjhuff, line 182: Yes, intentional.

app/src/redux/shell/remote.ts
```diff
@@ -36,3 +37,51 @@ export function appShellRequestor<Data>(
...
+      shellMessage,
+    } = deserializeNotifyMessage(message)
+
+    if (hostname === shellHostname && topic === shellTopic) {
+      eventEmitter.emit('data', shellMessage)
+    }
+  })
+  return eventEmitter
+
+  function deserializeNotifyMessage(
```
mjhuff: MQTT serializes to binary before sending messages. We convert these to strings from buffers on the app-shell layer. I'm hesitant to deserialize to an object until we make it through ipcRenderer, so that's why it's done here.

app/src/redux/shell/remote.ts
```diff
@@ -36,3 +37,51 @@ export function appShellRequestor<Data>(
...
+    message: string
+  ): {
+    shellHostname: string
+    shellTopic: NotifyTopic
+    shellMessage: string | Object
+  } {
+    const delimiter = ':'
+    const firstIndex = message.indexOf(delimiter)
+    const secondIndex = message.indexOf(delimiter, firstIndex + 1)
```
mjhuff: Not so fun fact: .split() can't be used to achieve this without potentially chopping off parts of the message object.

discovery-client/src/health-poller.ts
```diff
@@ -72,7 +72,7 @@ export function createHealthPoller(options: HealthPollerOptions): HealthPoller {
           onPollResult(result)
           lastCompletedPollTimeByIp[ip] = pollTime
         } else {
-          log('debug', 'Stale poll result ignored', { ip, port, result })
+          // log('debug', 'Stale poll result ignored', { ip, port, result })
```
mjhuff, line 75: Also intentional.

react-api-client/src/maintenance_runs/useCurrentMaintenanceRun.ts
mjhuff, line 1: This pattern avoids the infinite, unintentional polling mentioned in the PR summary. Polling will now only occur on the refetchInterval, if specified.

robot-server/robot_server/runs/run_data_manager.py
```diff
@@ -73,11 +75,17 @@ class RunDataManager:
...
+        engine_store: EngineStore,
+        run_store: RunStore,
+        task_runner: TaskRunner,
+        notification_client: NotificationClient,
     ) -> None:
         self._engine_store = engine_store
         self._run_store = run_store
         self._task_runner = task_runner
+        self._notification_client = notification_client
+        self._stop_polling_event = asyncio.Event()
```
mjhuff, line 88: Totally open to hear if anyone has any better ideas on how to do the polling here. This does work as it should, at least.

app-shell/src/main.ts
```diff
@@ -44,7 +45,14 @@ if (config.devtools) app.once('ready', installDevtools)
 
 app.once('window-all-closed', () => {
   log.debug('all windows closed, quitting the app')
-  app.quit()
+  closeAllNotifyConnections()
```
sfoster1, line 48: do we have a possible hang case here? should we wrap this in a `useTimeout` that just quits if it exceeds and leave the os to clean the sockets?

app/src/redux/shell/remote.ts
sfoster1: all we put in these flag messages is going to be POJOs anyway, right? IMO go straight from buffer to object in the shell.

app/src/resources/useNotifyService.ts
mjhuff, line 1: It seems difficult to mock emitters in Jest, and I can't really test ipcRenderer without making a mock electron app. If anyone has any ideas on how to unit test this, that would be great.

app/src/redux/shell/remote.ts
```diff
@@ -36,3 +37,51 @@ export function appShellRequestor<Data>(
 
   return remote.ipcRenderer.invoke('usb:request', configProxy)
 }
+
+export function appShellListener(
+  hostname: string | null,
+  topic: NotifyTopic
+): EventEmitter {
+  const eventEmitter = new EventEmitter()
```
sfoster1, line 45: I'm a little concerned about the inspectability of this stuff. It won't show up in the devtools networking window and it kind of won't show up in logging and it won't show up in the redux message tracker. Can we maybe pipe some of this through redux, or log it out or something?

app-shell/src/notify.ts
```diff
@@ -0,0 +1,260 @@
...
+function establishListeners({
+  client,
+  browserWindow,
+  hostname,
+  topic,
+}: ListenerParams): void {
+  client.on('message', (topic, message, packet) => {
+    browserWindow.webContents.send(
+      'notify',
+      `${hostname}:${topic}:${message.toString()}`
```
sfoster1: `message` is a Buffer containing the json that the robot server sends, right? and then we `toString()` it here so it's a json string?

app/src/redux/shell/remote.ts
sfoster1: Yeah I have to admit packing the hostname, topic, and message together into a string, using a delimiter that's guaranteed to be in the message, is striking me as kind of bug-prone. mqtt is giving us those elements separately anyway, can we send an object with named fields or even a 3-length array and let the electron implementation worry about serialization? then we can avoid this entirely.

app-shell/src/notify.ts
mjhuff: > `message` is a Buffer containing the json that the robot server sends, right? and then we `toString()` it here so it's a json string?

Correct. My reasoning for not doing the full deserialization on the node layer had to do with sending actual data, but I agree that it probably makes more sense to cross that bridge when we get there. 

I'll move all the serialization down to the node layer.

app/src/redux/shell/remote.ts
mjhuff: Sounds good. I'll do this on the browser layer.

app-shell/src/main.ts
mjhuff: The first optional arg to end is forceClose, but it's hard to determine based on the docs if this is truly sufficient. I've added in a 2 second timeout just to be safe.

mjhuff:
Just realized the Flex build I linked in the Test Plan was broken. I've updated the link to a new build.

robot-server/robot_server/notification_client.py
```diff
@@ -0,0 +1,160 @@
...
+)
+
+log: logging.Logger = logging.getLogger(__name__)
+
+
+class NotificationClient:  # noqa: D101
+    def __init__(
+        self,
+        host: str = "127.0.0.1",
+        port: int = 1883,
```
jbleon95, line 37: for cleanliness we should move these to be private constants, they can live in this folder for now

robot-server/robot_server/notification_client.py
```diff
@@ -0,0 +1,160 @@
...
+        keepalive: int = 60,
+        protocol_version: int = mqtt.MQTTv5,
+        default_qos: int = 1,
```
jbleon95: I know that qos in mqtt is defined as integers, but again for readability we should either use an MQTT imported constant or just make our own

robot-server/robot_server/notification_client.py
```diff
@@ -0,0 +1,160 @@
...
+        retain_message: bool = False,
+    ) -> None:
+        """Returns a configured MQTT client."""
+        self.host = host
```
jbleon95: make these attributes private unless there is a reason we need them to be public

jbleon95 (approved):
The Robot Server side of the code looks good for now, besides the minor Python style comments for notification client and the asyncio code in run_data_manager, which per our discussion is being actively refactored and we can iterate more on it in that future PR.

PR #14346, opened 2024-01-24
refactor(robot-server): Add a placeholder migration for schema 3

SyntaxColoring:
# Overview

This is a step towards the database improvement work that we have planned for this release (RSS-98 and RSS-132). It also happens to close RSS-190.

# Test plan

We're well-covered by automated tests, so far.

We'll do further manual testing after follow-up PRs merge, so we can test the changes together.

# Changelog

PR #14329 added a new versioning and migration system for robot-server's persisted data, but it didn't put it into use yet.

This PR adds the first migration under the new system. It's currently a no-op migration that just copies files.

Before:

```
/data/opentrons_robot_server/
    protocols/
    deck_configuration.json
    robot_server.db
```

After:

```
/data/opentrons_robot_server/
    protocols/
    deck_configuration.json
    robot_server.db
    3/
        protocols/
        deck_configuration.json
        robot_server.db
```

Follow-up PRs will edit this migration in-place to add their actual migration work. We will batch them up in the integration branch and release them as one migration.

# Review requests

* Am I forgetting about any files?
* Are the names understandable? Is the organization understandable?

# Risk assessment

Low.

### DIFF
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -41,20 +41,6 @@ def _migrate_db(
     source_transaction: sqlalchemy.engine.Connection,
     dest_transaction: sqlalchemy.engine.Connection,
 ) -> None:
-    # TODO: This migration table is what we used to use to keep track of what schema
-    # the database is at. It's not needed anymore now that we have this
-    # subdirectory-based migration infrastructure.
-    #
-    # We should either drop this table entirely from schema 3 and up, or, if we
-    # want to keep it, we should add a row here to stamp it with schema "3".
-    copy_rows_unmodified(
-        schema_2.migration_table,
-        schema_3.migration_table,
-        source_transaction,
-        dest_transaction,
-        order_by_rowid=False,
-    )
-
     copy_rows_unmodified(
         schema_2.protocol_table,
         schema_3.protocol_table,
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -1,0 +1,88 @@
+from contextlib import ExitStack
+from pathlib import Path
+
+import sqlalchemy
+
+from .._database import create_schema_2_sql_engine, create_schema_3_sql_engine
+from .._folder_migrator import Migration
+from .._tables import schema_2, schema_3
+from ._util import copy_rows_unmodified, copy_if_exists, copytree_if_exists
+
+
+# TODO: Define a single source of truth somewhere for these paths.
+_DECK_CONFIGURATION_FILE = "deck_configuration.json"
+_PROTOCOLS_DIRECTORY = "protocols"
+_DB_FILE = "robot_server.db"
+
+
+class MigrationUpTo3(Migration):  # noqa: D101
+    def migrate(self, source_dir: Path, dest_dir: Path) -> None:
+        """Migrate the persistence directory from schema 2 to 3."""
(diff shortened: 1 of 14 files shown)


robot-server/robot_server/persistence/_migrations/up_to_3.py
```diff
@@ -0,0 +1,88 @@
...
+        with ExitStack() as exit_stack:
+            # If the source is schema 0 or 1, this will migrate it to 2 in-place.
+            source_db = create_schema_2_sql_engine(source_dir / _DB_FILE)
+            exit_stack.callback(source_db.dispose)
+
+            dest_db = create_schema_3_sql_engine(dest_dir / _DB_FILE)
+            exit_stack.callback(dest_db.dispose)
+
+            with source_db.begin() as source_transaction, dest_db.begin() as dest_transaction:
+                _migrate_db(source_transaction, dest_transaction)
```
SyntaxColoring, line 37: A note about how I'm copying the .db file:

I'm creating a fresh .db file, setting up its schema from scratch, and then copying records one-by-one from the old .db file. As opposed to doing a filesystem-level copy of the old file and then editing the new one in-place.

The advantages to this are that it solves RSS-190, it avoids [gotchas](https://www.sqlite.org/howtocorrupt.html#roguejrnl) with copying .db files at the filesystem level, and it makes it easy to change column types and do other alterations that would otherwise be [very difficult](https://www.sqlite.org/lang_altertable.html#making_other_kinds_of_table_schema_changes).

The disadvantage is that it incurs compute overhead from extracting and reinserting the records.

I think that overhead is acceptable for this migration because the solutions to RSS-98 and RSS-132 are going to incur it anyway, but we might need to come up with something faster for future migrations.

robot-server/robot_server/persistence/_migrations/up_to_3.py
```diff
@@ -0,0 +1,88 @@
...
+def _migrate_db(
+    source_transaction: sqlalchemy.engine.Connection,
+    dest_transaction: sqlalchemy.engine.Connection,
+) -> None:
+    # TODO: This migration table is what we used to use to keep track of what schema
+    # the database is at. It's not needed anymore now that we have this
+    # subdirectory-based migration infrastructure.
+    #
+    # We should either drop this table entirely from schema 3 and up, or, if we
+    # want to keep it, we should add a row here to stamp it with schema "3".
```
SyntaxColoring: I'm leaning towards just dropping the table. Anyone else have an opinion?

SyntaxColoring: I've decided to drop it.

robot-server/robot_server/persistence/_migrations/up_to_3.py
CaseyBatten: Agree that this is acceptable for the first implementation of these changes. Question, will this mean that for every time we create an upgraded version we will be going through the same record by record db file migration from the last valid version, hence the overhead being inherited by each subsequent version? If so, this becomes a problem we can worry about solving for the next version of the database, and for now we can focus on a stable overall layout.

CaseyBatten (approved):
Agree low risk and migration methodology is appropriate for current goal set.

robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring: > Question, will this mean that for every time we create an upgraded version we will be going through the same record by record db file migration from the last valid version, hence the overhead being inherited by each subsequent version?

Basically yes.

PR #14347, opened 2024-01-24
fix(shared-data): widen gripper error bounds

sfoster1:
Unfortunately, the gripper assemblies vary by enough that without calibration (which we haven't yet implemented) we are not accurate (i.e., encoder position matching physical position) enough to trigger on small errors. A larger value should still catch the cases where the gripper closes on air.

The next step is to implement jaw calibration during home to bring this value down again (RET-1416)

Closes RQA-2242


SyntaxColoring (approved):
I do indeed allow this grip error.

sanni-t (approved):
👍

PR #14348, opened 2024-01-24
perf(robot-server): Store one command per row

SyntaxColoring:
# Overview

Closes RSS-132. See that ticket for background.

# Test Plan

I think we're sufficiently covered by existing automated integration tests.

[I've reviewed some of the new queries with SQLite's `EXPLAIN QUERY PLAN`](https://gist.github.com/SyntaxColoring/0207c40f82a6ad5b16fb269e37927a0f). They look like they're all using fast index-based searches instead of slow full table scans.

# Changelog

* Remove the `run.commands` column, where each value was a large list of commands.
* In its place, add a `run_commands` table, where each row holds just a single command.

  | **run_id** | **index_in_run** | **command_id** | **command** |
  |------------|------------------|----------------|-------------|
  | run1       | 0                | abcd           | [blob]      |
  | run1       | 1                | efgh           | [blob]      |
  | run2       | 0                | ijkl           | [blob]      |
  | run2       | 1                | mnop           | [blob]      |
  | ...        | ...              | ...            | ...         |

* Update `RunStore` to use the new table.
* Add a migration.

# Review requests

This new table will be our biggest, by far. Tens of thousands of records, as opposed to ~20 in our existing tables. One of SQL's traps is that it's easy to accidentally write a very inefficient query. If the right indexes aren't set up, SQLite will degrade to a full O(n) table scan. So scrutinize my queries to make sure we're not doing that.

Also see my inline review comments.

# Risk assessment

Medium. See review requests above.

### DIFF
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -4,11 +4,7 @@ from typing import Any, Dict, List
 
 import sqlalchemy
 
-from .._database import (
-    create_schema_2_sql_engine,
-    create_schema_3_sql_engine,
-    sqlite_rowid,
-)
+from .._database import create_schema_2_sql_engine, create_schema_3_sql_engine
 from .._folder_migrator import Migration
 from .._tables import schema_2, schema_3
 from ._util import copy_rows_unmodified, copy_if_exists, copytree_if_exists
@@ -80,44 +76,34 @@ def _migrate_run_table(
     source_transaction: sqlalchemy.engine.Connection,
     dest_transaction: sqlalchemy.engine.Connection,
 ) -> None:
-    select_everything_except_commands = sqlalchemy.select(
-        schema_2.run_table.c.id,
-        schema_2.run_table.c.created_at,
-        schema_2.run_table.c.protocol_id,
-        schema_2.run_table.c.state_summary,
-        schema_2.run_table.c.engine_status,
-        schema_2.run_table.c._updated_at,
-    ).order_by(sqlite_rowid)
+    select_old_run = sqlalchemy.select(schema_2.run_table)
     insert_new_run = sqlalchemy.insert(schema_3.run_table)
-
-    for old_run_row in (
-        source_transaction.execute(select_everything_except_commands).mappings().all()
-    ):
-        # Insert one at a time to retain sqlite rowid ordering.
-        # Providing many rows at once to execute() may not preserve order.
-        # https://www.mail-archive.com/db-sig@python.org/msg02071.html
-        dest_transaction.execute(insert_new_run, old_run_row)
-
-    select_commands = sqlalchemy.select(
-        schema_2.run_table.c.id, schema_2.run_table.c.commands
-    )
     insert_new_command = sqlalchemy.insert(schema_3.run_command_table)
 
-    for old_command_row in source_transaction.execute(select_commands).all():
-        run_id = old_command_row.id
-        commands: List[Dict[str, Any]] = old_command_row.commands or []
+    for old_run_row in source_transaction.execute(select_old_run).all():
+        dest_transaction.execute(
+            insert_new_run,
+            id=old_run_row.id,
+            created_at=old_run_row.created_at,
+            protocol_id=old_run_row.protocol_id,
+            state_summary=old_run_row.state_summary,
+            engine_status=old_run_row.engine_status,
+            _updated_at=old_run_row._updated_at,
+        )
+
+        commands: List[Dict[str, Any]] = old_run_row.commands or []
         new_command_rows = [
             {
-                "run_id": run_id,
+                "run_id": old_run_row.id,
                 "index_in_run": index_in_run,
                 "command_id": command["id"],
                 "command": command,
             }
             for index_in_run, command in enumerate(commands)
         ]
-        # Insert all the commands in one go, to avoid the overhead of separate
-        # statements, and since we had to bring them all into memory at once in order to
-        # parse them anyway.
+        # Insert all the commands for this run in one go, to avoid the overhead of
+        # separate statements, and since we had to bring them all into memory at once
+        # in order to parse them anyway.
         if len(new_command_rows) > 0:
             # This needs to be guarded by a len>0 check because if the list is empty,
             # SQLAlchemy misinterprets this as inserting a single row with all default
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -1,9 +1,14 @@
 from contextlib import ExitStack
 from pathlib import Path
+from typing import Any, Dict, List
 
 import sqlalchemy
 
-from .._database import create_schema_2_sql_engine, create_schema_3_sql_engine
+from .._database import (
+    create_schema_2_sql_engine,
+    create_schema_3_sql_engine,
+    sqlite_rowid,
+)
 from .._folder_migrator import Migration
 from .._tables import schema_2, schema_3
 from ._util import copy_rows_unmodified, copy_if_exists, copytree_if_exists
@@ -57,12 +62,9 @@ def _migrate_db(
         order_by_rowid=True,
     )
 
-    copy_rows_unmodified(
-        schema_2.run_table,
-        schema_3.run_table,
+    _migrate_run_table(
         source_transaction,
         dest_transaction,
-        order_by_rowid=True,
     )
 
     copy_rows_unmodified(
@@ -72,3 +74,52 @@ def _migrate_db(
         dest_transaction,
         order_by_rowid=True,
     )
+
+
+def _migrate_run_table(
+    source_transaction: sqlalchemy.engine.Connection,
+    dest_transaction: sqlalchemy.engine.Connection,
+) -> None:
+    select_everything_except_commands = sqlalchemy.select(
+        schema_2.run_table.c.id,
+        schema_2.run_table.c.created_at,
+        schema_2.run_table.c.protocol_id,
+        schema_2.run_table.c.state_summary,
+        schema_2.run_table.c.engine_status,
+        schema_2.run_table.c._updated_at,
+    ).order_by(sqlite_rowid)
+    insert_new_run = sqlalchemy.insert(schema_3.run_table)
+
+    for old_run_row in (
+        source_transaction.execute(select_everything_except_commands).mappings().all()
+    ):
+        # Insert one at a time to retain sqlite rowid ordering.
+        # Providing many rows at once to execute() may not preserve order.
+        # https://www.mail-archive.com/db-sig@python.org/msg02071.html
+        dest_transaction.execute(insert_new_run, old_run_row)
+
+    select_commands = sqlalchemy.select(
+        schema_2.run_table.c.id, schema_2.run_table.c.commands
+    )
+    insert_new_command = sqlalchemy.insert(schema_3.run_command_table)
+
+    for old_command_row in source_transaction.execute(select_commands).all():
+        run_id = old_command_row.id
+        commands: List[Dict[str, Any]] = old_command_row.commands or []
+        new_command_rows = [
+            {
+                "run_id": run_id,
+                "index_in_run": index_in_run,
+                "command_id": command["id"],
+                "command": command,
+            }
+            for index_in_run, command in enumerate(commands)
+        ]
+        # Insert all the commands in one go, to avoid the overhead of separate
+        # statements, and since we had to bring them all into memory at once in order to
+        # parse them anyway.
+        if len(new_command_rows) > 0:
+            # This needs to be guarded by a len>0 check because if the list is empty,
+            # SQLAlchemy misinterprets this as inserting a single row with all default
+            # values.
+            dest_transaction.execute(insert_new_command, new_command_rows)
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -4,7 +4,11 @@ from typing import Any, Dict, List
 
 import sqlalchemy
 
-from .._database import create_schema_2_sql_engine, create_schema_3_sql_engine
+from .._database import (
+    create_schema_2_sql_engine,
+    create_schema_3_sql_engine,
+    sqlite_rowid,
+)
 from .._folder_migrator import Migration
 from .._tables import schema_2, schema_3
 from ._util import copy_rows_unmodified, copy_if_exists, copytree_if_exists
@@ -76,11 +80,11 @@ def _migrate_run_table(
     source_transaction: sqlalchemy.engine.Connection,
     dest_transaction: sqlalchemy.engine.Connection,
 ) -> None:
-    select_old_run = sqlalchemy.select(schema_2.run_table)
+    select_old_runs = sqlalchemy.select(schema_2.run_table).order_by(sqlite_rowid)
     insert_new_run = sqlalchemy.insert(schema_3.run_table)
     insert_new_command = sqlalchemy.insert(schema_3.run_command_table)
 
-    for old_run_row in source_transaction.execute(select_old_run).all():
+    for old_run_row in source_transaction.execute(select_old_runs).all():
         dest_transaction.execute(
             insert_new_run,
             id=old_run_row.id,
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -1,3 +1,13 @@
+"""Migrate the persistence directory from schema 2 to 3.
+
+Summary of changes from schema 2:
+
+- Run commands were formerly stored as monolithic blobs in the run table,
+  with each row storing an entire list. This has been split out into a new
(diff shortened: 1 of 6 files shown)


robot-server/robot_server/persistence/_tables/schema_3.py
```diff
@@ -127,3 +121,26 @@
...
+    ),
+    sqlalchemy.Column("index_in_run", sqlalchemy.Integer, nullable=False),
+    sqlalchemy.Column("command_id", sqlalchemy.String, nullable=False),
+    sqlalchemy.Column(
+        "command",
+        # TODO: Should be JSON.
+        sqlalchemy.PickleType(pickler=legacy_pickle, protocol=PICKLE_PROTOCOL_VERSION),
+        nullable=False,
+    ),
+    sqlalchemy.PrimaryKeyConstraint("run_id", "command_id"),
```
SyntaxColoring, line 139: As a quirk of Protocol Engine and our HTTP API, commands are addressable by a string UUID *or* by an integer index. The two are basically redundant with each other. We need to support efficient access through both.

As far as I can tell, it's arbitrary which one we choose to be the primary key? 🤷

robot-server/robot_server/persistence/_tables/schema_3.py
```diff
@@ -127,3 +121,26 @@
...
+    sqlalchemy.Index(
+        "ix_run_run_id_index_in_run",  # An arbitrary name for the index.
+        "run_id",
+        "index_in_run",
+        unique=True,
+    ),
```
SyntaxColoring: As a quirk of Protocol Engine and our HTTP API, commands are addressable by a string UUID *or* by an integer index. We need to support efficient access through both.

Here, I'm picking the `(run_id, command_id)` to be the primary key, and setting up a unique index for `(run_id, index_in_run)`. We could also do it the other way around. As far as I can tell, it's totally arbitrary? 🤷 

[Edit: Clarified composite keys]

robot-server/robot_server/persistence/_migrations/up_to_3.py
```diff
@@ -86,3 +84,42 @@ def _migrate_db(
...
+            for index_in_run, command in enumerate(commands)
+        ]
+        # Insert all the commands for this run in one go, to avoid the overhead of
+        # separate statements, and since we had to bring them all into memory at once
+        # in order to parse them anyway.
+        if len(new_command_rows) > 0:
+            # This needs to be guarded by a len>0 check because if the list is empty,
+            # SQLAlchemy misinterprets this as inserting a single row with all default
+            # values.
+            dest_transaction.execute(insert_new_command, new_command_rows)
```
SyntaxColoring, line 125: The fact that I need to add this `len()` check makes me feel like I'm doing this wrong.

robot-server/robot_server/runs/run_store.py
```diff
@@ -368,22 +389,33 @@ def remove(self, run_id: str) -> None:
...
+        ).scalar_one()
+        return result
+
     def _clear_caches(self) -> None:
         self.has.cache_clear()
         self.get.cache_clear()
         self.get_all.cache_clear()
         self.get_state_summary.cache_clear()
         self.get_command.cache_clear()
-        self._get_all_unparsed_commands.cache_clear()
```
SyntaxColoring, line 386: This cache was mostly saving the computational work of repeatedly parsing the whole list of commands. It's removed here "accidentally," because we never parse the whole list of commands anymore: we only parse the slices that are requested, when they're requested.

If we wanted to retain the cache, I guess the spiritual sequel would be a pair of caches whose keys are `(run_id, command_index)` and `(run_id, command_id)`. It wouldn't be as simple as an `@lru_cache` anymore. We could do this, I guess, but I'm skeptical that it's worthwhile. Our app [no longer requests big batches of run commands](https://opentrons.slack.com/archives/CSCLVUW3C/p1706206819746719). And if it did, I'd urge us to find ways to speed up the actual underlying processing, like what we did in #13425.

SyntaxColoring: I suppose for consistency, we should remove the cache on `get_command(id)` too.

robot-server/robot_server/persistence/_migrations/up_to_3.py
sfoster1: IMO if you're concerned about something like this, write out the sql you want. Don't trust sqlalchemy. if you're running `INSERT INTO commands (run_id, index_in_run, command_id, command) VALUES ();` and the table has defaults, then it will add a new row with default values, that's pretty much all there is to it.

robot-server/robot_server/persistence/_tables/schema_3.py
```diff
@@ -127,3 +121,26 @@
...
+    "run_command",
+    metadata,
+    sqlalchemy.Column(
+        "run_id", sqlalchemy.String, sqlalchemy.ForeignKey("run.id"), nullable=False
+    ),
+    sqlalchemy.Column("index_in_run", sqlalchemy.Integer, nullable=False),
+    sqlalchemy.Column("command_id", sqlalchemy.String, nullable=False),
+    sqlalchemy.Column(
+        "command",
+        # TODO: Should be JSON.
```
sfoster1: what will let us do the todo?

robot-server/robot_server/persistence/_tables/schema_3.py
sfoster1: If I'm understanding the way this works it is most definitely not arbitrary because there is one table for all the commands across all the runs, and therefore index_in_run will not be unique across the table - you'd have to use a composite key or something.

The alternative, which I think I support, actually is using an autogenerated primary key with the command id and the index in run both as just random columns. The reason to do this is that the responsibility of primary key is not the same as the responsibility of good unique way to disambiguate columns - it's only a partial overlap. If I'm understanding it right, and I may not be, it's usually a very good idea to have your primary key be autogenerated and not something you care about at all, so that if you change something else - in this case, the command id i suppose - you don't have to rekey the table.

Also primary keys really should be unique, so either you're missing a uniqueness constraint on the command id - which is presumably fine since it's a uuid - or we absolutely have to do the autogen key thing.

sfoster1: Although now that I take a second look, where exactly are you setting the command id to be the primary key? It isn't set in line 32

robot-server/robot_server/runs/run_store.py
```diff
@@ -78,28 +83,42 @@ def update_run_state(
...
                 _convert_state_to_sql_values(
                     run_id=run_id,
-                    commands=commands,
                     state_summary=summary,
                     engine_status=summary.status,
                 )
             )
         )
+
+        delete_existing_commands = sqlalchemy.delete(run_command_table).where(
```
sfoster1, line 92: is this to prevent the table from growing, or concern about command id key collisions? if the latter, why not do a pre-delete and upsert, like
- have a "deleted" column
- `UPDATE commands SET deleted=true WHERE run_id == run_id; INSERT INTO commands  VALUES (command_id, index_in_run, command, run_id, false) ON CONFLICT (command_id,) UPDATE (index_in_run, run_id, command, deleted) ; DELETE commands WHERE deleted=true` (or even leave off that last clause - marking items invalid instead of deleting them is a reasonably common practice, though we're more space constrained than most dbs)

it would accomplish a similar thing but have some kind of nicer semantics. Also I'm getting this from eg https://www.sqlite.org/lang_upsert.html and I might have syntax slightly wrong and I don't know how sqlalchemy does it.

robot-server/robot_server/runs/run_store.py
```diff
@@ -305,22 +308,41 @@ def get_commands_slice(
...
+            actual_cursor = cursor if cursor is not None else count_result - length
+            # Clamp to [0, count_result).
+            actual_cursor = max(0, min(actual_cursor, count_result - 1))
+
+            select_slice = (
+                sqlalchemy.select(
+                    run_command_table.c.index_in_run, run_command_table.c.command
+                )
+                .where(
+                    run_command_table.c.run_id == run_id,
```
sfoster1, line 329: https://www.sqlitetutorial.net/sqlite-between/ ?

sfoster1 (commented):
comments on db stuff

robot-server/robot_server/persistence/_tables/schema_3.py
SyntaxColoring: > you'd have to use a composite key or something.

Yes, sorry, I was unclear.

This is currently setting the primary key to be a composite of `(run_id, command_id)` (see next point), and the alternative that I had in mind was setting it to be a composite of `(run_id, index_in_run)`.

> Although now that I take a second look, where exactly are you setting the command id to be the primary key? It isn't set in line 32

Line 139. `sqlalchemy.PrimaryKeyConstraint("run_id", "command_id")`

> Also primary keys really should be unique, so either you're missing a uniqueness constraint on the command id - which is presumably fine since it's a uuid - or we absolutely have to do the autogen key thing.

My understanding is that the `PrimaryKeyConstraint` implies uniqueness (because SQL primary keys are always unique), but I will double-check this.

So in other words, my understanding is that I've effectively told SQLite to have two composite indexes: `(run_id, command_id)`, and `(run_id, index_in_run)`. Both of those have a unique constraint. Either one them can serve as the primary key, and the PR right now chooses `(run_id, command_id)` for that.

>  If I'm understanding it right, and I may not be, it's usually a very good idea to have your primary key be autogenerated and not something you care about at all, so that if you change something else - in this case, the command id i suppose - you don't have to rekey the table.

Mm. I've seen mixed advice on this. I'll take a closer look. My knee-jerk reaction is that I would rather not introduce a third way of addressing commands, on top of the command ID and the run index. I see the point about decoupling, but it doesn't seem like a valuable thing to decouple in our case.

robot-server/robot_server/runs/run_store.py
SyntaxColoring: `BETWEEN` does an inclusive check, so it would have had to be `index_in_run BETWEEN actual_cursor AND actual_cursor + length - 1`, which felt less clear to me than `index_in_run >= actual_cursor AND index_in_run < actual_cursor + length`. I can change it if you think it's clearer, though.

robot-server/robot_server/runs/run_store.py
SyntaxColoring: > is this to prevent the table from growing, or concern about command id key collisions?

Sort of both? `RunStore.update()`'s contract is that it replace's a run's list of commands, so this does that. (We could change `RunStore`'s API to provide a separate `append_commands()` method or something, instead, if that makes the SQL simpler, but I don't want to do that in this PR.)

> it would accomplish a similar thing but have some kind of nicer semantics.

Could you clarify this? What's the advantage of a `deleted` column instead of just deleting those rows?

sfoster1: It doesn't really matter _that_ much in this context; it's common in web systems where you might want to offer undo, or get analytics from past transactions, or have scheduled tasks that do the delete or something. In this case the deleted flag mostly exists to support the upsert so you can get rid of any command ids that no longer exist. I didn't realize the semantics were that strict, if they are a full delete then insert is probably better.

robot-server/robot_server/runs/run_store.py
sfoster1: I think it's clearer because we're trying to go "full sql" here. I'd rather do the check with the sql mechanism.

robot-server/robot_server/persistence/_tables/schema_3.py
sfoster1: Is it true that it's common for command ids to be reused between runs but unique within the run? I thought they were just UUIDs?

This is honestly sounding more and more like we should make the primary key autogenerated and forget about this forever. That's the value there.

SyntaxColoring: They are implemented as just UUIDs, yeah. But I view the API contract as them being unique within the protocol, not unique globally. For example, the HTTP API endpoint is `GET /runs/{run_id}/commands/{command_id}`, where the command is a subresource of the run, as opposed to `GET /runs/commands/{command_id}` or something.

Is the primary key being compound your concern, and you're trying to avoid that? Or is the point really to decouple the DB primary key from any application-level semantics, regardless of compound-ness? Like, do you feel the same way about run IDs and action IDs and protocol analysis IDs?

sfoster1: I want the database layer to be the database layer, organized as a good useful database layer. I think that the presentation of the API is something that should lead you to design database _access patterns_ to make it easy to do; it should not drive the storage model of the database. APIs can change, databases can change. Linking them explicitly means that whenever you want to change one, you either have to change the other or suddenly do the work of separating them. And fundamentally, if we're using a uuid for the command id there's no reason to use anything else for the primary key _for the purposes of the database_.

Work with the database to help it do the thing it's good at - storing and indexing data. Work with the API to do the thing it's good at - representing data in a clear and domain-relevant way. Don't force one to compromise for the other; instead, use sql to bridge the gap. Do the work of separating them now so we don't have to do it in the future.

robot-server/robot_server/runs/run_store.py
SyntaxColoring: Oh, I'm not sure if this is what you mean, but the `>=` and `<` are "full SQL" in the sense that they won't get evaluated in Python. SQLAlchemy does operator overloading magic and uses them to generate a real SQL expression, like this:

```
SELECT run_command.index_in_run, run_command.command
FROM run_command
WHERE run_command.run_id = ? AND run_command.index_in_run >= ? AND run_command.index_in_run < ? ORDER BY run_command.index_in_run
```

(This is pulled from sqlalchemy's logs, so the `?`s are placeholders where the actual parameters are supplied.)

Does that change your opinion, or did you mean something else by "full SQL"?

robot-server/robot_server/runs/run_store.py
SyntaxColoring: OK, thanks, that makes sense. We'll keep the full delete and insert for now.

In that kind of append-only SQL world, I wonder how we would represent the progression of a command through its `queued`->`running`->`completed` states. One entry for each state? That seems excessive to me, but maybe?

robot-server/robot_server/persistence/_tables/schema_3.py
SyntaxColoring: Updated with a ticket reference, TY.

robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring: > write out the sql you want. Don't trust sqlalchemy.

Do you mean manually concatenating a SQL string roughly like `f"INSERT INTO commands (blahblah) VALUES ({blahblahblah})"`, or am I misinterpreting?

robot-server/robot_server/persistence/_tables/schema_3.py
SyntaxColoring: Replaced the `(run_id, command_id)` primary key with a `row_id INTEGER` primary key and a `(run_id, command_id)` unique index.

robot-server/robot_server/runs/run_store.py
CaseyBatten: Utilizing the `BETWEEN` implementation case is slightly better for readability but in terms of logical operation there is nothing immediately wrong with the current approach of validating the position of `index_in_run` relative to the cursor in the code or the given SQL expression.

robot-server/robot_server/persistence/_migrations/up_to_3.py
CaseyBatten: I believe catching here is appropriate if we are going to continue to use SQLAlchemy in this way, because I'm not entirely sure it would only insert a new row of defaults if passed an empty list. Check out the following:
https://github.com/sqlalchemy/sqlalchemy/issues/9647#issuecomment-1743670202
If we ever were to pass a `executemany` with an `INSERT` while handling an empty list we would likely hit the raise case mentioned in that thread. Have we confirmed that currently without this check in place, when passed an empty command list, that the software doesn't encounter an exception? If so, it also may come down to a difference of versions at this appears to possibly be a 2.1 change, which would mean we are vulnerable to upgrades in the future without catching it ourselves here (assuming we stick with the SQLAlchemy approach, which I'm fine with).

CaseyBatten (approved):
Looks good overall, some comments below with questions and clarifications.

robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring: > Check out the following:
https://github.com/sqlalchemy/sqlalchemy/issues/9647#issuecomment-1743670202

Oh, excellent find, thanks.

> Have we confirmed that currently without this check in place, when passed an empty command list, that the software doesn't encounter an exception?

We'd hit a (correct) SQLite error because we would be (incorrectly) trying to insert a row with no values.

> this appears to possibly be a 2.1 change

Right, so, if I'm reading that thread correctly, our code here isn't doing anything especially weird. This is just a SQLAlchemy API trap that they're eventually hoping to fix in ~2.1. Given that, I'm happy to keep the `if len(...)` check.

robot-server/robot_server/runs/run_store.py
SyntaxColoring: Sticking with `>=`/`<` for now.

PR #14349, opened 2024-01-24
chore: resolve (or suppress) all Sphinx warnings

ecormany:
# Overview

Every Python docs build was spewing hundreds of Sphinx warnings. This prevented us from treating warnings as errors in CI. It also made it really hard to tell if you introduced some new problem, like a broken link.

This PR brings the current docs to zero _reported_ warnings. Warnings with easily fixable root causes are actually fixed. Warnings resulting from type hints in the API Reference have been suppressed. Is this a kludge? Kinda. But the only other way out — as far as I can tell — is to _actually publicly document every single thing in those type hints_. I believe we have multiple reasons not to do that. Hopefully each warning suppression pattern that I've added to `conf.py` matches all and only a category of things we have motivated reasons to not document.

Addresses RTC-198 (except for hardware docs).

# Test Plan

- Check out the sandboxes. Plural, because the CI changes could affect hardware, v1, and v2 docs.
- Check out the API docs build action for this branch and see how clean the v2 build is!

# Changelog

### Fix warnings
- Fix stuff only Sphinx cares about, like linking the same text twice on a page (yes really).
- Don't link to classes we've decided to not put in the reference, like `TrashBin` and `WasteChute`.
- Don't use fake class names, like `ModuleContext` to refer to `ThermocyclerContext`, `HeaterShakerContext`, etc.
- Give up on trying to link to `pop()` (what are the Python docs doing??)

### Suppress warnings
Adds a long list of `nitpick_ignore_regex` options.

### CI
- Switches the `v%` builds away from allowing warnings. (`v1` already had no warnings, hooray.)
- Deletes the unused lightbulb image and stops copying it around AWS and everywhere every time we build.

# Review requests

Is the warning suppression strategy valid? Are there things that should be cleaned up in source instead? Should we publicize and document some of these instead?

# Risk assessment

Low. Changes CI slightly but should only fail for actual Sphinx problems that we should catch as soon as they happen.


api/docs/v2/deck_slots.rst
```diff
@@ -150,7 +150,7 @@ Starting in API version 2.16, you must load trash bin fixtures in your protocol
 .. versionadded:: 2.16
 
 .. note::
-    The :py:class:`TrashBin` class doesn't have any callable methods, so you don't have to save the result of ``load_trash_bin()`` to a variable, especially if your protocol only loads a single trash container. Being able to reference the trash bin by name is useful when dealing with multiple trash containers.
+    The ``TrashBin`` class doesn't have any callable methods, so you don't have to save the result of ``load_trash_bin()`` to a variable, especially if your protocol only loads a single trash container. Being able to reference the trash bin by name is useful when dealing with multiple trash containers.
```
SyntaxColoring: Let's publicize `opentrons.protocol_api.TrashBin` and `opentrons.protocol_api.WasteChute`, I think. Doesn't have to happen in this PR though.

api/docs/v2/conf.py
```diff
@@ -424,3 +424,26 @@
...
+
+# -- Suppress autodoc warnings --------------------------------------------
+
+# Ignore warnings for deliberately missing/undocumented things that appear
+# in automatically generated type signatures.
+#
+# The goal here is to pass through any warnings for bad targets of MANUALLY
+# created links.
+nitpick_ignore_regex = [
+    ("py:class", r".*\._.*"),  # anything private
```
SyntaxColoring: We generally do want it to be an error if we accidentally mention something private in the docs. It looks like this is only triggering on `TrashBin` and `WasteChute` now, which makes sense. Once we publicize those, I hope we can delete this rule.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -393,17 +393,17 @@ def parent(self) -> Union[str, Labware, ModuleTypes, OffDeckType]:
             If the labware is directly on the robot's deck, the ``str`` name of the deck slot,
             like ``"D1"`` (Flex) or ``"1"`` (OT-2). See :ref:`deck-slots`.
 
-            If the labware is on a module, a :py:class:`ModuleContext`.
+            If the labware is on a module, a module context.
```
SyntaxColoring, line 396: We don't have to list them all out, but it might be worth linking some example of a module context here and in the other places where we're replacing broken `ModuleContext` links. Again, doesn't have to happen in this PR.

SyntaxColoring (approved):
Yay, thank you!

RIP lightbulb.jpg.

> Is the warning suppression strategy valid? Are there things that should be cleaned up in source instead? Should we publicize and document some of these instead?

Yeah, this makes sense.

I haven't looked very closely at the list of stuff that we need to ignore because I'm not very API-brained at the moment (sorry). I think it's likely that a lot of them are pointing to things that we should resolve "properly," like you said. But it's definitely a good idea to ignore them for now so we can treat new warnings as errors and stop things from getting worse.

sfoster1 (approved):
Looks good to me, thank you for fixing it all!

api/docs/v2/deck_slots.rst
ecormany: I think I'll do it here, since otherwise I'll have to refer back to this PR to figure out which occurrences I want to re-link.

api/docs/v2/conf.py
ecormany: Good point. I thought this was catching something else too, but it must be addressed by one of the less broad filters. 

I did have to add `types.DeckSlotName` in its place, because just including that in the reference was a can of worms, which is now RTC-386.

api/src/opentrons/protocol_api/labware.py
ecormany: A couple other options here:
- Link to the [top of the Modules section](http://docs.opentrons.com/v2/new_protocol_api.html#modules) of the reference (which starts with Heater-Shaker because it's first alphabetically). We could even insert some text into `new_protocol_api.rst` to explain that everything in that section is a "module context".
- Link to the [Module Setup](https://docs.opentrons.com/v2/modules/setup.html) section, which talks about module contexts, although somewhat obliquely.

PR #14350, opened 2024-01-24
fix(shared-data): load pick up current properly

sfoster1:
While we properly handled _most_ kinds of pipette config, there's one we didn't: pick up current. It gets special treatment now because we can properly set different currents for different numbers of tips, rather than being a value. So, we need to mark it as a setting that needs to be iterated over a dictionary in its leaf value rather than directly set.

This also fixes an issue that caused the OT-2 to not boot when loading a pipette configuration that had a pick up current customization, since this code is called when loading a pipette.

It may be closing the barn after the horse escapes, but to prevent similar issues preventing boot, I also tossed in a catch all except that logs a loud message but won't prevent starting the server.

## Testing
- [x] With this PR, modify a pipette's pick up current and check that the system reboots properly and the values are preserved across reboots
- [x] Downgrade an OT-2 to 6.3.0, modify a pipette's pick-up current, and then update to something with this PR
   - [x] note that this isn't really necessary because 6.3.0, 7.0.1, 7.1.0, and edge all produce exactly the same config override file

Closes RESC-187


vegano1 (approved):

ahiuchingau (approved):
Makes sense!

PR #11561, opened 2022-10-11
refactor(robot-server): Store Pydantic objects as JSON instead of pickles

SyntaxColoring:
# Overview

This changes the way robot-server stores certain data, to make things less fragile. Closes RSS-96 and RSS-98.

# Changelog

* When storing protocol analyses, run state summaries, and run commands, do not store them as pickled dicts anymore. Instead, store them as serialized JSON strings. This closes RSS-98.
* During `robot-server` startup, migrate existing pickle-based databases to the new JSON-based schema.
* Use a new custom SQLAlchemy column type to hold these serialized objects. This makes it obvious from the table declaration which Python types we're serializing in there. It also guards against type mismatch bugs that would otherwise be easy to miss in the SQLAlchemy soup. This closes RSS-96.

# Test plan

I've done this on an OT-2 with the Opentrons App:

First, poke around your robot's existing run and protocol records with Postman or the Opentrons App, to get an idea of what they're supposed to look like.

Then, push this branch.

In the Opentrons App, you should see the robot as unavailable for a few minutes, even after the light stops blinking. (This is RSS-217.) When it becomes available again, poke around your run and protocol records with Postman or the Opentrons App. They should be exactly as you left them. it should behave exactly as it did before the push. You should also see messages mentioning migrations in the robot logs.

# Code review requests

* Is the `PydanticCol`, `sql_to_pydantic()`, and `pydantic_to_sql()` stuff understandable?
* Please scrutinize the new code in `_migrations.py`, since that's what the tests are covering the least.

# Risk assessment

High.

* There is no downgrade path.

  If you downgrade your robot after performing this migration, many HTTP requests will fail with internal 5XX errors.
  
  Fundamentally, this problem is not new with this PR. We have the same problem any time we add a new Protocol Engine command, for example. But this PR makes it worse because it affects entire tables.
  
  Another fun hazard: if you downgrade, manage to perform some database operations successfully, and then re-upgrade, your database will have a mixture of v1 and v2 records. Not good.
  
  You can resolve all of this by factory-resetting the robot's run history through the Opentrons App. But this will also erase your labware offsets.
  
  RSS-162 tracks this general problem.

* Pydantic may have gotchas for things like field aliases or custom `__root__` types. So, there's a risk that some of our weirder types will get serialized incorrectly in the new schema. I want `PydanticCol` to have unit tests for this, but it doesn't yet because of time constraints.

[...]

PR #14355, opened 2024-01-26
refactor(robot-server): Store Pydantic objects as JSON instead of pickles, take 2

SyntaxColoring:
# Overview

Closes RSS-98. Takes over for #11561.

# Test Plan

I'm going to spend the day putting this on various robots and smoke-testing them to make sure it doesn't break any basic flows or cause any horrific performance regressions.

Other than that, we have okay automated test coverage. But see the risk assessment below.

# Changelog

* For `RunStore`, convert our storage of `Command`s and `StateSummary`s from pickled dictified Pydantic objects to JSON-serialized Pydantic objects.
* For `CompletedAnalysisStore`, we already did that conversion in #13425, in an experimental parallel column. Promote that column to be non-experimental, and delete the old one that had pickled dicts.

# Review requests

None in particular.

# Risk assessment

Medium.

## Correctness risks

Automated tests currently cover these:

* Freshly-created records can survive a round-trip in and out of the database.
* The migration doesn't outright break any endpoints—i.e., HTTP requests won't start returning HTTP 500 errors.

But they do *not* cover:

* HTTP requests will return exactly the same JSON data that they were returning before. In particular, dealing with JSON `null`/omitted fields is always a bit hazardous with Pydantic, and this migration introduces points where something can go wrong with that.

This omission is tracked as RSS-213.

Mitigating this risk is the fact that we've been running this scheme in production, experimentally, since #13425, and it apparently hasn't caused any problems there.

## Performance risks

[Prior testing](https://opentrons.atlassian.net/l/cp/iHJnoFgC) has shown parsing JSON to take about 1.7x as long as parsing our existing pickled dicts. So retrieving runs and commands may get slower. (Retrieving analyses is unaffected because of #13425).

Manual testing will have to bear out whether this is a problem in practice. I expect #14286 and #14348 to go a long way. If we need a nuclear option, we can do something like #13425 again.

The migration can take a long time. I think this is unavoidable to some extent and we just need to communicate it. I've done what I can here to mitigate this by combining this migration with #14348. RSS-217 makes this worse, but we can fix that separately.

## Risk tolerance

We should take seriously the risks listed above, and mitigate them as much as we have time for, but we also need to weigh them against the risk of *not* merging this.

* RSS-98 is really not a good thing to have around, and it gets harder to solve the longer we let it sit.
* We were previously mitigating RSS-98 by giving it ongoing manual attention (e.g. RSS-157, RSS-198). We haven't done that in a while (RSS-371) because we've been preoccupied with the Flex and with deck configuration.
* Because we don't have Datadog configured for the Flex yet, RSS-98 can be setting up Flex-specific traps that we don't know about.

### DIFF
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -67,12 +67,9 @@ def _migrate_db(
         order_by_rowid=True,
     )
 
-    copy_rows_unmodified(
-        schema_2.analysis_table,
-        schema_3.analysis_table,
+    _migrate_analysis_table(
         source_transaction,
         dest_transaction,
-        order_by_rowid=True,
     )
 
     _migrate_run_table(
@@ -141,3 +138,29 @@ def _migrate_run_table(
             # SQLAlchemy misinterprets this as inserting a single row with all default
             # values.
             dest_transaction.execute(insert_new_command, new_command_rows)
+
+
+def _migrate_analysis_table(
+    source_connection: sqlalchemy.engine.Connection,
+    dest_connection: sqlalchemy.engine.Connection,
+) -> None:
+    select_old_analyses = sqlalchemy.select(schema_2.analysis_table).order_by(
+        sqlite_rowid
+    )
+    insert_new_analysis = sqlalchemy.insert(schema_3.analysis_table)
+    for row in (
+        # The table is missing an explicit sequence number column, so we need
+        # sqlite_rowid to retain order across this copy.
+        source_connection.execute(select_old_analyses).all()
+    ):
+        dest_connection.execute(
+            insert_new_analysis,
+            # The new `completed_analysis` column has the data that used to be in
+            # `completed_analysis_as_document`. The separate
+            # `completed_analysis_as_document` column is dropped.
+            completed_analysis=row.completed_analysis_as_document,
+            # The remaining columns are unchanged:
+            id=row.id,
+            protocol_id=row.protocol_id,
+            analyzer_version=row.analyzer_version,
+        )
--- a/robot-server/robot_server/persistence/_migrations/up_to_3.py
+++ b/robot-server/robot_server/persistence/_migrations/up_to_3.py
@@ -2,9 +2,18 @@
 
 Summary of changes from schema 2:
 
-- Run commands were formerly stored as monolithic blobs in the run table,
+- Run commands were formerly stored as monolithic blobs in the `run` table,
   with each row storing an entire list. This has been split out into a new
-  run_command table, where each individual command gets its own row.
+  `run_command` table, where each individual command gets its own row.
+
+- All columns that were storing binary pickles have been converted to storing
+  JSON strings:
+  - `analysis.completed_analysis`
+  - `run.state_summary`
+  - `run_commands.command` (formerly `run.commands`; see above)
+
+- `analysis.completed_analysis_as_document` has been removed,
+  since the updated `analysis.completed_analysis` (see above) replaces it.
 """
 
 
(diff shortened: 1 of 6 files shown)


robot-server/robot_server/persistence/_migrations/up_to_3.py
```diff
@@ -109,25 +118,40 @@ def _migrate_run_table(
...
 
     for old_run_row in source_transaction.execute(select_old_runs).all():
+        old_state_summary = old_run_row.state_summary
+        new_state_summary = (
+            None
+            if old_run_row.state_summary is None
+            else pydantic_to_json(
+                pydantic.parse_obj_as(StateSummary, old_state_summary)
+            )
+        )
```
SyntaxColoring, line 114: We could add a `try: ... except: pass` around each run's migration to drop it if something goes wrong, instead of making the whole server unavailable.

robot-server/robot_server/protocols/completed_analysis_store.py
SyntaxColoring, line 1: Some background on this file:

This supports endpoints like `GET /protocols/{id}/analyses/{id}`.

Originally, we stored these records as pickled dicts, and fully validated them before returning them over HTTP.

In #13425, we introduced a parallel DB column *and* parallel endpoint: they store the records as JSON, and return them *without validating them.* This of course has downsides, but it seems to be working well.

This PR consolidates the parallel DB columns, but it leaves the parallel endpoints alone. So we're always reading JSON now. The difference is just whether we validate it before returning it to the client, which depends on which endpoint the client uses.

We can probably consolidate the parallel endpoints now, but that's beyond the scope of this PR.

SyntaxColoring:
Test results so far:

* The atomicity implemented by prior PRs seems to work, so that's cool. I haven't had problems with rebooting or power-cycling in the middle of the migration.
* I ran a few protocols with 1000 comment commands to simulate there being a lot of data to migrate. The migration took ~15 seconds. Extrapolating to a worst case based on 20 10000-command runs (stress test cases from [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3563028488/Performance+stress+test+cases)), that's like 25 minutes. For comparison, our ABR protocols seem to have ~50-200 commands.
    * I have a proof of concept branch that uses multiprocessing to migrate the runs in parallel, and it seems like it can cut that in ~half, so we might do that in a follow-up PR.

robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring: Yeah I've decided I should do this. Dropping runs silently is bad, but it's significantly better than failing the boot.

jbleon95 (approved):
I think this looks good for moving this whole database migration forward. We should definitely keep an eye on the potential performance issues you noted, but at this point we should have this out on our internal robots and respond from there.

CaseyBatten (approved):
A few thoughts on the risks section:

- To what extent has the new migration time estimate been communicated to other parties within opentrons? Is there a plan in place to implement user/public facing prompts regarding the migration process to prepare users for this?
- If the solutions within #14286 and #14348 do not provide the desired enhancement in performance we should be prepared to implement an as yet designed third option so as not to further slow communications on the system. I'm not exactly a fan of the nuclear option.

Outside of that, the changes here look appropriate and we can feel safe to move forward into more advanced testing based on the initially reported test results.

robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring: I haven't gotten to this yet. We should do it in a follow-up.

SyntaxColoring:
> To what extent has the new migration time estimate been communicated to other parties within opentrons? Is there a plan in place to implement user/public facing prompts regarding the migration process to prepare users for this?

I'll make a post internally, and for the public, we'll have to add release notes.

PR #14357, opened 2024-01-26
fix(ci): build-ts before push to npm

y3rsh:
### Overview

There is no typing present in 0.1.1 of 
- [shared-data](https://www.npmjs.com/package/@opentrons/shared-data)
- [components](https://www.npmjs.com/package/@opentrons/components)

### Test Plan

If this is the correct fix, merge it then tag - or - tag this branch? to test a new deploy.

### Changelog

- add `make build-ts` in the GitHub Actions job that publishes for `shared-data`
- add `make build-ts` in the GitHub Actions job that publishes for `components`

### Review requests

- [x] Is this the correct way to get the types distributed in the package?
- [x] in https://github.com/Opentrons/opentrons/blob/4371e799ac2d43e2eba511eacd71bb1fda0b0604/components/webpack.config.js#L6 we use the `barrel.ts` but https://github.com/Opentrons/opentrons/blob/4371e799ac2d43e2eba511eacd71bb1fda0b0604/components/package.json#L6 will look for types on index.  Will that be ok?

### Risk assessment

None to any functional code.  CI only change to build types before distributing components and shared-data.

### DIFF
--- a/.github/workflows/components-test-build-deploy.yaml
+++ b/.github/workflows/components-test-build-deploy.yaml
@@ -191,6 +191,8 @@ jobs:
           yarn config set cache-folder ./.yarn-cache
           yarn config set network-timeout 60000
           yarn
+      - name: 'build typescript'
+        run: make build-ts
       - name: 'build library'
         run: |
           make -C components lib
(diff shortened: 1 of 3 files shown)


y3rsh:
### Progress

On build `0.1.2-alpha.0` components types are working.
shared-data they are not.

BUT If I copy in the the contents of `shared-data/lib/js` from my local build of the monorepo, into node_modules, which is where the package.json says types will be, then shared-data works fine.

- [x] Need to figure out why in ci `shared-data/lib/js` is not present.

b-cooper (approved):
✅

PR #14361, opened 2024-01-26
feat(api): gripper error detection

sfoster1:
# Description
Overhauls gripper error detection to take into account various sources of error.

There are three sources of error in determining whether the gripper jaws are at a location that means it probably did (or didn't) successfully grip something:

## Where are the gripper jaws?

Grippers are manufactured with a critical dimension hard stop - when the jaws are closed. When the jaws are opened, the jaws may be in different places on different grippers. We say that the jaw is "homed" - and zero its encoder - when the jaw is opened, and then calculate position by subtracting the current encoder value from the "known" open position. That means that you might close a gripper all the way to the dimensionally accurate stop and see that it thinks it is at 58mm or 62mm instead of 60mm, because that unit's open position was actually physically 90mm or 94mm instead of 92mm.

The way to fix it is to run a quick calibration - open the jaws to the stops set the encoder zero; close the jaws to the stops and capture the position; save the difference between the captured position and the nominal jaw displacement as an offset. It's really fast, and since it's really fast and dependent on the gripper we can just do it whenever we home the gripper for the first time. Then we'll keep it in memory and use it for positioning the jaws, and implicitly fold it in to position calculations including critical point calculations for things on the jaws like the calibration probe.

Note that neither this change to add gripper jaw calibration nor the previous incorrect lack of gripper jaw calibration affects gripper _offset_ calibration, because the two jaws are equally and oppositely affected by the manufacturing variance, and their position is averaged.

fa1e3511e1e92505f22e3dadf0cd219686ec2fed is this change.

## How wide is the thing where the gripper gripped it?

Our labware definitions capture a bounding box for the labware as a whole - a rectangular prism that is as small as it can be while still encompassing the entire labware. We use that bounding box to estimate where gripper jaws should be when gripping a labware, and raise errors if the actual jaw position differs too much. However, the labware bounding box is not on its own a good estimate of the gripper jaw position for two reasons: taper, and lips. First, many labwares have a little lip on the bottom; the bounding box must include it, and it's never relevant to the gripper interaction. Second, many labwares taper slightly from bottom to top, and the position of the gripper jaws depends on the height at which the gripper grips the labware.

Both of these elements have nothing to do with the wells and therefore aren't included in the gripper definition. Plus, they're a pain to measure by hand, which is what often happens with custom labware. Rather than making more accurate outer dimension data required, we can establish bounds of accuracy - a size that the labware definitely will not exceed (the labware bounding box) and a size that definitely will not exceed the labware. A good candidate for that second one is a bounding box established by the _wells_. The wells are inside the labware, so if you create a box that spans the outer edges of all the wells then you get a reasonable lower bound for the labware side.

Once we have those bounds, we can improve the gripper error calculation to take them into account. A grip is only an error if the jaw position (minus remaining uncertainty, see below) is smaller than the well bounding box or larger than the labware bounding box.

This is in ca25e3971e7569458bd268ef38d70d150003520e

## How much did the gripper jaws flex?

The gripper jaws are light, and the mechanism that drives them is light, and they apply a fair amount of force very far from where they're attached to the gripper body. That means that when the gripper is applying force through its jaws, the pads on the jaws will be farther out than the position of the motor above because the jaws will flex or rotate under the force. That difference in position can only be estimated and characterized, not sensed, and is the source of the grip error sense tolerance that remains in shared-data.

This is (implicitly) in fa1e3511e1e92505f22e3dadf0cd219686ec2fed 

## Also hard limits

Finally, though, there are some reliable hard limits. If the gripper is all the way open or all the way closed, it almost certainly is not gripping something with the required amount of force. If it's all the way open, then some motion error must have occurred or the jaws are just stuck. Even if there's a very large labware that barely fit inside the jaws, the flexing of the jaws means that the sensed position will be inboard of the stops. If the gripper is all the way closed, even if the jaws are technically around something and flexing, then in everything but the most corner of corner cases the hard stops have hit before the jaws have exerted the desired amounts of force. That means that we can confidently say that if we're at the limits, we failed. 

This is in d357035a1bef873086c39060b594c41c5d2e7649

# Testing
- Find grippers that are near the manufacturing tolerance bounds for open jaw position
- Successfully move some labware
- Get an error when taking the labware away

Closes RET-1416


api/src/opentrons/hardware_control/instruments/ot3/gripper.py
```diff
@@ -169,6 +177,28 @@ def gripper_id(self) -> str:
     def reload_configurations(self) -> None:
         return None
 
+    def update_jaw_open_position_from_closed_position(
```
ahiuchingau, line 182: Should we sanity-check the `jaw_at_closed` value in case we're trying to save the value when the gripper jaw is stuck?

ahiuchingau (changes requested):
This is great! I love the implementation of the well bounding box.

I think we should add a check for the `jaw_at_closed` before updating it though

api/src/opentrons/hardware_control/instruments/ot3/gripper.py
sfoster1: that is probably a good idea

sfoster1: https://github.com/Opentrons/opentrons/pull/14361/commits/7b09895902964d70fb57add749c83964924f1e45

ahiuchingau (approved):
LGTM! 😲🙌 at the 3 mm

api/src/opentrons/protocol_engine/execution/labware_movement.py
```diff
@@ -156,10 +156,18 @@ async def move_labware_with_gripper(
...
-                            labware_width=self._state_store.labware.get_dimensions(
-                                labware_id
-                            ).y
+                            expected_grip_width=labware_bbox.y,
+                            grip_width_uncertainty_wider=abs(
+                                max(well_bbox.y - labware_bbox.y, 0)
+                            ),
+                            grip_width_uncertainty_narrower=abs(
+                                min(well_bbox.y - labware_bbox.y, 0)
+                            ),
```
SyntaxColoring, line 170: Is the `abs` in this `abs(max(...))` construction just for symmetry with the `min` one? As far as I can tell it's not doing anything?

SyntaxColoring, line 170: Given that we're calling `labware_bbox.y` the expected grip width, should `grip_width_uncertainty_wider` be 0?

For example, say you have a labware that's 86 mm at its widest and whose wells form a 76 mm wide rectangle. This calculation for `grip_width_uncertainty_narrower` says that the gripper should be prepared for the labware to be as little as 76 mm wide, which makes sense. But this calculation for `grip_width_uncertainty_wider` says that the gripper should be prepared for the labware to be as much as 96 mm wide, which is unnecessarily uncertain?

SyntaxColoring (commented):
A couple of questions, but the Protocol Engine parts otherwise looks good to me. Thanks!

vegano1 (approved):
This is awesome

api/src/opentrons/protocol_engine/execution/labware_movement.py
sfoster1: The idea here is that all these values are _numerically_ positive, the semantic difference is in the names. So just to make sure that's always true, `abs`

sfoster1: It should be 0 and in practice it will be zero - the calculation you're calling out is the reason for the `max`. If `well_bbox.y` = 76 and `labware_bbox.y` = 86, `max(well_bbox.y - labware_bbox.y, 0) = max(76 - 86, 0) = max(-10, 0) = 0`

PR #14363, opened 2024-01-26
feat(protocol-designer): test coverage to modals, FileSideBar, FilePage, EditModules

jerader:
partially closes RAUT-943

# Overview

Started adding some test coverage back to PD components since Enzyme was deprecated a few months ago so we lost a lot of PD test coverage. I will add more test coverage in a follow up PR.

# Test Plan

Just confirm that my tests make sense and the checks pass. I didn't alter any behavior in PD

# Changelog

- add a bunch of test coverage

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/FilePage.tsx
```diff
@@ -30,23 +30,12 @@ import formStyles from '../components/forms/forms.css'
...
-
-export interface Props {
-  formValues: FileMetadataFields
-  instruments: React.ComponentProps<typeof InstrumentGroup>
-  goToNextPage: () => unknown
-  saveFileMetadata: (fileMetaDataFields: FileMetadataFields) => void
-  swapPipettes: () => unknown
-  modules: ModulesForEditModulesCard
-  t: any
-}
```
jerader, line 43: this wasn't being used at all. I forgot to remove it in my class/connect function PR

protocol-designer/src/components/FileSidebar/FileSidebar.tsx
```diff
@@ -132,7 +132,7 @@ function getWarningContent({
     .map(moduleOnDeck => t(`modules:module_long_names.${moduleOnDeck.type}`))
     .join(' and ')
 
-  if (pipettesWithoutStep.length && modulesWithoutStep.length) {
+  if (pipettesWithoutStep.length > 0 && modulesWithoutStep.length > 0) {
```
mjhuff: This is ok! Nit, I do think `pipettesWithoutStep.length` is fine or `Boolean(pipettesWithoutStep.length)` is the more common way to do this if you want to be very explicit.

mjhuff (approved):
Looks great! Left one non-blocking comment. Thanks for adding all these tests!

protocol-designer/src/components/FileSidebar/FileSidebar.tsx
jerader: thank you! for some reason i thought this was a bug the way it was before. I'll clean it up using `Boolean(pipettesWithoutStep.length)`

PR #14369, opened 2024-01-29
chore(app-shell): update serial port error to say "Flex"

ecormany:
# Overview

Change "OT-3" to "Flex" in a serial error message. A user noticed this and was confused by it.

# Test Plan

Automated tests only?

# Changelog

Changed one error string and associated comment.

# Review requests

Make sure we don't expect the previous exact wording somewhere else.

# Risk assessment

v low.


mjhuff (approved):
📝 
Doesn't appear we use OT-3 elsewhere for debug messages.

PR #14370, opened 2024-01-29
docs(api): Hide `TrashBin.__init__()` signature and clarify it's for both robots

SyntaxColoring:
Users are never meant to construct `TrashBin`s directly. They're supposed to get them from `ProtocolContext.load_trash_bin()`. Our docs were showing `TrashBin`'s `__init__()` signature, so let's hide it to clean things up and avoid misleading people.

`TrashBin` also said it was only for Flexes, but it's returned for OT-2s also. (See `ProtocolContext.fixed_trash`.)

## Before

Screenshot 2024-01-29 at 11 17 26 AM

## After

Screenshot 2024-01-29 at 11 19 03 AM

### DIFF
--- a/api/docs/v2/conf.py
+++ b/api/docs/v2/conf.py
@@ -444,6 +444,5 @@ nitpick_ignore_regex = [
     ("py:class", r".*protocol_api\.deck.*"),
     ("py:class", r".*protocol_api\.config.*"),
     ("py:class", r".*opentrons_shared_data.*"),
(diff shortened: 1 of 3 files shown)


ecormany (approved):
LGTM locally and in sandbox!

We could/should also remove this line in conf.py: https://github.com/Opentrons/opentrons/blob/88de889e919413a7bbf39b6448b155aec6895f94/api/docs/v2/conf.py#L447

Only the ``TrashBin`` signature was generating that kind of warning.

PR #14371, opened 2024-01-29
fix(api): Addition of Cutout Fixtures to slot height checks

CaseyBatten:
# Overview
This seeks to fix issues related to RSS-375 and RSS-422.

Of note, this will only fix issues during runtime. Currently when in partial tip configuration the engine will attempt to check the height of object adjacent to the slot that an action (such as pickup tip) is occurring in. However, only the heigh of labware and module is checked, the system does not check for the height of fixtures loaded into the deck configuration.

To solve this, a method has been added to allow for searching of Cutout fixtures by deck slot name, and then the height of a returned fixture is then checked alongside the labware and module heights. If, for example a waste chute is loaded in D3 and a partial tip pickup is attempted in D2, the engine will raise an error.

# Test Plan
Ensure that a protocol with a waste chute loaded in D3  with a 96 channel in partial tip configuration set to pick up tips from column A12 of a tiprack in D2 using column A1 of the instrument raises an error during operation before a collision would occur.
The following protocol ought to produce the expected results:

```python

requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(protocol_context):
	tip_rack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "D2")
	instrument = protocol_context.load_instrument('flex_96channel_1000', mount="left", tip_racks=[tip_rack1])

	waste_chute = protocol_context.load_waste_chute()

	instrument.pick_up_tip(tip_rack1)
	instrument.drop_tip(waste_chute)
```

# Risk assessment
While the robot will prevent a collision during runtime, a protocol where a crash would occur would still pass analysis. This is in part due to the fact that we do not load a true deck configuration for analysis, so when checking the set of loaded fixtures we see that there are "None", even if that is untrue. This is yet another item that would become simpler with multipass analysis.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/addressable_areas.py
+++ b/api/src/opentrons/protocol_engine/state/addressable_areas.py
@@ -460,6 +460,19 @@ class AddressableAreaView(HasState[AddressableAreaState]):
             z=position.z,
         )
 
+    def get_fixture_by_deck_slot_name(self, slot_name: DeckSlotName) -> (str | None):
+        """Get the Cutout Fixture ID of a fixture currently loaded into a specific Deck Slot, if one exists."""
+        for cutout in CUTOUT_TO_DECK_SLOT_MAP:
+            if CUTOUT_TO_DECK_SLOT_MAP[cutout] == slot_name:
+                deck_config = (
+                    self.state.deck_configuration
+                )  # we need to use that get cutout fixture by id thing
+                if deck_config:
+                    for deck_item in deck_config:
+                        if cutout in deck_item:
+                            return deck_item[1]
+        return None
+
     def get_fixture_height(self, cutout_fixture_name: str) -> float:
         """Get the z height of a cutout fixture."""
         cutout_fixture = deck_configuration_provider.get_cutout_fixture(
--- a/api/src/opentrons/protocol_engine/state/addressable_areas.py
+++ b/api/src/opentrons/protocol_engine/state/addressable_areas.py
@@ -20,6 +20,7 @@ from ..errors import (
     AreaNotInDeckConfigurationError,
     SlotDoesNotExistError,
     AddressableAreaDoesNotExistError,
+    CutoutDoesNotExistError,
 )
 from ..resources import deck_configuration_provider
 from ..types import (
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Looks good to me, I like the way the logic is broken up.

SyntaxColoring:
Tested on a Flex. Unfortunately, it looks like this is raising an internal `KeyError`.

Screenshot 2024-01-31 at 4 39 20 PM

```
  File "/opt/opentrons-robot-server/opentrons/protocols/execution/execute_python.py", line 69, in run_python
    exec("run(__context)", new_globs)

  File "<string>", line 1, in <module>

  File "casey_pr_test.py", line 17, in run

  File "/opt/opentrons-robot-server/opentrons/protocols/api_support/util.py", line 383, in _check_version_wrapper
    return decorated_obj(*args, **kwargs)

  File "/opt/opentrons-robot-server/opentrons/protocol_api/instrument_context.py", line 933, in pick_up_tip
    self._core.pick_up_tip(

  File "/opt/opentrons-robot-server/opentrons/protocol_api/core/engine/instrument.py", line 404, in pick_up_tip
    deck_conflict.check_safe_for_pipette_movement(

  File "/opt/opentrons-robot-server/opentrons/protocol_api/core/engine/deck_conflict.py", line 213, in check_safe_for_pipette_movement
    _check_deck_conflict_for_96_channel(

  File "/opt/opentrons-robot-server/opentrons/protocol_api/core/engine/deck_conflict.py", line 352, in _check_deck_conflict_for_96_channel
    _check_conflict_with_slot_item(

  File "/opt/opentrons-robot-server/opentrons/protocol_api/core/engine/deck_conflict.py", line 336, in _check_conflict_with_slot_item
    slot_highest_z = engine_state.geometry.get_highest_z_in_slot(

  File "/opt/opentrons-robot-server/opentrons/protocol_engine/state/geometry.py", line 154, in get_highest_z_in_slot
    slot_item = self.get_slot_item(slot.slotName)

  File "/opt/opentrons-robot-server/opentrons/protocol_engine/state/geometry.py", line 702, in get_slot_item
    maybe_fixture = self._addressable_areas.get_fixture_by_deck_slot_name(

  File "/opt/opentrons-robot-server/opentrons/protocol_engine/state/addressable_areas.py", line 488, in get_fixture_by_deck_slot_name
    for fixture in potential_fixtures[slot_cutout_id]:
```

[Run log](https://gist.github.com/SyntaxColoring/5bada5f555ef4939cf619e9e0f8cc581)

PR #14373, opened 2024-01-29
fix(api): flex has no z retract distance

sfoster1:
On the OT-2, when we home the Z axes we move up until we hit the switch; set that to the home position; and then move down 2mm. We have a corresponding config element that captures this 2mm distance called z_retract_distance, and when computing the max instrument height we subtract it from the stored home position. When we worked on the flex, because we inherited most of the code, we inherited this behavior.

However, this is actually not correct behavior on the flex. On the flex, when we home we move up until we hit the switch (well, occlude the photointerrupter, but potato potato); we move down until the switch is released; and we set _that_ to be the home position. So if we later take 2mm off the top for a retract distance, then we'll command a move above the physical home position and hit the stops, causing a stall.

Closes RQA-2256


ahiuchingau (approved):
Yes yes yes yes yes

PR #14374, opened 2024-01-29
fix(app): handle stale protocol analysis in protocol list and slideouts

smb2268:
Fix RAUT-923

# Overview
Add a new analysisStatus for "stale" analysis. If an analysis is stale, we'll hide all protocol data and prompt reanalysis

# Test Plan
With protocols that have a stale analysis (ie missing liquids, modules, labware at the top level):
1. See that in the protocol details page data is missing and reanalysis banner appears
2. See that banner appears on send to flex slideout
3. See that the choose a protocol slideout no longer causes a white screen, and stale protocols are given failed analysis treatment
4. When clicking into a stale protocol to the details page, see that reanalysis is triggered 

Screen Shot 2024-01-29 at 3 54 57 PM
Screen Shot 2024-01-29 at 3 54 23 PM
Screen Shot 2024-01-29 at 3 53 55 PM

# Changelog

1. Add new stale analysis status
6. Anywhere we branch for status, add stale state that matches failed/missing data
7. Add out of date analysis banner with reanalyze option to protocols list
8. Add out of date banner to send to flex slideout
9. Use state analysis status to trigger reanalysis from protocol details page
10. Gate nested fields when analysis data is missing for choose a protocol slideout

# Review requests
Look over code, I ran through the test plan

# Risk assessment
Low


app/src/assets/localization/en/protocol_details.json
```diff
@@ -41,6 +41,7 @@
   "proceed_to_setup": "Proceed to setup",
   "protocol_designer_version": "Protocol Designer {{version}}",
   "protocol_failed_app_analysis": "This protocol failed in-app analysis. It may be unusable on robots without custom software configurations.",
+  "protocol_outdated_app_analysis": "This protocol has an out of date in-app analysis. If the protocol has changed since its last analysis, results may differ.",
```
ecormany: Trying to shorten this text so it doesn't run onto a 4th line in the banner. Maybe…
```suggestion
  "protocol_outdated_app_analysis": "This protocol's analysis is out of date. It may produce different results if you run it now.",
```

app/src/assets/localization/en/protocol_list.json
```diff
@@ -7,8 +7,10 @@
   "modules": "modules",
   "no_data": "No data",
   "protocol_analysis_failure": "Protocol analysis failure.",
+  "protocol_analysis_outdated": "Protocol analysis is out of date.",
   "protocol_deleted": "Protocol deleted",
   "reanalyze_or_view_error": "<analysisLink>Reanalyze</analysisLink> protocol or view <errorLink>error details</errorLink>",
+  "reanalyze_to_view": "<analysisLink>Reanalyze</analysisLink> protocol to view details",
```
ecormany: I think we should remove `to view details`. I think it's shorthand for "to see more info on the protocol details screen", but it's juxtaposed with "view error details" in failed analysis banners — a very different kind of "details".

ecormany (commented):
Left a couple comments on shortening and improving copy. 

We should probably get the final strings back into the designs, and use them as a prototype for anywhere else we want to display info about out-of-date analyses.

mjhuff (approved):
lgtm w/ Ed's feedback!

PR #14375, opened 2024-01-29
refactor(api): separate serial logging

sfoster1:
We have serial logging for anything that involves sending commands to other hardware. On the OT-2, that meant the smoothie/cmb and any modules. On the Flex, it means modules, rear panel, and canbus. All these things can interfere with each other and they all get jammed in `opentrons-api-serial`.

Instead, we'll keep the smoothie and module logs in `opentrons-api-serial` but pipe the can logs to
`opentrons-api-serial-can` and the rear panel to
`opentrons-api-serial-usbbin`. Then we'll gather them all up when you hit the logs endpoint, but now
- If you're in an ssh or serial term session and want to see just the can logs, you finally are able to: journalctl -f -t opentrons-api-serial-canbus
- If you're reading log exports after the fact and want to look at just the can logs, you can now search for opentrons-api-serial-canbus and it will do what you want

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/system/log_control.py
+++ b/api/src/opentrons/system/log_control.py
@@ -7,7 +7,7 @@ This is the implementation of the endpoints in
 import asyncio
 import logging
 import subprocess
-from typing import Tuple
+from typing import Tuple, List
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


vegano1 (approved):
lgtm

PR #14376, opened 2024-01-29
docs(api): overhaul `ProtocolContext` docstrings

ecormany:
# Overview

Fresh coat of paint on all the docstrings for `ProtocolContext` and its children.

Addresses RTC-203.

# Test Plan

Check the [sandbox](http://sandbox.docs.opentrons.com/docstrings-protocolcontext/v2/new_protocol_api.html#module-opentrons.protocol_api).
No warnings anymore! 🥳 

# Changelog

- Prose changes to nearly all docstrings.
- Added some cross-reference anchors.
- Pedantic change to an error string.

# Review requests

- A close read of the prose.
- A sanity check that I haven't restated something in a way that incorrectly captures real API behavior.

# Risk assessment

nil, docstrings

### DIFF
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -180,14 +180,17 @@ class ProtocolContext(CommandPublisher):
 
         This value is set when the protocol context
         is initialized. 
-        
-          - When the context is the argument of ``run()``, the ``"apiLevel"`` key of the metadata or requirements dictionary determines ``api_version``. 
-          - When the context is instantiated with :py:meth:`opentrons.execute.get_protocol_api` or :py:meth:`opentrons.simulate.get_protocol_api`, the value of its ``version`` argument determines ``api_version``.
-        
-        
-        It may be lower than the highest version supported
-        by the robot software. For the highest version supported by the
-        robot software, see ``protocol_api.MAX_SUPPORTED_VERSION``.
+
+          - When the context is the argument of ``run()``, the ``"apiLevel"`` key of the
+            metadata or requirements dictionary determines ``api_version``.
+          - When the context is instantiated with
+            :py:meth:`opentrons.execute.get_protocol_api` or
+            :py:meth:`opentrons.simulate.get_protocol_api`, the value of its ``version``
+            argument determines ``api_version``.
+
+        It may be lower than the :ref:`maximum version <max-version>` supported by the
+        robot software, which is accessible via the
+        ``protocol_api.MAX_SUPPORTED_VERSION`` constant.
         """
         return self._api_version
 
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -230,36 +230,21 @@ class ProtocolContext(CommandPublisher):
     @property
     @requires_version(2, 0)
     def max_speeds(self) -> AxisMaxSpeeds:
-        """Per-axis speed limits when moving this instrument.
+        """Per-axis speed limits for moving instruments.
 
-        Changing this value changes the speed limit for each non-plunger
-        axis of the robot, when moving this pipette. Note that this does
-        only sets a limit on how fast movements can be; movements can
-        still be slower than this. However, it is useful if you require
-        the robot to move much more slowly than normal when using this
-        pipette.
+        Changing values within this property sets the speed limit for each non-plunger
+        axis of the robot. Note that this property only sets upper limits on movement
+        speed.
 
-        This is a dictionary mapping string names of axes to float values
-        limiting speeds. To change a speed, set that axis's value. To
+        This property is a dict mapping string names of axes to float values
+        of maximum speeds in mm/s. To change a speed, set that axis's value. To
         reset an axis's speed to default, delete the entry for that axis
         or assign it to ``None``.
 
-        For instance,
-
-        .. code-block:: py
-
-            def run(protocol):
-                protocol.comment(str(right.max_speeds))  # '{}' - all default
-                protocol.max_speeds['A'] = 10  # limit max speed of
-                                               # right pipette Z to 10mm/s
-                del protocol.max_speeds['A']  # reset to default
-                protocol.max_speeds['X'] = 10  # limit max speed of x to
-                                               # 10 mm/s
-                protocol.max_speeds['X'] = None  # reset to default
+        See :ref:`axis_speed_limits` for examples.
 
         .. caution::
-            This property is not yet supported on
-            :ref:`API version <v2-versioning>` 2.14 or higher.
+            This property is not yet supported in API version 2.14 or higher.
         """
         if self._api_version >= ENGINE_CORE_API_VERSION:
             # TODO(mc, 2023-02-23): per-axis max speeds not yet supported on the engine
@@ -333,19 +318,18 @@ class ProtocolContext(CommandPublisher):
         location: Union[DeckLocation, OffDeckType],
         label: Optional[str] = None,
     ) -> Labware:
-        """Specify the presence of a piece of labware on the OT2 deck.
+        """Specify the presence of a labware on the deck.
 
-        This function loads the labware definition specified by `labware_def`
-        to the location specified by `location`.
+        This function loads the labware definition specified by ``labware_def``
+        to the location specified by ``location``.
 
-        :param labware_def: The labware definition to load
+        :param labware_def: The labware's definition.
         :param location: The slot into which to load the labware,
                          such as ``1``, ``"1"``, or ``"D1"``. See :ref:`deck-slots`.
         :type location: int or str or :py:obj:`OFF_DECK`
-        :param str label: An optional special name to give the labware. If
-                          specified, this is the name the labware will appear
-                          as in the run log and the calibration view in the
-                          Opentrons app.
+        :param str label: An optional special name to give the labware. If specified,
+            this is how the labware will appear in the run log, Labware Position
+            Check, and elsewhere in the Opentrons App and on the touchscreen.
         """
         load_params = self._core.add_labware_definition(labware_def)
 
@@ -369,7 +353,7 @@ class ProtocolContext(CommandPublisher):
     ) -> Labware:
         """Load a labware onto a location.
 
-        For labware already defined by Opentrons, this is a convenient way
+        For Opentrons-verified labware, this is a convenient way
         to collapse the two stages of labware initialization (creating
         the labware and adding it to the protocol) into one.
 
@@ -377,8 +361,8 @@ class ProtocolContext(CommandPublisher):
         later in the protocol.
 
         :param str load_name: A string to use for looking up a labware definition.
-            You can find the ``load_name`` for any standard labware on the Opentrons
-            `Labware Library <https://labware.opentrons.com>`_.
+            You can find the ``load_name`` for any Opentrons-verified labware on the
+            `Labware Library <https://labware.opentrons.com>`__.
 
         :param location: Either a :ref:`deck slot <deck-slots>`,
             like ``1``, ``"1"``, or ``"D1"``, or the special value :py:obj:`OFF_DECK`.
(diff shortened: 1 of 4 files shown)


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -169,12 +176,22 @@ def __init__(
...
+            :ref:`metadata <tutorial-metadata>` or :ref:`requirements
+            <tutorial-requirements>` dictionary determines ``api_version``.
+          - When the context is instantiated with
+            :py:meth:`opentrons.execute.get_protocol_api` or
+            :py:meth:`opentrons.simulate.get_protocol_api`, the value of its ``version``
+            argument determines ``api_version``.
+
+        It may be lower than the :ref:`maximum version <max-version>` supported by the
+        robot software, which is accessible via the
+        ``protocol_api.MAX_SUPPORTED_VERSION`` constant.
```
jwwojak, line 194: Hey, all of this is nicely done.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -193,10 +210,10 @@ def _hw_manager(self) -> HardwareManager:
...
         """Accessor for data files bundled with this protocol, if any.
 
-        This is a dictionary mapping the filenames of bundled datafiles, with
-        extensions but without paths (e.g. if a file is stored in the bundle as
-        ``data/mydata/aspirations.csv`` it will be in the dict as
-        ``'aspirations.csv'``) to the bytes contents of the files.
+        This is a dictionary mapping the filenames of bundled datafiles to the bytes
+        contents of the files. The filename keys are formatted with extensions but
+        without paths. For example, a file stored in the bundle as
+        ``data/mydata/aspirations.csv`` will have the key ``"aspirations.csv"``.
```
jwwojak: I understand the latter part of this text about the filename keys formatted with extensions and without paths, and the `aspirations.csv` example is clear and supportive of the previous sentence.

This part in _italics_ is confusing me:
> This is a dictionary mapping the filenames of bundled datafiles _to the bytes contents of the files_.

It is the "bytes contents" part that gives pause. I don't understand what this means. Might it be clearer to say just the contents of the files?

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -213,36 +230,21 @@ def __del__(self) -> None:
...
-        still be slower than this. However, it is useful if you require
-        the robot to move much more slowly than normal when using this
-        pipette.
+        Changing values within this property sets the speed limit for each non-plunger
+        axis of the robot. Note that this property only sets upper limits on movement
+        speed.
 
-        This is a dictionary mapping string names of axes to float values
-        limiting speeds. To change a speed, set that axis's value. To
+        This property is a dict mapping string names of axes to float values
```
jwwojak, line 240: Can we use a non-developer, non-Python replacement for the word "dict"?

This is one of those moments where the text is semantically correct, but the terminology obscures meaning. If you're not a Python dev, a regular Python user, or an enthusiast, I'd suggest this term, while properly used in this docstring context, reduces clarity rather than improves it. A plain English substitute is recommended here.  

The liturgy doesn't have to be pre-Vatican II.  Face the congregation, use the vernacular.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -338,16 +353,16 @@ def load_labware(
     ) -> Labware:
         """Load a labware onto a location.
 
-        For labware already defined by Opentrons, this is a convenient way
+        For Opentrons-verified labware, this is a convenient way
```
jwwojak, line 357: Simple change, huge improvement. Struck me as very nicely done.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -825,29 +846,32 @@ def load_instrument(
         tip_racks: Optional[List[Labware]] = None,
         replace: bool = False,
     ) -> InstrumentContext:
-        """Load a specific instrument required by the protocol.
+        """Load a specific instrument for use in the protocol.
+
+        When analyzing the protocol on the robot, instruments loaded with this method
+        are compared against the instruments attached to the robot. You won't be able to
+        start the protocol until the correct instruments are attached and calibrated.
```
jwwojak, line 854: Consider "compared to" instead of "compared with." Then, change the next sentence to "You cannot start the protocol ..."

```suggestion
        When analyzing the protocol on the robot, instruments loaded with this method
        are compared to the instruments attached to the robot. You cannot
        start the protocol until the correct instruments are attached and calibrated.
```

jwwojak (commented):
These are all good changes. Left a few comments for consideration, but those shouldn't stop the process. LGTM.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: Yeah, I didn't like "bytes contents" either. The whole point is that the value is returned as a `bytes` type. Gonna reword this.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: Elsewhere in the docs, sure. In the API Reference, the whole point is to give the exact structure.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: It's subtle, but I think the tense here makes a useful distinction. "When doing X…you cannot Y" makes X and Y sound contemporaneous. But analysis always precedes the user starting the protocol.

jwwojak (approved):
LGTM! 

:shipit:

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -87,17 +87,24 @@ class HardwareManager(NamedTuple):
...
+    Methods generally fall into one of two categories.
+
+      - They can change the state of the ``ProtocolContext`` object, such as adding
+        pipettes, hardware modules, or labware to your protocol.
+      - They can control the flow of a running protocol, such as pausing, displaying
+        messages, or controlling built-in robot hardware like the ambient lighting.
+
+    Do not instantiate a ``ProtocolContext`` directly for protocols that you plan to run
+    via the Opentrons App or touchscreen. The ``run()`` function of your protocol does
+    that for you. See the :ref:`Tutorial <run-function>` for more information.
```
SyntaxColoring: As far as I know, there's *never* any legitimate reason or legitimate way for a user to directly instantiate a `ProtocolContext`, so we can simplify this a little.

```suggestion
    Do not instantiate a ``ProtocolContext`` directly.
    The ``run()`` function of your protocol does that for you.
    See the :ref:`Tutorial <run-function>` for more information.
```

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -87,17 +87,24 @@ class HardwareManager(NamedTuple):
...
+
+    Use :py:meth:`opentrons.execute.get_protocol_api` to instantiate a ``ProtocolContext`` when
+    using Jupyter Notebook. See :ref:`advanced-control`.
```
SyntaxColoring, line 107: I've never been quite sure how to concisely describe this, but in addition to Jupyter Notebook, `opentrons.execute.get_protocol_api()` is how you get a `ProtocolContext` from a standalone Python script that you've manually installed on the robot.

Your call for whether that's worth mentioning here. Certainly, it's the `advanced-control` section's responsibility to explain all of those nuances, anyway, not this random docstring.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -213,36 +231,21 @@ def __del__(self) -> None:
...
 
-        Changing this value changes the speed limit for each non-plunger
-        axis of the robot, when moving this pipette. Note that this does
-        only sets a limit on how fast movements can be; movements can
-        still be slower than this. However, it is useful if you require
-        the robot to move much more slowly than normal when using this
-        pipette.
+        Changing values within this property sets the speed limit for each non-plunger
+        axis of the robot. Note that this property only sets upper limits on movement
+        speed.
```
SyntaxColoring: I don't think the old docstring was any better, but I think what it was getting at was that you can't get the robot to go zoomzoomzoom by setting these to `9999999999`. We have our own internal speed limits internally that you can't turn off.

I *think* that's true, anyway?

Your call for whether that deserves clarification.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -293,6 +296,20 @@ def on_command(message: cmd_types.CommandMessage) -> None:
...
+        You can evaluate the result of this method in an ``if`` statement to make your
+        protocol behave differently in different environments. For example, you could
+        refer to a data file on your computer when simulating and refer to a data file
+        stored on the robot when not simulating.
+
+        You can also use it to skip time-consuming aspects of your protocol. Most Python
+        Protocol API methods, like :py:meth:`.delay`, are designed to evaluate
+        instantaneously in simulation. But external methods, like those from the
+        :py:mod:`time` module, will run at normal speed if not skipped.
+        """
```
SyntaxColoring, line 312: 🙌

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -338,16 +354,16 @@ def load_labware(
...
         the labware and adding it to the protocol) into one.
 
         This function returns the created and initialized labware for use
         later in the protocol.
 
         :param str load_name: A string to use for looking up a labware definition.
-            You can find the ``load_name`` for any standard labware on the Opentrons
-            `Labware Library <https://labware.opentrons.com>`_.
+            You can find the ``load_name`` for any Opentrons-verified labware on the
+            `Labware Library <https://labware.opentrons.com>`__.
```
SyntaxColoring, line 366: I'm not sure about replacing "standard" with "Opentrons-verified."

* I think [the support team is calling it "standard labware"](https://support.opentrons.com/s/article/Using-labware-in-your-protocols), contrasted with "custom labware."
* Labware lie in strata of Opentrons-blessedness. In a lower stratum, there's just the fact that we include a definition for the labware by default. At a higher stratum, there's Opentrons ["testing with and recommending"](https://support.opentrons.com/s/article/What-labware-can-I-use-with-my-modules) the labware. [A labware being "recommended" for use with a module is better than it just being "compatible" with the module](https://support.opentrons.com/s/article/Protocol-Designer-Labware-and-module-compatibility). And then there's the gold "verified" badge on [labware that we sell](https://shop.opentrons.com/tough-0.2-ml-96-well-pcr-plate-full-skirt/), whatever the hell that means.

  Here, we're really just talking about labware whose definitions are included by default. I don't think we want to imply anything more than that.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -338,16 +354,16 @@ def load_labware(
     ) -> Labware:
         """Load a labware onto a location.
 
-        For labware already defined by Opentrons, this is a convenient way
+        For Opentrons-verified labware, this is a convenient way
```
SyntaxColoring, line 357: ditto previous point about "verified"

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -357,25 +373,30 @@ def load_labware(
...
-              * ``"opentrons"``, to load standard Opentrons labware definitions.
-              * ``"custom_beta"``, to load custom labware definitions created with the
-                `Custom Labware Creator <https://labware.opentrons.com/create>`_.
+              - ``"opentrons"``, to load standard Opentrons labware definitions.
+              - ``"custom_beta"``, to load custom labware definitions created with the
+                `Custom Labware Creator <https://labware.opentrons.com/create>`__.
 
             You might need to specify an explicit ``namespace`` if you have a custom
-            definition whose ``load_name`` is the same as an Opentrons standard
+            definition whose ``load_name`` is the same as an Opentrons-verified
```
SyntaxColoring, line 388: ditto previous point about "verified"

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1022,11 +1046,11 @@ def location_cache(self, loc: Optional[Location]) -> None:
     @requires_version(2, 0)
     def deck(self) -> Deck:
         """An interface to provide information about what's currently loaded on the deck.
-        This object is useful for determining if a slot in the deck is free.
+        This object is useful for determining if a slot on the deck is free.
```
SyntaxColoring, line 1049: Oh god, the "is labware *in* or *on* a slot" question goes all the way down.

Is a *deck* *in* a robot, or *on* it?!

SyntaxColoring (approved):
Love it. Thank you!

Various small comments. The one bigger thing is "standard labware" / "verified labware" terminology.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: 🥴 I forgot that was even an option. But of course, there is an `opentrons_execute` CLI tool.

I agree that this is best covered in the Advanced Control page, and not here.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: Slight clarification seems fair. There are other properties sprinkled throughout the API where you can set a value of eleventy-billion to no real effect, and we don't go too far out of our way to declare that (common sense) fact.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: We adopted this terminology sometime last year — I know it got checked for inclusion in the Flex manual. "Opentrons verified labware" == items on Labware Library.

PR #14377, opened 2024-01-29
fix(api): temporarily increase Z_L hold current to prevent 96-channel from falling when motor is re-engaged

ahiuchingau:
# Overview
The 96-channel pipette sometimes fall immediately when the Z motor is re-engaged from being held by the ebrake. 

We're theorizing that increasing the hold current temporarily when we re-engage the motor could help it overcome the detent torque asserted when the ebrake was applied, so that the motor would have a chance to maintain its position.


sfoster1 (approved):
Tested on a flex by e-stopping while the 96 was moving down to detach, then releasing e-stop and homing - no drop!

PR #14380, opened 2024-01-30
fix(hardware-testing): Pipette plunger production test homing tolerance

andySigler:
# Overview

Seems like a firmware update somewhere changed the homing behavior so that the reported position is not strictly `0.0` after a home but something very close to that.

The PR updates the pipette QC's homing test to not strictly look for `0.0` after homing but instead `+/-0.005`.

# Test Plan

# Changelog

# Review requests

# Risk assessment


andySigler:
nvm, should base this change off edge, not the old tag. will reopen a new PR in a minute

PR #14381, opened 2024-01-30
fix(hardware-testing): Plunger homing tolerance expanded to match firmware changes

andySigler:
# Overview

Seems like a firmware update somewhere changed the homing behavior so that the reported position is not strictly 0.0 after a home but something very close to that.

The PR updates the pipette QC's homing test to not strictly look for 0.0 after homing but instead +/-0.005.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Looks good to me if it works!

PR #14382, opened 2024-01-30
fix(app): fix app progress bar not displaying 100% when update downloaded

mjhuff:
Closes RQA-2266

# Overview
Fixes a 1x issue where the app progress bar did not show 100% after the update was downloaded. A good cover-all solution for these update issues is to have the progress bar report 100% when the node layer emits that the update was successfully downloaded (as we do with the modal text). 

Screenshot 2024-01-30 at 10 13 30 AM (2)

# Test Plan
Just reason through the code. It's a 1x reported issue that I can't repro. 

# Changelog
- App update progress bar will always be at 100% when the download is complete.

# Risk assessment
low


ncdiehl11 (approved):
couldn't repro either but the logic makes sense to me

PR #14384, opened 2024-01-30
fix(robot-server): fix mqtt dependencies

sfoster1:
We need to add paho-mqtt as an install dep in our setup.py for good-practice reasons and to fix it in the pipfile.lock. More importantly, we need to add it as a buildroot dep in the local buildroot toolchain file so the OT-2 gets it, since otherwise the server can't boot because it won't be present when notification_client tries to import it.

Closes RQA-2270

## testing
- [x] make a buildroot build for this branch and check that it gets paho-mqtt installed and can boot ( https://github.com/Opentrons/buildroot/actions/runs/7714642474 ) 
- [ ] make an openembedded build for this branch and check that it gets paho-mqtt installed and can boot ( https://github.com/Opentrons/oe-core/actions/runs/7714634061 )


vegano1 (approved):
nice, let's see if this builds!

mjhuff (commented):
Thanks for fixing this! Good to know for the future 😅 
The paho-types are there for lint check reasons. I'm not sure how to handle those tests if we remove that dependency.

PR #14386, opened 2024-01-30
feat(api,hardware,robot-server): enable firmware updates for the HEPA/UV.

vegano1:
# Overview

The HEPA/UV can update over CAN, this PR adds the missing defines for this to work properly.

# Test Plan

- [x] Make sure we can update the HEPA/UV fan when we power on the Flex.
- [x] Make sure the robot server starts up properly after updating the HEPA/UV
- [x] Make sure the robot server starts properly without a HEPA/UV filter.

# Changelog

- Add hepa_uv defines for SubSystem, FirmwareUpdateType, and HWSubSystem

# Review requests
# Risk assessment

- low, Flex only, unreleased device


sfoster1 (approved):
Excellent!

PR #14387, opened 2024-01-30
fix(app): ensure firmware takeover modal won't open uneccessarily

smb2268:
fix RQA-2236

# Overview
This PR adds a few layers of protection against opening the ODD Firmware Update Takeover modal unnecessarily.

# Test Plan
1. Poke around on the ODD after having updated an instrument, ensure the modal doesn't pop up again
2. If you attach an instrument that needs an update, see the modal but don't interact with it before removing the instrument, the pop up should close

# Changelog
1. Add logic to the use effect that will wipe the `instrumentsToUpdate` array if the modal isn't open and there are currently no instruments that need updating. This will ensure that if an instrument is updated elsewhere in the app, it won't remain in this array.
3. Reset the `instrumentsToUpdate` array and `indexToUpdate` field when closing the takeover modal

# Review requests
With this branch on an ODD, play around with different scenarios that have been producing this bug

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/FirmwareUpdateModal/FirmwareUpdateTakeover.tsx
+++ b/app/src/organisms/FirmwareUpdateModal/FirmwareUpdateTakeover.tsx
@@ -81,10 +81,19 @@ export function FirmwareUpdateTakeover(): JSX.Element {
     ) {
       setShowUpdateNeededModal(true)
     }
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Tested, looks good to me!

PR #14388, opened 2024-01-30
fix(robot-server): Blink the OT-2's button light while the persistence layer initializes

SyntaxColoring:
# Overview

Closes RSS-217.

# Test Plan

* [x] This should not affect the Flex's status bar during its startup sequence.
* [x] On an OT-2, the front button light should blink during startup (as it currently does). After hardware initialization completes or database migration completes — *whichever comes later* — it should turn solid blue.
  * [x] Test case 1: no migration, and the robot needs to home, so hardware initialization completes later
  * [x] Test case 2: migration required, so database initialization completes later

# Changelog

|                                | **Before this PR**                    | **With this PR**                                                                                      |
|--------------------------------|---------------------------------------|--------------------------------------------------------------------------------------------------------|
| **When do we start blinking?** | As soon as hardware becomes available | As soon as hardware becomes available                                                                  |
| **When do we stop blinking?**  | When hardware is fully initialized    | When hardware is fully initialized *or* when persistence is fully initialized, whichever happens later |

# Review requests

Any architectural feedback? robot-server's startup and shutdown code has become a bit of a mess, and this is making it worse. I have some rough ideas...

1. We could combine this new code with the Flex's existing `LightController`, which controls the Flex's status bar. Then, whatever mess we have to tolerate, we only have to tolerate it once.

   I haven't done that here because it begs a product design question: to what extent do we want the OT-2's front button light and the Flex's status bar to be the "same thing"? The Flex's status bar currently displays run state, whereas the OT-2's front button light doesn't. But maybe they both should?

1. The `done_callbacks` thing that I've added to `start_initializing_persistence()` copies what we did in `start_initializing_hardware()`. But I'd rather not do this, if we can help it. Passing these callbacks around is obscuring control flow and making the concurrency and exception handling harder to get right.

   It seems like this is caused by `start_initializing_persistence()` and `start_initializing_hardware()` doing too much. They coordinate startup logic, *and then* they insert the results into `app.state` for HTTP endpoints to pick up later. This was good for encapsulation, but now it's causing problems because our initialization code can't easily chain things together. Maybe this gets better if `app_setup.py` becomes the place where we insert things into `app.state`.

# Risk assessment

Mmmmmmedium? It's concurrency and it's in a place that's difficult to test, except manually.


robot-server/robot_server/app_setup.py
```diff
@@ -74,12 +79,17 @@ async def on_startup() -> None:
...
             (mark_light_control_startup_finished, False),
+            # OT-2 light control:
+            (start_blinking_front_button_light, True),
         ],
     )
     start_initializing_persistence(
-        app_state=app.state, persistence_directory_root=persistence_directory
+        app_state=app.state,
+        persistence_directory_root=persistence_directory,
+        done_callbacks=[stop_blinking_front_button_light],
```
SyntaxColoring: I think this is wrong because the persistence layer's initialization will race against hardware initialization. We need to stop blinking only when *both* are done.

robot-server/robot_server/hardware.py
```diff
@@ -127,6 +131,111 @@ async def clean_up_hardware(app_state: AppState) -> None:
...
+
+    def _all_complete(self) -> bool:
+        return self._persistence_init_complete and self._hardware_init_complete
+
+
+def fbl_init(app_state: AppState) -> None:
+    if should_use_ot3():
+        # This is only for the OT-2's front button light.
+        # The Flex's status bar is handled elsewhere -- see LightController.
+        return
```
SyntaxColoring, line 193: I think we could actually omit this check and always call `hardware.set_lights(button=...)`, even on a Flex. On a Flex, [it looks like it will be a no-op](https://github.com/Opentrons/opentrons/blob/2bf5015fac9d9f335e1f185f4e974812c959523c/api/src/opentrons/hardware_control/backends/ot3controller.py#L1185-L1189).

Anyone have a preference for what would be less confusing?

CaseyBatten: I think I'd prefer if this remained, just because I'd rather not have a call made on Flex that isn't intended to actually do something, but if it no-ops I suppose it's harmless? Just a difference of style opinions.

CaseyBatten (approved):
Test cases provided appear comprehensive for the purposes of merging this.

As to the product question, I don't have a direct answer. Is there an existing standard as to what the light indicators are meant to convey to the user (code chart, etc) that we would be violating by adding this status to the lights? 

Initialization approach is appropriate for the time being, but to address your concerns maybe we can add this to the list of reasons a refactor would be desirable in the future  to start moving things over to `app_setup.py`. 

Overall, looks good to me and satisfies the needs of RSS-217.

robot-server/robot_server/hardware.py
SyntaxColoring: OK cool, thanks for the second opinion. Leaving it in.

SyntaxColoring:
> Is there an existing standard as to what the light indicators are meant to convey to the user (code chart, etc) that we would be violating by adding this status to the lights?

Aha, [yes, I think there is](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3544940776/Approved+-+Status+light+bar+colors+and+patterns). This is for the Flex. I don't think anything exists for the OT-2. I will ask if the OT-2 should ideally match the Flex.

PR #14389, opened 2024-01-30
chore(hardware): dont ask for barcode serial with print/clear action.

vegano1:
# Overview

Change the order of operations so we don't get asked to input a serial number from barcode when printing/clearing the eeprom, we should only get the serial from barcode when we don't specify any properties to write.

# Test Plan

- [x] Make sure the default action `write` is not changed and we are asked to input a serial number if no property is given.
- [x] Make sure we are not asked to input a serial number when we use the `print` and `clear` actions

# Changelog

- Only get serial from barcode if we are doing a `write` action and there are no properties

# Review requests
# Risk assessment


sfoster1 (approved):
nice

PR #14390, opened 2024-01-30
fix(api): ot2 subsystems is a property

sfoster1:
The OT2 hardware control API actually has an attached_subsystems property, except it wasn't a property, which causes annoying error messages on OT-2.


vegano1 (approved):

vegano1:
lol nice

ecormany:
It works 😏 
image

PR #14391, opened 2024-01-30
fix(g-code-testing): relock for new r-s deps

sfoster1:
Need to get the paho dep in there


SyntaxColoring (changes requested):
I think I fixed this accidentally in https://github.com/Opentrons/opentrons/pull/14280. If we leave edge as-is, it should be good now, with `paho-mqtt` in there. If we merge this, it'll clobber the `notify-server` deletions.

sfoster1:
Oh, great call, thanks for catching that!

PR #14392, opened 2024-01-30
fix(components): make labware info typography styles explicit

b-cooper:
# Overview

In order to prevent wonky rendering in environments with differing global css, be explicit about the font size, weight, and line-height of the text within the labware infor of the Protocol Deck

# Risk assessment

low


y3rsh (approved):
:shipit: 
0.1.3-alpha.0 looks good now in PL
image

PR #14394, opened 2024-01-31
fix(app): fix banners displaying when resetting a run on the desktop app

mjhuff:
Closes RQA-2250

# Overview

This PR fixes a bug in which banners from a cancelled, previous run would render while resetting the run. The problem is more noticable on the OT-2.

### Before

https://github.com/Opentrons/opentrons/assets/64858653/8562f362-7ec0-47d7-81f5-75d001b1b6ba

### With Fix

https://github.com/Opentrons/opentrons/assets/64858653/f9b8edc4-bcc0-4977-89c9-8c6382e86efb

# Test Plan

You can use this protocol to test. Just follow the videos above as a guide. 
[testing_ot2api_decorated_methods.py.zip](https://github.com/Opentrons/opentrons/files/14113952/testing_ot2api_decorated_methods.py.zip)

# Changelog
- Fixed a bug where banners would display for a cancelled run while the run is resetting.

# Risk assessment
low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/ProtocolRunHeader.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/ProtocolRunHeader.tsx
@@ -170,6 +170,7 @@ export function ProtocolRunHeader({
   const [pipettesWithTip, setPipettesWithTip] = React.useState<
     PipettesWithTip[]
   >([])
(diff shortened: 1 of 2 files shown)


ncdiehl11 (approved):
tested on QABOT2, looks good to me

PR #14395, opened 2024-01-31
refactor(robot-server): prevent using pipette settings on flex

sfoster1:
They are fundamentally not necessary or really functional on flex pipettes, and the settings don't really work on the 96 anyway.

### DIFF
--- a/robot-server/robot_server/service/legacy/routers/settings.py
+++ b/robot-server/robot_server/service/legacy/routers/settings.py
@@ -9,6 +9,7 @@ from opentrons_shared_data.errors import ErrorCodes
 from opentrons.hardware_control import (
     HardwareControlAPI,
     dev_types as hardware_dev_types,
+    API,
 )
 from opentrons.hardware_control.types import HardwareFeatureFlags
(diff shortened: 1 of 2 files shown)


robot-server/robot_server/service/legacy/routers/settings.py
```diff
@@ -258,13 +264,13 @@ async def get_robot_settings(
 
 @router.get(
     "/settings/pipettes",
-    description="List all settings for all known pipettes by id",
+    description="List all settings for all known pipettes by id. Only available on OT-2.",
```
sfoster1, line 267: @SyntaxColoring is there something else you'd like to see here about the change? The app wasn't presenting fields for these anyway

SyntaxColoring: Nope, nothing else.

SyntaxColoring (approved):
Awesome, thanks for the cleanup.

You might want to call this a `refactor(robot-server)` instead of a `chore` since it does affect production behavior.

PR #14396, opened 2024-01-31
fix(robot-server): mqtt disconnect cb signature

sfoster1:
The disconnect callback is called with a properties instance if we're running mqtt 5, as with the connect callback. I don't think we're running mqtt 5, but it doesn't matter because it looks like we call with this property during shutdown anyway so this prevents an annoying log message.


mjhuff (approved):
Thanks for catching this! We actually are running v5, which is why MQTT is complaining...I ran into this problem with the _on_connect signature but forgot to update this one.

PR #14398, opened 2024-01-31
fix(components): only show nest labware info on protocol deck when expected

b-cooper:
# Overview

Within the Protocol Deck component, the `showLabwareInfo` prop should also manage the presence or absence of the labware info on module-nested labware.
 
# Risk assessment

Low


y3rsh (approved):
Makes sense to me, good catch, I didn't even know to look for this.

PR #14400, opened 2024-01-31
fix(api): Revert #14371 Addition of Cutout Fixtures to slot height checks

CaseyBatten:
Reverts Opentrons/opentrons#14371 due to bug found on Flex


sfoster1 (approved):
Sure let's do it

PR #14401, opened 2024-01-31
docs(api): overhaul `Labware` docstrings

ecormany:
# Overview

Freshen up the `Labware` class in the API Reference.

Addresses RTC-205.

# Test Plan

Check the [sandbox](http://sandbox.docs.opentrons.com/docstrings-labware/v2/new_protocol_api.html#labware).

# Changelog

- General cleanup of docstrings.
- Added one cross-reference target.

# Review requests

Take a careful look at the simplifications I made in `set_offset`. I think we need to more fully address what the current state of that function is. The ticket referenced in the TODO just above the error handling logic has been closed. I think that function should be usable again, but I won't put the docs ahead of the API's actual capabilities.

# Risk assessment

none, docs

### DIFF
--- a/api/src/opentrons/protocol_api/labware.py
+++ b/api/src/opentrons/protocol_api/labware.py
@@ -7,6 +7,7 @@ and labware calibration offsets. It contains all the code necessary to
 transform from labware symbolic points (such as "well a1 of an opentrons
 tiprack") to points in deck coordinates.
 """
+
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_api/labware.py
```diff
@@ -380,14 +374,16 @@ def __getitem__(self, key: str) -> Well:
     def uri(self) -> str:
         """A string fully identifying the labware.
 
-        :returns: The URI, ``"namespace/loadname/version"``
+        The URI has three parts and follows the pattern ``"namespace/load_name/version"``.
```
jwwojak, line 378: Would it help to add on a "for example" here after defining the pattern?

api/src/opentrons/protocol_api/labware.py
```diff
@@ -740,16 +756,20 @@ def rows_by_index(self) -> Dict[str, List[Well]]:
...
+        For example, access column 1 with ``labware.columns()[0]``.
+        On a standard 96-well plate, this will output a list of :py:class:`.Well`
+        objects containing A1 through H1.
 
-        With indexing one can treat it as a typical python nested list.
-        For example, access row A with ``labware.columns()[0]``.
-        This will output ``['A1', 'B1', 'C1', 'D1'...]``.
+        .. note::
+            This method takes args for backward compatibility, but using args is
+            deprecated and will be removed in a future versions of the API.
```
jwwojak: When we use "deprecated" here, are we using it as a substitute for "don't use or do this thing" or are we saying "you can, but really shouldn't"? It seems the latter.

But just for thought: sometimes it seems like we (in a broader technology community sense, not just Opentrons), use "deprecated" as a too polite way of saying "no" or "don't." If we really mean "don't do this," let's make that stronger and clear.

jwwojak (commented):
Left a couple of comments. Overall, LGTM. Defer to Max for a follow on review/approval.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -296,29 +297,21 @@ def __hash__(self) -> int:
...
-       labware.wells()[0]
-       labware.rows()[0][0]
-       labware.columns()[0][0]
-       labware.rows_by_name()['A'][0]
-       labware.columns_by_name()[0][0]
+    This class represents a piece of labware.
+
+    Labware available in the API generally fall under two categories.
+
+      - Standard labware: well plates, tube racks, reservoirs, tip racks, etc.
```
SyntaxColoring: Is there a word we can use other than "standard" so it won't conflict with the historical distinction of *standard* vs. *custom* labware? (You've told me that distinction is now *verified* vs. *custom* labware, but I still see support articles talking about standard labware, so I suspect there's potential for confusion.)

Even just saying *"Normal" labware* might work.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -1000,7 +1036,7 @@ def return_tips(self, start_well: Well, num_channels: int = 1) -> None:
 
     @requires_version(2, 0)
     def reset(self) -> None:
-        """Reset all tips in a tip rack.
+        """Reset tip tracking for a tip rack.
```
SyntaxColoring, line 1038: Maybe something like:

```suggestion
        """Reset the tracking for which tips have been used.
```

Since I don't know if it's otherwise clear what we mean by "tip tracking."

SyntaxColoring (approved):
Yay, thanks.

api/src/opentrons/protocol_api/labware.py
ecormany: This is "deprecated" in an API contract sense. If anything, this notice is redundant, because that sense means "This feature still exists but we have decided that it will go away."

Making some changes that lean more heavily on the definition of "deprecated" and redirect users to indexing.

api/src/opentrons/protocol_api/labware.py
ecormany: Going with "consumable" — we use that in the Labware chapter of the Flex manual, and it's generally true of all those categories that we think of as "labware-labware" (Ghomeshi et al.: 2004)

jwwojak (commented):
LGTM. No comments.

:shipit:

api/src/opentrons/protocol_api/labware.py
ecormany: Added a lot of explanation. Hopefully not too much!

PR #14402, opened 2024-01-31
fix(app): fix query caching issues with last run command key

mjhuff:
Closes RQA-2275

# Overview

Follow up to #14399. useNotify hooks do not need to support returning useable data from the MQTT broker, since the broker only ever supplies refetch flags currently. Assuming the broker could return real command data was causing strange edge cases in which last run command keys would sometimes return null. Just use HTTP response data for all notify hooks. I'll follow this up with a refactor to simplify the useNotifyService hook as well.

# Test Plan
- Run a protocol, there should be no command flickering. 

# Risk assessment
low


ncdiehl11 (approved):
logic makes sense, tested and verified fix on Desktop and ODD 👍

PR #14405, opened 2024-02-01
chore(monorepo): migrate frontend bundling from webpack to vite

shlokamin:
# Overview

This PR:
- migrates monorepo frontend bundling from webpack to vite
- migrates all jest tests to vitest
- updates storybook

# Test Plan

Because this PR is so massive, the best way to test this is to smoke test: 
- PD
- The desktop app
- The ODD app (on a Flex)
- Labware Library
- Components library (storybook)

# Changelog

- Adopt vite as the frontend bundler
- Adopt vitest as the frontend testing library

# Review requests
This PR is massive so it's not going to be a good use of time to go through every file change. Important ones to review though:

- vitest config
- vite config in each project
- package json in each project
- electron bundler config
- makefiles in each project

# Risk assessment

on a scale from 1-10 probably like a 47


koji:
ToDo for future
 - [ ] Fix hover case
 Seems that `fireEvent.mouseOver(element)` doesn't work.
 Also userEvent.hover() doesn't work...
 
 ```ts
   // with async
    await userEvent.hover(button)
    await waitFor(() => {
      expect(button).toHaveStyle(`background-color: ${COLORS.blue55}`)
    })
 ```
 
the official page's sample doesn't work...  (updated @testing-library/user-event to the latest)
https://testing-library.com/docs/user-event/convenience/#hover
 
 This will be fixed when [this](https://github.com/testing-library/jest-dom/pull/568) is fixed.

mjhuff (approved):
🥳

PR #14411, opened 2024-02-01
feat(app, components, shared-data): disable invalid slots during module cal

ncdiehl11:
closes [RQA-2164](https://opentrons.atlassian.net/browse/RQA-2164)

# Overview

During module calibration, the DeckLocationSelect component needs to be aware of deck configuration. The only compatible cutout fixtures for module calibration should be single slots on the margins (columns 1 and 3). As a placeholder until detailed designs are final, if no slots are available for module calibration, the "confirm placement" button should be disabled. Each disabled slot shows a tooltip on hover describing the reason for the slot being disabled for selection.

# Test Plan

- verified combinations of deck configuration and module location select
Screen Shot 2024-02-01 at 3 50 33 PM
Screen Shot 2024-02-01 at 4 31 47 PM

# Changelog

- query deck config in module calibration orchestration component to limit available slots to single slots fixtures in module-compatible cutouts
- disable proceed button if no locations available
- show tooltip on hover for any disabled slot

# Risk assessment

medium. want to be sure slot selection is enabled for valid module locations

[RQA-2164]: https://opentrons.atlassian.net/browse/RQA-2164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/ModuleWizardFlows/SelectLocation.tsx
```diff
@@ -66,21 +69,23 @@ export const SelectLocation = (
...
+          showTooltipOnDisabled={true}
         />
       }
       bodyText={bodyText}
       proceedButtonText={t('confirm_location')}
       proceed={handleOnClick}
+      proceedIsDisabled={slotName == null}
+      disableProceedReason={
+        slotName == null
+          ? 'Current deck configuration prevents module placement'
```
jerader, line 86: can you add this to i18n?

app/src/organisms/ModuleWizardFlows/index.tsx
```diff
@@ -104,7 +125,7 @@ export const ModuleWizardFlows = (
     createTargetedMaintenanceRun,
     isLoading: isCreateLoading,
   } = useCreateTargetedMaintenanceRunMutation({
-    onSuccess: response => {
+    onSuccess: (response: { data: { id: any } }) => {
```
jerader: is there a better type to use here instead of typing id as `any`?

components/src/hooks/useSelectDeckLocation/index.tsx
```diff
@@ -103,11 +141,13 @@ export function DeckLocationSelect({
             addressableArea.id === 'fixedTrash'
         )
         .map(slot => {
+          // eslint-disable-next-line react-hooks/rules-of-hooks
```
jerader: can you give a little more detail in the comment about why this is needed?

components/src/hooks/useSelectDeckLocation/index.tsx
```diff
@@ -103,11 +141,13 @@ export function DeckLocationSelect({
...
           const slotLocation = { slotName: slot.id }
-          const isDisabled = disabledLocations.some(
-            l =>
-              typeof l === 'object' && 'slotName' in l && l.slotName === slot.id
-          )
+          const isDisabled = !availableSlotNames.some(l => l === slot.id)
```
jerader: nit: can the l be more descriptive? maybe something like `slotName`? 

```suggestion
          const isDisabled = !availableSlotNames.some(slotName => slotName === slot.id)
```

components/src/hooks/useSelectDeckLocation/index.tsx
```diff
@@ -220,6 +271,49 @@ export function DeckLocationSelect({
...
+              fontSize="1.5rem"
+              backgroundColor={COLORS.black90}
+              padding={SPACING.spacing8}
+              borderRadius="3px"
+            >
+              {hoveredData.disabledReason != null
+                ? `A ${getFixtureDisplayName(
+                    hoveredData.disabledReason
+                  ).toLowerCase()} is currently specified here on the deck configuration`
+                : 'Slot unavailable'}
```
jerader: can you add these to i18n?

jerader (commented):
left some comments but this is looking good so far!

app/src/organisms/ModuleWizardFlows/index.tsx
ncdiehl11: thanks for the catch— this is a remnant of testing that I forgot to remove

components/src/hooks/useSelectDeckLocation/index.tsx
ncdiehl11: thanks for this catch— we no longer need this

components/src/hooks/useSelectDeckLocation/index.tsx
ncdiehl11: will do!

app/src/organisms/ModuleWizardFlows/index.tsx
ncdiehl11: good question— I don't think this was introduced in this PR, but I will resolve

mjhuff (approved):
Looks and works great! Awesome job tackling this one! I think it's good to merge after @jerader 's feedback and passing CI.

shared-data/js/fixtures.ts
```diff
@@ -50,6 +50,26 @@ export const FLEX_CUTOUT_BY_SLOT_ID: { [slotId: string]: CutoutId } = {
   D4: 'cutoutD3',
 }
 
+// mapping of Flex cutouts to deck slots
+export const FLEX_SLOT_BY_CUTOUT_ID: { [CutoutId: string]: string } = {
```
brenthagen: `FLEX_SLOT_BY_CUTOUT_ID` can't be mapped from cutout id to slot because a cutout id can provide more than one slot. `cutoutA4` for example doesn't exist - slot `A4` is provided by `cutoutA3` (and same for B4, C4, D4).

I think instead this can be renamed `FLEX_SINGLE_SLOT_BY_CUTOUT_ID` to accomplish what we need, omitting column 4 slots.

app/src/organisms/ModuleWizardFlows/index.tsx
```diff
@@ -64,10 +71,26 @@ export const ModuleWizardFlows = (
       : attachedPipettes.right
 
   const moduleCalibrationSteps = getModuleCalibrationSteps()
+  const deckConfig = useDeckConfigurationQuery().data ?? []
+  const occupiedCutouts = deckConfig.filter(
+    (fixture: CutoutConfig) =>
+      !(fixture.cutoutFixtureId as SingleSlotCutoutFixtureId)?.includes(
```
brenthagen: there's a shared-data constant `SINGLE_SLOT_FIXTURES` and we do this check as `!SINGLE_SLOT_FIXTURES.includes(fixture.cutoutFixtureId` elsewhere

brenthagen (commented):
module calibration looks good, but with the shared code we need to be sure not to affect drop tip wizard (or do so intentionally). I think we're getting the below because `availableSlotNames` is not provided:

Screen Shot 2024-02-05 at 4 59 43 PM

shared-data/js/fixtures.ts
ncdiehl11: This makes sense for a 1:1 relationship between single slots and cutouts

app/src/organisms/ModuleWizardFlows/index.tsx
ncdiehl11: got it, will change this check

brenthagen (approved):
LGTM !

PR #14412, opened 2024-02-01
feat(protocol-designer): deprecate Formik, migrate to React-Hook-Form

jerader:
closes RAUT-953

# Overview

This PR deprecates the usage of Formik in PD and migrates to React-Hook-Form. Additionally, YUP is updated to the latest version.

# Test Plan

Smoke test PD! In particular, mess around with editing modules, configuration (staging areas, trash, waste chute) and pipettes and adding liquids. Those areas were particularly heavy with formik

# Changelog

- refactor a bunch of components to use React-hook-form

# Review requests

see test plan

# Risk assessment

medium-ish, this is a big refactor and deprecates Formik which is a legacy library we have been using.


protocol-designer/src/constants.ts
```diff
@@ -95,43 +95,43 @@ export const MODELS_FOR_MODULE_TYPE: Record<
 > = {
   [MAGNETIC_MODULE_TYPE]: [
     {
-      name: 'Magnetic',
+      name: 'GEN1',
```
jerader, line 98: I definitely introduced this bug in the i18n PR i merged in a few weeks ago. The name values needed to be the GEN instead of the module's short name.

protocol-designer/package.json
```diff
@@ -49,6 +50,6 @@
     "reselect": "4.0.0",
     "ua-parser-js": "^0.7.23",
     "uuid": "3.3.2",
-    "yup": "0.26.6"
+    "yup": "1.3.3"
```
mjhuff, line 53: Wow, nice version bump!

mjhuff (approved):
Looks great! Smoke tested PD without any problems. React-Hook-Form seems like the right choice. Thanks for solving YUP as well.

PR #14413, opened 2024-02-01
fix(api): Center instrument over moveable trash regardless of configuration

CaseyBatten:
# Overview
This seeks to solve problems identified in: RSS-419, RSS-434

Rather than cover many addressable area types, we have chose to special case the moveable trash given the likelihood of an adjacent item obstruction causing an overlap in various partial tip configurations. 

The solution to this is to ensure that the waypoint destination does not utilize the active channel of the pipette (such as a single column on a 96 channel) but rather the center of the destination instead.

We also override the adjacent slot checks in cases such as this, as we know that we will not overlap their slots in instances involving the trashbins because of this reconfiguraiton.

# Test Plan
Ensure the a protocol switching between partial and full tip pickup drops tips within the bounds of a moveable trash the same way regardless of configuration, and does not at any point overlap with adjacent slots as identified in various collision cases.
Test Case:
```

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16",
}

def run(protocol_context: protocol_api.ProtocolContext):
    adapter1 = protocol_context.load_adapter("opentrons_flex_96_tiprack_adapter", "A2")
    adapter2 = protocol_context.load_adapter("opentrons_flex_96_tiprack_adapter", "A1")

    trashA3 = protocol_context.load_trash_bin("A3")

    tiprack1 = adapter1.load_labware("opentrons_flex_96_tiprack_200ul")
    tiprack2 = adapter2.load_labware("opentrons_flex_96_tiprack_200ul")
    partial_tiprack= protocol_context.load_labware("opentrons_flex_96_tiprack_200ul", "B1")

        "flex_96channel_1000", mount="left", tip_racks=[tiprack1, tiprack2, partial_tiprack]
    )

    #Perform two full tip pickup and drops, watch the overlap they drop in
    pipette.configure_nozzle_layout(style=ALL)
    pipette.pick_up_tip(tiprack1)
    pipette.drop_tip()
    pipette.pick_up_tip(tiprack2)
    pipette.drop_tip()

    #Perform two partial tip drops. Ensure they occupy the same tip drop alternation locations as the full drops.
    pipette.configure_nozzle_layout(style=COLUMN, start="A12")
    pipette.pick_up_tip(partial_tiprack.wells()[0])
    pipette.drop_tip()
    pipette.pick_up_tip(partial_tiprack.wells()[8])
    pipette.drop_tip()
```
This protocol should also be tested for an 8 channel pipette with two full pickups (no partial necessary at this time) to ensure that we have not caused the pipette to incorrectly center over A1 when moving to the moveable Trash.

# Risk assessment
Waste chutes were omitted from special casing on this version, so as not to interrupt the current addressable areas provided by the waste chute with cover fixture. This fixture supports single channel tip drop from a 96 channel, but if we centered the partially configured 96 channel on the waste chute we would not be able to make use of the slot. Right now, the 96ch adapter is taller than the waste chute. This means a 96ch adapter in D2 could potentially obstruct the 96ch if partially offset on the Waste Chute, however we would need to verify if such a collision was even possible.


api/src/opentrons/protocol_engine/state/motion.py
```diff
@@ -177,7 +177,10 @@ def get_movement_waypoints_to_addressable_area(
             destination = base_destination + Point(offset.x, offset.y, offset.z)
 
         # TODO(jbl 11-28-2023) This may need to change for partial tip configurations on a 96
-        destination_cp = CriticalPoint.XY_CENTER
+        if "movableTrash" in addressable_area_name:
+            destination_cp = None
+        else:
+            destination_cp = CriticalPoint.XY_CENTER
```
SyntaxColoring: **Fundamental question.** From your PR description:

> Rather than cover many addressable area types, we have chose to special case the moveable trash given the likelihood of an adjacent item obstruction causing an overlap in various partial tip configurations.

Can you elaborate on this?

The `moveToAddressableArea` command is [deliberately specified](https://github.com/Opentrons/opentrons/blob/10c68ecb88d96a3ad5b0431484fa3e0007bf0783/api/src/opentrons/protocol_engine/commands/move_to_addressable_area.py#L32-L34) to center all of its nozzles over the given addressable area—independent of whether the addressable area is a trash, or something else. We should aim for the implementation to match that specification. Are there any challenges with doing that?

In other words, I don't think there should be anything here that's conditional on the addressable area.

api/src/opentrons/protocol_engine/state/motion.py
```diff
@@ -177,7 +177,10 @@ def get_movement_waypoints_to_addressable_area(
             destination = base_destination + Point(offset.x, offset.y, offset.z)
 
         # TODO(jbl 11-28-2023) This may need to change for partial tip configurations on a 96
-        destination_cp = CriticalPoint.XY_CENTER
+        if "movableTrash" in addressable_area_name:
```
SyntaxColoring: In general, instead of matching addressable areas based on string prefixes, let's use semantic properties from the deck definition, like `areaType`. I'm aware that there are some places in the code that are already doing string matching, but those were a hack for the v7.1.0 crunch and we need to cut the habit.

api/src/opentrons/protocol_engine/state/motion.py
```diff
@@ -177,7 +177,10 @@ def get_movement_waypoints_to_addressable_area(
...
+            destination_cp = None
```
SyntaxColoring: Can you clarify what `destination_cp = None` is intended to do here and how it works?

Our critical point API confuses me, so I might have this wrong, but my understanding is that `None` uses the pipette's default critical point, which depends on what tips are attached. I don't know what the default will be here, but I suspect it doesn't reliably do the centering that we want, does it?

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -323,6 +323,22 @@ def get_addressable_area(self, addressable_area_name: str) -> AddressableArea:
         else:
             return self._get_addressable_area_from_deck_data(addressable_area_name)
 
+    def get_addressable_area_by_deck_slot_name(
+        self, slot_name: DeckSlotName
+    ) -> AddressableArea:
```
SyntaxColoring: This function doesn't make sense to me, conceptually? There are potentially multiple addressable areas "on" a single deck slot, right?

Like, a waste chute on D3 has a `1ChannelWasteChute`, `8ChannelWasteChute`, ....

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -324,10 +324,17 @@ def _check_deck_conflict_for_96_channel(
...
+    addressable_area = (
+        engine_state.addressable_areas.get_addressable_area_by_deck_slot_name(
+            labware_slot
+        )
+    )
+    if "movableTrash" not in addressable_area.area_name:
+        if primary_nozzle == "A12":
+            adjacent_slot_num = get_west_slot(destination_slot_num)
+        elif primary_nozzle == "A1":
+            adjacent_slot_num = get_east_slot(destination_slot_num)
```
SyntaxColoring: See my other note about treating all addressable areas the same, instead of special-casing trashes. If we do that, I think this whole diff can go away?

*Otherwise,* if we do still have this diff... 

1. See my other note about not matching based on string prefixes.
2. The comment describes the trash *or* the waste chute, but the code looks like it only handles `movableTrash`es?
3. The comment `# if we are handling commands in the trash or in the waste chute, skip these checks`. Can we edit this to explain *why* we're skipping those checks?

SyntaxColoring (dismissed):
Thank you! I have one fundamental question, and a number of smaller things that follow from that.

api/src/opentrons/protocol_engine/state/addressable_areas.py
CaseyBatten: Thanks for pointing that out, I have another PR out that introduces the ability to get a cutout by deckslot name, so I will switch it to using that function once #14415 is merged into edge. Then I will adjust the logic elsewhere to check using the data provided by the fixture instead of the addressable areas.

api/src/opentrons/protocol_engine/state/motion.py
CaseyBatten: The idea was to utilize the default critical point in trashing movements. My hope was that the default critical point would be the instruments center rather than the center of the partially configured channels, and the intention was to test this to confirm that. Unfortunately, setting destination_cp to None instead centered the entire pipette to A1. I'll need to modify this to identify the same critical point as a Full Configuration instead, and no longer set destination_cp to None.

api/src/opentrons/protocol_engine/state/motion.py
CaseyBatten: The current behavior is that the center of the chosen pipettes is the "critical point" that moves to the center of the addressable area. This means that when a pipette is in partial configuration, the critical point becomes the center of that selection. For example, the center of a partially configured 96 channel in COLUMN configuration for column A1 would be the middle of that column (so somewhere between D1 and E1?). That is my current understanding anyway. Whether or not we want to use the instrument center for ANY AND ALL move to addressable area commands is a difficult question to answer. 

In the case of movableTrash, we specifically want to use the center of the instrument, because we're in the trash! We don't want to overlap with anything outside of the bins walls.

In the case of the wasteChute with a cover attached, we would instead want the partially configured 96 channel with single column pickup to use only its active channel as the critical point, so that it can still utilize the cover slot (unless we want to remove that support). 

There may be some other examples as well, but these are the two examples related to this PR.

SyntaxColoring: > Whether or not we want to use the instrument center for ANY AND ALL move to addressable area commands is a difficult question to answer.

Okay, I think this is the crux of the confusion. *For the purposes of what this code needs to do,* [we've already answered](https://github.com/Opentrons/opentrons/blob/10c68ecb88d96a3ad5b0431484fa3e0007bf0783/api/src/opentrons/protocol_engine/commands/move_to_addressable_area.py#L32-L34) that question. The answer is "yes, use the instrument center always."

(For the broader purposes of future extensions to the deck configuration system, and generically supporting lots of fixtures and lots of pipettes, yes, it gets more difficult. But even there, I don't think the solution will be to special-case different addressable areas; it will be to make it configurable per-area, or add a parameter to `moveToAddressableArea`.)

> In the case of the wasteChute with a cover attached, we would instead want the partially configured 96 channel with single column pickup to use only its active channel as the critical point, so that it can still utilize the cover slot (unless we want to remove that support).

As far as I know, this was never supported. We can double-check this, but for example, if a Python protocol does `p96.drop_tip(waste_chute)`, I think it will emit a `moveToAddressableArea` command with `"addressableAreaName": "96ChannelWasteChute"`, not `"addressableAreaName": "8ChannelWastechute"`. That would be the proper use of the Protocol Engine API, as documented.

api/src/opentrons/protocol_engine/commands/move_to_addressable_area_for_drop_tip.py
```diff
@@ -64,6 +64,15 @@ class MoveToAddressableAreaForDropTipParams(PipetteIdMixin, MovementMixin):
             " If False, the tip will be dropped at the top center of the area."
         ),
     )
+    ignoreTipConfiguration: Optional[bool] = Field(
+        False,
+        description=(
+            "Whether to utilize the critical point of the tip configuraiton when moving to an addressable area."
+            " If True, this command will ignore the tip configuration and use the center of the entire instrument"
+            " of the entire instrument as the critical point for movement."
```
TamarZanzouri:
```suggestion
            "Whether to utilize the critical point of the tip configuraiton when moving to an addressable area."
            " If True, this command will ignore the tip configuration and use the center of the entire instrument"
            " as the critical point for movement."
```

TamarZanzouri: duplicate

api/src/opentrons/protocol_engine/commands/move_to_addressable_area_for_drop_tip.py
```diff
@@ -64,6 +64,15 @@ class MoveToAddressableAreaForDropTipParams(PipetteIdMixin, MovementMixin):
             " If False, the tip will be dropped at the top center of the area."
         ),
     )
+    ignoreTipConfiguration: Optional[bool] = Field(
```
SyntaxColoring, line 67: 1. I don't think we should introduce this field at this time. I don't see a pressing need for it, and it smells like unnecessary complexity.
2. If we do introduce this, *strictly,* I think it needs to default to `True` instead of `False`, to preserve prior semantics.

CaseyBatten:
Utilizing the new instrument critical point approach we achieved the desired result of the protocol outlined in the test plan above. When dropping tips in a trash bin, regardless of configuration, the pipette correctly distributed them within the bounds of the trashbin in the expected safe tip drop locations.

api/src/opentrons/protocol_engine/commands/move_to_addressable_area_for_drop_tip.py
CaseyBatten: We want this to default to False as the purpose of this flag is to enable and disable the ignoring of partial tip configuration. We want to utilize tip configuration unless otherwise specifically overloaded with a True. The usefulness of this is actually quite high as a field, for now it is only being used from move_to_addressable_area_to_drop_tip for the purposes of this PR, however eventually it could be utilized by any move to addressable area command regardless of intention. The idea being that in the future we could expand this to include actions such as blow outs and dispenses in the trash (or in any other relevant addressable area, depending on need).

SyntaxColoring: > We want this to default to False as the purpose of this flag is to enable and disable the ignoring of partial tip configuration. We want to utilize tip configuration unless otherwise specifically overloaded with a True.

For a couple of reasons, I don't think we actually do want that.

1. Centering with the active tips does not actually seem like a better choice for a general default. I mean, RSS-419 and RSS-434, the bugs we're solving here, are a case in point. To date, for every single addressable area that we move to, the shape of the pipette is a more important factor than the shape of the tips, because we need to avoid colliding with things or going out of bounds.
2. Even if it were a better choice, it's a breaking change from very explicitly documented semantics (as described in [the other thread](https://github.com/Opentrons/opentrons/pull/14413#discussion_r1478861579)). We can make breaking changes in extenuating circumstances (I did that [here](https://github.com/Opentrons/opentrons/pull/14253), for example), but they need to be justified, and I'm just not seeing the justification here.

> The usefulness of this is actually quite high as a field, for now it is only being used from move_to_addressable_area_to_drop_tip for the purposes of this PR, however eventually it could be utilized by any move to addressable area command regardless of intention. The idea being that in the future we could expand this to include actions such as blow outs and dispenses in the trash (or in any other relevant addressable area, depending on need).

Okay, so at the time of writing, no client has any active need for which centering the tips is the solution, right?

I appreciate evolving APIs to try to solve problems the right way, but all of this seems too vague right now for us to be confident that tip-centering *is* the right way. When I actually dig into concrete use cases, it seems less and less helpful.

For example, suppose you want to support a 96-channel dropping a column of tips into an `8ChannelWasteChute`, with the lid. You might think you can solve that with tip-centering. But then, how do you know whether the pipette is in bounds? How do you know it won't collide with anything? Is that Protocol Engine's responsibility, or Protocol Designer's? If it's Protocol Engine's, how do you surface errors early in the Protocol Designer UI? If it's Protocol Designer's, then Protocol Designer must know a lot about the geometry of the robot and pipettes and labware—so why is it using `moveToAddressableArea` at all, instead of `moveToCoordinates`?

CaseyBatten:
Resolution to noted issue by defaulting ignoreTipConfiguraiton behavior to true. Behavior change ensures that trash bins and waste chutes always operate with center of instrument. Tested on Flex, performed as expected. Of note, covered waste chutes are already incompatible with the 96 channel, so we are safe to proceed.

sanni-t (approved):
LGTM! Thanks for making the changes!

laviera (approved):

PR #13630, opened 2023-09-22
feat(shared-data): add custom & default thermocycler overlap offsets for labware

sanni-t:
# Overview

Closes RSS-353

- Adds the `stackingOverlapWithModule` for Biorad & NEST PCR plates on TC Gen2.
- Also adds a default overlap for any labware placed on a thermocycler on the OT2 such that the final height of the labware is the same as what it would have been prior to introducing stacking overlaps for thermocycler labware.

# Test Plan

The ticket above has a protocol attached for testing. That protocol places a nest PCR plate, a biorad plate and a custom pcr plate on the thermocycler one ofter the other to check for location precision. You will need to create and upload a custom PCR plate definition as mentioned in the protocol.

1. Upload this protocol to a Flex and check that:
    - [ ] when you run LPC, the pipette moves to accurate z-positions for the NEST and Biorad plates, and moves too high for the custom pcr plate
    - [ ] during the protocol run the gripper is able to move the NEST & biorad plates correctly but can't move the custom pcr plate because the gripper moves too high

2. Adjust the protocol for running on OT2 and upload to an OT2:
    - [ ] when you run LPC, the pipette moves to accurate z-positions for the NEST and Biorad plates, and an okay position for custom pcr plate (unless it happens to be based on an armadillo plate). Also note that positional accuracy of the OT2 for labware on modules is not as high as the Flex so I wouldn't stress too much if there is an offset of a couple of mm.

# Changelog

- updated Biorad & NEST PCR plate definitions
- added backwards-compatible stacking overlap for OT2 based TC in `state.labware. get_module_overlap_offsets()`

# Review requests

- Do above tests on both Flex & OT-2
- Do we need to bump labware definition revision number for this?

# Risk assessment

Low. Just adds the missing overlap values. Low risk if tests show expected results.

PR #14414, opened 2024-02-02
fix(api): use old TC Gen2 labwareOffset values in legacy core

sanni-t:
Fixes issue in RESC-196

# Overview

[In this PR](https://github.com/Opentrons/opentrons/pull/13567), we changed how we calculate the final origin of any labware placed on a TC gen2. Previously we assumed that all labware sit 98.26mm above the OT2 slot; we call it the TC labware offset. This offset was correct for the Armadillo plate; while other pcr plates would sit +/-0.1mm differently. In 7.0.1, we updated this labware offset to be the top plane of metal cylinders of the thermocycler (108.96mm), and introduced a `stackingOffsetWithModule` parameter in the labware definitions to inform us of the distance a labware actually rests at from top of those cylinders. For labware that didn’t have these offsets, we [added a fallback](https://github.com/Opentrons/opentrons/pull/13630) of using the previous labware offset of 98.26.

Issue is, we forgot to add that fallback in v2.13, which is a legacy core based API so now the backend assumes all labware sit at 108.96mm on TC Gen2, ignoring any `stackingOffsetWithModule` because the legacy core has no concept of stacking offsets.

[I compiled all this, with options for solutions in this spreadsheet](https://docs.google.com/spreadsheets/d/1WrQNtfWEtFBqY7iKXv0lUZ_sutG1KYjYZb5C6k5Tiac/edit?usp=sharing) and we decided to go with option 1- revert the behavior of API v2.13 to what it was prior to the change in 7.0.1.

# Test Plan

- Load a TC Gen2 protocol on the OT2 without loading any previous LPC offsets and verify that pipettes go to the correct z height for the labware on the thermocycler, rather than moving 1cm above the labware.
- Also verify that TC gen1 nor any other modules are affected.

# Changelog

- when loading module labware offsets in module geometry, it will now load separate, older offset values for TC gen2
- other minor code refactoring

# Review requests

- Is it enough to save the older offsets in shared-data like that or do we want to have a better system for it?

# Risk assessment

Medium. This adds/ reverts a change in positioning such that previous LPC offsets should not used or they'll result in pipettes crashing into labware bottom. We should also convey this to users.

### DIFF
--- a/api/src/opentrons/protocol_api/core/legacy/module_geometry.py
+++ b/api/src/opentrons/protocol_api/core/legacy/module_geometry.py
@@ -16,6 +16,7 @@ import numpy as np
 
 from opentrons_shared_data import module
 from opentrons_shared_data.module.dev_types import ModuleDefinitionV3
+from opentrons_shared_data.module import OLD_TC_GEN2_LABWARE_OFFSET
 
 from opentrons.types import Location, Point, LocationLabware
 from opentrons.motion_planning.adjacent_slots_getters import (
@@ -29,6 +30,7 @@ from opentrons.hardware_control.modules.types import (
     ModuleModel,
     ModuleType,
     module_model_from_string,
+    ThermocyclerModuleModel,
 )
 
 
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Changes and cocnept LGTM

PR #14415, opened 2024-02-02
fix(api): Tip pickup call validation near wastechute

CaseyBatten:
**Special Note:** This is a replacement for #14371 due to a potentially problematic keyerror bug that would occur on any height checking that resulted in a fixture being the highest item in the adjacent slots. This seeks to add back the capability of #14371 as well as fix that bug.

# Overview
This seeks to fix issues related to RSS-375 and RSS-422.

Of note, this will only fix issues during runtime. Currently when in partial tip configuration the engine will attempt to check the height of object adjacent to the slot that an action (such as pickup tip) is occurring in. However, only the heigh of labware and module is checked, the system does not check for the height of fixtures loaded into the deck configuration.

To solve this, a method has been added to allow for searching of Cutout fixtures by deck slot name, and then the height of a returned fixture is then checked alongside the labware and module heights. If, for example a waste chute is loaded in D3 and a partial tip pickup is attempted in D2, the engine will raise an error.

# Test Plan
Ensure that a protocol with a waste chute loaded in D3  with a 96 channel in partial tip configuration set to pick up tips from column A12 of a tiprack in D2 using column A1 of the instrument raises an error during operation before a collision would occur.
The following protocol ought to produce the expected results:

```python

requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(protocol_context):
	tip_rack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "D2")
	instrument = protocol_context.load_instrument('flex_96channel_1000', mount="left", tip_racks=[tip_rack1])

	waste_chute = protocol_context.load_waste_chute()

	instrument.pick_up_tip(tip_rack1)
	instrument.drop_tip(waste_chute)
```

# Risk assessment
While the robot will prevent a collision during runtime, a protocol where a crash would occur would still pass analysis. This is in part due to the fact that we do not load a true deck configuration for analysis, so when checking the set of loaded fixtures we see that there are "None", even if that is untrue. This is yet another item that would become simpler with multipass analysis.


CaseyBatten:
Tested the protocol in the Testing Plan on Flex, resulted correctly with the following expected error:
**Error 2004: PartialTipMovementNotAllowedError**
```
Moving to Opentrons Flex 96 Tip Rack 50 µL in slot D2 with pipette column A1 nozzle configuration will result in collision with items in deck slot D3.
```

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -462,6 +470,30 @@ def get_addressable_area_center(self, addressable_area_name: str) -> Point:
             z=position.z,
         )
 
+    def get_fixture_by_deck_slot_name(
+        self, slot_name: DeckSlotName
+    ) -> Optional[CutoutFixture]:
+        """Get the Cutout Fixture of a fixture currently loaded into a specific Deck Slot."""
```
SyntaxColoring: Terminology thing: 
```suggestion
        """Get the Cutout Fixture currently loaded where a specific Deck Slot would be."""
```

* Fixtures are not loaded into deck slots; deck slots are a kind of fixture, and fixtures are loaded into cutouts.
* "Cutout Fixture of a fixture" seems like it was just a typo?

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -462,6 +470,30 @@ def get_addressable_area_center(self, addressable_area_name: str) -> Point:
...
+                            cutout_fixture_id, self.state.deck_definition
+                        )
+                    )
+                    break
+            if slot_cutout_fixture is None:
+                raise CutoutDoesNotExistError(
+                    f"No Cutout was found in the Deck that matched provided slot {slot_name}."
+                )
+            return slot_cutout_fixture
+        return None
```
SyntaxColoring: Your call, but I think you can simplify this control flow a little bit by `return`ing where you're currently `break`ing. Then you don't have to keep track of the `None` value for `slot_cutout_fixture`.

```suggestion
        if deck_config:
            slot_cutout_id = DECK_SLOT_TO_CUTOUT_MAP[slot_name]
            # This will only ever be one under current assumptions
            for cutout_id, cutout_fixture_id in deck_config:
                if cutout_id == slot_cutout_id:
                    slot_cutout_fixture = (
                        deck_configuration_provider.get_cutout_fixture(
                            cutout_fixture_id, self.state.deck_definition
                        )
                    )
                    return slot_cutout_fixture
            raise CutoutDoesNotExistError(
                f"No Cutout was found in the Deck that matched provided slot {slot_name}."
            )
        return None
```

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -462,6 +470,30 @@ def get_addressable_area_center(self, addressable_area_name: str) -> Point:
...
+                    slot_cutout_fixture = (
+                        deck_configuration_provider.get_cutout_fixture(
+                            cutout_fixture_id, self.state.deck_definition
+                        )
+                    )
+                    break
+            if slot_cutout_fixture is None:
+                raise CutoutDoesNotExistError(
+                    f"No Cutout was found in the Deck that matched provided slot {slot_name}."
+                )
```
SyntaxColoring, line 496: Would this comment be correct?

```suggestion
                # If this happens, it's a bug. Either DECK_SLOT_TO_CUTOUT_MAP
                # is missing an entry for the slot, or the deck configuration is missing
                # an entry for the cutout.
                raise CutoutDoesNotExistError(
                    f"No Cutout was found in the Deck that matched provided slot {slot_name}."
                )
```

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -687,21 +690,33 @@ def get_extra_waypoints(
...
         """Get the item present in a deck slot, if any."""
         maybe_labware = self._labware.get_by_slot(
             slot_name=slot_name,
         )
 
         if isinstance(slot_name, DeckSlotName):
+            maybe_fixture = self._addressable_areas.get_fixture_by_deck_slot_name(
+                slot_name
+            )
+            # Ignore generic single slot fixtures
```
SyntaxColoring, line 705: Could you clarify in this comment why we're ignoring these fixtures specifically? What would happen if we didn't?

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -687,21 +690,33 @@ def get_extra_waypoints(
 
     def get_slot_item(
         self, slot_name: Union[DeckSlotName, StagingSlotName]
-    ) -> Union[LoadedLabware, LoadedModule, None]:
+    ) -> Union[LoadedLabware, LoadedModule, CutoutFixture, None]:
```
SyntaxColoring, line 695: I think we should be pretty wary of this interface change.

This is saying that a fixture is a kind of "slot item," something that's in a slot. But the first principles of deck configuration are closer to the other way around: a slot is a special kind of fixture, which is in a cutout.

This is also why you had to add a `DECK_SLOT_TO_CUTOUT_MAP` constant. The new code isn't quite operating on the official affordances of our deck configuration system.

It looks like we're attempting things this way because there's [tip pickup validation code](https://github.com/Opentrons/opentrons/blob/4f98c75e153fb60d144823e2233bbc407300250d/api/src/opentrons/protocol_api/core/engine/deck_conflict.py#L353) that finds "the slot adjacent to the target," and then gets the height of whatever is inside it. It's tempting to do it this way so we can leave that validation code unchanged.

But ideally, that validation code really does need to understand that it's insufficient to check adjacent slots. It needs to check *all adjacent obstacles,* including those that are not in slots, or those that occupy more than one slot. We should be wary of introducing conceptual confusion over here just for the sake of preserving code elsewhere that was already questionable.

SyntaxColoring: The proper solution for this is to check for collisions based on bounding boxes.

I know, I know. But this really doesn't have to be as big of a deal as one might fear. Actually, this tip pickup validation code is an excellent opportunity to do it. Here's how it would work:

1. Currently, `shared-data` defines fixtures' z-dimensions. We'd expand it to also define x-dimensions and y-dimensions.
2. The tip pickup code is already computing waypoints to the tip rack. Before actually traveling through those waypoints, we'd check where the last one is, and use it to build a bounding box of the pipette body.
3. We'd check if the pipette body intersects with the fixture.

SyntaxColoring: If we really want to keep this as-is for the sake of squeezing it into v7.2.0, let's add a tech-debt ticket and a `# todo` comment. And I think we should resolve it before we pick up more feature work. We wouldn't want to let this sit any longer than, say, a month, given the fact that a bounding-box solution seems so readily attainable.

SyntaxColoring (approved):
This looks good to merge if it's been tested.

I have a few code-level comments, plus one bigger comment about how we're approaching this validation, architecturally.

api/src/opentrons/protocol_engine/state/addressable_areas.py
CaseyBatten: Just a nomenclature typo, thanks for the catch.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -462,6 +470,30 @@ def get_addressable_area_center(self, addressable_area_name: str) -> Point:
...
 
+    def get_fixture_by_deck_slot_name(
+        self, slot_name: DeckSlotName
+    ) -> Optional[CutoutFixture]:
+        """Get the Cutout Fixture of a fixture currently loaded into a specific Deck Slot."""
+        deck_config = self.state.deck_configuration
+        if deck_config:
+            slot_cutout_id = DECK_SLOT_TO_CUTOUT_MAP[slot_name]
+            slot_cutout_fixture = None
+            # This will only ever be one under current assumptions
```
sanni-t, line 481: Not sure what this means. Can you clarify?

api/src/opentrons/protocol_engine/state/geometry.py
CaseyBatten: The deck configuration will always contain some kind of fixture for each cutout, so if we didn't ignore generic single slots then "get_slot_item" would always at least return a single slot fixture of some kind as a bare minimum. This is assuming no other fixture was installed there and there were no labware or modules there either. While technically correct, other functions may depend upon an "empty" slot being "None", as in the case of the following:
```
def run(context):
    assert context.deck['A1'] is None
    del context.deck['A1'] # no op
```

This currently should pass analysis. If we were to return empty generic single slots instead of "None" it would fail. The idea currently is that the "item" in an empty single slot fixture is "None". There is "No Item" in the "Slot". Conversely, a waste chute WOULD be "an item" in the "slot". If there were a labware or module or fixture-that-is-not-a-generic-slot loaded into that "Slot", then we'd care about returning it.

Perhaps my approach here is wrong, but it seems appropriate to me in that entities calling the get_slot_item() function would want to know about that item, be it a labware, a module or a fixture. The only thing that they reasonably would not care about is an empty generic fixture, as the previous version of this function assumed that an empty slot was "None" items.

api/src/opentrons/protocol_engine/state/addressable_areas.py
```diff
@@ -462,6 +470,30 @@ def get_addressable_area_center(self, addressable_area_name: str) -> Point:
...
+                if cutout_id == slot_cutout_id:
+                    slot_cutout_fixture = (
+                        deck_configuration_provider.get_cutout_fixture(
+                            cutout_fixture_id, self.state.deck_definition
+                        )
+                    )
+                    break
+            if slot_cutout_fixture is None:
+                raise CutoutDoesNotExistError(
+                    f"No Cutout was found in the Deck that matched provided slot {slot_name}."
```
sanni-t, line 495:
```suggestion
                    f"No Cutout matching the provided slot ({slot_name}) was found in the deck configuration."
```

api/src/opentrons/protocol_engine/state/geometry.py
CaseyBatten: The decision to do this as part of the existing validation was one made with the informed notion that as development proceeds on conflict checking that we will transition away from it in the future as the bounding box capability comes into scope. One such case is exactly what you mentioned, that we should instead be checking the geometric overlaps rather than the adjacent slots. While that is superior, and would ideally be what we are developing, the solution here is meant to prevent dangerous behavior that our system is capable of now with the current release. In the future, these checks will ideally become unnecessary as a supervisory body that is responsible for identifying bounding box conflicts will instead be responsible for handling the case covered by this PR (among others across the deck and stack). A `#todo` to track this item would be appropriate.

api/src/opentrons/protocol_engine/state/addressable_areas.py
CaseyBatten: Thats accurate. This raise case is here to protect against the currently impossible. There is no way with our current concept of deck configuration to have a deck slot without a cutout fixture, nor to provide a deckslot here that isn't included within the currently provided selection of known deckslots to cutouts. However, on the off chance that information like that was provided we will catch it here. Hypothetically, if we decide someday you can load a deck with only some of its fixtures and the rest left literally empty (as in a big hole in the deck), this would catch instances where entities calling this function attempt to reference the "holes" in the deck.

api/src/opentrons/protocol_engine/state/geometry.py
sanni-t: > * Currently, `shared-data` defines fixtures' z-dimensions. We'd expand it to also define x-dimensions and y-dimensions.
> * The tip pickup code is already computing waypoints to the tip rack. Before actually traveling through those waypoints, we'd check where the last one is, and use it to build a bounding box of the pipette body.
> * We'd check if the pipette body intersects with the fixture.

Oho! Glad you mention this approach! I'm trying to get there with https://github.com/Opentrons/opentrons/pull/14426

We can keep the approach in this PR for 7.2.0 and make changes in the follow-up

sanni-t (commented):
Sorry for getting this in late. Added comments

PR #14417, opened 2024-02-02
feat(app, app-shell, robot-server): Add notification support for GET /runs and /runs/:runId

mjhuff:
Closes RAUT-951, RSS-452, RQA-2278, RQA-2280

# Overview

This PR adds refetch flag notification support for the /runs and /runs/:runId endpoints. Because a growing number of routes are supported, the notification logic on the robot-server side is refactored to better scale and reduce notification_client complexity/responsibility. 

On the app side, there are a few changes as well. Logs are now considerably less spammy. Additionally, there are a few logic problems that need fixing. First, over-subscribing/unsubscribing isn't properly managed. Although over-subscribing isn't an issue, we don't want to unsubscribe from a topic until ALL interested components are no longer interested in the topic (this wasn't causing bugs yet, but it would after the addition of these routes). Second, if a subscription to a topic occurs before the client has fully connected, the subscribe action never occurs (this is causing bugs). We need retry logic to solve this. Lastly, useNotifyService needs to support dynamic topics, such as /runs/:runId. The old logic assumes that a component must be freshly mounted in order to express interest in a new topic, but this isn't true in practice (nor should it be true). Simply adding topic as a useEffect dependency and cleaning up the way we prevent unnecessary subscriptions solves this problem. 

# Test Plan

- [x] Smoke test running a protocol, pausing, cancelling, on the desktop app/ODD. These actions should be much snappier.
- [x] Look at app-shell/app-shell-odd logs and/or mosquitto logs to confirm that subscribing/unsubscribing is occurring appropriately.
- [x] Ensure that app-shell logs in general are considerably less spammy.
- [x] Ensure that run commands still display, and starting a maintenance run flow is just as snappy as it is in alpha-4.

# Changelog
- Add MQTT support for polling protocol endpoints.

# Risk assessment
High. This is the heaviest of all the polling migrations. I've smoke tested considerably to make sure things work, but this touches protocol flow.


robot-server/robot_server/runs/run_store.py
mjhuff, line 1: I'm doing synchronous publishing here. There's negligible time difference b/t sync/async. I thought that may be simpler and less risky than doing async here, but I'm happy to refactor it over if someone has a case for it.

robot-server/robot_server/maintenance_runs/maintenance_run_data_manager.py
```diff
@@ -154,9 +152,7 @@ async def delete(self, run_id: str) -> None:
         if run_id == self._engine_store.current_run_id:
             await self._engine_store.clear()
 
-            await self._notification_client.publish(
-                topic="robot-server/maintenance_runs",
-            )
+            await self._maintenance_runs_publisher.publish_current_maintenance_run()
```
CaseyBatten, line 155: A question on what this was doing before and what it is doing now. This is publishing an equivalent to `/maintenance_runs/current_run`, where as previous the publisher was simply `/maintenance_runs` (which only has an equivalent POST if a /{run_id} is not supplied for a GET). So I assume that publishing the `/current_run` endpoint was the intended behavior of the original publish statement that has been replaced here?

CaseyBatten (commented):
This looks like a good implementation on the robot-server side of things, just looking for some clarification on some of the maintenance run behavior.

robot-server/robot_server/runs/run_store.py
sfoster1: When you say negligible time difference are you talking about wall clock from call-site to app receiving, or wall clock duration of the call? If it's the first, we need to think about the second too.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -0,0 +1,102 @@
...
+        get_current_command: Callable[[str], Optional[CurrentCommand]],
+        run_id: str,
+    ) -> None:
+        """Asynchronously publish new current commands.
+
+        Args:
+            get_current_command: Retrieves the engine store's current command.
+            run_id: ID of the current run.
+        """
+        while not self._run_data_manager_polling.is_set():
```
sfoster1, line 70: Let's drop some todos in here to extend reactivity farther downward. We can definitely get this behavior done by triggers on run data manager methods.

robot-server/robot_server/maintenance_runs/maintenance_run_data_manager.py
mjhuff: Good question - so the functional behavior is identical to the previous PR. My intention with the refactor here is to move all the route handling logic outside of everything but the Publisher classes (I don't think it's great to define the topic strings directly in the data manager). 

As for renaming, I changed the topic string to mirror the naming of the HTTP resource - on the app side, this notification tells the app to refetch `GET /maintenance_runs/current_run`. Calling it `/maintenance_runs` was just a subpar naming decision on my part.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
mjhuff: > When you say negligible time difference are you talking about wall clock from call-site to app receiving, or wall clock duration of the call? If it's the first, we need to think about the second too.

Wall clock duration of the call is what I was interested in, although call-site to app is also negligible when tested.

CaseyBatten (approved):
Let's do it!

PR #11912, opened 2022-12-20
Unpin Python requirements

ecederstrand:
# Overview
Fixes #11905

# Changelog
- Allow dependency versions up to the next major version when pinning is not strictly necessary. This makes it easier to combine opentrons with other Python packages that have shared dependencies.

PR #14420, opened 2024-02-02
chore: update python deps to breaking change breakpoints

sfoster1:
This is a PRs to update python dependencies to modern versions. This one stops at the major breaking change point of pydantic 2.0. While Pydantic 2.0 is apparently much much faster (it's written in rust if you can believe it) it also makes a bunch of breaking api changes and _in theory_ provides back-compat patches. Similarly, fastapi 0.100.0 switches internally to depend on pydantic 2, with similarly theoretical back compat patches. So, we've updated to, 

## Dependency Version Changes
In prod, major deps (full list is essentially entirely in the robot-server pipfile, see below for why):
- fastapi 0.99.0 (from 0.68.1)
- pydantic 1.10.12 (from 1.9.2)
- uvicorn 0.27.0 (from 0.14.0)
- sqlalchemy 1.4.51 (from 1.4.32 - bigger change than you think, and breaking changes above this)

In dev tooling,
- mypy 1.8.0 (from 0.981)
- flake8 7 (from like 3 or something we never updated this)
- mock 5 (from 4)
- decoy 2 (from 1)
- pytest 7.4.4 except robot-server, which is limited by tavern to 7.3 (from various)
- tavern 2.9.1 (from 1.6)

## Version Definition Changes
Some people want to use our python libraries that are published on pypi. The problem with this is that our libraries - `opentrons` and `opentrons_shared_data` have explicit version pinning in their setup.py `install_requires`, which makes it incredibly hard for them to coexist with other python packages that aren't comaintained by us (see #11912 , #12839 ). One way to fix this would be version ranges in the setup.py and explicit versions (that are contained within those ranges, and that match or define what's present on the robot) in the pipfiles. We couldn't do this because pipenv had problems with it.

Now, however, we've upgraded pipenv, and that strategy works! And since I was going around bumping all the deps anyway, I could figure out what the actual functional dependency version boundaries were. So as part of this, `opentrons` (`api/setup.py`) and `opentrons_shared_data` (`shared-data/python/setup.py`) now have version ranges for all of their `install_requires` that aren't other opentrons packages, and I'm pretty sure about those version ranges. They may be smaller than would be ideal, but they're real.

## Testing (and how to get this out of draft)
This PR should stay in draft until the system builders are known to have the right dependency versions. 
- [x] Flex build with appropriate deps
- [x] OT-2 build with appropriate deps
- [x] usb-bridge works on flex
- [x] robot-server works on flex
- [x] robot-server works on ot-2
- [x] update-server works on flex
- [x] update-server works on ot-2
- [x] jupyter works on ot-2
- [x] jupyter works on flex


sfoster1:
Works on flex with https://github.com/Opentrons/oe-core/pull/132 (build: https://builds.opentrons.com/ot3-oe/7790572076/ot3-system.zip )

sfoster1:
Works on ot-2 with https://github.com/Opentrons/buildroot/pull/220 (build: https://builds.opentrons.com/ot2-br/7807327676/ot2-system.zip )

CaseyBatten (commented):
Works well on Flex during run and no issues with analysis on app or robot. FastAPI changelog might want to be appraised to be sure that there isn't anything that might conflict with how we've used it so far.

vegano1 (approved):
Sweet,
once we start using this more then that might expose specific issues.
until then, this looks good to me!

api/src/opentrons/hardware_control/execution_manager.py
```diff
@@ -74,7 +74,9 @@ async def wait_for_is_running(self) -> None:
         async with self._condition:
             if self._state == ExecutionState.PAUSED:
                 await self._condition.wait()
-                if self._state == ExecutionState.CANCELLED:
+                # type-ignore needed because this is a reentrant function and narrowing cannot
+                # apply
+                if self._state == ExecutionState.CANCELLED:  # type: ignore[comparison-overlap]
```
SyntaxColoring, line 79: Oof, this seems like bad mypy behavior.

api/src/opentrons/protocols/models/json_protocol.py
```diff
@@ -672,7 +672,7 @@ class Model(BaseModel):
     modules: Optional[Dict[str, Modules]] = Field(
         None, description="All modules used in this protocol"
     )
-    commands: List[AllCommands] = Field(
+    commands: Optional[List[AllCommands]] = Field(
```
SyntaxColoring: The underlying JSON schema says this is required. I'm looking into whether it's safe to change the `None` default to `...`.

PR #14425, opened 2024-02-05
feat(app): detect an A1 deck config conflict for thermcycler

brenthagen:
# Overview

provides special-case logic to generate a conflict if a protocol uses a therocycler and the robot has a trash bin configured in A1. special-casing is needed because A1 is not directly referenced in the protocol. in a future where modules are fixtures, we won't need this.

Updates the location conflict modal for the thermocycler to reference both slots, and update both slots.

closes RQA-2029

Screen Shot 2024-02-05 at 1 44 40 PM
Screen Shot 2024-02-05 at 3 18 16 PM

# Test Plan

 - verified that a conflict is generated when protocol requires a thermocycler and trash bin is configured in A1. verified that conflict is resolved by location conflict modal

# Changelog

 - Detects an A1 deck config conflict for thermcycler

# Review requests

load a thermocycler protocol, add a trash bin in A1 and look for conflict

# Risk assessment

low


mjhuff (approved):
Tested and works as expected. LGTM.

PR #14426, opened 2024-02-05
feat(api): check all possible collisions with deck items during partial tip movement

sanni-t:
Closes RSS-437

# Overview

Previously we only checked for possible collisions with slots to the East and West of the target labware, now we check all cardinal directions around the slot for possible deck conflicts. These new checks are based on knowing the exact position of the pipette and the deck area covered by the pipette, _regardless of the pipette's configuration_. 

NOTE: Checks for collisions with trash bin & waste chute are still not available in analysis. They work only during actual runs.

# Test Plan
The following protocol is expected to raise the following error in various configurations (uncomment and recomment different load calls and configuration types to test this):
**Expected Error:**
```
PartialTipMovementNotAllowedError [line 37]: Error 2004 MOTION_PLANNING_FAILURE (PartialTipMovementNotAllowedError): Moving to Bio-Rad 96 Well Plate 200 µL PCR in slot C2 with pipette column A1 nozzle configuration will result in collision with items in deck slot D3.
```

**Protocol:**
```

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16",
}

def run(protocol_context: protocol_api.ProtocolContext):
    trashA3 = protocol_context.load_trash_bin("A3")
    tiprack_safe = protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "A1")
    #tiprack1= protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C1")
    #tiprack2= protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "D1")
    #tiprack3= protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "D2")
    tiprack4= protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "D3")
    #tiprack5= protocol_context.load_labware("opentrons_flex_96_tiprack_50ul", "C3")
    plate = protocol_context.load_labware('biorad_96_wellplate_200ul_pcr', "C2")

        "flex_96channel_1000", mount="left", tip_racks=[tiprack_safe]
    )

    #pipette.configure_nozzle_layout(style=COLUMN, start="A12")
    #pipette.pick_up_tip(tiprack_safe.wells()[0])

    pipette.configure_nozzle_layout(style=COLUMN, start="A1")
    pipette.pick_up_tip(tiprack_safe.wells()[88])


    #Both should raise an error
    #West/SouthWest/South
    # pipette.aspirate(50, plate.wells()[4])
    # pipette.drop_tip()

    #East/SouthEast/South
    pipette.aspirate(50, plate.wells()[92])
    pipette.drop_tip()
```
# Changelog

- added pipette nozzle bounds getter to pipette store
- updated the conflict check to figure out slots overlapping with pipette position (instead of assuming east/west slot overlap based on just nozzle configuration), and checking if any of these slots have items that are tall enough to cause a collision with the pipette.
- misc supporting changes like test & fixtures updates

# Review requests
- Test on a robot for various combinations of pipette, deck items and deck configurations. It should raise expected errors during both analysis and runs
- *currently the bounding box only comprises of nozzles.* I think we should add a margin for it in x, y directions for the 96 channel
- I had to add `do_compatibility_check` bool to the addressable area getter methods since even fetching info about an area that has a conflict with a loaded addressable raises error. I think we should not raise that error unless we're trying to actually load an addressable area in an already configured area. But for now, I've kept the existing addressableArea behavior as is while only suppressing the the check in the getter methods being used in deck_conflict.
- I have some in-line comments & questions that will appreciate info/opinions if you have.

# Risk assessment
Medium


CaseyBatten (commented):
Performed functionality test on app and on Flex, correctly raises errors in the case of tipracks surrounding labware to the East, West, Southeast, Southwest and South of the targeted labware. Functionally looks great!

Odd raise case identified:
If the pipette is configured in A1 column configuration, and is targeting well D8 of a labware loaded into C3, with a trashbin loaded into D3, the following unexpected error is raised:
`IncompatibleAddressableAreaError [line 36]: Error 4000 GENERAL_ERROR (IncompatibleAddressableAreaError): Cannot use Slot D3, not compatible with one or more of the following fixtures: Trash Bin in D3`

This issue is non-blocking for this PR, but good to keep an eye on. Relates to a conflict in hierarchy of the AddressableArea checks vs deck conflict checks.

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
```diff
@@ -202,4 +208,11 @@ def get_pipette_static_config(pipette_dict: PipetteDict) -> LoadedStaticPipetteD
         # https://opentrons.atlassian.net/browse/RCORE-655
         home_position=0,
         nozzle_offset_z=0,
+        # TODO (spp): Confirm that the nozzle map is populated by the hardware api by default
```
sanni-t, line 211: This seems to be correct

CaseyBatten (approved):
Issue with fixtures in adjacent slots has been corrected, functionally raises errors in app, on protocol setup, and during protocol execution on Flex as expected.

api/src/opentrons/hardware_control/dev_types.py
```diff
@@ -95,7 +95,7 @@ class PipetteDict(InstrumentDict):
     has_tip: bool
     default_push_out_volume: Optional[float]
     supported_tips: Dict[PipetteTipType, SupportedTipsDefinition]
-    current_nozzle_map: Optional[NozzleMap]
+    current_nozzle_map: NozzleMap  # spp: why was this Optional?
```
SyntaxColoring, line 98: Is this comment left over from debugging, or is this still something we need to investigate?

SyntaxColoring (commented):
Nice.

api/src/opentrons/hardware_control/dev_types.py
sanni-t: Kinda both. I did a primary check and it seemed like it doesn't need to be optional since it gets populated at the time of pipette instance creation- for both virtual and actually connected pipettes. But we should do a thorough check too, and also update the methods that assume that this field is optional (there's a few of those in hardware control & engine). So kept the comment in to not forget.

PR #14427, opened 2024-02-05
feat(app): refactor analytics settings modal

jerader:
closes RAUT-946

# Overview

refactor analytics modal to match the new designs, also only allowing the user to accept.

Screen Shot 2024-02-05 at 4 40 40 PM

# Test Plan

see that the modal works and appears as expected. The easiest way to test this is to change your privacy setting on the tab to opt out, then the modal should pop up immediately!

# Changelog

- update the modal to match designs
- delete unneeded component

# Review requests

see test plan

# Risk assessment

low


app/src/organisms/AnalyticsSettingsModal/index.tsx
```diff
@@ -1,38 +1,90 @@
...
+      <Flex flexDirection={DIRECTION_COLUMN}>
+        <Flex gridGap={SPACING.spacing10} flexDirection={DIRECTION_COLUMN}>
+          <Flex gridGap={SPACING.spacing10} flexDirection={DIRECTION_COLUMN}>
+            <Trans
+              t={t}
+              i18nKey={t('privacy_body')}
+              components={{ block: <StyledText as="p" /> }}
+            />
+          </Flex>
+          <Link
```
brenthagen: there's an existing `ExternalLink` component with an icon, can that be used?

jerader: thanks Brent, I forgot about that component. It seems to do the trick.

mjhuff (commented):
Modal looks good! Looking over the ticket, do we actually want to show this modal to users who have previously opted out? From what I can tell, it seems that what we want is to only show this to _new_ users who haven't yet opted in/out. If someone has opted out previously, they don't get this modal.

jerader:
> Modal looks good! Looking over the ticket, do we actually want to show this modal to users who have previously opted out? From what I can tell, it seems that what we want is to only show this to _new_ users who haven't yet opted in/out. If someone has opted out previously, they don't get this modal.

@mjhuff , oh good call out! I'll post in slack to confirm.

app/src/organisms/AnalyticsSettingsModal/index.tsx
```diff
@@ -1,38 +1,70 @@
...
   const dispatch = useDispatch<Dispatch>()
   const seen = useSelector(getAnalyticsOptInSeen)
-  const setSeen = (): unknown => dispatch(setAnalyticsOptInSeen())
+
+  const handleClick = (): void => {
+    dispatch(setAnalyticsOptInSeen())
+    dispatch(toggleAnalyticsOptedIn())
+  }
 
   return !seen ? (
```
jerader: @mjhuff , the `seen` boolean here is from the `getAnalyticsOptInSeen` selector, so i think this modal does only pop up if the user has never seen it before, aka a new user. That is what we think the expected behavior is right?

mjhuff: Yep, looking at migrate.ts, this seems to be correct!

mjhuff (approved):
Per slack convo, we want to show this modal to new users and users who have previously opted out. LGTM!

PR #14428, opened 2024-02-06
fix(robot-server): Fix indefinite protocol cancel state

mjhuff:
Closes [RQA-2286](https://opentrons.atlassian.net/browse/RQA-2286)

# Overview

This PR fixes a bug in which cancelling a protocol from the desktop app/ODD results in an indefinite stop-requested state, which does not resolve unless the user navigates away and back to the protocol from the desktop app. Certain EngineStatus states are internal to PE and not handled by the robot-server, including stop-requested. This means that when the app requests a STOP, robot-server doesn't internally manage the transition from stop-requested->stopped. Simply emitting refetch flags whenever the robot-server updates the state is therefore insufficient. The current solution is to do what we do for current_commands, which is to poll PE and emit to the app if there's an update. In the future, we should bubble up this event to robot-server from PE (the TODO is already there).

There's a cheeky removal of the RUN_STATUS_STOP_REQUESTED as a run status that currently disables refetching on the app side. After spending some time looking at it with @shlokamin, we were unable to determine why this conditional is here, or why things appear to work even when it is here. Let's just get rid of it. 

# Test Plan
- Initiate a protocol run.
- Cancel out of said protocol run.
- The cancel modal should not be indefinite now.

# Changelog
- Fixed indefinite "cancelling run" desktop app/ODD state when pressing "cancel" during a protocol run.

# Risk assessment
low. This PR just adds another notification push. Worst case scenario, the app refetches more often than it should.

[RQA-2286]: https://opentrons.atlassian.net/browse/RQA-2286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/RunTimeControl/hooks.ts
+++ b/app/src/organisms/RunTimeControl/hooks.ts
@@ -77,11 +77,9 @@ export function useRunStatus(
     refetchInterval: DEFAULT_STATUS_REFETCH_INTERVAL,
     enabled:
       lastRunStatus.current == null ||
-      !([
-        RUN_STATUS_STOP_REQUESTED,
-        RUN_STATUS_FAILED,
-        RUN_STATUS_SUCCEEDED,
-      ] as RunStatus[]).includes(lastRunStatus.current),
+      !([RUN_STATUS_FAILED, RUN_STATUS_SUCCEEDED] as RunStatus[]).includes(
+        lastRunStatus.current
+      ),
     onSuccess: data => (lastRunStatus.current = data?.data?.status ?? null),
     ...options,
   })
--- a/robot-server/robot_server/service/notifications/publishers/runs_publisher.py
+++ b/robot-server/robot_server/service/notifications/publishers/runs_publisher.py
@@ -2,9 +2,7 @@ from fastapi import Depends
 import asyncio
 from typing import Union, Callable, Optional
 
-from opentrons.protocol_engine import (
-    CurrentCommand,
-)
+from opentrons.protocol_engine import CurrentCommand, StateSummary, EngineStatus
 
 from server_utils.fastapi_utils.app_state import (
     AppState,
@@ -23,11 +21,13 @@ class RunsPublisher:
         self._client = client
         self._run_data_manager_polling = asyncio.Event()
         self._previous_current_command: Union[CurrentCommand, None] = None
+        self._previous_state_summary_status: Union[EngineStatus, None] = None
 
     # TODO(jh, 2023-02-02): Instead of polling, emit current_commands directly from PE.
     async def begin_polling_engine_store(
         self,
         get_current_command: Callable[[str], Optional[CurrentCommand]],
+        get_state_summary: Callable[[str], Optional[StateSummary]],
         run_id: str,
     ) -> None:
         """Continuously poll the engine store for the current_command.
@@ -38,7 +38,9 @@ class RunsPublisher:
         """
         asyncio.create_task(
             self._poll_engine_store(
-                get_current_command=get_current_command, run_id=run_id
+                get_current_command=get_current_command,
+                run_id=run_id,
+                get_state_summary=get_state_summary,
             )
         )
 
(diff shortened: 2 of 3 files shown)


SyntaxColoring (approved):
Yep, makes sense, thanks.

Have you given any thought to how we can catch this kind of thing with automated tests? It seems like we need some kind of integration test; unit tests would have been insufficient because the robot-server code was doing the "intended thing," it's just that what was intended was not sufficient for correct end-to-end behavior.

robot-server's current integration tests are doing a bit of polling themselves, for things like "wait until the run has completed." We can write a notification client in Python and replace that polling with subscriptions. That won't give us coverage of every field. But maybe it's the best bang for buck nonetheless?

app/src/organisms/RunTimeControl/hooks.ts
```diff
@@ -77,11 +77,9 @@ export function useRunStatus(
     refetchInterval: DEFAULT_STATUS_REFETCH_INTERVAL,
     enabled:
       lastRunStatus.current == null ||
-      !([
-        RUN_STATUS_STOP_REQUESTED,
```
shlokamin, line 81: still confused about how this worked lol

shlokamin (approved):

mjhuff:
> Yep, makes sense, thanks.
> 
> Have you given any thought to how we can catch this kind of thing with automated tests? It seems like we need some kind of integration test; unit tests would have been insufficient because the robot-server code was doing the "intended thing," it's just that what was intended was not sufficient for correct end-to-end behavior.
> 
> robot-server's current integration tests are doing a bit of polling themselves, for things like "wait until the run has completed." We can write a notification client in Python and replace that polling with subscriptions. That won't give us coverage of every field. But maybe it's the best bang for buck nonetheless?

Good question. Totally agree that we need integration testing here - working that out is my next ticket item. Thanks for pointing me towards the existing logic - I might see if I can repurpose it instead of inventing a new wheel entirely.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -69,12 +72,23 @@ async def _poll_engine_store(
...
             ):
                 await self._publish_current_command()
                 self._previous_current_command = current_command
+
+            if (
+                current_state_summary_status is not None
+                and self._previous_state_summary_status != current_state_summary_status
+            ):
+                await self._publish_runs_async(run_id=run_id)
+                self._previous_state_summary_status = current_state_summary_status
```
sanni-t, line 91: There will be certain status transitions (like IDLE -> RUNNING -> PAUSED) that will be handled by the robot server and hence (I'm guessing) will have a notification sent out. How is the publisher differentiating between such a transition vs one that requires polling. Because would it not send out duplicate notifications otherwise? Or does that not matter since it's just a tiny redundancy?

sanni-t (approved):
I haven't gone through all details of the notifications work done so far so I might be missing something, but this lgtm. Just a question about possible duplicate notifications.

SyntaxColoring:
> Good question. Totally agree that we need integration testing here - working that out is my next ticket item. Thanks for pointing me towards the existing logic - I might see if I can repurpose it instead of inventing a new wheel entirely.

Here's one example of where we're doing that, FYI. Grepping for `max_retries` in `*.tavern.yaml` shows ~22 instances of this kind of thing.

https://github.com/Opentrons/opentrons/blob/d81e1c340768ad343a923aa7343d2f9022a59d4e/robot-server/tests/integration/http_api/runs/test_protocol_run.tavern.yaml#L133

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
mjhuff: Good question (and sorry to have missed this before the merge) - it's a redundancy here. As you note, it's relatively tiny compared to what the app does now. 

The idea is that this polling is pretty temporary - we're planning on implementing the callback strategy we discussed yesterday as a more permanent solution to this problem, and we should revisit any redundancy and clean it up then. I think it's sufficient for now, though!

sanni-t: Got it. Thanks for the clarification!

PR #14430, opened 2024-02-06
chore: Update .nvmrc from 16 to 18

SyntaxColoring:
# Overview

Update the Node version in the monorepo-wide .nvmrc from 16 to 18, to match PR #14314.

# Test Plan

Make sure CI keeps passing.

# Risk assessment

Low.


sfoster1 (approved):
Seems like a good idea

PR #14431, opened 2024-02-06
fix(api): use motion lock when updating firmware

ahiuchingau:
# Overview

The 96-channel pipette is reported to drop after a firmware update in testing. This is because while the update is happening, the pipette cannot respond to any other CAN requests. If the hardware controller probes the network at this time, it would not be able to detect the presence of the 96-channel pipette and mistakenly lower the Z-axis motor current required to hold the pipette in place.

This PR fixes this by making sure the update firmware call block other motion tasks until all firmware has been updated.


sfoster1 (approved):
LGTM!

PR #14433, opened 2024-02-06
fix(api): analysis and simulation of OT-2 protocols no longer hang

jbleon95:
# Overview

Fixes RSS-462 and RQA-2277.

In a previous PR (#14358) we added having trash bins and waste chutes immediately added to the addressable area store when they are loaded in a python protocol. Previously they'd only be added to the engine store when they were first referenced, which could cause confusing line numbers in error messages and didn't allow any deck conflict related checks to rely on them being in the store during analysis.

Unfortunately, the method used to add it to the store relied on the `SyncClient`, and since this was being called in the `__init__` of the `ProtocolContext` for OT-2s for api levels 2.16 and above, this was causing a deadlock since this was being called from the main thread and not the thread running the PAPI protocol.

This PR fixes this in the most simple and straightforward way, which is introducing a new, optional boolean argument to `append_disposal_location` that skips this call. This is now used in the call in the `__init__` and only there, so Flex trashes and the waste chute will still be added to the engine store immediately. The OT-2 fixed trash will now be added to the store when it is first referenced (either by a `drop_tip`, `dispense`, `blow_out` or a `move_to` call), which is how it acted before #14358.

This does have the consequence that the OT-2 fixed trash, unlike any other trash bin, will behave slightly differently and inconsistently. In practice, this shouldn't lead to any issues or false positives in analysis. The only checks using the addressable area store for trashes involve the 96 channel pipette, which is Flex only, and all other checks don't rely on the addressable area store.

# Test Plan

This empty OT-2 protocol does not hang during analysis or simulation but instead passes/completes.

```
requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.16"
}


def run(context):
    pass
```

# Changelog

- added boolean argument to protocol core `append_disposal_location` to allow skipping adding the area to the engine store
- OT-2 fixed trashes no longer automatically get added to the engine store at start of protocol

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -132,7 +132,9 @@ class ProtocolCore(
                 )
 
     def append_disposal_location(
-        self, disposal_location: Union[Labware, TrashBin, WasteChute]
+        self,
+        disposal_location: Union[Labware, TrashBin, WasteChute],
+        skip_add_to_engine: bool = False,
     ) -> None:
         """Append a disposal location object to the core"""
         if isinstance(disposal_location, TrashBin):
@@ -150,7 +152,8 @@ class ProtocolCore(
                 existing_labware_ids=list(self._labware_cores_by_id.keys()),
                 existing_module_ids=list(self._module_cores_by_id.keys()),
             )
-            self._engine_client.add_addressable_area(disposal_location.area_name)
+            if not skip_add_to_engine:
+                self._engine_client.add_addressable_area(disposal_location.area_name)
         elif isinstance(disposal_location, WasteChute):
             # TODO(jbl 2024-01-25) hardcoding this specific addressable area should be refactored
             #   when analysis is fixed up
@@ -163,7 +166,8 @@ class ProtocolCore(
             self._engine_client.state.addressable_areas.raise_if_area_not_in_deck_configuration(
                 "1ChannelWasteChute"
             )
-            self._engine_client.add_addressable_area("1ChannelWasteChute")
+            if not skip_add_to_engine:
+                self._engine_client.add_addressable_area("1ChannelWasteChute")
         self._disposal_locations.append(disposal_location)
 
     def get_disposal_locations(self) -> List[Union[Labware, TrashBin, WasteChute]]:
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -150,10 +150,10 @@ class ProtocolContext(CommandPublisher):
         }
         self._bundled_data: Dict[str, bytes] = bundled_data or {}
 
-        # With the addition of Moveable Trashes and Waste Chute support, it is not necessary
+        # With the addition of Movable Trashes and Waste Chute support, it is not necessary
         # to ensure that the list of "disposal locations", essentially the list of trashes,
         # is initialized correctly on protocols utilizing former API versions prior to 2.16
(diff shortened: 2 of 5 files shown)


y3rsh:
### Analyses snapshot test run

 [ Analyses Snapshot Test #76](https://github.com/Opentrons/opentrons/actions/runs/7805816606)

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -168,7 +168,13 @@ def __init__(
...
                 location=DeckSlotName.FIXED_TRASH, addressable_area_name="fixedTrash"
             )
-            self._core.append_disposal_location(_fixed_trash_trashbin)
+            # We have to skip adding this fixed trash bin to engine because this __init__ is called in the main thread
+            # and any calls to sync client will cause a deadlock. This means that OT-2 fixed trashes are not added to
+            # the engine store until one is first referenced. This should have minimal consequences for OT-2 given that
+            # we do not need to worry about the 96 channel pipette and partial tip configuration with that pipette.
+            self._core.append_disposal_location(
+                _fixed_trash_trashbin, skip_add_to_engine=True
+            )
```
SyntaxColoring, line 177: Do you need the same modification on the other `append_disposal_location()` call, on line 163?

jbleon95: No, because that's a labware and that doesn't go down the code path that calls the sync client (since OT-2 trash labware is instantiated in protocol engine start)

sanni-t (approved):
LGTM!

api/src/opentrons/protocol_api/core/engine/protocol.py
```diff
@@ -150,7 +152,8 @@ def append_disposal_location(
                 existing_labware_ids=list(self._labware_cores_by_id.keys()),
                 existing_module_ids=list(self._module_cores_by_id.keys()),
             )
-            self._engine_client.add_addressable_area(disposal_location.area_name)
+            if not skip_add_to_engine:
+                self._engine_client.add_addressable_area(disposal_location.area_name)
```
SyntaxColoring, line 156: > This PR fixes this in the most simple and straightforward way, which is introducing a new, optional boolean argument to append_disposal_location that skips this call.

Hm, okay. Did you get a chance to try this alternative from RSS-464?

> * Have `ProtocolContext.__init__()` call `ProtocolEngine.add_addressable_area()` directly, without going through the SyncClient.

I wouldn't expect that to be much more difficult than this, but I admit I didn't try it.

I'm uncomfortable (as I think you are) with introducing the `skip_add_to_engine` complication to a system that's already difficult to reason about (as alluded to in https://github.com/Opentrons/opentrons/pull/14358#pullrequestreview-1846378144). I'm fine with merging this to unbreak `edge`, since it seems like an implementation detail that we can silently improve in another PR. But how do you feel about doing that improvement in an immediate follow-up? As in, we'd actually do it this week, before picking up any additional work, not just leave it as a `# todo` and write a tech debt ticket.

jbleon95: My initial attempts at getting that method to work ran into issues that made me decide to go along with the one this PR implements, but I agree with your assessment that this introduces an additional complication into the system and that with further work and more time there can be a more holistic solution.

I am willing and able to continue to iterate on this this week to make this cleaner, while merging this PR to unbreak edge.

api/src/opentrons/protocol_api/protocol_context.py
SyntaxColoring: o i c

SyntaxColoring (approved):
Thanks for jumping on this so quickly. It’s a spicy one.

PR #14437, opened 2024-02-06
fix(api): home before retrieving gantry position for disengaged Z mounts in FLEX

ahiuchingau:
# Overview
On the FLEX, we _always_ disengage the Gripper mount or the 96-channel Pipette mount whenever they're idle (i.e. at the home position and not being used). This allows us to better secure the instruments in place using the high-torque electromagnetic brakes, instead of relying solely on the motor hold current. 

There's one caveat: the motor loses its position tracking when it becomes disengaged. Whenever we want a motor to come out of the idle state, we have to home the motor first to reset its position before we can move it again. 

This means that we need to provide the hardware controller API user a way to re-enable an idle motor before accessing it's position. This PR does that by presenting a public facing call `prepare_for_mount_movement()`. This function readies a mount motor for future movement and position-retrieval by getting it out of the idle state. 

We can now use this function in MoveToMaintenancePosition to ensure the left mount is already ready to move!

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1158,13 +1158,13 @@ class OT3API(
         else:
             checked_max = None
 
+        await self.prepare_for_mount_movement(realmount)
         await self._move(
             target_position,
             speed=speed,
             max_speeds=checked_max,
             expect_stalls=_expect_stalls,
         )
-        self._last_moved_mount = realmount
 
     async def move_axes(  # noqa: C901
         self,
@@ -1269,6 +1269,8 @@ class OT3API(
             checked_max: Optional[OT3AxisMap[float]] = max_speeds
         else:
             checked_max = None
+
+        await self.prepare_for_mount_movement(realmount)
         await self._move(
             target_position,
             speed=speed,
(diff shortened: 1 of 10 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -247,6 +247,22 @@ def estop_cb(event: HardwareEvent) -> None:
...
+
+        If gripper mount is NOT the last moved mount, it's idle.
+        If a 96-channel pipette is attached, the mount is idle if it's not
+        the last moved mount.
+        """
+        realmount = OT3Mount.from_mount(mount)
+        if not self._last_moved_mount or realmount == self._last_moved_mount:
+            return False
+
+        return (self._gantry_load == GantryLoad.HIGH_THROUGHPUT) or (
```
```suggestion
        return (realmount == OT3Mount.LEFT and self._gantry_load == GantryLoad.HIGH_THROUGHPUT) or (
```

sfoster1: right?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1298,8 +1317,16 @@ async def _cache_and_maybe_retract_mount(self, mount: OT3Mount) -> None:
```
sfoster1, line 1321: the gripper mount above should get this treatment right?

sfoster1 (changes requested):
I think there's some small details of the logic to iron out here - I think we can use the idle mount stuff for the gripper too, but its logic might need to change slightly

sfoster1 (approved):
Okay, code looks good to me!

PR #14441, opened 2024-02-07
chore: verbose pypi deploys to catch errors

sfoster1:
Our pypi deploys seem to have succeeded in that the build is on pypi (https://pypi.org/project/opentrons/7.1.1a0/#files) but failed in that the action said there was an error and you should run with --verbose, so let's do that.

### DIFF
--- a/scripts/python.mk
+++ b/scripts/python.mk
@@ -69,7 +69,7 @@ endef
 # parameter 2: repository url
 # parameter 3: the wheel file to upload
 define python_upload_package
-$(python) -m twine upload --repository-url $(2) $(1) $(3)
+$(python) -m twine upload $(and $(CI),--verbose) --repository-url $(2) $(1) $(3)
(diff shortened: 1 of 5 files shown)


scripts/python.mk
```diff
@@ -68,8 +68,9 @@ endef
 # parameter 1: auth arguments for twine
 # parameter 2: repository url
 # parameter 3: the wheel file to upload
+# parameter 4: the sdist to upload
 define python_upload_package
-$(python) -m twine upload --repository-url $(2) $(1) $(3)
+$(python) -m twine upload $(and $(CI),--verbose) --repository-url $(2) $(1) $(3) $(4)
```
SyntaxColoring, line 73: I'd be down for `--verbose` to be unconditional, if that simplifies things.

SyntaxColoring (approved):

PR #14383, opened 2024-01-30
feat(robot-server): Support downgrades without clearing robot-server data

SyntaxColoring:
I meant to merge #14329 into `edge`, but I accidentally merged it into a throwaway branch. This PR merges it into `edge`.

No code review should be necessary, since that already happened in #14329.

PR #14442, opened 2024-02-07
fix(robot-server): Fix protocols being stored in the wrong directory

SyntaxColoring:
# Overview

Closes RSS-469 and associated issues, like RQA-2294 and RQA-2295.

# Test Plan

I've confirmed that on `edge`, these steps reproduce the error in described in RQA-2295, and with this branch, they do not:

1. Put a lot of different protocols ([≥20 if you're using a Flex](https://github.com/Opentrons/oe-core/blob/9942b8aa873109af2000e42929dd057ab6f4d28a/layers/meta-opentrons/recipes-robot/opentrons-robot-server/files/opentrons-robot-server.service#L16) or [≥5 if you're using a dev server](https://github.com/Opentrons/opentrons/blob/8f4b709fd3cd865ea427e6e5dd0861f7397f75a4/robot-server/robot_server/settings.py#L88)) on a v7.1.0 server.
2. Upgrade to this branch.
3. Upload a new protocol. This will auto-delete one of the old ones to make room.
4. Downgrade back to v7.1.0.
5. Make sure all the original protocols are still available (either by manually issuing HTTP GET requests or by clicking through the Opentrons App).

# Changelog

robot-server stores stuff in a `protocols/` directory and in a `robot_server.db` file. They need to be kept in sync.

The bug, introduced in #14383, is that after a migration to schema 3, we were still using the `protocols/` directory from schema 2. So the server was working from the new `robot_server.db` file but the old `protocols/` directory. When you downgraded back to v7.1.0, the server would hit errors from the old `protocols/` directory now being out of sync with the old `robot_server.db` file. This was caused by a pretty embarrassing mistake, and the fix is one line.

By fixing that, we expose another problem. The `POST /settings/reset` endpoint lets you clear deck configuration. But it can only do that after the persistence layer has initialized. And that endpoint is supposed to be highly available; it's supposed to work even if the database is corrupt, or taking a million hours to migrate. This PR's quick and dirty solution is to return HTTP 503 if you asked to clear deck configuration but things aren't yet initialized far enough to do that.

# Review requests

1. Does my treatment of `POST /settings/reset` seem sensible? It's not really clear to me how that endpoint should ideally behave. Should it block until it's able to reset everything that was requested? Should `deckConfiguration` even be an option there, as opposed to it being a separate endpoint like `DELETE /deck_configuration`?
2. What do we want to do with the internal robots that have already been infected with this borked migration? Unfortunately, I don't see a sensible way of recovering from it. We can reset those robots' runs and protocols, but then we lose some QA coverage of the migration. I could run around the office and manually fix things up, I guess? There are "only" 10-15 robots that have gone through this update so far, right?

# Risk assessment

Medium. As the bugs show, this upgrade/downgrade stuff is not easy to catch with our existing automated tests.


robot-server/robot_server/persistence/_fastapi_dependencies.py
```diff
@@ -220,7 +220,9 @@ async def get_active_persistence_directory(
     If this is called before that initialization completes, this will raise an
     appropriate HTTP-facing error to indicate that the server is busy.
     """
-    initialize_task = _root_persistence_directory_init_task_accessor.get_from(app_state)
+    initialize_task = _active_persistence_directory_init_task_accessor.get_from(
+        app_state
+    )
```
SyntaxColoring, line 225: This is the main bug fix. I was using `_root_persistence_directory_init_task_accessor` when I should have been using `_active_persistence_directory_init_task_accessor`.

robot-server/robot_server/deck_configuration/fastapi_dependencies.py
```diff
@@ -30,9 +34,33 @@ async def get_deck_configuration_store(
...
+    if deck_configuration_store is None:
+        path = persistence_directory / _DECK_CONFIGURATION_FILE_NAME
+        # If this initialization becomes async, we will need to protect it with a lock,
+        # to protect against the bug described in https://github.com/Opentrons/opentrons/pull/11927.
+        deck_configuration_store = DeckConfigurationStore(deck_type, path)
+        _accessor.set_on(app_state, deck_configuration_store)
+    return deck_configuration_store
+
+
+# TODO(mm, 2024-02-07): Resolve the duplication between these two implementations.
```
SyntaxColoring, line 47: In hindsight, it looks like we want a pattern like this:

```python
async def get_foo_store(
    foo_store_future: asyncio.Future[FooStore] = fastapi.Depends(FooStore),
) -> FooStore:
    """A FastAPI dependency to get the server's singleton FooStore.

    If the FooStore is still initializing, this will block until it's done.
    If the FooStore has failed to initialize, this will raise that exception.
    """
    return await foo_store_future.result()


async def get_foo_store_future(
    app_state: AppState,
    persistence_directory_future: asyncio.Future[Path] = fastapi.Depends(get_persistence_directory_future)
) -> asyncio.Future[FooStore]:
    """A FastAPI dependency to get the server's singleton FooStore, as a Future.

    This is like `get_foo_store()`, except this will always return immediately and
    never fail. It's useful for endpoints that will implement their own logic
    for tolerating FooStore unavailability.
    """
    ...
```

That seems pretty straightforward. It just happens to not be the first thing that I got to work here.

CaseyBatten (approved):
Looks good to me! On the deck configuration issues:
We should ticket out that issue (as well as the semi-related issue that resetting protocol run history also resets the deck configuration) to solve in a separate PR. The issues seems too broad for this one PR as it will inevitably effect pieces on the APP/UI end, and the failing to reset the the deck configuration during database migration is not a large enough blocker to prevent merging this PR.

SyntaxColoring:
> On the deck configuration issues: We should ticket out that issue

RSS-465

> (as well as the semi-related issue that resetting protocol run history also resets the deck configuration)

RSS-403

SyntaxColoring:
> What do we want to do with the internal robots that have already been infected with this borked migration?

I discussed this with @nusrat813. I think it's mostly a moot point. Most or all of these robots have already had their protocols and runs reset.

PR #14443, opened 2024-02-07
fix(api): raise errors from modules disconnecting

sfoster1:
When a module gets disconnected, the hardware controller cleans up the module instance, which stops the poller. What it doesn't do is cancel anything that was waiting on the next poll, or in fact prevent new things from waiting on the now-stopped poller. That means that if a module disconnects
- During a module method waiting for the next poll, that module method hangs (well, awaits) forever
- During a module method right _before_ waiting for the next poll, that module method would start waiting and continue waiting forever

This PR forwards cancellations into the registered poll waiters when the poller task is cancelled, and prevents the registration of new poll waiters on a cancelled poll task.

However, it does not fix the following issue: the protocol engine does not maintain a lifetime associated with a particular module connection, but looks up a module object on each call based on the serial number. That means that if a module is disconnected and reconnected during a protocol, as long as the engine isn't actually communicating with it at that exact time the protocol will continue as if nothing has happened - even if, for instance, the module disconnected because it was powered down, and the user-specified module state has been lost.

### DIFF
--- a/api/src/opentrons/hardware_control/poller.py
+++ b/api/src/opentrons/hardware_control/poller.py
@@ -3,6 +3,7 @@ import contextlib
 import logging
 from abc import ABC, abstractmethod
 from typing import AsyncGenerator, List, Optional
+from opentrons_shared_data.errors.exceptions import ModuleCommunicationError
 
 
 log = logging.getLogger(__name__)
@@ -48,6 +49,8 @@ class Poller:
         async with self._use_read_lock():
             task.cancel()
             await asyncio.gather(task, return_exceptions=True)
+        for waiter in self._poll_waiters:
+            waiter.cancel(msg="Module was removed")
 
     async def wait_next_poll(self) -> None:
         """Wait for the next poll to complete.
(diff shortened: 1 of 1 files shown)


sanni-t (approved):
Changes make sense to me! Thanks for the quick fixes!
Can test this soon.

> However, it does not fix the following issue: the protocol engine does not maintain a lifetime associated with a particular module connection, but looks up a module object on each call based on the serial number. That means that if a module is disconnected and reconnected during a protocol, as long as the engine isn't actually communicating with it at that exact time the protocol will continue as if nothing has happened - even if, for instance, the module disconnected because it was powered down, and the user-specified module state has been lost.

Thinking about it a bit more, I think it's fine to keep this behavior of the engine. As you said, if the re-connected module is in the same state as before, then the protocol will continue without any problems, and that's good. If it's not in the same state as before, the engine will error out at the point where the diverged states will cause critical issues (like if a gripper is trying to move labware from a h/s whose engine state says it's latch is open but it's hardware state actually has latch closed). Maybe we can improve this sync between actual h/w and assumed logical state instead of making the protocol error out even if the reconnected module is in the expected state.

caila-marashaj (approved):

PR #14445, opened 2024-02-07
Abr asair sensor

rclarke0:
# Overview

This PR adds a script to the hardware-testing folder that will allow the robot to connect to an asair temperature and humidity sensor and record continuously to a google spread sheet. 

# Test Plan

I tested this script on robot DVT10 on software v7.1.1. It successfully measured the temperature and humidity and uploaded to the google sheet. 

# Changelog

Added abr-asair-sensor script
Script:

1. Asks user for robot name
2. Asks user for duration of recording (min)
3. Asks user for frequency of recording (min)
4. Robot reads ports and connects to asair sensor
5. robot collects data and appends to a list at the given frequency
6. At the end of the duration the robot saves a .csv file of the data to the testing-data folder within the robot and to the google sheet
I also updated the analyze-abr script to improve ease of analysis.

IMPORTANT: MAKE SURE ANY BREAKING CHANGES ARE PROPERLY COMMUNICATED
-->

# Review requests

# Risk assessment

- for the script to run successfully on the robot, google_sheets_tool.py and a google sheets credentials file needs to be uploaded manually onto the robot's jupyter notebook.
- for google_sheets_tool.py to run successfully within the script python libraries: httplib2, oauth2client, gspread, and pprintpp need to be installed onto the robots


hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
```diff
@@ -1,6 +1,126 @@
...
+try:
+    sys.path.insert(0, "/var/lib/jupyter/notebooks")
+    import google_sheets_tool  # type: ignore[import]
+except ImportError:
+    raise ImportError(
+        "Run on robot. Make sure google_sheets_tool.py is in jupyter notebook."
+    )
+
+
+def _get_user_input(list: List, some_string: str) -> str:
```
sfoster1: `list` is actually a predefined term in python - it is the name of the list type. defining a variable called `list` will shadow the global name in the scope where the variable is defined, which would lead to surprising results if you wanted to build a list in this function. Would you mind calling this something else, like `lst` or `input_list` or something?

sfoster1: Unrelatedly, `List` is a generic container type that is meant to be specialized with the type it contains - for instance, `List[str]` is a list of strings, `List[int]` is a list of ints.

hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
```diff
@@ -1,6 +1,126 @@
...
+    variable = input(some_string)
+    while variable not in list:
+        print(
+            f"Your input was {variable}. Expected input is one of the following: {list}"
+        )
+        variable = input(some_string)
+    return variable
+
+
+class _abr_asair_sensor:
```
sfoster1: fyi, common python style is for classes to be `PascalCase` - in this case, `_ABRAsairSensor` if the class isn't meant to be exported and `ABRAsairSensor` if it is

hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
rclarke0: changed to lst

rclarke0: added List[str]

hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
rclarke0: changed class name _ABRAsairSensor

hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
```diff
@@ -1,6 +1,130 @@
...
+        except FileNotFoundError:
+            print(
+                "There is no google sheets credentials. Make sure credentials in jupyter notebook."
+            )
+        results_list = []  # type: List
+        start_time = datetime.datetime.now()
+        while True:
+            env_data = sensor.get_reading()
+            timestamp = datetime.datetime.now()
+            new_timestamp = timestamp - datetime.timedelta(hours=5)
```
andySigler, line 66: is this b/c of the timezone? either leaving a comment to describe what this is doing or somehow specifying the system's timezone would make it more clear

hardware-testing/hardware_testing/scripts/abr_asair_sensor.py
```diff
@@ -1,6 +1,130 @@
...
+                print("Did not write row.")
+            # Delay for desired frequency minutes before the next iteration
+            t.sleep(frequency * 60)  # seconds
+
+        # Upload to robot testing data folder
+        result_string = ""
+        for sublist in results_list:
+            row_str = ", ".join(map(str, sublist)) + "\n"  # type: str
+            result_string += row_str
+            save_file_path = data.append_data_to_file(
```
andySigler, line 100: here you are appending the line to all previous lines, and then you appending that string to the file, and doing this for each line, which means that the file you are writing will be huge and filled with repeated lines.

This can be fixed by removing the `result_string` variable and just passing the `row_str` into `append_data_to_file()`.

Then also, you can append each line to the CSV at the same time that you write the data to the google sheet. Just move the `append_data_to_file()` up into while loop above. Then the CSV will be updated in realtime, plus you don't need to store data in the `results_list` variable anymore.

andySigler (approved):
This is awesome. Just a few comments that shouldn't block merging

PR #14446, opened 2024-02-07
fix(api): keep proto contents in bytes for longer

sfoster1:
When we parse python protocols, we were doing it by (1) making it into a string with decode('utf-8') and then (2) passing the string ast.parse(). The problem with this is that decode('utf-8') does not apply "universal newlines", which means that the code object created by compiling the ast will have line numbers that are around twice what they should be under certain circumstances (windows machine, crlf file, mercury in the seventh house, etc). Then, when we go and display a nice error message about a syntax error or whatever, the user says "why is this error message pointing to a place past the end of my protocol".

This should fix that by keeping the protocol contents in bytes form all the way through to passing ast.parse() a bytes that _has never been through str.decode('utf-8')_ which should preserve everything.

# Testing
- Use a protocol like the attached
[flex_mag_mod.py.zip](https://github.com/Opentrons/opentrons/files/14201855/flex_mag_mod.py.zip) (unzip it first) that has (1) an error and (2) crlf newlines
- On windows, simulate with `opentrons_simulate` and in the app
- check that it says the error is on the right line, and not the line*2

### DIFF
--- a/api/src/opentrons/protocols/parse.py
+++ b/api/src/opentrons/protocols/parse.py
@@ -12,7 +12,7 @@ import re
 import traceback
 from io import BytesIO
 from zipfile import ZipFile
-from typing import Any, Dict, Optional, Union, Tuple, TYPE_CHECKING, cast
+from typing import Any, Dict, Optional, Union, Tuple, TYPE_CHECKING
 
 import jsonschema  # type: ignore
 
@@ -500,17 +500,6 @@ def _has_api_v1_imports(parsed: ast.Module) -> bool:
     return bool(v1_markers.intersection(opentrons_imports))
 
 
-def _strvalue_from_str_or_byte_dict(
-    dct: Union[Dict[str, str], Dict[bytes, bytes]], key: str
-) -> Optional[str]:
-    key_bytes = key.encode("utf-8")
-    val_bytes = cast(Dict[bytes, bytes], dct).get(key_bytes, None)
-    val_str = cast(Dict[str, str], dct).get(key, None)
-    if val_bytes:
-        return val_bytes.decode("utf-8")
-    return val_str
-
-
 def _version_from_static_python_info(
     static_python_info: StaticPythonInfo,
 ) -> Optional[APIVersion]:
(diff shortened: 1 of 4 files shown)


api/src/opentrons/protocols/parse.py
```diff
@@ -489,6 +500,17 @@ def _has_api_v1_imports(parsed: ast.Module) -> bool:
...
 
+def _strvalue_from_str_or_byte_dict(
+    dct: Union[Dict[str, str], Dict[bytes, bytes]], key: str
+) -> Optional[str]:
+    key_bytes = key.encode("utf-8")
+    val_bytes = cast(Dict[bytes, bytes], dct).get(key_bytes, None)
+    val_str = cast(Dict[str, str], dct).get(key, None)
+    if val_bytes:
+        return val_bytes.decode("utf-8")
+    return val_str
```
SyntaxColoring: Hm, is this doing anything? As far as I can tell, `dct` is always going to be a `Dict[str, str]`, never a `Dict[bytes, bytes]`. Even when the source is a `bytes`:

```python
>>> b = b"requirements = {'hello': 'world'}"

>>> ast.parse(b).body[0]
<ast.Assign object at 0x105218a00>

>>> ast.literal_eval(ast.parse(b).body[0].value)
{'hello': 'world'}

>>> type(ast.literal_eval(ast.parse(b).body[0].value)["hello"])
<class 'str'>
```

SyntaxColoring (commented):
Wow, juicy.

Very happy to have less encoding/decoding. Thank you!

api/src/opentrons/protocols/parse.py
sfoster1: Without that, tests where we pass bytes-encoded protocols were failing (or so I thought). But you're right, and in fact if you look at `_extract_static_dict` we explicitly ensure those things are `str` instances, so maybe I got the wrong fix for the problem.

sfoster1: @SyntaxColoring https://github.com/Opentrons/opentrons/pull/14446/commits/dbe84179a803922379db5bbddce505ce23111174 Yeah totally not needed. Weird, I definitely thought this fixed a test, but there were other required test changes so it must have been something else

SyntaxColoring (approved):
Thank you! CI is failing but it looks like some Python environment thing unrelated to this PR. This LGTM if tests and linting pass locally.

PR #14447, opened 2024-02-07
fix(app): open external links in browser

shlokamin:
# Overview

This PR fixes a bug introduced by [updating electron](https://github.com/Opentrons/opentrons/pull/14314) where we were opening up external links in the electron window rather than an external browser window. 

As per the (poorly worded) [docs](https://www.electronjs.org/docs/latest/api/web-contents#contentssetwindowopenhandlerhandler), `contents.setWindowOpenHandler` is...

> Called before creating a window a new window is requested by the renderer, e.g. by window.open(), a link with target="_blank", shift+clicking on a link, or submitting a form with <form target="_blank">

So the checks we were doing before are unnecessary.

closes RQA-2309

# Test Plan

Verified internal links do not open in the browser, and verified external links do.

# Changelog

- Open external links in the browser (outside of electron)

# Review requests

Click around the app and make sure internal/external links work as expected.
# Risk assessment
Low

### DIFF
--- a/app-shell/src/ui.ts
+++ b/app-shell/src/ui.ts
@@ -57,14 +57,10 @@ export function createUi(): BrowserWindow {
   mainWindow.loadURL(url, { extraHeaders: 'pragma: no-cache\n' })
 
   // open new windows (<a target="_blank" ...) in browser windows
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
nice!

PR #14448, opened 2024-02-08
feat(app): Add Robot Initialization State

mjhuff:
Closes RAUT-944

# Overview
This PR adds initialization state to the desktop & ODD when the robot server does things like undergo a database migration. On the desktop app, initialization checks occur during the robot update progress modal and ensure that the robot server has initialized before proceeding to show the "update complete" modal. On the ODD, RecentRunProtocolCards now render their skeleton if initialization occurs.

Although the new useRobotInitializationStatus hook supports success/initializing/failure codes, for now we're keeping it simple and just showing the above while the robot server initializes. 

# Test Plan
TO TEST DESKTOP APP.
- Downgrade a robot to an early 7.2 alpha.
- SSH in, stop the robot server, `rm -rf /var/lib/opentrons-robot-server/3`, start the server.
- Restart the robot.
- For the desktop app, check out this branch and push [this special build ](https://opentrons.slack.com/archives/C81CR4VCZ/p1707337732789829)to the robot, which artificially sets a migration timer of 10 minutes. 

TO TEST ODD.
- Have a protocol card and stop the robot server. You should see the shimmer. NOTE: This won't work with the above build, you'll have to push app-shell-odd to test. 
- I can't actually get a protocol card to display after the custom update given how the artificial branch works (run history gets deleted on the artifical testing branch...I think this is filed as a bug). That being said, the logic is the same as the desktop (which works) and easy to reason about. I think seeing the shimmer when the init status == null is sufficient when coupled with the unit testing. 

# Changelog
- Added loading state when the robot server is initializing for the desktop app/ODD.

# Risk assessment
low


app/src/organisms/Devices/RobotSettings/UpdateBuildroot/RobotUpdateProgressModal.tsx
```diff
@@ -92,34 +93,18 @@ export function RobotUpdateProgressModal({
       installFromFileRef.current.click()
   }, [showFileSelect])
 
-  const hasStoppedUpdating = error || updateStep === 'finished'
+  const robotInitStatus = useRobotInitializationStatus()
+  const hasRobotCompletedInit =
+    updateStep === 'finished' && robotInitStatus !== INIT_STATUS.INITIALIZING
```
mjhuff, line 102: I guess it's theoretically possible for the server to fail initialization and show the "update successful" modal, but there aren't any designs for this, and the app sufficiently informs the user something is wrong with the robot. I'm open to discuss if we want to add anything here. I do think it's already pretty explicit the server is down when the user sees the red banner informing them of that, however.

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/index.tsx
```diff
@@ -50,11 +53,19 @@ const UpdateBuildroot = NiceModal.create(
 
     if (hasSeenSessionOnce.current)
       return (
-        <RobotUpdateProgressModal
-          robotName={robotName.current}
-          session={session}
-          closeUpdateBuildroot={modal.remove}
-        />
+        <ApiHostProvider
```
mjhuff, line 56: The nice thing about the ebay-nice-modal library is that modals aren't children of the components that render them, but this is an annoying effect of that. We have to add `ApiHostProvider` back in.

react-api-client/src/health/useHealth.ts
mjhuff, line 1: This hook isn't anywhere else in the app, so I generalized it a bit more (we don't return `response.data` now, just `response`).

app/src/organisms/Devices/RobotSettings/UpdateBuildroot/RobotUpdateProgressModal.tsx
```diff
@@ -313,3 +306,42 @@ function useCleanupRobotUpdateSessionOnDismount(): void {
...
+    case 'download':
+      modalBodyText = t('downloading_update')
+      break
+    case 'install':
+      modalBodyText = t('installing_update')
+      break
+    case 'restart':
+      if (robotInitStatus === INIT_STATUS.INITIALIZING) {
+        modalBodyText = t('robot_initializing')
+        subProgressBarText = t('do_not_turn_off', { minutes: 40 })
```
mjhuff, line 338: I've increased the default user exit time for the migration to 40 minutes. I think Max said 30 minutes max, but IMO I think adding a buffer here is fine.

app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
```diff
@@ -83,6 +85,9 @@ export function ProtocolWithLastRun({
   const onResetSuccess = (createRunResponse: Run): void =>
     history.push(`runs/${createRunResponse.data.id}/setup`)
   const { cloneRun } = useCloneRun(runData.id, onResetSuccess)
+  const robotInitStatus = useRobotInitializationStatus()
+  const isRobotInitializing =
+    robotInitStatus === INIT_STATUS.INITIALIZING || robotInitStatus == null
```
mjhuff, line 93: The `null` check is more for dev purposes than anything else (stopping the robot-server will cause a shimmer now). I think it's a safe default, though.

app/src/organisms/Devices/hooks/useRobotInitializationStatus.ts
```diff
@@ -0,0 +1,44 @@
...
+  INITIALIZING: 'INITIALIZING',
+  SUCCEEDED: 'SUCCEEDED',
+  FAILED: 'FAILED',
+} as const
+
+export type RobotInitializationStatus =
+  | typeof INIT_STATUS[keyof typeof INIT_STATUS]
+  | null
+
+export function useRobotInitializationStatus(): RobotInitializationStatus {
```
brenthagen: could consider putting this hook in `resources/health/hooks` or similar. we're trying to move away from the `organisms/Devices/hooks` directory and instead associate hooks with a resource when possible

app/src/organisms/Devices/hooks/__tests__/useRobotInitializationStatus.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  })
+
+  it('should return "FAILED" when response status code is 500', () => {
+    ;(mockUseHealthQuery as jest.Mock).mockImplementation(({ onSuccess }) => {
+      onSuccess({ status: 500 })
+
+      const { result } = renderHook(() => useRobotInitializationStatus())
+      expect(result.current).toBe(INIT_STATUS.FAILED)
+    })
+  })
```
brenthagen: some syntax stuff going on here - the expect statements should be outside of the mock implementation, and could assert `const mockUseHealthQuery = useHealthQuery as jest.Mock` directly above instead of asserting again in each test

brenthagen (approved):
tested update from desktop app and ODD recent run protocol change, looks good

PR #14449, opened 2024-02-08
fix(api): get_slice should return last executed command if a run failed 

TamarZanzouri:
# Overview

closes [RQA-2243](https://opentrons.atlassian.net/browse/RQA-2243).
if a run has failed and the cursor is not supplied we should return the last executed command. 

# Test Plan

- Upload a json protocol that will have a failed command.
- Once run has failed fetch `/runs/{runId}/commands` make sure you are getting back the failed command and not the last command. 
- Test python protocols as well
- Test successful protocols with and without cursor. 

- Added integration tests

# Changelog

- when getting command_slice check if there is a run_result (if the run ended). if it did and its in a failed status get the failed command index as the cursor.  

# Review requests

are you ok with this approach? is it better to store the failed command in a variable? 

# Risk assessment

low. should only affect failed protocols. 

[RQA-2243]: https://opentrons.atlassian.net/browse/RQA-2243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -466,6 +466,14 @@ def get_slice(
                 cursor = commands_by_id[running_command_id].index
             elif len(queued_command_ids) > 0:
                 cursor = commands_by_id[queued_command_ids.head()].index - 1
+            elif self._state.run_result and self._state.run_result == RunResult.FAILED:
+                # failed command only contains the error that occurred
+                last_executed = next(
```
CaseyBatten: Here we're returning the next item after the command with an error state. Would this return the Next command to execute, not the command that just executed and failed? Which do we want to return in practice? 
As an example, in a protocol with 5 commands, if the protocol failed on command 3 I believe this logic would return command 4.

CaseyBatten (approved):
Looks good and definitely fixes the issues faced by the current endpoint on JSON protocols. Tested on RSS Flex, performs as expected and does not raise incorrect modals in the case provided by RQA-2242.
Question below seeking some clarification:

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -466,6 +466,14 @@ def get_slice(
...
+                    command_id
+                    for command_id, command in self._state.commands_by_id.items()
+                    if command.command.error
+                )
```
SyntaxColoring: This is an O(n) scan of a list that has, potentially, tens of thousands of elements. Can we avoid this? The performance problems can cause things like sporadic disconnections.

sfoster1: IMO for any access to the commands list other than "get the next single thing"/"get the first thing"/"get the last thing" we probably should move towards using sql

TamarZanzouri: good point @SyntaxColoring. @sfoster1 we use sql for none current runs. for current runs we use the state. I can store the failed command in the view.

SyntaxColoring:
Does this need to be retargeted into the v7.2.0 release branch, or is this intended to be for a future release?

api/tests/opentrons/protocol_engine/state/test_command_view.py
```diff
@@ -793,6 +797,37 @@ def test_get_slice_default_cursor_no_current() -> None:
     )
 
 
+def test_get_slice_default_cursor_no_current_failed_command() -> None:
```
TamarZanzouri:
```suggestion
def test_get_slice_default_cursor_failed_command() -> None:
```

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -152,6 +152,9 @@ class CommandState:
     are stored on the individual commands themselves.
     """
 
+    failed_command: Optional[CommandEntry]
+    """The command, if any, that made the run fail."""
```
TamarZanzouri:
```suggestion
    """The command, if any, that made the run fail and the index in the command list."""
```

SyntaxColoring:
Recapping a conversation that RSS had in a call:

As-is, this PR only fixes the behavior for the current run, which is served from memory, and not for historical runs, which are served from the database. Ideally, the behavior between the two should certainly be consistent. We need to ask the app team what the priority is for solving this for historical runs.

One way to solve this for historical runs is:

* A new command status, like `run-stopped-before-this-command-started`. (Except with a better name.) We suspected we'd need this a while ago, in RCORE-390.
* A new database column for the command status, for efficient querying.

robot-server/tests/integration/http_api/runs/test_json_v6_run_failure.tavern.yaml
```diff
@@ -83,44 +83,9 @@ stages:
         links:
           current: !anydict
         meta:
-          cursor: 0
+          cursor: 3
```
SyntaxColoring, line 86: Nice.

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: I think this is addressed by Tamar's recent changes to store the last-failed command in a variable instead of scanning the list.

SyntaxColoring (approved):
Check out my two comments above about the target branch and historical runs, but when those are answered, this looks great to merge. Thank you!

TamarZanzouri:
> Recapping a conversation that RSS had in a call:
> 
> As-is, this PR only fixes the behavior for the current run, which is served from memory, and not for historical runs, which are served from the database. Ideally, the behavior between the two should certainly be consistent. We need to ask the app team what the priority is for solving this for historical runs.
> 
> One way to solve this for historical runs is:
> 
> * A new command status, like `run-stopped-before-this-command-started`. (Except with a better name.) We suspected we'd need this a while ago, in RCORE-390.
> * A new database column for the command status, for efficient querying.

spoke to @shlokamin and historical runs are not urgent. will create a ticket

[Edit]: RSS-470

sfoster1 (approved):
Looks good to me!

PR #14451, opened 2024-02-08
fix(app): ensure protocol doesn't have stale status before analysis completion

smb2268:
fix RQA-2297

# Overview
During the few seconds between when `isAnalyzing` is true and before the analysis has resolved, the protocol card was showing a stale protocol banner. To avoid this, we should only return a `stale` status if `analysis != null`

# Test Plan
Upload a protocol, see that it no longer flashes a stale banner between the loading and completed states -- see test video here:
https://github.com/Opentrons/opentrons/assets/14302493/37059d7b-ab42-4296-80fc-f3bbe52eef88

# Changelog
Add `analysis != null` to condition for returning a `stale` status

# Review requests
Look over video and code

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/ProtocolsLanding/utils.ts
+++ b/app/src/organisms/ProtocolsLanding/utils.ts
@@ -10,7 +10,7 @@ export function getAnalysisStatus(
 ): AnalysisStatus {
   if (isAnalyzing) {
     return 'loading'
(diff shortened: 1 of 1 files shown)


jerader (approved):
changes make sense to me!

PR #14458, opened 2024-02-08
 feat(api, shared-data): make speed & distance pickUpTipConfigurations values keyed by tip count

sanni-t:
Closes RSS-443

# Overview

This PR looks big but it actually has only 2 real changes-
`speed` and `distance` properties of a pipette's `pickUpTipConfigurations` are now changed to `speedByTipCount` and `distanceByTipCount`. These values are now dictionaries keyed by number of tips being used for tip pickup.

# Test Plan

- Run a protocol or go through a maintenance run procedure that moves the pipette around and make sure that the pipettes move the same as before. 
- Test the above on both Flex and the OT-2.

# Changelog

- updated pipette schema for the new `speedByTipCount` & `distanceByTipCount` properties
- updated ot2 & ot3 pipette handlers to use the updated properties
- updated the pipette definitions 

# Review requests

- this was a bit too easy! Probably a testament to hardware control architecture, but maybe I've also overlooked something?
- *would appreciate a second set of eyes going through the updated definitions to make sure I've only updated the properties' shapes and kept the speed and distance values the same as before*

# Risk assessment

High if I messed up pipette definition updates.


sanni-t:
Totally forgot about the schema v1 build scripts. Updating now

sfoster1 (approved):
Code looks good, and I see you noticed the fairly cursed way we do mutable configurations, awesome.

It feels a little silly and error-prone to have three different maps with the same key rather than one map to objects with three members though. Can you either invert that (so it's like `configByTipCount: { "1": { "current": 0.1, "distance": 10, "speed": 20}, }`) or add some verification to the python modals, or even just to the tests, to make sure that nobody forgets an entry in one of the three?

sanni-t:
> It feels a little silly and error-prone to have three different maps with the same key rather than one map to objects with three members though. Can you either invert that (so it's like configByTipCount: { "1": { "current": 0.1, "distance": 10, "speed": 20}, }) or add some verification to the python modals, or even just to the tests, to make sure that nobody forgets an entry in one of the three?

That is certainly a good idea. Will try the first method as it seems the most robust.

SyntaxColoring:
Notwithstanding the changes discussed above, I tested this on a Flex and didn't notice anything out of the ordinary. I used the drop-tip wizard and whatever 96-channel protocol happened to already be on the robot.

sanni-t:
> That is certainly a good idea. Will try the first method as it seems the most robust.

@sfoster1 On second thought, ~I'll add tests right now to validate the entries.~ Will refactor the shape of the properties in a follow-up. I'll fix the failing tests and merge this to make hardware testing a bit easier.

sanni-t:
> Notwithstanding the changes discussed above, I tested this on a Flex and didn't notice anything out of the ordinary. I used the drop-tip wizard and whatever 96-channel protocol happened to already be on the robot.

Awesome! Thanks @SyntaxColoring !

sanni-t:
Tested on OT-2 and verified that:
- pickup speed & distance works just like before
- speed & distance pipette settings via app apply correctly

PR #14460, opened 2024-02-08
 fix(app): add prop so run setup pipette flows are strung together

smb2268:
Fix RQA-2305, RQA-2292, RQA-2284

# Overview

This PR fixes a few bugs (or missing implementations) in pipette flows on the ODD and app. It ensures pipette detach -> attach flows are properly strung together for ODD instrument setup and adds a "Robot is Moving" loading screen after attaching a mounting plate during 96-channel attach

# Test Plan

On ODD, 
1. From run setup with the wrong instruments attached, the ODD should now initiate a "Replace pipette" flow:
IMG_6670

2. From run setup with a single channel on the left mount for a protocol that requires an ODD, the replace pipette flow should show the correct pipette name in the detach step and lower the correct mount:
IMG_6672
IMG_6673

3. When attaching a 96-channel pipette, there should be a loading screen while the gantry moves after attaching a mounting plate

# Changelog

# Review requests
Look over code and screenshots

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/InstrumentMountItem/ProtocolInstrumentMountItem.tsx
+++ b/app/src/organisms/InstrumentMountItem/ProtocolInstrumentMountItem.tsx
@@ -19,6 +19,7 @@ import {
   NINETY_SIX_CHANNEL,
   PipetteName,
   SINGLE_MOUNT_PIPETTES,
+  LoadedPipette,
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
All LGTM!

PR #14465, opened 2024-02-09
perf(robot-server): Parallelize the migration to schema 3

SyntaxColoring:
# Overview

Closes RSS-451.

# Changelog

* During the 2->3 schema migration, use a pool of subprocesses to parse, transform, and reserialize run command data.

# Test plan

For testing for correctness, I think we're decently covered by automated tests.

For performance, I've tested this on an OT-2 along with #14480 and found that it's in line with the initial estimates in RSS-451. With a run history lifted from one of the ABR robots, with 88k commands, this cuts the 2->3 migration time from 10-11 minutes (625s) to 4-5 minutes (261s). Extrapolating to the worst case as defined in RSS-451, it would be a cut from 23-24 minutes to 9-10 minutes. 

# Review requests

See inline comments.

# Risk assessment

Medium. This is the first time (?) we're introducing multiprocessing into robot-server, and so we have to work through [some gotchas](https://github.com/Opentrons/opentrons/pull/14465#discussion_r1484989089.). And it's part of database migration code, so it's sort of a complexity turducken situation.


robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring, line 1: Here is a story that starts off about a performance problem, but then it becomes about a safety problem!

I noticed that although this PR makes the migration ~50% faster on real-world robots, it makes the persistence snapshot tests ~50% *slower* on macOS. What gives.

It turns out there is an utterly obscene amount of constant overhead on macOS. On my machine, each worker ~~thread~~ process takes 1500 ms (!!!) to start. This is partially the platform's fault, and partially our fault.

The part that's the platform's fault is: on macOS, [Python starts each subprocess with `spawn` instead of `fork`](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). They're forced to do this because [`fork` is unusably broken or something](https://bugs.python.org/issue?@action=redirect&bpo=33725), I dunno. Anyway, this means that each subprocess needs to initialize the Python interpreter and import our code from scratch, instead of inheriting all of that from the parent process.

The part that's our fault is: importing this code takes 1500 ms. I've confirmed that this is almost entirely from our `__init__.py`s re-exporting shit. Each of these subprocesses is effectively importing an entire copy of robot-server; they can't import anything without accidentally pulling in unrelated stuff like our FastAPI routers. We can save like 50% off the bat just by removing [this one re-export](https://github.com/Opentrons/opentrons/blob/4c4f49109a8e577b1da21599ebcbd439d455cc30/robot-server/robot_server/__init__.py#L6-L10), for example.

On robots, this isn't causing any *performance* problems right now because the default for Linux is still `fork`. However, for the same sort of reasons `fork` isn't used on macOS, it will cause *correctness* problems. There are [credible](https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555) [safety](https://discuss.python.org/t/switching-default-multiprocessing-context-to-spawn-on-posix-as-well/21868/22) [criticisms](https://web.archive.org/web/20201109035255/https://pythonspeed.com/articles/python-multiprocessing/) against `fork`, and [real-world horror stories](https://github.com/python/cpython/issues/84559#issuecomment-1340081137) of it causing mysterious deadlocks, and [it's getting deprecated in Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#:~:text=On%20POSIX%20platforms%2C%20os.fork()%20can%20now%20raise%20a%20DeprecationWarning%20when%20it%20can%20detect%20being%20called%20from%20a%20multithreaded%20process.).

So we have to choose between the code being fast or being reliable. Or we need to put in a bit more work to do both at once. That would entail:

1. Using `spawn` or `forkserver` instead of the platform-specific default.
2. Making that fast enough by making our imports cheaper, and possibly using [`set_forkserver_preload()`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_forkserver_preload). Remember that 1500 ms on one of our laptops is more like [1 minute on an OT-2](https://opentrons.atlassian.net/browse/RSS-133).

SyntaxColoring: I'm conflicted now over whether to leave the broken `fork`, attempt to fix it with `spawn` or `forkserver` as described above, or just abandon this whole thing and leave RSS-451 unresolved in v7.2.0.

If not for this PR, we would have hit this eventually in RSS-93.

robot-server/robot_server/persistence/_migrations/up_to_3.py
```diff
@@ -173,3 +190,95 @@ def _migrate_analysis_table(
...
+        # One worker per core of the OT-2's Raspberry Pi.
+        # We're compute-bound, so more workers would just thrash.
+        #
+        # Napkin math for the memory footprint:
+        # Suppose a very large protocol has ~10MB of commands (see e.g. RQA-443).
+        # The maximum number of runs at the time of writing is 20,
+        # so that's at most ~200MB total, which should be fine.
+        processes=4
+    ) as pool:
+        pool.map(
```
SyntaxColoring: Ideally, add logging here for when each run is started and completed, if we can get that to work with subprocesses.

robot-server/robot_server/persistence/_migrations/up_to_3.py
```diff
@@ -173,3 +190,95 @@ def _migrate_analysis_table(
...
+            _migrate_db_commands_for_run,
+            ((source_db_file, dest_db_file, run_id, lock) for run_id in run_ids),
+        )
+
+
+def _migrate_db_commands_for_run(
```
SyntaxColoring: Probably wrap this whole thing in a `try`/`except: pass`.

robot-server/robot_server/persistence/_migrations/up_to_3.py
sfoster1: > It turns out there is an utterly obscene amount of constant overhead on macOS. On my machine, each worker thread takes 1500 ms (!!!) to start. This is partially the platform's fault, and partially our fault.

Kind of a nit but the difference is important here: these are worker _processes_, not worker _threads_. This is where the difference comes in.

> The part that's our fault is: importing this code takes 1500 ms. I've confirmed that this is almost entirely from our `__init__.py`s re-exporting shit. Each of these subprocesses is effectively importing an entire copy of robot-server; they can't import anything without accidentally pulling in unrelated stuff like our FastAPI routers. We can save like 50% off the bat just by removing [this one re-export](https://github.com/Opentrons/opentrons/blob/4c4f49109a8e577b1da21599ebcbd439d455cc30/robot-server/robot_server/__init__.py#L6-L10), for example.
> 
> On robots, this isn't causing any _performance_ problems right now because the default for Linux is still `fork`. However, for the same sort of reasons `fork` isn't used on macOS, it will cause _correctness_ problems. 
> So we have to choose between the code being fast or being reliable. Or we need to put in a bit more work to do both at once. That would entail:
> 
>     1. Using `spawn` or `forkserver` instead of the platform-specific default.
> 
>     2. Making that fast enough by making our imports cheaper, and possibly using [`set_forkserver_preload()`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_forkserver_preload). Remember that 1500 ms on one of our laptops is more like [1 minute on an OT-2](https://opentrons.atlassian.net/browse/RSS-133).

We should do both, or at least the lowest-hanging-fruit of both. Specifically
- We should certainly remove that re-export you identified that drops sub-import times by 50%, and maybe profile it and see if there's a couple other big wins (big wins here to me would be semantically-irrelevant changes like changing import paths that give us 10% or more uplift)
- We should certainly specify `forkserver` for multiprocessing, and we should preload it. This is the way you do fast worker processes. It sucks that we have to do worker processes as opposed to worker threads, but That's Python (tm).
    - Note: we use `subprocess` elsewhere, and we might use `multiprocessing` more in the future. For that reason (and for general hygiene) we should use the for-some-reason-not-as-emphasized-as-it-should-be [multiprocessing.get_context](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context) for each semantically-contained area that uses multiprocessing so start strategies and preloads and such can be set independently. I believe that this api still has the defect that it doesn't close itself because it emulates the whole `multiprocessing` module, so we should (sadly) create the context manager in a python module that we conditionally import.

What do you think?

SyntaxColoring: Thanks, yeah, that sounds great if the `forkserver`+preload thing doesn't strike you as overly complicated.

> I believe that this api still has the defect that it doesn't close itself because it emulates the whole multiprocessing module, so we should (sadly) create the context manager in a python module that we conditionally import.

~~I'm not following this part. In my proof of concept, I'm doing [this](https://github.com/Opentrons/opentrons/blob/6c0763a740fe8725d66ab8ac162edb627bff5fc5/robot-server/robot_server/persistence/_migrations/up_to_3.py#L218-L243), following [this example](https://docs.python.org/3/library/multiprocessing.html#multiprocessing-start-methods:~:text=Alternatively%2C%20you%20can%20use%20get_context()%20to%20obtain%20a%20context%20object.%20Context%20objects%20have%20the%20same%20API%20as%20the%20multiprocessing%20module%2C%20and%20allow%20one%20to%20use%20multiple%20start%20methods%20in%20the%20same%20program.). Why would we need a conditional import? I don't think `get_context()` affects global state and I don't think its return value is a thing that needs to be closed—it's the things that you create *in* the context, like a `Pool`, that need to be closed. Right?~~

Edit: Discussed in-person. We think when you're using `forkserver`, there *is* something that needs to be cleaned up: the long-running process that the workers get forked off of. Possibly relevant: [bpo-37507](https://bugs.python.org/issue37507)

Edit 2: From [bpo-37507](https://bugs.python.org/issue37507) and [this discussion](https://mail.python.org/archives/list/python-dev@python.org/thread/NJCLS7IOKKO6W7LBGPQI2VFAO57EJEPW/#DIOMN256E24NUBSPKWFQA556F43IOWXR), I'm getting the sense that the `multiprocessing` authors, in general, are reluctant for subprocesses to be cleaned up explicitly. They want to clean them up for us implicitly in `__del__` or whatever. We disagree with that, but it doesn't seem like a productive thing for us to fight, so we'll keep it naive and follow [the pattern in the examples](https://docs.python.org/3/library/multiprocessing.html#multiprocessing-start-methods:~:text=Alternatively%2C%20you%20can%20use%20get_context()%20to%20obtain%20a%20context%20object.%20Context%20objects%20have%20the%20same%20API%20as%20the%20multiprocessing%20module%2C%20and%20allow%20one%20to%20use%20multiple%20start%20methods%20in%20the%20same%20program.).

SyntaxColoring: I've implemented `forkserver` in this PR, and the import speedups in #14480.

robot-server/robot_server/persistence/_database.py
SyntaxColoring, line 1: Changes in this file are to separate connecting to the database (`sqlalchemy.create_engine()`) from adding all the tables to the database (`metadata.create_all()`).

The new migration code has to close and reopen the database several times, and we don't want to re-add all the tables every time we do that. I think it would technically be harmless, but it felt confusing, and it spammed the logs.

robot-server/robot_server/persistence/_database.py
```diff
@@ -49,21 +42,11 @@ def create_schema_2_sql_engine(path: Path) -> sqlalchemy.engine.Engine:
...
-    """Create a SQL engine for a schema 3 database.
-
-    Unlike `create_schema_2_sql_engine()`, this assumes the database is already
-    at schema 3. Migration is done through other mechanisms.
-    """
-    sql_engine = sqlalchemy.create_engine(sql_utils.get_connection_url(path))
-
+@contextmanager
+def sql_engine_ctx(path: Path) -> Generator[sqlalchemy.engine.Engine, None, None]:
+    """Like `create_sql_engine()`, but clean up when done."""
```
SyntaxColoring, line 47: The only reason we haven't been defining this as a context manager this whole time is that it was difficult to integrate with FastAPI.

In [recent versions of FastAPI](https://fastapi.tiangolo.com/release-notes/#0930), that's fixed, yay! So at some point, we can switch everything to this and delete the non-context-manager version.

robot-server/robot_server/persistence/_tables/schema_2.py
```diff
@@ -105,7 +105,7 @@
     # column added in schema v1
     sqlalchemy.Column(
         "commands",
-        sqlalchemy.PickleType(pickler=legacy_pickle, protocol=PICKLE_PROTOCOL_VERSION),
+        sqlalchemy.LargeBinary,
```
SyntaxColoring, line 108: This `sqlalchemy`-only change (*not* affecting the actual data, or the SQL) is to let us unpickle the data without holding the database lock the whole time.

SyntaxColoring: Separately: I'm increasingly unsure of how to document historical schema stuff like this. Just seeing `LargeBinary` is definitely not helpful if you're a new contributor and you're trying to figure out how this old migration code works. But it's difficult to clarify this without just being like "see `RunStore`," and I don't know if that's much better.

CaseyBatten: Regarding what to do about documenting schema info/transitions as we implement the migration, a summary of the old command storage method describing the migration to our new storage method exists elsewhere, yes? I would recommend directing contributors there from here, and if need be add details regarding the migration in that singular location.

CaseyBatten (approved):
Overall looks solid, the performance improvements will be very advantageous. The `forkserver` adaptation is appropriate as well.

robot-server/robot_server/persistence/_tables/schema_2.py
SyntaxColoring: > a summary of the old command storage method describing the migration to our new storage method exists elsewhere, yes?

[Yeah, although it's not super detailed.](https://github.com/Opentrons/opentrons/blob/e20f49446bbc55cce9433f4b3ac541fd81841cb3/robot-server/robot_server/persistence/_migrations/up_to_3.py#L9-L13)

> I would recommend directing contributors there from here, and if need be add details regarding the migration in that singular location.

Hmm. Not the other way around?

robot-server/robot_server/persistence/_migrations/up_to_3.py
SyntaxColoring: Not done in this PR because subprocess logging looks a little tricky.

PR #14466, opened 2024-02-09
test(robot-server): Check the number of returned commands in persistence snapshot tests

SyntaxColoring:
# Overview

robot-server's persistence snapshot tests currently only check that the server can return *some number of* run commands. They don't check that the server returns the *correct number,* or the *correct contents.* That was good enough before, but it's dicey now because we're doing a bunch of database migration stuff in this release that deals with commands. (e.g. #14348,  #14465.)

This PR asserts that the server returns the correct number of commands. Because of time constraints, we're still not attempting to check that the *contents* are correct—that's RSS-213.

# Test Plan

I've run these tests on a commit before all of v7.2.0's database changes, to be more certain that we're locking in good numbers instead of bad numbers.

# Changelog

* For each run in each snapshot, hard-code the expected number of commands.
* Fix a funny test bug where it was fetching commands redundantly, because the code to do that was accidentally nested inside an unrelated loop.

# Review requests

None in particular.

# Risk assessment

Very low.

### DIFF
--- a/robot-server/tests/integration/http_api/persistence/test_compatibility.py
+++ b/robot-server/tests/integration/http_api/persistence/test_compatibility.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from shutil import copytree
 from tempfile import TemporaryDirectory
-from typing import List, NamedTuple
+from typing import List
 
 import anyio
 import pytest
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Yeah, that looks like a good idea, thanks!

PR #14467, opened 2024-02-11
docs: update docs to reflect node v18 is current

strangemonad:
# Overview

There's already a .nvmrc file but if someone is just manually following the docs, there are references to node v16 and then either `make setup` will fail if you forced the current node version to be v16 or the setup docs will be confusing.

# Test Plan
Ran locally with both `nvm` and `nvs` in both auto mode and manual mode
- nvs auto changes to v18 when I cd into the dirrectory
- `nvm install` and `nvm use` also work as expected.

# Changelog
- Updated dev setup docs to reflect that the project depends on Node v18

# Review requests

n/a

# Risk assessment
n/a

### DIFF
--- a/DEV_SETUP.md
+++ b/DEV_SETUP.md
@@ -13,7 +13,7 @@ You will need the following tools installed to develop on the Opentrons platform
 - curl
 - ssh
 - Python v3.10
-- Node.js v16
+- Node.js v18
 
 ### macOS
 
@@ -82,10 +82,9 @@ Close and re-open your terminal to confirm `nvs` is installed.
 nvs --version
 ```
 
-Now we can use nvs to install Node.js v16 and switch on `auto` mode, which will make sure Node.js v16 is used any time we're in the `opentrons` project directory.
+Now we can use `nvs` to install the currently required Node.js version set in `.nvmrc`. We think it's easiest if you switch on `auto` mode which will make sure the correct version of Node.js is used any time we're in the `opentrons` project directory. ^[If you don't want to use `nvs` auto mode you'll have to manually run `use` or `install` when you want to work on the project.]
 
 ```shell
-nvs add 16
 nvs auto on
 ```
 
@@ -202,7 +201,7 @@ Once you are inside the repository for the first time, you should do two things:
 3. Run `python --version` to confirm your chosen version. If you get the incorrect version and you're using an Apple silicon Mac, try running `eval "$(pyenv init --path)"` and then `pyenv local 3.10.13`. Then check `python --version` again.
 
 ```shell
-# confirm Node v16
+# confirm Node v18
 node --version
 
 # set Python version, and confirm


koji:
Hi @strangemonad, thank you for opening a PR!

koji (commented):
LGTM (node version parts)

@ecormany, could you check this PR?

DEV_SETUP.md
```diff
@@ -82,10 +82,9 @@ Close and re-open your terminal to confirm `nvs` is installed.
 nvs --version
 ```
 
-Now we can use nvs to install Node.js v16 and switch on `auto` mode, which will make sure Node.js v16 is used any time we're in the `opentrons` project directory.
+Now we can use `nvs` to install the currently required Node.js version set in `.nvmrc`. We think it's easiest if you switch on `auto` mode which will make sure the correct version of Node.js is used any time we're in the `opentrons` project directory. ^[If you don't want to use `nvs` auto mode you'll have to manually run `use` or `install` when you want to work on the project.]
```
ecormany, line 85:
```suggestion
Now we can use `nvs` to install the currently required Node.js version set in `.nvmrc`. We enable `auto` mode to use the correct version of Node.js any time we're in the `opentrons` project directory. Without `auto` mode, we would have to manually run `use` or `install` each time we work on the project.
```
Markdown footnotes don't render inline in GitHub, which is where most users will read this. So I would remove that syntax. Plus some light prose edits.

ecormany (changes requested):
Thanks for this addition! I made some minor suggestions, mostly to make sure that the text renders well on GitHub.

DEV_SETUP.md
koji: I think we'd like to remove "mode" from use `nvs` auto mode since it is a command or switching the word from mode to command would be better.

https://github.com/jasongin/nvs?tab=readme-ov-file#command-reference

ecormany: How about:
```
The `auto` command selects the correct version of Node.js any time we're in the `opentrons` project directory.
Without `auto`, we would have to manually run `use` or `install` each time we work on the project.
```

koji: I think `the correct version` -> `the long-term support (LTS) version` would be accurate.
what do you think about that, @ecormany ?

ecormany: I used "correct" just to mean "it will work" 😁 

I don't think we have to be more precise here. (And I don't know enough about how `nvs` behaves to be more precise.)

koji: For nvs/nvm, we set 18 in `.nvmrc` so if someone uses `nvs auto` or `nvm use`, nvs/nvm will install LTS version of nodejs. Right now it is v 18.19.0.

strangemonad: @ecormany @koji any final thoughts on this as to the final version of the wording? I don't have strong feelings so happy to follow whatever you settle on as best

koji: Hi @strangemonad 

If you adopted the following Ed's suggestion, that would be great. 
```
The `auto` command selects the correct version of Node.js any time we're in the `opentrons` project directory.
Without `auto`, we would have to manually run `use` or `install` each time we work on the project.
```

koji:
@strangemonad 
I will close this PR since the PR hasn't been updated for 3 weeks.
We love your addition, and we'd like to add that as soon as possible.
I opened a new PR to follow up this PR.

We really appreciate your contribution!

PR #14470, opened 2024-02-11
fix: allow dev Dockerfile to build

strangemonad:
# Overview

There were 2 issues that prevented this dockerfile from building: 1) The base ubuntu image doesn't have git installed. This is required by `./scripts/python_build_utils.py` when normalizing version strings in setup.py 2) robot-server depends on server-utils but that was never copied or installed in the dockerfile

# Test Plan

Verified that a standalone docker build works

```
docker buildx build -t ot-dev .
```

Verified that the dev robot stack can be brought up and connected to 127.0.0.1 with the app.

```
docker compose up
```

# Changelog
- Fixed dev robot stack Dockerfile to include git in the builder stage and install server-utils

# Risk assessment

Unclear. This isn't used in any of the github actions. Since it was broken and `server-utils` has existed for a while, it's unclear if anyone is using the local docker dev workflow

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,16 @@
 FROM ubuntu as base
 ENV TZ=Etc/UTC
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
-RUN apt-get update && apt-get install --yes python3 pip pkg-config libsystemd-dev
+RUN apt-get update && apt-get install --yes python3 pip pkg-config libsystemd-dev git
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me. You're right that nobody here really uses it but it should still work.

strangemonad:
@sfoster1 sweet, thanks for merging. on our end, being able to build a docker image for a release tag and pushing that to our internal docker registry will be the easier way to run our integration tests rather than having to build from source.

PR #14474, opened 2024-02-12
 fix(app): remove [Object, object] tooltip from appearing on hover

jerader:
closes RQA-2323

# Overview

This was a tricky bug, I narrowed down the issue occurring with the `LegacyModal` and there must have been some sort of conflict with passing down the props, so changing it from `...props` (passing down all props even the ones already written out) to `...styleProps` (passing down only styled components props) ensured that only the style props were being passed down. I also fixed a i18n bug i introduced in the modal.

Before:

https://github.com/Opentrons/opentrons/assets/66035149/f9dbadbb-cf52-4658-aa1d-1a934d42f0d3

After:

https://github.com/Opentrons/opentrons/assets/66035149/11fccfc4-6204-4755-9f4a-1cd039fd9075

# Test Plan

See the videos but also test for yourself if you want. An easy way to test is to change the `seen` boolean logic from `!seen` to `seen`.

# Changelog

- fix i18n bug with the `Trans` component
- change `LegacyModal`'s spread operator props.

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/molecules/LegacyModal/index.tsx
+++ b/app/src/molecules/LegacyModal/index.tsx
@@ -68,7 +68,7 @@ export const LegacyModal = (props: LegacyModalProps): JSX.Element => {
       onOutsideClick={closeOnOutsideClick ?? false ? onClose : undefined}
       // center within viewport aside from nav
       marginLeft={styleProps.marginLeft ?? '7.125rem'}
-      {...props}
+      {...styleProps}
       footer={footer}
     >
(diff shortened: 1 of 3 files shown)


mjhuff (approved):
Nice! I can see how this was the issue. It looks like this doesn't interfere with how `LegacyShellModal` uses its custom `LegacyModalShellProps`, given how these are declared separately. 

Thanks for fixing this (and the translation string)!

PR #14475, opened 2024-02-12
fix(api): OT-2 trash bins in 2.16 get added to engine at the start of protocol

jbleon95:
# Overview

This PR refactors #14433 to fix a known inconsistency it introduced.

When the new `TrashBin` and `WasteChute` objects were introduced in api level 2.16 in version 7.1.0 of the robot, there was a peculiarity of how they were loaded in PAPI versus when they would actually appear in the engine during analysis. Because of the way the protocol engine commands were designed, they'd only be recognized when first used i.e. whenever the pipette would first move to the trash for a drop tip, dispense, or blow out. This caused analysis caught errors to display wrong line numbers and made relying on the addressable area store in engine during analysis more difficult.

To fix all this, in #14325 a protocol action was added to allow the addressable areas representing the trash bin and waste chute to be immediately added to engine on a load. This worked for Flex protocols, but because of a threading issue caused OT-2 protocols to hang during analysis. To fix this issue, #14433 reverted OT-2s to load the fixed trash addressable area on first reference, not at the start of the protocol.

This PR fixes that by refactoring the protocol core methods that add trash bins and waste chutes to the engine and internal list of disposal locations. Now there are two methods, one that just adds a disposal location (a `Labware`, `TrashBin` or `WasteChute` to the engine core's `self._disposal_locations` list and another one that does that _and_ adds and validates the area, only for the latter two objects. Now for OT-2s, in the `ProtocolContext.__init__` method the OT-2 fixed trash `TrashBin` is only appended to the core list. In order to fix the inconsistency, in `create_protocol_context` the protocol engine method is called directly to add that trash.

# Test Plan

Re-using test protocols from previous PRs, this protocol passes analysis/simulate without deadlocking
```
requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.16"
}


def run(context):
    pass
```

These protocols correctly fail analysis
```
metadata = {
    'protocolName': 'Thermocycler conflict 1',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}


def run(context):
    thermocycler = context.load_module("thermocyclerModuleV2")
    trash = context.load_trash_bin('A1')
```
```
metadata = {
    'protocolName': 'Heater-shaker conflict OT-2',
}

requirements = {
    "robotType": "OT-2",
    "apiLevel": "2.16"
}


def run(context):
    # Both these lines should raise
    heater_shaker = context.load_module("heaterShakerModuleV1", '11')
    heater_shaker_2 = context.load_module("heaterShakerModuleV1", '9')
```

# Changelog

- removes `skip_add_to_engine` boolean flag for `append_disposal_location` and remove adding it to engine in that method
- add core method `add_disposal_location_to_engine` to add and verify against the engine/deck conflict for new trash objects
- add a `should_load_fixed_trash_area_for_python_protocol` method to see if we should add the area in `create_protocol_context` and to make sure it stays in sync with `ProtocolContext.__init__`

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/create_protocol_context.py
+++ b/api/src/opentrons/protocol_api/create_protocol_context.py
@@ -15,10 +15,14 @@ from opentrons.util.broker import Broker
 from opentrons.protocol_engine import ProtocolEngine
 from opentrons.protocol_engine.clients import SyncClient, ChildThreadTransport
 from opentrons.protocols.api_support.types import APIVersion
+from opentrons.protocols.api_support.deck_type import (
+    should_load_fixed_trash_area_for_python_protocol,
+)
 from opentrons.protocols.api_support.definitions import MAX_SUPPORTED_VERSION
 
 from .protocol_context import ProtocolContext
 from .deck import Deck
+from ._trash_bin import TrashBin
 
 from .core.common import ProtocolCore as AbstractProtocolCore
 from .core.legacy.deck import Deck as LegacyDeck
@@ -148,7 +152,7 @@ def create_protocol_context(
     # this swap may happen once `ctx.move_labware` off-deck is implemented
     deck = None if isinstance(core, ProtocolCore) else cast(Deck, core.get_deck())
 
-    return ProtocolContext(
+    context = ProtocolContext(
         api_version=api_version,
         # TODO(mm, 2023-05-11): This cast shouldn't be necessary.
         # Fix this by making the appropriate TypeVars covariant?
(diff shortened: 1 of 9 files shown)


api/src/opentrons/protocol_api/create_protocol_context.py
```diff
@@ -158,3 +162,17 @@ def create_protocol_context(
         deck=deck,
         bundled_data=bundled_data,
     )
+    # If we're loading an engine based core into the context, and we're on api level 2.16 or above, and on an OT-2
+    # we need to insert a fixed trash addressable area into the protocol engine for correctness in anything that relies
```
TamarZanzouri: :+1:

TamarZanzouri (approved):
Changes makes sense to me!! thank you for the comments they helped! if tested its good to merge imo

api/src/opentrons/protocol_api/create_protocol_context.py
```diff
@@ -158,3 +162,17 @@ def create_protocol_context(
...
+    # on knowing what addressable areas have been loaded (and any checks involving trash geometry). Because the method
+    # that uses this in the core relies on the sync client
```
SyntaxColoring: Incomplete sentence?

SyntaxColoring (approved):
Thank you!

sanni-t (approved):
Nice!!

PR #14476, opened 2024-02-12
refactor(app): refactor protocol event analytics hooks

ncdiehl11:
# Overview

Recent implementation of analytics hooks for protocol events broke a rule of hooks, resulting in broken functionality when trying to proceed to protocol setup. Here, I refactor the use of 'useRobot' into 'useTrackProtocolRunEvent' to follow the rule of hooks and avoid the use of 'useRobot' in the parsing utility function 'parseProtocolRunAnalyticsData'.

# Review requests

@mjhuff per finding the bug

# Risk assessment

low


mjhuff (approved):
Code looks and good and fix works as expected!

PR #14477, opened 2024-02-12
fix(app-shell, app-shell-odd): Fix intermittently dropped notifications

mjhuff:
Closes [RQA-2339](https://opentrons.atlassian.net/browse/RQA-2339), [RQA-2321](https://opentrons.atlassian.net/browse/RQA-2321), [RQA-2319](https://opentrons.atlassian.net/browse/RQA-2319), [RAUT-962](https://opentrons.atlassian.net/browse/RAUT-962)

# Overview

Our subscribe/unsubscribe logic is directly tied to the component lifecycle. If a component dismounts and unsubscribes to a given topic while a new component mounts and subscribes to the same topic, a race condition occurs in which the shell layer often thinks the subscribing component is subscribed when it is in fact not. This race condition occurs when switching pages that contain components on each page interested in the same topic. 

The most straightforward fix is to set a small timeout any time an unsubscribe request goes through to accommodate any additional rendering that might occur. After the timeout, then check to see if the connection manager really wants to unsubscribe. 

I think this solution makes the most sense, but I'm open to alternatives. 

Some things I considered:
1) We could in theory never unsubscribe to topics, but that means the MQTT broker holds more subscriptions in memory until the robot is power cycled (or the broker is restarted). 
2) Alternatively, we could somehow not handle subscriptions within components. This sounds great in theory, but our routes are not always entirely static (ex, we subscribe to runs/abc123, which we may not need 5 minutes from now). 

# Test Plan
- Push this branch to the ODD.
- On the desktop app, navigate to the DeviceDetails page.
- Launch a protocol. The desktop app should display the run status in the upper right corner.
- On the ODD, pause/play a protocol repeatedly. The desktop app run status should update. 
- The ODD buttons should update as expected. 
- Cancelling a protocol on the ODD should cause the "cancelling" modal to eventually unrender. 

# Changelog
- Fixed an issue in which the app and ODD did not correctly reflect the current protocol run status.

# Risk assessment
low

[RQA-2339]: https://opentrons.atlassian.net/browse/RQA-2339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RQA-2321]: https://opentrons.atlassian.net/browse/RQA-2321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RQA-2319]: https://opentrons.atlassian.net/browse/RQA-2319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RAUT-962]: https://opentrons.atlassian.net/browse/RAUT-962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RQA-2346]: https://opentrons.atlassian.net/browse/RQA-2346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -184,30 +184,38 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
   }
 }
 
+// Because subscription logic is directly tied to the component lifecycle, it is possible
+// for a component to trigger an unsubscribe event on dismount while a new component mounts and
+// triggers a subscribe event. For the connection store and MQTT to reflect correct topic subscriptions,
+// do not unsubscribe and close connections before newly mounted components have had time to update the connection store.
+const RENDER_TIMEOUT = 250
+
 function unsubscribe(notifyParams: NotifyParams): Promise<void> {
   const { hostname, topic } = notifyParams
   return new Promise<void>(() => {
     if (hostname in connectionStore) {
-      const { client } = connectionStore[hostname]
-      const subscriptions = connectionStore[hostname]?.subscriptions
-      const isLastSubscription = subscriptions[topic] <= 1
+      setTimeout(() => {
+        const { client } = connectionStore[hostname]
+        const subscriptions = connectionStore[hostname]?.subscriptions
+        const isLastSubscription = subscriptions[topic] <= 1
 
-      if (isLastSubscription) {
-        client?.unsubscribe(topic, {}, (error, result) => {
-          if (error != null) {
-            log.warn(
-              `Failed to unsubscribe on ${hostname} from topic: ${topic}`
-            )
-          } else {
-            log.info(
-              `Successfully unsubscribed on ${hostname} from topic: ${topic}`
-            )
-            handleDecrementSubscriptionCount(hostname, topic)
-          }
-        })
-      } else {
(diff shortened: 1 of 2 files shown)


mjhuff:
See https://opentrons.atlassian.net/browse/RAUT-959 for discussion on the better way to do this. Changes involving base connectivity won't make it into the 7.2 release but will be prioritized for future MQTT work. Once that's done, moving unsubscribe logic to the server side (per @sfoster1 's advice) will be do-able.

ncdiehl11 (approved):
tested, looks good

PR #14478, opened 2024-02-12
fix(api): do not include HEPA-UV in motor nodes

ahiuchingau:
# Overview
We should exclude hepa uv and its bootloader node from the motor nodes, so we don't get a timeout response fetching motor position from them.

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3utils.py
+++ b/api/src/opentrons/hardware_control/backends/ot3utils.py
@@ -351,13 +351,13 @@ def motor_nodes(devices: Set[FirmwareTarget]) -> Set[NodeId]:
         NodeId.head_bootloader,
         NodeId.gripper_bootloader,
     }
-    hepa_nv_nodes = {
+    hepa_uv_nodes = {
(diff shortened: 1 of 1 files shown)


api/src/opentrons/hardware_control/backends/ot3utils.py
```diff
@@ -351,8 +351,13 @@ def motor_nodes(devices: Set[FirmwareTarget]) -> Set[NodeId]:
         NodeId.head_bootloader,
         NodeId.gripper_bootloader,
     }
+    hepa_nv_nodes = {
```
vegano1: typo?
hepa_uv_nodes

vegano1 (approved):
oh yeah, nice catch
Looks good, just left a small comment.

sfoster1 (approved):
Looks good, let's fix the typo tho

api/src/opentrons/hardware_control/backends/ot3utils.py
ahiuchingau: lol yep u != n, who would have thought

PR #14479, opened 2024-02-12
feat(hardware, api): check motor engaged status

ahiuchingau:
# Overview

Up until now, the hardware controller doesn't actually know if a motor has been engaged. We've been assuming a motor has been disengaged by checking the stepper position status, which doesn't provide the full story and causes us to engage the motor when it's already enabled (while homing the 96-channel pipette mount). 

This PR enables the hardware controller to prompt for the motor enabled status and allows us to finally implement the `OT3controller.engaged_axes` method.

This PR also updated the phony axis bound to values that are closer to reality; this prevents us from getting stuck in a extremely long homing move (like > 33 minutes) that never times out if a limit switch malfunctions.

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -86,6 +86,8 @@ from opentrons_hardware.hardware_control.motor_enable_disable import (
     set_disable_motor,
     set_enable_tip_motor,
     set_disable_tip_motor,
+    get_motor_enabled,
+    get_tip_motor_enabled,
 )
 from opentrons_hardware.hardware_control.motor_position_status import (
     get_motor_position,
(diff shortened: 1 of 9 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1151,37 +1155,59 @@ async def watch(self, loop: asyncio.AbstractEventLoop) -> None:
...
         return {
-            Axis.Z_L: phony_bounds,
-            Axis.Z_R: phony_bounds,
-            Axis.P_L: phony_bounds,
-            Axis.P_R: phony_bounds,
-            Axis.X: phony_bounds,
-            Axis.Y: phony_bounds,
-            Axis.Z_G: phony_bounds,
-            Axis.Q: phony_bounds,
+            Axis.Z_L: (0, 300),
```
sfoster1, line 1164: how close to reality are these bounds?

sfoster1 (changes requested):
Let's get some tests at least for the `opentrons_hardware` side. I also wonder if `get_motor_status` is maybe better implemented with a broadcast message so we don't need so many canbus back and forths

sfoster1 (approved):
LGTM!

sfoster1 (approved):
LGTM if it works!

PR #14480, opened 2024-02-12
perf(robot-server): Remove slow re-exports from __init__.py and protocols/__init__.py

SyntaxColoring:
# Overview

This rearranges some of robot-server's internal `import` statements to make it significantly faster to launch certain subprocesses. See https://github.com/Opentrons/opentrons/pull/14465#discussion_r1484989089 for background. This goes towards RSS-451.

This is paired with a PR in oe-core: https://github.com/Opentrons/oe-core/pull/134

# Benchmarks

This brings the startup time of #14465's worker processes down from **~74 sec to ~31 sec**. Tested on an OT-2 with this command:

```
python -m timeit -r1 -n1 'import robot_server.persistence._migrations._up_to_3_helper; import robot_server.persistence.legacy_pickle; robot_server.persistence.legacy_pickle._get_legacy_ot_types()'
```

(The call to `_get_legacy_ot_types()` is to make sure we account for the cost of its dynamic import. `_up_to_3_helper` is a thing from my work in progress for #14465 and it may not make it into the final PR, so don't be surprised if you can't import it; but it's basically just the code for that PR's worker process, extracted to a standalone file.)

# Explanation

In Python, if you have a tree like this:

```
some_package/
    __init__.py
        # re-export .module_1.foo, module_2.bar, module_3.baz, etc.
    module_1.py
    module_2.py
    module_3.py
```

And then you do this:

```python
import some_package.module_1
```

The Python interpreter will evaluate `module_1`, `module_2`, *and* `module_3`, because they were mentioned in `some_package/__init__.py`. When you factor in all the subdependencies of `module_2` and `module_3`, this can be very very slow.

In the case of #14465, worker processes for database migration code were wasting nearly 45 seconds importing things like FastAPI routers, which are irrelevant to them.

# Review requests

We will definitely regress on this performance improvement unless we deliberately guard against it.

I propose, going forward, we adopt a house rule to "very rarely" re-export things from `__init__.py` files. Acceptable re-exports are ones like [`opentrons.protocol_api.<...>`](https://github.com/Opentrons/opentrons/blob/c142da102967ea7d2f6fb13c5d72b754fdc560cb/api/src/opentrons/protocol_api/__init__.py#L39-L63), where we have a strict public interface that we want to be ergonomic for users. All other re-exports are questionable.

Then, without those re-exports, we need a way to delineate private stuff from public stuff. I propose we lean harder into prefixing the filenames with underscores. We've already done this in a few places, but it's been pretty ad-hoc so far. Underscore-prefixed filenames are a common, but not universal, Python convention. (Examples: [h11](https://github.com/python-hyper/h11/tree/master/h11), [trio](https://github.com/python-trio/trio/tree/b65afe1686e915071ac3c67ae7dbb35236736a84/src/trio), [requests](https://github.com/psf/requests/blob/4f3f189d6b7d2a19ff73dfb4a3e39713d4abe01a/src/requests/_internal_utils.py); counterexamples: [sqlalchemy](https://github.com/sqlalchemy/sqlalchemy/tree/62b3ca476e879bacb20bb0c520c7c91feca576c0/lib/sqlalchemy), [pydantic](https://github.com/pydantic/pydantic/tree/main/pydantic).)

Does all of that sound good?

Then, we could take this a lot further. `opentrons.protocol_engine`'s imports, for example, are pretty heavy. You can't just import command models without also pulling in all of the execution logic, and all of the dependencies of that execution logic, including the hardware API, etc. etc.

# Test plan

* [x] `make dev` still works
* [x] Still boots on an OT-2
* [x] Still boots on a Flex
    * Remember this needs a build with https://github.com/Opentrons/oe-core/pull/134

# Risk assessment

Pretty low.

### DIFF
--- a/robot-server/Makefile
+++ b/robot-server/Makefile
@@ -71,7 +71,7 @@ clean_all_cmd = $(clean_cmd) dist
 #     probably POSIX-only.
 dev_port ?= "31950"
 dev_host ?= "localhost"
-run_dev ?= uvicorn "robot_server:app" --host $(dev_host) --port $(dev_port) --ws wsproto --lifespan on --reload
+run_dev ?= uvicorn "robot_server.app:app" --host $(dev_host) --port $(dev_port) --ws wsproto --lifespan on --reload
 
 .PHONY: all
 all: clean sdist wheel
--- a/robot-server/opentrons-robot-server.service
+++ b/robot-server/opentrons-robot-server.service
@@ -18,7 +18,7 @@ Type=notify
 # /run/aiohttp.sock matches where our reverse proxy expects to find us.
 # It refers to aiohttp even though this server doesn't use that framework
 # for historical reasons.
-ExecStart=uvicorn robot_server:app --uds /run/aiohttp.sock --ws wsproto --lifespan on
+ExecStart=uvicorn robot_server.app:app --uds /run/aiohttp.sock --ws wsproto --lifespan on
 
 Environment=OT_SMOOTHIE_ID=AMA
 Environment=RUNNING_ON_PI=true
--- a/robot-server/robot_server/__init__.py
+++ b/robot-server/robot_server/__init__.py
@@ -2,9 +2,3 @@
 
 This server provides the main control interface for an Opentrons robot.
 """
-
-from .app_setup import app
-
-__all__ = [
-    "app",
-]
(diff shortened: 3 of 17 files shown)


sfoster1 (approved):
Definitely looks good, huge uplift. I totally agree with the house style concept.

sfoster1:
flex build here: https://github.com/Opentrons/oe-core/actions/runs/7893890421

sfoster1:
Oops I mean THIS build (got the refs swapped): https://github.com/Opentrons/oe-core/actions/runs/7903462058

sfoster1:
And here's an ot-2 one: https://github.com/Opentrons/buildroot/actions/runs/7903479028

PR #14482, opened 2024-02-12
fix(app-shell): detect usb devices after update

sfoster1:
When we updated to electron 27, our old package for detecting devices didn't work anymore. Instead we now use the npm package `usb`, which is a reimplementation of node-usb that also provides a WebUSB frontend.

The integration here had a couple small problems; the webusb just straight up doesn't seem to work with these devices, and we were doing the callback glue slightly wrong.

With both those problems fixed, we should once again detected device add and removal basically instantly, which fixes some issues with robot detection and removal.

Closes RQA-2324

### DIFF
--- a/app-shell/src/system-info/usb-devices.ts
+++ b/app-shell/src/system-info/usb-devices.ts
@@ -19,23 +19,30 @@ export interface UsbDeviceMonitor {
 
 const log = createLogger('usb-devices')
 
+const decToHex = (number: number): string =>
+  number.toString(16).toUpperCase().padStart(4, '0')
+const idVendor = (device: usb.Device): string =>
+  decToHex(device.deviceDescriptor.idVendor)
+const idProduct = (device: usb.Device): string =>
+  decToHex(device.deviceDescriptor.idProduct)
+
 const descriptorToDevice = (
   descriptors: usb.Device,
   manufacturerName?: string,
   serialNumber?: string,
-  productName?: string
+  productName?: string,
+  systemIdentifier?: string
 ): UsbDevice => ({
   vendorId: descriptors.deviceDescriptor.idVendor,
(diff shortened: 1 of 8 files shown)


app-shell/src/system-info/usb-devices.ts
```diff
@@ -12,30 +12,142 @@ export type UsbDeviceMonitorOptions = Partial<{
...
+          }
+        ),
+      ])
+    )
+
+    .then(([manufacturer, serialNumber, productName]) => {
+      return descriptorToDevice(device, manufacturer, serialNumber, productName)
+    })
+    .finally(() => {
+      device.close()
```
brenthagen: i'm getting `[shell] 15:29:48.027 [usb-devices] error: Failed to format added device vid=5ac pid=5ac for upstream: Error: Can't close device with a pending request` from this close call on device add, trying to see why. I'm assuming we need to close the device here?

brenthagen: putting the close call on a 1 sec timeout seems to work but...why is it necessary

sfoster1: I assume so yeah, definitely a bug in how we call these promises

sfoster1:
Couple tests left to pass but @brenthagen i think that issue is fixed. right now the only problem appears to be that we can't reliably pull mfr and product name on windows at least for composites

sfoster1:
i'm worried that might be this: https://github.com/libusb/libusb/commit/fdab67b14ac32891158dc210636b622c42e24e22 which is fixed in libusb 0.27.0 which is of course one later version than what node-usb vendors

brenthagen (approved):
talked through with @sfoster1 , checked device add/remove at least on the macOS side without issue (including Flex)

PR #14485, opened 2024-02-13
feat(protocol-designer): update React DnD from version 6.0.0 to 16.0.1

jerader:
close RAUT-964

# Overview

This PR updates React-DnD from version `6` to `16.0.1` and it updates the react dnd backend from `mouse-backend` to `HTML5-backend` (this backend has been updated recently, since mouse-backend was last update 5 years ago 💀 )

NOTE: i didn't add test coverage in this because once the vite migration is done, we will be using vitest - figured that is a better time to add test coverage

# Test Plan

Test the drag and drop features which include:
1. dragging/dropping labware to different slots on the deck, check dragging/dropping on top of modules, adapters and on the deck, and swapping labware
2. dragging/dropping to reorder the step items. The easiest way to do this is import a protocol fixture and drag/drop the items to rearrange the steps. Try this with `doItAllV8` or something! Also test deleting & duplicating a step -> which you can access through right clicking on the step item and seeing the "delete" option in the menu that appears.

Please call out anything strange you might see. I tried to smoke test all the options as thoroughly as possible.

# Changelog

- update react DnD and backend version in package.json
- rename robotCoordinateWorkSpace component to something that includes the ref since it no longer needs the DOM coords
- update each component to using the react dnd hooks 
- remove all now unnecessary `@ts-expect-errors` due to the react dnd version

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/DeckSetup/LabwareOverlays/AdapterControls.tsx
```diff
@@ -15,66 +15,108 @@ import {
...
+            draggedDef.parameters.loadName
+          ) || isCustomLabware
+        )
+      }
+      return true
+    },
+    drop: (item: DroppedItem) => {
+      const droppedLabware = item
+      if (newSlot != null) {
+        dispatch(moveDeckItem(newSlot, labwareId))
```
jerader, line 92: the reason why i am checking for `newSlot` here (also in `SlotControls`) is because I can't get `item` from `useDrag` to update to the latest state. If you drag and drop the labware initially, it doesn't update to the new slot in `item` - this was causing a bug where you could only drag and drop the labware once. I tried to find documentation about how to fix this but i couldn't find anything so this was my workaround. Any refactor suggestions are welcome!

protocol-designer/src/components/DeckSetup/index.tsx
```diff
@@ -486,7 +478,6 @@ export const DeckSetupContents = (props: ContentsProps): JSX.Element => {
           </React.Fragment>
         )
       })}
-      <DragPreview getRobotCoordsFromDOMCoords={getRobotCoordsFromDOMCoords} />
```
jerader, line 489: When i updated from mouse backend to html5 backend, the DOM coordinates changed and resulted in the drag preview being mirrored. I tried to fix it but couldn't quite get it right because the `EditLabware` was being shown in replace of the cursor and it was being layered on top of the drag preview that was not in the correct place. so I opted to delete it. I showed the current visuals to Felix and he said they are ok! the deck map is getting a whole redesign this year (for 8.2) so this change will be only for the upcoming PD release (8.1).

mjhuff (approved):
Smoke tested without any issues or strange behavior!

Bummer about the `item` from `useDrag` not updating, but the `newSlot` approach is easy enough to follow. LGTM! Thanks for spending the time to update 10 versions 🥳

PR #14491, opened 2024-02-14
 fix(api): associate thermocycler with all slots it occupies

sanni-t:
Addresses RQA-2311

# Overview

This PR fixes a long-standing issue that the thermocycler is associated only with the slot it was loaded, even though it occupies more than one slot on both OT-2 & Flex. This marks the first step in making our backend aware of multiple slots used by any module.

The observable change of this PR is that `protocol_context.deck` in API 2.14 & above now returns the thermocycler module for items of slot 8, 10, 11 on OT-2 & A1 on Flex, when there is a thermocycler loaded in slot 7/ B1

Currently non-observable changes (due to practical use cases and non-public API features):
- when moving a 96-channel pipette to slot A2 in a way that the pipette partially moves over thermocycler in slot A1, the deck conflict checker will do conflict checking with thermocycler. Practically, since the thermocycler sits under the deck on the Flex, we will most likely run into an out-of-bounds issue before the 96 ch nozzles collide with the thermocycler.
- when moving an 8-channel in partial config on an OT-2 with thermocycler, moving to a slot such that the pipette is at least partially over 5/8/11 will raise a deck conflict error. Currently we don't have public API available to partially configure an 8-channel so this won't happen to existing protocols.

# Test Plan

The changes in this PR can be tested with analysis. Change API levels to make sure analysis passes for all versions >= 2.14.

```
requirements = {
	"robotType": "OT-2",
	"apiLevel": "2.14",
}

	thermocycler        = protocol.load_module('thermocycler module gen2')
	
	assert protocol.loaded_modules == {7: thermocycler}
	assert protocol.deck["7"] ==  thermocycler
	assert protocol.deck["8"] == thermocycler
	assert protocol.deck["10"] == thermocycler
	assert protocol.deck["11"] == thermocycler

```

- Protocol analysis snapshot tests would be very useful for this.

- Other than that, we just need to make sure that protocols run fine and as usual with these changes.

# Changelog

- module state store now also stores any additional slots occupied by a module
- `state.geometry.get_slot_item(slot)` now returns the module that has the given slot in its 'additional slots occupied' list
- `state.geometry.get_highest_z_in_slot` now returns height of module occupying the slot (regardless of if it was loaded in it)

# Review requests

- Is it okay to not put this behind a new version change? 
  - we discussed on slack and the consensus is that this fixes a bug and hence, should be applied to all versions. It won't be applied to older, non-engine protocols, though.
- Are we okay merging this into release? Or should this go into edge for thorough testing?

# Risk assessment

Low-medium. 
For suggested use of our API, this changes actually fixes a long-standing bug and is an improvement over the current API at a low risk. But for any protocols that might have been using loop-holes in the deck class, this change would break the behavior.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -153,7 +153,11 @@ class GeometryView:
     def get_highest_z_in_slot(
         self, slot: Union[DeckSlotLocation, StagingSlotLocation]
     ) -> float:
-        """Get the highest Z-point of all items stacked in the given deck slot."""
+        """Get the highest Z-point of all items stacked in the given deck slot.
+
+        This height includes the height of any module that occupies the given slot
+        even if it wasn't loaded in that slot (e.g., thermocycler).
+        """
         slot_item = self.get_slot_item(slot.slotName)
         if isinstance(slot_item, LoadedModule):
             # get height of module + all labware on it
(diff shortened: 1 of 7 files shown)


CaseyBatten (approved):
LGTM, Let's do it!

PR #14496, opened 2024-02-14
chore: default all py versions to robot-stack

sfoster1:
We used to have different default projects (which happen, it turns out, in the deploy environment) for sdists and wheels and stuff because sdists only existed for flex dev and we hadn't released it yet. Now we have, and we should always be using our robot-stack project unless ot_project is set to something else.

Keep the other defaults around but make them ir (for internal release) instead of ot3.

This should hopefully prevent the pypi uploads from posting internal-release sdists.

Should fix this: https://github.com/Opentrons/opentrons/actions/runs/7902877476/job/21569838106


SyntaxColoring (approved):
Spicy!

PR #14499, opened 2024-02-15
fix(robot-server): fix infinite cancelling protocol bug on ODD

mjhuff:
Closes [RQA-2346](https://opentrons.atlassian.net/browse/RQA-2346)

# Overview

This PR fixes an infinite cancelling bug on the ODD caused by incorrect ordering of polling logic. The protocol engine poller used by notifications initializes before the protocol run is created, causing the poller to fail silently. The fix is simply to start the poller after the protocol run is created.

More sanity checks are added to the PE poller - we now stop the current poller, clear it, and then create a new poller if a new run is created before the old run is deleted. The polling flag is properly reset now, too.

There's also a small refactor for the Topic Enum!

# Test Plan
- Smoke tested all the major flows and a protocol run.
- Confirmed that cancelling a protocol run on the ODD actually cancels the run.

# Changelog
- Fixed an infinitely cancelling protocol run bug on the ODD.

# Risk assessment
low

[RQA-2346]: https://opentrons.atlassian.net/browse/RQA-2346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/runs/run_data_manager.py
+++ b/robot-server/robot_server/runs/run_data_manager.py
@@ -112,11 +112,6 @@ class RunDataManager:
             EngineConflictError: There is a currently active run that cannot
                 be superceded by this new run.
         """
-        await self._runs_publisher.begin_polling_engine_store(
-            get_current_command=self.get_current_command,
-            get_state_summary=self._get_state_summary,
-            run_id=run_id,
-        )
         prev_run_id = self._engine_store.current_run_id
         if prev_run_id is not None:
             prev_run_result = await self._engine_store.clear()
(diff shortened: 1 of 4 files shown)


robot-server/robot_server/runs/run_data_manager.py
mjhuff, line 1: This is where the bug occurred - everything else is just a refactor.

sfoster1 (approved):
Nice, thank you!

PR #14500, opened 2024-02-15
docs(api): Update reference docs for `ProtocolContext.pause()`

SyntaxColoring:
# Changelog

`ProtocolContext.pause()` has confusing (RESC-165, RESC-201) behavior on Python Protocol API version 2.13 and below. This *was* documented, but we didn't document that it's fixed in Protocol API v2.14. Do that here.

# Review requests

None in particular.

# Risk assessment

No risk.


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -979,9 +979,9 @@ def pause(self, msg: Optional[str] = None) -> None:
 
         A human can resume the protocol in the Opentrons App or on the touchscreen.
 
-        This function returns immediately, but the next function call that
-        is blocked by a paused robot (anything that involves moving) will
-        not return until the protocol is resumed.
+        .. note::
+            In Python Protocol API versions 2.13 and earlier, the pause will only
```
ecormany:
```suggestion
            In Python Protocol API version 2.13 and earlier, the pause will only
```
We use singular here because reasons 🧐

ecormany (approved):
Thanks for clarifying this. Part of me wants for this to be a `.. versionchanged` message, but I think a proper explanation would be too long in that format. So the `.. note` is better.

jwwojak (commented):
👍 to Ed's comment. Otherwise, LGTM.

:shipit:

PR #14502, opened 2024-02-15
fix(api): reconfigure instrument cache after a firmware update

ahiuchingau:
# Overview
The hardware controller sets the proper motor currents for each axis when the instruments are cached. Whenever a subsystem gets rebooted/updated, the motor current setting is lost. This means we must re-configure the instrument cache after we update a firmware, to make sure the subsystem motor is set up correctly.

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -727,7 +727,7 @@ class OT3API(
         await self.set_gantry_load(self._gantry_load_from_instruments())
         await self.refresh_positions()
         await self.reset_tip_detectors(False)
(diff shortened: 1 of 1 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -523,6 +524,7 @@ async def update_firmware(
                     message="Update failed because of uncaught error",
                     wrapping=[PythonException(e)],
                 ) from e
+            self._configured_since_update = False
```
sfoster1: probably wants to be in a `finally:`

sfoster1 (approved):
Looks good other than the `finally` thing!

PR #14505, opened 2024-02-15
fix(protocol-designer): pronoun agreement typo in unused item modal

ecormany:
# Overview

`these staging area slots…it` -> `them`

# Test Plan

Seems no unit test for this string.

# Review requests

We could turn this into a broader style/copy pass on this modal or we could just merge the needed typo fix.

# Risk assessment

nil


jerader (approved):
lgtm, thanks!

PR #14506, opened 2024-02-15
fix(app): return calibration flow even if pipette is already calibrated

smb2268:
fix RQA-2357

# Overview
In setup to run when a pipette is correctly attached and calibrated, we should still open the calibration flow when the user presses "recalibrate"

# Test Plan
On ODD, with a pipette attached and calibrated, go to run setup and press "Recalibrate" - the calibration flow should launch as expected

# Changelog
1. Return calibration steps from `getPipetteWizardStepsForProtocol` if a pipette is properly attached, regardless of whether it is already calibrated
2. Return `null` instead of `[]` for both `getPipetteWizardStepsForProtocol` and `getPipetteWizardSteps` to take advantage of null handling in the `PipetteWizardFlows` component

# Review requests
Look over code - I've already tested this

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/PipetteWizardFlows/__tests__/getPipetteWizardStepsForProtocol.test.tsx
+++ b/app/src/organisms/PipetteWizardFlows/__tests__/getPipetteWizardStepsForProtocol.test.tsx
@@ -24,29 +24,13 @@ const mockSingleMountPipetteAttached = {
 }
 
 describe('getPipetteWizardStepsForProtocol', () => {
-  it('returns an empty array of info when the attached pipette matches required pipette', () => {
-    const mockFlowSteps = [] as PipetteWizardStep[]
-    expect(
-      getPipetteWizardStepsForProtocol(
-        mockSingleMountPipetteAttached,
-        [
-          {
-            id: '123',
-            pipetteName: 'p1000_single_flex',
-            mount: 'left',
-          },
-        ],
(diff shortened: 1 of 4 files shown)


app/src/organisms/PipetteWizardFlows/__tests__/getPipetteWizardStepsForProtocol.test.tsx
```diff
@@ -24,29 +24,13 @@ const mockSingleMountPipetteAttached = {
...
-  })
   it('returns an empty array when there is no pipette attached and no pipette is needed', () => {
-    const mockFlowSteps = [] as PipetteWizardStep[]
+    const mockFlowSteps = null as PipetteWizardStep[] | null
     expect(
       getPipetteWizardStepsForProtocol({ left: null, right: null }, [], LEFT)
     ).toStrictEqual(mockFlowSteps)
   })
-  it('returns the calibration flow only when correct pipette is attached but there is no pip cal data', () => {
+  it('returns the calibration flow only when correct pipette is attached even if htere is pip cal data', () => {
```
mjhuff:
```suggestion
  it('returns the calibration flow only when correct pipette is attached even if there is pip cal data', () => {
```

mjhuff (approved):
LGTM!

PR #14507, opened 2024-02-15
fix(app): remove subsystem update link from instrument card

smb2268:
fix RQA-2328

# Overview
We're going to remove the ability to update a subsystem from the app-side instrument card and force users to initiate these updates on the ODD.

# Test Plan
Attach a pipette that needs a firmware update to the robot (outside of an in-app attachment flow)
See that on the app's instrument card, the new copy is displayed in the red banner

# Changelog
1. Remove firmware update modal from the pipette and gripper cards
2. Update copy to direct users to go to the ODD to update the instrument

# Review requests
Look over changes

# Risk assessment
Low


sfoster1 (approved):
Nice, that's way easier.

PR #14508, opened 2024-02-15
test(robot-server): Add persistence snapshot test for v7.1.1

SyntaxColoring:
# Overview

This adds robot-server integration tests for backwards compatibility with v7.1.1 data. This closes RSS-371.

The snapshot data comes from our ABR robots. I've taken a few runs from each robot and combined them into a single Frankenstein snapshot. I also had to massage the data somewhat to correct RSS-469.

[Here is the raw ABR robot data](https://github.com/Opentrons/opentrons/files/14303446/ABR.zip) (after correcting RSS-469), and [here are the scripts I used to create a single snapshot out of it](https://gist.github.com/SyntaxColoring/ffe789309673c57d70d1e496a3a8128b).

Also, configure Git to leave line endings alone for files in the snapshots. At least one of the protocols added here happens to have Windows CRLFs, and it seems like a good idea to keep them.

# Review requests

None in particular.

Note that CI is currently failing because of RSS-471, which I'm fixing in #14509.

# Risk assessment

No risk.


sfoster1 (approved):
Looks good to me if the tests all pass. Nice work.

robot-server/Makefile
```diff
@@ -36,6 +36,8 @@ tests ?= tests
 cov_opts ?= --cov=$(SRC_PATH) --cov-report term-missing:skip-covered --cov-report xml:coverage.xml
 test_opts ?=
 
+black_opts := --extend-exclude tests/integration/persistence_snapshots/
+
```
SyntaxColoring, line 40: image

PR #14510, opened 2024-02-16
fix(robot-server): append error to analysis error list in case an internal error occurred

TamarZanzouri:
# Overview

closes [RSS-169](https://opentrons.atlassian.net/browse/RSS-169).
append internal error to analysis errors list in case a un expected error occured. 

# Test Plan

Tested this by adding a raise in the `runner.run` method. made sure that the analysis is not stuck in a pending state. 
Can do the same test on a robot. 
Make sure analysis completed with "NOT_OK" result and that the error appears.  
```
{
    "data": [
        {
            "id": "45aeefdd-8450-470d-8289-bde0d6c17597",
            "status": "completed",
            "result": "not-ok",
            "robotType": "OT-2 Standard",
            "commands": [],
            "labware": [],
            "pipettes": [],
            "modules": [],
            "liquids": [],
            "errors": [
                {
                    "id": "internal-error",
                    "createdAt": "2024-02-16T02:24:08.423768+00:00",
                    "errorCode": "4000",
                    "errorType": "PythonException",
                    "detail": "RuntimeError: No active exception to reraise",
                    "errorInfo": {
                        "args": "('No active exception to reraise',)",
                        "traceback": "  File \"/Users/tamarzanzouri/opentrons/robot-server/robot_server/protocols/protocol_analyzer.py\", line 37, in analyze\n    result = await runner.run(\n\n  File \"/Users/tamarzanzouri/opentrons/api/src/opentrons/protocol_runner/protocol_runner.py\", line 307, in run\n    raise\n",
                        "class": "RuntimeError"
                    },
                    "wrappedErrors": []
                }
            ]
        }
    ],
    "meta": {
        "cursor": 0,
        "totalLength": 1
    }
}
```

# Changelog

- Added a "catch all" for errors that the PE does not catch internally in `analyze`.
- Create an error from the internal error and append it to the analysis error list. 

# Risk assessment

low.

[RSS-169]: https://opentrons.atlassian.net/browse/RSS-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/protocols/protocol_analyzer.py
+++ b/robot-server/robot_server/protocols/protocol_analyzer.py
@@ -2,10 +2,13 @@
 import logging
 
 from opentrons import protocol_runner
+from opentrons.protocol_engine.errors import ErrorOccurrence
+import opentrons.util.helpers as datetime_helper
+
+import robot_server.errors.error_mappers as em
 
 from .protocol_store import ProtocolResource
 from .analysis_store import AnalysisStore
-
 
 log = logging.getLogger(__name__)
 
@@ -30,9 +33,31 @@ class ProtocolAnalyzer:
             robot_type=protocol_resource.source.robot_type,
             protocol_config=protocol_resource.source.config,
         )
-        result = await runner.run(
-            protocol_source=protocol_resource.source, deck_configuration=[]
-        )
+        try:
+            result = await runner.run(
+                protocol_source=protocol_resource.source, deck_configuration=[]
+            )
+        except BaseException as error:
+            internal_error = em.map_unexpected_error(error=error)
+            await self._analysis_store.update(
+                analysis_id=analysis_id,
+                robot_type=protocol_resource.source.robot_type,
+                commands=[],
+                labware=[],
+                modules=[],
+                pipettes=[],
+                errors=[
+                    ErrorOccurrence.from_failed(
+                        # TODO(tz, 2-15-24): replace with a different error type
+                        #  when we are able to support different errors.
+                        id="internal-error",
+                        createdAt=datetime_helper.utc_now(),
+                        error=internal_error,
+                    )
+                ],
+                liquids=[],
+            )
+            return
 
         log.info(f'Completed analysis "{analysis_id}".')
 
--- a/robot-server/tests/protocols/test_protocol_analyzer.py
+++ b/robot-server/tests/protocols/test_protocol_analyzer.py
@@ -17,10 +17,30 @@ from opentrons.protocol_engine import (
 )
 import opentrons.protocol_runner as protocol_runner
 from opentrons.protocol_reader import ProtocolSource, JsonProtocolConfig
+import opentrons.util.helpers as datetime_helper
 
 from robot_server.protocols.analysis_store import AnalysisStore
 from robot_server.protocols.protocol_store import ProtocolResource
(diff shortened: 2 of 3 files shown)


robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -30,9 +33,31 @@ async def analyze(
...
+                commands=[],
+                labware=[],
+                modules=[],
+                pipettes=[],
+                errors=[
+                    ErrorOccurrence.from_failed(
+                        # TODO(tz, 2-15-24): replace with a different error type
+                        #  when we are able to support different errors.
+                        id="internal-error",
+                        createdAt=datetime.now(tz=timezone.utc),
```
SyntaxColoring: Can we use `opentrons.util.helpers.utc_now()` for this?

It's not a huge deal, but the `tz=timezone.utc` thing is easy to forget, so I'd rather not encourage our future selves to be in the habit of doing raw calls to `datetime.now()`.

There's also a similar helper defined in `robot_server` somewhere, but it's `async` to dodge FastAPI weirdness, so...I dunno.

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -30,9 +33,31 @@ async def analyze(
...
         )
-        result = await runner.run(
-            protocol_source=protocol_resource.source, deck_configuration=[]
-        )
+        try:
+            result = await runner.run(
+                protocol_source=protocol_resource.source, deck_configuration=[]
+            )
+        except Exception as error:
+            internal_error = await em.map_unexpected_error(error=error)
```
SyntaxColoring: This can be non-async, right?

robot-server/tests/protocols/test_protocol_analyzer.py
```diff
@@ -146,3 +158,78 @@ async def test_analyze(
...
         ),
     )
+
+
+async def test_analyze_updates_pending_on_error(
+    decoy: Decoy,
+    analysis_store: AnalysisStore,
+    subject: ProtocolAnalyzer,
+) -> None:
+    """It should be able to analyze a protocol."""
```
SyntaxColoring: Looks like this docstring came copy-pasted from the other test function. Can we change it to be specific to this test?

SyntaxColoring (approved):
Looks good, just a few small suggestions. Thank you!

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -30,9 +33,31 @@ async def analyze(
...
             protocol_config=protocol_resource.source.config,
         )
-        result = await runner.run(
-            protocol_source=protocol_resource.source, deck_configuration=[]
-        )
+        try:
+            result = await runner.run(
+                protocol_source=protocol_resource.source, deck_configuration=[]
+            )
+        except Exception as error:
```
sfoster1:
```suggestion
        except BaseException as error:
```
Just in case?

sfoster1 (approved):
Looks good to me, couple tiny fixes maybe

sanni-t (commented):
Looks like there was some rebasing mixup? Should rebase on top of `chore_release-7.2.0`.

PR #14512, opened 2024-02-16
refactor(api): Save and load tip length calibrations from tiprack uri

Laura-Danielle:
# Overview

This is an effort to fix the problem described in an escalations case [here](https://opentrons.atlassian.net/browse/RSS-468?atlOrigin=eyJpIjoiMTBmOTdlYTQ0YjU4NGRkYmFjNDkwZjY3ZmU2YmU2ZmEiLCJwIjoiamlyYS1zbGFjay1pbnQifQ). Where the tip length data does not persist during a protocol run, but does in LPC. @SyntaxColoring discovered that it seemed to be a result of the labware hash not always being the same once a dict has been converted to/from a pydantic model. We both agreed that the easiest path forward would be to lookup tip length calibrations by labware URI instead.

# Note
The decision was made to keep around tiprack hash for the delete endpoint until we decide to bump the version header of the API. A TODO was included in the endpoint.

# Test Plan
- Without modifying tip length data on a robot, ensure that behavior does not change (i.e. robot moves to all the correct places).
- Modify the tip length calibration to be purposefully off (as described in the ticket) and verify that the tip length persists both in LPC _and_ the protocol run.
- Delete tip length calibrations and start fresh. Ensure that behavior is normal.

# Changelog

- Load and save tip length calibrations by tiprack uri rather than hash. If the old format is detected, we should automatically migrate the data shape.
- Updated relevant locations for the new data shape

# Review requests

Check out the code and make sure everything makes sense to you.

# Risk assessment

High. This is modifying a critical component of loading tip length calibration.


api/src/opentrons/calibration_storage/ot2/tip_length.py
```diff
@@ -35,17 +36,21 @@ def _conver_tip_length_model_to_dict(
...
 def tip_lengths_for_pipette(
     pipette_id: str,
-) -> typing.Dict[str, v1.TipLengthModel]:
+) -> typing.Dict[LabwareUri, v1.TipLengthModel]:
     tip_lengths = {}
     try:
         tip_length_filepath = config.get_tip_length_cal_path() / f"{pipette_id}.json"
         all_tip_lengths_for_pipette = io.read_cal_file(tip_length_filepath)
-        for tiprack, data in all_tip_lengths_for_pipette.items():
+        for tiprack_uri, data in all_tip_lengths_for_pipette.items():
```
sfoster1:
```suggestion
        for tiprack_identifier, data in all_tip_lengths_for_pipette.items():
```
kind of a nit, but with this plus line 47 I got pretty confused about what this data represented - can we change the name and maybe add a comment about what it's supposed to be?

robot-server/robot_server/service/tip_length/router.py
```diff
@@ -80,17 +81,19 @@ async def get_all_tip_length_calibrations(
 @router.delete(
     "/calibration/tip_length",
     description="Delete one specific tip length calibration by pipette "
-    "serial and tiprack hash",
+    "serial and tiprack uri",
```
sfoster1, line 84: If this isn't called and we don't use it let's deprecate it if not remove it.

sfoster1 (approved):
Looks good to me with the couple of small nits.

I think this is going to need a pretty broad testing pass and I'm not sure what the timing will be, so for now let's
- hold it open until we can test it on or after the 26th
- decide then whether it goes to edge or release

robot-server/robot_server/service/tip_length/router.py
Laura-Danielle: Agree. Just waiting on RSS to figure out the best way to do this.

api/src/opentrons/calibration_storage/ot2/models/v1.py
```diff
@@ -32,8 +32,8 @@ class TipLengthModel(BaseModel):
         default_factory=CalibrationStatus,
         description="The status of the calibration data.",
     )
-    uri: typing.Union[LabwareUri, Literal[""]] = Field(
-        ..., description="The tiprack URI associated with the tip length data."
+    definitionHash: str = Field(
+        ..., description="The tiprack hash associated with the tip length data."
```
SyntaxColoring: I'm not sure if this is the same thing that Seth was [asking about](https://github.com/Opentrons/opentrons/pull/14512/files#r1492629242), but why are we changing this from a URI (we think URIs are good) to a hash (we think hashes are bad)?

Laura-Danielle: the way the info was previously stored:

`tiprack_hash: {..rest of data here including uri}`. 

the way the data is now stored:

`tiprack_uri: {..rest of data (we don't want to lose hash info}`

When the app receives this info, it's flattened to a single level dict with all of that info.

robot-server/tests/integration/test_tip_length_access.tavern.yaml
```diff
@@ -148,14 +148,14 @@ marks: *cal_marks
 stages:
   - name: DELETE request with correct pipette AND tiprack
     request:
-      url: "{ot2_server_base_url}/calibration/tip_length?pipette_id=321&tiprack_hash=130e17bb7b2f0c0472dcc01c1ff6f600ca1a6f9f86a90982df56c4bf43776824"
+      url: "{ot2_server_base_url}/calibration/tip_length?pipette_id=321&tiprack_uri=opentrons/opentrons_96_filtertiprack_200ul/1"
```
SyntaxColoring, line 151: It doesn't look like this test is asserting anywhere that the calibration is actually deleted. i.e. if you `GET /calibration/tip_length` again, the correct entry has been removed.

Do we need to add that here, or are you satisfied with automated test coverage elsewhere?

One thing we'd want to make sure is working properly, for example, is that the slashes in the labware URI don't get messed up by HTTP URL encoding.

api/src/opentrons/calibration_storage/ot2/tip_length.py
```diff
@@ -129,7 +134,7 @@ def get_custom_tiprack_definition_for_tlc(labware_uri: str) -> "LabwareDefinitio
 # Delete Tip Length Calibration
 
 
-def delete_tip_length_calibration(tiprack: str, pipette_id: str) -> None:
+def delete_tip_length_calibration(tiprack: LabwareUri, pipette_id: str) -> None:
```
SyntaxColoring: Looks like this function docstring needs updating to reflect that the `tiprack` arg is a URI now, not a hash.

robot-server/tests/integration/test_tip_length_access.tavern.yaml
Laura-Danielle: Added an additional check but not in tavern because I was lazy.

api/src/opentrons/calibration_storage/ot2/tip_length.py
```diff
@@ -35,17 +36,21 @@ def _conver_tip_length_model_to_dict(
...
-) -> typing.Dict[str, v1.TipLengthModel]:
+) -> typing.Dict[LabwareUri, v1.TipLengthModel]:
     tip_lengths = {}
     try:
         tip_length_filepath = config.get_tip_length_cal_path() / f"{pipette_id}.json"
         all_tip_lengths_for_pipette = io.read_cal_file(tip_length_filepath)
-        for tiprack, data in all_tip_lengths_for_pipette.items():
+        for tiprack_identifier, data in all_tip_lengths_for_pipette.items():
+            # check if tiprack hash
+            if len(tiprack_identifier.split("/")) == 1:
```
SyntaxColoring: Nit: Is this clearer?

```suggestion
            if "/" not in tiprack_identifier:
```

api/src/opentrons/calibration_storage/ot2/models/v1.py
SyntaxColoring, line 40: If we're modifying this v1 model in-place instead of making a v2, can we leave a comment like this for posterity?

```suggestion
    # Old data may have a `uri` field, replaced later by `definitionHash`.
    # uri: typing.Union[LabwareUri, Literal[""]] = Field(
    #    ..., description="The tiprack URI associated with the tip length data."
    # )
    definitionHash: str = Field(
        ..., description="The tiprack hash associated with the tip length data."
```

api/src/opentrons/calibration_storage/ot2/tip_length.py
```diff
@@ -35,17 +36,21 @@ def _conver_tip_length_model_to_dict(
...
     pipette_id: str,
-) -> typing.Dict[str, v1.TipLengthModel]:
+) -> typing.Dict[LabwareUri, v1.TipLengthModel]:
     tip_lengths = {}
     try:
         tip_length_filepath = config.get_tip_length_cal_path() / f"{pipette_id}.json"
         all_tip_lengths_for_pipette = io.read_cal_file(tip_length_filepath)
-        for tiprack, data in all_tip_lengths_for_pipette.items():
+        for tiprack_identifier, data in all_tip_lengths_for_pipette.items():
+            # check if tiprack hash
```
SyntaxColoring: Mind if we elaborate this comment a bit?

I'm thinking something like this.

```suggestion
            # We normally key these calibrations by their tip rack URI,
            # but older software had them keyed by their tip rack hash.
            # Migrate from the old format, if necessary.
```

SyntaxColoring (commented):
This looks good. Various nitpicks and small suggestions as I get familiar with this.

robot-server/robot_server/service/tip_length/router.py
SyntaxColoring: I see this note in the PR description:

> The decision was made to keep around tiprack hash for the delete endpoint until we decide to bump the version header of the API. A TODO was included in the endpoint.

Has this been resolved, or are there still open questions?

api/src/opentrons/hardware_control/instruments/ot2/instrument_calibration.py
```diff
@@ -133,15 +133,15 @@ def load_tip_length_for_pipette(
     # TODO (lc 09-26-2022) We shouldn't have to do a hash twice. We should figure out what
```
TamarZanzouri: do we still need this comment?

TamarZanzouri (approved):
I agree with a few of Max's comment but besides that looks great!

robot-server/robot_server/service/tip_length/router.py
Laura-Danielle: I think @smb2268 and I came up with a sufficient solution.

SyntaxColoring (commented):
Sweet, this looks good, pending client-side changes (JS linting is currently failing). Thank you for taking this on. I'm very happy that we're moving away from the fundamentally problematic hashing instead of trying to patch it up.

I agree with you and Seth that we'll have to test this carefully on a robot. I'll do that as soon as I can.

nusrat813 (approved):
Tested on OT2 refresh 
- Tip lengths persisted in LPC and Run 
- Tip length cals were able to be deleted

PR #14517, opened 2024-02-16
chore: fix ci setup-py

sfoster1:
CI's setup-py step has been failing with mysterious pipenv problems. It also only fails once. So try again if it fails. I am not proud

Also somehow an empty notify-server workflow got snuck back in, so remove it


vegano1 (approved):
weird

PR #14518, opened 2024-02-16
refactor(protocol-designer): ot-2 slot map is now a normal size

jerader:
closes RAUT-1005

# Overview

See the ticket. no idea when this error was introduced - apparently after i did the migration to react-hook-form?? Weird. But anyway, I fixed it here and also cleaned up the SlotMap components, getting rid of the no longer needed `ConnectSlotMap`

Bonus: also there was a small bug with the well selection field where it was defaulting the value to 0 which is different from PD in prod. So i fixed that up really quick too. 

# Test Plan

Create an Ot-2 protocol and edit the modules. See that the slot map is no longer way too big. 

# Changelog

- clean up the `SlotMap` component and associated components
- fix test
- fix bug with well selection field where it was defaulting the value to 0 if the length of well array was 0.

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/StepEditForm/fields/WellSelectionField/WellSelectionField.tsx
```diff
@@ -40,9 +40,10 @@ export const WellSelectionField = (props: Props): JSX.Element => {
   const stepId = useSelector(getSelectedStepId)
   const pipetteEntities = useSelector(stepFormSelectors.getPipetteEntities)
   const wellSelectionLabwareKey = useSelector(getWellSelectionLabwareKey)
-  const primaryWellCount = Array.isArray(selectedWells)
-    ? selectedWells.length.toString()
-    : undefined
+  const primaryWellCount =
+    Array.isArray(selectedWells) && selectedWells.length > 0
+      ? selectedWells.length.toString()
+      : undefined
```
jerader, line 46: unrelated to the bug but this fixed another small bug i found that was introduced by my PR that refactored class components into functional components 😬

mjhuff (approved):
Looks good! IMO, I wouldn't be opposed to naming `SlotMap` to something like `OT2SlotMap` or leaving a docstring explaining that this is an OT-2 specific component, since we have an analogous `FlexSlotMap`.

jerader:
> Looks good! IMO, I wouldn't be opposed to naming `SlotMap` to something like `OT2SlotMap` or leaving a docstring explaining that this is an OT-2 specific component, since we have an analogous `FlexSlotMap`.

@mjhuff good call! thank you! `SlotMap` used to be used for both robot types so I forgot that I made a `FlexSlotMap` later. I'll rename it to `OT2SlotMap`

PR #14521, opened 2024-02-16
refactor(protocol-designer): nicknames now properly import

jerader:
closes RQA-2358, RQA-1927

# Overview

This bug has been reported by Support numerous times so it is a high priority bug to get in with the next PD release. Looking into it, its definitely a bug I introduced 7 months ago by accidentally listing the container info via the labware's loadname instead of the labware Id. Oops. 

# Test Plan

Create a protocol either flex or ot-2. When adding a labware, add a nickname. Also add duplicate tipracks. Add a command that uses the trash. Export. Import the protocol, see that the nicknames on the labware and the tiprack (the numbering at the end of the tiprack) are preserved. 

# Changelog

- fix the reducer to list the containers via the labware Id instead of the loadname

# Review requests

see test plan

# Risk assessment

low


koji (approved):
tested and the sandbox worked as expected.
- numbers were displayed
- nick names were displayed on Desktop app

PR #14522, opened 2024-02-16
feat(api): raise an error if pipette nozzles might collide with thermocycler lid clips

sanni-t:
Addresses RQA-2311

# Overview

This is the second part of the 3-part fix for RQA-2311.

# Test Plan

The following protocol attempts to move the 96-ch to A2 in a way that it would collide with the TC pins. Make sure error is raised during analysis as well as during the actual run on the robot.

Change the destination well and verify that it does NOT raise for target positions that don't hit the thermocycler.

```
from opentrons import types

requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16",
}

	thermocycler  = protocol.load_module('thermocycler module gen2')
	tiprack       = protocol.load_labware('opentrons_flex_96_tiprack_200ul', 'A2')
	p1000 		  = protocol.load_instrument("flex_96channel_1000", "left")
	thermocycler.open_lid()
	p1000.configure_nozzle_layout(style=COLUMN, start="A12", tip_racks=[tiprack])
	p1000.pick_up_tip(tiprack.wells()[0].center().move(types.Point(x=-10, y=10, z=-10)))
```

# Changelog

- added absolute measurements of the TC lid's clips
- added collision check for the no-go zone around the lid clips
- minor refactors

# Review requests

- any questions/ concerns about the approach?

# Risk assessment

Low. Only adds a new no-go zone for the 96-channel.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
+++ b/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
@@ -10,9 +10,11 @@ from typing import (
     overload,
     Union,
     TYPE_CHECKING,
+    List,
 )
 
 from opentrons_shared_data.errors.exceptions import MotionPlanningFailureError
+from opentrons_shared_data.module import FLEX_TC_LID_CLIP_POSITIONS_IN_DECK_COORDINATES
 
 from opentrons.hardware_control.nozzle_manager import NozzleConfigurationType
 from opentrons.hardware_control.modules.types import ModuleType
@@ -30,7 +32,10 @@ from opentrons.protocol_engine import (
     DropTipWellLocation,
 )
 from opentrons.protocol_engine.errors.exceptions import LabwareNotLoadedOnModuleError
-from opentrons.protocol_engine.types import StagingSlotLocation, Dimensions
+from opentrons.protocol_engine.types import (
+    StagingSlotLocation,
+    Dimensions,
+)
 from opentrons.types import DeckSlotName, StagingSlotName, Point
 from ..._trash_bin import TrashBin
 from ..._waste_chute import WasteChute
@@ -195,7 +200,8 @@ def check(
     )
 
 
-def check_safe_for_pipette_movement(  # noqa: C901
+# TODO (spp, 2023-02-16): move pipette movement safety checks to its own separate file.
+def check_safe_for_pipette_movement(
     engine_state: StateView,
     pipette_id: str,
     labware_id: str,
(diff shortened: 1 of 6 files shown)


shared-data/python/opentrons_shared_data/module/__init__.py
```diff
@@ -16,9 +16,16 @@
 
 OLD_TC_GEN2_LABWARE_OFFSET = {"x": 0, "y": 68.06, "z": 98.26}
 
+# TODO (spp, 2023-02-14): make these a part of thermocycler/ deck definitions
+FLEX_TC_LID_CLIP_POSITIONS_IN_DECK_COORDINATES = {
+    "left_clip": {"x": -3.25, "y": 402, "z": 205},
+    "right_clip": {"x": 97.75, "y": 402, "z": 205},
+}
```
SyntaxColoring: You mentioned on a call that you measured these experimentally. That sounds good, in the absence of an official value from technical drawings. But can we leave a comment explaining that, so it won't confuse people if we get a different official value later on?

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
```diff
@@ -249,44 +255,94 @@ def check_safe_for_pipette_movement(  # noqa: C901
...
+    pipette_bounds: Tuple[Point, Point, Point, Point],
+    surrounding_regular_slots: List[DeckSlotName],
+) -> bool:
+    """Return whether the pipette might collide with thermocycler's lid on a Flex."""
+    if (
+        DeckSlotName.SLOT_A1 in surrounding_regular_slots
+        and engine_state.modules.is_flex_deck_with_thermocycler()
+    ):
+        tc_right_clip_pos = FLEX_TC_LID_CLIP_POSITIONS_IN_DECK_COORDINATES["right_clip"]
+        for bound_vertex in pipette_bounds:
```
CaseyBatten, line 349: So currently we're only doing this check on the outside edge of the right clip. What happens in the following case:
If a single channel pipette were to move inside the thermocycler then manually offsets to a position in which it would collide with the *inside* edge (so left edge) of the right clip, would we not catch it?

Same question applies to anything that would collide with the left clip.

CaseyBatten (approved):
Looks like this will help to satisfy the ticket that it seeks to solve, so good on that front. I have a question down below regarding cases we may not be catching with this logic however.

api/src/opentrons/protocol_api/core/engine/deck_conflict.py
sanni-t: This check looks for anything left of the right edge of the right clip so it covers your case. That said, these checks are currently only enabled for partially configured pipettes so it won't raise for single channels. 

But it just made me realize that this might actually raise an unnecessary error if an 8-channel is to the left of the left clip. Not sure if it's actually possible for an 8-channel to be in that nook safely. But I can update this check regardless.

PR #14523, opened 2024-02-20
docs(api): new `ProtocolContext.deck` behavior for multi-slot modules

ecormany:
# Overview

Follow up to dev work in #14491. 

# Test Plan

Check the [sandbox](http://sandbox.docs.opentrons.com/docs-tc-slot-bugfix/v2/).

# Changelog

- cherrypick commit to keep 2.17 docs out of 7.2.0 feature branch
- remove known issue about TC leaving slot A1 open on Flex
- expand docstring for `ProtocolContext.deck`

# Review requests

Would like an all clear on removing the known issue. You could still get in a bad jam on 7.1.x.

# Risk assessment

v low, docs

### DIFF
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -1092,12 +1092,11 @@ class ProtocolContext(CommandPublisher):
             reflect the new deck state, add a :py:meth:`.pause` or use
             :py:meth:`.move_labware` instead.
 
-        .. versionchanged:: 2.15
-           ``del`` sets the corresponding labware's location to ``OFF_DECK``.
-
         .. versionchanged:: 2.14
            Includes the Thermocycler in all of the slots it occupies.
 
+        .. versionchanged:: 2.15
+           ``del`` sets the corresponding labware's location to ``OFF_DECK``.
         """
         return self._deck
 
--- a/api/src/opentrons/protocol_api/protocol_context.py
+++ b/api/src/opentrons/protocol_api/protocol_context.py
@@ -1075,7 +1075,8 @@ class ProtocolContext(CommandPublisher):
         relevant slots. Currently, the only multiple-slot module is the Thermocycler.
         When loaded, the :py:class:`ThermocyclerContext` object is the value for
         ``deck`` keys ``"A1"`` and ``"B1"`` on Flex, and ``7``, ``8``, ``10``, and
-        ``11`` on OT-2.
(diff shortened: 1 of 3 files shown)


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1066,9 +1066,16 @@ def deck(self) -> Deck:
...
-        labware, a module context if the slot contains a hardware
-        module, or ``None`` if the slot doesn't contain anything.
+        The value for each key depends on what is loaded in the slot:
+          - A :py:obj:`~opentrons.protocol_api.Labware` if the slot contains a labware.
+          - A module context if the slot contains a hardware module.
+          - ``None`` if the slot doesn't contain anything.
+
+        A module that occupies multiple slots is set as the value for all of the
+        relevant slots. Currently, the only multiple-slot module is the Thermocycler.
+        When loaded, the :py:class:`ThermocyclerContext` object is the value for
```
jwwojak, line 1076: Hey, maybe I'm reading this through a straw and not seeing the bigger context. I had to pause and think about this one for a moment.

> When loaded, the `ThermocyclerContext` object is the value for `deck` keys `A1` and `B1` on Flex, and ... on OT-2.

So if there's a Thermocycler on the deck, then something like `deck[A1]` returns `thermocycler` because the robot knows that item occupies A1 and B1?

jwwojak (commented):
Aside from some small amount of confusion about the text related to the `deck` property, this looks fine.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -1085,6 +1092,9 @@ def deck(self) -> Deck:
             reflect the new deck state, add a :py:meth:`.pause` or use
             :py:meth:`.move_labware` instead.
 
+        .. versionchanged:: 2.14
+           Includes the Thermocycler in all of the slots it occupies.
```
sanni-t, line 1097: Should we mention that in prior versions it returns `None` for those slots even when they are occupied by the thermocycler?

sanni-t (approved):
Looks great! 
Just a question about whether we document old vs new behavior, or just state the latest behavior.

api/src/opentrons/protocol_api/protocol_context.py
ecormany: Sure, I'll put that in the docstring body (to keep this directive short).

jwwojak (commented):
LGTM if you want the extra 2 cents.

🪙

PR #14525, opened 2024-02-20
refactor(step-generation): openLatch should emit last in heaterShaker…

jerader:
… compound command

closes RQA-2361, RESC-195

# Overview

Very strange that this bug was never reported until now. Basically, if a user adds a timer step in the heater-shaker form with an open latch, the open latch occurs before the heater and shaker are deactivated. This is not allowed in protocols so it would fail analysis. For the fix, i swapped the order so the open latch is emitted last.

# Test Plan

1. upload the attached protocol to the app. see that it fails analysis with an error saying the H-S can't heat or shake with the latch open
2. Upload the attached protocol to PD and then export then upload that to the app. it should pass analysis.

This is the customer's protocol
[test-1 (2).json](https://github.com/Opentrons/opentrons/files/14349054/test-1.2.json)

# Changelog

- change the command order in the H-S compound command
- add a test case

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/__tests__/heaterShaker.test.ts
+++ b/step-generation/src/__tests__/heaterShaker.test.ts
@@ -127,7 +127,7 @@ describe('heaterShaker compound command creator', () => {
       },
     ])
   })
-  it('should NOT delay and deactivate the heater shaker when a user specificies a timer tthat is 0 seconds', () => {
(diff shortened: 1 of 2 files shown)


mjhuff (approved):
lgtm! thanks for adding a test for the bug, too!

PR #14529, opened 2024-02-20
refactor(api): Save full nozzle map configuration and update state store accordingly

Laura-Danielle:
# Overview
To make a few operations easier in protocol engine, we will keep the nozzle map in state always. I will think about this further in RSS-443, but for now it should unblock other partial tip work. 

# Test Plan

General sanity check that partial tip still works and regular protocols operate as normal.

# Changelog

- Pull in full nozzle map from the pipette dict
- Remove the back left/right values on the `LoadedStaticPipetteData`

# Review requests

Check that I didn't miss anything to update.

# Risk assessment

Low/medium. Currently not using this information anywhere aside from back left/right values which will remain the same. Tests only updated due to new helper function pulling in mock data.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/pipettes.py
+++ b/api/src/opentrons/protocol_engine/state/pipettes.py
@@ -102,6 +102,7 @@ class StaticPipetteConfig:
     home_position: float
     nozzle_offset_z: float
     bounding_nozzle_offsets: BoundingNozzlesOffsets
+    default_nozzle_map: NozzleMap
 
 
 @dataclass
@@ -170,6 +171,7 @@ class PipetteStore(HasState[PipetteState], HandlesActions):
                     back_left_offset=config.nozzle_map.back_left_nozzle_offset,
                     front_right_offset=config.nozzle_map.front_right_nozzle_offset,
                 ),
+                default_nozzle_map=config.nozzle_map,
             )
             self._state.flow_rates_by_id[private_result.pipette_id] = config.flow_rates
             self._state.nozzle_configuration_by_id[
@@ -191,7 +193,11 @@ class PipetteStore(HasState[PipetteState], HandlesActions):
             self._state.aspirated_volume_by_id[pipette_id] = None
             self._state.movement_speed_by_id[pipette_id] = None
             self._state.attached_tip_by_id[pipette_id] = None
-            self._state.nozzle_configuration_by_id[pipette_id] = None
+            static_config = self._state.static_config_by_id.get(pipette_id)
+            if static_config:
+                self._state.nozzle_configuration_by_id[
+                    pipette_id
+                ] = static_config.default_nozzle_map
 
         elif isinstance(command.result, (AspirateResult, AspirateInPlaceResult)):
             pipette_id = command.params.pipetteId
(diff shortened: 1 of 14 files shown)


sfoster1 (approved):
This looks good as long as it has the right serializability - will this get persisted?

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -167,11 +167,14 @@ def _handle_command(  # noqa: C901
...
-                    back_left_offset=config.back_left_nozzle_offset,
-                    front_right_offset=config.front_right_nozzle_offset,
+                    back_left_offset=config.nozzle_map.back_left_nozzle_offset,
+                    front_right_offset=config.nozzle_map.front_right_nozzle_offset,
                 ),
             )
             self._state.flow_rates_by_id[private_result.pipette_id] = config.flow_rates
+            self._state.nozzle_configuration_by_id[
+                private_result.pipette_id
+            ] = config.nozzle_map
```
SyntaxColoring, line 179: Can we make `nozzle_configuration_by_id` values non-`Optional` now?

SyntaxColoring:
> This looks good as long as it has the right serializability - will this get persisted?

I think we're good. I'm not seeing any modification of Pydantic `BaseModel`s here, so it doesn't look like any of this will be stored in the database.

api/src/opentrons/protocol_engine/state/pipettes.py
SyntaxColoring: Discussed with @Laura-Danielle. We'll defer this change. Might want to leave a `# TODO` explaining what needs to happen, though.

SyntaxColoring (approved):
Sweet. Thanks!

We won't have a chance to test this on any robots for a bit. Your call whether you want to merge this before then. I'm not sure how much other stuff is blocked behind this.

api/tests/opentrons/protocol_engine/resources/test_pipette_data_provider.py
```diff
@@ -81,8 +77,7 @@ def test_configure_virtual_pipette_for_volume(
         ),
         tip_configuration_lookup_table=result1.tip_configuration_lookup_table,
         nominal_tip_overlap=result1.nominal_tip_overlap,
-        back_left_nozzle_offset=Point(x=-8.0, y=-22.0, z=-259.15),
-        front_right_nozzle_offset=Point(x=-8.0, y=-22.0, z=-259.15),
+        nozzle_map=result1.nozzle_map,
```
sanni-t, line 80: Hmm, just realizing these aren't great asserts. They're just comparing an item with itself; like `result1.nozzle_map` == `result1.nozzle_map`. We should check it against the actual expected nozzle map.

api/tests/opentrons/protocol_engine/resources/test_pipette_data_provider.py
```diff
@@ -105,8 +100,7 @@ def test_configure_virtual_pipette_for_volume(
         ),
         tip_configuration_lookup_table=result2.tip_configuration_lookup_table,
         nominal_tip_overlap=result2.nominal_tip_overlap,
-        back_left_nozzle_offset=Point(x=-8.0, y=-22.0, z=-259.15),
-        front_right_nozzle_offset=Point(x=-8.0, y=-22.0, z=-259.15),
+        nozzle_map=result2.nozzle_map,
```
sanni-t, line 103: same as above

sanni-t (approved):
LGTM! Just the comment about better test assertions

PR #14533, opened 2024-02-21
refactor(protocol-designer): prevent zombie children in stepEditForm

jerader:
# Overview

🧟🧟‍♂️🧟‍♀️
 
I was running into stale props issues with adding multiple step forms and could not figure out why! Turns out, it is because of this edge case in redux: https://react-redux.js.org/api/hooks#stale-props-and-zombie-children. I thought that it would no longer be an issue for some reason since this was listed as an issue 2 years ago, but it is def still an issue haha. So this PR just reverts the component back to how it used to be before i refactored it.

# Test Plan

Create a flex or ot-2 protocol. Add a transfer step. Then add a mix step, see that the fields (in particular the mix volume field) aren't erroring before they were dirty.

# Changelog

revert the `StepEditForm` component back to using the connect function.

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepEditForm/index.tsx
+++ b/protocol-designer/src/components/StepEditForm/index.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
-import { useDispatch, useSelector } from 'react-redux'
 import { useTranslation } from 'react-i18next'
+import { connect } from 'react-redux'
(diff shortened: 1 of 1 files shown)


mjhuff (approved):
Agree with the diagnosis. As there's no change in functionality and performance, I think reverting is fine. ✅

PR #12193, opened 2023-02-24
feat(robot-server): hash protocol + labware contents to avoid duplicate entries 

shlokamin:
# Overview
This PR adds a content hash property to protocol resources. The hash is generated when `POST`ing a protocol in the robot server, and on robot boot when generating protocol sources into memory.

closes RCORE-614

# Test Plan

- Used postman to `POST` to the `/protocols` endpoint multiple times with the same protocol
- Validated that the first `POST` returned a 201 status, and all subsequent returned 200s
- Validated that when hitting `GET` `/protocols` I only got one entry back
- Restarted a robot and made sure i still got 200 statuses and new protocol entries were not being created

# Changelog

- Hash protocol and labware contents to avoid duplicate entries 

# Review requests

Follow test plan above

# Risk assessment

High — this touches a core API, and one that the app heavily depends on

PR #14535, opened 2024-02-21
refactor(robot-server): Ensure SQL selects are ordered

### LINKED ISSUE #10304: Ensure all SQL selects are ordered
# Overview

By default, SQL `select` statements return results in an undefined order. If you want ordered results, you have to ask for it.

We have a way of asking for resources to be returned in the order we originally inserted them:

https://github.com/Opentrons/opentrons/blob/9b91905903ecbafb400ffe8fe66fdc8f133be4f7/robot-server/robot_server/persistence.py#L116-L128

But we don't use this everywhere yet. For example:

https://github.com/Opentrons/opentrons/blob/9b91905903ecbafb400ffe8fe66fdc8f133be4f7/robot-server/robot_server/protocols/protocol_store.py#L289

# Acceptance criteria

* SQLAlchemy `.select()` statements are audited. Where appropriate, they have an `.order_by(sqlite_rowid)` clause added. 
* This ordering guarantee is mentioned in the method docstrings.
* This ordering guarantee is tested. Probably in the store's unit tests.
* This ordering guarantee is mentioned in the HTTP API docs, if appropriate.

SyntaxColoring:
# Overview

These are some low-priority `robot-server` fixes.

# Changelog

* Document our HTTP API's ordering guarantees for runs, commands, actions, and protocols.
* Make sure that our SQL `SELECT` statements all have explicit `ORDER BY` clauses to support those guarantees. We were omitting this in a lot of places, which is undefined behavior. It just happened not to cause any problems. This closes #10304.
* Update the `POST /protocols` documentation for PR #12193.

# Test plan

We're covered by existing automated tests.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/robot-server/robot_server/protocols/protocol_store.py
+++ b/robot-server/robot_server/protocols/protocol_store.py
@@ -188,7 +188,10 @@ class ProtocolStore:
 
     @lru_cache(maxsize=_CACHE_ENTRIES)
     def get_all(self) -> List[ProtocolResource]:
-        """Get all protocols currently saved in this store."""
+        """Get all protocols currently saved in this store.
+
+        Results are ordered from first-added to last-added.
+        """
         all_sql_resources = self._sql_get_all()
         return [
             ProtocolResource(
@@ -333,7 +336,7 @@ class ProtocolStore:
     def _sql_get_all_from_engine(
         sql_engine: sqlalchemy.engine.Engine,
     ) -> List[_DBProtocolResource]:
-        statement = sqlalchemy.select(protocol_table)
+        statement = sqlalchemy.select(protocol_table).order_by(sqlite_rowid)
         with sql_engine.begin() as transaction:
             all_rows = transaction.execute(statement).all()
         return [_convert_sql_row_to_dataclass(sql_row=row) for row in all_rows]
--- a/robot-server/robot_server/protocols/protocol_store.py
+++ b/robot-server/robot_server/protocols/protocol_store.py
@@ -297,17 +297,16 @@ class ProtocolStore:
 
         See the `runs` module for information about runs.
 
-        Results are ordered with the oldest-added (NOT created) run first.
+        Results are ordered with the oldest run first.
         """
-        select_referencing_run_ids = sqlalchemy.select(run_table.c.id).where(
-            run_table.c.protocol_id == protocol_id
+        select_referencing_run_ids = (
+            sqlalchemy.select(run_table.c.id)
+            .where(run_table.c.protocol_id == protocol_id)
+            .order_by(sqlite_rowid)
         )
 
         with self._sql_engine.begin() as transaction:
-            referencing_run_ids = (
-                transaction.execute(select_referencing_run_ids).scalars().all()
-            )
-        return referencing_run_ids
+            return transaction.execute(select_referencing_run_ids).scalars().all()
 
     def _sql_insert(self, resource: _DBProtocolResource) -> None:
         statement = sqlalchemy.insert(protocol_table).values(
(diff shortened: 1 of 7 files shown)


robot-server/robot_server/protocols/protocol_store.py
```diff
@@ -297,17 +300,16 @@ def get_referencing_run_ids(self, protocol_id: str) -> List[str]:
 
         See the `runs` module for information about runs.
 
-        Results are ordered with the oldest-added (NOT created) run first.
```
SyntaxColoring, line 300: @shlokamin Do you remember what the intent of this was? Reading this now, I don't understand the distinction between oldest-added and oldest-created runs.

robot-server/robot_server/protocols/router.py
```diff
@@ -129,11 +132,18 @@ class ProtocolLinks(BaseModel):
         When too many protocols already exist, old ones will be automatically deleted
         to make room for the new one.
         A protocol will never be automatically deleted if there's a run
-        referring to it, though.
+        referring to it, though. (See the `/runs/` endpoints.)
```
TamarZanzouri, line 135: nice!

robot-server/robot_server/protocols/router.py
```diff
@@ -129,11 +132,18 @@ class ProtocolLinks(BaseModel):
...
+        If you upload the exact same set of files multiple times, the first protocol
+        resource will be returned instead of creating duplicate ones.
+
+        When a new protocol resource is created, an analysis is started for it.
+        See the `/protocols/{id}/analyses/` endpoints.
         """
     ),
     status_code=status.HTTP_201_CREATED,
     responses={
+        status.HTTP_200_OK: {"model": SimpleBody[Protocol]},
```
TamarZanzouri, line 146: is this for the case there was already the same uploaded protocol?

robot-server/robot_server/runs/run_store.py
```diff
@@ -237,26 +241,28 @@ def get(self, run_id: str) -> RunResource:
...
                     .limit(length)
                 )
                 # need to select the last inserted runs and return by asc order
                 runs = list(reversed(transaction.execute(select_runs).all()))
             else:
-                runs = transaction.execute(
-                    select_runs.order_by(sqlite_rowid.asc())
-                ).all()
+                select_runs = sqlalchemy.select(*_run_columns).order_by(
+                    sqlite_rowid.asc()
```
TamarZanzouri, line 263: probably don't need `.asc()` but its more explicit

TamarZanzouri (approved):
looks great! thank you for doing this! if I am not mistaking the default is `.asc()` but its better to specify the order we want it returned.

robot-server/robot_server/protocols/router.py
SyntaxColoring: Yep! Keeping the documentation in sync with the behavior down here:

https://github.com/Opentrons/opentrons/blob/791a26dd5103438ef12ea5fd67c3dc2b332273b0/robot-server/robot_server/protocols/router.py#L226-L227

robot-server/robot_server/runs/run_store.py
SyntaxColoring: Yep, this is just for explicitness and symmetry, since we're doing `.desc()` elsewhere in this function.

robot-server/robot_server/protocols/protocol_store.py
shlokamin: i really have no idea what i meant. i think i probably just worded this poorly and incorrectly. my best guess is that i was trying to convey was that results are ordered by when the run record was created (NOT when the run was started)

PR #14536, opened 2024-02-21
feat(step-generation): error if gripper moves when pipette has tip on it

jerader:
closes RAUT-907, addresses RQA-2098

# Overview

This was an error that was reported awhile back (RQA-2098) and was since addressed in PE but we needed to address it in PD as well. So if you move the gripper when any pipette has a tip on it still, it will have a timeline error to avoid collisions

Screen Shot 2024-02-21 at 1 34 05 PM

# Test Plan

1. Create a flex protocol and add the gripper
2. Add a transfer step from 1 well to another well. 
3. Add a move labware step with the gripper
4. Add another transfer step from 1 well to another well but this time, modify the the `Change Tip` option to `never`. This means that step-generation will not drop tip after the 1st transfer, therefore leaving the tip on the pipette when the move labware step occurs. 
5. observe the timeline error
6. modify the move labware step to not use the gripper and the timeline error should go away

# Changelog

- check if there is a tip on any pipette, if so and the gripper is in use, emit error.
- add new error creator for tip on pipette
- add test case

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/localization/en/alert.json
```diff
@@ -175,6 +175,10 @@
       "CANNOT_MOVE_WITH_GRIPPER": {
         "title": "Cannot move with gripper",
         "body": "The gripper cannot move aluminum blocks. Edit the step and deselect the 'Use Gripper' checkbox."
+      },
+      "PIPETTE_HAS_TIP": {
+        "title": "Collision warning due to pipette tip",
+        "body": "The gripper cannot be moved when any pipettes have tips on them, please drop tip in order to continue."
```
jerader: @ecormany, please help me rewrite this when you have minute. This error occurs when a user tries to move labware with the gripper and the pipette still has a tip on it. Thanks!

ecormany:
```suggestion
        "title": "Possible collision with tip",
        "body": "The gripper cannot pick up labware while pipettes have tips attached. Drop all tips before this move labware step."
```
How's this?

jerader: This is great! much clearer. thank you!

mjhuff (approved):
Thanks for fixing! 🦾

PR #14537, opened 2024-02-21
revert(app): Privacy policy changes and opt-in strategy

b-cooper:
# Overview

Revert changes to the privacy settings toggle messaging and prompts in the Desktop and On Device Display apps.  Behavior should be consistent with 7.1.1 as far as opting-in/out of analytics.

# Review requests

- Confirm that there is nothing unintentional reverted within the 4 constituent commits.

# Risk assessment
low


jerader (commented):
Privacy tab seems to be fine & working as expected everywhere except for here. Not sure why, i'll dig into it.

Screen Shot 2024-02-26 at 10 22 59 AM

jerader:
> Privacy tab seems to be fine & working as expected everywhere except for here. Not sure why, i'll dig into it.
> 

@b-cooper  I did some digging, do we need to add log aggregation back? Looks like its not being returned from the `getRobotSettings` selector.

jerader (approved):
approving since we discussed on slack that the bug I commented about seems dev server specific.

PR #14541, opened 2024-02-22
fix(app): Center app and robot update modals

mjhuff:
Closes [RQA-2349](https://opentrons.atlassian.net/browse/RQA-2349)

# Overview

This PR fixes an issue in which the app and robot update modals are not properly centered (as a bonus, the module setup modal is also properly centered). 

There's a lot we can do to make this issue less likely to occur in the future, and I've added a note to revisit this topic in the [tech debt paydown doc](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3917479962/App+UI+Pay+off+Tech+Debt+Improve+Codebase). Because we don't have standardized modal sizing, custom modal widths require a custom left margin offset in order to be properly centered. 

At a bare minimum, adding testing to custom width/left-margin modals is a good idea. This at least shows we've gone out of the way to center the modal properly, and anyone who tries to adjust width only will fail a test and have to re-center the modal.

A better, longer-term solution off the top of my head would be to standardize modal sizes (convo w/ design).

# Test Plan
- Downloaded [this build of the app. ](https://opentrons.slack.com/archives/C81CR4VCZ/p1708619272432719)
- Verified that the app update modal is centered (select the alpha channel if you aren't already on it).
- Opened up a robot update modal and verify the robot update modal is centered.
- Verified that no other modals were negatively impacted by this change.
- If you'd like to try this yourself and you don't have a robot, you may have to finagle the code a bit to get the banners to show, but it is possible!

# Changelog
- Desktop app update modals are now properly centered.

# Risk assessment
low

[RQA-2349]: https://opentrons.atlassian.net/browse/RQA-2349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/assets/localization/en/device_settings.json
+++ b/app/src/assets/localization/en/device_settings.json
@@ -301,7 +301,7 @@
   "update_robot_software_description": "Bypass the Opentrons App auto-update process and update the robot software manually.",
   "update_robot_software_link": "Launch Opentrons software update page",
   "updating": "Updating",
-  "updating_robot_system": "Updating the robot software requires restarting the robot",
+  "update_requires_restarting": "Updating the robot software requires restarting the robot",
   "usage_settings": "Usage Settings",
   "usb": "USB",
   "usb_to_ethernet_description": "Looking for USB-to-Ethernet Adapter info?",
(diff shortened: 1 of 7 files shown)


koji (approved):
LGTM!

PR #14543, opened 2024-02-23
refactor(robot-server): Remove most __init__.py re-exports

SyntaxColoring:
# Overview

Rip off the band-aid in `robot-server`: remove re-exports from most `__init__.py` files, following [our new Python guidelines](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3997401136/Imports+and+module+structure). See there for a rationale.

Testing with the same command as #14480 suggests this will provide a slight speedup from ~31 sec. to ~29 sec. I don't consider that worthwhile to merge into `chore_release-7.2.0`, so this is targeted to `edge`. This is really more intended to set a consistent pattern to follow, rather than make any real performance improvements. Real performance improvements will come from doing this in `api`.

# Test plan

We're covered by automated tests.

# Review requests

As described in [our new Python guidelines](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3997401136/Imports+and+module+structure), we should now use `_underscore_prefixes` to indicate that a module is private. I've done that inside `robot_server.persistence` because I just worked on that code and I'm familiar with it, but I haven't done it elsewhere. So this PR might accidentally make a bunch of files look more public than before. Is that acceptable, or do we want to put more effort into tidying that up right now?

# Risk assessment

Very low.


sfoster1 (approved):
Very very nice

PR #14545, opened 2024-02-23
fix(app-shell, app-shell-odd): Fix notification fallback for multiple subscriptions

mjhuff:
Closes [RQA-2376](https://opentrons.atlassian.net/browse/RQA-2376)

# Overview

This PR fixes a bug in which the notification connection manager assumes that a subscription request to a topic is successful if a previous request to the same topic has already been made. This assumption does not hold when multiple subscription requests to the same topic occur at the same time or when a subsequent request is made but before the first subscription request receives a response. 

This was causing the "fallback to HTTP polling" logic to break noticeably during protocol runs - the current run step was not properly updating. This impacts the ODD and the desktop app (for the Flexes without MQTT running and all OT-2s). 

### Current Behavior

IMG_3778

### With Fix

Screenshot 2024-02-23 at 3 56 49 PM

# Test Plan
- Run a protocol on a Flex with the mosquitto service stopped. Alternatively, run a protocol on the OT-2. 
- Notice that in the app, the run step will still properly update. You can confirm this by observing the network requests to `commands?pageLength=1` every 3 seconds.
- If you'd like, you can push app-shell-odd to a Flex and note that the proper run command is displayed as well. Again, make sure the mosquitto service is stopped.

# Changelog
- Fixed a bug causing the current run step not to display when the MQTT server is not functioning. 

# Risk assessment
low

[RQA-2376]: https://opentrons.atlassian.net/browse/RQA-2376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -115,24 +115,21 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
   // true if the connection store has an entry for the hostname.
   else {
     const subscriptions = connectionStore[hostname]?.subscriptions
-    if (subscriptions && subscriptions[topic] > 0) {
-      subscriptions[topic] += 1
-      return Promise.resolve()
-    } else {
-      if (subscriptions) {
-        subscriptions[topic] = 1
-      }
-      return new Promise<void>(() => checkIfClientConnected()).catch(() => {
-        log.warn(`Failed to subscribe on ${hostname} to topic: ${topic}`)
-        sendToBrowserDeserialized({
-          browserWindow,
-          hostname,
-          topic,
-          message: 'ECONNFAILED',
-        })
-        handleDecrementSubscriptionCount(hostname, topic)
-      })
+    if (subscriptions) {
+      if (subscriptions[topic] > 0) subscriptions[topic] += 1
+      else subscriptions[topic] = 1
     }
+
+    return checkIfClientConnected().catch(() => {
+      log.warn(`Failed to subscribe on ${hostname} to topic: ${topic}`)
+      sendToBrowserDeserialized({
+        browserWindow,
+        hostname,
+        topic,
+        message: 'ECONNFAILED',
+      })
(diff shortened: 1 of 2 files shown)


app-shell-odd/src/notify.ts
```diff
@@ -115,24 +115,21 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
   // true if the connection store has an entry for the hostname.
   else {
     const subscriptions = connectionStore[hostname]?.subscriptions
-    if (subscriptions && subscriptions[topic] > 0) {
-      subscriptions[topic] += 1
-      return Promise.resolve()
```
mjhuff, line 120: This was the actual bug. We shouldn't resolve here - every subscription should ensure that we actually have a connection established independently. Everything else is just a promise refactor. There were unhandled promise rejections getting thrown in the app-shell logs that made sense to clean up.

app-shell-odd/src/notify.ts
```diff
@@ -151,36 +148,38 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
+          reject(
+            new Error(
+              `Maximum number of subscription retries reached for hostname: ${hostname}`
+            )
           )
-        )
-      }
-    }, 500)
+        }
+      }, 500)
```
koji: nit

maybe define a const for this?

app-shell/src/notify.ts
```diff
@@ -147,36 +144,38 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
+          reject(
+            new Error(
+              `Maximum number of subscription retries reached for hostname: ${hostname}`
+            )
           )
-        )
-      }
-    }, 500)
+        }
+      }, 500)
```
koji: same

koji (approved):
The changes make sense to me.
The comment I left is nit.

PR #14419, opened 2024-02-02
feat(api): publish commands for trash bins, waste chutes, and moving labware for Opentrons Simulate

jbleon95:
# Overview

Closes RSS-416 and RSS-377

This PR adds the `publish_context` context manager and associated infrastructure to cover `drop_tip`, `dispense`, `blow_out`, `move_to` for waste chutes and 2.16 and above trash bins, as well as `move_labware` in general.

The most practical and immediate effect of this is that Opentrons Simulate will now correctly log these commands when run in CLI. This will allow a better representation of the simulated protocol. These are _not_ hooked up to the `LegacyCommandMapper`, as these commands are not legacy commands and don't need to go through that class.

# Test Plan

Running the following protocols produces the following logs when run with opentrons simulate (`pipenv run python -m opentrons.simulate`)

```
metadata = {
    'protocolName': 'Trash Bin and Waste Chute Methods',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.16"
}


def run(context):
    trash_bin_A1 = context.load_trash_bin('A1')
    trash_bin_D1 = context.load_trash_bin('D1')
    waste_chute = context.load_waste_chute()

    tiprack = context.load_labware('opentrons_flex_96_tiprack_50ul', 'A3')
    plate = context.load_labware('opentrons_96_wellplate_200ul_pcr_full_skirt', 'C2')

    pipette = context.load_instrument('flex_1channel_50', 'left', tip_racks=[tiprack])

    pipette.pick_up_tip()
    pipette.move_to(trash_bin_A1)
    pipette.aspirate(20, plate['A1'].top())
    pipette.dispense(20, trash_bin_A1)
    pipette.blow_out(trash_bin_A1)
    pipette.drop_tip(trash_bin_A1)

    pipette.pick_up_tip()
    pipette.move_to(trash_bin_D1)
    pipette.aspirate(20, plate['A1'].top())
    pipette.dispense(20, trash_bin_D1)
    pipette.blow_out(trash_bin_D1)
    pipette.drop_tip(trash_bin_D1)

    pipette.pick_up_tip()
    pipette.move_to(waste_chute)
    pipette.aspirate(20, plate['A1'].top())
    pipette.dispense(20, waste_chute)
    pipette.blow_out(waste_chute)
    pipette.drop_tip(waste_chute)

    context.move_labware(plate, waste_chute, use_gripper=True)
```

produces

```
Belt calibration not found.
[...]

PR #14546, opened 2024-02-23
fix(app): don't publish "Dropping tip into…" for steps that don't drop tips

ecormany:
# Overview

Run log entries in the Opentrons App for movement to trash bins read as "Dropping tip into" instead of "Moving to". In other words, they would say tips were being dropped even when they weren't, like when dispensing or blowing out into the trash.

This is a string fix band-aid 🩹 just to make the run log sensical and correct. We would probably like these commands to behave more like the `opentrons_simulate` output introduced in #14419 (i.e., combining the movement and action step into a single line). But that's for another day and another release.

# Test Plan

- Tested in app with sample protocol from #14419.
- Unit tests pass.

# Changelog

- Changed `move_to_addressable_area_drop_tip`  localization string and corresponding test.

# Review requests

OK solution for now? Does this break any other publishing? It shouldn't as far as I know.

# Risk assessment

low

### DIFF
--- a/app/src/assets/localization/en/protocol_command_text.json
+++ b/app/src/assets/localization/en/protocol_command_text.json
@@ -37,7 +37,7 @@
   "move_to_slot": "Moving to Slot {{slot_name}}",
   "move_to_well": "Moving to well {{well_name}} of {{labware}} in {{labware_location}}",
   "move_to_addressable_area": "Moving to {{addressable_area}}",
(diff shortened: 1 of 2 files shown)


CaseyBatten (approved):
Clean and simple, I like it! It is likely this oversight occurred with the addition of "move to addressable area for drop tip" as a unique command of its own, as that command is meant to be the preamble for a drop tip (but doesn't actually drop the tip!).

PR #14547, opened 2024-02-23
feat(api): add pipette outer dimensions when checking movement conflict

sanni-t:
Closes RQA-2311
Closes RSS-490

# Overview

Does two things-
1. shared-data: adds (to pipette definitions) the offsets of back left and front right corners of all pipettes' outer casing from their respective mount position
2. api: updates the pipette movement conflict checker to use the above mentioned pipette bounds that include the outer casing of pipettes, instead of just the bounding nozzles.

# Test Plan

Main objective of this PR is to make sure a conflict with thermocycler lid is caught in analysis. 

On top of that, since this PR also updates the logic for checking potential collisions, make sure that previously failing cases aren't passing now and vice versa. Good news is that previous integration tests for pipette movement conflict are still passing as expected.

Test protocol:

```

requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(ctx):
  tip_rack1 = ctx.load_labware("opentrons_flex_96_tiprack_50ul", "B3", adapter="opentrons_flex_96_tiprack_adapter")
  tip_rack2 = ctx.load_labware("opentrons_flex_96_tiprack_50ul", "D3")
  instrument = ctx.load_instrument('flex_96channel_1000', mount="left")
  
  my_pcr_plate = ctx.load_labware('nest_96_wellplate_200ul_flat', "C2")
  my_other_plate = ctx.load_labware('nest_96_wellplate_200ul_flat', "C1")

  thermocycler = ctx.load_module('thermocyclerModuleV2')
  tc_adjacent_plate = ctx.load_labware("nest_96_wellplate_200ul_flat", "A2")
  ctx.load_trash_bin("A3")
  
  instrument.configure_nozzle_layout(style=COLUMN, start="A12", tip_racks=[tip_rack2])
  
  instrument.pick_up_tip()  
  instrument.dispense(20, my_other_plate.wells_by_name()["A2"])
  
  # Should error out because conflict with thermocycler lid
  instrument.dispense(20, tc_adjacent_plate.wells_by_name()["A1"])

  instrument.drop_tip()
```

# Changelog

Shared-data:
- added `pipetteBoundingBoxOffsets` to pipette definitions

api:
- updated `PipetteDict` to include pipette bounding box offsets
- added pipette bounding box offsets to engine pipette state's config data
- replaced bounding nozzles fetchers with pipette bounding box fetchers
- updated the logic that checks whether a pipette is over/ overlapping with a certain slot. 

# Review requests

- any objections to this addition of `pipetteBoundingBoxOffsets` to the definitions?
- anything I'm missing with the changes in hardware controller?

# Risk assessment

Medium. Changes a pretty central part of the pipette movement conflict checker but it's well-tested and only used during 96 channel partial config.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/pipettes.py
+++ b/api/src/opentrons/protocol_engine/state/pipettes.py
@@ -725,6 +725,8 @@ class PipetteView(HasState[PipetteState]):
             - primary_nozzle_offset
             + pipette_bounds_offsets.front_right_corner
         )
+        # TODO (spp, 2024-02-27): remove back right & front left;
+        #  return only back left and front right points.
         pip_back_right_bound = Point(
             pip_front_right_bound.x, pip_back_left_bound.y, pip_front_right_bound.z
         )
--- a/api/src/opentrons/protocol_engine/state/pipettes.py
+++ b/api/src/opentrons/protocol_engine/state/pipettes.py
@@ -703,7 +703,7 @@ class PipetteView(HasState[PipetteState]):
         pipette_id: str,
         destination_position: Point,
     ) -> Tuple[Point, Point, Point, Point]:
-        """Get the given pipette's bounding nozzles' positions when primary nozzle is at the given destination position."""
+        """Get the pipette's bounding offsets when primary nozzle is at the given position."""
         primary_nozzle_offset = self.get_primary_nozzle_offset(pipette_id)
         tip = self.get_attached_tip(pipette_id)
         # Primary nozzle position at destination, in deck coordinates
--- a/api/src/opentrons/protocol_engine/state/pipettes.py
+++ b/api/src/opentrons/protocol_engine/state/pipettes.py
@@ -715,7 +715,6 @@ class PipetteView(HasState[PipetteState]):
         pipette_bounds_offsets = self.get_config(
             pipette_id
         ).pipette_bounding_box_offsets
-        # TODO (spp): add a margin to these bounds
         pip_back_left_bound = (
             primary_nozzle_position
             - primary_nozzle_offset
--- a/api/src/opentrons/protocol_engine/state/pipettes.py
+++ b/api/src/opentrons/protocol_engine/state/pipettes.py
@@ -87,7 +87,7 @@ class BoundingNozzlesOffsets:
 
 @dataclass(frozen=True)
 class PipetteBoundingBoxOffsets:
-    """Offsets of the bounding nozzles of the pipette."""
+    """Offsets of the corners of the pipette's bounding box."""
 
     back_left_corner: Point
     front_right_corner: Point
--- a/api/src/opentrons/protocol_engine/state/pipettes.py
+++ b/api/src/opentrons/protocol_engine/state/pipettes.py
@@ -86,6 +86,14 @@ class BoundingNozzlesOffsets:
 
 
 @dataclass(frozen=True)
+class PipetteBoundingBoxOffsets:
+    """Offsets of the bounding nozzles of the pipette."""
+
+    back_left_corner: Point
+    front_right_corner: Point
+
+
+@dataclass(frozen=True)
 class StaticPipetteConfig:
     """Static config for a pipette."""
 
@@ -101,6 +109,7 @@ class StaticPipetteConfig:
     nominal_tip_overlap: Dict[str, float]
     home_position: float
     nozzle_offset_z: float
+    pipette_bounding_box_offsets: PipetteBoundingBoxOffsets
     bounding_nozzle_offsets: BoundingNozzlesOffsets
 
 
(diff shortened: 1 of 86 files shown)


api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -101,6 +109,7 @@ class StaticPipetteConfig:
     nominal_tip_overlap: Dict[str, float]
     home_position: float
     nozzle_offset_z: float
+    pipette_bounding_box_offsets: PipetteBoundingBoxOffsets
     bounding_nozzle_offsets: BoundingNozzlesOffsets
```
sanni-t, line 113: the bounding nozzle offsets are no longer needed and will be removed in a follow-up. Not doing it in this PR since it requires changes made to the pipette state store in edge.

shared-data/pipette/schemas/2/pipetteGeometrySchema.json
```diff
@@ -32,6 +32,14 @@
       "description": "The path to a valid Opentrons shared schema relative to the shared-data directory, without its extension. For instance, #/pipette/schemas/2/pipetteGeometrySchema.json is a reference to this schema."
     },
     "nozzleOffset": { "$ref": "#/definitions/xyzArray" },
+    "pipetteBoundingBoxOffsets": {
+      "description": "Offsets from the pipette mount position to the back-left & front-right corners of the pipette's casing around the nozzles.",
```
SyntaxColoring: From your numbers, it looks like you're wrapping the bounding box around the casing *and* the nozzles, but this description makes it sound like it's supposed to just wrap around the casing.

Totally up to you how it should be, but if it's intended to wrap around both the casing *and* the nozzles, here's a suggested rewording:

```suggestion
      "description": "Offsets from the pipette mount position to the back-left & front-right corners of the bounding box containing the whole pipette, including its nozzles and casing.",
```

shared-data/pipette/definitions/2/geometry/ninety_six_channel/p1000/1_0.json
```diff
@@ -2,6 +2,10 @@
   "$otSharedSchema": "#/pipette/schemas/2/pipetteGeometrySchema.json",
   "pathTo3D": "pipette/definitions/2/geometry/ninety_six_channel/p1000/placeholder.gltf",
   "nozzleOffset": [-36.0, -25.5, -259.15],
+  "pipetteBoundingBoxOffsets": {
+    "backLeftCorner": [-67.0, -3.5, -259.15],
```
SyntaxColoring, line 6: The -3.5 here is kind of surprising to me. The back of the 96-channel is flat, so it doesn't jut in the +y direction any farther than the surface of the mount that it's attached to. I guess it depends on how "the position of the mount" is defined, but I guess I would have expected this to be 0. Does -3.5 make sense to you?

SyntaxColoring (commented):
Thank you! The approach here looks really good.

I'm still working my way through this, but here are some preliminary questions.

shared-data/pipette/definitions/2/geometry/ninety_six_channel/p1000/1_0.json
sanni-t: Good catch! Let me find out

sanni-t: OK, so there are a couple of attachment structures that do jut out from the back of the 96-channel in the +y direction. I am guessing that makes the pipette sit 3.5mm away from the mounting panel or the mounting position of the pipette. Will verify with the embedded team.

SyntaxColoring: Discussed with Seth: it's the mount plate, which adds a small gap between the mount surface and the pipette casing.

sanni-t:
Analysis of the test protocol errors out as expected.

Tested on Flex with 96 channel & thermocycler. Running the test protocol makes the run fail before moving to the conflict location. So it's safe to say that pipette location data fetching during actual execution is working correctly too.

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -85,6 +85,14 @@ class BoundingNozzlesOffsets:
     front_right_offset: Point
 
 
+@dataclass(frozen=True)
+class PipetteBoundingBoxOffsets:
+    """Offsets of the bounding nozzles of the pipette."""
```
SyntaxColoring:
```suggestion
    """Corners of the bounding box of the pipette."""
```

SyntaxColoring (approved):
Looks great, pending the documentation comments above. Thank you!

api/src/opentrons/protocol_engine/state/pipettes.py
sanni-t: Hmm, they are offsets tho. How about-
```suggestion
    """Offsets of the corners of the pipette's bounding box."""
```
?

PR #14548, opened 2024-02-26
fix(app, app-shell, app-shell-odd): Fix host context for notifications

mjhuff:
Closes [RAUT-1018](https://opentrons.atlassian.net/browse/RAUT-1018)

# Overview

This PR contains a few fixes for a few spots in the app that did not properly pass a host object to the notification utility, therefore causing the app to use polling instead of MQTT. 

Also, this updates MQTT analytics reporting - it's necessary to only report Flexes, because currently all OT-2s will report that MQTT is blocked (there's no MQTT on OT-2s). Filtering out false positives for the 7.2 release would be nice!

# Test Plan
- Verified that maintenance runs flows are still accessible on the ODD under the InstrumentDetails page (and utilizing MQTT). 
- Verified slideouts use MQTT as well when possible. 
- Verified via Mixpanel that OT-2s are no longer reporting. 
- Verified that an annoying shell message no longer displays. 

# Changelog

# Risk assessment
low

[RAUT-1018]: https://opentrons.atlassian.net/browse/RAUT-1018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -193,8 +193,8 @@ const RENDER_TIMEOUT = 10000 // 10 seconds
 function unsubscribe(notifyParams: NotifyParams): Promise<void> {
   const { hostname, topic } = notifyParams
   return new Promise<void>(() => {
-    if (hostname in connectionStore) {
-      setTimeout(() => {
+    setTimeout(() => {
+      if (hostname in connectionStore) {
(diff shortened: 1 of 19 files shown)


koji (approved):
The changes look good to me.

PR #14550, opened 2024-02-26
refactor(app-shell, app-shell-odd): reduce port block analytics reporting frequency

mjhuff:
Closes [RQA-2385](https://opentrons.atlassian.net/browse/RQA-2385)

# Overview

Instead of reporting every port block event that can happen in an app session (which can be excessive), let's just report one event per session.

# Test Plan
- Verified on my Mixpanel acct that port block events aren't excessive.

# Review requests
- Just verify the logic is sound. I think reporting this event once per app session is fine. At the end of the day, we just want to use this number to gauge the # of users who experience port blocking issues. We can filter by robot, ID, etc. in Mixpanel to fine tune the results. Let's just cut down on the overall chatter here. 

# Risk assessment
low

[RQA-2385]: https://opentrons.atlassian.net/browse/RQA-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -15,6 +15,11 @@ import type { Action, Dispatch } from './types'
 // However, redundant subscriptions are permitted and result in the broker sending the retained message for that topic.
 // To mitigate redundant connections, the connection manager eagerly adds the host, removing the host if the connection fails.
 
+const FAILURE_STATUSES = {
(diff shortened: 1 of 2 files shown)


koji (approved):
The changes look good to me.

PR #14551, opened 2024-02-26
chore(release): Add release notes for RSS-468

SyntaxColoring:
# Overview

Add a release note for RSS-468 / PR #14512.

# Review requests

* Can this be made less wordy?
* Do we say "OT-2" or "The OT-2" or "OT-2 robots" these days?
* Is the Capitalization of "Labware Position Check" and "Tip Length Calibration" Appropriate?

# Risk assessment

No risk.

### DIFF
--- a/api/release-notes.md
+++ b/api/release-notes.md
@@ -23,6 +23,7 @@ If you don't care about preserving your labware offsets and run history, you can
 ### Bug Fixes
 
 - The OT-2 now consistently applies tip length calibration. There used to be a height discrepancy between Labware Position Check and protocol runs. If you previously compensated for the inconsistent pipette height with labware offsets, re-run Labware Position Check to avoid pipette crashes.
+- The OT-2 now accurately calculates the position of the Thermocycler. If you previously compensated for the incorrect position with labware offsets, re-run Labware Position Check to avoid pipette crashes.
 - The Flex Gripper will no longer pick up large labware that could collide with tips held by an adjoining pipette.
 - Flex now properly configures itself when connected by Ethernet directly to a computer.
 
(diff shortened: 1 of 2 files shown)


ecormany (commented):
> Can this be made less wordy?

pushed a commit with copy edits.

> Do we say "OT-2" or "The OT-2" or "OT-2 robots" these days?

went with "The OT-2" for parallelism with https://github.com/Opentrons/opentrons/blob/bfd31454c6c67a9c307a355333e984225d3c60c2/app-shell/build/release-notes.md?plain=1#L26
no hard and fast rule though.

> Is the Capitalization of "Labware Position Check" and "Tip Length Calibration" Appropriate?

title case for Labware Position Check. lowercase for tip length calibration. at least per the [Glossary](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/917274747/Opentrons+Glossary)

**remaining question**
that line cited above ☝️ is in the `app-shell` notes. i think that was because of the focus on "hey, re-run LPC, an app feature." this is a pretty similar fix / required user action. so it should probably be moved or copied to the other notes file as well.

jwwojak (commented):
This is reasonable and clear. LGTM.
:shipit:

SyntaxColoring:
> **remaining question** that line cited above ☝️ is in the `app-shell` notes. i think that was because of the focus on "hey, re-run LPC, an app feature." this is a pretty similar fix / required user action. so it should probably be moved or copied to the other notes file as well.

I probably wouldn't have added the Thermocycler thing to the `app-shell` notes. Taking that for granted, though, it sounds like both notes should be in both places, so I've copied them.

Another vote for merging the release note files, I guess.

ecormany (approved):
Looks good 👯 

Further votes on _the future of release notes_ can go in RTC-135.

PR #14563, opened 2024-02-28
fix(api): Add proper tip mismatch for flex

Laura-Danielle:
# Overview
Turns out we never actually updated the tiprack mismatch warning in the python api to handle flex pipette capabilities. Adds in the proper tip types and a test.

### DIFF
--- a/api/src/opentrons/protocols/api_support/instrument.py
+++ b/api/src/opentrons/protocols/api_support/instrument.py
@@ -73,11 +73,14 @@ def tip_length_for(
 
 
 VALID_PIP_TIPRACK_VOL = {
-    "p10": [10, 20],
-    "p20": [10, 20],
-    "p50": [50, 200, 300],
-    "p300": [200, 300],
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Looks good, though it strikes me as being perfect for having a warning rather than error.

ecormany:
I know this is going into `edge`, but we need to do some significant docs and comms before putting a hard stop to using 200 and 1000 tips with the 50 pipette.

SyntaxColoring (approved):
Sweet, thank you! Appreciate the `caplog` test. The string parsing makes me nervous, but I guess it worked like that before, and it doesn't need to be addressed right now.

SyntaxColoring:
> Looks good, though it strikes me as being perfect for having a warning rather than error.

It is a warning, unless I'm missing something. (I said aloud at one point that this was an error, but I was definitely wrong, so I’m sorry if I’ve caused any confusion here.)

PR #14570, opened 2024-02-29
fix(api): Do not pass loop arg to subproc

sfoster1:
asyncio subprocess used to need loop; now it doesn't. For a minimum-affected fix, just don't pass loop to the subprocess invocations while keeping it in the top level environment.

Closes RQA-2417


Laura-Danielle (approved):
you are the weakest loop, goodbye

DerekMaggio:
Installed this branch on the robot, was able to update my Heater-Shaker successfully. Thanks!

PR #14571, opened 2024-02-29
fix(app): fix desktop app attach flow for the 96ch

mjhuff:
Closes [RQA-2408](https://opentrons.atlassian.net/browse/RQA-2408)

# Overview

The right pipette card on the desktop app conditionally renders based on whether or not a 96 channel is attached. This poses an issue when a user tries to attach a 96ch from the right card, since the card will unrender as soon as the pipette is recognized, unrendering the pipette wizard as well. This is a natural spot for our Nice Modal library, which decouples the conditional logic for rendering the modal from the logic persisting the modal.

I also randomly refactored a Nice Modal test that was testing things it didn't need to test and causing issues.

# Test Plan
- Run the attach flow for the 96ch on the desktop app from the _right pipette card_. Verify the flow works.
- Feel free to smoke test other pipette flows from the pipette cards as well.

# Changelog
- Attaching the 96ch pipette from the overflow menu of the right pipette card on the desktop app will no longer cause the pipette wizard to disppear as soon as the 96ch pipette is attached.

# Risk assessment
low

[RQA-2408]: https://opentrons.atlassian.net/browse/RQA-2408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/Devices/PipetteCard/index.tsx
+++ b/app/src/organisms/Devices/PipetteCard/index.tsx
@@ -26,6 +26,7 @@ import {
 import {
   useCurrentSubsystemUpdateQuery,
   usePipetteSettingsQuery,
+  useHost,
 } from '@opentrons/react-api-client'
 
 import { LEFT } from '../../../redux/pipettes'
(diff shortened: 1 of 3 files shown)


smb2268 (approved):
Tested this out and it works as expected. Changes look good - thanks for fixing!

PR #14575, opened 2024-02-29
fix(app): change firmware update progress bar to spinner on ODD modals

smb2268:
fix RQA-2314

# Overview

We previously displayed a progress bar for each firmware item being updated on the ODD takeover modal. This caused some visual confusion because multiple items would be updated in sequence and the bar would flip from empty to full. Instead, design has requested that we show a spinner.

# Test Plan
Initiate a firmware update by updating the robot or via the takeover modal on the ODD. See that there is now a spinner and no progress bar on the resulting modal.

# Changelog
Swap progress bar for spinner icon in UpdateInProgressModal

# Review requests
Look over changes, they're pretty minor

# Risk assessment
Low


mjhuff (approved):
LGTM! This looks much better.
Screenshot 2024-02-29 at 6 05 06 PM

PR #14576, opened 2024-03-01
chore(app): remove robot calibration link from labware offsets modal

ecormany:
# Overview

A link in the "How labware offsets work" modal goes to a help center page that only covers OT-2. The link is present when setting up Flex or OT-2 protocols. See comments on RQA-2411.

labware offset modal with calibration link

It would be a big lift to rewrite the help center page, and it only provides supplemental information — it won't directly help users run LPC and getting the protocol started.

We decided to remove the link for now.

# Test Plan

It's gone 🕳️ 

labware offset modal without calibration link

# Changelog

- remove `<Link>` 
- update tests
- remove unused localization string

# Review requests

I'm pretty sure everything I touched here is contained to only this modal — would love a double check though.

# Risk assessment

low.

### DIFF
--- a/app/src/assets/localization/en/protocol_setup.json
+++ b/app/src/assets/localization/en/protocol_setup.json
@@ -98,7 +98,6 @@
   "learn_how_it_works": "Learn how it works",
   "learn_more_about_offset_data": "Learn more about Labware Offset Data",
   "learn_more_about_robot_cal_link": "Learn more about robot calibration",
-  "learn_more_about_robot_cal_offset": "Learn more about Robot Calibration",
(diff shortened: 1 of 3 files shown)


mjhuff (commented):
How do we feel about conditionally rendering the link if the robot is an OT-2?

ecormany:
@mjhuff I think auditing all of these outbound help center links and doing some conditionals by robot type would be an interesting and potentially useful project…after we get 7.2 out the door. I took the quick approach so we could get this in today and not add any new logic to QA.

mjhuff (approved):
Makes sense. LGTM!

PR #14579, opened 2024-03-01
Analyzes ABR logs and saves to csv

rclarke0:
# Overview

This PR adds edits to the abr_read_logs.py file. The script reads a folder of run logs, determines which run ids are on the sheet vs in the folder, and then adds missing run log information to the sheet.

# Test Plan

# Changelog

1. Read whole folder of run logs and wrote csv file with run log information : 
Robot | Run_ID | Protocol_Name | Software Version | Date | Start_Time | End_Time | Run_Time (min) | Errors | Error_Code | Error_Type | Error_Instrument | Error_Level | Left Mount | Right Mount | Extension | heaterShakerModuleV1 | temperatureModuleV2 | magneticBlockV1 | thermocyclerModuleV2 
2. Script was edited to only add rows if a new run id was read
3. error level.csv was added and used to fill out error level based off of error code.

# Review requests

Advise if there are any spots in my scripts that can be made more consistent and concise. 
Advise if there are any spots in my script that are too specific to ABR and would make the scrip unusable by other people/robots.

# Risk assessment

Pulling run data while robot is running may slow it down.


hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,224 @@
...
+        "heaterShakerModuleV1",
+        "temperatureModuleV2",
+        "magneticBlockV1",
+        "thermocyclerModuleV2",
+    )
+    all_modules = {key: None for key in modList}
+    for module in file_results.get("modules", []):
+        if isinstance(module, dict) and module.get("model") in modList:
+            try:
+                all_modules[module["model"]] = module["serialNumber"]
```
sfoster1: you can use python [dictionary's `get`](https://docs.python.org/3/library/stdtypes.html#dict.get) method to do this:
```python
all_modules[module["model"]] = module.get("serialNumber", "")
```

the second argument to `get()` is a default value to return if the key wasn't found. So this can replace the `try`/`except`.

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,224 @@
...
+    return all_modules
+
+
+def get_error_info(file_results: Dict[str, Any]) -> Tuple[int, str, str, str, str]:
+    """Determines if errors exist in run log and documents them."""
+    error_levels = []
+    # Read error levels file
+    with open(ERROR_LEVELS_PATH, "r") as error_file:
+        error_levels = list(csv.reader(error_file))
+    num_of_errors: int = len(file_results["errors"])
```
sfoster1:
```suggestion
    num_of_errors = len(file_results["errors"])
```
type checkers are usually smart enough to figure out what type a variable should be when you initialize it by calling a function with a return type. For instance, `len()` can only ever return an `int`, so type checkers can infer that `num_of_errors` must be an int.

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,224 @@
...
+    if num_of_errors > 0:
+        commands_of_run = file_results.get("commands", [])  # type: List[Dict[str, Any]]
```
sfoster1: I'm not sure if this is on purpose, but these comments will be parsed by type checkers - this is how types were done before annotation was introduced. IMO, it's better to have these be annotations, like `commands_of_run: List[Dict[str, any]] = ...`

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def get_error_info(file_results: Dict[str, Any]) -> Tuple[int, str, str, str, str]:
+    """Determines if errors exist in run log and documents them."""
+    error_levels = []
+    # Read error levels file
+    with open(ERROR_LEVELS_PATH, "r") as error_file:
+        error_levels = list(csv.reader(error_file))
+    num_of_errors: int = len(file_results["errors"])
+    if num_of_errors > 0:
```
sfoster1: one thing that can improve readability for scripts here is to invert this logic to something like:
```python
if num_of_errors == 0:
    error_type = ""
    error_code = ""
    error_instrument = ""
    error_level = ""
    return 0, error_type, error_code, error_instrument, error_level, 
```
and then don't even have an `else` - just have the rest of the logic continue. That way, in the more extensive logic that follows, it doesn't have to be nested under that `else`.

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,224 @@
...
+        error_str = len(run_command_error.get("error", ""))  # type: int
+        if error_str > 1:
+            error_type = run_command_error["error"].get("errorType", None)
+            error_code = run_command_error["error"].get("errorCode", None)
+            try:
+                # Instrument Error
+                error_instrument = run_command_error["error"]["errorInfo"].get(
+                    "node", None
+                )
+            except KeyError:
```
sfoster1: I don't think this will do what you want. Because you're using `.get()` when you access `run_command_error["error"]["error_info"]`, if the error info doesn't have a `"node"` key then `error_instrument` will be none but no exception will be thrown. The only time this `except` would trigger is if either `run_command_error` has no key `"error"`, or if `run_command_error["error"]` has no key `"errorInfo"`. Here, to get the behavior you want, you should actually do `error_instrument = run_command_error["error"]["errorInfo"]["node"]` and _not_ use get, since you want the key error.

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,224 @@
...
+        error_instrument = ""
+        error_level = ""
+    return num_of_errors, error_type, error_code, error_instrument, error_level
+
+
+def create_abr_data_sheet(storage_directory: str) -> None:
+    """Creates csv file to log ABR data."""
+    sheet_location = os.path.join(storage_directory, "ABR-run-data.csv")
+    if os.path.exists(sheet_location):
+        print(f"File {sheet_location} located.")
```
sfoster1: This means that if a file by the specified name already exists, you _will not_ write to it. This is probably a good idea, but IMO the print message should say like "not overwriting, exiting" or something.

hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
```diff
@@ -84,26 +84,37 @@ def get_run_data(one_run: Any, ip: str) -> Dict[str, Any]:
...
     return run
 
 
 def save_runs(runs_to_save: Set[str], ip: str, storage_directory: str) -> None:
     """Saves runs to user given storage directory."""
     for a_run in runs_to_save:
         data = get_run_data(a_run, ip)
-        robot_name = data["robot_name"]
-        data_file_name = data["robot_name"] + "_" + data["run_id"] + ".json"
+        # robot_name = data["robot_name"]
```
sfoster1: let's not hceck in commented-out code.

hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
```diff
@@ -84,26 +84,37 @@ def get_run_data(one_run: Any, ip: str) -> Dict[str, Any]:
...
+    run["API_Version"] = health_data["api_version"]
     try:
         robot_serial = health_data["robot_serial"]
-    except UnboundLocalError:
+    except KeyError:
         robot_serial = "unknown"
-    run["robot_name"] = robot_name
+    try:
+        run["robot_name"] = robot_name
+    except UnboundLocalError:
```
sfoster1: this is usually a sign that something should be refactored - what was the error that made you write this?

sfoster1 (changes requested):
This is awesome. I think there's some small changes that could improve it, but really good!

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
rclarke0: added additional print statement

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
rclarke0: thanks for explaining that

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
rclarke0: I was adding that to get past the lint checkers. thanks for explaining best practice for these annotations.

hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
rclarke0: one robot was returning something weird for name but it is not doing that anymore

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
sfoster1: You actually don't even need this `if` on line 40 anymore - `num_of_errors` must be >0, because `len()` only returns `>=0`, and if it were strictly-equals to 0 it would have returned on line 39.

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,221 @@
...
+    num_of_errors = len(file_results["errors"])
+    if num_of_errors == 0:
+        error_type = ""
+        error_code = ""
+        error_instrument = ""
+        error_level = ""
+        return 0, error_type, error_code, error_instrument, error_level
+    if num_of_errors > 0:
+        commands_of_run: List[Dict[str, Any]] = file_results.get("commands", [])
+        run_command_error = commands_of_run[-1]  # type: Dict[str, Any]
```
sfoster1:
```suggestion
        run_command_error: Dict[str, Any] = commands_of_run[-1]
```

hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -0,0 +1,221 @@
...
+        error_str = len(run_command_error.get("error", ""))  # type: int
```
sfoster1:
```suggestion
        error_str = len(run_command_error.get("error", ""))
```
(should be inferred)

sfoster1 (approved):
Couple more small things on the same themes that I think we could improve, but definitely ok with merging it even if you don't!

PR #14580, opened 2024-03-01
fix(app): handle error during exit from pipette wizard flow

ncdiehl11:
closes [RQA-2340](https://opentrons.atlassian.net/browse/RQA-2340)

# Overview

Pipette wizard flows need to be able to handle errors after exiting, namely, during homing. Behavior is updated here to surface the error reason to the user and expose only the option to exit the modal, without issuing another home command. The modal immediately closes and deletes the maintenance run on this final exit click.

# Test Plan

- begin a pipette wizard flow (e.g. attach pipette)
- from any step, exit the wizard flow
- create or simulate an error as the gantry attempts to home. this can be done by copying the following into the promise `catch` block on line 201:
```
        .catch(error => {
          setIsExiting(true)
          setShowErrorMessage(error.message)
        })
```
- verify error screen
- click exit again
- verify that modal closes immediately

# RIsk Assessment
- medium

[RQA-2340]: https://opentrons.atlassian.net/browse/RQA-2340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/PipetteWizardFlows/index.tsx
+++ b/app/src/organisms/PipetteWizardFlows/index.tsx
@@ -182,7 +182,9 @@ export const PipetteWizardFlows = (
   }
   const handleClose = (): void => {
     if (onComplete != null) onComplete()
-    deleteMaintenanceRun(maintenanceRunData?.data.id)
+    if (maintenanceRunData != null) {
+      deleteMaintenanceRun(maintenanceRunData?.data.id)
+    }
     closeFlow()
   }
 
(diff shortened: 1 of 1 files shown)


app/src/organisms/PipetteWizardFlows/index.tsx
```diff
@@ -181,18 +181,14 @@ export const PipetteWizardFlows = (
     }
   }
   const handleClose = (): void => {
-    setIsExiting(false)
-    closeFlow()
     if (onComplete != null) onComplete()
+    deleteMaintenanceRun(maintenanceRunData?.data.id)
```
smb2268:
```suggestion
    if (maintenanceRunData != null) deleteMaintenanceRun(maintenanceRunData?.data.id)
```

smb2268 (approved):
Changes look good! Saw this tested out and I think this is the best approach.

PR #14581, opened 2024-03-01
fix(app): poll modules from protocol run module controls

smb2268:
fix RQA-2419

# Overview
Once a run is completed, the module cards on the controls section of the run view were not updating to reflect the most recent module information. This is because we were only polling for modules in the Setup section, which is disabled after the run begins. To fix this, I've enabled polling for modules from the Module Controls tab.

# Test Plan
1. Run a protocol that uses modules - I've linked one in the ticket that I tested earlier today on app&ui bot
2. Once the run has finished, navigate to the Module Controls tab
3. Take action with one of the modules by disabling it, setting temp, etc
4. See that the module card updates with the changing information
5. In the network tab of dev tools, you should see a `/modules` response every 5 seconds. If you dismiss the run by X'ing the run completed banner, the Module Controls tab should become disabled and this poll should stop

# Changelog
Add optional `pollModules` param to `useModuleRenderInfoForProtcolById` that will poll `useAttachedModules` every 5 seconds when true 
Pass this prop as `true` from `ProtocolRunModuleControls`

# Review requests
Test this out and look over change

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/ProtocolRunModuleControls.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/ProtocolRunModuleControls.tsx
@@ -82,7 +82,8 @@ export const ProtocolRunModuleControls = ({
   } = usePipetteIsReady()
 
   const moduleRenderInfoForProtocolById = useModuleRenderInfoForProtocolById(
-    runId
+    runId,
+    true
   )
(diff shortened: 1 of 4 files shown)


mjhuff (approved):
Nice!

PR #14583, opened 2024-03-01
Shenzhen Scale Script

rclarke0:
# Overview

Use Scale in Shenzhen to get stable reading

# Test Plan

# Changelog
Looks for scale
Reads stable_value, mass
Logs stable_value, mass to csv
stops logging when stable value is reached.

# Review requests

# Risk assessment

Concerned shenzhen will have a different vid pid for the scale.


hardware-testing/hardware_testing/scripts/shenzhen_scale.py
```diff
@@ -0,0 +1,36 @@
...
+    results_list = []
+    while is_stable is False:
+        grams, is_stable = scale.read_mass()
+        print(f"Scale reading: grams={grams}, is_stable={is_stable}")
+        time_now = datetime.datetime.now()
+        row = [time_now, grams, is_stable]
+        results_list.append(row)
+        if is_stable is True:
+            print(f"Scale stable reading {grams}.")
+            break
```
andySigler: Might be nice to let the script just continue running, instead of needing to restart for every reading. Add the following 2x lines would do that I think:
```suggestion
            ui.get_user_ready("to measure weight again")
            is_stable = False
```

andySigler (approved):
LGTM, just added one suggestion for usability

hardware-testing/hardware_testing/scripts/shenzhen_scale.py
rclarke0: I added a few more lines of code to ensure that the file save data would be different from the previous sample. 
`y_or_no = input("Do you want to weight another sample? (Y/N): ")
            if y_or_no == "Y":
                is_stable = False
                save_file_variables = _create_scale_reading_file()
            elif y_or_no == "N":
                break_all = True
        if break_all:
            break`

PR #14587, opened 2024-03-04
test(shared-data): add pressFit pickUpTipConfigurations validation test

sanni-t:
# Overview

I started working towards RSS-496 and as I was updating the definitions to use `configsByTipCount`, I realized that that schema was more prone to human errors and was too verbose. Even though the current schema looks a bit redundant with three properties keyed by tip count, having a property written as: 

```
currentByTipCount: {"1": 0.1, "2": 0.2, "3": 0.3....}
```
as opposed to 
```
configByTipCount: { "1": { "current": 0.1, "speed": 12, "distance": 22}, "2": {.....}}
```

is much more readable, easy to peruse at a quick glance and less prone to human errors because you can spot an anomaly in the pattern right away.

Now the present and easily readable schema still has the problem of tip count keys being missed or out of sync. So this PR adds a test that scans all pipette definitions to verify that each definition has identical keys for `speedByTipCount`, `distanceByTipCount` & `currentByTipCount`.

# Test Plan

Checked that removing a key from any definition raises error in the added test.

# Review requests

Any opposition to not changing the schema?

# Risk assessment

None. Just a test

### DIFF
--- a/shared-data/python/tests/pipette/test_validate_schema.py
+++ b/shared-data/python/tests/pipette/test_validate_schema.py
@@ -35,3 +35,36 @@ def test_check_all_models_are_valid() -> None:
                 )
 
                 assert isinstance(loaded_model, PipetteConfigurations)
+
+
+def test_pick_up_configs_tip_count_keys() -> None:
+    """Verify that speed, distance & current of pickUpTipConfigurations have same tip count keys."""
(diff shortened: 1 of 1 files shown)


SyntaxColoring (approved):
Cool, TY!

PR #14588, opened 2024-03-04
refactor(app-shell, app-shell-odd): reduce superfluous MQTT traffic

mjhuff:
# Overview

This PR reduces superfluous MQTT network traffic on the app side by keeping track of robots that have known blocked MQTT ports (or have an offline message broker). Currently, any time a new component wants to attempt subscription to a blocked host, a new MQTT connection handshake will occur or the component will wait for the handshake to complete if one is already in progress. This creates unnecessary network traffic and latency. 

Instead, keep track of blocked hosts for the app session, and if a host is known to be blocked, immediately tell the component attempting subscription to fallback to the old HTTP polling. This does not account for the edge case when an MQTT server that is blocked and becomes unblocked during the same app session, but the effort to make this work isn't worth the effort considering the connectivity logic will soon be shifted to discovery-client. Forcing the app to use HTTP polling for the remainder of the session is fine.

Although MQTT elegantly handles repeated subscriptions for the same topic, this outcome is not ideal, since it implies that we are sending unneeded traffic over the network. As we do for connection attempts, where we only let the first component that wants to use notifications to initiate a connection to the broker, let's do the exact same thing for subscriptions: only the first component that expresses interest in a topic initiates a subscription request. As with the connection logic, this means that all other components that want to subscribe to a topic while the connection manager is still negotiating the subscription will simply wait for the subscription to complete or timeout after 2 seconds.

# Test Plan
- [x] Verified via Wireshark that the app isn't sending connection packets to known port blocked hosts. You can also search the logs for `Failed to connect to <hostname>` and confirm there's only one entry for the robot, even after you navigate away and back to that robot's device details page.
- [x] Successfully smoke tested all notification flows on the desktop app for the Flex (maintenance run flows and running a protocol with pauses, plays, etc). 
- [x] Successfully smoke tested all notification flows on the ODD for the Flex (maintenance run flows and running a protocol with pauses, plays, etc). 
- [x] Successfully smoke tested all notification flows on the desktop app for the OT-2 (maintenance run flows and running a protocol with pauses, plays, etc). 

# Risk assessment
low

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -25,6 +25,7 @@ interface ConnectionStore {
   [hostname: string]: {
     client: mqtt.MqttClient | null
     subscriptions: Record<NotifyTopic, number>
+    pendingSubs: Set<NotifyTopic>
   }
 }
 
@@ -47,8 +48,8 @@ const connectOptions: mqtt.IClientOptions = {
 
 /**
  * @property {number} qos: "Quality of Service", "at least once". Because we use React Query, which does not trigger
-  a render update event if duplicate data is received, we can avoid the additional overhead 
-  to guarantee "exactly once" delivery. 
+  a render update event if duplicate data is received, we can avoid the additional overhead
+  to guarantee "exactly once" delivery.
  */
 const subscribeOptions: mqtt.IClientSubscribeOptions = {
   qos: 1,
@@ -94,15 +95,19 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
       client: null,
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
       subscriptions: { [topic]: 1 } as Record<NotifyTopic, number>,
+      pendingSubs: new Set(),
     }
     return connectAsync(`mqtt://${hostname}`)
       .then(client => {
+        const { pendingSubs } = connectionStore[hostname]
         log.info(`Successfully connected to ${hostname}`)
         connectionStore[hostname].client = client
+        pendingSubs.add(topic)
         establishListeners({ ...notifyParams, client })
-        return new Promise<void>(() =>
+        return new Promise<void>(() => {
           client.subscribe(topic, subscribeOptions, subscribeCb)
-        )
+          pendingSubs.delete(topic)
+        })
       })
       .catch((error: Error) => {
         log.warn(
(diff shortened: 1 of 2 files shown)


app-shell-odd/src/notify.ts
```diff
@@ -86,36 +88,50 @@ interface NotifyParams {
   topic: NotifyTopic
 }
 
-function subscribe(notifyParams: NotifyParams): Promise<void> {
+function subscribe(notifyParams: NotifyParams): Promise<void> | void {
   const { hostname, topic, browserWindow } = notifyParams
+  if (unreachableHosts.has(hostname)) {
+    sendToBrowserDeserialized({
```
sfoster1, line 94: want to wrap this in an immediately-fulfilling promise just to avoid that kinda gross signature above

app-shell-odd/src/notify.ts
```diff
@@ -156,42 +172,47 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
       })
-      handleDecrementSubscriptionCount(hostname, topic)
+      setTimeout(() => {
+        if (Object.keys(connectionStore[hostname].subscriptions).length <= 0) {
+          connectionStore[hostname].client?.end()
+        }
+      }, RENDER_TIMEOUT)
     } else {
       // log.info(`Successfully subscribed on ${hostname} to topic: ${topic}`)
+      if (subscriptions[topic] > 0) subscriptions[topic] += 1
```
sfoster1: please use curly braces cmon

app-shell-odd/src/notify.ts
```diff
@@ -156,42 +172,47 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
 
         counter++
         if (counter === MAX_RETRIES) {
           clearInterval(intervalId)
-          reject(
-            new Error(
-              `Maximum number of subscription retries reached for hostname: ${hostname}`
-            )
-          )
+          // log.warn(`Failed to subscribe on ${hostname} to topic: ${topic}`)
```
sfoster1, line 230: purposeful?

app-shell-odd/src/notify.ts
```diff
@@ -156,42 +172,47 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
+          sendToBrowserDeserialized({
```
sfoster1: maybe do this in whatever is calling this function in a `catch`? makes the job of this function a little simpler

ncdiehl11 (approved):
inspected with WireShark with @mjhuff and smoketested devices route, device details page, starting protocol setup, etc. 👍

sfoster1 (approved):
We could fix the stuff I've noticed above but overall looks good.

PR #14589, opened 2024-03-04
fix(app-shell, app-shell-odd): Handle HTTP fallback during active subscription

mjhuff:
# Overview

Currently, components know to fallback to HTTP if the broker is unresponsive during the initial connection/subscription request to the MQTT broker. However, there is no logic to handle the edge case of falling back to HTTP during an active subscription (the component successfully subscribes to a topic, but during the active subscription, the broker becomes unresponsive).

This PR implements the appropriate fallback logic by notifying all components subscribed to any topic that they should fallback to HTTP if the broker becomes unresponsive after successfully subscribing. 

# Test Plan
- Start a protocol run on the Flex. [Here is my favorite protocol.](https://github.com/Opentrons/opentrons/files/14484080/GripperMoveTesting.py.zip)
- During the run ssh into the robot and verify the broker is running via `systemctl status mosquitto`.
- After confirming the broker is working, `systemctl stop mosquitto`.
- Verify that the current run step still successfully tracks steps on both the desktop app and ODD. (You can also verify this via the network tab -- polling for the active run command should occur). 

# Changelog
- Notification fallback logic after successfully subscribing to the MQTT broker works properly. 

# Risk assessment
low

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -47,8 +47,8 @@ const connectOptions: mqtt.IClientOptions = {
 
 /**
  * @property {number} qos: "Quality of Service", "at least once". Because we use React Query, which does not trigger
-  a render update event if duplicate data is received, we can avoid the additional overhead 
-  to guarantee "exactly once" delivery. 
+  a render update event if duplicate data is received, we can avoid the additional overhead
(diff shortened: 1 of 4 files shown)


koji (approved):
followed the test plan and it worked as expected.
the changes look good to me

PR #14591, opened 2024-03-04
feat(protocol-designer): add Opentrons Tough PCR plate as compatible …

jerader:
…with aluminum block

closes [AUTH-53 ](https://opentrons.atlassian.net/browse/AUTH-53)

# Overview

A customer noticed that the 96 well aluminum block is only compatible with the biorad and NEST plates but not the Opentrons Tough 96 well plate. So this PR fixes that. Additionally, this plate is already supported in the labware definition so it should work in python already. 

Screen Shot 2024-03-04 at 11 44 55 AM

# Test Plan

Create a flex protocol. Add a temperature module and the gripper. On the temperature module, add the 96 well aluminum block under the adapter selection. Then on the deck, add the same aluminum block. Add the Opentrons tough plate on top of one of them. Then add a move labware step with the gripper to move the labware from one aluminum block to the other.

That should work as expected.

# Changelog

- add the opentrons tough plate as compatible with the 96 well aluminum block both with labware selection and moving the labware

# Review requests

see test plan

# Risk assessment

low


koji (approved):
Followed the test case and the sandbox worked as expected.

PR #14593, opened 2024-03-04
docs(api): section on Position Relative to Trash Containers

ecormany:
# Overview

Follow up to further explain behavior introduced in #14560.

Addresses RTC-400.

# Test Plan

- [sandbox](http://sandbox.docs.opentrons.com/docs-trash-offsets/v2/robot_position.html#position-relative-to-trash-containers)
- simulated snippets in 2.16 / `edge`

# Changelog

- new § on Labware and Deck Positions page
- updated description and example for `move_to()` to cover new acceptable `location`s
- version added statements for `TrashBin`, `WasteChute`, and their methods in API Reference 

# Review requests

- double check code snippets. not only simulate, but do what i say they do?
- confirm version added for the methods. it feels like we may need to gate this to 2.17, because otherwise there will be a big functionality gap between robot system 7.1 (which can't do any of this) and 7.2 (which can). the classes are unambiguously _New in version 2.16._

# Risk assessment

low…but see API version gate caveat above.

### DIFF
--- a/api/docs/v2/robot_position.rst
+++ b/api/docs/v2/robot_position.rst
@@ -123,9 +123,10 @@ You should only adjust labware offsets in your Python code if you plan to run yo
 Position Relative to Trash Containers
 =====================================
 
-In API version 2.15 and earlier, trash containers are :py:class:`.Labware` objects that have a single well. Use the above techniques for labware to adjust position relative to trash containers in protocols specifying these API versions.
+Movement to :py:class:`.TrashBin` or :py:class:`.WasteChute` objects is based on the horizontal *center* of the pipette. This is different than movement to labware, which is based on the primary channel (the back channel on 8-channel pipettes, and the back-left channel on 96-channel pipettes in default configuration). Using the center of the pipette ensures that all attached tips are over the trash container for blowing out, dropping tips, or other disposal operations.
 
-Starting in API version 2.16, trash containers are :py:class:`.TrashBin` or :py:class:`.WasteChute` objects. The API calculates movement to these objects based on the horizontal *center* of the pipette, rather than its primary channel (the back channel on 8-channel pipettes, and the back-left channel on 96-channel pipettes in default configuration). This ensures that all tips attached to the pipette are placed over the trash container for blowing out, dropping tips, or other operations.
+.. note::
+    In API version 2.15 and earlier, trash containers are :py:class:`.Labware` objects that have a single well. See :py:obj:`.fixed_trash` and :ref:`position-relative-labware` above.
 
 You can adjust the position of the pipette center with the :py:meth:`.TrashBin.top` and :py:meth:`.WasteChute.top` methods. These methods allow adjustments along the x-, y-, and z-axes. In contrast, ``Well.top()``, :ref:`covered above <well-top>`, only allows z-axis adjustment. With no adjustments, the "top" position is centered on the x- and y-axes and is just below the opening of the trash container.
 
@@ -140,8 +141,7 @@ You can adjust the position of the pipette center with the :py:meth:`.TrashBin.t
 
 .. versionadded:: 2.18
 
-.. note::
-    Another difference between the trash container ``top()`` methods and ``Well.top()`` is that they return an object of the same type, not a :py:class:`.Location`. This helps prevent performing undesired actions in trash containers. For example, you can :py:meth:`.aspirate` at a location or from a well, but not from a trash container. On the other hand, you can :py:meth:`.blow_out` at a location, well, trash bin, or waste chute.
+Another difference between the trash container ``top()`` methods and ``Well.top()`` is that they return an object of the same type, not a :py:class:`.Location`. This helps prevent performing undesired actions in trash containers. For example, you can :py:meth:`.aspirate` at a location or from a well, but not from a trash container. On the other hand, you can :py:meth:`.blow_out` at a location, well, trash bin, or waste chute.
 
 .. _protocol-api-deck-coords:
 
--- a/api/docs/v2/robot_position.rst
+++ b/api/docs/v2/robot_position.rst
@@ -138,7 +138,7 @@ You can adjust the position of the pipette center with the :py:meth:`.TrashBin.t
     trash.top(z=10)  # 10 mm higher
     trash.top(y=10)  # 10 mm towards back, default height
 
-.. versionadded:: 2.16
+.. versionadded:: 2.18
 
 .. note::
     Another difference between the trash container ``top()`` methods and ``Well.top()`` is that they return an object of the same type, not a :py:class:`.Location`. This helps prevent performing undesired actions in trash containers. For example, you can :py:meth:`.aspirate` at a location or from a well, but not from a trash container. On the other hand, you can :py:meth:`.blow_out` at a location, well, trash bin, or waste chute.
--- a/api/docs/v2/robot_position.rst
+++ b/api/docs/v2/robot_position.rst
@@ -140,16 +140,8 @@ You can adjust the position of the pipette center with the :py:meth:`.TrashBin.t
 
 .. versionadded:: 2.16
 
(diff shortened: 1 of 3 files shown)


api/docs/v2/robot_position.rst
```diff
@@ -116,6 +116,39 @@ All positions relative to labware are adjusted automatically based on labware of
...
+=====================================
+
+In API version 2.15 and earlier, trash containers are :py:class:`.Labware` objects that have a single well. Use the above techniques for labware to adjust position relative to trash containers in protocols specifying these API versions.
+
+Starting in API version 2.16, trash containers are :py:class:`.TrashBin` or :py:class:`.WasteChute` objects. The API calculates movement to these objects based on the horizontal *center* of the pipette, rather than its primary channel (the back channel on 8-channel pipettes, and the back-left channel on 96-channel pipettes in default configuration). This ensures that all tips attached to the pipette are placed over the trash container for blowing out, dropping tips, or other operations.
+
+You can adjust the position of the pipette center with the :py:meth:`.TrashBin.top` and :py:meth:`.WasteChute.top` methods. These methods allow adjustments along the x-, y-, and z-axes (in constrast to :py:meth:`.Well.top`, which only allows z-axis adjustment). With no adjustments, the "top" position is centered on the x- and y-axes and is just below the opening of the trash container.
```
jwwojak: Spelling typo: "constrast" --> contrast

Wondering about the parenthetical comparison to the `Well.top` method. Could the contrast be stronger if this became its own sentence? Make it stand out some.  "By contrast, the `Well.top` method only allows z-axis adjustments."

Or something like that.

api/docs/v2/robot_position.rst
```diff
@@ -116,6 +118,39 @@ All positions relative to labware are adjusted automatically based on labware of
...
+
+Another difference between the trash container ``top()`` methods and ``Well.top()`` is that they return an object of the same type, not a :py:class:`.Location`. This means that you can use their output anywhere that accepts objects of those types.
+
+One example is the :py:obj:`.InstrumentContext.trash_container` property. Because this property is settable, if you always want to drop tips at an offset from the default location, you can overwrite the default for a given pipette::
+
+    chute = protocol.load_waste_chute()
+
+    pipette.trash_container = chute.top(z=-5)
+    pipette.pick_up_tip()
+    pipette.drop_tip()  # drop in waste chute, 5 mm lower than default
```
jbleon95: This isn't the case with this right now, since currently if you provide no arguments to drop tip, tip drop alternation overrides any set offset. Its only if you explicitly provide that offseted disposal location that we use that.

api/src/opentrons/protocol_api/disposal_locations.py
```diff
@@ -78,6 +78,8 @@ class TrashBin(_DisposalLocation):
     """Represents a Flex or OT-2 trash bin.
 
     See :py:meth:`.ProtocolContext.load_trash_bin`.
+
+    .. versionadded:: 2.16
```
jbleon95: This is targeted for 2.18 in the future 7.3 release (or whatever we end up calling it)

jbleon95 (changes requested):

api/docs/v2/robot_position.rst
ecormany: 🙏 Thank you for flagging this. I wasn't sure if I had overextended here, since neither simulation nor run preview say anything about offsets — just the parent object the offset is applied to.

I'll remove this bit.

api/src/opentrons/protocol_api/disposal_locations.py
ecormany: So, the _class_ is added in 2.16, since we're already using it.

The `top()` method isn't going to be in the next API version though? That would be 2.17.

ecormany: Update: 2.17 indeed went with robot system 7.2.0. So I'll bump these to 2.18. And target this on a new docs feature branch.

ecormany:
OK, after a near-miss rebasing this onto `docs_release.2-18`, this is ready for another review. Not much prose has changed (except some added in the API reference). Version added/changed notices have changed and moved around quite a bit — make sure I've tagged everything properly, especially related features added _prior_ to 2.18.

api/docs/v2/robot_position.rst
```diff
@@ -116,6 +118,31 @@ All positions relative to labware are adjusted automatically based on labware of
 
 You should only adjust labware offsets in your Python code if you plan to run your protocol in Jupyter Notebook or from the command line. See :ref:`using_lpc` in the Advanced Control article for information.
 
+.. _position-relative-trash:
+
+Position Relative to Trash Containers
+=====================================
+
+In API version 2.15 and earlier, trash containers are :py:class:`.Labware` objects that have a single well. Use the above techniques for labware to adjust position relative to trash containers in protocols specifying these API versions.
```
jwwojak: Word choice: "techniques" stands out as unusual to me. The content referred to are methods.

Might it make sense to write something like, "Use the labware _methods_ above to adjust position relative to trash containers ..."

api/docs/v2/robot_position.rst
```diff
@@ -116,6 +118,31 @@ All positions relative to labware are adjusted automatically based on labware of
 
 You should only adjust labware offsets in your Python code if you plan to run your protocol in Jupyter Notebook or from the command line. See :ref:`using_lpc` in the Advanced Control article for information.
 
+.. _position-relative-trash:
+
+Position Relative to Trash Containers
+=====================================
+
```
jwwojak, line 125: I'm putting a comment here because this refers to the intro sentences to the first 2 paragraphs, lines 126 and 128.

While the intros are clear, each paragraph tells the reader that the API treats trash containers differently depending on the API verison. I wonder if we're softening two very important pieces of information by discussing them in simple, unadorned paragraphs. The differences may get lost or overlooked by someone hastily browsing through the docs looking for an answer.

One way to resolve this, and make the 2 paragraphs stand, out is to put them in a table with the version in one column and a description in another. However, this is unusual compared to current practice in the API docs and I can see that it might not be desirable.

Maybe this section needs a short intro sentence immediately following the section title. Something simple like, "Depending on the API version, the API treats trash containers as either generic labware objects or distinct trash bin or waste chute objects." Something to prime the reader along the lines of, "hey you, pay attention here! you might not see the distinctions being made in these paragraphs and you need to."

While the text is fine as is, I still wonder about what could be done to make the differences between these 2 paragraphs pop more.

jwwojak (commented):
Added review comments for consideration. Defer to others re: go/no go approvals.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -306,27 +306,34 @@ def dispense(  # noqa: C901
...
+
+                            - If unspecified, the pipette will
+                              dispense at its current position.
 
                             If only a ``location`` is passed (e.g.,
                             ``pipette.dispense(location=plate['A1'])``), all of the
                             liquid aspirated into the pipette will be dispensed (the
                             amount is accessible through :py:attr:`current_volume`).
 
+            .. versionchanged:: 2.18
```
jbleon95: This and the other methods in instrument context can accept TrashBin and WasteChute starting in version 2.16

jbleon95 (approved):

api/src/opentrons/protocol_api/instrument_context.py
ecormany: I knew I had confused something. Thanks for checking!

api/docs/v2/robot_position.rst
ecormany: Went for a restructure here. Re-requesting a review…

(Also, you can comment on multiple lines. Click the first line, shift-click the last line, then click the `+` _also on the last line_. It's a little fussy, but it works!)

jwwojak: Ed, Yes to the changes here! Great revision. This aligns the text with the latest version functionality. Nice fix. LGTM.

jwwojak (approved):
Great revision.

:shipit:

PR #14597, opened 2024-03-04
fix(api): fix infinite homing and z drops around fw update

sfoster1:
This PR fixes two issues.
1. Often, after updating the firmware of the head or the gripper, homing that subsystem's z axis would not cause the motor to move. The root cause for this is in the firmware, but lessening the distance that we home the axes also makes the errors happen after 2 minutes instead of 20.
2. Sometimes, after updating the firmware of the head or the gripper, that axis would fall down. This is in fact pretty much the same issue - it's to do with mechanics around how we handle the emergency brakes - and is also fixed in firmware; but disabling these axes when we're about to update them will help.

Also, wrap the entirety of `cache_instruments` in the motion lock rather than just sections, which would sometimes lead to part of the calls happening long before the rest of them.

Closes RQA-2301, RQA-2429


ahiuchingau (approved):
Unbelizeable!

PR #14600, opened 2024-03-05
chore(app-shell-odd): update electron builder config

koji:
# Overview
Update electron in builder config file.

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app-shell-odd/electron-builder.config.js
+++ b/app-shell-odd/electron-builder.config.js
@@ -1,8 +1,9 @@
 'use strict'
+const { app } = require('electron')
 
 module.exports = {
   appId: 'com.opentrons.odd',
-  electronVersion: '27.0.0',
+  electronVersion: app.getVersion(),
   npmRebuild: false,
   files: [
     '**/*',
--- a/app-shell-odd/electron-builder.config.js
+++ b/app-shell-odd/electron-builder.config.js
@@ -1,9 +1,8 @@
 'use strict'
-import pkg from '../package.json'
 
 module.exports = {
   appId: 'com.opentrons.odd',
-  electronVersion: JSON.stringify(pkg.devDependencies.electron),
+  electronVersion: '27.0.0',
   npmRebuild: false,
   files: [
     '**/*',
--- a/app-shell-odd/electron-builder.config.js
+++ b/app-shell-odd/electron-builder.config.js
@@ -1,9 +1,9 @@
 'use strict'
-const { app } = require('electron')
+import pkg from '../package.json'
 
 module.exports = {
   appId: 'com.opentrons.odd',
-  electronVersion: app.getVersion(),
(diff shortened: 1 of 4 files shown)


app-shell-odd/electron-builder.config.js
```diff
@@ -1,8 +1,9 @@
 'use strict'
+const { app } = require('electron')
 
 module.exports = {
   appId: 'com.opentrons.odd',
-  electronVersion: '23.3.13',
+  electronVersion: app.getVersion(),
```
shlokamin: i dont think this is quite right: this will get the version of the app, not electron: https://www.electronjs.org/docs/latest/api/app#appgetversion

i think we want to import our `package.json` directly: 

```js
import pkg from './package.json'

module.exports = {
...
  electronVersion: JSON.stringify(pkg.devDependencies.electron),
}
```

shlokamin (approved):
lgtm but lets put this in edge instead of the the release branch to minimize variables since they want to release ASAP

also maybe lets wait to merge this into edge until the vite branch is merged?

sfoster1:
Corresponding openembedded branch: https://github.com/Opentrons/oe-core/pull/137

y3rsh (commented):
Along with this would we also want to dynamically define electron version in /app-shell/Makefile ?  Not sure off the top of my head if that is easy to do in a reliable way. 
https://github.com/Opentrons/opentrons/blob/release/app-shell/Makefile#L62

sfoster1:
With this build https://github.com/Opentrons/oe-core/actions/runs/8238878130 of the oe-core PR and with putting all the built in solibs in, the app doesn't haeve a crash and all its processes are running but is still not displaying

sfoster1:
Latest commit alters init behavior. Build with it is here https://github.com/Opentrons/oe-core/actions/runs/8332543203

That openembedded PR will definitely be needed because it also now adds a command line flag that tells chrome to allow all-origin requests for devtools websockets.

sfoster1:
> Along with this would we also want to dynamically define electron version in /app-shell/Makefile ? Not sure off the top of my head if that is easy to do in a reliable way. https://github.com/Opentrons/opentrons/blob/release/app-shell/Makefile#L62

I'm just noting this and uhhhhhhhhhhh what

sfoster1:
> Along with this would we also want to dynamically define electron version in /app-shell/Makefile ? Not sure off the top of my head if that is easy to do in a reliable way. https://github.com/Opentrons/opentrons/blob/release/app-shell/Makefile#L62

OK, had some time to think about this. I think that we should, but maybe in a follow up pr, and honestly I think the most reliable thing to do is use `jq`.

I think we _do not_ want to do this for the ODD. There, we just accept that the version might be different than the desktop.

PR #14602, opened 2024-03-05
fix(api): FLEX fast home collision hangs because of deadlock

ahiuchingau:
RQA-2312

# Overview
When a stall happens during a FLEX fast home move, it tries to acquire the motion lock while trying to refresh position - while it already has the lock and causes the robot to hang.


sfoster1 (approved):
Looks good to me!

PR #14604, opened 2024-03-06
chore: current version numbers in Python API docs

ecormany:
# Overview

Well, we deliberately [changed our setup](https://github.com/Opentrons/opentrons/pull/13991) so we'd err on the side of caution and have to update the `|apiLevel|` substitution value manually.

So naturally, I erred and forgot to bump it to 2.17 before publishing `docs@2.17`. This fixes that and a mention of the latest robot system version.

# Test Plan

[box of sand](http://sandbox.docs.opentrons.com/docs-go-to-2.17-for-real/v2/versioning.html#maximum-supported-versions)

# Changelog

🆙 

# Review requests

While we're at it, do we want to validate the tutorial scripts against 2.17 and bump all those numbers as well? Or happy to leave them on 2.16 for the time being?

# Risk assessment

zippo

### DIFF
--- a/api/docs/v2/conf.py
+++ b/api/docs/v2/conf.py
@@ -99,7 +99,7 @@ extensions += ['sphinx-prompt', 'sphinx_substitution_extensions']
 # use rst_prolog to hold the subsitution
 # update the apiLevel value whenever a new minor version is released
 rst_prolog = f"""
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
Whoops

PR #14605, opened 2024-03-07
feat(shared-data): create util so FE apps can access pip schema v2 defs 

jerader:
closes AUTH-51

# Overview

This PR creates a new util for the FE apps to grab the pipette schema v2 definitions. There are some keys in the v2 definitions that PD will need to have access to (specifically bounding box and nozzle map). 

In a follow up PR, I am going to plug this new util into every instance of `getPipetteNameSpecs` in PD

# Test Plan

Review the util and types and test. 

If you want, find a place where `getPipetteNameSpecs` is being used and plug it in the new util and test it to see that it works as expected! A component i've been testing with is `PipetteTypeTile` and plugging `    console.log('pip specs', getPipetteSpecsV2(o.value))` into line 174 and checking that it logs correctly.

# Changelog

- create new types that match the json shapes
- create a new `getPipetteSpecsV2` util. It takes in the same props as `getPipetteNameSpecs` and `getPipetteModelSpecs` and parses out the matching pipette v2 definition file paths. Then it filters through all the json files in the given paths to return 3+ json files that match the pipetteName's specs. Finally, it combines the json files into 1, including all of the liquid files (default and low volume) and the utility returns that combined object.

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/shared-data/js/__tests__/pipettes.test.ts
+++ b/shared-data/js/__tests__/pipettes.test.ts
@@ -1,7 +1,11 @@
 // tests for pipette info accessors in `shared-data/js/pipettes.js`
 import { describe, expect, it } from 'vitest'
-import { getPipetteSpecsV2, PipetteV2LiquidSpecs, PipetteV2Specs } from '..'
-import { getPipetteNameSpecs, getPipetteModelSpecs } from '../pipettes'
+import {
+  getPipetteSpecsV2,
+  getPipetteNameSpecs,
+  getPipetteModelSpecs,
+} from '../pipettes'
+import type { PipetteV2LiquidSpecs, PipetteV2Specs } from '../types'
 
 const PIPETTE_NAMES = [
   'p10_single',
--- a/shared-data/js/__tests__/pipettes.test.ts
+++ b/shared-data/js/__tests__/pipettes.test.ts
@@ -1,5 +1,6 @@
 // tests for pipette info accessors in `shared-data/js/pipettes.js`
 import { describe, expect, it } from 'vitest'
+import { getPipetteSpecsV2, PipetteV2LiquidSpecs, PipetteV2Specs } from '..'
 import { getPipetteNameSpecs, getPipetteModelSpecs } from '../pipettes'
 
 const PIPETTE_NAMES = [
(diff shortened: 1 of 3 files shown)


shared-data/js/pipettes.ts
```diff
@@ -87,3 +114,93 @@ export const getIncompatiblePipetteNames = (
...
+  const matchingJsons = allJsons.reduce(
+    //  NOTE: combinedModules = array of 3 jsons, module = 1 json
+    (combinedModules: CombinedModules[], module: ModuleSpecs, index) => {
+      const path = Object.keys(combinedGlobs)[index]
+      V2_DEFINITION_TYPES.forEach(type => {
+        if (
+          (type !== 'liquid' &&
+            `../pipette/definitions/2/${type}/${channels}/${pipetteModel}/${version}.json` ===
+              path) ||
+          (type === 'liquid' &&
```
sfoster1: We should export all the "liquid class" liquid definitions, not just default

sfoster1 (changes requested):
Two things -
- If we're going to export the liquid specs, we should definitely export both the default and low volume specs
- Do we have a test or lint job or something that will fail if the ts types don't match the jsons?

shared-data/js/pipettes.ts
jerader: Oh, i didn't realize there was a `lowVolumeDefault` folder, thanks! Yes I should definitely add that in. Since the keys are the same as the `Default`, i wonder if i should add an optional prop to the util to allow for returning either default or low volume.

jerader:
> * Do we have a test or lint job or something that will fail if the ts types don't match the jsons?

@sfoster1, that is a good point, thank you! I will add a test.

shared-data/js/pipettes.ts
sfoster1: Yeah I think that would be fine, or have it be a mapping. In theory we will add more kinds of liquid classes.

shared-data/js/pipettes.ts
```diff
@@ -87,3 +114,95 @@ export const getIncompatiblePipetteNames = (
...
+
+const V2_DEFINITION_TYPES = ['general', 'geometry', 'liquid']
+
+/* takes in pipetteName such as 'p300_single' or 'p300_single_gen1' 
+or PipetteModel such as 'p300_single_v1.3' and converts it to channels,
+model, and version from generation in order to return the correct pipette schema v2 json files. 
+**/
+export const getPipetteSpecsV2 = (
+  name: PipetteName | PipetteModel,
+  lowVolume: boolean
```
sfoster1: Could we do something like `liquidClass?: 'default' | 'lowVolume'` with unspecified meaning default or something? or even `liquidClass?: string`

sfoster1 (changes requested):

shared-data/js/pipettes.ts
jerader: oh mapping makes a lot of sense actually and could accommodate more liquid classes as we make them without having to refactor. Okay i'm gonna do that instead. thanks!

jerader:
- [x] write test after vite migration merges into edge

shared-data/js/pipettes.ts
```diff
@@ -87,3 +111,124 @@ export const getIncompatiblePipetteNames = (
...
+      return '3_0'
+    }
+    default: {
+      return null
+    }
+  }
+}
+
+const V2_DEFINITION_TYPES = ['general', 'geometry']
+const LIDUID_TYPES = ['default', 'lowVolumeDefault']
```
sfoster1: can we discover these dynamically from what directories are present or do we have to list them here?

sfoster1 (approved):
Looks great to me! Thank you!

shared-data/js/pipettes.ts
```diff
@@ -89,3 +116,133 @@ export const getIncompatiblePipetteNames = (
...
+          path
+        ) {
+          combinedModules.push(module.default)
+        }
+      })
+      return combinedModules
+    },
+    []
+  )
+  const generalGeometricCombined = Object.assign({}, ...matchingJsons)
```
b-cooper: This should probably be wrapped into the reduce for clarity?

shared-data/js/pipettes.ts
```diff
@@ -89,3 +116,133 @@ export const getIncompatiblePipetteNames = (
...
+const V2_DEFINITION_TYPES = ['general', 'geometry']
+
+/* takes in pipetteName such as 'p300_single' or 'p300_single_gen1' 
+or PipetteModel such as 'p300_single_v1.3' and converts it to channels,
+model, and version in order to return the correct pipette schema v2 json files. 
+**/
+export const getPipetteSpecsV2 = (
+  name: PipetteName | PipetteModel
+): PipetteV2Specs | null => {
+  const nameSplit = name.split('_')
```
koji, line 168: nit

```ts
const {[0]: pipetteModel, [1]: pipetteChannel, [2]: pipetteGen } = name.split('_')
```

shared-data/js/pipettes.ts
```diff
@@ -89,3 +116,133 @@ export const getIncompatiblePipetteNames = (
...
+  const liquidTypes: string[] = []
+
+  const liquidJsons = Object.keys(liquid).map(
+    path => liquid[path]
+  ) as LiquidSpecs[]
+
+  //  liquidMatchingJsons returns an array of all liquid definition jsons
+  const liquidMatchingJsons = liquidJsons.reduce(
+    (combinedModules: PipetteV2LiquidSpecs[], module: LiquidSpecs, index) => {
+      const path = Object.keys(liquid)[index]
```
b-cooper: You should be able to remove some redundant code if you reduce `Object.entries` instead of creating `liquidJsons`?

shared-data/js/__tests__/pipettes.test.ts
```diff
@@ -56,4 +61,180 @@ describe('pipette data accessors', () => {
...
+        displayCategory: 'FLEX',
+        displayName: 'Flex 1-Channel 1000 μL',
+        dropTipConfigurations: { plungerEject: { current: 1, speed: 10 } },
+        liquids: {
+          default: {
+            $otSharedSchema:
+              '#/pipette/schemas/2/pipetteLiquidPropertiesSchema.json',
+            defaultTipOverlapDictionary: {
+              default: 10.5,
+              'opentrons/opentrons_flex_96_tiprack_1000ul/1': 10.5,
```
koji, line 87: nit
maybe define as const?

shared-data/js/types.ts
```diff
@@ -394,6 +394,119 @@ export interface FlowRateSpec {
...
+}
+
+export interface PipetteV2GeometrySpecs {
+  nozzleOffset: number[]
+  pipetteBoundingBoxOffsets: {
+    backLeftCorner: number[]
+    frontRightCorner: number[]
+  }
+  pathTo3D: string
+  orderedRows: any
```
koji: Can we switch the type from any to a specific type?
I guess Brent upgraded the TypeScript version so now defining a type would be more flexible than before.

shared-data/js/types.ts
```diff
@@ -394,6 +394,119 @@ export interface FlowRateSpec {
...
+    dispense: {
+      default: {
+        1: TipData
+      }
+    }
+  }
+}
+
+export interface PipetteV2LiquidSpecs {
+  $otSharedSchema: string
```
koji, line 499: question

`$otSharedSchema`.  Do you know why we use `$` in an object item name?

shared-data/js/pipettes.ts
jerader: some of the names don't have a pipetteGen (like `p10_single`)  so i'm not sure if this makes it more clear or not

shared-data/js/types.ts
jerader: it represents the schema, i thought about removing it because i don't think we care about them in the frontend

koji (approved):
LGTM!

shared-data/js/pipettes.ts
```diff
@@ -89,3 +118,109 @@ export const getIncompatiblePipetteNames = (
...
+
+  let version: string
+  if (nameSplit.length === 2) {
+    version = '1_0'
+  } else if (gen != null) {
+    version = gen //  ex: gen1 -> 1_0
+  } else {
+    const versionDot = nameSplit[2].split('v')[1]
+    version = versionDot.replace('.', '_') // ex: v1.1 -> 1_1
+  }
```
sanni-t: Maybe it's my lack of js code comprehension but does this handle the 96-channel correctly? We load it using the `p1000_96` name in engine and I'm guessing that's what is used in the client side. Is that case covered here?

jerader: Yup! The 96 channel will fall under the `nameSplit.length === 2` case since `p1000_96` will be split up into an array length of 2.

sanni-t: But it looks like it'll always default to using the `1_0` version?

jerader: Oh, should it be defaulting to the `3_0` version?

b-cooper: It depends on this function's purpose. Right now the function takes in either a `model` or a `name`.  If a name is given you're going to end up giving the user version specific information, do the front-ends actually want the most modern version specific information?  If so, then the logic here should specifically find the most modern model of a given pipette name. 

It might be helpful to organize this by splitting the internals of this function on how it handles `pipetteName` vs. `pipetteModel`?

jerader: the specific version will be taken care of in the `else`.

If you have the `pipetteName`, it is either something like `p10_single` or `p10_single_gen1`. If it has the gen attached, then the version will be `1_0` for gen1, `2_0` for gen2, `3_0` for flex. If it doesn't have the gen, then i've been defaulting to `1_0` (since not having the gen means it is a gen1 pipette, except for the 96-channel). But seems like the 96-channel, even though it is a gen1, needs to use `3_0`.

sanni-t: > do the front-ends actually want the most modern version specific information?

Sorry, this is what I'm trying to understand too. The 96 channel does have a `1_0` definition as default. But ya for non-default we'll have to use the `3_..` versions. 
Checking what we do in python.

Where does the app get pipette model/ version from? Is that from `/instruments`? 
And I guess the question of default values is more relevant to PD since PD won't know exact versions of pipettes to use?

jerader: > And I guess the question of default values is more relevant to PD since PD won't know exact versions of pipettes to use?

Correct, PD will always get the specs from the pipetteName.

> Checking what we do in python.

looks like in python in` version_from_generation`, we are defaulting to `3_0`, `2_0`, or `1_0`. - looks like 96-channel is using `1_0`. I definitely jumped the gun and added branching logic to change it to `3_0`, i will change it back.

sanni-t: Got it. Sorry about the confusion.

As for the question of whether we should use the latest version of a pipette as default, this was brought up previously but I'm not sure if there was a decision made. I'll check and post an update somewhere relevant later.

sanni-t (approved):
LGTM! This is really awesome

shared-data/js/pipettes.ts
jerader: no worries, and thank you!

PR #14606, opened 2024-03-07
fix(api): simulate not logging drop_tip with no args

DerekMaggio:
# Overview

Fixes [RESC-214](https://opentrons.atlassian.net/browse/RESC-214)

# Test Plan

- [x] Add simulate test that runs a protocol using `drop_tip` with no args

# Changelog

- Add publisher.publish_context context manager when calling drop tip with no args
- Add above test case to `test_simulate.py`

# Review requests

- It seems that the simulate drop tip functionality could benefit from more robust test coverage. Should make sure that we validate all branches inside of drop_tip. But is this PR the place to do it? 

# Risk assessment

Very low. Just added a message and a test

[RESC-214]: https://opentrons.atlassian.net/browse/RESC-214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1024,11 +1024,17 @@ class InstrumentContext(publisher.CommandPublisher):
             if isinstance(trash_container, labware.Labware):
                 well = trash_container.wells()[0]
             else:  # implicit drop tip in disposal location, not well
-                self._core.drop_tip_in_disposal_location(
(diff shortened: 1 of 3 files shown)


ecormany:
It works 🎉

sfoster1 (approved):
Looks like it'll work!

PR #14607, opened 2024-03-07
Uploads ABR Runs to Google Sheet

rclarke0:
# Overview

Uploads ABR runs to google sheet

# Test Plan

1. Uploaded to empty google sheet
2. Uploaded additional lines to filled google sheet
3. Uploaded to sheet during temp/humidity sensor data collection and verified that the other google sheet was not affected.

# Changelog

- Added time conversion to abr_run_logs.py to convert from UTC to EST
- Added check in error logging if error was not documented in final command
- removed addition of instrument model from serial number since model information can be extracted from serial.
- Added lines to connect to google sheet
- Added lines to write to google sheet
- Changed try exception block when reading json files to an if statement:read if ends with .json and if run_id exists to avoid reading google_sheets_tools.py and credentials file
- In liquid measurement analysis script (analyze_abr.py) changed from reading whole entire file to just the last line because the file stops logging when scale is stable. The last line will always be the stable line.

# Review requests

# Risk assessment

Users need a credentials file and google_sheets_tools.py saved in the folder given as an argument when running ABR_read_logs

### DIFF
--- a/hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
+++ b/hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
@@ -1,4 +1,4 @@
-"""Read ABR run logs and save data to ABR testing csv."""
+"""Read ABR run logs and save data to ABR testing csv and google sheet."""
 from .abr_run_logs import get_run_ids_from_storage, get_unseen_run_ids
 from .error_levels import ERROR_LEVELS_PATH
 from typing import Set, Dict, Tuple, Any, List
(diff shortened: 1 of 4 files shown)


hardware-testing/hardware_testing/abr_tools/abr_read_logs.py
```diff
@@ -97,13 +109,9 @@ def create_data_dictionary(
     runs_and_robots = {}
     for filename in os.listdir(storage_directory):
         file_path = os.path.join(storage_directory, filename)
-        try:
+        if file_path.endswith(".json"):
```
sfoster1, line 112: should we have an
```
else:
     continue
```
here?

sfoster1 (approved):
Looks good, though I think we might want to ignore files that aren't in json

sfoster1 (approved):

PR #14608, opened 2024-03-07
fix(robot-server): dont let /instruments block

sfoster1:
It calls cache_instruments and that can block because it takes the motion lock, but we really don't want that. We don't mind if cache_instruments doesn't get called on the flex because it's sort of a secondary functionality, so just bail early in this case.

## Testing
- Run on a robot and do some `GET /instruments` while things are homign and note that it instantly returns
- Do an attach flow and note that it still returns the new instruments.

Closes EXEC-298

### DIFF
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -431,7 +431,9 @@ class API(
         return False
 
     async def cache_instruments(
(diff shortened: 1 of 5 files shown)


SyntaxColoring (approved):
Interesting!

ahiuchingau (approved):
Noice

PR #14609, opened 2024-03-07
fix(api): raise error if fast home is stalling

ahiuchingau:
# Overview
Closes RQA-2312

We previously swallowed collision errors during fast home move and proceeded to slow home because we used to not handle encoder overflow properly and would mistakenly raise. Now that we have fixed those encoder issues, we should actually raise the error when the motor stalls.


sfoster1 (approved):
Ooh, nice!

PR #14611, opened 2024-03-08
docs(api): Python Protocol API 2.18

ecormany:
# Overview

This is the feature branch for Python API version 2.18 docs.

_Please base all PRs with 2.18-related docs here, rather than on `edge`, if possible._ (But get code into whatever branch makes sense.)


jbleon95 (approved):
Did a quick scan through, everything looks accurate!

PR #14612, opened 2024-03-08
chore(monorepo): remove jest-when from repo

koji:
# Overview
vite migration branch was merged so we don't need jest-when any more. Removing jest will be done in a following pr since it will require some eslint modification.

# Test Plan
the following commands work as expect
- make check-js
- make lint-js
- make lint-css
- make test-js

# Changelog

# Review requests

# Risk assessment
low


mjhuff (approved):
Looks good, thank you!

PR #14613, opened 2024-03-08
feat(api): remove z safety margin from pipette movement check

sanni-t:
Closes RESC-216

# Overview

See the escalations ticket for more info.
Now that we use actual nozzle positions in movement conflict check, we shouldn't need to add such a big safety margin (10mm). We will prefer allowing the movement if it's in the realm of 'slight chance of collision'.

# Review requests

Added an integration test. Feel free to throw some tests at it that check for collisions when moving to places where a pipette goes very close to a surrounding labware but falls just short of actually touching it.

**Question: There is another way we can handle this and it's keeping the safety margin but making it low, like 2mm. Should we do that instead? See risk assessment for reason.**

# Risk assessment

Medium- low: there is a slight chance- due to equipment manufacturing tolerance that the pipette might touch/collide a deck item.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
+++ b/api/src/opentrons/protocol_api/core/engine/deck_conflict.py
@@ -79,9 +79,6 @@ A1_column_back_right_bound = Point(
     x=A12_column_back_right_bound.x - _NOZZLE_PITCH * 11, y=506.2
 )
 
-# Arbitrary safety margin in z-direction
-Z_SAFETY_MARGIN = 10
-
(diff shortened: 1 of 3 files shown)


SyntaxColoring (approved):
Makes sense, thanks.

I agree with removing the safety margin entirely until we can report these close shaves as nonblocking warnings.

sanni-t:
> Makes sense, thanks.
> 
> I agree with removing the safety margin entirely until we can report these close shaves as nonblocking warnings.

That's a good point about warnings!

PR #14614, opened 2024-03-08
chore(api): define command note in engine

sfoster1:
The CommandNote is a piece of data that may be attached to a Command instance. It will also be exported in CommandSummaries.

Notes are data that are attached to commands by the code that executes them or the code that dispatches them. They aren't created by authorship software.

Commands are intended for consumption as part of a run record - the long-lasting record of what actions the robot took. For instance, the desktop app or the ODD might consume them to display information about commands that have been executed.

Notes have very little structure to them, which is on purpose for robustness of serialization especially across versions. Most fields are strings, and the model bakes in no references to other pieces of data. Instead, notes might be attached to other things - Commands might get an array of notes, for instance. This means that a single note will relate to exactly one other thing - one command might have 0 or more notes, but one Note will only ever refer to one Command.

Closes EXEC-287

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/command.py
+++ b/api/src/opentrons/protocol_engine/commands/command.py
@@ -6,7 +6,16 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from datetime import datetime
 from enum import Enum
-from typing import TYPE_CHECKING, Generic, Optional, TypeVar, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Generic,
+    Optional,
+    TypeVar,
+    Tuple,
+    Union,
+    Literal,
+    List,
+)
 
 from pydantic import BaseModel, Field
 from pydantic.generics import GenericModel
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -27,6 +36,29 @@
 
 CommandPrivateResultT = TypeVar("CommandPrivateResultT")
 
+NoteKind = Union[Literal["warning", "information"], str]
```
DerekMaggio, line 38: Am I correct to assume the `str` will be removed down the road when we start getting more solid definitions?

DerekMaggio (approved):

api/src/opentrons/protocol_engine/commands/command.py
sfoster1: Nope! That will stay part of this definition forever. It's meant to serve as a sentinel for future refinements but will always maintain its presence as a backstop for even more future assignments

SyntaxColoring (commented):
Oops, sorry I missed this—LGTM.

PR #14616, opened 2024-03-08
feat(api,robot-server): export notes on commands

sfoster1:
Commands in runs can have notes attached to them. Each command can have zero or more notes, and zero notes can be expressed as any of (notes=None, notes is-not-present, notes is empty-list). This is to make the public models a little more robust across versions.

Closes EXEC-288

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/command.py
+++ b/api/src/opentrons/protocol_engine/commands/command.py
@@ -14,6 +14,7 @@ from typing import (
     Tuple,
     Union,
     Literal,
+    List,
 )
 
 from pydantic import BaseModel, Field
--- a/api/src/opentrons/protocol_engine/commands/command.py
+++ b/api/src/opentrons/protocol_engine/commands/command.py
@@ -175,6 +175,13 @@ class BaseCommand(GenericModel, Generic[CommandParamsT, CommandResultT]):
             " a command that is part of a calibration procedure."
         ),
     )
+    notes: Optional[List[CommandNote]] = Field(
+        None,
+        description=(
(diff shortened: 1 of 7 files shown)


api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -176,6 +176,13 @@ class BaseCommand(GenericModel, Generic[CommandParamsT, CommandResultT]):
             " a command that is part of a calibration procedure."
         ),
     )
+    notes: Optional[List[CommandNote]] = Field(
```
DerekMaggio, line 179: Will the ordering of the notes end up being important?

DerekMaggio (dismissed):
Looks straightforward to me.

DerekMaggio (approved):
Did some digging to ensure that the requirements of the ticket are being met. Commands are sent via HTTP, with a note attached, the note is validated by querying for said command, and then checking its value. 

Approval, on the condition that all tests and stuff end up passing

api/src/opentrons/protocol_engine/commands/command.py
sfoster1: I think probably not

TamarZanzouri (approved):
Looks great! thank you! do we not need to update the json commands schema?

SyntaxColoring (approved):
No notes (ha ha ha ha ha)

sfoster1:
> Looks great! thank you! do we not need to update the json commands schema?

We do not, because the commands schema actually contains the `CommandCreate` union. Notes do not appear in protocols, only runs, so they don't need to be there.

PR #14619, opened 2024-03-11
fix(app, shared-data): fix null check issue

koji:
# Overview
fix a couple of null check issues

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/__mocks__/electron-store.js
+++ b/__mocks__/electron-store.js
@@ -1,7 +1,7 @@
 // mock electron-store
 'use strict'
 import { vi } from 'vitest'
-import { DEFAULTS_V12, migrate } from '../app-shell-odd/src/config/migrate'
+// import { DEFAULTS_V12, migrate } from '../app-shell-odd/src/config/migrate'
 
 // will by default mock the config dir. if you need other behaavior you can
 // override this mock (see app-shell/src/__tests__/discovery.test.ts for an example)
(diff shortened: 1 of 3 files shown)


__mocks__/electron-store.js
```diff
@@ -1,7 +1,7 @@
 // mock electron-store
 'use strict'
 import { vi } from 'vitest'
-import { DEFAULTS_V12, migrate } from '../app-shell-odd/src/config/migrate'
+// import { DEFAULTS_V12, migrate } from '../app-shell-odd/src/config/migrate'
```
mjhuff: Are we sure we want to comment out lines like this instead of just removing them completely? What's the actual reason we are commenting this out? 

I think what we really want to do is update the import location?

app/src/resources/runs/useNotifyRunQuery.ts
```diff
@@ -24,7 +24,7 @@ export function useNotifyRunQuery<TError = Error>(
   useNotifyService({
     topic: `robot-server/runs/${runId}` as NotifyTopic,
     setRefetchUsingHTTP,
-    options: { ...options, enabled: options.enabled && runId != null },
+    options: { ...options, enabled: options?.enabled != null && runId != null },
```
mjhuff: On L17, `options` will always be defined. 
```suggestion
    options: { ...options, enabled: options.enabled != null && runId != null },
```

mjhuff (approved):
Looks good, thank you!

PR #14620, opened 2024-03-11
refactor(robot-server): Utilize unsubscribe flags for dynamic topics

mjhuff:
Closes [EXEC-305](https://opentrons.atlassian.net/browse/EXEC-305)

# Overview

This PR adds unsubscribe flags for dynamic notification topics, e.g. "/runs/:runId". 

Now that the dust has settled around what our push notification can (and can’t) do, it’s clear that the burden of unsubscribing should be on the robot-server, specifically for dynamic topics. 

In other words, it’s ok for the app to connect to a broker and remain subscribed to topics that will never change, ex. “/maintenance_runs/current_run” (and be subscribed to those topics as long as the app is connected). Dynamic topics, such as “/runs/:runId”, should tell the app to unsubscribe however, as requiring the broker to stores these topics unnecessarily increases memory load on the robot server during long-lived sessions. 

This model is much easier to reason about as opposed to the component interest model that we use now. Bug surface is reduced, too.

# Test Plan
- Start up postman and start a protocol run using the desktop app's dev mode. You'll notice in the app-shell logs a lot of subscribe requests to a /run/:runId. Make note of that runId and in postman, subscribe to /runs/<that run Id>.
- After the run concludes, start another protocol run. Verify that the previous run (the run you're subscribed to in Postman), receives a message containing `unsubscribe: true`. 

# Risk assessment
low

[EXEC-305]: https://opentrons.atlassian.net/browse/EXEC-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/service/json_api/response.py
+++ b/robot-server/robot_server/service/json_api/response.py
@@ -293,7 +293,7 @@ class NotifyRefetchBody(BaseResponseBody):
 class NotifyUnsubscribeBody(BaseResponseBody):
     """A notification response.
 
-    Returns flags for refetching via HTTP and unsubscribing from a topic.
+    Returns flags for unsubscribing from a topic.
     """
 
-    refetchAndUnsubscribe: bool = True
+    unsubscribe: bool = True
--- a/robot-server/robot_server/service/json_api/response.py
+++ b/robot-server/robot_server/service/json_api/response.py
@@ -288,3 +288,13 @@ class NotifyRefetchBody(BaseResponseBody):
     """A notification response that returns a flag for refetching via HTTP."""
 
     refetchUsingHTTP: bool = True
+
+
+class NotifyUnsubscribeBody(BaseResponseBody):
+    """A notification response.
+
+    Returns flags for refetching via HTTP and unsubscribing from a topic.
+    """
+
+    refetchUsingHTTP: bool = True
+    unsubscribe: bool = True
--- a/robot-server/robot_server/service/json_api/response.py
+++ b/robot-server/robot_server/service/json_api/response.py
@@ -296,5 +296,4 @@ class NotifyUnsubscribeBody(BaseResponseBody):
     Returns flags for refetching via HTTP and unsubscribing from a topic.
     """
 
-    refetchUsingHTTP: bool = True
-    unsubscribe: bool = True
+    refetchAndUnsubscribe: bool = True
--- a/robot-server/robot_server/service/json_api/response.py
+++ b/robot-server/robot_server/service/json_api/response.py
@@ -285,5 +285,6 @@ class ResponseList(BaseModel, Generic[ResponseDataT]):
 
 
 class NotifyRefetchBody(BaseResponseBody):
-    "A notification response that returns a flag for refetching via HTTP."
+    """A notification response that returns a flag for refetching via HTTP."""
+
     refetchUsingHTTP: bool = True
--- a/robot-server/robot_server/service/notifications/notification_client.py
+++ b/robot-server/robot_server/service/notifications/notification_client.py
@@ -78,7 +78,7 @@ class NotificationClient:
         await to_thread.run_sync(self.client.disconnect)
 
     async def publish_async(
-        self, topic: str, message: NotifyRefetchBody = NotifyRefetchBody()
+        self, topic: str, message: NotifyRefetchBody = NotifyRefetchBody.construct()
     ) -> None:
         """Asynchronously Publish a message on a specific topic to the MQTT broker.
 
@@ -89,7 +89,7 @@ class NotificationClient:
         await to_thread.run_sync(self.publish, topic, message)
 
     def publish(
-        self, topic: str, message: NotifyRefetchBody = NotifyRefetchBody()
+        self, topic: str, message: NotifyRefetchBody = NotifyRefetchBody.construct()
     ) -> None:
         """Publish a message on a specific topic to the MQTT broker.
 
--- a/robot-server/robot_server/service/notifications/notification_client.py
+++ b/robot-server/robot_server/service/notifications/notification_client.py
@@ -3,10 +3,10 @@ import logging
 import paho.mqtt.client as mqtt
 from anyio import to_thread
 from fastapi import Depends
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
 from enum import Enum
 
-from ..json_api import NotifyRefetchBody
+from ..json_api import NotifyRefetchBody, NotifyUnsubscribeBody
 from server_utils.fastapi_utils.app_state import (
     AppState,
     AppStateAccessor,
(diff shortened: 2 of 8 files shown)


robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -56,24 +57,26 @@ async def begin_polling_engine_store(
             )
 
     async def stop_polling_engine_store(self) -> None:
-        """Stops polling the engine store."""
+        """Stops polling the engine store. Run-related topics will publish as the poller is cancelled."""
         if self._poller is not None:
             self._run_data_manager_polling.set()
```
DerekMaggio, line 68: Doing some research, there are other ways that a CancelledError can be triggered. The ones we might encounter would be: timeouts or when an asyncio task is awaiting another task that gets canceled.

To handle this, should all of the `= None` statements inside `stop_polling_engine_store()` be moved to inside the `except asyncio.CancelledError`? This way we ensure, regardless of what raises the `CancelledError`, we are fully cleaning up the RunPublisher?

DerekMaggio (commented):
Unsubscribing from the run after it is finished makes sense to me, as that run shouldn't be receiving any updates, so it makes no sense to keep listening for events.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
mjhuff: This makes a whole lot of sense. Good catch!

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -87,36 +86,55 @@ async def _poll_engine_store(
...
+                    self._previous_current_command = current_command
+
+                if self._previous_state_summary_status != current_state_summary_status:
+                    await self._publish_runs_async(run_id=run_id)
+                    self._previous_state_summary_status = current_state_summary_status
+                await asyncio.sleep(1)
+        except asyncio.CancelledError:
+            self._clean_up_poller()
+            await self._publish_runs_async(run_id=run_id, should_unsubscribe=True)
+            await self._client.publish_async(topic=Topics.RUNS_CURRENT_COMMAND)
```
DerekMaggio: More on error handling.
Do we also need a catch-all `except` to handle any other errors thrown in the while loop? 
Something like, 

```python
except Exception as e:
    log.error(e)
    await self.stop_polling_engine_store()

    # Not sure if we should restart or not.
    # Seems like it could be prone to getting stuck in an infinite loop 
    # of stopping and restarting if it keeps hitting the 
    # same error over and over
    await self.begin_polling_engine_store() 
```

mjhuff: I think I'll add an Exception and just log the error. That's not a bad idea. I'm not too keen on touching this logic extensively because it's temporary (and it hasn't given us issues for the past couple of months). This polling logic won't exist in a couple weeks. It's only a few ticket items down the list for me to tackle. If I can avoid breaking anything or sinking time covering weird edge-cases, I'm cool with that if you are.

Here's [the epic if you're interested](https://opentrons.atlassian.net/browse/EXEC-157). I just added the [PE refactor ticket to it](https://opentrons.atlassian.net/browse/EXEC-310).

DerekMaggio: Oh, perfect.
Just logging sounds perfect then. I saw the Epic but wasn't sure when it was being addressed.

robot-server/robot_server/service/json_api/response.py
```diff
@@ -285,5 +285,16 @@ class ResponseList(BaseModel, Generic[ResponseDataT]):
...
+
+    refetchUsingHTTP: bool = True
+
+
+class NotifyUnsubscribeBody(BaseResponseBody):
+    """A notification response.
+
+    Returns flags for refetching via HTTP and unsubscribing from a topic.
+    """
+
```
sfoster1, line 298: I do not think `NotifyUnsubscribeBody` should inherit from `NotifyRefetchBody` like this. I think these messages have different semantic meanings and having them both be in the same message kind of encourages the sort of functions I object to later in the review. _In practice_, I think that clients maybe should use the reception of a `NotifyUnsubscribe` as a cue to refetch as well as cut the connection, but I think it's a separate meaning.

DerekMaggio (approved):

robot-server/robot_server/service/notifications/notification_client.py
```diff
@@ -77,26 +77,28 @@ async def disconnect(self) -> None:
         self.client.loop_stop()
         await to_thread.run_sync(self.client.disconnect)
 
-    async def publish_async(
-        self, topic: str, message: NotifyRefetchBody = NotifyRefetchBody()
-    ) -> None:
+    async def publish_async(self, topic: str, should_unsubscribe: bool = False) -> None:
```
sfoster1: This function, and its synchronous cousin, are functions that do two completely different things - semantically - based on a boolean parameter. 

With the parameter `False` or unspecified, this causes a refetch advice notification; with the parameter `True`, it also causes the client to hang up forever. Those are two very different things. It's also true of the message definitions (see above).

This is something we try and avoid as much as we possibly can. I think we should avoid it here.

For now, I'd advice having a `publish_advise_unsubscribe()` method or something; maybe even change the general `publish()` to `publish_advise_refetch()`. As (if) we get more messages that can send via the notifications channel, we may want to refactor this into something that takes a message object, but hey we may end up refactoring that back into a function-like interface too eventually.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -56,24 +61,22 @@ async def begin_polling_engine_store(
...
             self._run_data_manager_polling.set()
             self._poller.cancel()
-            self._poller = None
-            self._run_data_manager_polling.clear()
-            self._previous_current_command = None
-            self._previous_state_summary_status = None
-            await self._client.publish_async(topic=Topics.RUNS_CURRENT_COMMAND)
 
-    def publish_runs(self, run_id: str) -> None:
+    def publish_runs(self, run_id: str, should_unsubscribe: bool = False) -> None:
```
sfoster1: I think this is the same problem as the other message - does something different, for one of its subtopics, depending on a flag like this. Let's factor it into two methods.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -87,36 +90,57 @@ async def _poll_engine_store(
...
-            current_state_summary = get_state_summary(run_id)
-            current_state_summary_status = (
-                current_state_summary.status if current_state_summary else None
-            )
-
-            if self._previous_current_command != current_command:
-                await self._publish_current_command()
-                self._previous_current_command = current_command
+        try:
+            while not self._run_data_manager_polling.is_set():
```
sfoster1: more of a nit but if we're going to do a top-level catchall handler, I think we should make another function we call out to inside the `try/except`

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -87,36 +90,57 @@ async def _poll_engine_store(
...
-                self._previous_state_summary_status = current_state_summary_status
-            await asyncio.sleep(1)
+                if self._previous_current_command != current_command:
+                    await self._publish_current_command()
+                    self._previous_current_command = current_command
+
+                if self._previous_state_summary_status != current_state_summary_status:
+                    await self._publish_runs_async(run_id=run_id)
+                    self._previous_state_summary_status = current_state_summary_status
+                await asyncio.sleep(1)
```
sfoster1: we should make this a global constant tunable while we're here

sfoster1 (changes requested):
I think we should refactor this so we don't have a function with a boolean flag that drastically alters its semantics.

sfoster1 (approved):
Thanks for making the changes, LGTM!

mjhuff:
Will merge alongside https://github.com/Opentrons/opentrons/pull/14640

PR #14621, opened 2024-03-11
fix(app): properly manage the ipcRenderer notify event emitter

mjhuff:
Closes [RQA-2459](https://opentrons.atlassian.net/browse/RQA-2459)

# Overview

Do not instantiate new notification ipcRenderer event listeners for each component that intends to subscribe to the notification server. Instead re-use the same ipcRenderer listener. This not only reduces the number of emitters created, but fixes a bug in which ipcRenderer listeners were not properly disposed.

# Test Plan
- Verified via Chrome console that there is no memory leak warning, which often appears after navigating to the devices page or the robot details page.
- Verified via smoke test that notifications still work as expected.
- Using [this build of the app](https://opentrons.slack.com/archives/C81CR4VCZ/p1710161213430769), update a robot on the devices tab via the robot update banner, click a lot, let the app sit on the device details page for a while, etc. etc., and verify the app doesn't whitescreen.

# Changelog
- Fixed a memory leak.

# Review requests
- Read the ticket if you'd like more context. I've asked QA to try and repro this as well. Regardless of whether or not this actually solves the whitescreen (although I think this does), this should make it into 7.2.1.

# Risk assessment
medium-low. The code touches the app's ipcRenderer for notifications, but it's pretty straightforward: if notifications still work as expected, we're good. 

[RQA-2459]: https://opentrons.atlassian.net/browse/RQA-2459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/redux/shell/remote.ts
+++ b/app/src/redux/shell/remote.ts
@@ -37,20 +37,50 @@ export function appShellRequestor<Data>(
   return remote.ipcRenderer.invoke('usb:request', configProxy)
 }
 
-export function appShellListener(
-  hostname: string | null,
-  topic: NotifyTopic,
+interface CallbackStore {
+  [hostname: string]: {
+    [topic in NotifyTopic]: Array<(data: NotifyResponseData) => void>
+  }
+}
+const callbackStore: CallbackStore = {}
+
+interface AppShellListener {
+  hostname: string
+  topic: NotifyTopic
   callback: (data: NotifyResponseData) => void
-): void {
-  remote.ipcRenderer.on(
-    'notify',
-    (_, shellHostname, shellTopic, shellMessage) => {
-      if (
-        hostname === shellHostname &&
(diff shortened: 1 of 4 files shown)


mjhuff:
We explicitly prevent testing of `remote.ts`, since it breaks all our Jest tests, meaning writing a unit test for `appShellListener` requires moving it out `remote.ts` and doing some import/export changes. This seems like a big workaround for Jest considering we no longer use it in edge, and we don't seem to have the same limitation of mocking the whole file path. My plan is to do the 7.2.1 mergeback into edge with the unit test, but I'm open to suggestions if we think it's worth changing the plan.

sfoster1 (approved):
Looks good to me, nice fix!

PR #7136, opened 2020-12-16
refactor(api, robot-server): change tip length data model to include uri

Laura-Danielle:
# Overview

Add the uri for a tiprack in the tip length calibration data model and in the endpoint that returns tip length calibration. Since there is no way to reverse a hash into a uri, we can only update the uri key once a tip length has been saved again.

# Changelog

- Modify the data model in calibration storage
- Modify the data model in robot server that sends tip length info over
- Fix tests related to this.

# Review requests
Push to a robot, make sure nothing breaks. I tested both tip length calibration and uploading a protocol to make sure the tip length still shows up in the calibration panel.

# Risk assessment

Low because we are only adding the key in when saving next, and otherwise, loading an empty string.

PR #14622, opened 2024-03-11
fix(app): Tolerate old tip length calibration records without a `uri` field

SyntaxColoring:
# Overview

This fixes RESC-215.

Apparently, old tip length calibration records don't have a `uri` field. (It looks like `uri` was introduced in #7136, v4.1.0.) We used to safely drop these records via [this `except ValidationError` block](https://github.com/Opentrons/opentrons/blob/a1cde96c41b8cbd065fe17057e3c6119b6b252f0/api/src/opentrons/calibration_storage/ot2/tip_length.py#L46-L50). But PR #14512 started [accessing the `uri` field manually](https://github.com/Opentrons/opentrons/blob/1556c1c442d4de3619a38a56b07d20b42b60c0fc/api/src/opentrons/calibration_storage/ot2/tip_length.py#L50), accidentally bringing it outside the coverage of the `except` block and turning it into an HTTP 500 error.

# Test Plan

To reproduce this manually, you can download the `data/` directory from RESC-215 and run this in the `api` project directory:

```shell
OT_API_TIP_LENGTH_CALIBRATION_DIR=/path/to/downloaded/data/tip_lengths \
  pipenv run python -c \
  'from opentrons import calibration_storage; print(calibration_storage.get_all_tip_length_calibrations())'
```

On v7.1.1, it will work. On v7.2.0, it will raise an error. On this branch, it will work again.

I've also added a unit test.

# Changelog

* If `uri` is missing from an old tip length calibration record, silently ignore that record.
* Fix a misplaced `JSONDecodeError` catch. (This turned out to not be the problem, but it's still probably a good idea to fix.)
* Add exception info to a bunch of parsing warnings to make these problems easier to debug. (These warnings turned out to not cover this problem, but it's still probably a good idea to fix.)

# Review requests

* Is there a better way to handle these records than skipping over them entirely?

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/calibration_storage/ot2/deck_attitude.py
+++ b/api/src/opentrons/calibration_storage/ot2/deck_attitude.py
@@ -79,7 +79,7 @@ def get_robot_deck_attitude() -> Optional[v1.DeckCalibrationModel]:
         pass
     except (json.JSONDecodeError, ValidationError):
         log.warning(
-            "Deck calibration is malformed. Please factory reset your calibrations."
+            "Deck calibration is malformed. Please factory reset your calibrations.",
+            exc_info=True,
         )
-        pass
     return None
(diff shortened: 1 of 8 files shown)


Laura-Danielle (approved):
Nice debugging and fixing max :) I also don't know when we added in URI because I thought we've had that component for a long time. Maybe I only added it in when doing calibration stuff for the flex, but it's been a while so I can't remember.

PR #14626, opened 2024-03-11
fix(protocol-designer): fix custom labware uploads afrer vite migration

shlokamin:
# Overview
Oopsy I definitely used the wrong JSON schema when doing the vite migration. Custom labware uploads were using the protocol schema instead of the labware schema 😨


jerader (approved):
lol thanks for the fix!!

PR #14627, opened 2024-03-11
refactor(protocol-designer): export button disabled before file created

jerader:
# Overview

Fix bug from when i refactored the `FileSidebar` component to not be a class component

# Test Plan

make sure the export button is disabled when you first open PD but then after you create a file, it is enabled

# Changelog

fix the oopsy. initially what i had, it made the export button always enabled.

# Review requests

see test plan

# Risk assessment
low


koji (approved):
followed the test case and the sandbox worked as expected.

PR #14633, opened 2024-03-12
refactor(app-shell, app-shell-odd): Refactor app to use unsubscribe flags

mjhuff:
Closes [EXEC-306](https://opentrons.atlassian.net/browse/EXEC-306)

# Overview

This is the frontend implementation of https://github.com/Opentrons/opentrons/pull/14620. See that PR and its corresponding ticket for the explanation behind this refactor.

# Test Plan
- Open the desktop app and smoke test running a protocol and opening/closing a maintenance run flow. Note that there shouldn't be any unsubscribe actions dispatched.
- During a protocol run, cancel the run.
- You should see an app-shell(-odd) log about being "successfully unsubscribed." 
- You can repeat the same steps on the ODD if you'd like, but the code is nearly identical. 

# Risk assessment
low

[EXEC-306]: https://opentrons.atlassian.net/browse/EXEC-306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app-shell-odd/src/notify.ts
```diff
@@ -235,43 +216,42 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
-              // )
-              handleDecrementSubscriptionCount(hostname, topic)
-            }
-          })
+    if (hostname in connectionStore && !pendingUnsubs.has(topic)) {
+      pendingUnsubs.add(topic)
+      const { client } = connectionStore[hostname]
+      client?.unsubscribe(topic, {}, (error, result) => {
+        if (error != null) {
+          // log.warn(
```
sfoster1, line 238: might want this back but definitely don't want it commented - let's uncomment or delete

app-shell-odd/src/notify.ts
```diff
@@ -235,43 +216,42 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
...
         }
-      } else {
-        // log.info(
-        //   `Attempted to unsubscribe from unconnected hostname: ${hostname}`
-        // )
-      }
-    }, RENDER_TIMEOUT)
+      })
+    } else {
+      // log.info(
```
sfoster1, line 251: let's uncomment or delete

app-shell-odd/src/notify.ts
```diff
@@ -434,3 +391,14 @@ function sendToBrowserDeserialized({
 
   browserWindow.webContents.send('notify', hostname, topic, deserializedMessage)
 }
+
+function deserialize(message: string): string | Record<string, unknown> {
```
sfoster1, line 395: Is there a case where this will be (a) valid and (b) not json? if so can we remove that case, and make this be a `Promise<Record<string, unknown>>` or even better a `Promise<NotifyMessage>` with a small list of valid models that rejects if it doesn't parse?

mjhuff: Yeah good point

mjhuff:
Soooo, I borked this PR (but the feedback I'll carry over!). 

Turns out I manufactured the end result that assumes discovery-client handles the connection logic alongside the unsubscribe logic. That's what I get for doing too many things on too many branches at once. 

Correcting this PR without creating a complex intermediary solution that utilizes the backend unsubscribe flags seems pretty pointless, as does changing one variable on the app side that refetches on the unsubscribe flag. 

I'll just re-open a PR with the end result that includes the discovery-client logic. That way, the node layer actually makes use of the new unsubscribe flags.

PR #14634, opened 2024-03-12
fix(step-generation): getNextTiprack now accounts for 4th column tips…

jerader:
… correctly

closes RQA-2491

# Overview

`sortLabwareBySlot`  previously wasn't listing the 4th column slots last. Kinda weird that a customer hasn't reported this yet since it seems like a bug that would commonly occur. But anyway this PR fixes it.

# Test Plan

Upload the protocol in the bug ticket and see that you don't run into a timeline error for the last step. Then delete a tiprack in the 2nd or 3rd column. See that the timeline error says that tipracks in the 4th column can't be pipetted into.

# Changelog

- change the sortLabwareBySlot logic to list the 4th column slots last

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/robotStateSelectors.ts
+++ b/step-generation/src/robotStateSelectors.ts
@@ -1,7 +1,6 @@
 import assert from 'assert'
 // TODO: Ian 2019-04-18 move orderWells somewhere more general -- shared-data util?
 import min from 'lodash/min'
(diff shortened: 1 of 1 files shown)


koji (approved):
lgtm
the test case worked as expected

PR #14636, opened 2024-03-12
feat(hardware): add a performance analysis system to the lld data processing script

ryanthecoder:
# Overview
This additions to the lld-data-script allows us to grab a whole directory full of final_report.csv's and process them, then using all of these aggregated results we can flag any failing run and score each algorithm for accuracy, precision, and combined performance.

Also added one more algorithm that uses the threshold logic, but using smoothing instead of raw data.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Very nice!

PR #14638, opened 2024-03-12
feat(robot-server): add runTimeParameters field to analysis response

sanni-t:
Closes AUTH-93

# Overview

First PR of RTP backend work! 
Updates the shape of completed analysis response to include `runTimeParameters`. This field will contain a list of all run time parameters defined in a protocol, with param values set to either the values sent by the client or default values from protocol if the client didn't provide any.

For this PR, the field will only return an empty list. Actual param parsing is part of a separate ticket.

# Test Plan

Integration and unit tests are sufficient for this but if need to test, 
- Upload a protocol via `/protocols` and verify that the completed analysis response contains an empty `runTimeParameters` list
- Check cli analysis response too

# Review requests

- mainly shape of the responses and naming of stuff
- Question: how should we handle older analyses that won't contain this field? Do we need to make this field optional to handle that case?

# Risk assessment

Medium. Modifies a critical resource but is very isolated

### DIFF
--- a/api/src/opentrons/protocol_engine/types.py
+++ b/api/src/opentrons/protocol_engine/types.py
@@ -842,7 +842,7 @@ class TipPresenceStatus(str, Enum):
 class RTPBase(BaseModel):
     """Parameters defined in a protocol."""
 
-    displayLabel: str = Field(..., description="Display string for the parameter.")
+    displayName: str = Field(..., description="Display string for the parameter.")
     variableName: str = Field(..., description="Python variable name of the parameter.")
     description: str = Field(..., description="Detailed description of the parameter.")
     suffix: Optional[str] = Field(
@@ -881,11 +881,18 @@ class FloatParameter(RTPBase):
     )
 
 
+class EnumChoice(BaseModel):
+    """Components of choices used in RTP Enum Parameters."""
+
+    displayName: str = Field(..., description="Display string for the param's choice.")
+    value: str = Field(..., description="Enum value of the param's choice.")
+
+
 class EnumParameter(RTPBase):
     """A string enum defined in a protocol."""
 
-    choices: List[str] = Field(
-        ..., description="List of valid values for this parameter."
+    choices: List[EnumChoice] = Field(
+        ..., description="List of valid choices for this parameter."
     )
     default: str = Field(
         ...,
--- a/api/src/opentrons/protocol_engine/types.py
+++ b/api/src/opentrons/protocol_engine/types.py
@@ -837,3 +837,60 @@ class TipPresenceStatus(str, Enum):
             HwTipStateType.PRESENT: TipPresenceStatus.PRESENT,
             HwTipStateType.ABSENT: TipPresenceStatus.ABSENT,
         }[state]
+
+
+class RTPBase(BaseModel):
+    """Parameters defined in a protocol."""
+
+    displayLabel: str = Field(..., description="Display string for the parameter.")
+    variableName: str = Field(..., description="Python variable name of the parameter.")
+    description: str = Field(..., description="Detailed description of the parameter.")
+    suffix: Optional[str] = Field(
+        None,
+        description="Units (like mL, mm/sec, etc) or a custom suffix for the parameter.",
+    )
+
+
+class IntParameter(RTPBase):
+    """An integer parameter defined in a protocol."""
+
+    min: int = Field(
+        ..., description="Minimum value that the integer param is allowed to have."
+    )
+    max: int = Field(
+        ..., description="Maximum value that the integer param is allowed to have."
+    )
+    default: int = Field(
+        ...,
+        description="Default value of the parameter, to be used when there is no client-specified value.",
+    )
+
+
+class FloatParameter(RTPBase):
+    """A float parameter defined in a protocol."""
+
+    min: float = Field(
+        ..., description="Minimum value that the float param is allowed to have."
+    )
+    max: float = Field(
+        ..., description="Maximum value that the float param is allowed to have."
+    )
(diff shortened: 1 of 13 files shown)


api/src/opentrons/protocol_engine/types.py
```diff
@@ -837,3 +837,60 @@ def from_hw_state(cls, state: HwTipStateType) -> "TipPresenceStatus":
...
+    default: float = Field(
+        ...,
+        description="Default value of the parameter, to be used when there is no client-specified value.",
+    )
+
+
+class EnumParameter(RTPBase):
+    """A string enum defined in a protocol."""
+
+    choices: List[str] = Field(
```
y3rsh: Each choice will need to have a `displayLabel` and a `value`.
For example you are allowing a choice for the left pipette.
Top level `displayLabel` for the EnumParameter is `Left Pipette`
We want to have a friendly name in the dropdown to select.
```json
{
    "displayLabel": "Flex 1-Channel 1-50 µL",
    "value": "flex_1channel_50"
},
{
    "displayLabel": "Flex 1-Channel 5-1000 µL",
    "value": "flex_1channel_1000"
}
```

api/src/opentrons/protocol_engine/types.py
```diff
@@ -837,3 +837,60 @@ def from_hw_state(cls, state: HwTipStateType) -> "TipPresenceStatus":
             HwTipStateType.PRESENT: TipPresenceStatus.PRESENT,
             HwTipStateType.ABSENT: TipPresenceStatus.ABSENT,
         }[state]
+
+
+class RTPBase(BaseModel):
+    """Parameters defined in a protocol."""
+
+    displayLabel: str = Field(..., description="Display string for the parameter.")
```
jerader: thoughts about calling this `displayName` instead? imo, `displayName` is a bit more straight forward. `label` comes across as a nickname or short-hand name or something to me.

api/src/opentrons/protocol_engine/types.py
```diff
@@ -837,3 +837,60 @@ def from_hw_state(cls, state: HwTipStateType) -> "TipPresenceStatus":
...
+    min: int = Field(
+        ..., description="Minimum value that the integer param is allowed to have."
+    )
+    max: int = Field(
+        ..., description="Maximum value that the integer param is allowed to have."
+    )
+    default: int = Field(
+        ...,
+        description="Default value of the parameter, to be used when there is no client-specified value.",
+    )
```
SyntaxColoring, line 866: Beware that Pydantic has very nasty type coercion behavior by default. When parsing an `int` field, it will silently truncate a JSON `3.14159` value into `3`.

I think the way to fix this in our version Pydantic, if you want to, is `pydantic.StrictInt`.

robot-server/robot_server/protocols/analysis_models.py
```diff
@@ -102,6 +104,16 @@ class CompletedAnalysis(BaseModel):
             " in analyses that were originally created on older versions."
         ),
     )
+    runTimeParameters: Optional[List[RunTimeParameter]] = Field(
+        default_factory=list,  # Should this be None?
```
SyntaxColoring: I think it probably doesn't matter, but I would do either of these:

* Type it as `Optional[List[RunTimeParameter]]` and do `default=None`
* Type it as `List[RunTimeParameter]` and do `default_factory=list`

sanni-t: Picked option 2 to be consistent with (most) other fields, and also since it provides a consistent shape of analysis response. 

Verified that it doesn't impact backwards or cross-version server-client compatibility:
- previous analyses that don’t have `runTimeParameters` field, get returned from `/protocols/protocol_id/analyses` with `runTimeParameters: []` in the analysis response
- if the app is on an older version and fetching analysis of a protocol that has `runTimeParameters`, then it ignores that field
- if a robot has analyses that contain `runTimeParameters` and the bot gets downgraded to before RTP support, then the response from `/protocols/protocol_id/analyses` for that protocol doesn’t show the runTimeParameters and that field just gets ignored

api/src/opentrons/protocol_engine/types.py
sanni-t: I just remembered that we've refrained from using `StrictInt` in fields that are similar to this one where the expected type on the backend is an integer. We've mostly left the fields be `floats` with purposeful validation/ conversion down the line. I don't remember the exact reasons (maybe you or @shlokamin know why) but that actually makes things simpler and more straightforward in this case. 

 If we only have a `float` param model instead of separate `int` & `floats`, (we will still have different int & float param API functions) and rely on the engine-based param validator and app's validation (which we will implement anyway), then we won't run into weird pydantic coercions.

sanni-t: The other option is to keep the two different models and make them strict type so that `Union[intParam, floatParam]` won't try to coerce one type into other, but it feels a bit unnecessary since we will actually have float vs int validation on both ends of this client-server communication.

Let me know if I'm missing something though.

SyntaxColoring: Sure, whatever is easiest for you all.

sanni-t: Cool, I'm going to merge this as is once it gets an approval since I am blocked on it, but will make a change in the follow-up. It's not a critical issue and I'm sure we'll have a few more changes to this.

y3rsh (approved):
LGTM!

jerader (approved):
lgtm, thank you!

SyntaxColoring (approved):
All looks good to me. Thanks!

PR #14639, opened 2024-03-12
refactor(api): Add an `enableErrorRecoveryExperiments` feature flag

SyntaxColoring:
# Overview

Closes [EXEC-317](https://opentrons.atlassian.net/browse/EXEC-317).

# Test Plan

There's nothing meaningful to test here until further work wires stuff up to this feature flag.

# Changelog

Add an `enableErrorRecoveryExperiments` feature flag, which currently doesn't do anything, but will help with future work in EXEC-284.

# Review requests

I've left `internal_only` set to `False`, which I guess means this will show up in the Opentrons App. I figure we want to make it easy for our internal engineers to toggle this on and off.

Is the description text sufficient to turn away people who shouldn't be messing with this, or do we want to set `internal_only` to `True`?

# Risk assessment

Low.

[EXEC-317]: https://opentrons.atlassian.net/browse/EXEC-317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -219,6 +219,19 @@ settings = [
         description="When this setting is on, Flex will continue its activities regardless of pressure changes inside the pipette. Do not turn this setting on unless you are intentionally causing pressures over 8 kPa inside the pipette air channel.",
         robot_type=[RobotTypeEnum.FLEX],
     ),
+    SettingDefinition(
+        _id="enableErrorRecoveryExperiments",
+        title="Enable error recovery experiments",
+        description=(
+            "Do not enable."
+            " This is an Opentrons internal setting to experiment with"
+            " in-development error recovery features."
+            " This will interfere with your protocol runs,"
+            " corrupt your robot's storage,"
+            " bring misfortune and pestilence upon you and your livestock, etc."
+        ),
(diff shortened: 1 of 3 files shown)


api/src/opentrons/config/advanced_settings.py
```diff
@@ -219,6 +219,19 @@ class Setting(NamedTuple):
...
+        description=(
+            "Do not enable."
+            " This is an Opentrons internal setting to experiment with"
+            " in-development error recovery features."
+            " This will interfere with your protocol runs,"
+            " corrupt your robot's storage,"
+            " bring misfortune and pestilence upon you and your livestock, etc."
+        ),
+        robot_type=[RobotTypeEnum.FLEX],
+    ),
```
DerekMaggio: I am going to need more fire and brimstone to approve this. Something about curses upon your children should suffice.

DerekMaggio (approved):
See comment, otherwise looks good to me

sfoster1 (dismissed):
We should set it to `internal_only`. When there's more behind it and it's stable enough we can set it to be externally visible.

sfoster1 (commented):
Nice!

PR #14640, opened 2024-03-12
refactor(app-shell, app-shell-odd): Refactor app to use unsubscribe flags

mjhuff:
Closes [EXEC-306](https://opentrons.atlassian.net/browse/EXEC-306)

# Overview

This is the frontend implementation of https://github.com/Opentrons/opentrons/pull/14620. See that PR and its corresponding ticket for the explanation behind this refactor. This also includes the feedback from https://github.com/Opentrons/opentrons/pull/14633.

Note: This PR does NOT implement any shell logic changes based on the new unsubscribe flag, as that would unnecessarily increase the complexity of the shell logic without any current benefit. The payoff of using unsubscribe flags is contingent upon app-shell connectivity logic being tied to discovery-client (which will happen in a follow up PR) and the ODD connectivity logic not trying to account for connecting to other hosts (again, another PR). 

# Test Plan
- Open the desktop app and smoke test running a protocol and opening/closing a maintenance run flow. 
- During a protocol run, cancel the run. The app should successfully handle the close without getting caught in an infinite loop (this is where the new "unsubscribe flag" is dispatched). 
- You can repeat the same steps on the ODD if you'd like, but the code is nearly identical. 

# Risk assessment
low

[EXEC-306]: https://opentrons.atlassian.net/browse/EXEC-306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -257,6 +257,8 @@ function unsubscribe(notifyParams: NotifyParams): Promise<void> {
           client?.unsubscribe(topic, {}, (error, result) => {
             if (error == null) {
               handleDecrementSubscriptionCount(hostname, topic)
+            } else {
+              log.warn(`Failed to subscribe on ${hostname} to topic: ${topic}`)
             }
           })
         } else {
--- a/app-shell-odd/src/notify.ts
+++ b/app-shell-odd/src/notify.ts
@@ -190,7 +190,6 @@ function subscribe(notifyParams: NotifyParams): Promise<void> {
   function subscribeCb(error: Error, result: mqtt.ISubscriptionGrant[]): void {
     const { subscriptions } = connectionStore[hostname]
     if (error != null) {
-      // log.warn(`Failed to subscribe on ${hostname} to topic: ${topic}`)
       sendToBrowserDeserialized({
         browserWindow,
         hostname,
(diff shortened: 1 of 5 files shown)


app-shell-odd/src/notify.ts
```diff
@@ -252,24 +255,13 @@ function unsubscribe(notifyParams: NotifyParams): Promise<void> {
...
           client?.unsubscribe(topic, {}, (error, result) => {
-            if (error != null) {
-              // log.warn(
-              //   `Failed to unsubscribe on ${hostname} from topic: ${topic}`
-              // )
-            } else {
-              // log.info(
-              //   `Successfully unsubscribed on ${hostname} from topic: ${topic}`
-              // )
+            if (error == null) {
```
sfoster1, line 258: I think we do actually want to log if this occurred, don't we? Wouldn't errors during unsubscribe cause a memory leak?

sfoster1 (approved):
Looks good to me, maybe we should bring back some of that logging though.

app-shell-odd/src/notify.ts
mjhuff: Bringing back the logging next PR™ when discovery-client doing the connection logic will make things much less spammy.

Yeah, true true. They definitely could, but because topics are constricted to a few types and there are only a finite amount of hostnames on a given network, it shouldn't be anything substantial.

mjhuff: On second thought, No, you're right. This one can't be spammy and it can't hurt to have it 👍

PR #14646, opened 2024-03-13
feat(api): Pause run after recoverable errors

SyntaxColoring:
# Overview

This closes most of EXEC-301 and closes EXEC-320.

# Changelog

## What works now (or is intended to work, anyway)

* When running a JSON protocol on a Flex, if a `pickUpTip` command fails because of a missing tip, the run pauses now, instead of homing and showing an error message.
* You can then either:
    * Cancel the run in the usual way, by issuing a `stop` action via `POST /runs/{id}/actions`.
    * Resume the run from the next command, by issuing a `resume-from-recovery` action via `POST /runs/{id}/actions`.

## What doesn't work

* After a `pickUpTip` command failure, the robot seems to simultaneously think it does not have a tip attached (a subsequent `aspirate` command will fail with a "tip not attached" error) and that it does have a tip attached (a subsequent `moveTo` will act like there's a tip). I haven't had a chance to look into this yet. So the practical usefulness of this is pretty limited so far: you can only meaningfully recover if the protocol doesn't have any aspirates or dispenses.
* After recovering from an error and proceeding through the rest of the run, the run will end as `failed` despite the recovery. I see what's causing this, I just haven't gotten to it yet.
* You can't issue any fixup commands yet.
* You can't recover from errors in Python protocols yet. EXEC-339 is the next step to accomplish that.
* This isn't remotely supported by the Opentrons App or Flex on-device display. You'll see weird things like buttons without labels.

# Test Plan

## Setup

Enable the hidden error recovery feature flag by manually issuing a `POST /settings` request with `{"id": "enableErrorRecoveryExperiments", "value": true}`.

## Test

1. Upload this test protocol to a Flex: [Recovery test.json.zip](https://github.com/Opentrons/opentrons/files/14654980/Recovery.test.json.zip). This comes from Protocol Designer, with `aspirate`s and `dispense`s hand-replaced by `moveToWell`s in order to work around the tip-attached confusion that I've described above.
2. Fill random wells in the leftmost column of the tip rack with tips.
3. Start the run.
4. When the run reaches a missing tip, it should pause. Resume it by issuing a `{"data": {"actionType": "resume-from-recovery"}}` action to `POST /runs/{id}/actions`, or cancel it by issuing a `{"data": {"actionType": "stop"}}` action.

## Cleanup

1. Delete any runs that you just created, to avoid database incompatibilities when you push a different branch. You can do this with the Opentrons App or with manually-issued `DELETE /runs/{id}` requests.
5. `POST /settings` with `{"id": "enableErrorRecoveryExperiments", "value": false}`.

# Review requests

None in particular.

The easiest way to review this might be to filter by commit. I've organized this into refactor commits vs. feature commits.

# Risk assessment

Medium. This touches core internal Protocol Engine logic. Although it's all technically well covered by unit tests, those tests are mostly too isolated and low-level to catch actual bugs.

### DIFF
--- a/api/src/opentrons/protocol_engine/actions/actions.py
+++ b/api/src/opentrons/protocol_engine/actions/actions.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
 from typing import Optional, Union
+from opentrons.protocol_engine.error_recovery_policy import ErrorRecoveryType
 
 from opentrons.protocols.models import LabwareDefinition
 from opentrons.hardware_control.types import DoorState
@@ -129,18 +130,13 @@ class UpdateCommandAction:
 
 @dataclass(frozen=True)
 class FailCommandAction:
-    """Mark a given command as failed.
+    """Mark a given command as failed."""
 
-    The given command and all currently queued commands will be marked
-    as failed due to the given error.
-    """
-
-    # TODO(mc, 2021-11-12): we'll likely need to add the command params
-    # to this payload for state reaction purposes
     command_id: str
     error_id: str
     failed_at: datetime
     error: EnumeratedError
+    type: ErrorRecoveryType
 
 
 @dataclass(frozen=True)
--- a/api/src/opentrons/protocol_engine/error_recovery_policy.py
+++ b/api/src/opentrons/protocol_engine/error_recovery_policy.py
@@ -1,0 +1,76 @@
+# noqa: D100
+
+import enum
+from typing import Protocol
+
+from opentrons_shared_data.errors import EnumeratedError, ErrorCodes
+
+from opentrons.config import feature_flags as ff
+from opentrons.protocol_engine.commands import Command
+
+
+class ErrorRecoveryType(enum.Enum):
+    """Ways to handle a command failure."""
+
+    FAIL_RUN = enum.auto()
+    """Permanently fail the entire run.
+
+    TODO(mm, 2024-03-18): This might be a misnomer because failing the run is not
+    a decision that's up to Protocol Engine. It's decided by what the caller supplies
+    to `ProtocolEngine.finish()`. For example, a Python protocol can
+    theoretically swallow the exception and continue on.
+    """
+
+    WAIT_FOR_RECOVERY = enum.auto()
+    """Stop and wait for the error to be recovered from manually."""
+
+    # TODO(mm, 2023-03-18): Add something like this for
+    # https://opentrons.atlassian.net/browse/EXEC-302.
+    # CONTINUE = enum.auto()
+    # """Continue with the run, as if the command never failed."""
+
+
+class ErrorRecoveryPolicy(Protocol):
+    """An interface to decide how to handle a command failure.
+
+    This describes a function that Protocol Engine calls after each command failure,
+    with the details of that failure. The implementation should inspect those details
+    and return an appropriate `ErrorRecoveryType`.
+    """
+
+    @staticmethod
+    def __call__(  # noqa: D102
+        failed_command: Command, exception: Exception
+    ) -> ErrorRecoveryType:
+        ...
+
+
+def error_recovery_by_ff(
+    failed_command: Command, exception: Exception
+) -> ErrorRecoveryType:
+    """Use API feature flags to decide how to handle an error.
+
+    This is just for development. This should be replaced by a proper config
+    system exposed through robot-server's HTTP API.
+    """
+    # todo(mm, 2024-03-18): Do we need to do anything explicit here to disable
+    # error recovery on the OT-2?
+    if ff.enable_error_recovery_experiments() and _is_recoverable(
+        failed_command, exception
+    ):
+        return ErrorRecoveryType.WAIT_FOR_RECOVERY
+    else:
+        return ErrorRecoveryType.FAIL_RUN
+
+
+def _is_recoverable(failed_command: Command, exception: Exception) -> bool:
+    if (
+        failed_command.commandType == "pickUpTip"
+        and isinstance(exception, EnumeratedError)
+        # Hack(?): It seems like this should be ErrorCodes.TIP_PICKUP_FAILED, but that's
+        # not what gets raised in practice.
+        and exception.code == ErrorCodes.UNEXPECTED_TIP_REMOVAL
+    ):
+        return True
+    else:
+        return False
--- a/api/src/opentrons/protocol_engine/execution/command_executor.py
+++ b/api/src/opentrons/protocol_engine/execution/command_executor.py
@@ -11,6 +11,8 @@ from opentrons_shared_data.errors.exceptions import (
     PythonException,
 )
 
+from opentrons.protocol_engine.error_recovery_policy import ErrorRecoveryPolicy
+
 from ..state import StateStore
 from ..resources import ModelUtils
 from ..commands import (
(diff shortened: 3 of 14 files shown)


api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -122,18 +123,13 @@ class UpdateCommandAction:
...
-    as failed due to the given error.
-    """
-
-    # TODO(mc, 2021-11-12): we'll likely need to add the command params
-    # to this payload for state reaction purposes
     command_id: str
     error_id: str
     failed_at: datetime
     error: EnumeratedError
+    type: ErrorRecoveryType
```
DerekMaggio, line 139: What about separating this into `FailCommandRecoverableAction` and `FailCommandAction`? That way we maintain current functionality for fatal errors. Then all new error recovery functionality is isolated explicitly? 

Otherwise, we will end up with completely diverging functionality inside the same class based off of whether the error is recoverable

sfoster1: We definitely want to keep this as just the one `FailCommandAction` because whether the error was later recoverable or not, _this command failed_. Nothing we do in error recovery or handling changes that. The command always fails, the engine always reacts to it in some way. The difference is that now the engine may react to it by suspending and waiting for recovery, rather than by cancelling; it may react by continuing. But either way, this command failed.

SyntaxColoring: Some notes from the trenches:

* I'm embracing the intrusiveness. Error recovery is such a fundamental change to Protocol Engine that, in the end, we won't be able to isolate it and leave the old code alone. So we might as well confront intrusive changes like this up-front.
* Even setting that aside, there's not a huge difference between a single action with a `type` discriminator vs. two different actions. It's just whichever turns out to be easiest to follow internally. So far, I think that's turning out to be a single action with a `type` discriminator, but it won't be too hard to change later if we want.

api/src/opentrons/protocol_engine/error_recovery_policy.py
```diff
@@ -0,0 +1,73 @@
...
+    if (
+        failed_command.commandType == "pickUpTip"
+        and isinstance(exception, EnumeratedError)
+        # Hack(?): It seems like this should be ErrorCodes.TIP_PICKUP_FAILED, but that's
+        # not what gets raised in practice.
+        and exception.code == ErrorCodes.UNEXPECTED_TIP_REMOVAL
+    ):
+        return True
+    else:
+        return False
```
DerekMaggio, line 76: I can foresee this function getting huge. As we add more recoverable states, will this eventually be replaced with a more standardized approach?

DerekMaggio (commented):
Took a look at your new commits. 
I like your todos for future tickets. I am stealing that.

Not directly related to your PR, but the logic in handle_action of protocol_engine/state/commands.py is quite complex and easy to get lost in. 
I wonder if there is a way to end up cleaning that up? Maybe deferring each of the handling logic to their own function?

api/src/opentrons/protocol_engine/error_recovery_policy.py
SyntaxColoring: Yeah, I'm also worried about this, and I don't know.

The opposite way to organize this would be to distribute this choice across the command implementations. So, instead of `PickUpTipImplementation` returning an error, and this deciding whether that error is recoverable, `PickUpTipImplementation` would return an error *and* a `recoverable` boolean (or something like that).

DerekMaggio: Is it true that any executed command, that failed and is recoverable, would throw an EnumeratedError exception? 

If so, could the command implementations be required to have an `is_recoverable(error_code: ErrorCodes)` method?

SyntaxColoring: There are probably places where that's not true today, but we can (and might have to) aim for that to become true as part of this project.

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -177,6 +181,18 @@ async def execute(self, command_id: str) -> None:
...
                     failed_at=self._model_utils.get_timestamp(),
+                    # todo(mm, 2024-03-13):
+                    # When a command fails recoverably, and we handle it with
+                    # WAIT_FOR_RECOVERY or CONTINUE, we want to update our logical
+                    # protocol state as if the command succeeded. (e.g. if a tip
+                    # pickup failed, pretend that it succeeded and that the tip is now
+                    # on the pipette.) However, this currently does the opposite,
+                    # acting as if the command never executed.
+                    type=self._error_recovery_policy(
+                        # todo: should possibly be command_with_new_notes
```
TamarZanzouri: can you explain this todo?

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -288,13 +312,25 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
                         command_id=id, failed_at=action.failed_at, error_occurrence=None
                     )
                 self._state.queued_setup_command_ids.clear()
+            elif (
+                prev_entry.command.intent == CommandIntent.PROTOCOL
+                or prev_entry.command.intent is None
```
TamarZanzouri, line 317: sorry I might be confusing something but when is this set to None? I thought we use a default value for this

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -288,13 +312,25 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
...
+                else:
+                    assert_never(action.type)
             else:
-                other_command_ids_to_fail = self._state.queued_command_ids
-                for id in other_command_ids_to_fail:
-                    self._update_to_failed(
-                        command_id=id, failed_at=action.failed_at, error_occurrence=None
-                    )
-                self._state.queued_command_ids.clear()
+                assert_never(prev_entry.command.intent)
```
TamarZanzouri, line 333: this is new to me :-) good to know about this guy!

TamarZanzouri (approved):
a few comments but makes total sense and it looks great! nice job!

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: 🤷 The model [defines it](https://github.com/Opentrons/opentrons/blob/edba100b11a8fda57e5b655096668cf3dfe20def/api/src/opentrons/protocol_engine/commands/command.py#L146-L154) as defaulting to `None`. It might have to do with wanting HTTP commands to default to `"setup"` but other commands to default to `"protocol"`.

api/src/opentrons/protocol_engine/execution/command_executor.py
SyntaxColoring:
```python
# todo: should possibly be command_with_new_notes
```

☝️ This one looks left over from development, sorry. I'll remove it.

```python
# todo(mm, 2024-03-13):
# When a command fails recoverably, and we handle it with
# WAIT_FOR_RECOVERY or CONTINUE, we want to update our logical
# protocol state as if the command succeeded. (e.g. if a tip
# pickup failed, pretend that it succeeded and that the tip is now
# on the pipette.) However, this currently does the opposite,
# acting as if the command never executed.
```

☝️ This one is related to this point from EXEC-301:

> * The failed command should affect the engine's logical state as if it had succeeded, per Slack discussion [link]

Because this is a command failure, not a success, Protocol Engine will *not* currently accomplish that point.

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: I think it's probably "new" with our recent mypy and typing-extensions updates!

SyntaxColoring:
I've run through the test plan in the PR description, and also run a couple of other random protocols as a smoke test to make sure this didn't break anything else.

PR #9170, opened 2021-12-27
perf(protocol-engine): avoid iterating all commands for every state update

mcous:
## Overview

This PR refactors the logic used to lookup the next queued command ID. Previously, this logic was O(n) in the worst case, and it was called by the engine's `QueueWorker` on every single state update. Legacy protocol execution triggered the worst case due to its bypass of the `queued` command state.

Closes #9109, closes #8739. Hopefully also has an impact on #9133.

## Changelog

- Maintain a separate list of queued command IDs to ensure queue lookup is always O(1)
- Maintain PE state in mutable dataclasses vs immutable ones
    - Were not using immutability at the store level
    - Going mutable gets us benefits for performance _and_ typesafety
    - If/when we need immutability, we should evaluate a performant library
- Further improve run status reporting to remove reliance on iterating over the commands and errors lists 

**Important:** This PR _also_ removes the "bail on first PE command failure" logic because:

- It was easier to avoid re-implementing it
- We don't really want to keep the behavior, per #8739

This **does not affect protocol running bails**, just bails during the "live" LPC portion. **This probably has client-facing impacts**.

### Client callouts

- The session status will no longer switch back and forth between `running` and `idle` during the non-LPC phase of the run
- A failed command during LPC will not make the run unusable, nor will it interfere with run status reporting
- A new run status has been added: `finishing`. This status is set when the run has completed and the robot is currently shutting down the hardware (and, in the future, dropping tips and homing)

## Review requests

- [ ] Smoke tests of complex protocol analyses and runs
- [ ] Smoke tests of LPC

## Risk assessment

Medium low

PR #14647, opened 2024-03-13
refactor(api): Split CommandStore

SyntaxColoring:
# Background

[This file](https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/protocol_engine/state/commands.py) has been a bit uncomfortably bloated for a while. It's responsible for at least two broad areas:

1. Run-level concerns, like getting "the overall run error," or choosing "the next command to execute."
2. Lower-level data organization concerns, to support stuff like O(1) retrieval of a command by its ID, or O(1) retrieval of the next queued command.

The other problem is that it uses the old test style of splitting the [store](https://github.com/Opentrons/opentrons/blob/edge/api/tests/opentrons/protocol_engine/state/test_command_store.py) from the [view](https://github.com/Opentrons/opentrons/blob/edge/api/tests/opentrons/protocol_engine/state/test_command_view.py). This creates thousands of lines of tests that don't pull their weight in catching bugs, and, I find, make it harder to work on this file.

All of this has contributed to maintenance problems (here are apparently two [different](https://github.com/Opentrons/opentrons/blob/392d83bc23d85c261b440723df8f5b3f5c9ba3c8/api/src/opentrons/protocol_engine/state/commands.py#L470-L471) [ways](https://github.com/Opentrons/opentrons/blob/392d83bc23d85c261b440723df8f5b3f5c9ba3c8/api/src/opentrons/protocol_engine/state/commands.py#L536-L544) of computing the same thing, for example), and performance problems [past](https://github.com/Opentrons/opentrons/pull/9170) and [ongoing](https://github.com/Opentrons/opentrons/blob/392d83bc23d85c261b440723df8f5b3f5c9ba3c8/api/src/opentrons/protocol_engine/state/commands.py#L534-L535).

This is going to get worse because of error recovery, which will add new run states and maybe command states.

So, the plan here is:

* Pull the lower-level data organization stuff out into its own data structure. Its responsibility will be to provide O(1) access to commands, without relying on knowledge of higher-level things like runs.
* That single type will take over for at least [these fields](https://github.com/Opentrons/opentrons/blob/392d83bc23d85c261b440723df8f5b3f5c9ba3c8/api/src/opentrons/protocol_engine/state/commands.py#L103-L116) in CommandState.
* Unit-test as a pure functional black box.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/ordered_set.py
+++ b/api/src/opentrons/ordered_set.py
@@ -81,7 +81,7 @@ class OrderedSet(Generic[_SetElementT]):
     def head(
         self, default_value: Union[_DefaultValueT, _NOT_SPECIFIED] = _NOT_SPECIFIED()
     ) -> Union[_SetElementT, _DefaultValueT]:
-        """Get the head of the set.
+        """Get the head (oldest-added element) of the set.
 
         Args:
             default_value: A value to return if set is empty.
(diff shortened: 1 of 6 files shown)


SyntaxColoring:
Closing for now to prioritize other things.

PR #14648, opened 2024-03-13
refactor(app-shell): Migrate desktop app notify connectivity to discovery-client

mjhuff:
Closes [EXEC-304](https://opentrons.atlassian.net/browse/EXEC-304)

# Overview

Migrating MQTT connectivity logic to discovery-client simplifies  app-shell connectionStore manager logic (maybe), reduces bug surface, and gives us connection retry logic relatively for free. 

Most importantly, we will not depend on components to express interest in a broker and topics, which implicitly requires all components and notify helper functions in the app to behave appropriately (not re-render like crazy). 

Note that there's some browser layer logic that is no longer relevant (like UNSUBSCRIBE actions), but these still need to be here until app-shell-odd is refactored. While the app still dispatches these actions, they don't actively do anything.

App-shell notifications behave as follows:
- Broker connections are added or forcefully removed based on health status changes reported by discovery-client.
- Subscriptions are handled "lazily", in which a component must express interest in a topic before a subscription request is made (no network request will be made if already subscribed tot he topic). 
- Unsubscribe requests only occur if an "unsubscribe" flag is received from the broker. That means the only thing we actively unsubscribe from at the moment is "/runs/:runId".
- Pending subs and unsubs are used to prevent unnecessary network and broker load.
- A blocked port event can only be reported a maximum of one time per robot per app session (no change here). 

# Test Plan
- The easiest way to test this is to do a `setInterval(() => {console.log(connectionStore)}, 1000)` within `notify.ts` as this gives direct feedback on what's happening.
- Start the app in dev mode, you should see a bunch of app-shell logs in which the app tries to connect to all robots that are _available_. Note that on our network, there are a lot of OT-2s and Flexes on 7.0.2, so it's fine to see a bunch of failed connections. Some should still go through through.
- Click on the Devices page. In the app-shell logs, you should see a bunch of subscribe actions going through the IPC and the app-shell should report successful subscribes. 
- Turn off a robot. Whenever the discovery-client polls and updates the available robots tab, you should see that robot be removed from the connection store. You won't actually see the "closed connection" log, because the robot is completely offline.
- Run a protocol and cancel it. I don't remember if it's when you restart a run or as soon as the cancel process is complete, but at some point the app-shell logs should say "successfully unsubscribed..." with the topic string that matches the run ID (this is the only topic with an unsubscribe flag for now). 

# Risk assessment
medium-ish. This is a pretty big refactor of how notifications work for the desktop app. 

[EXEC-304]: https://opentrons.atlassian.net/browse/EXEC-304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app-shell/src/notify.ts
```diff
@@ -55,128 +51,144 @@ const connectOptions: mqtt.IClientOptions = {
...
-            })
-            .catch(() => {
-              sendToBrowserDeserialized({
-                browserWindow,
-                hostname,
-                topic,
-                message: FAILURE_STATUSES.ECONNFAILED,
-              })
+        if (!subscriptions.has(topic)) {
+          if (!pendingSubs.has(topic)) {
```
DerekMaggio: I am missing something, what is the `pendingSubs` check doing here? 

My understanding: 
`client.subscribe` is blocking, so the client will not be able to try to subscribe to the same topic multiple times. 

Does it have something to do with electron calling the registerNotify actionHandler super frequently?

DerekMaggio: Never mind I think it is because of the waitUnitActiveOrErrored retry logic. @mjhuff, for my understanding, can you weight in

mjhuff: You're correct with the app layer calling the action super frequently. `Client.subscribe` is blocking only for the current `subscribe` call, which is why `pendingSubs` exists - this limits network traffic if there are multiple subscribe actions to the same topic dispatched at once - only one subscribe packet is sent instead of the number of subscribe actions.

MQTT does a good job of handling situations in which multiple subscribe packets are sent simultaneously, so nothing breaks if we accidentally make two identical subscribe requests. Therefore the goal of `pendingSubs` is only to limit unnecessary network traffic.

DerekMaggio: Thank you! That makes sense

app-shell/src/notify.ts
```diff
@@ -55,156 +51,172 @@ const connectOptions: mqtt.IClientOptions = {
...
+  robots: DiscoveryClientRobot[]
+): string[] {
+  const reachableRobotIPs = getHealthyRobotIPsForNotifications(robots)
+  cleanUpUnreachableRobots(reachableRobotIPs)
+  addNewRobotsToConnectionStore(reachableRobotIPs)
+
+  return reachableRobotIPs
+
+  // This is the discovery-client equivalent of "available" robots when viewing the Devices page in the app.
+  function getHealthyRobotIPsForNotifications(
```
sfoster1: can we pop this out of `handleNotificaitonConnectionsFor`? particularly with being defined after the rest of the body of the function, it's a little hard to parse this way and it doesn't seem to need to capture anything

app-shell/src/notify.ts
```diff
@@ -55,156 +51,172 @@ const connectOptions: mqtt.IClientOptions = {
...
+        delete connectionStore[hostname]
+        unreachableHosts.delete(hostname)
+        return true
+      }
+      return false
+    })
+    void closeConnectionsForcefullyFor(unreachableRobots)
+  }
+
+  function addNewRobotsToConnectionStore(robots: string[]): void {
```
sfoster1: let's have this return `Promise<void>[]` or something since that's what it's doing anyway, IMO we should default to returning instead of dangling promises for as long as we can

app-shell/src/notify.ts
```diff
@@ -55,156 +51,172 @@ const connectOptions: mqtt.IClientOptions = {
...
+    )
+  }
+
+  function cleanUpUnreachableRobots(healthyRobotIPs: string[]): void {
+    const healthyRobotIPsSet = new Set(healthyRobotIPs)
+    const unreachableRobots = Object.keys(connectionStore).filter(hostname => {
+      // The connection is forcefully closed, so remove from the connection store immediately to reduce disconnect packets.
+      if (hostname in connectionStore && !healthyRobotIPsSet.has(hostname)) {
+        delete connectionStore[hostname]
+        unreachableHosts.delete(hostname)
```
sfoster1: this means that if the robot 
1. is mqtt-unreachable
2. disappears from the health poller output
3. later reappears in the health poller output

we'll try to connect to it again, right?

app-shell/src/notify.ts
```diff
@@ -55,156 +51,172 @@ const connectOptions: mqtt.IClientOptions = {
...
+    robots: DiscoveryClientRobot[]
+  ): string[] {
+    return robots.flatMap(robot =>
+      robot.addresses
+        .filter(address => address.healthStatus === HEALTH_STATUS_OK)
+        .map(address => address.ip)
+    )
+  }
+
+  function cleanUpUnreachableRobots(healthyRobotIPs: string[]): void {
```
sfoster1: same here

app-shell/src/notify.ts
sfoster1: also let's move it out from the enclosing function

sfoster1 (changes requested):
I think I understand what this is doing, but it's getting really confusing. There's a lot of free-variable associative stores at the file scope, robots can be in multiple of them at once, there's several different pending states.

I think a lot of this will be cleared up and also testable if you consolidate some of this into a single object or closure with a couple methods that move robots between states atomically. Then have a file-scope instance of that, and have the various actions and callbacks use it to move robots around. At that point, it's pure logic and testable.

app-shell/src/notify.ts
mjhuff: Yep, exactly

mjhuff:
While testing I realized that I didn't account for the common scenario of discovery-client reporting multiple IPs for one robot (ethernet, wifi). We don't want multiple broker connections from the same app, since that's extremely inefficient. So the solution:

- IPs can be associated based on "name" (which is really robot name). This is guaranteed to be network unique, meaning that two IPs with the same name point to the same robot (and therefore MQTT broker).
- We take the first MDNS reported IP and try connecting to it for that robot. During in person convo w/ Seth, we brought up the idea of prioritizing connection based on selecting an IP with an identical subnet to the host. However, that introduces another layer of complexity - how do we decide to handle connections while MDNS is initially polling? Do we not connect to any broker for the first X seconds? Do we connect and then disconnect if we find an identical subnet? I think for now, it's fine to assume that if a robot is connected to multiple networks, it is free to use any of those networks. 
- If a network is port blocked, we will try connecting on alternative networks. Ex, if we connect to a robot's wifi network before ethernet, return an ECONNREFUSED during the connection ACK, we will try the ethernet network (if all networks fail, we just default to polling as before).
- Restarting a robot doesn't cause any funky issues. I've tested this thoroughly.  

Anyway, I think this is ready for review again. Let me know your thoughts. I will add testing once we've decided we're ok with this (hopefully final) iteration of the connection manager.

app-shell/src/notifications/store.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   *
+   * @description Marks the host as unreachable with an error status derived from the MQTT returned error object.
+   *
+   */
+  public setErrorStatus(ip: string, errorMessage: string): Promise<void> {
+    return new Promise((resolve, reject) => {
+      if (ip in this.hosts) {
+        if (this.hosts[ip].unreachableStatus == null) {
```
sfoster1: at a glance it seems like this will never resolve the wrapping promise if `ip in this.hosts` and `this.hosts[ip].unreachableStatus != null` but github's lineup of brackets might be deceiving me

app-shell/src/notifications/store.ts
```diff
@@ -0,0 +1,299 @@
...
+            FAILURE_STATUSES.ECONNREFUSED
+          )
+            ? FAILURE_STATUSES.ECONNREFUSED
+            : FAILURE_STATUSES.ECONNFAILED
+
+          this.hosts[ip].unreachableStatus = errorStatus
+          if (errorStatus === FAILURE_STATUSES.ECONNREFUSED) {
+            this.knownPortBlockedIPs.add(ip)
+          }
+          resolve()
```
sfoster1: i think this wants to go outside the `if (this.hosts[ip].unreachableStatus == null)` check since either way if we're in `if (ip in this.hosts)` we're resolving

app-shell/src/notifications/store.ts
```diff
@@ -0,0 +1,299 @@
...
+    status: 'pending' | 'subscribed'
+  ): Promise<void> {
+    return new Promise((resolve, reject) => {
+      if (ip in this.hosts) {
+        const { pendingSubs, subscriptions } = this.hosts[ip]
+        if (status === 'pending') {
+          pendingSubs.add(topic)
+        } else {
+          subscriptions.add(topic)
+          pendingSubs.delete(topic)
```
sfoster1, line 152: does `delete` throw if the argument wasn't present? if so we should document a requirement that a topic be pended before being added

app-shell/src/notifications/store.ts
```diff
@@ -0,0 +1,299 @@
...
+    status: 'pending' | 'unsubscribed'
+  ): Promise<void> {
+    return new Promise((resolve, reject) => {
+      if (ip in this.hosts) {
+        const { pendingUnsubs, subscriptions } = this.hosts[ip]
+        if (subscriptions.has(topic)) {
+          if (status === 'pending') {
+            pendingUnsubs.add(topic)
+          } else {
+            pendingUnsubs.delete(topic)
```
sfoster1, line 176: same question about `delete` throwing

app-shell/src/notifications/store.ts
```diff
@@ -0,0 +1,299 @@
...
+    })
+  }
+
+  public deleteAllAssociatedIPsGivenRobotName(
+    robotName: string
+  ): Promise<void> {
+    return new Promise((resolve, reject) => {
+      const associatedHosts = this.getAssociatedIPsFromRobotName(robotName)
+      associatedHosts.forEach(hostname => {
+        delete this.hosts[hostname]
```
sfoster1: based on `associateIPWithExistingHostData`, won't it sometimes be the case that two ips refer to not just the same broker but the same js object by-ref? does deleting the same thing twice cause problems, or is `delete` here punned to just be "pop from object" or decref or something

sfoster1 (changes requested):
More details in comments, but broad-strokes stuff:
- Since we appear to be focused on keeping one connection per broker, does it make sense to have the connection store have `robotNameByIp: Record<string, string>` and `hostsByRobotName: Record<string, HostData>`? that way we don't have to do a search for matching hostnames every time.

sfoster1:
Uh I submitted that too early I'll have followups

app-shell/src/notifications/connect.ts
```diff
@@ -0,0 +1,215 @@
...
+    void closeConnectionsForcefullyFor(unreachableRobots)
+  })
+}
+
+export function establishConnections(
+  healthyRobots: RobotData[]
+): Promise<void> {
+  return new Promise(() => {
+    const newConnections = healthyRobots.filter(({ ip, robotName }) => {
+      if (connectionStore.isConnectedToBroker(ip)) {
```
sfoster1: This might be a bit clearer if we treat `returns` as terminating, so rather than doing
```
if (cond) {
    return true 
} else if (cond2) {
 ...
}```

we do

```
if (cond) {
   return true
}
if (cond2) {
  ...
}
```

sfoster1 (approved):
Ok. Fundamentally looks good to me but I think there's some refactors we can do for clarity.

app-shell/src/notifications/store.ts
mjhuff: ah yep nice catch

app-shell/src/notifications/store.ts
mjhuff: Nope, this doesn't throw an error. Sets in JS fail silently for all operations that I can think of.

app-shell/src/notifications/store.ts
mjhuff: Yeah, like a Set it just "fails silently", but weirdly enough it more accurately "succeeds silently", since deleting a non-existent property returns true when caste as a boolean. Fun stuff...

PR #14650, opened 2024-03-13
fix(protocol-designer): cannot move trash into slot with a module

jerader:
closes RQA-2498 RQA-2499

# Overview

There seems to be an issue with this `getSlotEmpty` util all the time 😅  PD on edge was incorrectly stating that a slot was empty but it had a module in it and it was an issue with cutoutId vs slot. This PR fixes it

# Test Plan

create a flex protocol and add a thermocycler and the trash bin. Move the trash bin in slot A1 and B1 and see that it errors saying the slot is occupied for both slots. 

# Changelog

modify the util logic and account for `cutoutId`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/EditModulesModal/index.tsx
+++ b/protocol-designer/src/components/modals/EditModulesModal/index.tsx
@@ -395,7 +395,11 @@ const EditModulesModalComponent = (
             {slotIssue ? (
               <PDAlert
                 alertType="warning"
-                title={validation.selectedSlot}
+                title={
+                  robotType === OT2_ROBOT_TYPE
+                    ? t('alert:module_placement.SLOT_OCCUPIED.title')
+                    : validation.selectedSlot
+                }
                 description={''}
               />
             ) : null}
--- a/protocol-designer/src/step-forms/utils/index.ts
+++ b/protocol-designer/src/step-forms/utils/index.ts
@@ -120,6 +120,7 @@ export const getSlotIdsBlockedBySpanning = (
 
   return []
 }
+//  TODO(jr, 3/13/24): refactor this util it is messy and confusing
 export const getSlotIsEmpty = (
   initialDeckSetup: InitialDeckSetup,
   slot: string,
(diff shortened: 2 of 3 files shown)


protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -128,6 +128,13 @@ export const getSlotIsEmpty = (
   includeStagingAreas?: boolean
 ): boolean => {
   if (
+    slot === 'cutoutA1' &&
```
jerader, line 133: this is special casing for the thermocycler A1 slot

protocol-designer/src/components/modals/EditModulesModal/index.tsx
```diff
@@ -395,7 +395,11 @@ const EditModulesModalComponent = (
             {slotIssue ? (
               <PDAlert
                 alertType="warning"
-                title={validation.selectedSlot}
+                title={
+                  robotType === OT2_ROBOT_TYPE
+                    ? t('alert:module_placement.SLOT_OCCUPIED.title')
+                    : validation.selectedSlot
```
jerader, line 401: the `validation.selectedSlot` text is too long to fit on the ot-2 modal (the validation text was fixing a bug when i migrated to react hook form). so i changed it to just `cannot place module` text which is what we had before.

protocol-designer/src/step-forms/utils/index.ts
koji, line 1: Could you add `type` to L14?

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -120,14 +120,23 @@ export const getSlotIdsBlockedBySpanning = (
...
   slot: string,
   /* we don't always want to count the slot as full if there is a staging area present
      since labware/wasteChute can still go on top of staging areas  **/
   includeStagingAreas?: boolean
 ): boolean => {
+  //   special-casing the TC's slot A1 for the Flex
   if (
+    slot === 'cutoutA1' &&
+    Object.values(initialDeckSetup.modules).find(
+      module => module.type === 'thermocyclerModuleType'
```
koji:
```suggestion
      module => module.type === THERMOCYCLER_MODULE_TYPE
```

koji (approved):
I left a couple of comments but the changes make sense to me and the sandbox worked as expected.

PR #14651, opened 2024-03-13
refactor(app,api): Add an "awaiting-recovery" run status

SyntaxColoring:
# Overview

Closes EXEC-299.

# Changelog

* Set up our types for a new `awaiting-recovery` run status. It's not actually used by Protocol Engine or emitted by robot-server yet. See the Python docstring for the semantics that we have in mind so far.
* Attempt to update the many little bits of logic in the app that use the run status.

# Test plan

There's nothing substantial to test here yet because the new status isn't used by Protocol Engine or emitted by robot-server yet.

I've deliberately omitted unit tests because I think, at this stage, we want to keep it cheap to fundamentally rework the semantics of this status. But I'm happy to go back and add them if anyone thinks they'd be helpful.

# Review requests

Any questions or suggestions about the semantics of this new status?

# Risk assessment

Low until this new status actually gets emitted.

### DIFF
--- a/api-client/src/maintenance_runs/types.ts
+++ b/api-client/src/maintenance_runs/types.ts
@@ -4,7 +4,11 @@ import type {
   LoadedModule,
   LoadedPipette,
 } from '@opentrons/shared-data'
-import type { RunCommandSummary, LabwareOffsetCreateData, RunStatus } from '../runs'
+import type {
+  RunCommandSummary,
+  LabwareOffsetCreateData,
+  RunStatus,
+} from '../runs'
 
 export interface MaintenanceRunData {
   id: string
--- a/api-client/src/maintenance_runs/types.ts
+++ b/api-client/src/maintenance_runs/types.ts
@@ -4,35 +4,12 @@ import type {
   LoadedModule,
   LoadedPipette,
 } from '@opentrons/shared-data'
-import type { RunCommandSummary, LabwareOffsetCreateData } from '../runs'
-
-export const ENGINE_STATUS_IDLE = 'idle' as const
-export const ENGINE_STATUS_RUNNING = 'running' as const
-export const ENGINE_STATUS_PAUSE_REQUESTED = 'pause-requested' as const
-export const ENGINE_STATUS_PAUSED = 'paused'
-export const ENGINE_STATUS_STOP_REQUESTED = 'stop-requested' as const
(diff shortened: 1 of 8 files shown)


api-client/src/maintenance_runs/types.ts
```diff
@@ -4,35 +4,12 @@ import type {
...
-  | typeof ENGINE_STATUS_IDLE
-  | typeof ENGINE_STATUS_RUNNING
-  | typeof ENGINE_STATUS_PAUSE_REQUESTED
-  | typeof ENGINE_STATUS_PAUSED
-  | typeof ENGINE_STATUS_STOP_REQUESTED
-  | typeof ENGINE_STATUS_STOPPED
-  | typeof ENGINE_STATUS_FAILED
-  | typeof ENGINE_STATUS_FINISHING
-  | typeof ENGINE_STATUS_SUCCEEDED
-  | typeof ENGINE_STATUS_BLOCKED_BY_OPEN_DOOR
```
SyntaxColoring, line 30: This looks like it was a duplication between maintenance run statuses and regular run statuses. These constants (the maintenance run ones) were unused in practice, and I don't think we have a reason to keep them separated, so I just deleted them in favor of the regular run constants.

app/src/organisms/RunProgressMeter/index.tsx
```diff
@@ -83,6 +83,8 @@ export function RunProgressMeter(props: RunProgressMeterProps): JSX.Element {
   const runCommandsLength = allCommandsQueryData?.meta.totalLength
 
   const downloadIsDisabled =
+    // todo(mm, 2024-03-13): Given that this includes RUN_STATUS_RUNNING, does this
+    // intentionally exclude RUN_STATUS_PAUSED, RUN_STATUS_BLOCKED_BY_OPEN_DOOR, etc.?
```
mjhuff: The answer is almost certainly yes. My guess as to why is we don't want to let users download run logs in spots where the log can be currently updated while we're downloading.

mjhuff (approved):
Looks like you got all the relevant spots in the app! This LGTM.

sfoster1 (approved):
Python code looks good to me and so does the js changes although I think someone a little more in tune with how the app works probably should take a look there. We should pretty rapidly follow this up with re-adding the ability to cancel from this state at least.

PR #14652, opened 2024-03-13
feat(api): command executors can add notes

sfoster1:
This PR adds an interface for command executors to add notes to the command that they're executing, succeed or fail, and uses that interface to implement an example note warning of aspirate volume rounding.

The interface is done by callbacks from the command implementations to the command executor that is executing each command. Implementations, or the functions they call, are responsible for creating the CommandNote instances. The executor just gathers a list of commands and then issues an UpdateCommandAction with the new list. If there are already notes on the command before execution, the new ones are appended. While this doesn't seem necessary right now, we'll want this behavior generically when we have notes also coming from hardware.

One thing that's a little annoying about this interface is that the callbacks have to be threaded through all the way to whatever wants to add a note. For instance, in the example aspirate-rounded note, we have to pass the callback all the way into the pipetting executor. An alternative would be to have an interface that can add notes on one of the state instances that can be called from anything that can access the state instance. The thing is, that function won't have the context to know when commands are currently executing or are done executing without having some sort of state machine that only looks at UpdateCommandActions from elsewhere. That code would be pretty ugly. The alternative there would be to issue a new UpdateCommandAction for each note; this would have race condition issues if it provided new values for the notes list. The alternative _there_ would be a new action called like AddCommandNoteAction that carries the attention of adding to the notes list along with it. Of course, that command is not idempotent.

This interface works well enough for now that I think we should roll with it, and feel free to change it later.

## Still todo
- [x] add tests for actually submitting a new command note

## testing (once todos are todone)
- [x] in simulation, request an aspirate that is 0.0000001 ul higher than the pipette max and see the note in the output
- [x] on an executing flex, do the same
~- [ ] on an executing ot-2, do the same~ will test this in followups

Closes EXEC-291

### DIFF
--- a/api/src/opentrons/protocol_engine/execution/command_executor.py
+++ b/api/src/opentrons/protocol_engine/execution/command_executor.py
@@ -184,7 +184,7 @@ class CommandExecutor:
                 "result": result,
                 "status": CommandStatus.SUCCEEDED,
                 "completedAt": self._model_utils.get_timestamp(),
-                **_append_notes_if_notes(running_command, note_tracker.get_notes())
+                **_append_notes_if_notes(running_command, note_tracker.get_notes()),
             }
             completed_command = running_command.copy(update=update)
             self._action_dispatcher.dispatch(
--- a/api/src/opentrons/protocol_engine/execution/command_executor.py
+++ b/api/src/opentrons/protocol_engine/execution/command_executor.py
@@ -1,7 +1,7 @@
 """Command side-effect execution logic container."""
 import asyncio
 from logging import getLogger
-from typing import Optional
+from typing import Optional, List, Dict, Any, Protocol
 
 from opentrons.hardware_control import HardwareControlAPI
 
@@ -18,10 +18,12 @@ from ..commands import (
     AbstractCommandImpl,
     CommandResult,
     CommandPrivateResult,
+    Command,
 )
 from ..actions import ActionDispatcher, UpdateCommandAction, FailCommandAction
 from ..errors import RunStoppedError
 from ..errors.exceptions import EStopActivatedError as PE_EStopActivatedError
+from ..notes import CommandNote, CommandNoteTracker
 from .equipment import EquipmentHandler
 from .movement import MovementHandler
 from .gantry_mover import GantryMover
@@ -34,6 +36,29 @@ from .status_bar import StatusBarHandler
 
 
 log = getLogger(__name__)
+
+
+class CommandNoteTrackerProvider(Protocol):
+    """The correct shape for a function that provides a CommandNoteTracker.
+
+    This function will be called by the executor once for each call to execute().
+    It is mostly useful for testing harnesses.
+    """
+
+    def __call__(self) -> CommandNoteTracker:
+        """Provide a new CommandNoteTracker."""
+        ...
+
+
+class _NoteTracker(CommandNoteTracker):
+    def __init__(self) -> None:
+        self._notes: List[CommandNote] = []
+
+    def __call__(self, note: CommandNote) -> None:
+        self._notes.append(note)
+
+    def get_notes(self) -> List[CommandNote]:
+        return self._notes
 
 
 class CommandExecutor:
@@ -58,6 +83,7 @@ class CommandExecutor:
         rail_lights: RailLightsHandler,
         status_bar: StatusBarHandler,
         model_utils: Optional[ModelUtils] = None,
+        command_note_tracker_provider: Optional[CommandNoteTrackerProvider] = None,
     ) -> None:
         """Initialize the CommandExecutor with access to its dependencies."""
         self._hardware_api = hardware_api
@@ -73,6 +99,9 @@ class CommandExecutor:
         self._rail_lights = rail_lights
         self._model_utils = model_utils or ModelUtils()
         self._status_bar = status_bar
+        self._command_note_tracker_provider = (
+            command_note_tracker_provider or _NoteTracker
+        )
 
     async def execute(self, command_id: str) -> None:
         """Run a given command's execution procedure.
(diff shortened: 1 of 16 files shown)


SyntaxColoring:
Yeah, this fundamental interface looks good to me.

I agree that we don't want command executors to add notes by emitting `UpdateCommandAction`s or `AddCommandNoteAction`s. In addition to the reasons you mentioned, it's nice if we can continue to treat a command's execution as one atomic step that's committed with one `UpdateCommandAction`.

Re the annoyance of having to pass the callback all the way into the pipetting executor: it does seem doable to me to have something "globally" accessible that knows what the current note target is. It can delegate to `CommandView.get_current()`; it doesn't have to be a whole new independent state machine that monitors `UpdateCommandAction`s from scratch. We could even wire that up to the standard Python [warnings](https://docs.python.org/3/library/warnings.html) system, maybe.

I agree that this interface is good enough that we should roll with it.

DerekMaggio (commented):
This interface looks good to me. 
I do like Max's idea of having something global so you don't have to pass the callback so much.

sfoster1:
Yeah I honestly don't love the idea of a global callback like that because it gets pretty nonlocal

TamarZanzouri:
Spoke privately with Seth about this, I kind of prefer we use `AddNoteAction` instead of UpdateCommandAction. I feel like there is too much conditional logic for `UpdateCommandAction` and we know exactly what needs to be done when we dispatch a more specific action. `UpdateCommandAction` feels like its more in charge of the command status and timestamps.

TamarZanzouri (commented):
Spoke again to Seth about this, might make more sense to use UpdateCommandAction to append all notes at one time. This logic makes sense to me.

api/tests/opentrons/protocol_engine/note_utils.py
```diff
@@ -0,0 +1,63 @@
+"""Test utilities for dealing with notes."""
+import re
+from typing import Optional
+from opentrons.protocol_engine.notes import CommandNote
+
+
+class CommandNoteMatcher:
```
SyntaxColoring, line 7: Neat.

api/tests/opentrons/protocol_engine/execution/test_pipetting_handler.py
```diff
@@ -487,16 +502,33 @@ async def test_aspirate_volume_validation(
...
+        mock_command_note_adder(
+            cast(
+                CommandNote,
+                CommandNoteMatcher(
+                    matching_noteKind_regex="warning",
+                    matching_shortMessage_regex="Aspirate clamped to 1 µL",
+                ),
+            )
+        )
+    ).then_return(None)
```
SyntaxColoring: Because of the way Decoy works, I think this isn't actually testing anything right now.

In Decoy, when the test subject does a wrong function call (wrong in the sense that we haven't rehearsed it with `decoy.when(...)`), it will see a `None` return value and the test will continue. So if the subject here calls the command note adder wrongly, or forgets to call it at all, it's indistinguishable to this test from the subject calling it properly.

I would fix that by changing this `decoy.when(mock_command_note_adder(...))` to `decoy.verify(mock_command_note_adder(...))` and moving it after the `await subject.aspirate_in_place(...)` call.

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -36,6 +38,29 @@
...
+    def __init__(self) -> None:
+        self._notes: List[CommandNote] = []
+
+    def __call__(self, note: CommandNote) -> None:
+        self._notes.append(note)
+
+    def get_notes(self) -> List[CommandNote]:
+        return self._notes
+
+
```
SyntaxColoring, line 63: For simplicity, I'd be down for `CommandExecutor` to just construct each `CommandNoteTracker` directly, in a way that can't be overridden by `CommandExecutor`'s unit tests.

The downside is we couldn't test that it instantiates a separate tracker for each command, and I'm fine with that.

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -138,15 +180,27 @@ async def execute(self, command_id: str) -> None:
...
-                }
+            update = {
+                "result": result,
+                "status": CommandStatus.SUCCEEDED,
+                "completedAt": self._model_utils.get_timestamp(),
+            }
+            update.update(
+                _append_notes_if_notes(running_command, note_tracker.get_notes())
             )
+            completed_command = running_command.copy(update=update)
```
SyntaxColoring: No preference from me, but you could write this like this if you wanted to avoid the `update.update` thing.

```suggestion
            update = {
                "result": result,
                "status": CommandStatus.SUCCEEDED,
                "completedAt": self._model_utils.get_timestamp(),
                **_append_notes_if_notes(running_command, note_tracker.get_notes())
            }
            completed_command = running_command.copy(update=update)
```

SyntaxColoring (approved):
Looks good, thanks.

One thing in the tests that might matter, and various small suggestions.

api/src/opentrons/protocol_engine/execution/command_executor.py
sfoster1: yeah that's a good idea, will reduce the diff

sfoster1: @SyntaxColoring mind if we put this off actually?  the way those tests are currently implemented, we'd have to change a bunch of stuff to get access to the test command constructor arguments

SyntaxColoring: Sure, NP.

SyntaxColoring (approved):
LGTM if tests and linting are passing.

PR #14654, opened 2024-03-13
fix(protocol-designer): fix conditional rendering of contents in edit…

jerader:
…Labware

closes AUTH-108

# Overview

I don't quite understand why this bug was happening after updating react DnD. Basically, if you add another labware on to the deck, the last labware added is unable to drag/drop until another labware is dragged (unless the labware was a duplicate or a tiprack). This is because a duplicate and tiprack do not render the `NameThisLabware` component. The issue is with conditionally rendering the `contents`. If the component returned `NameThisLabware`, it never updated to the correct contents unless another labware was moved. The way i have it now works, it isn't quite the normal behavior since you can now drag a labware around before you give it a nickname but i don't think that really matters.

# Test Plan

create a flex or ot-2 protocol and add a labware. make sure the labware can drag/drop to a different slot.

# Changelog

- change the component to only return contents instead of the option of early returning the `NameThisLabware` components

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/DeckSetup/LabwareOverlays/EditLabware.tsx
+++ b/protocol-designer/src/components/DeckSetup/LabwareOverlays/EditLabware.tsx
@@ -71,11 +71,8 @@ export const EditLabware = (props: Props): JSX.Element | null => {
           dispatch(moveDeckItem(draggedLabware.slot, labwareOnDeck.slot))
         }
       },
-
-      hover: (item: DroppedItem, monitor: DropTargetMonitor) => {
-        if (monitor.canDrop()) {
-          setHoveredLabware(labwareOnDeck)
-        }
+      hover: () => {
+        setHoveredLabware(labwareOnDeck)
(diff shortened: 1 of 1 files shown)


protocol-designer/src/components/DeckSetup/LabwareOverlays/EditLabware.tsx
```diff
@@ -71,11 +71,8 @@ export const EditLabware = (props: Props): JSX.Element | null => {
           dispatch(moveDeckItem(draggedLabware.slot, labwareOnDeck.slot))
         }
       },
-
-      hover: (item: DroppedItem, monitor: DropTargetMonitor) => {
-        if (monitor.canDrop()) {
```
jerader, line 76: `setHoveredLabware` was never being called lol

shlokamin (approved):
thanks for fixing!

PR #14657, opened 2024-03-14
feat(app): add Parameters tab to odd protocolDetails

jerader:
closes AUTH-113, AUTH-124, AUTH-115, AUTH-114

# Overview

Creates the parameters tab in the ProtocolDetails section of the ODD also modify the empty section component for parameters

# Test Plan

go to protocol details on the odd and see that the parameters tab exists. clicking on the tab should display the empty parameters component. Observe the Parameters page and click on it to make sure the snackbar appears.

# Changelog

- add support to `EmptySection` for parameters
- create new `Parameters` component and test
- add `Parameters` as an option in the types to add it as a tab
- add snackbar to the parameters component 

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/pages/ProtocolDetails/Parameters.tsx
+++ b/app/src/pages/ProtocolDetails/Parameters.tsx
@@ -10,6 +10,7 @@ import {
   WRAP,
 } from '@opentrons/components'
 import { StyledText } from '../../atoms/text'
+import { useToaster } from '../../organisms/ToasterOven'
 import { useRequiredProtocolHardware } from '../Protocols/hooks'
 import { EmptySection } from './EmptySection'
 
@@ -37,12 +38,12 @@ const TableDatum = styled('td')`
   white-space: break-spaces;
   text-overflow: ${WRAP};
   &:first-child {
-    border-top-left-radius: ${BORDERS.borderRadiusSize4};
-    border-bottom-left-radius: ${BORDERS.borderRadiusSize4};
+    border-top-left-radius: ${BORDERS.borderRadius4};
+    border-bottom-left-radius: ${BORDERS.borderRadius4};
   }
   &:last-child {
-    border-top-right-radius: ${BORDERS.borderRadiusSize4};
-    border-bottom-right-radius: ${BORDERS.borderRadiusSize4};
+    border-top-right-radius: ${BORDERS.borderRadius4};
+    border-bottom-right-radius: ${BORDERS.borderRadius4};
   }
 `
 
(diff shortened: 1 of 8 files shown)


app/src/pages/ProtocolDetails/index.tsx
```diff
@@ -256,6 +258,9 @@ const ProtocolSectionContent = ({
         />
       )
       break
+    case 'Parameters':
```
jerader, line 275: not sure an easy way to put this behind the ff? 🤔

shlokamin: the new option in `protocolSectionTabOptions` should be behind a FF. you can either make a new constant or conditionally spread it in to the list. but as long as you do that you dont need to put stuff in `ProtocolSectionContent` cuz it wont be possible to select the params tab

app/src/pages/ProtocolDetails/Parameters.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const getRange = (parameter: RunTimeParameters): string => {
+    const { type } = parameter
+    const min = 'min' in parameter ? parameter.min : 0
+    const max = 'max' in parameter ? parameter.max : 0
+    const numChoices = 'choices' in parameter ? parameter.choices.length : 0
+    let range: string | null = null
+    if (numChoices === 2 && 'choices' in parameter) {
+      range = `${parameter.choices[0].displayName}, ${parameter.choices[1].displayName}`
+    }
```
koji, line 69: Oh, you are right. we just need to handle 2 choices this can be better than what I did.

app/src/pages/ProtocolDetails/Parameters.tsx
```diff
@@ -0,0 +1,160 @@
...
+            return choice.displayName
+          }
+        }
+        break
+    }
+    return ''
+  }
+
+  return runTimeParameters.length > 0 ? (
+    <Table onClick={makeSnack} data-testid="Parameters_table">
```
koji, line 113: I thought this was fine when we talked about this yesterday's meeting.
But now I think this might show the snack bar unexpectedly.

Right now I'm having the issue of ssh to a dev kit and flexs so not 100% sure but I guess this implementation may show the snack bar when scrolling the lists?

If that doesn't happen, this implementation is okay.

koji (approved):
If the scroll doesn't cause the issue I mentioned, the changes make sense to me.

app/src/pages/ProtocolDetails/Parameters.tsx
jerader: hmm i might merge this in to not block you but i think you're right that this will cause an issue, i'll post in slack about it. Maybe we need it to appear during a long press

PR #14661, opened 2024-03-14
refactor(app,components): borderRadius4 override

sfoster1:
Border radius 4.  Don't storybook because the pipette render is broken. I think this is it for components from figma.

Closes EXEC-331


mjhuff:
Definitely good for us not to hardcode these, thanks for changing that! The ticket is focused on changing borderRadius values not caught by then mass "search and replace" PR that are not currently 4px to 4px (with the shiny new constant). 

Figma:
https://www.figma.com/file/bJxzgdG6IUVuN2eHZs8dBI/Design-Variables-Playground?type=design&node-id=209-9667&mode=design&t=gBA432JSuz8c8Jpb-0

For example: The LocationConflictModal has a couple boxes that have borderRadius4 (the grey boxes - you can see this if you click it in figma). We need to either:

- Add a borderRadius4 if there isn't a border radius.
- Change the border radius from whatever it was to borderRadius4. 

I did a few of these already. I've pushed them up as examples if you'd like to take the rest of BR4!

mjhuff:
Alright, looks good now! 🚢

mjhuff (approved):

PR #14662, opened 2024-03-14
 refactor: use Public Sans and Reddit Mono on PAPI docs site

ecormany:
# Overview

Updating the Python API docs to use our design system proportional and fixed-width fonts.

# Test Plan

- public [sansbox](http://sandbox.docs.opentrons.com/docs-public-sans/v2/)
- tested on macOS Safari and Chrome, on computers with and without the fonts installed locally

# Changelog

- Open Sans -> Public Sans
- using Reddit Mono for monospace
- other minor CSS tweaks

# Review requests

is it pretty?

should we also get rid of Akko H1s?

is there a way to get rid of the 1px of white space between the light grey background box of `<a><code>` and the dotted border below it? i've tried every CSS kludge i know of.
Screenshot 2024-08-19 at 1 37 30 PM

also need testing assistance on

- [ ] Windows Chrome
- [ ] Windows Edge
- [ ] Linux browsers

# Risk assessment

low

### DIFF
--- a/api/docs/static/override_sphinx.css
+++ b/api/docs/static/override_sphinx.css
@@ -1,11 +1,11 @@
-@import url('https://fonts.googleapis.com/css?family=Open+Sans:300,400,400i,600,700');
+@import url('https://fonts.googleapis.com/css?family=Public+Sans:300,400,400i,600,700');
 
 /* OT NAV */
 
 body {
     padding: 0;
     margin: 0;
-    font-family: "Open Sans", "sans-serif";
+    font-family: "Public Sans", "sans-serif";
 }
 
 .highlight-none, .mi, .literal {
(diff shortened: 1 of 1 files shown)


api/docs/static/override_sphinx.css
```diff
@@ -35,7 +36,18 @@ div.document [id] {
 
 div.body p {
     line-height: 20pt;
-    font-family: "Open Sans", "sans-serif";
+    font-family: "Public Sans", "sans-serif";
+}
+
+pre, tt, code {
+    font-size: 0.9em;
```
koji, line 43: need to use the parent element as the basis?
if not, recommend you use `rem` instead `em`

api/docs/static/override_sphinx.css
```diff
@@ -35,7 +36,18 @@ div.document [id] {
...
+    font-family: "Reddit Mono", "Consolas", "Lucida Console", monospace;
+}
+
+/* classes for API Reference docstring signatures */
+.sig, .sig-name, code.descname, .sig-prename, .optional, .sig-paren {
+    font-size: 1em;
```
koji, line 49: same question

koji (approved):
nice!
font was changed from `Open Sans` to `Public Sans`

ecormany:
Going ahead and merging this so I can continue on bigger style changes (new top and bottom nav). If there are any issues caused by this PR, I can address it as part of that or in a separate followup.

PR #14664, opened 2024-03-14
refactor(api): Remove failed command state

SyntaxColoring:
# Overview

This goes towards EXEC-301.

This removes this piece of internal Protocol Engine state:

https://github.com/Opentrons/opentrons/blob/7689521103796ee548cea6503ad1f6b6503c0cd9/api/src/opentrons/protocol_engine/state/commands.py#L155-L156

It's only used as a pagination helper for robot-server, and it's sort of semantically problematic in a world where a run has many failed commands and none of them are necessarily fatal.

# Test Plan

To do.

# Changelog

To do.

# Review requests

To do.

# Risk assessment

To do.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -451,6 +451,7 @@ class CommandView(HasState[CommandState]):
         cursor: Optional[int],
         length: int,
     ) -> CommandSlice:
+        # todo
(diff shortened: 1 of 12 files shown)


SyntaxColoring:
Closing for now: this tempts us down a robot-server pagination rabbithole that doesn't seem productive right now. I'll go with a different approach that doesn't involve removing this piece of state.

PR #14668, opened 2024-03-15
feat(api): runtime parameters API for adding and using default parameters in protocols

jbleon95:
# Overview

Addresses AUTH-67 and AUTH-68. 

This PR implements both the current proposed runtime parameter API for adding and defining parameters to your protocol, as well as the means to do so. This only allows usage of the defaults in the protocol, you cannot set it to other values and they are not reported in analysis.

This is accomplished by injecting a new `ParameterContext` into an `add_parameters` function that is defined in the protocol, similar to how we inject `ProtocolContext` into the `run` function. This contains the user facing API for adding parameters to your protocol. 

Once all parameters are parsed and validated (if they fail this they will raise a user friendly error with the correct line number), they are injected into the protocol context under a new `context.params` object. This will have all the parameters as dynamically set attributes of this class using the set variable name (and as `property`s to enforce them being read only) that can then be used in the protocol as needed.

For parameter definitions, both the constraints and the default values are verified to ensure the protocol will not analyze unless parameters have been defined properly.

# Test Plan

This protocol is able to analyze on the app and robot, as well as run without any errors.

```
metadata = {
    'protocolName': 'RTP Test 2 - Default only, all types',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.17"
}


    parameters.add_int(
        display_name="Sample count",
        variable_name="sample_count",
        default=6,
        minimum=1,
        maximum=12,
    )
    parameters.add_float(
        variable_name="volume",
        default=20.0,
        choices=[
            {"display_name": "Low Volume (10.0µL)", "value": 10.0},
            {"display_name": "Medium Volume (20.0µL)", "value": 20.0},
            {"display_name": "High Volume (50.0µL)", "value": 50.0},
        ],
        unit="µL"
    )
    parameters.add_bool(
        display_name="Dry Run",
        variable_name="dry_run",
        default=False,
    )
    parameters.add_str(
        display_name="Pipette Name",
        variable_name="pipette",
        choices=[
        ],
        default="flex_1channel_50",
    )



def run(context):

        pipette.pick_up_tip()

        pipette.drop_tip()
```

# Changelog

- Added `ParameterContext` class for use in `add_parameters` special protocol function
- `ProtocolContext` now has a `params` attribute that holds all the parameters as dynamically set attributes
- Added `ParameterDefinition` class to hold all parameter definition data that will be reported in analysis and passed to the robot server
- Added parameter validation when creating definitions

# Review requests

General code structure feedback

# Risk assessment

Low/medium, this is a brand new API but it is completely encapsulated from the run context. The most dangerous part of this is the dynamically set attributes but I believe we are doing enough checks on variable name and checking existing attributes to prevent the user from doing anything dangerous.


api/src/opentrons/protocol_api/_parameters.py
```diff
@@ -0,0 +1,31 @@
...
+    def _getparam(self, variable_name: str) -> Any:
+        return getattr(self, f"_{variable_name}")
+
+    def _initialize_parameter(self, variable_name: str, value: AllowedTypes) -> None:
+        if not hasattr(self, variable_name) and not hasattr(self, f"_{variable_name}"):
+            setattr(self, f"_{variable_name}", value)
+            prop = property(
+                fget=lambda s, v=variable_name: Parameters._getparam(s, v)  # type: ignore[misc]
+            )
+            setattr(Parameters, variable_name, prop)
```
SyntaxColoring: A couple of suggestions for simplifying this.

-----

First, instead of the user-facing API being like this:

```python
def run(context):
    sample_count = context.params.sample_count
```

Have you considered making it this?

```python
def run(context):
    sample_count = context.params["sample_count"]
```

* This gives us namespacing in case we ever need to add more methods like `params.get_all()`.
* This matches parameter usage more closely to parameter declaration. In the declaration, the protocol author wrote `variable_name="sample_count"`, with a string literal, so they pass the same string literal here.
* This is arguably more straightforward and easier to explain.

-----

Or, taking the `context.params.sample_count` syntax for granted, an implementation-level suggestion.

Right now, when we initialize the parameter `foo="bar"`, we:

1. Set a private attribute `self._foo = "bar"`
2. Set a public attribute `self.foo` to a read-only property that returns `self._foo`
3. Store `"foo": "bar"` in the `self._values` dict

Does this get easier if you treat `self._values` as the single source of truth, and override `__getattr__()` (and possibly `__setattr__()`) to delegate to it? In other words, instead of dynamically adding attributes, implement dynamic attribute lookup. Then you don't need the `f"_{variable_name}"` magic.

api/src/opentrons/protocol_api/parameter_context.py
SyntaxColoring, line 1: I would prefix this file name with an underscore unless you want `from opentrons.protocol_api import parameter_context` to be a thing that protocol authors are allowed to do.

api/src/opentrons/protocols/parameters/types.py
```diff
@@ -0,0 +1,25 @@
+from typing import TypeVar, Union, TypedDict
+
+
+AllowedTypes = Union[str, int, float, bool]
+
+ParamType = TypeVar("ParamType", bound=AllowedTypes)
+
+
+class ParameterChoices(TypedDict):
```
SyntaxColoring: Naming nitpick: should this be a singular `ParameterChoice`?

api/src/opentrons/protocols/parameters/validation.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def ensure_variable_name(variable_name: str) -> str:
+    """Validate variable name is a valid python variable name."""
+    if not variable_name.isidentifier():
+        raise ParameterNameError(
+            "Variable name must only contain alphanumeric characters, underscores, and cannot start with a digit."
+        )
+    if keyword.iskeyword(variable_name):
+        raise ParameterNameError("Variable name cannot be a reserved Python keyword.")
```
SyntaxColoring, line 34: I'm not sure `iskeyword` is the right thing to use here; it doesn't seem necessary or sufficient for preventing "bad" variable names.

Some slightly contrived examples:

* `def` is a keyword, ~~but there's no technical reason someone shouldn't be able to name something `params.def`.~~ It might even be a reasonable thing to name a parameter for a labware definition load name.
* Meanwhile, things like `__class__` and `__eq__` are potentially scarier, but they're *not* keywords, so they'd pass this check.

This is kind of reaffirming to me that the user-facing syntax should be `params["foo"]`. It just seems inherently safer than `params.foo`. It's your call, but if we stick with `params.foo`, we should put some thought into a threat model.

api/src/opentrons/protocols/parameters/validation.py
```diff
@@ -0,0 +1,141 @@
...
+    ParameterChoices,
+    ParameterNameError,
+    ParameterValueError,
+    ParameterDefinitionError,
+)
+
+
+UNIT_MAX_LEN = 10
+DISPLAY_NAME_MAX_LEN = 30
+DESCRIPTION_MAX_LEN = 100
```
SyntaxColoring, line 15: Is this to protect against layout problems in the Opentrons App and ODD?

If so: It's up to you, but I think you'll have better results if you leave it to the frontend to truncate or ellipsize overly long names. Here's a contrived display name of `len()==15` that will certainly not fit in a space designed to fit "up to 30 characters."

> ﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽﷽

Or, if this is just to protect against extreme resource usage, I would set the limits to something like 1000.

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -215,6 +217,10 @@ def bundled_data(self) -> Dict[str, bytes]:
         """
         return self._bundled_data
 
+    @property
+    def params(self) -> Parameters:
+        return self._params
```
SyntaxColoring: Might want a `@requires_version` decorator or a ticket for it, before we forget.

SyntaxColoring (approved):
A few suggestions, but LGTM!

api/src/opentrons/protocol_api/parameter_context.py
jbleon95: The would be no code execution reason to import the module, but since its the most public facing part of this (as far as the API) I'm alright keeping it without an underscore. Plus it would allow as user to do this pattern of typing the `add_parameters` function.
```
from opentrons.protocol_api.parameter_context import ParameterContext

def add_parameters(parameters: ParameterContext):
    ...
```

api/src/opentrons/protocols/parameters/validation.py
jbleon95: `params.def`, using the attribute method of parameter usage, cannot actually be used in code even though it can be set and will raise a `SyntaxError` when called.

As for the second set of examples, those will be caught by the `hasattr` check in the `Parameters` class, so there is no chance of that being overwritten (and I think `setattr` wouldn't even let you do that in the first place)

api/src/opentrons/protocols/parameters/validation.py
jbleon95: We've made a deliberate choice not to truncate or ellipsize long names on the frontend to maximize the user experience for lab operators and force protocol authors to choose their names with care. But you bring up a good point with this and I'll bring it up with app&ui on the protocol authorship team.

api/src/opentrons/protocol_api/_parameters.py
jbleon95: For now I'm going to leave the syntax the same, but I will take your suggestion to simplify things by using `self._values` as the source of truth for the `fget` function of the property.

api/src/opentrons/protocol_api/parameter_context.py
SyntaxColoring: I would solve the type annotation problem by re-exporting the class so it’s

`from opentrons.protocol_api import ParameterContext`

api/src/opentrons/protocol_api/_parameters.py
SyntaxColoring: I think this will be moot if you use `self._values` as the source of truth, but I've also just realized that in `setattr(Parameters, variable_name, prop)`, it looks like that's mutating the `Parameters` class, which will leak between protocols?

sanni-t (approved):
Looks great! I like the dynamic property creation! @SyntaxColoring has some good points that it's not as easily safe as having property getters but we'll do thorough testing and adding safeguards so that there's no scope for misuse. 

I think we definitely should put the `context.params` property behind a requires decorator. Are you planning on doing that in a follow-up?

api/src/opentrons/protocol_api/protocol_context.py
sanni-t: We can create a feature flag for now and put it behind that. Replace it with API level check before release. The feature flag will be useful for other backend things too.

PR #14670, opened 2024-03-15
refactor(api-client): Delete unused code that was supporting actions on maintenance runs

SyntaxColoring:
# Overview

Remove some dead code and redundant type definitions that I noticed while working on EXEC-300.

# Test Plan

None needed.

# Changelog

* Delete unused HTTP API client code that would have hit a `/maintenance_runs/{id}/actions` endpoint. That endpoint doesn't exist. As far as I can tell, maintenance runs, unlike regular runs, do not have actions.
    * Even if maintenance run actions were to exist in the future, I think we would want to find a way to combine them with the `/runs/{id}/actions` code, to avoid having to maintain two parallel trees that are mostly identical.
* Delete types and constants for maintenance run actions. Instead of those, reuse the types for regular run actions, where appropriate.

# Review requests

Is this a good idea?

# Risk assessment

Low.


mjhuff (approved):
Makes sense and LGTM!

PR #14673, opened 2024-03-15
refactor(hardware): give options for sensor data output during probe

caila-marashaj:
# Overview

These are some small changes that help accommodate liquid-level detection testing in the hardware layer. 
EXEC-337
# Changelog

- add `AddSensorLinearMoveRequest`, `SendAccumulatedPressureDataRequest` CAN messages
- add `sensor_report` Move Condition to allow us to request that the firmware accumulate data in the necessary manner for hardware testing.
- modify `tool_sensors.py` to allow its users to choose between reporting sensor data to the CAN bus only, writing sensor data to a csv file, or not reporting data at all during a pass. 

# Review requests

The CAN messages mentioned above will probably be temporary additions to `edge`, but I think merging them and potentially later on removing them may be the best way to go about this. Let me know if not.


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1341,7 +1342,8 @@ async def liquid_probe(
             plunger_speed,
             mount_speed,
             threshold_pascals,
-            log_pressure,
+            OutputOptions.none,
```
sfoster1: The sheer number of arguments here was OK when everything was passed through with its semantic variable name, but I think once we're passing in stuff like two variants of `None` it's time to switch this call to be by keyword arg

hardware/opentrons_hardware/firmware_bindings/messages/message_definitions.py
```diff
@@ -977,3 +990,16 @@ class GetHepaUVStateResponse(BaseMessage):
     message_id: Literal[
         MessageId.get_hepa_uv_state_response
     ] = MessageId.get_hepa_uv_state_response
+
+
+@dataclass
+class SendAccumulatedPressureDataRequest(BaseMessage):
```
sfoster1, line 1007: what response message does this use?

hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -60,6 +74,70 @@ def _build_pass_step(
...
+            ax: float64(speed[ax] * copysign(1.0, distance[ax])) for ax in movers
+        },
+        acceleration={},
+        # use any node present to calculate duration of the move, assuming the durations
+        #   will be the same
+        duration=float64(abs(distance[movers[0]] / speed[movers[0]])),
+        present_nodes=pipette_nodes,
+        stop_condition=MoveStopCondition.sensor_report,
+    )
+    print(f"Move group for probe {move_group}")
```
sfoster1: accidentally checked in?

hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -60,6 +74,70 @@ def _build_pass_step(
...
+    for node in pipette_nodes:
+        move_group[node] = pipette_move[node]
+    print(f"Move group for probe {move_group}")
```
sfoster1: accidentally checked in?

sfoster1 (changes requested):
Couple small things to change like removing prints and making a function invocation kwargs. I think I would have done the split of the CSV stuff slightly differently but I think this way will work and I'm interested to see how it turns out.

hardware/opentrons_hardware/firmware_bindings/messages/message_definitions.py
caila-marashaj: this prompts all of the `ReadFromSensorResponse` messages to be dumped from the array the firmware accumulates during liquid probe

ryanthecoder: It sends the same ReadFromSensorResponse as we did before

hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -60,6 +74,68 @@ def _build_pass_step(
...
+        duration=float64(abs(distance[movers[0]] / speed[movers[0]])),
+        present_nodes=pipette_nodes,
+        stop_condition=MoveStopCondition.sensor_report,
+    )
+    for node in pipette_nodes:
+        move_group[node] = pipette_move[node]
+    return move_group
+
+
+async def run_pass_output_to_csv(
```
ryanthecoder: We still need a way to gate the behavior here to see if we want to use the stream can messages to csv behavior we had before, or the new buffer-on-pipette then write we need. 
the way we run the move group runner is different with the custom firmware. 

the new custom firmware needs this logic

       messenger.add_listener(sensor_capturer, None)

        async with sensor_capturer:
            print("starting move group runner")
            positions = await sensor_runner.run(can_messenger=messenger)
            await messenger.send(
                node_id=tool,
                message=SendAccumulatedPressureDataRequest(
                    payload=SendAccumulatedPressureDataPayload(
                        sensor_id=SensorIdField(sensor_id)
                    )
                ),
            )
            await asyncio.sleep(10)
            messenger.remove_listener(sensor_capturer)
        await messenger.send(
            node_id=tool,
            message=BindSensorOutputRequest(
                payload=BindSensorOutputRequestPayload(
                    sensor=SensorTypeField(SensorType.pressure),
                    sensor_id=SensorIdField(sensor_id),
                    binding=SensorOutputBindingField(SensorOutputBinding.none),
                )
            ),
        )

and the normal firmware needs what you have here

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1334,17 +1335,18 @@ async def liquid_probe(
...
-            num_baseline_reads,
-            sensor_id_for_instrument(probe),
+            messenger=self._messenger,
+            tool=tool,
+            head_node=head_node,
+            max_z_distance=max_z_distance,
+            plunger_speed=plunger_speed,
+            mount_speed=mount_speed,
+            threshold_pascals=threshold_pascals,
+            output_format=OutputOptions.none,
```
ryanthecoder: we need to hook this up one more layer to ot3_api,
and if the Output Options chosen is the one that needs the custom firm we should assert here that 
 `self._subsystem_manager.device_info[SubSystem.of_mount(mount)].revision.tertiary == "1"`

hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -60,6 +74,68 @@ def _build_pass_step(
...
+        distance={ax: float64(abs(distance[ax])) for ax in movers},
+        velocity={
+            ax: float64(speed[ax] * copysign(1.0, distance[ax])) for ax in movers
+        },
+        acceleration={},
+        # use any node present to calculate duration of the move, assuming the durations
+        #   will be the same
+        duration=float64(abs(distance[movers[0]] / speed[movers[0]])),
+        present_nodes=pipette_nodes,
+        stop_condition=MoveStopCondition.sensor_report,
```
ryanthecoder, line 107: This also needs to be gated on using the custom firmware or not, the normal firmware cannot process this stop condition
And we'll need to grab the changes from move_group_runner to handle this stop condition too

ryanthecoder (approved):

sfoster1 (approved):
Okay, looks good to me!

PR #14674, opened 2024-03-15
feat(robot-server,api): Add the skeleton for a new `complete-recovery` run action

SyntaxColoring:
# Overview

Closes EXEC-300.

# Changelog

* Add a new action type for `POST /runs/{id}/actions`: `complete-recovery`.
* Connect it to `ProtocolEngine` and update unit tests along the way.
* For now, have `ProtocolEngine` raise a `NotImplementedError` whenever it encounters it. The actual implementation will be done in future PRs.
* Update TypeScript bindings.

# Test Plan

I've tested with Postman that I can post a `complete-recovery` action to a run, and it raises the expected `NotImplementedError`.

# Review requests

Do we like the name `complete-recovery` for this?

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -666,7 +666,7 @@ class CommandView(HasState[CommandState]):
         """Get whether engine is in a terminal state."""
         return self._state.run_result is not None
 
-    def validate_action_allowed(
+    def validate_action_allowed(  # noqa: C901
         self,
         action: Union[
             PlayAction,
@@ -694,6 +694,17 @@ class CommandView(HasState[CommandState]):
             SetupCommandNotAllowedError: The engine is running, so a setup command
                 may not be added.
         """
+        if self.get_status() == EngineStatus.AWAITING_RECOVERY:
+            # While we're developing error recovery, we'll conservatively disallow
+            # all actions, to avoid putting the engine in weird undefined states.
+            # We'll allow specific actions here as we flesh things out and add support
+            # for them.
+            raise NotImplementedError()
+
+        if isinstance(action, ResumeFromRecoveryAction):
+            # https://opentrons.atlassian.net/browse/EXEC-301
+            raise NotImplementedError()
+
         if self._state.run_result is not None:
             raise RunStoppedError("The run has already stopped.")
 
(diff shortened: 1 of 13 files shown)


api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -701,7 +710,10 @@ def validate_action_allowed(
                     "Setup commands are not allowed after run has started."
                 )
 
-        elif self.get_status() == EngineStatus.AWAITING_RECOVERY:
+        elif (
+            isinstance(action, CompleteRecoveryAction)
+            or self.get_status() == EngineStatus.AWAITING_RECOVERY
```
DerekMaggio: Why is this `or` instead of `and`?

SyntaxColoring: Currently, doing a `CompleteRecoveryAction` is not implemented, and, separately, doing anything while the engine is `AWAITING_RECOVERY` is also not implemented. I want them both to raise the same `NotImplementedError`.

SyntaxColoring: Although as I'm reading this now, this is not actually implemented correctly. For example, I wanted `PlayAction` to raise a `NotImplementedError` if it was issued while the engine status was `AWAITING_RECOVERY`, but it does not. I'll fix and clarify this.

sfoster1 (approved):
modulo the correction on the state hnadling that Derek pointed out, looks good. I'm interested in your reasoning about two things:
1. why call this `complete-recovery` as opposed to some `resume` thing? To me it carries the implication that the engine will be _doing something_ to complete the recover, as opposed to relying on the client to have done something and then naively resuming
2. in fact now that i'm looking at it, why not have this be `PlayAction`? That moves us towards a system where we have fewer, more common keywords that do different things based on state - a state machine where logic is biased to "what state am I in".

Like I said, it looks good to merge either way, but I'm interested in hearing your thoughts.

SyntaxColoring:
> why call this `complete-recovery` as opposed to some `resume` thing?

No reasoning. `resume-from-recovery` does sound clearer to me. I'll switch to that unless anyone has a better suggestion.

> in fact now that i'm looking at it, why not have this be `PlayAction`?

A couple of loose reasons:

* Depending on how pause-on-open-door shakes out, we might need `PlayAction`  to resume the recovery sequence from a door pause without *exiting* the recovery sequence.
* We don't yet have a server-enforced way of preventing clients from stepping on each other's toes, like any kind of remote lock or HTTP conditional request. Given that, having HTTP actions that are only valid in certain contexts is a cheap (but admittedly half-assed) way to decrease the odds of two clients accidentally interacting in a dangerous way.

PR #14676, opened 2024-03-15
adding ip address and google sheet path as arguments

rclarke0:
# Overview

Added Google Sheet Path as Arguments and IP import from storage_directory

# Test Plan

Tested with different google sheet to ensure that new names work.

# Changelog

**abr_run_logs.py**
In get_all_run_logs function added
` try:
        sys.path.insert(0, storage_directory)
        import IPs  # type: ignore[import]

        ip_address_list = IPs.ip_address_list
    except ImportError:
        raise ImportError("Make sure Ip address file is saved in storage directory.")`
**abr_read_logs.py**
changed credentials file to be called credentials.json instead of abr.json
added google sheet path and tab as arguments.
`parser.add_argument(
        "file_name",
        metavar="FILE_NAME",
        type=str,
        nargs=1,
        help="Name of google sheet and local csv to save data to.",
    )
    parser.add_argument(
        "google_sheet_tab_number",
        metavar="GOOGLE_SHEET_TAB_NUMBER",
        type=int,
        nargs=1,
        help="Google sheet tab number.",
    )
    args = parser.parse_args()
    storage_directory = args.storage_directory[0]
    file_name = args.file_name[0]
    tab_number = args.google_sheet_tab_number[0]`

changed local file name to be same as google sheet name
`def create_abr_data_sheet(storage_directory: str, file_name: str) -> str:
    """Creates csv file to log ABR data."""
    file_name_csv = file_name + ".csv"
    sheet_location = os.path.join(storage_directory, file_name_csv)`
# Review requests

Determine if there is any other lines in this code that make it too abr specific

# Risk assessment


hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
```diff
@@ -1,11 +1,12 @@
 """ABR Run Log Pull."""
-from .abr_robots import ABR_IPS
+# from .abr_robots import ABR_IPS
```
sfoster1: delete rather than commenting

hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
```diff
@@ -112,8 +113,16 @@ def get_all_run_logs(storage_directory: str) -> None:
     Read each robot's list of unique run log IDs and compare them to all IDs in storage.
     Any ID that is not in storage, download the run log and put it in storage.
     """
+    try:
+        sys.path.insert(0, storage_directory)
+        import IPs  # type: ignore[import]
```
sfoster1: it's usually not a great idea to do these kind of dynamic imports; it makes it tough to evaluate the code and it's a confusing error to handle. What if you saved them in a json file or something instead?

sfoster1 (changes requested):
Let's make the IP addresses a json file or something - using piython files for data storage is not a great idea because errors are tough to parse and it makes the program invalid if it's not present.

Other than that looks great!

hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
rclarke0: changed this to be read in json file format.

hardware-testing/hardware_testing/abr_tools/abr_run_logs.py
rclarke0: deleted.

sfoster1 (approved):
Looks good! I think probably a good future improvement is to have the ips and the sheets credentials files be defaults that can be overridden by optional command line arguments using `parser.add_argument('-c', '--credentials', type=argparse.FileType('r'), default=os.path.join(storage_directory, 'credentials.json'))` and similar but that's a long way in the future I think.

PR #14677, opened 2024-03-15
docs(api): runtime parameters in Python API 2.18

ecormany:
# Overview

Full documentation for setting up and using runtime parameters in Python protocols.

Closes RTC-416, RTC-428, RTC-429, RTC-430, RTC-451

# Test Plan

- Check content and formatting of all pages in the [sandbox](http://sandbox.docs.opentrons.com/docs-rtp/v2/runtime_parameters.html)
- Simulate all [code from use case pages](https://gist.github.com/ecormany/9f4d871851ec4baceb9fae177663dd00).

# Changelog

- New landing page for Runtime Parameters
- 6 new subpages
- Unhide `ProtocolContext.params` in API Reference. ~Note, this still needs a docstring, either here or in a separate PR.~ Added [docstring](http://sandbox.docs.opentrons.com/docs-rtp/v2/new_protocol_api.html#opentrons.protocol_api.ProtocolContext.params).

# Review requests

Read some or all of it, try the code, tell me what can be better.

# Risk assessment

None


api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,96 @@
...
+        variable_name="pipette",
+        display_name="Pipette Name",
+        choices=[
+            {"display_name": "1-Channel 50µL", "value": "flex_1channel_50"},
+            {"display_name": "8-Channel 50µL", "value": "flex_8channel_50"},
+        ],
+        default="flex_1channel_50",
+        description="What pipette to use during the protocol.",
+    )
+
```
y3rsh, line 178: ## Exciting!  Coming together in a nice shape 📦 

Will we not provide all possible shapes here for brevity or better to include them all?

- bool ✅
- int min/max ✅
- int choice ❌ 
- float min/max ❌ 
- float choice✅
- str ✅

ecormany: I think we'll include them all. This is just what I had pre-written code for.

api/docs/v2/parameters/choosing.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Consider our earlier example of parameterizing serial dilution. Each of the three numerical parameters have logical upper and lower bounds, which we need to enforce to get sensible results.
+
+- *Number of dilutions* must be between 0 and 11 on a 96-well plate. And it may make sense to require at least 1 dilution.
+- *Dilution factor* is a ratio, which we can express as a decimal number that must be between 0 and 1.
+- *Number of rows* must be between 1 and 8 on a 96-well plate.
+
+What if you wanted to perform a dilution with 20 repetitions? It's possible with two 96-well plates, or with a 384-well plate. You could set the maximum for the number of dilutions to 24 and allow for these possibilities — either switching the plate type or loading an additional plate based on the provided value. 
+
+But what if the user wanted to do just 8 repetitions on a 384-well plate? That would require an additional parameter, an additional choice by the user, and additional logic in your protocol code. It's up to you as the protocol author to decide if adding more parameters adds unnecessary complexity. It may well be better to have two or three protocols that each work simply and efficiently.
```
jwwojak: First, wow! This whole `choosing.rst` file is great. It's clear, precise, a great piece of writing.

My one observation here is with the very last sentence. It seems to pull its punches. I'm reading all this and coming up to the end, expecting the message about simplicity to be hammered home. You still get a payoff, but the message about simplicity comes at the very end and feels softer by its positioning. "Simple" should come earlier in the sentence. Something like, "It may be better to work with two or three simple protocols than with a large multi-parameter protocol."

And if we do want to discourage potentially big and unwieldy protocols, make the end slightly more opinionated than "It may well be." For example, "Sometimes it's better to work with a few simple protocols rather than one that's long and complex."

Anyway, all of this is good as I noted. This very last sentence just kinda felt off.

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+*******************
+Defining Parameters
+*******************
+
+To use parameters, you need to define them in :ref:`a separate function <add-parameters>` within your protocol. Each parameter definition has two main purposes: to specify acceptable values, and to inform the protocol user what the parameter does.
+
+Depending on the :ref:`type of parameter <rtp-types>`, you'll need to specify some or all of the following.
+
+.. list-table::
```
jwwojak, line 13: Tables, yesssss.

api/docs/v2/parameters/style.rst
```diff
@@ -0,0 +1,137 @@
+:og:description: Style and usage guidance for parameters in Opentrons Python protocols.
+
+.. _rtp-style:
+
+*********************
+Parameter Style Guide
+*********************
+
+Defining parameters creates an important part of the user experience for the technicians who run your protocols. They rely on your parameter names and descriptions being clear, consistent, and ultimately easy to use.
```
jwwojak: I wonder if we could reconsider this intro, always a difficult thing to write.

Defining the parameters, a verb, isn't an experience for people running the protocol. They come after that action. They experience parameters as a noun, a thing that exists that they have to deal with.

I think what this is trying to get at is parameters with clear, simple names and definitions/descriptions are what's important to the technician's experience. Like a good news headline, the parameter name should convey information about what it does or is. A name shouldn't be vague or obscure purpose. A suggestion:

"Parameter names and definitions should be simple and reflect their function. Clarity helps improve the user experience for technicians who run your protocols. They rely on your parameter names and descriptions to better understand what your protocol, and the robot, are going to do."

Or something like that.

api/docs/v2/parameters/style.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Adopting the advice of this guide also makes your protocols more consistent with those in the `Opentrons Protocol Library <https://library.opentrons.com>`_, which can help others access and replicate your science. 
+
+General Guidance
+================
+
+**Parameter names are nouns.** Parameters should be discrete enough that you can describe them in a single word or short noun phrase. ``display_name`` is limited to 30 characters, and you can add more context in the description.
```
jwwojak, line 16: Could these **bold** sentences be be H3s? Might be nice if they're discoverable in the left nav/TOC. May not work given how many there are though.

api/docs/v2/parameters/style.rst
```diff
@@ -0,0 +1,137 @@
...
+        - ✅ How much to aspirate from each sample.
+        - ❌ Volume that the pipette will aspirate
+    * - Dilution factor
+      - 
+        - ✅ Each step uses this ratio of total liquid to original solution. Express the ratio as a decimal.
+        - ❌ total/diluent ratio for the process
+
+Not every parameter requires a description! For example, in a protocol that uses only one pipette, it would be difficult to explain a parameter named "Pipette type" without repeating yourself. In a protocol that offers parameters for two different pipettes, it may be useful to summarize what steps each pipette performs.
+    
+**Use sentence case for readability**. Sentence case means adding a capital letter to *only* the first word of the name and description. This gives your parameters a professional appearance. Keep proper names capitalized as they would be elsewhere in a sentence. For example:
```
jwwojak, line 55: From my suggestion above, if this were an H3, I'd revise to "Use sentence case" and drop the "for readability" part.

api/docs/v2/parameters/use_case_dry_run.rst
```diff
@@ -0,0 +1,127 @@
...
+Skipping Delays
+===============
+
+Many protocols have built-in delays, either for a module to work or to let a reaction happen passively. Lengthy delays just get in the way when verifying a protocol with a dry run. So wherever the protocol calls for a delay, we can check the value of ``protocol.params.dry_run`` and make the protocol behave accordingly.
+
+To start, let's consider a simple :py:meth:`.delay` command. We can wrap it in an ``if`` statement such that the delay will only execute when the run is *not* a dry run::
+
+    if protocol.params.dry_run == False:
+        protocol.delay(minutes=5)
+
```
jwwojak: Now this is very slick, super-cool. Most impressive.

api/docs/v2/parameters/use_case_dry_run.rst
```diff
@@ -0,0 +1,127 @@
+:og:description: How to set up and use a dry run parameter in an Opentrons Python protocol.
```
jwwojak, line 1: Holy hell. This whole thing is friggin' amazing. I can see how it would be extremely useful and make code and dry runs efficient. Wowzers.

api/docs/v2/parameters/use_case_sample_count.rst
```diff
@@ -0,0 +1,273 @@
...
+                {"display_name": "24", "value": 24},
+                {"display_name": "32", "value": 32},
+            ]
+        )
+
+All of the possible values are multiples of 8, because the protocol will use an 8-channel pipette to process an entire column of samples at once. Considering how 8-channel pipettes access wells, it may be more useful to operate with a *column count* in code. We can set a ``column_count`` very early in the ``run`` function by accessing the value of ``params.sample_count`` and dividing it by 8::
+
+    def run(protocol):
+
+        column_count = protocol.params.sample_count // 8
```
jwwojak: There's an assumption here that end users know how to use math operators in Python. While you can quickly guess that this text "and dividing it by 8" followed by the code sample that shows `// 8` suggests that `//` is the division operator, that doesn't seem fair. It might be better to link to a resource that shows the arithmetic operators in Python or explain this in the paragraph.

api/docs/v2/parameters/use_case_sample_count.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Most examples below will use ``column_count``, rather than redoing (and retyping!) this calculation multiple times.
+
+Loading Tip Racks
+=================
+
+Tip racks come first in most protocols. To ensure that a protocol runs to completion, we need to load enough tip racks to avoid running out of tips.
+
+We could load as many tip racks as are needed for our maximum number of samples, but that would be suboptimal. Run setup is faster when the technician doesn't have to load extra items onto the deck. So it's best to examine the protocol's steps and determine how many racks are needed for each value of ``sample_count``.
```
jwwojak, line 55: Word choice: "suboptimal". There's no rule here, I just dislike the word. I'd have written this as, "We could load as many tip racks as required for our maximum number of samples, but that is inefficient."

jwwojak (commented):
Awesome additions to the API docs. I did leave some comments. Please take a look.

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+       - Maximum [10/TK] characters.
+
+
+
+.. _add-parameters:
+
+The ``add_parameters()`` Function
+=================================
+
+All parameter definitions are contained in a Python function, which must be named ``add_parameters`` and takes a single argument. You must define ``add_parameters`` before the ``run`` function that contains protocol commands.
```
jbleon95: It isn't strictly true that you need to define `add_parameters` before `run`, due to how we parse it and the fact that they're "independent" functions outside the context of our API, a protocol will analyze even if they are in the other order. But, I absolutely agree that it's best practice to have it before and I'm fine leaving it how it is.

ecormany: Good to know, but I agree it's good to be a little prescriptive about protocol structure.

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+        default=5.0,
+        choices=[
+            {"display_name": "Low (2.5 µL)", "value": 2.5},
+            {"display_name": "Medium (5 µL)", "value": 5.0},
+            {"display_name": "High (10 µL)", "value": 10.0},
+        ],
+        unit="µL"
+    )
+
+Remember, you can't mix types in a parameter. So values of ``2.5``, ``5.0``, and ``10.0`` are valid, but ``2.5``, ``5``, and ``10`` will raise an error.
```
jbleon95: The first contrary example is a float here, and also we do accept `int` values for floats since integers are a subset of floating points. This is not true the other way around (even for something like `1.0`). But again, though to a lesser degree then my previous comment, it is best practice to make them explicit floating points.

api/docs/v2/parameters/using_values.rst
```diff
@@ -0,0 +1,72 @@
...
+
+- ``add_bool`` with ``variable_name="dry_run"``
+- ``add_int`` with ``variable_name="sample_count"``
+- ``add_float`` with ``variable_name="volume"``
+
+Then ``params`` will gain three attributes: ``params.dry_run``, ``params.sample_count``, and ``params.volume``. You can use these attributes anywhere you want to access their values, including directly as arguments of methods.
+
+.. code-block::
+
+    if protocol.params.dry_run == False:
```
jbleon95: Python style should have this as either `if protocol.params.dry_run is False:` or `if not protocol.params.dry_run:`

ecormany: Oh good call. I prefer `is False` — more explicit and easier to understand.

`==` comparison for all other param types, yes?

api/docs/v2/parameters/choosing.rst
```diff
@@ -0,0 +1,53 @@
...
+1. **Add flexibility.** Accommodate changes from run to run or from lab to lab.
+2. **Work efficiently.** Don't burden run setup with too many choices or confusing options.
+3. **Avoid errors.** Ensure that every combination of parameters produces an analyzable, runnable protocol.
+
+Build on a Task
+===============
+
+Consider what scientific task is at the heart of your protocol, and build parameters that contribute to, rather than diverge from it.
+
+For example, it makes sense to add a parameter for number of samples to a protocol that uses a particular DNA prep reagent kit. But it wouldn't make sense to add a parameter for *which reagent kit* to use for DNA prep. That kind of parameter would affect so many aspects of the protocol that it would make more sense to maintain two protocols, one for each kit.
```
sanni-t:
```suggestion
For example, it makes sense to add a parameter for number of samples to a protocol that uses a particular DNA prep reagent kit. But it wouldn't make sense to add a parameter for *which reagent kit* to use for DNA prep. That kind of parameter would affect so many aspects of the protocol that it would make more sense to maintain a separate protocol for each kit.
```

api/docs/v2/parameters/choosing.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Consider Contradictions
+=======================
+
+Here's a common time-saving use of parameters: your protocol requires a 1-channel pipette and an 8-channel pipette, but it doesn't matter which mount they're attached to. Without parameters, you would have to assign the mounts in your protocol. Then if the robot is set up in the reverse configuration, you'd have to either physically swap the pipettes or modify your protocol.
+
+One way to get this information is to ask which mount the 1-channel pipette is on, and which pipette the 8-channel pipette is on. But if a user answers "left" to both questions — even by accident — the API will raise an error, because you can't load two pipettes on a single mount. It's no better to flip things around by asking which pipette is on the left mount, and which pipette is on the right mount. Now the user can say that both mounts have a 1-channel pipette. This is even more dangerous, because it *might not* raise any errors in analysis. The protocol could run "successfully" on a robot with two 1-channel pipettes, but produce completely unintended results.
```
sanni-t:
```suggestion
One way to get this information is to ask which mount the 1-channel pipette is on, and which mount the 8-channel pipette is on. But if a user answers "left" to both questions — even by accident — the API will raise an error, because you can't load two pipettes on a single mount. It's no better to flip things around by asking which pipette is on the left mount, and which pipette is on the right mount. Now the user can say that both mounts have a 1-channel pipette. This is even more dangerous, because it *might not* raise any errors in analysis. The protocol could run "successfully" on a robot with two 1-channel pipettes, but produce completely unintended results.
```

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+       - For numeric or string parameters.
+       - Provides a fixed list of values to choose from.
+       - Each choice has its own display name and value.
+       - Can't be used at the same time as ``minimum`` and ``maximum``.
+   * - ``units``
+     -
+       - Optional, for numeric parameters only.
+       - Displays after the number during run setup.
+       - Does not affect the parameter's value or protocol execution.
+       - Maximum [10/TK] characters.
```
sanni-t: What does TK mean? Is this a typo?

api/docs/v2/parameters/using_values.rst
jbleon95: Yup, strings, ints and floats should all use `==`, unless you want to get into the messy world of explaining cPython string and integer interning (you don't).

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+During run setup, users can toggle between the two values. In the Opentrons App, Boolean parameters appear as a toggle switch. On the touchscreen, they appear as *On* or *Off*, for ``True`` and ``False`` respectively.
+
+.. versionadded:: 2.18
+
+Integer Parameters
+------------------
+
+Integer parameters either accept a range of numbers or a list of numbers. You must specify one or the other; you can't prompt for any integer.
```
sanni-t: Not sure what "you can't prompt for any integer" is trying to say. Although that might be just a me problem.

api/docs/v2/parameters/defining.rst
ecormany: Did we loosen that for _defining_ parameters as well, or only for _entering_ values in run setup?

api/docs/v2/parameters/style.rst
ecormany: It would absolutely blow up the side nav, unless I can do some sort of TOC Sphinx override just for this page. All headers are linkable though, so it might still be worthwhile. I'll spend 5 minutes looking in the Sphinx reference but if I can't find anything I'm gonna leave it as is.

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+        variable_name="volume",
+        display_name="Aspirate volume",
+        description="How much to aspirate from each sample.",
+        default=20,
+        choices=[
+            {"display_name": "Low (10 µL)", "value": 10},
+            {"display_name": "Medium (20 µL)", "value": 20},
+            {"display_name": "High (50 µL)", "value": 50},
+        ],
+        unit="µL"
```
sanni-t: It's best to leave `unit` out of choices examples since they are actually not used. It won't throw an error but will not be executed either.

sanni-t (commented):
This is awesome!
Still going through the PR, just minor corrections so far.

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+        variable_name="volume",
+        display_name="Aspirate volume",
+        description="How much to aspirate from each sample.",
+        default=5.0,
+        choices=[
+            {"display_name": "Low (2.5 µL)", "value": 2.5},
+            {"display_name": "Medium (5 µL)", "value": 5.0},
+            {"display_name": "High (10 µL)", "value": 10.0},
+        ],
+        unit="µL"
```
sanni-t: Same as above, avoid `unit` in example of choices

api/docs/v2/parameters/defining.rst
```diff
@@ -0,0 +1,181 @@
...
+Remember, you can't mix types in a parameter. So values of ``2.5``, ``5.0``, and ``10.0`` are valid, but ``2.5``, ``5``, and ``10`` will raise an error.
+
+.. versionadded:: 2.18
+
+String Parameters
+-----------------
+
+String parameters only accept a list of values. You can't currently prompt for free text entry of a string value.
+
+To specify a list of strings, include ``choices``. Each choice is a dictionary with entries for display name and value. In most cases, the display name and value *will not* match. A common use for string display names is to provide an easy-to-read version of an API load name. You can also use them to briefly explain the effect each choice will have.
```
sanni-t: > In most cases, the display name and value *will not* match.

This sentence causes more confusion for me. Is it necessary to say this?

api/docs/v2/parameters/style.rst
ecormany: Nope, trying to insert `:tocdepth: 2` breaks all the subnav for this page. And since I have H3s for different parameter types below, I'd have to refactor the structure to make all the entries be at the same level and look the same. The juice is not worth the squeeze.

api/docs/v2/parameters/style.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Strings
+-------
+
+**Don't use string choices when a Boolean will work.** Use a Boolean when your string choices could be rephrased in terms of "true/false" or "on/off". Note that this does not apply to all two-choice situations.
+
+    - ✅ Blue, Red
+    - ✅ Left-to-right, Right-to-left
+    - ❌ Include, Exclude
+    - ❌ Yes, No
```
sanni-t, line 137: Had to re-read this to understand what the examples are saying because the 'strings' section basically just talks about booleans. Can we phrase this section better? Or maybe a simple, 'Use strings for: Blue, Red.. choices' and 'Use boolean for: Yes/ No choices' would explain this better.

api/docs/v2/parameters/defining.rst
jbleon95: We loosened it for defining it too, since it seemed overly strict to complain about a number that can be losslessly coerced into a float. We do internally convert it into a float, so when they see the parameter it should be represented as a floating point.

ecormany: Thanks, will delete this sentence as no longer needed.

sanni-t (approved):
Just the above comments from me. Everything else looks great!
Doesn't have to be in this PR but I'm wondering if it would be useful to have screenshots of the app flow to help understand the significance of all the parameters' arguments (display name/ description/ choices, etc).

api/docs/v2/parameters/use_case_sample_count.rst
ecormany: I think we have to assume this basic level of Python savvy.

api/docs/v2/parameters/defining.rst
ecormany: oooops i need to search for TKs in general. https://en.wikipedia.org/wiki/To_come_(publishing)

ecormany: verified that it's still 10 in code.

api/docs/v2/parameters/defining.rst
ecormany: Replaced with `Only the display name will appear during run setup.` which is probably more important information.

ecormany:
> Just the above comments from me. Everything else looks great! Doesn't have to be in this PR but I'm wondering if it would be useful to have screenshots of the app flow to help understand the significance of all the parameters' arguments (display name/ description/ choices, etc).

Thanks for the comments! 

I've deliberately kept the UI out of the Python docs (here and elsewhere). Although I hope that we'll have a video or some other visual resource that I can link to from one of these pages, either in time for release or shortly after.

PR #14679, opened 2024-03-18
refactor(protocol-designer): show title of protocol at final deck state

jerader:
closes RQA-2510

# Overview

This bug was introduced when refactoring out the class components in PD. There was no title specified for the end deck state

# Test Plan

Follow the steps outlined in the ticket. Create a new protocol (ot2 or flex) and go to deck map. See that the title renders with the title bar. Then click on the final deck state and see that the title bar still includes the title.

# Changelog

- define title outside of the if/else statement

# Review requests

see test plan

# Risk assessment

low


koji (commented):
lgtm

The title showed up properly.

koji (approved):

PR #14684, opened 2024-03-18
feat(app, api, shared-data, robot-server): Add module fixtures to deck configuration

CaseyBatten:
# Overview
Covers RESC-209, PLAT-247, PLAT-248, PLAT-249 and PLAT-250
App side coverage also includes PLAT-251, PLAT-252 and PLAT-254
Build out the backend and frontend to support loading modules into the deck configuration, including tracking modules by serial number in the persistent deck configuration directory.

# Known Issues to be addressed in follow-up

- During Protocol Setup on the ODD, the Modules and Deck section still separates out Modules and their Fixtures which leads to a confusing visual bug where modules are listed twice. This is non-blocking as the modules still need to be configured, calibrated, and connected. Once this is true you will be able to start the run. A following PR will implement the new designs that consolidate all deck hardware into a single table instead of splitting modules from fixtures.
- the Deck map rendering shows white space beneath modules now, this is a visual only bug that will be fixed in a following PR
- ~the Deck Configurator that is accessible from the ODD in the three dot menu of the robot dashboard, doesn't currently support adding and removing modules. Until this is added in a following PR, configure modules from the desktop app.~ (fixed)
- The Modules & Deck view on the ODD from the protocol setup view cannot be used to solve conflicts due to restrictions between the entanglement of module resolution and fixture resolution. Instead the APP should be used to resolve or the Deck Configuration view from the ODD main page.

# Test Plan
 The following Flex protocol should execute without issue, allowing the user to resolve all deck conflicts and add all modules in order to run the protocol.
```

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.17",
}


def run(protocol_context):
	tc_mod = protocol_context.load_module(module_name="thermocyclerModuleV2")
	tc_plate = protocol_context.load_labware("nest_96_wellplate_100ul_pcr_full_skirt", "C1")

	temp_mod = protocol_context.load_module(module_name="temperatureModuleV2", location="D3")
	temp_mod2 = protocol_context.load_module(module_name="temperatureModuleV2", location="D1")
	temp_adapter = temp_mod.load_adapter("opentrons_96_well_aluminum_block")
	temp_plate = protocol_context.load_labware("nest_96_wellplate_100ul_pcr_full_skirt", "D2")
	temp_adapter2 = temp_mod2.load_adapter("opentrons_96_well_aluminum_block")
	temp_plate2 = protocol_context.load_labware("nest_96_wellplate_100ul_pcr_full_skirt", "B2")
	magnetic_block = protocol_context.load_module(module_name="magneticBlockV1", location="C2")
	mag_plate = protocol_context.load_labware("nest_96_wellplate_100ul_pcr_full_skirt", "B3")

	tc_mod.open_lid()

	tiprack1 = protocol_context.load_labware("opentrons_flex_96_tiprack_1000ul", "C3")

	# move all labware into place on modules
	protocol_context.move_labware(tc_plate, tc_mod, use_gripper=True)
	protocol_context.move_labware(temp_plate, temp_adapter, use_gripper=True)
	protocol_context.move_labware(temp_plate2, temp_adapter2, use_gripper=True)
	protocol_context.move_labware(mag_plate, magnetic_block, use_gripper=True)

		"flex_1channel_1000", mount="left", tip_racks=[tiprack1]
	)
	pipette.pick_up_tip()

	pipette.aspirate(volume=10, location=tc_plate.wells()[0])
	pipette.aspirate(volume=10, location=temp_plate.wells()[0])
	pipette.aspirate(volume=10, location=temp_plate2.wells()[0])
	pipette.aspirate(volume=10, location=mag_plate.wells()[0])

	trash = protocol_context.load_trash_bin("A3")
	pipette.drop_tip()
```

# Changelog

- Updates /deck_configuration endpoint with optional `opentronsModuleSerialNumber` field
- Update deck configuration JSON file for persistent storage
- Introduces Deck Definition Schema V5 adding fixture definitions for modules
- Validates that a given module exists within the deck configuration when it is loaded in a python protocol
- Removes compatibleModuleTypes from OT3 deck definitions, maintains them for OT2 deck config support
- Serial number from deck configuration now utilized in validation of modules during load/calibration solving support for multiples of a module across all networked modules
- Introduces Fixture Groups to deck schema to allow a "fixture" that covers multiple cutouts to be broken up into individual per-cutout sections (each section a cutout fixture) and and those resulting sections be dependently grouped together into a map for load validation
- App changes to support new fixtures defs and fixture grouping
- App changes to module configuration during protocol setup to allow modules to be added to the deck configuration
- App changes for submission of module serial to deck configuration upon confirmation

# Review requests

# Risk assessment
The addition of modules to the deck configuration has changed a signification amount of backend infrastructure regarding the method in which items are validated in the deck and the way in which deck configurations are validated as well. The front end flow for protocol setup regarding the deck, module installation and calibration, and protocol run has also been greatly modified. Thorough testing should be executed on the sustained behavior and usability of all modules and protocol types across OT2 and Flex.


api/src/opentrons/protocol_api/core/legacy/legacy_protocol_core.py
```diff
@@ -288,6 +288,7 @@ def load_module(
         model: ModuleModel,
         deck_slot: Optional[DeckSlotName],
         configuration: Optional[str],
+        addressable_area: Optional[str],
```
b-cooper: this seems extraneous right?  we aren't really loading a module into an addressable area. It's more that we're loading it into a cutout right?

api/src/opentrons/protocol_api/validation.py
```diff
@@ -332,6 +336,109 @@ def ensure_and_convert_trash_bin_location(
...
+    robot_type: RobotType,
+    model: ModuleModel,
+) -> Optional[str]:
+    """Ensure module fixture load location is valid.
+
+    Also, convert the deck slot to a valid module fixture addressable area.
+    """
+
+    if robot_type == "OT-2 Standard":
+        # OT-2 Utilizes the existing compatibleModulelTypes list of traditional addressable areas
```
b-cooper:
```suggestion
        # OT-2 Utilizes the existing compatibleModuleTypes list of traditional addressable areas
```

api/src/opentrons/hardware_control/modules/types.py
```diff
@@ -64,6 +65,22 @@ def from_model(cls, model: ModuleModel) -> ModuleType:
         if isinstance(model, MagneticBlockModel):
             return cls.MAGNETIC_BLOCK
 
+    @classmethod
+    def to_module_fixture_id(cls, module_type: ModuleType) -> str:
```
b-cooper, line 68: looks like these cutout fixture ids still need the model versions added in

shared-data/deck/definitions/5/ot3_standard.json
```diff
@@ -0,0 +1,1012 @@
...
+        "cutoutD3": ["temperatureModuleV2D3"],
+        "cutoutC3": ["temperatureModuleV2C3"],
+        "cutoutB3": ["temperatureModuleV2B3"],
+        "cutoutA3": ["temperatureModuleV2A3"]
+      },
+      "fixtureGroup": [],
+      "height": 40
+    },
+    {
+      "id": "magneticBlockModuleV1",
```
b-cooper: for consistency, maybe we either rename this cutout fixture id to `magneticBlockV1` or rename it's addressable Areas to `magneticBlockModuleV1XY` ?  Slight preference for the former as we don't really call it the "magnetic block module" elsewhere

api/src/opentrons/protocol_api/core/legacy/legacy_protocol_core.py
CaseyBatten: Could be moved elsewhere potentially. Right now the logic flow is:

1. PAPI load_module command begins the load process and supplies a deck location.
2. Module fixture location is converted to a valid addressable area for that Module, if one exists (could be None, like for OT-2). Up until this point we do the same thing for fixtures commands like load_trash_bin(...).
3. Engine core load_module is passed the identified addressable_area value (Str | None)
4. Legacy load_module engine method doesn't use this value, current load_module engine method will supply the engine client load_module method with the validated addressable area if it is not None

I could probably collapse this down to occur inside the engine-side load_module logic instead if we want.

shared-data/deck/definitions/5/ot3_standard.json
```diff
@@ -0,0 +1,1014 @@
...
+    {
+      "id": "thermocyclerModuleV2Front",
+      "expectOpentronsModuleSerialNumber": true,
+      "mayMountTo": ["cutoutB1"],
+      "displayName": "Front Slot portion of the Thermocycler Moduler",
+      "providesAddressableAreas": {
+        "cutoutB1": ["thermocyclerModuleV2"]
+      },
+      "fixtureGroup": ["thermocyclerModuleV2Rear", "thermocyclerModuleV2Front"],
+      "height": 170.66
```
SyntaxColoring: 1. The schema [defines this](https://github.com/Opentrons/opentrons/blob/0974d69a62d40012ec2dcc2c79737f43adc9d565/shared-data/deck/schemas/5.json#L302-L305) as "The vertical distance (mm) **from the cutout fixture's origin** to its tallest physical feature that an instrument could collide with." The origin of the cutout fixture is identical to the origin of the cutout, right? So will this accidentally say that the Thermocycler juts 170.66 mm above the deck, whereas in fact it's approximately flush with the deck?
2. I would really really love to keep the module definitions as the single source of truth for this kind of thing.

shared-data/deck/definitions/5/ot3_standard.json
```diff
@@ -0,0 +1,1014 @@
...
+    {
+      "id": "thermocyclerModuleV2Rear",
+      "expectOpentronsModuleSerialNumber": true,
+      "mayMountTo": ["cutoutA1"],
+      "displayName": "Rear Slot portion of the Thermocycler Moduler",
+      "providesAddressableAreas": {
+        "cutoutA1": []
+      },
+      "fixtureGroup": ["thermocyclerModuleV2Rear", "thermocyclerModuleV2Front"],
+      "height": 211.91
```
SyntaxColoring: 1. See my other note about this number possibly being too large, depending on the origin.
2. See my other note about keeping the module definitions as the single source of truth for this kind of thing.
3. The schema [defines this](https://github.com/Opentrons/opentrons/blob/0974d69a62d40012ec2dcc2c79737f43adc9d565/shared-data/deck/schemas/5.json#L302-L305) as "The vertical distance (mm) from the cutout fixture's origin to its tallest physical feature **that an instrument could collide with.**" The back of the Thermocycler lid is mostly out of the way. I think for the 96-channel, we have some bounding-box collision prediction added by @sanni-t. Will this make a bunch of good protocols stop working because, for example, the gripper can't lift a labware high enough to go above the Thermocycler lid?

shared-data/deck/definitions/5/ot3_standard.json
```diff
@@ -0,0 +1,1014 @@
...
+    },
+    {
+      "id": "thermocyclerModuleV2Front",
+      "expectOpentronsModuleSerialNumber": true,
+      "mayMountTo": ["cutoutB1"],
+      "displayName": "Front Slot portion of the Thermocycler Moduler",
+      "providesAddressableAreas": {
+        "cutoutB1": ["thermocyclerModuleV2"]
+      },
+      "fixtureGroup": ["thermocyclerModuleV2Rear", "thermocyclerModuleV2Front"],
```
SyntaxColoring: As a thought experiment, what would we have to do in order to not chop up the Thermocycler like this?

shared-data/deck/definitions/5/ot3_standard.json
```diff
@@ -0,0 +1,1014 @@
...
+        "id": "heaterShakerV1D1",
+        "areaType": "heaterShaker",
+        "offsetFromCutoutFixture": [0.0, 0.0, 0.0],
+        "boundingBox": {
+          "xDimension": 128.0,
+          "yDimension": 86.0,
+          "zDimension": 0
+        },
+        "displayName": "Heater Shaker in D1"
+      },
```
SyntaxColoring: I'm confused about what these new addressable areas represent. Is it the slot that the caddy provides, into which the module is loaded? Or is it the slot that the module provides, into which the labware is loaded?

SyntaxColoring (commented):
Thanks! Preliminary questions and comments.

components/src/hardware-sim/DeckConfigurator/HeaterShakerFixture.tsx
```diff
@@ -0,0 +1,101 @@
...
+  Y_ADJUSTMENT,
+} from './constants'
+
+import type {
+  CutoutFixtureId,
+  CutoutId,
+  DeckDefinition,
+} from '@opentrons/shared-data'
+
+// TODO(BC, 2024-03-21): This component is almost identical to TemperatureModuleFixture, consider consolidating?
```
brenthagen, line 23: there's a possible shared abstraction between all of these "config" fixtures but in thinking it over for deck config the benefit seems limited. a shared abstraction could be useful for quick transfer though

app/src/resources/deck_configuration/hooks.ts
```diff
@@ -107,10 +91,7 @@ export function useDeckConfigurationCompatibility(
           cutoutFixtureId,
           requiredAddressableAreas: requiredAddressableAreasForCutoutId,
           // Thermocycler requires an "empty" (single slot) fixture in A1 that is not referenced directly in protocol
```
brenthagen: should delete this comment along with the rest of the excellent deletion in this file

app/src/organisms/ModuleWizardFlows/PlaceAdapter.tsx
```diff
@@ -60,6 +60,9 @@ export const PlaceAdapter = (props: PlaceAdapterProps): JSX.Element | null => {
   } = props
   const { t } = useTranslation('module_wizard_flows')
   const mount = attachedPipette.mount
+  const slotName = deckConfig.find(cc => (
+    cc.opentronsModuleSerialNumber === attachedModule.serialNumber
+  ))?.cutoutId?.replace('cutout', '') ?? null
```
brenthagen: there's a mapping `FLEX_SINGLE_SLOT_BY_CUTOUT_ID` in shared-data

brenthagen (approved):
very solid on the js side. there are some edge cases to fix but should be safe to merge

Screen Shot 2024-04-12 at 4 22 32 PM
Screen Shot 2024-04-15 at 10 33 02 AM

vegano1 (approved):
The robot-server and protocol engine changes look good to me, we also spoke about some potential optimizations we could make in the future.

1. The definition for each cutout might not be required for fixtures that have the same offsets/dimensions, ex "heaterShakerA1”, “heaterShakerA3”,…,”heaterShakerD1”. We might be able to have one definition that is referenced in software for multiple cutouts.
2. OT-2 still uses the "legacy" deck definition, it would be good to eventually port over this deck config work to encompass the OT-2 and remove the 2 logic paths we currently have.

y3rsh (commented):
Please review the snapshot changes that this branch causes.
These are reviewable in #14909
Do not merge it in, only use it as another point of data.
Ping me if you have questions.

PR #14687, opened 2024-03-19
Abr command data

rclarke0:
# Overview

Upload command data from a folder of run logs to a google sheet

# Test Plan

Ran script on my PC

# Changelog

**in abr_read_logs.py**
- Added google_sheet_name, google_sheet_tab_number, and header as arguments to make script less specific to abr
- made local csv and google sheet the same name
**Added abr_command_data script.**
Script reads run logs and organizes commandData into pipetteCommands, moduleCommands, setupCommands, and movementCommands. 
data dictionaries get written to google sheet, with each set of commands on a seperate tab.
**abr_asair_sensor**
converted user inputs into arguments to speed up execution of script
**abr_run_logs**
removed abr_ips import
# Review requests

# Risk assessment


hardware-testing/hardware_testing/abr_tools/abr_command_data.py
```diff
@@ -0,0 +1,523 @@
...
+                (
+                    labware.get("displayName", labware.get("loadName", ""))
+                    if labware["id"] == labware_id
+                    and labware.get("displayName") is not None
+                    else labware.get("loadName", "")
+                    for labware in all_labware
+                ),
+                "",
+            )
+            group_key = (commandType, labware_name)
```
sfoster1, line 217: I think this means that all `moveTo` to _any_ e.g. `nest_96_wellplate_2ml_deep`, even if there's more than one `nest_96_wellplate_2ml_deep` on the deck, gets grouped together. Maybe htat's what you want, but if it isn't you may want to base the group key on the labware ID or something.

sfoster1 (approved):
Looks good to me, though I don't know if you want to group the movement commands like this.

hardware-testing/hardware_testing/abr_tools/abr_command_data.py
rclarke0: changed it to default to labwareid if no display name

PR #14688, opened 2024-03-19
feat(robot-server, api): accept RTP overrides via /protocols and inject them in python executor

sanni-t:
Closes AUTH-64, AUTH-65

# Overview

Updates the `POST /protocols` API to accept an additional `runTimeParameterValues` form-data field. This field is expected to be a stringified JSON composed of key-value pairs of the parameter variable name and its override value.

These override values are converted into a dictionary and passed into the protocol runner, which makes them available to the python executor.

# Test Plan

- [ ] Send a POST protocol request and verify that an analysis is created
- [ ] POST the same protocol again and verify that the response still gives the same analysis & protocol IDs
- [ ] Send a protocol with `runTimeParameterValues` in request data and verify the request goes through, and a new analysis is created on the same protocol ID

# Known issues

- every time a protocol creation request contains a non-empty `runTimeParameterValues` data, analysis is triggered, regardless of whether we have previously analyzed protocols with the same param values
- if no param values are sent with an existing protocol, a new analysis is NOT triggered, regardless of whether the previous analyses were done using any RTP values or not

These will be addressed in AUTH-70 

# Review requests

- check for usual code sanity

# Risk assessment

High. Affects the correctness of order and contents of analysis summaries returned by the `POST /protocols` endpoint when using run time parameters. This behavior will remain broken until AUTH-70 is implemented. Until then, please be cautious that a protocol analysis request that uses custom RTP values will modify an existing protocol analysis database such that the app might pick the wrong analysis for a run. I will recommend deleting such protocols (and runs) from the database after you are done testing, especially if you are testing on a shared robot.


robot-server/robot_server/protocols/router.py
```diff
@@ -205,11 +211,32 @@ async def create_protocol(
...
+            task_runner.run(
+                protocol_analyzer.analyze,
+                protocol_resource=resource,
+                analysis_id=analysis_id,
+                run_time_param_overrides=parsed_rtp,
+            )
+            analysis_store.add_pending(
+                protocol_id=cached_protocol_id,
+                analysis_id=analysis_id,
+            )
```
SyntaxColoring: This will probably need to be implemented in a different PR, but at some point before releasing this, we'll need to cap the maximum number of protocol analyses. I'm curious if you have any thoughts on what the limit will be?

sanni-t: Ya, we'll have to figure that out. How did we land on the max number of protocols and runs to store? Is there a hard cut off for memory availability we need to keep in mind?

SyntaxColoring: It's mostly how long it would take to parse them all into Pydantic models if we had to do some kind of migration.

api/src/opentrons/protocols/execution/execute.py
```diff
@@ -16,7 +17,11 @@
 MODULE_LOG = logging.getLogger(__name__)
 
 
-def run_protocol(protocol: Protocol, context: ProtocolContext) -> None:
+def run_protocol(
+    protocol: Protocol,
+    context: ProtocolContext,
+    run_time_param_overrides: Optional[Dict[str, Union[float, bool, str]]] = None,
```
jbleon95, line 24: We should use the PE type here if possible

robot-server/robot_server/protocols/router.py
```diff
@@ -184,6 +189,7 @@ async def create_protocol(
     Arguments:
         files: List of uploaded files, from form-data.
         key: Optional key for client-side tracking
+        runTimeParameterValues: Key value pairs of run-time parameters defined in a protocol.
```
jbleon95: Let's make this snake_case for consistency

jbleon95 (approved):
Some small style nitpicks but otherwise LGTM!

robot-server/robot_server/protocols/router.py
sanni-t: It's actually camel case for robot-server HTTP API consistency purposes.

api/src/opentrons/protocols/execution/execute.py
sanni-t: I had a TODO for this exact thing. I/ pycharm must have removed it during refactor. Putting it back in.

robot-server/robot_server/protocols/router.py
```diff
@@ -165,6 +166,10 @@ async def create_protocol(
             " protocol resources on the robot."
         ),
     ),
+    runTimeParameterValues: Optional[str] = Form(
+        default=None,
+        description="Key value pairs of run-time parameters defined in a protocol.",
+    ),
```
SyntaxColoring: It might be worth clarifying in this description that:

* This is a string holding a JSON object.
* This affects the auto-started analysis resource (`GET /protocols/{id}/analyses/{id}`).

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
```diff
@@ -28,7 +28,7 @@ stages:
 
   - name: Check that the analysis summary is present in /protocols/:id; retry until it says it's completed
     max_retries: 5
-    delay_after: 1
+    delay_after: 5
```
SyntaxColoring: A 5-second retry interval might make the tests feel pretty slow if it misses the first poll. If this was timing out for you, can we fix that by increasing `max_retries` instead?

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
```diff
@@ -84,3 +84,27 @@ stages:
...
+      method: POST
+      data:
+        runTimeParameterValues: '{{"volume": 123, "dry_run": true, "pipette": "p10_single"}}'
+      files:
+        files: 'tests/integration/protocols/basic_transfer_standalone.py'
+    response:
+      save:
+        json:
+          protocol_id: data.id
+          analysis_id: data.analysisSummaries[1].id
```
SyntaxColoring: Looks like this is unused.

```suggestion
```

robot-server/robot_server/protocols/router.py
```diff
@@ -205,11 +211,32 @@ async def create_protocol(
         assert file.filename is not None
     buffered_files = await file_reader_writer.read(files=files)  # type: ignore[arg-type]
 
+    if isinstance(runTimeParameterValues, str):
+        # We have to do this isinstance check because if `runTimeParameterValues` is
+        # not specified in the request, then it gets assigned a Form(None) value
+        # instead of just a None. \(O.o)/
+        parsed_rtp = json.loads(runTimeParameterValues)
```
SyntaxColoring: Nit, doesn't have to happen in this PR: This will raise a 500 error if the client passes invalid JSON. Ideally, it would return a 422 instead. I wonder if there's a way to make a custom "run-time parameter multipart form field" Pydantic type so FastAPI can do that for us.

SyntaxColoring (approved):
hype

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
sanni-t: oh shoot! Forgot to undo this. I was just debugging something

robot-server/robot_server/protocols/router.py
sanni-t: Hmm.. interesting idea. I'll add a TODO

PR #14689, opened 2024-03-19
refactor(protocol-designer): metadata fields update with new protocol

jerader:
closes RQA-2527

# Overview

Fixes an error with react-hook-form in `FilePage`. Basically, if the default values change, the changes aren't reflected in the File page's input fields because they are getting the values from react-hook-form's `watch`. But we need to still use `watch` because it re-renders when the field is updated. The fix is to add a `useEffect` to update the fields when the default values update to keep things synchronous.

# Test Plan

Create a ot-2 or flex protocol and add a protocol name, description, and author. See that the fields were populated correctly once you complete the create file wizard in the Information section. Now create a new protocol and add different info to those fields. They should update correctly in the Information section.

# Changelog

- add use effect

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/FilePage.tsx
+++ b/protocol-designer/src/components/FilePage.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { createPortal } from 'react-dom'
 import { Controller, useForm } from 'react-hook-form'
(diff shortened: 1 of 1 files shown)


koji (approved):

PR #14691, opened 2024-03-19
Upload Scale Value to google sheet

rclarke0:
# Overview

Reads scale value until it becomes stable and uploads value to google sheet and local csv.

# Test Plan

Tested with ABR scale.

# Changelog

Moved abr_scale script to abr_tools folder
Added labware, robot, measurement step, storage directory, google sheet name as arguments
saves only the stable value, rather than all of the unstable values
asks the user if they want to measure another sample to reduce the number of times all arguments need to be put in .
deleted analyze_abr script because no longer needed. Analysis is automated on the google sheet. 

# Review requests

# Risk assessment


hardware-testing/hardware_testing/abr_tools/abr_scale.py
```diff
@@ -0,0 +1,120 @@
...
+import datetime
+from hardware_testing.drivers import find_port, list_ports_and_select
+from hardware_testing.drivers.radwag import RadwagScale
+from typing import Any, List
+import argparse
+import csv
+from .abr_read_logs import create_abr_data_sheet
+
+
+def write_to_sheets(file_name_csv: str, google_sheet: Any, row_list: List) -> None:
```
sfoster1, line 13: We're using this function quite a lot, right? Like we use it in pretty much all of our scripts at this point? Might be better to put it in a reusable module.

hardware-testing/hardware_testing/abr_tools/abr_scale.py
```diff
@@ -0,0 +1,120 @@
...
+    """Write list to google sheet and csv."""
+    sheet_location = os.path.join(storage_directory, file_name_csv)
+    with open(sheet_location, "a", newline="") as f:
+        writer = csv.writer(f)
+        writer.writerow(row_list)
+        print(f"Written {row_list} point to {file_name_csv}")
+        # Read Google Sheet
+        if google_sheet.creditals.access_token_expired:
```
sfoster1, line 21: `creditals`? `credentials`?

sfoster1 (changes requested):

hardware-testing/hardware_testing/abr_tools/abr_scale.py
```diff
@@ -0,0 +1,120 @@
...
+    try:
+        google_sheet = google_sheets_tool.google_sheet(
+            credentials_path, file_name, tab_number=0
+        )
+        print("Connected to google sheet.")
+    except FileNotFoundError:
+        print("No google sheets credentials. Add credentials to storage notebook.")
+    # Scale Loop
+    break_all = False
+    while is_stable is False:
```
DerekMaggio, line 102: Does this blow up your terminal with print messages waiting for the scale to stabilize? If so, you might add a small sleep at the end of your while false statement

rclarke0: Yes. I like seeing that though because it gives you an indication of how long the scale is taking.

DerekMaggio: You could take the time before you enter the while false loop and then take it when you enter the true loop and print that difference. 
Whatever you prefer though.

hardware-testing/hardware_testing/abr_tools/abr_scale.py
rclarke0: that's the function spelling in this module
`import gspread
import socket
import httplib2
from oauth2client.service_account import ServiceAccountCredentials
from typing import (Any, Dict, Iterator, List,
                    Optional, Sequence, Set, Tuple, Union)

"""This module requires a creditials.json file before getting started.
The following link provides a step by step tutorial on how to get startd.
https://towardsdatascience.com/accessing-google-spreadsheet-data-using-python-90a5bc214fd2"""

class google_sheet:
    """
    This module was designed for the testing team to automate tests by
    ultizing google sheets API to record data. The following attributes that this
    class needs is the file_name and the tab_number define.
    """
    def __init__(self, creditals, file_name, tab_number):
        self.scope = ['https://spreadsheets.google.com/feeds',
                                        'https://www.googleapis.com/auth/drive']
        self.creditals = \
            ServiceAccountCredentials.from_json_keyfile_name(creditals, self.scope)
        self.gc = gspread.authorize(self.creditals)
        self.file_name = file_name
        self.tab_number = tab_number
        self.spread_sheet = self.open_google_sheet()
        self.worksheet = self.open_worksheet(self.tab_number)
        self.row_index = 1

    def open_google_sheet(self):
        """Open Google Spread Sheet"""
        sheet = self.gc.open(self.file_name)
        return sheet

    def open_worksheet(self, tab_number: int )-> classmethod:
        """Open individual worksheet within a googlesheet"""
        return self.spread_sheet.get_worksheet(tab_number)

    def create_worksheet(self, tab_name: int):
        """Create a worksheet with a title for a tab. This method need
        to have an existing spreadsheet for it to work."""
        try:
            self.spread_sheet.add_worksheet(tab_name, rows="1000", cols="26")
        except gspread.exceptions.APIError:
            print('Work Sheet already exists')

    def write_header(self, header: List):
        """Write Header to first row, this method as determines if the first row
        has a header that identical to the header the user is writing too."""
        header_list = self.worksheet.row_values(1)
        print(header_list)
        if header_list != header:
            self.worksheet.insert_row(header, self.row_index)

    def write_to_row(self, data: List):
        """Write data into a row in a List[] format, this method skips the first
        row of the worksheet due to headers. """
        try:
            self.row_index += 1
            self.worksheet.insert_row(data, index= self.row_index)
        except socket.gaierror:
            pass
        except httplib2.ServerNotFoundError:
            print('UNABLE TO CONNECT TO SERVER!!, CHECK CONNECTION')
        except Exception as error:
            print(error.__traceback__)

    def delete_row(row_index: int):
        """Delete Row from google sheet"""
        self.worksheet.delete_row(row_index)

    def update_cell(self, row: int, column: int, single_data: any):
        """Update an individual cell according to a row and column. Note that
        this function only updates one individual cell"""
        self.worksheet.update_cell(row, column, data)

    def update_cells(self, cells, data):
        cell_list = self.worksheet.range(cells)
        self.worksheet.update_cells(data)

    def get_all_data(self)-> Dict[str, any]:
        """Get all the Data recorded from worksheet."""
        return self.worksheet.get_all_records()

    def get_column(self, column_number: int)-> int:
        return self.worksheet.col_values(column_number)

    def get_index_row(self):
        """Check for the next available row to write too."""
        row_index = len(self.get_column(1))
        print("Row Index: ", row_index)
        return row_index

    def update_row_index(self):
        """Update self.row_index instance variable"""
        self.row_index = self.get_index_row()


    def get_all_sheets(self):
        """List all tabs in the spreadsheets"""
        worksheets = self.spread_sheet.worksheets()
        return worksheets

    def get_sheet_by_name(self, title: str):
        #worksheet = sh.worksheet("January")
        try:
            worksheet = self.spread_sheet.worksheet(title)
            return worksheet
        except gspread.exceptions.WorksheetNotFound:
            raise 'Worksheet does not exist!!, create a \
                    worksheet first using the create_worksheet() function.'

    def token_check():
        if self.creditals.access_token_expired:
            self.gc.login()
    
    
`

sfoster1 (approved):
Ok, looks good!

PR #14692, opened 2024-03-19
feat(app): add value key to RTP type

jerader:
closes AUTH-216

# Overview

Add the top level value key to the run time parameters type.

# Test Plan

- just check the code. i updated the mocks and the type

# Changelog

- update mocks
- update type

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/shared-data/js/types.ts
+++ b/shared-data/js/types.ts
@@ -621,6 +621,7 @@ interface BaseRunTimeParameter {
   variableName: string
   description: string
   type: RunTimeParameterType
+  value: unknown
   suffix?: string
 }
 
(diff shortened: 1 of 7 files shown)


koji (approved):
lgtm
thank you for updating

shared-data/js/types.ts
```diff
@@ -621,6 +621,7 @@ interface BaseRunTimeParameter {
   variableName: string
   description: string
   type: RunTimeParameterType
+  value: unknown
```
koji, line 624: maybe 

```suggestion
  value: string | number | boolean
```

jerader: @shlokamin mentioned in standup that the value should be type `any` or `unknown` (but `unknown` is better) since i guess technically, a user could specify anything for their parameters. The value type in the `choice` object is also `unknown` for the same reason.

PR #14693, opened 2024-03-19
fix(app): fix font size in parameters screen odd

koji:
# Overview
the design team requested us to update the font size of value in parameters screen. (22px -> 28px)
close AUTH-218

[related thread](https://opentrons.slack.com/archives/C06M6R72UDS/p1710863687506439)

Screenshot 2024-03-19 at 2 22 56 PM
Screenshot 2024-03-19 at 2 22 40 PM

# Test Plan

# Changelog

# Review requests
If there is more efficient way to fix this except re-structuring a component, please let me know.

# Risk assessment
low

### DIFF
--- a/app/src/pages/ProtocolSetup/index.tsx
+++ b/app/src/pages/ProtocolSetup/index.tsx
@@ -117,7 +117,7 @@ interface ProtocolSetupStepProps {
   //  optional removal of the icon
   hasIcon?: boolean
   //  optional enlarge the font size
-  hasLargeFont?: boolean
+  fontSize?: string
 }
 
 export function ProtocolSetupStep({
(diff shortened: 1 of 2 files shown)


app/src/pages/ProtocolSetup/index.tsx
```diff
@@ -116,6 +116,8 @@ interface ProtocolSetupStepProps {
   description?: string
   //  optional removal of the icon
   hasIcon?: boolean
+  //  optional enlarge the font size
+  hasLargeFont?: boolean
```
jerader: do we know if this is the only other font size? maybe instead of a boolean, we could call it "fontSize" and then default it to `p`.

jerader (approved):
yay! thanks for addressing the design feedback from Sue. Just left a comment

app/src/pages/ProtocolSetup/index.tsx
koji: Actually not sure... , but yeah at this moment what you suggested is better than passing `boolean` prop.
I will update this. Thanks!

PR #14697, opened 2024-03-19
refactor(protocol-designer): fix dropdown selection for moveLabware

jerader:
closes RQA-2503

# Overview
The bug outlined in the ticket shows that once you create a timeline error, the dropdown options are incorrect when editing. This PR fixes that. The logic to determine if the presaved step was a `moveLabware` or `moveLiquid` was not good enough for if there was a timeline error because the presaved step info gets deleted and returns null.

The fix was to split the selector up to accommodate the 2 cases since idk what info to rely on since we can't rely on the presaved step info.

# Test Plan

Create a flex protocol and add a gripper. Add an off deck labware. Move the offdeck labware to the deck using the gripper. The "labware off-deck" error should arise. Edit it by changing the labware to the tiprack. The error should resolve and the step should be properly created. 

# Changelog

- split the selector into 2 selectors to remove the brittle branching logic, create a new `MoveLiquidField` to keep the logic for move labware and move liquid separate
- fix test

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/StepEditForm/fields/MoveLabwareField.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { useSelector } from 'react-redux'
+import { getMoveLabwareOptions } from '../../../ui/labware/selectors'
+import { StepFormDropdown } from './StepFormDropdownField'
+import type { FieldProps } from '../types'
+
+export const MoveLabwareField = (props: FieldProps): JSX.Element => {
```
koji:
```suggestion
export function MoveLabwareField(props: FieldProps): JSX.Element {
```

protocol-designer/src/components/StepEditForm/fields/index.ts
```diff
@@ -12,6 +12,7 @@ export { DelayFields } from './DelayFields'
 export { DisposalVolumeField } from './DisposalVolumeField'
 export { FlowRateField } from './FlowRateField'
 export { LabwareField } from './LabwareField'
+export { MoveLabwareField } from './MoveLabwareField'
 export { LabwareLocationField } from './LabwareLocationField'
```
koji: nit

```suggestion
export { LabwareLocationField } from './LabwareLocationField'
export { MoveLabwareField } from './MoveLabwareField'
```

protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -35,30 +39,63 @@ export const _sortLabwareDropdownOptions = (options: Options): Options =>
...
+
+  let nickName = nicknamesById[labwareId]
+  if (module != null) {
+    nickName = `${nicknamesById[labwareId]} in ${module}`
+  } else if (isOffDeck) {
+    nickName = `${nicknamesById[labwareId]} off-deck`
+  } else if (isLabwareInColumn4) {
+    nickName = `${nicknamesById[labwareId]} in staging area slot`
+  }
+  return nickName
```
koji: nit

```suggestion
  const nickName = nicknamesById[labwareId]
  if (module != null) {
    return  `${nickname} in ${module}`
  } else if (isOffDeck) {
    return `${nickname} off-deck`
  } else if (isLabwareInColumn4) {
    return `${nickname} in staging area slot`
  }
  return nickName
```

protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -35,30 +39,63 @@ export const _sortLabwareDropdownOptions = (options: Options): Options =>
     return a.name.localeCompare(b.name)
   })
 
-/** Returns options for labware dropdowns.
+const getNickNames = (
```
koji: return type is string so getNickName would be a better name.

```suggestion
const getNickName = (
```

protocol-designer/src/ui/labware/selectors.ts
```diff
@@ -73,66 +110,86 @@ export const getLabwareOptions: Selector<Options> = createSelector(
...
+          form =>
+            form.stepType === 'moveLabware' &&
+            form.labware === labwareId &&
+            form.newLocation === wasteChuteLocation
         )
-        const module =
-          moduleOnDeck != null ? getModuleShortNames(moduleOnDeck.type) : null
 
-        const isLabwareInColumn4 = getLabwareInColumn4(
+        const isAdapter = labwareEntity.def.allowedRoles?.includes('adapter')
```
koji: isAdapter maybe undefined

```suggestion
        const isAdapter = labwareEntity.def.allowedRoles?.includes('adapter') || false
```

koji (approved):
left a few comments, but overall the changes make sense to me.

protocol-designer/src/ui/labware/selectors.ts
jerader: oh ya good catch!

PR #14699, opened 2024-03-20
chore: Make analyses-snapshot-test PR titles usable as commit titles

SyntaxColoring:
# Overview

Something in our GitHub config changed recently where, when you click the green merge button on a PR like #14586, it autocompletes the commit message like this:

> Evaluate Analyses Snapshot Update edge ([PR number])

We want it to autocomplete like this instead:

> fix(app-testing): snapshot failure capture ([PR number])

# Test Plan

None needed.

# Changelog

* Change the titles of PRs created by the `analyses-snapshot-test.yaml` workflow, because they're used to autocomplete the squash commit message.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/.github/workflows/analyses-snapshot-test.yaml
+++ b/.github/workflows/analyses-snapshot-test.yaml
@@ -65,7 +65,7 @@ jobs:
       uses: peter-evans/create-pull-request@v5
       with:
           commit-message: 'fix(app-testing): snapshot failure capture'
-          title: 'Evaluate Analyses Snapshot Update ${{ env.TARGET }}'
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Oh, thanks! This was actually a change we made to make squash-merge outputs default to PR title, because
- This was actually always true for PRs with more than one commit
- But PRs with just one commit (as you noticed) would use the first line of that commit

So we would get a lot of merges that had bad titles. This way, as long as the PR title is ok, the commit will default to that.

This change looks good!

PR #14702, opened 2024-03-20
fix(app): refactor Flex pipette card to reference /instruments

smb2268:
fix RQA-2433

# Overview
Currently, the pipette cards on the desktop app for Flex reference `/pipettes` while those on ODD reference `/instruments`. This has caused confusion because occasionally the pipettes would be visible on desktop but not ODD. Seth made a change in https://github.com/Opentrons/opentrons/pull/14608 that makes the `/instruments` response report cached data like `/pipettes` does, which should have eliminated this discrepancy. This PR ensures that by using `/instruments` as the one source of truth for pipettes on Flex.

# Test Plan
Smoke test pipette cards and all possible overflow menu actions for both Flex and OT-2 pipettes. This includes OT-2 pipette settings, attach, detach, and calibrate flows, drop tip wizard, and the about pipette menu.

# Changelog
1. Create new `FlexPipetteCard` and pull all `Flex` specific logic out of `PipetteCard` into here.
2. Cleanup `PipetteCard` and `PipetteOverflowMenu`, removing all `Flex` logic which has been moved to `FlexPipetteCard` and some leftover OT-2 calibration logic that is no longer used since we only allow pipette cal from the calibration dashboard now.
3. From `InstrumentsAndModules` only enable `usePipettesQuery` for OT-2 and `useInstrumentsQuery` for Flex. Display `PipetteCard` or `FlexPipetteCard` accordingly.
4. Update all tests, refactoring some logic to bring them in line with our vitest best practices

# Review requests
Look through the code - did I miss anything? And test this out if you can!

# Risk assessment
Medium

### DIFF
--- a/app/src/molecules/InstrumentCard/index.tsx
+++ b/app/src/molecules/InstrumentCard/index.tsx
@@ -2,6 +2,7 @@ import * as React from 'react'
 
 import {
   ALIGN_FLEX_START,
+  ALIGN_CENTER,
   BORDERS,
   Box,
   COLORS,
@@ -80,13 +81,19 @@ export function InstrumentCard(props: InstrumentCardProps): JSX.Element {
         </Flex>
       ) : null}
       {instrumentDiagramProps?.pipetteSpecs != null ? (
-        <InstrumentDiagram
-          pipetteSpecs={instrumentDiagramProps.pipetteSpecs}
-          mount={instrumentDiagramProps.mount}
-          transform="scale(0.3)"
-          size="3.125rem"
-          transformOrigin="20% -10%"
-        />
+        <Flex
+          alignItems={ALIGN_CENTER}
+          width="3.75rem"
+          height="3.375rem"
+          paddingRight={SPACING.spacing8}
+        >
+          <InstrumentDiagram
+            pipetteSpecs={instrumentDiagramProps.pipetteSpecs}
+            mount={instrumentDiagramProps.mount}
+            transform="scale(0.3)"
+            transformOrigin={'-5% 52%'}
+          />
+        </Flex>
       ) : null}
       <Flex
         alignItems={ALIGN_FLEX_START}
--- a/app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
+++ b/app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
@@ -98,7 +98,7 @@ export function FlexPipetteCard({
   const handleChoosePipette: React.MouseEventHandler<HTMLButtonElement> = () => {
     setShowChoosePipette(true)
   }
-  const handleAttach = () => {
+  const handleAttach = (): void => {
     setShowChoosePipette(false)
     handleLaunchPipetteWizardFlows(FLOWS.ATTACH)
   }
--- a/app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
+++ b/app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
@@ -1,0 +1,281 @@
+import * as React from 'react'
+import { Trans, useTranslation } from 'react-i18next'
+import { css } from 'styled-components'
+import { SPACING, TYPOGRAPHY, StyledText } from '@opentrons/components'
+import {
+  NINETY_SIX_CHANNEL,
+  SINGLE_MOUNT_PIPETTES,
+  FLEX_ROBOT_TYPE,
+  LEFT,
+} from '@opentrons/shared-data'
+import {
+  useCurrentSubsystemUpdateQuery,
+  useHost,
+} from '@opentrons/react-api-client'
+import { Banner } from '../../../atoms/Banner'
+import { InstrumentCard } from '../../../molecules/InstrumentCard'
+import { ChoosePipette } from '../../PipetteWizardFlows/ChoosePipette'
+import { FLOWS } from '../../PipetteWizardFlows/constants'
+import { handlePipetteWizardFlows } from '../../PipetteWizardFlows'
+import { DropTipWizard } from '../../DropTipWizard'
+
+import { AboutPipetteSlideout } from './AboutPipetteSlideout'
+
+import type {
+  BadPipette,
+  Mount,
+  PipetteData,
+  HostConfig,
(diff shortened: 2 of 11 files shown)


app/src/molecules/InstrumentCard/index.tsx
```diff
@@ -80,13 +81,19 @@ export function InstrumentCard(props: InstrumentCardProps): JSX.Element {
...
       {instrumentDiagramProps?.pipetteSpecs != null ? (
-        <InstrumentDiagram
-          pipetteSpecs={instrumentDiagramProps.pipetteSpecs}
-          mount={instrumentDiagramProps.mount}
-          transform="scale(0.3)"
-          size="3.125rem"
-          transformOrigin="20% -10%"
-        />
+        <Flex
+          alignItems={ALIGN_CENTER}
```
sfoster1, line 85: This is a separate bug, right? Like it's logically equivalent with the new `transformOrigin` at the standard viewpoint size, but it reflows better?

sfoster1 (approved):
Code looks good to me!

app/src/molecules/InstrumentCard/index.tsx
smb2268: Yeah this code wasn't hit previously because we weren't using `InstrumentCard` for pipettes - so I pulled the logic from what was actually being hit which lived here https://github.com/Opentrons/opentrons/blame/edge/app/src/organisms/Devices/PipetteCard/index.tsx#L261 and was recently updated by Nick to fix a styling bug

app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+export function FlexPipetteCard({
+  pipetteModelSpecs,
+  attachedPipette,
+  mount,
+  isRunActive,
+  isEstopNotDisengaged,
+}: FlexPipetteCardProps): JSX.Element {
+  const { t, i18n } = useTranslation(['device_details', 'shared'])
+  const host = useHost() as HostConfig
```
jerader, line 69: do you need to type this? isn't host already type `HostConfig`?

app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
```diff
@@ -0,0 +1,282 @@
...
+      refetchInterval: POLL_DURATION_MS,
+    }
+  )
+  const pipetteDisplayName = pipetteModelSpecs?.displayName
+  // we should poll for a subsystem update from the time a bad instrument is
+  // detected until the update has been done for 5 seconds
+  // this gives the instruments endpoint time to start reporting
+  // a good instrument
+  React.useEffect(() => {
+    if (attachedPipette?.ok === false) {
```
jerader, line 134: nit
```suggestion
    if (!attachedPipette?.ok) {
```

app/src/organisms/Devices/PipetteCard/__tests__/FlexPipetteCard.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+import { useCurrentSubsystemUpdateQuery } from '@opentrons/react-api-client'
+import { i18n } from '../../../../i18n'
+import { handlePipetteWizardFlows } from '../../../PipetteWizardFlows'
+import { AboutPipetteSlideout } from '../AboutPipetteSlideout'
+import { FlexPipetteCard } from '../FlexPipetteCard'
+import { ChoosePipette } from '../../../PipetteWizardFlows/ChoosePipette'
+import { DropTipWizard } from '../../../DropTipWizard'
+
+import type { PipetteData } from '@opentrons/api-client'
+import { mockLeftSpecs } from '../../../../redux/pipettes/__fixtures__'
```
jerader: import can go higher

app/src/organisms/Devices/PipetteCard/index.tsx
```diff
@@ -237,127 +124,66 @@ export const PipetteCard = (props: PipetteCardProps): JSX.Element => {
...
+                alignItems={ALIGN_CENTER}
+                width="3.75rem"
+                height="3.375rem"
+                paddingRight={SPACING.spacing8}
+              >
+                <InstrumentDiagram
+                  pipetteSpecs={pipetteModelSpecs}
+                  mount={mount}
+                  transform="scale(0.3)"
+                  transformOrigin={'20% 52%'}
```
jerader: nit: 
```suggestion
                  transformOrigin="20% 52%"
```

app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
```diff
@@ -0,0 +1,282 @@
...
+      ) : null}
+      {attachedPipette?.ok && showAboutPipetteSlideout && (
+        <AboutPipetteSlideout
+          pipetteId={attachedPipette.serialNumber}
+          pipetteName={pipetteDisplayName ?? attachedPipette.instrumentName}
+          firmwareVersion={attachedPipette.firmwareVersion}
+          isExpanded={showAboutPipetteSlideout}
+          onCloseClick={() => setShowAboutPipetteSlideout(false)}
+        />
+      )}
```
jerader: nit, to keep the same pattern for easier readability:

```suggestion
      {attachedPipette?.ok && showAboutPipetteSlideout ? (
        <AboutPipetteSlideout
          pipetteId={attachedPipette.serialNumber}
          pipetteName={pipetteDisplayName ?? attachedPipette.instrumentName}
          firmwareVersion={attachedPipette.firmwareVersion}
          isExpanded={showAboutPipetteSlideout}
          onCloseClick={() => setShowAboutPipetteSlideout(false)}
        />
      ) : null}
```

jerader (approved):
code lgtm, just left some cleanup comments

app/src/organisms/Devices/PipetteCard/FlexPipetteCard.tsx
smb2268: I think this needs to be explicit -- we don't want to enter this if statement if `attachedPipette` is null or undefined, only if it does exist and `ok` is `false`

PR #14704, opened 2024-03-20
EXEC-308 feat(api): Add robot extents per pipette mount to shared data

Laura-Danielle:
# Overview

Add robot extent data to shared data. Note this is in robot coordinates on purpose, we will convert in EXEC-309.

# Test Plan

None yet. This will need to be tested with EXEC-309.

# Changelog

- add new key in pipette geometry related to specific extents based on the attached pipettes for both OT2 and Flex.
- add relevant data to static pip configs

# Review requests

Check naming, data shape etc.

# Risk assessment

Low, not doing anything with the data yet


shared-data/pipette/schemas/2/pipetteGeometrySchema.json
```diff
@@ -32,6 +40,14 @@
       "description": "The path to a valid Opentrons shared schema relative to the shared-data directory, without its extension. For instance, #/pipette/schemas/2/pipetteGeometrySchema.json is a reference to this schema."
     },
     "nozzleOffset": { "$ref": "#/definitions/xyzArray" },
+    "robotPositions": {
+      "description": "The pipette specific robot extents which change depending on the mount.",
+      "type": "object",
+      "properties": {
+        "robotFrontLeftPosition": { "$ref": "#/definitions/mountDefinition" },
```
sanni-t, line 47: Suggestion: `robotFrontLeftCornerPosition`

shared-data/pipette/schemas/2/pipetteGeometrySchema.json
```diff
@@ -21,6 +21,14 @@
       "type": "string",
       "pattern": "[0-9]+",
       "description": "The name of a column of nozzles"
+    },
+    "mountDefinition": {
+      "type": "object",
+      "required": ["left", "right"],
```
sanni-t, line 27: Suggestion: `leftMount`, `rightMount`

It's not clear what 'left' and 'right' are until you check the schema. I think adding the 'mount' suffix makes this self-explanatory in the definitions and also reads better.

shared-data/pipette/schemas/2/pipetteGeometrySchema.json
```diff
@@ -21,6 +21,14 @@
       "type": "string",
       "pattern": "[0-9]+",
       "description": "The name of a column of nozzles"
+    },
+    "mountDefinition": {
```
sanni-t, line 25: Suggestion: `positionByMount`

shared-data/pipette/schemas/2/pipetteGeometrySchema.json
sanni-t: Or, if you prefer keeping this just `left' & 'right', then I think we should rename the position properties to be `..byMount`. As in, `robotHomePositionByMount`. 

I think I do like the `..byMount` naming better. Matches other property names we have like `..ByTipCount`..

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
```diff
@@ -40,6 +47,18 @@ class LoadedStaticPipetteData:
...
     back_left_corner_offset: Point
     front_right_corner_offset: Point
+    robot_home_min_position: robotPositionsDict
+    robot_front_left_max_position: robotPositionsDict
+
+
+def _convert_robot_positions_dict(
+    robot_position: pipette_definition.PipetteMountPositions,
+) -> robotPositionsDict:
+    """Internally helper function for converting a robot positions dict."""
```
sanni-t, line 57:
```suggestion
    """Internal helper function for converting a robot positions dict."""
```

shared-data/pipette/definitions/2/geometry/single_channel/p50/3_5.json
```diff
@@ -10,5 +10,12 @@
   "orderedColumns": [{ "key": "1", "orderedNozzles": ["A1"] }],
   "nozzleMap": {
     "A1": [-8.0, -22.0, -259.15]
+  },
+  "robotPositions": {
+    "robotHomePosition": { "left": [0.3, -4.3], "right": [-53.7, -4.3] },
+    "robotFrontLeftPosition": {
+      "left": [537.8, 401.515],
+      "right": [483.8, 401.515]
+    }
```
sanni-t, line 19: An xyz array requires minimum 3 items but these are only specifying x & y.
I think we can include the z in the definitions as well. It'll just be the mount z and hence same for all pipettes on both mounts of that robot. That info can be used to check for z extents as well (how high can a pipette go).

OR, if you think the z info is not useful, we can define a new `xyArray` definition

sanni-t (commented):
This looks good! I have some naming suggestions and a potential schema issue.

shared-data/pipette/definitions/2/geometry/single_channel/p50/1_5.json
```diff
@@ -10,5 +10,12 @@
   "orderedColumns": [{ "key": "1", "orderedNozzles": ["A1"] }],
   "nozzleMap": {
     "A1": [0.0, 0.0, 25.0]
+  },
+  "robotPositions": {
+    "robotHomePosition": { "left": [0.0, 0.0], "right": [-35, 0.0] },
```
sfoster1, line 15: This isn't right for OT-2s. On the OT-2, deck coordinates and machine coordinates are the same orientation and of similar origin. The origin is in the front left at the deck, and coordinates increase towards the right, the back, and up away from the deck. When an OT-2 is homed, its x smoothie coordinate is `+421`: https://github.com/Opentrons/SmoothiewareOT/blob/25e5cea9a652d2199a9d54be3677adf8332cda2e/src/config.default#L136-L137 . When an OT-2 is at its leftmost gantry extent, its x smoothie coordinate should be below zero (because the gantry can move farther than the slot).

shared-data/pipette/definitions/2/geometry/single_channel/p50/3_0.json
```diff
@@ -10,5 +10,12 @@
   "orderedColumns": [{ "key": "1", "orderedNozzles": ["A1"] }],
   "nozzleMap": {
     "A1": [-8.0, -22.0, -259.15]
+  },
+  "robotPositions": {
+    "robotHomePosition": { "left": [0.3, -4.3], "right": [-53.7, -4.3] },
```
sfoster1, line 15: On the flex, you have to be a little more precise about coordinate systems than just "machine coordinates". We track offsets for each mount here https://github.com/Opentrons/opentrons/blob/1ff87e89b83adf36ac3021c52282d55494971b29/api/src/opentrons/config/defaults_ot3.py#L74-L77 and use them pretty frequently. This code has to be more specific about which critical point it applies to: the carriage? the mount position of that pipette? the pipette nozzle? the union of all the motor controller position trackers? If it's the latter, why aren't all the back-right positions (0,0,0)?

shared-data/pipette/schemas/2/pipetteGeometrySchema.json
```diff
@@ -32,6 +40,14 @@
       "description": "The path to a valid Opentrons shared schema relative to the shared-data directory, without its extension. For instance, #/pipette/schemas/2/pipetteGeometrySchema.json is a reference to this schema."
     },
     "nozzleOffset": { "$ref": "#/definitions/xyzArray" },
+    "robotPositions": {
+      "description": "The pipette specific robot extents which change depending on the mount.",
+      "type": "object",
+      "properties": {
+        "robotFrontLeftPosition": { "$ref": "#/definitions/mountDefinition" },
+        "robotHomePosition": { "$ref": "#/definitions/mountDefinition" }
```
sfoster1, line 48: Can we call this something with `backRight` rather than home? I think we want to separate those semantics of homing vs position extents.

sfoster1 (changes requested):
These coordinates don't make sense universally, I think. The OT-2's machine coordinate system is very different from the flex's, and we now have enough layers of position offsets that we need to define things a little more tightly when we talk about "machine coordinates". 

There's a note on one of the ot-2 only pipettes that describes the problem of the different coordinate systems, and a note on one of the flex-only pipettes that describes the offset issue.

I don't quite get why, on either machine, the back right _machine coordinates_ would be anything other than the homing position if the machine coordinates we're talking about are the sum of the axis positions. That's exactly what the machine-coordinates homing position is: the sum of the axis positions when all axes are in the back right. (I do kind of support having these coordinates there, just for completeness's sake, but I don't see either why they wouldn't be the home coordinates or why they would vary between pipettes).

Similarly, I don't quite see why the front-left coordinates would vary between pipette mounts on the OT-2 at all if they're machine coordinates. The pipettes on an OT-2 are under a shroud, and the shroud is the thing that would collide with internal geometry, and it will do that regardless of what pipette is attached where. Ditto with the axis bounds. If these coordinates are the sum of the motor controller coordinates, what's the physical origin of motor coordinates at maximum front-left extent changing based on pipette mount?

On the flex, the pipette mounts are different and we definitely will see different front-left bounding box corner positions if there's a 96 vs other kinds of pipette, and possibly model to model or mount to mount on the other kinds of pipettes, so I'm glad we're capturing that.

I think fundamentally my problem is that if we're specifying stuff at the level of machine coordinates, we just can't treat the Flex and the OT-2 the same. They're fundamentally different at this level. Getting the same data shape and semantics of content on OT-2 pipettes and Flex pipettes can only result in having a bunch of excess or confusing data on at least one of them. Let's have different models for the two at the level of shared-data and let's get the shared-data content correct now. We can present the same data shape to the protocol engine across the two machines from the hardware controller.

Also, we should really nail down terminology about these things. If these are bounding box positions, they should say bounding box somewhere. If they're about physical extents, they should talk about physical concepts like "front left" and "back right", not semantic concepts like "home position".

shared-data/pipette/definitions/2/geometry/single_channel/p50/3_0.json
Laura-Danielle: The offsets linked here are more meant to convert the coordinate system for the Flex back into the front left position being 0,0,0 though. The idea I had with the values in shared data was to store them in "machine coordinates" and the convert the values to "deck coordinates" afterwards for EXEC-309.

Maybe it makes more sense for the coordinates saved in shared data to have front left position be ~ 0,0,0 but that feels like it's moving into deck coordinates territory. These values are the "physical" extents of a pipette based on constraints related to the mount it's attached to.

shared-data/pipette/definitions/2/geometry/single_channel/p50/3_5.json
Laura-Danielle: Decided to define an xyArray to make it clear that we're dealing in XY only (because we also have many different offsets for Z and we would probably end up ignoring the value in code/making it maybe more confusing).

Laura-Danielle:
Closing in favor of relying on data we have already for determining robot extents per pipette.

PR #14707, opened 2024-03-20
chore: new project abr-testing

sfoster1:
Add a new project for ABR test files so they do not have to adhere to the requirements that hardware-testing does, and vice versa.

### DIFF
--- a/.github/workflows/abr-testing-lint-test.yaml
+++ b/.github/workflows/abr-testing-lint-test.yaml
@@ -10,7 +10,7 @@ on:
       - 'scripts/**/*.mk'
       - 'scripts/**/*.py'
       - '.github/workflows/abr-testing-lint-test.yaml'
-      - '.github/actions/ython/**'
+      - '.github/actions/python/**'
     branches:
       - 'edge'
     tags-ignore:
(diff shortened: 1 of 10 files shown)


.github/workflows/abr-testing-lint-test.yaml
```diff
@@ -0,0 +1,59 @@
...
+name: 'abr-testing lint/test'
+on:
+  push:
+    paths:
+      - 'Makefile'
+      - 'abr-testing/**'
+      - 'scripts/**/*.mk'
+      - 'scripts/**/*.py'
+      - '.github/workflows/abr-testing-lint-test.yaml'
+      - '.github/actions/ython/**'
```
SyntaxColoring:
```suggestion
      - '.github/actions/python/**'
```

abr-testing/setup.py
SyntaxColoring, line 1: Should this have a setup.py if we don't intend on packaging and publishing it anywhere?

SyntaxColoring (approved):
LGTM when CI passes.

abr-testing/setup.py
sfoster1: yes because maybe we will in the future and it's easier. I want the uniformity

PR #14711, opened 2024-03-21
feat(robot-server): Use a BadRun when we cant load

sfoster1:
Up to now, if there's a run saved in the persistence layer that cannot be loaded - which is typically because it contains data from a version of the robot server or api package that whatever's currently running can't handle - we error when trying to retrieve it. That includes both a 500 error when trying to access that particular run, which clients can broadly handle, and a 500 error when trying to list all runs, which clients cannot. Without being able to list out all runs, there's no way for clients to find the problematic run - the IDs are UUIDs and cannot be enumerated - and remove it. The only recourse is to delete all the run storage.

A different way to handle this problem is to consider a "bad run", a run whose run metadata or engine state summary cannot be loaded, as a first class entity that can be returned from run access endpoints wherever a run could be, without an HTTP level error. This is done everywhere for consistency in this commit, though the argument could be made that it should only be done in the list-all-runs access and other endpoints should continue to error.

This bad run contains error information about the cause of the invalid data using a new enumerated error. The bad run will carry all the information that could be loaded - in effect, if the state summary is bad then the run metadata will still be present, and the ID should generally be accessible.

Closes EXEC-344

## Review requests
- [ ] normal stuff
- [ ] from an app perspective, does this seem reasonable to work with?

## Testing
- [x] Put it on some robots and see that we get some useful data

### DIFF
--- a/robot-server/robot_server/runs/router/base_router.py
+++ b/robot-server/robot_server/runs/router/base_router.py
@@ -36,7 +36,7 @@ from robot_server.protocols.router import ProtocolNotFound
 
 from ..run_models import RunNotFoundError
 from ..run_auto_deleter import RunAutoDeleter
-from ..run_models import Run, RunCreate, RunUpdate
+from ..run_models import Run, BadRun, RunCreate, RunUpdate
 from ..engine_store import EngineConflictError
 from ..run_data_manager import RunDataManager, RunNotCurrentError
 from ..dependencies import get_run_data_manager, get_run_auto_deleter
@@ -99,7 +99,7 @@ class AllRunsLinks(BaseModel):
 async def get_run_data_from_url(
     runId: str,
     run_data_manager: RunDataManager = Depends(get_run_data_manager),
-) -> Run:
+) -> Union[Run, BadRun]:
     """Get the data of a run.
 
     Args:
@@ -144,7 +144,7 @@ async def create_run(
     deck_configuration_store: DeckConfigurationStore = Depends(
         get_deck_configuration_store
     ),
-) -> PydanticResponse[SimpleBody[Run]]:
+) -> PydanticResponse[SimpleBody[Union[Run, BadRun]]]:
     """Create a new run.
 
     Arguments:
@@ -206,7 +206,7 @@ async def create_run(
         "Get a list of all active and inactive runs, in order from oldest to newest."
     ),
     responses={
-        status.HTTP_200_OK: {"model": MultiBody[Run, AllRunsLinks]},
+        status.HTTP_200_OK: {"model": MultiBody[Union[Run, BadRun], AllRunsLinks]},
     },
 )
 async def get_runs(
@@ -220,7 +220,7 @@ async def get_runs(
         ),
     ),
     run_data_manager: RunDataManager = Depends(get_run_data_manager),
-) -> PydanticResponse[MultiBody[Run, AllRunsLinks]]:
+) -> PydanticResponse[MultiBody[Union[Run, BadRun], AllRunsLinks]]:
     """Get all runs, in order from least-recently to most-recently created.
 
     Args:
@@ -248,13 +248,13 @@ async def get_runs(
     summary="Get a run",
     description="Get a specific run by its unique identifier.",
     responses={
-        status.HTTP_200_OK: {"model": SimpleBody[Run]},
+        status.HTTP_200_OK: {"model": SimpleBody[Union[Run, BadRun]]},
         status.HTTP_404_NOT_FOUND: {"model": ErrorBody[RunNotFound]},
     },
 )
 async def get_run(
     run_data: Run = Depends(get_run_data_from_url),
-) -> PydanticResponse[SimpleBody[Run]]:
+) -> PydanticResponse[SimpleBody[Union[Run, BadRun]]]:
     """Get a run by its ID.
 
     Args:
@@ -316,7 +316,7 @@ async def update_run(
     runId: str,
     request_body: RequestModel[RunUpdate],
     run_data_manager: RunDataManager = Depends(get_run_data_manager),
-) -> PydanticResponse[SimpleBody[Run]]:
+) -> PydanticResponse[SimpleBody[Union[Run, BadRun]]]:
     """Update a run by its ID.
 
     Args:
--- a/robot-server/robot_server/runs/run_data_manager.py
+++ b/robot-server/robot_server/runs/run_data_manager.py
@@ -22,7 +22,6 @@ from .run_store import RunResource, RunStore, BadRunResource, BadStateSummary
 from .run_models import Run, BadRun, RunLoadingError
 
 from opentrons.protocol_engine.types import DeckConfigurationType
-from opentrons.protocol_engine import ErrorOccurrence
 
 
 def _build_run(
--- a/robot-server/robot_server/runs/run_data_manager.py
+++ b/robot-server/robot_server/runs/run_data_manager.py
@@ -1,9 +1,9 @@
 """Manage current and historical run data."""
 from datetime import datetime
-from typing import List, Optional
+from typing import List, Optional, Union
 
 from opentrons_shared_data.labware.labware_definition import LabwareDefinition
-
+from opentrons_shared_data.errors.exceptions import InvalidStoredData, EnumeratedError
 from opentrons.protocol_engine import (
     EngineStatus,
     LabwareOffsetCreate,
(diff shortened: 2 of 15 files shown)


sfoster1:
with a robot with a run using a state from the future we get this:
```
{
      "id": "312a3e2c-5156-45dd-b3de-ea84ff6734b1",
      "dataError": {
        "id": "RunLoadingError",
        "title": "Run Loading Error",
        "detail": "There was no engine state data for this run.",
        "meta": {
          "type": "InvalidStoredData",
          "code": "4008",
          "message": "There was no engine state data for this run.",
          "detail": {},
          "wrapping": []
        },
        "errorCode": "4008"
      },
      "createdAt": "2024-03-19T19:08:58.821381+00:00",
      "status": "stopped",
      "current": false,
      "actions": [],
      "errors": [],
      "pipettes": [],
      "modules": [],
      "labware": [],
      "liquids": [],
      "labwareOffsets": [],
      "protocolId": "8aa13211-066a-4b80-b24f-bac669cff09a"
    }
```

robot-server/robot_server/runs/run_data_manager.py
```diff
@@ -18,44 +18,90 @@
...
+            errors=state.errors,
+            labware=state.labware,
+            labwareOffsets=state.labwareOffsets,
+            pipettes=state.pipettes,
+            modules=state.modules,
+            current=current,
+            completedAt=state.completedAt,
+            startedAt=state.startedAt,
+            liquids=state.liquids,
+        )
```
DerekMaggio: Take a look at this version of the logic. 
It's less nested, passes all the tests, and doesn't have a case where we get a logic error.

It does make the assumption, if you have no errors, you didn't have a bad run. Is that correct?

```suggestion

    errors: List[EnumeratedError] = []

    if isinstance(state_summary, BadStateSummary):
        state = StateSummary.construct(
            status=EngineStatus.STOPPED,
            errors=[],
            labware=[],
            labwareOffsets=[],
            pipettes=[],
            modules=[],
            liquids=[],
        )
        errors.append(state_summary.dataError)
    else:
        state = state_summary

    if not run_resource.ok:
        errors.append(run_resource.error)

    if len(errors) == 0: 
        return Run.construct(
            id=run_resource.run_id,
            protocolId=run_resource.protocol_id,
            createdAt=run_resource.created_at,
            actions=run_resource.actions,
            status=state_summary.status,
            errors=state_summary.errors,
            labware=state_summary.labware,
            labwareOffsets=state_summary.labwareOffsets,
            pipettes=state_summary.pipettes,
            modules=state_summary.modules,
            current=current,
            completedAt=state_summary.completedAt,
            startedAt=state_summary.startedAt,
            liquids=state_summary.liquids,
        )
    if len(errors) == 1:
        run_loading_error = RunLoadingError.from_exc(errors[0])
    else:
        run_loading_error = RunLoadingError.from_exc(
            InvalidStoredData(
                message=(
                    "Data on this run is not valid. The run may have been "
                    "created on a future software version."
                ),
                wrapping=errors,
            )
        )

    return BadRun.construct(
        dataError=run_loading_error,
        id=run_resource.run_id,
        protocolId=run_resource.protocol_id,
        createdAt=run_resource.created_at,
        actions=run_resource.actions,
        status=state.status,
        errors=state.errors,
        labware=state.labware,
        labwareOffsets=state.labwareOffsets,
        pipettes=state.pipettes,
        modules=state.modules,
        current=current,
        completedAt=state.completedAt,
        startedAt=state.startedAt,
        liquids=state.liquids,
    )
```

mjhuff (commented):
Nice! From an app perspective, this does seem reasonable to work with.

shared-data/python/opentrons_shared_data/errors/codes.py
```diff
@@ -85,6 +85,7 @@ class ErrorCodes(Enum):
     COMMAND_PARAMETER_LIMIT_VIOLATED = _code_from_dict_entry("4005")
     INVALID_PROTOCOL_DATA = _code_from_dict_entry("4006")
     API_MISCONFIGURATION = _code_from_dict_entry("4007")
+    INVALID_STORED_DATA = _code_from_dict_entry("4008")
```
SyntaxColoring, line 88: TY

robot-server/robot_server/runs/run_models.py
```diff
@@ -20,9 +20,20 @@
 )
 from opentrons_shared_data.errors import GeneralError
 from robot_server.service.json_api import ResourceModel
+from robot_server.errors.error_responses import ErrorDetails
 from .action_models import RunAction
 
 
+class RunLoadingError(ErrorDetails):
```
SyntaxColoring: Nit: I might call this "run data error" or "run storage error" or something. "Run loading error" sounds to me like an error from creating it.

robot-server/robot_server/runs/run_store.py
```diff
@@ -442,28 +492,48 @@ def _clear_caches(self) -> None:
...
+        actions = [
             RunAction(
                 id=action_row.id,
                 createdAt=action_row.created_at,
                 actionType=RunActionType(action_row.action_type),
             )
             for action_row in action_rows
-        ],
+        ]
+    except BaseException as be:
```
SyntaxColoring: I think we only want to catch regular `Exception`s. We don't want to accidentally convert `KeyboardInterrupt`s or `SystemExit`s or `asyncio.CancelledError`s into `BadRunResource`s, for example.

```suggestion
    except Exception as be:
```

robot-server/robot_server/runs/run_store.py
```diff
@@ -442,28 +492,48 @@ def _clear_caches(self) -> None:
...
+        return BadRunResource(
```
SyntaxColoring, line 519: It probably makes sense to log a warning here, right?

```suggestion
        _log.warning("Error reading actions for run ID {run_id}:", exc_info=True)
        return BadRunResource(
```

robot-server/robot_server/runs/run_store.py
```diff
@@ -296,11 +337,20 @@ def get_state_summary(self, run_id: str) -> Optional[StateSummary]:
...
                 if row.state_summary is not None
-                else None
+                else BadStateSummary(
+                    dataError=InvalidStoredData(
+                        message="There was no engine state data for this run."
+                    )
+                )
             )
         except ValidationError as e:
             log.warning(f"Error retrieving state summary for {run_id}: {e}")
```
SyntaxColoring: Not related to your changes, but this improves the warning in case the stringification of `e` doesn't include enough information to debug. (It doesn't for `KeyError`s, for example.)

```suggestion
            log.warning(f"Error retrieving state summary for {run_id}", exc_info=True)
```

robot-server/robot_server/runs/router/base_router.py
```diff
@@ -144,7 +144,7 @@ async def create_run(
     deck_configuration_store: DeckConfigurationStore = Depends(
         get_deck_configuration_store
     ),
-) -> PydanticResponse[SimpleBody[Run]]:
+) -> PydanticResponse[SimpleBody[Union[Run, BadRun]]]:
```
SyntaxColoring, line 147: Technically, this is a breaking HTTP API change, right? A client that was doing `response.data.actions.length`, for example, will now error. I guess the right way to deal with that would be:

* For `Opentrons-Version` ≥ *n*, Return `BadRun`s as you are now.
* For `Opentrons-Version` < *n*, simply filter out `BadRun`s.

I'm also happy for this to be deemed not worthwhile to worry about.

robot-server/robot_server/runs/router/base_router.py
```diff
@@ -248,13 +248,13 @@ async def get_runs(
...
     responses={
-        status.HTTP_200_OK: {"model": SimpleBody[Run]},
+        status.HTTP_200_OK: {"model": SimpleBody[Union[Run, BadRun]]},
         status.HTTP_404_NOT_FOUND: {"model": ErrorBody[RunNotFound]},
     },
 )
 async def get_run(
     run_data: Run = Depends(get_run_data_from_url),
-) -> PydanticResponse[SimpleBody[Run]]:
+) -> PydanticResponse[SimpleBody[Union[Run, BadRun]]]:
```
SyntaxColoring, line 257: > though the argument could be made that it should only be done in the list-all-runs access and other endpoints should continue to error.

Yeah, I'm of that opinion. Can `GET /runs/{id}` return an HTTP 500 whenever it returns a `BadRun`?

My thinking is that we have a lot of Python integration test code (and maybe also JS client code) that does stuff like:

```python
run_response = client.get_run()
run_response.raise_if_not_http_ok()
```

And I think it's more correct for it *not* to proceed as normal in these cases.

SyntaxColoring (approved):

robot-server/robot_server/runs/router/base_router.py
mjhuff: For what it's worth, the client side doesn't currently have any conditional behavior based on an HTTP error for `GET /runs/{id}`, but there could be something more implicit that I'm missing. 

I don't have any strong convictions on this either way. I think having the "all runs" and "this run" resources behave uniformly would be preferred, but if it's going to greatly interfere with existing code (or at least make us think something insidious could break), then sure, I'm of the opinion we can continue to error.

robot-server/robot_server/runs/run_data_manager.py
sfoster1: I think I'll unnest it but I really like having the no-error case first

robot-server/robot_server/runs/run_store.py
sfoster1: Sure!

robot-server/robot_server/runs/router/base_router.py
sfoster1: The integration test coverage is enough that I did in fact have to change some behavior, and I think that's good.

I do think that semantically, what we're doing here is adding this new kind of resource, and getting a resource successfully - which this now counts as - gives you a 200. I think we move "there was invalid run data" problem out of the realm of an HTTP API concern and into the system that API provides access to and modeling of.

robot-server/robot_server/runs/router/base_router.py
sfoster1: To me that's an argument for making the actions not `Optional[]` and having it be an empty list instead

robot-server/robot_server/runs/router/base_router.py
sfoster1: > My thinking is that we have a lot of Python integration test code (and maybe also JS client code) that does stuff like:
> 
> ```python
> run_response = client.get_run()
> run_response.raise_if_not_http_ok()
> ```


I'm also quite happy to change all this stuff Where is it?

robot-server/robot_server/runs/router/base_router.py
SyntaxColoring: Yeah, that works in this case. It wouldn't work if there were ever a problem reading a scalar like `createdAt`, but that's not a problem that we have right now.

robot-server/robot_server/runs/router/base_router.py
SyntaxColoring: > I'm also quite happy to change all this stuff Where is it?

At least some of the uses of [`RobotClient.get_run()`](https://github.com/Opentrons/opentrons/blob/4221574fdbbcc14de773e9c3ef554059f16e0152/robot-server/tests/integration/robot_client.py#L183-L187) (haven't looked at them all yet, sorry)

mjhuff (approved):
Types look good with passing CI.

PR #14712, opened 2024-03-21
refactor(modules): return has update for unexpected module fw filename

caila-marashaj:
## Overview
When we flash local firmware changes to a module using an ST-Link, the file produced will have the usual firmware filename for the module, except with a commit hash amended in the middle of it. 

Currently, having the filename altered like this causes the robot server not to recognize that a module is present at all. Instead, this change allows the app to reflect that the module is there and offer an available update.

## Changelog

- catch and handle an `InvalidVersion` error for both current and available device versions

### DIFF
--- a/api/src/opentrons/hardware_control/modules/mod_abc.py
+++ b/api/src/opentrons/hardware_control/modules/mod_abc.py
@@ -4,7 +4,7 @@ import logging
 import re
 from pkg_resources import parse_version
 from typing import ClassVar, Mapping, Optional, cast, TypeVar
-
+from packaging.version import InvalidVersion
 from opentrons.config import IS_ROBOT, ROBOT_FIRMWARE_DIR
 from opentrons.drivers.rpi_drivers.types import USBPort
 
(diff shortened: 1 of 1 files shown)


sfoster1 (approved):
Looks good to me!

api/src/opentrons/hardware_control/modules/mod_abc.py
```diff
@@ -90,6 +90,8 @@ def has_available_update(self) -> bool:
             device_version = parse_version(self.device_info["version"])
             available_version = parse_version(self._bundled_fw.version)
             return cast(bool, available_version > device_version)
+        elif self.device_info and not self._bundled_fw:
+            return True
```
SyntaxColoring: 1. I'm not really familiar with how module updates work, but would it not be better to update the regex in `get_bundled_fw()` to understand the commit hash suffix? If you click the "update module" button in the Opentrons App, doesn't something need to find that weirdly-named update file, anyway?
2. If we do keep this, can we leave a comment explaining that it's a workaround for dev builds?

SyntaxColoring (approved):

SyntaxColoring (approved):
LGTM, thanks!

sfoster1 (approved):
lgtm!

PR #14713, opened 2024-03-21
feat(protocol-designer, step-generation, shared-data): migrate to using pipette v2 schema

jerader:
closes AUTH-52, AUTH-231

# Overview

Migrate from using `getPipetteNameSpecs` to `getPipetteV2Specs` in PD. This change is necessary both for pipette collision warnings and so flow rates are accurate/reflect the api

There are a few followup action items:
1. add the filtered tips to the default tips array in the pipette definitions https://opentrons.atlassian.net/browse/AUTH-232
2. @sanni-t to investigate why there are no `min`s and `max`s for the flowrates in the v2 schema definitions - this information is needed for PD to accurately have a range defining custom flow rates.
3. determin if the util should return the latest version or not see https://opentrons.atlassian.net/browse/AUTH-243

# Test Plan

Smoke test PD to make sure it works as expected. In particular, test these areas:
1. the create file wizard
2. adding a transfer and mix step
4. edit the transfer/mix step's advance settings by changing the flow rates

# Changelog

- migrate all instances of `getPipetteNameSpecs` to `getPipetteV2Specs`
- create util for returning the flow rates
- fix tests
-
# Review requests

see test plan

# Risk assessment

low


sanni-t:
The absence of min & max values just seems to be an oversight and we should add them into the v2 definitions in a follow-up. Let me know if you need any help adding them.

protocol-designer/src/components/modals/CreateFileWizard/__tests__/PipetteTipsTile.test.tsx
```diff
@@ -98,10 +98,10 @@ describe('PipetteTipsTile', () => {
   afterEach(() => {
     cleanup()
   })
-  it('renders default tiprack options for 50uL flex pipette and btn ctas work', () => {
+  it('renders default tiprack options for 1000uL flex pipette and btn ctas work', () => {
     render(props)
-    screen.getByText('Choose tips for Flex 1-Channel 50 μL')
-    screen.getByText('mock EquipmentOption')
+    screen.getByText('Choose tips for Flex 1-Channel 1000 μL')
+    screen.getAllByText('mock EquipmentOption')
```
koji, line 104: need to use `AllByText` instead of `getByText`?

koji (approved):
The changes look good to me.

protocol-designer/src/components/modals/CreateFileWizard/__tests__/PipetteTipsTile.test.tsx
jerader: yeah with the change in pipette, more equipment options (tiprack options) rendered! had to change it a bit since filtered tips aren't included in the pipette's default tips array right now.

PR #14714, opened 2024-03-21
fix(protocol-designer): display correct tip position offset

shlokamin:
# Overview
This PR fixes a bug where PD was rendering the wrong tip position field.

closes RESC-223

# Test Plan
- Verified that when updating tip position in the move liquid form the form field updates
- Verified that when updating tip position in the mix form the form field updates
- Verified that when updating tip position in the batch edit form the form field updates

# Changelog

- Display correct tip position offset

# Review requests

See test plan

# Risk assessment
Low


ncdiehl11 (approved):
tested and looks good 👍

PR #14715, opened 2024-03-21
refactor(components, app): move StyledText from app to components

koji:
# Overview
move `StyledText` from app to components for Chip component update+unification.
https://opentrons.slack.com/archives/C06M6R72UDS/p1710857517041999?thread_ts=1710857252.462119&cid=C06M6R72UDS

Chip component uses `StyledText` and components has been released as a npm package.

close AUTH-228

# Test Plan
- make teardown-js && make setup-js`
- Desktop/ODD displays text properly

# Changelog
- app/src/atom/text -> components/src/atom/StyledText
- update all import paths of StyledText
- fix a few typos that I founded during updating the paths

# Review requests

# Risk assessment
mid since this PR updated 389 files


jerader (approved):
nice! I did some smoke testing in app/odd/storybook and things look good. Once this merges, I think you should post it in the app and ui channel so all frontend devs are aware. This refactor also means that PD can use `StyledText` instead of having its own version, so that's awesome.

PR #14716, opened 2024-03-22
test(robot-server): Fail tests quickly when `GET /health` returns 500

SyntaxColoring:
# Overview

@CaseyBatten ran into a confusing thing with robot-server's integration tests. If we introduce a bug in robot-server's database migration code that makes it fail to start, many of our integration tests hang for a very long time before eventually timing out, obscuring the bug.

This fixes our test code so it fails immediately when robot-server's startup fails with a fatal error.

# Test Plan

I've cherry-picked this atop a known-bad commit and confirmed that it makes the tests fail quickly with the expected "500 internal server error" message.

# Changelog

* Teach the `RobotClient` test helper the difference between an HTTP 503 "busy" and other HTTP error codes. If it's busy, keep waiting. If it's any other kind of error, propagate it.
* Similarly, remove `RobotClient.wait_until_dead()`. Some tests want to make sure that there's no existing server before starting a new one, and that's fine, but they can do that by just checking `RobotClient.dead()` once. They don't need to keep polling; all that does is forestall the inevitable failure.
* While we're here, have `RobotClient.wait_until_alive()` raise a timeout error instead of returning `False`, so callers don't have to remember to `assert` it.

# Review requests

None in particular.

# Risk assessment

Low. We're just changing some test helpers.

### DIFF
--- a/robot-server/tests/integration/http_api/persistence/test_compatibility.py
+++ b/robot-server/tests/integration/http_api/persistence/test_compatibility.py
@@ -152,9 +152,7 @@ async def test_protocols_analyses_and_runs_available_from_older_persistence_dir(
     async with RobotClient.make(
         base_url=f"http://localhost:{_PORT}", version="*"
     ) as robot_client:
-        assert (
-            await robot_client.wait_until_dead()
-        ), "Dev Robot is running and must not be."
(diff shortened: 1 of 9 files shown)


sfoster1 (approved):
Nice, looks like a good idea.

PR #14718, opened 2024-03-22
feat(api, robot-server): add bool param type, add 'type' field to all param definitions

sanni-t:
Addresses AUTH-125

# Overview

Updates the shape of the parameter definition models in analysis & run responses according to the agreed upon structure in AUTH-125.

# Review requests

- make sure that the shape is agreeable to both the API & HTTP client interfaces.

# Risk assessment

None. Just data model changes


jbleon95 (approved):
🙌

PR #14720, opened 2024-03-22
refactor(app): create utils to format value for rtp

koji:
# Overview
create `formatRunTimeParameterValue` to format rtp value and replace a local function with that.

I noticed a small bug I made. On the UI, the following code looks fine at first glance.
However, if the data doesn't have `suffix`, the return has an unnecessary space.
ex
value: 6 without `suffix` -> `'6 '` (including unnecessary space)  but for this case we should return `'6'`

Screenshot 2024-03-22 at 8 00 00 PM

```ts
  case 'int':
  case 'float':
    return `${defaultValue.toString()} ${suffix}`
```

close AUTH-236

# Test Plan
Desktop app
- protocol details parameters tab
- protocol run setup parameters tab

ODD
- protocol details parameters tab
- protocol setup parameters screen
- view-only parameters screen

# Changelog
- create a utils file and add `formatRunTimeParameterValue` + its test

# Review requests
If there is a better place this util function, please let me know.

# Risk assessment
low


app/src/organisms/ProtocolDetails/ProtocolParameters/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+      ? runTimeParameter.suffix
+      : null
+  switch (type) {
+    case 'int':
+    case 'float':
+      return suffix !== null
+        ? `${defaultValue.toString()} ${suffix}`
+        : defaultValue.toString()
+    case 'boolean':
+      return Boolean(defaultValue) ? 'On' : 'Off'
```
jerader: maybe the util should take in `t` as a prop so `On` and `Off` can be a in i18n?

jerader (approved):
thanks for creating this util! Left a suggestion about adding `t` as a prop

app/src/organisms/ProtocolDetails/ProtocolParameters/utils.ts
koji: updated the util function.

PR #14722, opened 2024-03-25
feat(app): sort robot lists alphabetically

mjhuff:
Closes [EXEC-350](https://opentrons.atlassian.net/browse/EXEC-350)

# Overview

Back during a standup in January, appui agreed that there is no current logic when presenting lists of robots on the Devices tab, sending protocols slideouts, etc. Let's sort robots alphabetically. Doing this via selectors is best because:

1. Selectors are memoized.
2. Sorting already filtered lists of robots means we aren't computing this often.
3. Doing this on every slideout, page, etc. doesn't make much sense, since when we use these selectors, we are using them to display lists of robots.

Screenshot 2024-03-25 at 9 51 27 AM

Screenshot 2024-03-25 at 9 51 12 AM

# Test Plan
- Open the app. Observe robots are sorted on the Devices page and robot slideouts. 

# Changelog
- Robots are sorted alphabetically on the Devices page and various slideouts. 

# Risk assessment
low

[EXEC-350]: https://opentrons.atlassian.net/browse/EXEC-350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/redux/discovery/selectors.ts
+++ b/app/src/redux/discovery/selectors.ts
@@ -3,6 +3,7 @@ import concat from 'lodash/concat'
 import head from 'lodash/head'
 import isEqual from 'lodash/isEqual'
 import find from 'lodash/find'
+import orderBy from 'lodash/orderBy'
 import { createSelector, createSelectorCreator, defaultMemoize } from 'reselect'
 import semver from 'semver'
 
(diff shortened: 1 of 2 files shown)


app/src/redux/discovery/selectors.ts
```diff
@@ -146,34 +147,59 @@ export const getDiscoveredRobots: (
     })
   }
 )
-
+// TOME: Just test this works on the slideouts, devices page, etc, and you're good to go! I think I'd rather just sort these compound selectors
+// since that's less TC intensive than sorting ALL the discovered robots each time.
```
SyntaxColoring: Comment left over from development?

app/src/redux/discovery/selectors.ts
```diff
@@ -146,34 +147,59 @@ export const getDiscoveredRobots: (
...
 export const getConnectableRobots: GetConnectableRobots = createSelector(
   getDiscoveredRobots,
-  robots => robots.flatMap(r => (r.status === CONNECTABLE ? [r] : []))
+  robots =>
+    orderBy(
+      robots.flatMap(r => (r.status === CONNECTABLE ? [r] : [])),
+      [robot => robot.displayName.toLowerCase()],
+      ['asc']
+    )
```
SyntaxColoring, line 158: > Selectors are memoized.

For my edification, what defines that memoization? It's smart enough to compare arrays by semantic equality instead of by reference equality?

SyntaxColoring (approved):
Yay!

app/src/redux/discovery/selectors.ts
mjhuff: Ah yep, woops. Thank you!

app/src/redux/discovery/selectors.ts
mjhuff: Yep exactly. We get semantic equality with Reselect via `createSelector` but not redux. In this case though, we default back to shallow equality, since we don't actually specify an equality function (but that's sufficient for our MDNS obtained robots).

app/src/redux/discovery/selectors.ts
koji, line 1: nit maybe creating a new function since orderBy is used by multiple functions. 🤔

```ts
const sortRobots = (robots) => orderBy(
  robots,
  [robot => robot.displayName.toLowerCase()],
  ['asc']
)

export const getReachableRobots: GetReachableRobots = createSelector(
  getDiscoveredRobots,
  robots => sortRobots(robots.flatMap(r => (r.status === REACHABLE ? [r] : [])))
)

export const getUnreachableRobots: GetUnreachableRobots = createSelector(
  getDiscoveredRobots,
  robots => sortRobots(robots.flatMap(r => (r.status === UNREACHABLE ? [r] : [])))
)


```

koji (approved):
left a comment but it is optional.
the changes look good to me.

app/src/redux/discovery/selectors.ts
mjhuff: Good thinking. Personally, I think the typing gets a bit convoluted when trying to apply this to `getAllRobots` and `getViewableRobots`. I think this is pretty borderline...I'll probably just keep it as is.

PR #14723, opened 2024-03-25
fix(robot-server): Update tests to properly check new bad run records

SyntaxColoring:
# Overview

Follow-ups for https://github.com/Opentrons/opentrons/pull/14711#discussion_r1534599540.

#14711 added safer error propagation for when robot-server encounters bad stored run data. As part of that, if it finds a run where the `state_summary` SQL column is `NULL`, it treats that as bad data and propagates the error to HTTP clients.

When you restart the robot while there is an active run, then no state summary will be inserted (this only happens when the run is ended and the state moves from engine to sql) and the run will be bad. 

We say that this is in fact a bad run because to the client, there is no distinction between state summary and run. A run with an empty state summary does not have correct data and does not represent what occurred. 

Add a regression test to make sure this is how we handle runs that did not have state summaries persisted.

# Testing
- [x] create a run with this branch on a flex and restart the flex (or kill the robot server process - this isn't about the details of when things are written to disk, just the lifetime of the data here) and see that the run is now bad

### DIFF
--- a/robot-server/tests/integration/http_api/persistence/test_compatibility.py
+++ b/robot-server/tests/integration/http_api/persistence/test_compatibility.py
@@ -198,7 +198,8 @@ async def test_protocols_analyses_and_runs_available_from_older_persistence_dir(
             assert all_run_ids == [r.id for r in snapshot.expected_runs]
 
             for expected_run in snapshot.expected_runs:
-                await robot_client.get_run(run_id=expected_run.id)
+                run = (await robot_client.get_run(run_id=expected_run.id)).json()
+                assert run["data"].get("dataError") is None
 
                 all_command_summaries = (
                     await robot_client.get_run_commands(
--- a/robot-server/tests/integration/http_api/persistence/test_compatibility.py
+++ b/robot-server/tests/integration/http_api/persistence/test_compatibility.py
@@ -1,8 +1,10 @@
+from __future__ import annotations
+
 from dataclasses import dataclass, field
 from pathlib import Path
 from shutil import copytree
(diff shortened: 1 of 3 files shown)


sfoster1:
Do we actually want to treat these differently? Is there a difference? Like, this is in fact a run that could not have its data loaded and is in fact bad.

TamarZanzouri (approved):
LGTM once CI passes

robot-server/tests/integration/http_api/persistence/test_compatibility.py
```diff
@@ -198,7 +198,8 @@ async def test_protocols_analyses_and_runs_available_from_older_persistence_dir(
             assert all_run_ids == [r.id for r in snapshot.expected_runs]
 
             for expected_run in snapshot.expected_runs:
-                await robot_client.get_run(run_id=expected_run.id)
+                run = (await robot_client.get_run(run_id=expected_run.id)).json()
+                assert run["data"].get("dataError") is None
```
SyntaxColoring: OK, so this is failing because some of the old runs are now marked bad.

I think we want to test that:

1. If bad runs "existed before," the migration should preserve them.
2. The migration should not introduce any new bad runs.

I would do this by introducing an `expected_bad: bool = False` attribute to the `Run` test helper dataclass.

To figure out which run IDs, specifically, are expected to be bad...I don't have a great answer other than temporarily running the test with a `print()` in place of the `assert`, or something like that.

SyntaxColoring (commented):
I can't approve this because I originated the PR, but LGTM. Thank you!

PR #14724, opened 2024-03-25
refactor(app-shell-odd): Utilize robot-server unsubscribe flags

mjhuff:
Closes [EXEC-319](https://opentrons.atlassian.net/browse/EXEC-319)

# Overview

This is the app-shell-odd equivalent of #14648. It's similar to the app-shell logic, but significantly simpler, since we don't have to manage multiple robots, worry about localhost port blocking, and multiple IPs per robot. The real change lies in the initial connect and final disconnect on app shutdown. Otherwise, the changes are primarily in the ConnectionStore. Because the app no longer utilizes unsubscribe actions in any capacity, we can safely remove those references. 

# Test Plan
- Smoke test the ODD, ensuring MQTT works.
- After the end of the protocol (or at the beginning of a new protocol run, I don't remember which), you should see a log message about unsubscribing from the old protocol run.

# Risk assessment
medium-ish, like the desktop app refactor.

[EXEC-319]: https://opentrons.atlassian.net/browse/EXEC-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Looks good to me!

PR #14725, opened 2024-03-25
refactor(api): only ignore stalls for downward portion of force pickup routine

caila-marashaj:
## Overview
During force tip-pickup, the z stage ignores stalls for all parts of the movement, including when it retracts back up towards the home position. This is believed to be the cause of some 'unknown position' problems popping in testing, so this changes the code to only ignore stalls during the parts of the routine that are actually pushing into the tiprack.

## Changelog
- make ignoring stalls during the pickup routine conditional on the direction of movement
- change the `target_down` variable to `target` for readability purposes


ahiuchingau (approved):
Nice - LGTM once tested on the FLEX!

sfoster1 (approved):
Okay, good once tested on a flex

PR #14726, opened 2024-03-25
refactor(api): Split UpdateCommandAction

SyntaxColoring:
# Overview

Closes EXEC-340.

# Changelog

* Delete the monolithic `UpdateCommandAction` and replace it with new, distinct `RunCommandAction` and `SucceedCommandAction` command transitions. So we end up with this:

  ```mermaid
  stateDiagram-v2
      [*] --> queued: QueueCommandAction
      queued --> running: RunCommandAction
      running --> succeeded: SucceedCommandAction
      running --> failed: FailCommandAction
  ```

  Compare with [the old state diagram](https://gist.github.com/SyntaxColoring/34cbcd8d68c2f2bad1be1dee3e75913a), where you could use `UpdateCommandAction` to jump around and skip over states.

* Use `assert`s to enforce the preconditions of the state transitions. (For example, you can't run a command when a different one is already running.)

* Make sure all completed commands have a `notes` list. If there are no notes, make it an empty list. Formerly, `notes` would not be present (`None` in Python, omitted in JSON) if there were no notes. Now, it's sort of like `completedAt`, where `None` means "hasn't happened yet."

  This entails a million billion test updates, unfortunately.

  This seemed like it would simplify things, but in hindsight, it doesn't make a big difference. I'm happy to undo this if anyone wants.

* Update the `LegacyContextPlugin`. This was the really annoying part. Before, it emitted `UpdateCommandAction`s to created `succeeded` and `failed` commands, skipping over `queued` and `running`. Now, because we're being stricter, it has to explicitly transition each command through `queued` and `running`.

  This bumps into a couple of ugly scary things, which I'll describe in separate threads below.

# Test plan

* [x] Make sure legacy OT-2 protocols (PAPIv≤2.13, JSONv≤5) still work.
    * [x] Make sure pausing and playing still works.
    * [x] Make sure the run progress monitor in the Opentrons App still tracks the current step and shows the predicted future steps.
    * [x] Make sure the robot still homes before the protocol begins.
* [x] Make sure Labware Position Check (a kind of live HTTP run) still works.
* [x] Make sure more modern protocols (PAPIv≥2.14, JSONv≥6) still work.

# Review requests

See my review comments.

This diff got huge, sorry. Most of the files only had trivial changes.

# Risk analysis

Medium. This touches some sensitive and messy code having to do with legacy (PAPIv≤2.13, JSONv≤5) protocols. Other than that, the changes are simple.


api/src/opentrons/protocol_runner/legacy_context_plugin.py
```diff
@@ -119,26 +127,34 @@ def _handle_legacy_command(self, command: LegacyCommand) -> None:
...
+            # theoretically active, which means it's constantly watching in the
+            # background to execute any commands that it finds `queued`.
+            #
+            # For example, one of these action batches will often want to
+            # instantaneously create a running command by having a queue action
+            # immediately followed by a run action. We cannot let the
+            # ProtocolEngine's background task see the command in the `queued` state,
+            # or it will try to execute it, which the legacy protocol is already doing.
+            for action in action_batch:
+                self.dispatch(action)
```
SyntaxColoring, line 160: Here is ugly scary thing number 1.

The `LegacyContextPlugin` injects fake commands into a `ProtocolEngine` to mirror a legacy protocol's activity. Formerly, it did this by skipping the `queued` state and going straight to `running`. Now, that's not a valid state transition, and it has to explicitly pass through `queued`. That introduces a concurrency hazard if the `ProtocolEngine` ever finds the command in the `queued` state.

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -190,17 +190,21 @@ async def load(
...
+            await self._protocol_engine.add_and_execute_command(
+                request=initial_home_command
+            )
+            await self._legacy_executor.execute(
+                protocol=protocol,
+                context=context,
+                run_time_param_values=run_time_param_values,
+            )
+
+        self._task_queue.set_run_func(run_func)
```
SyntaxColoring, line 207: Here is ugly scary thing number 2.

Formerly, running a legacy protocol worked like this:

1. We enqueue (but do not execute) the initial home command into the `ProtocolEngine`.
2. Someone plays the run.
3. Concurrently (???):
    1. The legacy protocol starts controlling the robot the way they always have, by directly accessing the hardware API. The `LegacyContextPlugin` monitors its activity and injects fake commands into the `ProtocolEngine` for the sake of providing a uniform Protocol Engine façade to higher-level code.
    2. Protocol Engine finds the queued home command and executes it.
    3. It's unclear to me how this concurrency didn't cause mayhem, and how the home command always won in practice.

This started raising `AssertionError`s with this PR because we're being stricter about state transitions now. This sequence was happening:

1. Enqueuing the initial home command does a `QueueCommandAction`.
2. Playing the run does a `PlayAction`
3. Protocol Engine beginning to execute the initial home command does a `RunCommandAction`.
4. `FinishAction` (unclear to me why this happened here)
5. The `LegacyContextPlugin` maps the protocol's first `load_labware()` (or whatever) call into a `QueueCommandAction` and `RunCommandAction`.
    * This raises  an `AssertionError` because the initial home command is still running, and we're now `assert`ing that only one command can be running at a time.

My solution here is to let the initial home command finish before letting the legacy Python protocol take the reins. For non-legacy Python protocols, I believe this is unnecessary but harmless. Alternatively, we could remove the `assert`, but...I don't see a legitimate reason to do this.

api/src/opentrons/protocol_runner/legacy_context_plugin.py
SyntaxColoring: If we want to fix this, I think we should do it by separating legacy protocol execution further from Protocol Engine, instead of doing this plugin thing.

sfoster1: That would be a pretty big change, yeah. What about if we added a new command intent called like `FOLLOW` or something? and have the execute machinery no-op it through the three states? Then we wouldn't have to have either a plugin or a major change to how we execute legacy protocols.

api/src/opentrons/protocol_runner/protocol_runner.py
sfoster1: I think this change makes sense.

sfoster1:
After looking at it I think these are good concurrency hazards to fix and I think we can definitely do it especially if we're open to alternative means of fixing things. The home command looks pretty solved, and I think with the alternate intent we can handle the legacy python pretty well.

api/tests/opentrons/protocol_engine/execution/test_command_executor.py
```diff
@@ -519,144 +445,20 @@ def _ImplementationCls(self) -> Type[_TestCommandImpl]:
...
-        action_dispatcher.dispatch(
-            UpdateCommandAction(private_result=None, command=running_command)
-        ),
         action_dispatcher.dispatch(
-            UpdateCommandAction(private_result=None, command=completed_command)
-        ),
-    )
-
-
-async def test_executor_forwards_notes_on_command_failure(
```
SyntaxColoring, line 592: The deletions of `test_executor_forwards_notes_on_command_success` and `test_executor_forwards_notes_on_command_failure` are because I'm merging them into `test_execute` and `test_execute_raises_protocol_engine_error`.

api/src/opentrons/protocol_runner/legacy_context_plugin.py
SyntaxColoring: Oh oh, wait, maybe we don't have to do either of those. If @TamarZanzouri hoists out `ProtocolEngine`'s background executor, that can solve it too.

api/src/opentrons/protocol_engine/actions/actions.py
SyntaxColoring, line 1: Meaningful changes in this file.

api/src/opentrons/protocol_engine/execution/command_executor.py
SyntaxColoring, line 1: Meaningful changes in this file.

api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -121,21 +122,42 @@ class QueueCommandAction:
...
+
+
+@dataclass(frozen=True)
+class SucceedCommandAction:
+    """Mark a given command as succeeded.
+
+    At the time of dispatching this action, the command must be running.
+    """
 
     command: Command
```
sfoster1, line 143: when `UpdateCommandAction` was the everything-doer, having the command embedded made sense, but now having it in `SucceedCommandAction` looks a little weirder amongst all its siblings that only reference the id. Does it maybe make sense to have `SucceedCommandAction` carry `notes: List[CommandNote]`, `result: CommandResult`, `private_result: CommandPrivateResult` independently?

api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -121,21 +122,42 @@ class QueueCommandAction:
...
 
 @dataclass(frozen=True)
 class FailCommandAction:
-    """Mark a given command as failed."""
+    """Mark a given command as failed.
+
+    At the time of dispatching this action, the command must be running.
+    """
 
     command_id: str
```
sfoster1, line 156: if we don't want to make `SucceedCommandAction` carry everything independently, maybe we should make `FailCommandAction` embed the command too. Both of these actions are going to cause pretty equivalent modifications to the command dataclass anyway.

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -132,29 +136,24 @@ async def execute(self, command_id: str) -> None:
...
         self._action_dispatcher.dispatch(
-            UpdateCommandAction(command=running_command, private_result=None)
+            RunCommandAction(command_id=queued_command.id, started_at=started_at)
         )
+        running_command = self._state_store.commands.get(queued_command.id)
 
         try:
             log.debug(
-                f"Executing {command.id}, {command.commandType}, {command.params}"
+                f"Executing {queued_command.id}, {queued_command.commandType}, {queued_command.params}"
```
sfoster1: kind of a nit, but to make this easier on readers can we make the flow of this function cleanly go from `queued_command` to `running_command` to `completed_command` in terms of which variable we use to get information? So for instance, from this line until the creation of `completed_command`, we'd reference `running_command` in the `command_impl.execute` calls and the logging calls. They should be exactly equivalent in terms of the data we're accessing, right?

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -163,24 +162,14 @@ async def execute(self, command_id: str) -> None:
...
-                    )
-                )
 
             self._action_dispatcher.dispatch(
                 FailCommandAction(
                     error=error,
                     command_id=command_id,
                     error_id=self._model_utils.generate_id(),
                     failed_at=self._model_utils.get_timestamp(),
+                    notes=note_tracker.get_notes(),
```
sfoster1, line 172: these notes changes means that command notes from before the beginning of execution will be lost. this might be fine, but we should note (hehe) it somewhere

api/src/opentrons/protocol_runner/legacy_command_mapper.py
```diff
@@ -410,7 +443,11 @@ def _build_liquid_handling_command(
                 # or aspirate() with a volume of 0, which behaves roughly like
                 # move_to(). Protocol Engine aspirate and dispense commands must have
                 # volume > 0, so we can't map into those.
-                return pe_commands.MoveToWell.construct(
+                #
+                # TODO(mm, 2024-03-22): I don't think this has been true since
+                # https://github.com/Opentrons/opentrons/pull/14211. Can we just use
+                # aspirate and dispense commands now?
```
sfoster1, line 449: don't see why not, but maybe in a followup

sfoster1 (approved):
Looks good in structure with caveats inline.

I'm a little curious about the changes in the diff to add `notes: []` everywhere. I don't think you made that value non optional in the command dataclass here - why add all that stuff but keep it optional?

api/src/opentrons/protocol_engine/actions/actions.py
TamarZanzouri: I agree with @sfoster1 we might want to add these to the `SucceedCommandAction`. I agree `UpdateCommandAction` had too many responsibilities

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -242,7 +242,13 @@ async def add_and_execute_command(
         await self.wait_for_command(command.id)
         return self._state_store.commands.get(command.id)
 
-    def estop(self, maintenance_run: bool) -> None:
+    def estop(
+        self,
+        # TODO(mm, 2024-03-26): Maintenance runs are a robot-server concept that
+        # ProtocolEngine should not have to know about. Can this be simplified or
+        # defined in other terms?
+        maintenance_run: bool,
```
TamarZanzouri, line 250: true!

api/src/opentrons/protocol_engine/actions/actions.py
SyntaxColoring: Yep, that makes sense.

I’m pretty sure some of the states stores are reading `command.params`, too. So to replace that, we’d either need to add `params` to `CompleteCommandAction`, or give every state store access to the `CommandView` so they can retrieve the params themselves.

SyntaxColoring:
> I'm a little curious about the changes in the diff to add `notes: []` everywhere. I don't think you made that value non optional in the command dataclass here - why add all that stuff but keep it optional?

The idea was to keep it `None`/omitted until the command completes, at which point it becomes the real final list.

I don’t know if this was a good idea.

TamarZanzouri (approved):
Changes make sense! I prefer dispatching specific actions instead of one action to accommodate all these changes. lets talk more about hoisting the commands queue in the task_queue although looks like you brought us closer there too :-)

PR #14727, opened 2024-03-25
feat(step-generation): blowOut emits before touchTip

jerader:
closes AUTH-3, RAUT-581

# Overview

One of the items for PD 8.1. Basically, we just want step-generation to swap the order in which blow out and touch tip are emitted if those settings are both turned on, in the advanced settings.

Digging through the code, looks like it is already in the correct order for `mix` and `distribute` when aspirating. So this fix only applies to `transfer` and `consolidate`

# Test Plan

There is a cypress test that tests this change. But you can create a simple protocol with a transfer step, under advanced settings, add a touch tip and blow out to the dispense. Export the protocol and open the json file. See that the blow out is emitted first.

# Changelog

- swap order in compound commands
- fix unit tests
- fix cypress test

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/fixtures/protocol/8/example_1_1_0MigratedToV8.json
+++ b/protocol-designer/fixtures/protocol/8/example_1_1_0MigratedToV8.json
@@ -6,7 +6,7 @@
     "author": "Author name",
     "description": "Description here",
     "created": 1560957631666,
-    "lastModified": 1709309281554,
+    "lastModified": 1711649174279,
     "category": null,
     "subcategory": null,
     "tags": []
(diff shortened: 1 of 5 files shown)


ecormany:
Just FYI, here's the [order](https://docs.opentrons.com/v2/complex_commands/order_operations.html#step-sequence) we have documented for the Python API. I presume we want them to be the same.

1. Pick up tip
2. Mix at source
2. Aspirate from source
2. Touch tip at source
2. Air gap
2. Dispense into destination
2. Mix at destination
2. Touch tip at destination
2. Blow out
2. Drop tip

jerader:
> Just FYI, here's the [order](https://docs.opentrons.com/v2/complex_commands/order_operations.html#step-sequence) we have documented for the Python API. I presume we want them to be the same.
> 
> 1. Pick up tip
> 2. Mix at source
> 3. Aspirate from source
> 4. Touch tip at source
> 5. Air gap
> 6. Dispense into destination
> 7. Mix at destination
> 8. Touch tip at destination
> 9. Blow out
> 10. Drop tip

@ecormany, so we are swapping 8 and 9 in this PR per a bunch of customers and science's request. I'll post something in the slack channel to discuss this so we are all aligned!

koji (approved):
lgtm

PR #14729, opened 2024-03-25
fix(robot-server): ensure robot-server starts after mosquitto

mjhuff:
Closes [EXEC-351](https://opentrons.atlassian.net/browse/EXEC-351)

# Overview

@SyntaxColoring brought up the valid point that we should ensure mosquitto starts before the robot server, since the robot-server assumes this is true. This doesn't appear to be an issue in practice (so far), but let's fix it. 

# Risk assessment
low

[EXEC-351]: https://opentrons.atlassian.net/browse/EXEC-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/opentrons-robot-server.service
+++ b/robot-server/opentrons-robot-server.service
@@ -4,7 +4,7 @@ Description=Opentrons Robot HTTP Server
 
 Requires=nginx.service
 After=nginx.service
-
+After=mosquitto.service
 After=opentrons-status-leds.service
 
 
--- a/robot-server/opentrons-robot-server.service
+++ b/robot-server/opentrons-robot-server.service
@@ -4,9 +4,11 @@ Description=Opentrons Robot HTTP Server
 
 Requires=nginx.service
 After=nginx.service
-After=mosquitto.service
+
 After=opentrons-status-leds.service
 
+Wants=mosquitto.service
+After=mosquitto.service
 
 [Service]
 


robot-server/opentrons-robot-server.service
```diff
@@ -4,7 +4,7 @@ Description=Opentrons Robot HTTP Server
 
 Requires=nginx.service
 After=nginx.service
-
+After=mosquitto.service
```
SyntaxColoring: Do we also want to add `Wants=mosquitto.service`?

If we omit that, and we _only_ have `After=mosquitto.service`, the sequencing will be correct any time systemd starts both services, but systemd won't necessarily know that it has to start both services. Concretely: if you're in a state where nothing is started, and then you do `systemctl start opentrons-robot-server`, it won't also start `mosquitto.service`.

I'm not sure if this will matter in practice, since `mosquitto.service` is started by `multi-user.target`.

SyntaxColoring (approved):
LGTM once this has been tested with [the buildroot changes.](https://github.com/Opentrons/buildroot/pull/227)

robot-server/opentrons-robot-server.service
mjhuff: Hm, I don't feel strongly about it, but I could see that behavior as being potentially confusing. As long as mosquitto is guaranteed to start on boot (which I believe is always the case), then I think we're good. 

If you feel like it should be there, I don't mind adding it.

sfoster1: Please do also add `Wants` it's good practice. systemd is a general-purpose constraint solver and so if another, tighter constraint is active this wants will have no effect, but it's always a good idea to express intent in unit files like this.

sfoster1 (changes requested):

sfoster1 (approved):

PR #14732, opened 2024-03-25
fix(modules): use parse from packaging module

caila-marashaj:
Fix for https://github.com/Opentrons/opentrons/pull/14712.

We were using parse_version from `pkg_resources,` which can raise an `InvalidVersionError`, and importing `InvalidVersionError` from a separate packaging module. This imports the `parse` function from the same module the error would be raised from.

### DIFF
--- a/api/src/opentrons/hardware_control/modules/mod_abc.py
+++ b/api/src/opentrons/hardware_control/modules/mod_abc.py
@@ -91,6 +91,8 @@ class AbstractModule(abc.ABC):
             device_version = parse_version(self.device_info["version"])
             available_version = parse_version(self._bundled_fw.version)
             return cast(bool, available_version > device_version)
+        elif self.device_info and not self._bundled_fw:
+            return True
         return False
 
     async def wait_for_is_running(self) -> None:
--- a/api/src/opentrons/hardware_control/modules/mod_abc.py
+++ b/api/src/opentrons/hardware_control/modules/mod_abc.py
@@ -76,7 +76,7 @@ class AbstractModule(abc.ABC):
             return None
         file_prefix = self.firmware_prefix()
 
-        MODULE_FW_RE = re.compile(f"^{file_prefix}@v(.*)[.](hex|bin).*$")
+        MODULE_FW_RE = re.compile(f"^{file_prefix}@v(.*)[.](hex|bin)$")
         for fw_resource in ROBOT_FIRMWARE_DIR.iterdir():  # type: ignore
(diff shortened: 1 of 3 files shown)


sfoster1 (changes requested):
Let's get some tests of the version handling by forcing it to evaluate some invalid versions. This was a very weird issue that happened because of packaging minutia lol

api/src/opentrons/hardware_control/modules/mod_abc.py
```diff
@@ -16,6 +15,14 @@
 TaskPayload = TypeVar("TaskPayload")
 
 
+def parse_fw_version(version: str) -> Version:
```
DerekMaggio, line 18: What is the difference when using parse instead of instantiating a Version object directly? 
The answer is nothing, absolutely nothing. 
https://github.com/pypa/packaging/blob/main/src/packaging/version.py#L47-L56

It is for backward compatibility when `parse` used to return a Version or a LegacyVersion object

DerekMaggio (approved):
LGTM!

api/src/opentrons/hardware_control/modules/mod_abc.py
caila-marashaj: yeah, I was mostly using it here to detect a bad version name so we can default it to 0.0.0

DerekMaggio: No, your function is totally good. What I mean is all `packaging.version.parse(string_version)` does is call `Version(string_version)` under the hood. Ignore me lol

robot-server/tests/modules/test_router.py
```diff
@@ -71,6 +76,24 @@ async def test_get_modules_empty(
...
+@pytest.mark.parametrize(
+    argnames=["device_version", "expected_result"],
+    argvalues=[
+        ["v1.0.4", Version("v1.0.4")],
+        ["v0.5.6", Version("v0.5.6")],
+        ["v1.0.4-dhfs", Version("v0.0.0")],
+        ["v3.0.dshjfd", Version("v0.0.0")],
+    ],
+)
+async def test_catch_invalid_fw_version(
```
sfoster1: this should be next to the actual code it's testing

sfoster1 (changes requested):
The test looks good but it should really be in like `api/tests/opentrons/hardware_control/`, not all the way over in the robot server

sfoster1 (approved):
remove the rest of the diff from the robot server and we're good, nice!

PR #14734, opened 2024-03-26
refactor(app, components): parametersTable moved to Components dir for PL

jerader:
closes AUTH-240

# Overview

This PR moves the parameters desktop app table and empty state from the app directory to the components directory and also moves the formate value util to shared data

# Test Plan

In the desktop app, check the protocol details parameters tab and make sure it renders as expected

# Changelog

- pull the table out into its own folder in the components directory: `parametersTable` and move the no parameters component in that folder as well
- add tests
- move the format value util to `Shared-data` and fix all instances of it
- pass down an optional `t` as a prop for the app since it uses i18n but PL does not

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
@@ -14,10 +14,9 @@ import type {
   CompletedProtocolAnalysis,
   RunTimeParameter,
 } from '@opentrons/shared-data'
-import type * as Components from '@opentrons/components'
 
 vi.mock('@opentrons/components', async importOriginal => {
-  const actual = await importOriginal<typeof Components>()
+  const actual = await importOriginal<typeof NoParameters>()
   return {
     ...actual,
     NoParameters: vi.fn(),
--- a/app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
+++ b/app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
@@ -3,9 +3,9 @@ import { describe, it, vi, beforeEach, afterEach, expect } from 'vitest'
 import { screen } from '@testing-library/react'
 import { when } from 'vitest-when'
 
+import { NoParameters } from '@opentrons/components'
 import { renderWithProviders } from '../../../../__testing-utils__'
 import { i18n } from '../../../../i18n'
-import { NoParameter } from '../../../ProtocolDetails/ProtocolParameters/NoParameter'
 import { useMostRecentCompletedAnalysis } from '../../../LabwarePositionCheck/useMostRecentCompletedAnalysis'
 
 import { ProtocolRunRuntimeParameters } from '../ProtocolRunRunTimeParameters'
@@ -14,8 +14,15 @@ import type {
   CompletedProtocolAnalysis,
   RunTimeParameter,
 } from '@opentrons/shared-data'
+import type * as Components from '@opentrons/components'
 
-vi.mock('../../../ProtocolDetails/ProtocolParameters/NoParameter')
+vi.mock('@opentrons/components', async importOriginal => {
+  const actual = await importOriginal<typeof Components>()
+  return {
+    ...actual,
+    NoParameters: vi.fn(),
+  }
+})
 vi.mock('../../../LabwarePositionCheck/useMostRecentCompletedAnalysis')
 
 const RUN_ID = 'mockId'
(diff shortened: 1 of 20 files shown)


shlokamin (commented):
can the folder be named `ParametersTable` to match the current pattern (see `CheckboxField`)

jerader:
> can the folder be named `ParametersTable` to match the current pattern (see `CheckboxField`)

@shlokamin oh ya, i was following the other folders at the `src` level in components and saw they were all camel case like `tabbedNav`. I will capitalize the P though!

app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
```diff
@@ -14,8 +14,15 @@ import type {
   CompletedProtocolAnalysis,
   RunTimeParameter,
 } from '@opentrons/shared-data'
+import type * as Components from '@opentrons/components'
```
koji:
```suggestion

```

app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
```diff
@@ -14,8 +14,15 @@ import type {
...
 
-vi.mock('../../../ProtocolDetails/ProtocolParameters/NoParameter')
+vi.mock('@opentrons/components', async importOriginal => {
+  const actual = await importOriginal<typeof Components>()
```
koji:
```suggestion
  const actual = await importOriginal<typeof NoParameters>()
```

shared-data/js/helpers/formatRunTimeParameterValue.ts
```diff
@@ -1,9 +1,8 @@
-import { useTranslation } from 'react-i18next'
-import type { RunTimeParameter } from '@opentrons/shared-data'
+import { RunTimeParameter } from '../types'
 
 export const formatRunTimeParameterValue = (
   runTimeParameter: RunTimeParameter,
-  t: ReturnType<typeof useTranslation>['t']
+  t?: any
```
koji, line 5: Didn't work the type?

components/src/parametersTable/index.tsx
```diff
@@ -0,0 +1,119 @@
+import * as React from 'react'
+import styled from 'styled-components'
+import { formatRunTimeParameterValue } from '@opentrons/shared-data'
+import { BORDERS } from '../helix-design-system'
+import { SPACING, TYPOGRAPHY } from '../ui-style-constants/index'
+import { StyledText } from '../atoms/StyledText'
+
+import type { RunTimeParameter } from '@opentrons/shared-data'
+
+export * from './NoParameters'
```
koji: Do we need this?

components/src/parametersTable/NoParameters.tsx
```diff
@@ -1,21 +1,16 @@
...
-  const { t } = useTranslation('protocol_details')
+import { BORDERS, COLORS } from '../helix-design-system'
+import { SPACING, TYPOGRAPHY } from '../ui-style-constants/index'
+import { StyledText } from '../atoms/StyledText'
+import { Icon } from '../icons'
+import { Flex } from '../primitives'
+import { ALIGN_CENTER, DIRECTION_COLUMN } from '../styles'
 
+interface NoParametersProps {
+  t?: any
```
koji: if the type in utils.test.ts, probably using that instead of any would be good.

components/src/parametersTable/index.tsx
koji, line 1: Can we move this `components/src/molecules/` since components follows `atomic design`?

components/src/parametersTable/index.tsx
jerader: its so you can export `NoParameters` from the top level of `@opentrons/components`

components/src/parametersTable/NoParameters.tsx
jerader: the type of `t` comes from `useTranslation` but components doesn't have `i18Next` as a dependency, so i left it typed as `any`. Idk if there are any better methods? Do you have suggestions?

components/src/parametersTable/index.tsx
koji: index.ts doesn't work?

components/src/parametersTable/NoParameters.tsx
koji: at this moment no, I will keep an eye on that.

components/src/molecules/index.tsx
```diff
@@ -1,2 +1,4 @@
 export * from './LocationIcon'
 export * from './RoundTab'
+export * from './ParametersTable'
+export * from './ParametersTable/NoParameters'
```
koji, line 4: 👍

koji (approved):
lgtm

components/src/parametersTable/NoParameters.tsx
jerader: cool thanks, same here. I also don't like typing it as `any` but i'm not sure how else to do it so the app is still translated.

PR #14735, opened 2024-03-26
feat(api, robot-server): set sent runtime values and report parameters in analysis

jbleon95:
# Overview

Completes the analysis portion of AUTH-222.

This PR hooks up the runtime parameter API code to setting sent override values as well as reporting the parameters in the analysis summary, using the Pydantic models created in #14638 and #14718.

To report the parameters in analysis, a new method was added to the definitions to convert them into the Pydantic models. This method can then be used by the `ParameterContext` to get a list of all defined definitions. In order to have access to these after run execution, the ProtocolRunner was refactored to own this class so we can have access to it after the parameters are parsed and validated.

A method to set the parameters was also added and hooked up to `excute` and `run_python`, which will validate and set all sent parameters to be used in the protocol, for both analysis and execution.

# Test Plan

Used this protocol and checked the analysis result to make sure all parameters were being properly reported.

```
metadata = {
    'protocolName': 'Parameter Analysis Reporting Test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}


    parameters.add_int(
        display_name="Sample count",
        variable_name="sample_count",
        default=6,
        minimum=1,
        maximum=12,
    )
    parameters.add_float(
        variable_name="volume",
        default=20.1,
        choices=[
            {"display_name": "Low Volume", "value": 10.23},
            {"display_name": "Medium Volume", "value": 20.1},
            {"display_name": "High Volume", "value": 50.5},
        ],
        unit="µL"
    )
    parameters.add_bool(
        display_name="Dry Run",
        variable_name="dry_run",
        default=False,
    )
    parameters.add_str(
        display_name="Pipette Name",
        variable_name="pipette",
        choices=[
        ],
        default="flex_1channel_50",
    )



def run(context):

        pipette.pick_up_tip()

        pipette.drop_tip()
```

and in the analysis
```
  "runTimeParameters": [
    {
      "displayName": "Sample count",
      "variableName": "sample_count",
      "description": "How many samples to process.",
      "type": "int",
      "min": 1,
      "max": 12,
      "value": 6,
      "default": 6
    },
    {
      "displayName": "Pipette volume",
      "variableName": "volume",
      "description": "How many microliters to pipette of each sample.",
      "type": "float",
      "choices": [
        {
          "displayName": "Low Volume",
          "value": 10.23
        },
        {
          "displayName": "Medium Volume",
          "value": 20.1
        },
        {
          "displayName": "High Volume",
          "value": 50.5
        }
      ],
      "value": 20.1,
      "default": 20.1
    },
    {
      "displayName": "Dry Run",
      "variableName": "dry_run",
      "description": "Skip aspirate and dispense steps.",
      "type": "bool",
      "value": false,
      "default": false
    },
    {
      "displayName": "Pipette Name",
      "variableName": "pipette",
      "description": "What pipette to use during the protocol.",
      "type": "str",
      "choices": [
        {
          "displayName": "Single channel 50µL",
          "value": "flex_1channel_50"
        },
        {
          "displayName": "Eight Channel 50µL",
          "value": "flex_8channel_50"
        }
      ],
      "value": "flex_1channel_50",
      "default": "flex_1channel_50"
    }
  ],
```

# Changelog

- Added methods to get a list of all defined parameter as Pydantic model objects to send to client
- Hooked up parameters to be reported in analysis
- Use the values sent over via the `protocols` endpoint (and the `runs` in the future`) to set defined parameters

# Review requests

# Risk assessment

Low, no significant architecture changes and mostly adding real data to stubs.


api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -150,6 +158,14 @@ def __init__(
             drop_tips_after_run=drop_tips_after_run,
             post_run_hardware_state=post_run_hardware_state,
         )
+        # We need to define this here so we can use it throughout this class, but it will be recreated with the proper
+        # api level version upon a protocol loading.
+        self._parameter_context = ParameterContext(api_version=MAX_SUPPORTED_VERSION)
```
sanni-t: This feels a bit unnecessary and confusing.
Why not initialize it to None, give it an Optional[ParameterContext] type?

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -150,6 +158,14 @@ def __init__(
             drop_tips_after_run=drop_tips_after_run,
             post_run_hardware_state=post_run_hardware_state,
         )
+        # We need to define this here so we can use it throughout this class, but it will be recreated with the proper
+        # api level version upon a protocol loading.
```
sanni-t:
```suggestion
        # api level version upon loading a protocol.
```
 🔄

api/src/opentrons/protocols/execution/execute.py
```diff
@@ -20,17 +21,23 @@
...
+    :param run_time_param_overrides: Any parameter values that are potentially overriding the defaults
     """
     if isinstance(protocol, PythonProtocol):
         if protocol.api_level >= APIVersion(2, 0):
-            run_python(protocol, context)
+            # If this is None here then we're either running simulate or execute, in any case we don't need to report
+            # this in analysis which is the reason we'd pass it to this function
+            if parameter_context is None:
+                parameter_context = ParameterContext(protocol.api_level)
+            run_python(protocol, context, parameter_context, run_time_param_overrides)
```
sanni-t: Ow I'd make this call specify key word args rather than positional args. Too many arguments to rely on the order.

api/src/opentrons/protocols/parameters/validation.py
```diff
@@ -53,6 +54,58 @@ def ensure_unit_string_length(unit: Optional[str]) -> Optional[str]:
...
 
+def ensure_value_type(
+    value: Union[float, bool, str], parameter_type: type
+) -> AllowedTypes:
+    """Ensures that the value type coming in from the client matches the given type."""
+    validated_value: AllowedTypes
+    if isinstance(value, float) and parameter_type is int and value.is_integer():
+        validated_value = int(value)
+    else:
+        validated_value = value
```
sanni-t, line 71: if value = 123.4 and the parameter type is `int` then this will assign 123.4 as the validated value instead of raising an error.

api/src/opentrons/protocols/parameters/validation.py
```diff
@@ -53,6 +54,58 @@ def ensure_unit_string_length(unit: Optional[str]) -> Optional[str]:
...
+    return validated_value
+
+
+def ensure_enum_type(value: AllowedTypes) -> Union[float, str]:
+    """Ensure that the value that is sent over from the client is the expected type."""
```
sanni-t:
```suggestion
    """Ensure that the value that is sent over to the client is the expected type."""
```

api/src/opentrons/protocols/parameters/validation.py
```diff
@@ -53,6 +54,58 @@ def ensure_unit_string_length(unit: Optional[str]) -> Optional[str]:
...
+    validated_value: Union[float, bool, str]
+    if isinstance(value, bool):
+        raise ParameterValueError("Cannot send a boolean value as an enum type")
+    elif isinstance(value, int):
+        validated_value = float(value)
+    else:
+        validated_value = value
+    return validated_value
```
sanni-t: Why is this necessary? I think pydantic will just accept the int without any issues. As for the boolean value, wouldn't that be covered by the conversion you do in `as_protocol_engine_type`?

api/src/opentrons/protocols/parameters/validation.py
sanni-t: Oh, is this value then going to go through `validate_type` and will then raise an error? 
In that case, can we add this explanation in the docstring?

api/tests/opentrons/protocols/parameters/test_validation.py
```diff
@@ -129,6 +129,70 @@ def test_validate_options_raises_name_error() -> None:
...
+        (2.0, float, 2.0),
+        (2.2, float, 2.2),
+        ("3.0", str, "3.0"),
+        (True, bool, True),
+    ],
+)
+def test_ensure_value_type(
+    value: Union[float, bool, str], param_type: type, result: AllowedTypes
+) -> None:
+    """It should ensure the correct type is there, converting an float to ints."""
```
sanni-t:
```suggestion
    """It should ensure the correct type is there, converting floats to ints."""
```

api/tests/opentrons/protocols/parameters/test_validation.py
```diff
@@ -129,6 +129,70 @@ def test_validate_options_raises_name_error() -> None:
...
+    """It should convert the type into a string for the NumberParameter model."""
+    assert result == subject.convert_type_string_for_num_param(param_type)
+
+
+@pytest.mark.parametrize(
+    "param_type",
+    [str, bool],
+)
+def test_convert_type_string_for_num_param_raises(param_type: type) -> None:
+    """It should raise if given a bool or str to convert to an enum type string."""
```
sanni-t:
```suggestion
    """It should raise if given a bool or str to convert to a number type string."""
```

sanni-t (approved):
🙌 🎉 So exciting!! This looks really good! 
Just have a few nitpicks

api/tests/opentrons/protocols/parameters/test_validation.py
jbleon95: I could have sworn I was getting mypy errors for this, but removing it has no issue. Guess it was the same memory I had of putting in the `set_parameters` method

api/src/opentrons/protocols/parameters/validation.py
jbleon95: I could have sworn I was getting mypy errors for this, but removing it has no issue. Guess it was the same memory I had of putting in the set_parameters method

sanni-t (approved):
Thanks for making the changes!

PR #14736, opened 2024-03-26
feat(app): Display bad runs on the ODD

sfoster1:
When runs are bad, which means they couldn't be loaded correctly, we now get standin records called BadRuns rather than 500 errors. We should display that a run is bad on the ODD. For now, let's try making it red and displaying a status text.

This is what this looks like. Thoughts?
image


mjhuff (approved):
Very nice! Just need CI to pass, which I think requires updating the app typing.

sfoster1:
> Very nice! Just need CI to pass, which I think requires updating the app typing.

Yeah I really wish typescript allowed this with coalescing... that's a weird union of unions to support the legacy data structure

vegano1:
@sfoster1 

Full stack developer

PR #14737, opened 2024-03-26
refactor(shared-data): add filter tipracks to pipette defaultTipracks

jerader:
closes AUTH-232

# Overview

This PR adds the filter tiprack `loadname`s to the default tipracks list in the Flex pipette definitions. This info is needed for PD to accurately display the selectable tipracks for the given pipette.

Additionally i added the info to `defaultTipOverlapDictionary` which i am not sure is necessary or not, but just followed the pattern from the gen2 and gen1 pipette definitions

# Test Plan

Review the definitions and make sure it looks good.

PD correctly shows the pipettes!

Screenshot 2024-03-26 at 17 00 36

# Changelog

- add the filter tipracks to the pipette flex definitions. i added the info to all versions and default/low volume liquid
- fix the unit test for the util that returns the data

# Review requests

see test plan

# Risk assessment

low


sfoster1:
I think @andySigler should look at this but it seeeeems fine?

andySigler (approved):
LGTM, I took a scan through the flex pipette definitions and I don't see any instances of a filtered tip being assigned a tip-overlap other than `10.5` the default, which would mean that this PR has no change in behavior so lgtm

PR #14740, opened 2024-03-27
feat(app): add tooltips to ChooseRobotToRunProtocolSlideout RTPs, update types

ncdiehl11:
closes [AUTH-100](https://opentrons.atlassian.net/browse/AUTH-100)

# Overview

According to designs, number type RTPs should display their respective numerical ranges as InputField captions. The optional RTP description should be displayed in a tooltip when an information icon is hovered next to the RTP title. As part of the scope of this change, I also refactor the runtime parameter types so that the type of the RTP (numerical vs choice vs boolean) will narrow to the appropriate properties of that specific RTP when mapping.

# Test Plan

- Select any protocol from Desktop (mock data is still used for now)
- Select `Start setup`
- Choose any robot and select `Continue to parameters`
- Observe style of numerical RTPs with hoverable information icon
- Select `Change robot`
- Verify that slideout moves back to robot select

Screenshot 2024-03-27 at 9 39 21 AM

# Changelog

# Risk assessment

low

[AUTH-100]: https://opentrons.atlassian.net/browse/AUTH-100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
@@ -242,19 +242,6 @@ export function ChooseRobotToRunProtocolSlideoutComponent(
               protocol_name: protocolDisplayName,
             })
       }
-      // title={
-      //   enableRunTimeParametersFF && runTimeParameters.length > 0
-      //     ? currentPage === 1
-      //       ? t('choose_robot_to_run', {
-      //           protocol_name: protocolDisplayName,
-      //         })
-      //       : t('select_parameters_for_robot', {
-      //           robot_name: selectedRobot?.name,
-      //         })
-      //     : t('choose_robot_to_run', {
-      //         protocol_name: protocolDisplayName,
-      //       })
-      // }
       runTimeParametersOverrides={runTimeParametersOverrides}
       setRunTimeParametersOverrides={setRunTimeParametersOverrides}
       footer={
--- a/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
@@ -63,7 +63,6 @@ export function ChooseRobotToRunProtocolSlideoutComponent(
   )
 
   // TODO: (nd: 3/20/24) remove stubs and pull parameters from analysis
-
   const mockRunTimeParameters: RunTimeParameter[] = [
     {
       displayName: 'Dry Run',
@@ -117,10 +116,11 @@ export function ChooseRobotToRunProtocolSlideoutComponent(
       default: 'none',
     },
   ]
+  const runTimeParameters: RunTimeParameter[] = mockRunTimeParameters
   const [
     runTimeParametersOverrides,
     setRunTimeParametersOverrides,
-  ] = React.useState<RunTimeParameter[]>(mockRunTimeParameters)
+  ] = React.useState<RunTimeParameter[]>(runTimeParameters)
 
   const offsetCandidates = useOffsetCandidatesForAnalysis(
     mostRecentAnalysis,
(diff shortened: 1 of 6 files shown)


app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
```diff
@@ -231,14 +231,35 @@ export function ChooseRobotToRunProtocolSlideoutComponent(
...
+      //       ? t('choose_robot_to_run', {
+      //           protocol_name: protocolDisplayName,
+      //         })
+      //       : t('select_parameters_for_robot', {
+      //           robot_name: selectedRobot?.name,
+      //         })
+      //     : t('choose_robot_to_run', {
+      //         protocol_name: protocolDisplayName,
+      //       })
+      // }
```
jerader: do you need this still?

shared-data/js/types.ts
```diff
@@ -602,22 +603,29 @@ interface Choice {
...
 
 interface BooleanParameter {
+  type: BooleanParameterType
   default: boolean
 }
 
-type RunTimeParameterType = 'int' | 'float' | 'str' | 'boolean'
+type NumberParameterType = 'int' | 'float'
+type BooleanParameterType = 'boolean'
+type StringParameterType = 'str'
```
jerader, line 618: why do we need to further specify the types? I think i missed the discussion on this 😅

jerader (commented):
ui looks good! i have a question about the type changes you made

app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
ncdiehl11: absolutely do not 😅 removing

shared-data/js/types.ts
ncdiehl11: My rationale here is that we should restrict each parameter type's respective `type` property. In other words, if a parameter's `type` is `'int'`, type narrowing will allow us to access its `min` and `max` properties.

ncdiehl11: In other words, there should never be a parameter of `type: 'boolean'` that has a `min` and `max` property

jerader: hmm, imo it seems a bit unnecessary since you can determine if a parameter has `min` and `max` and `choice` via the `"in" operator`. For example in `app/src/pages/ProtocolDetails/Parameters.tsx`:

`const min = 'min' in parameter ? parameter.min : 0`

jerader: my preference is how we have it currently but @shlokamin @koji do yall have thoughts?

koji: If I added new types,  I would change something like below.

```ts
type ParameterType = 'int' | 'float' | 'boolean' | 'str';

interface BaseParameter {
  displayName: string;
  variableName: string;
  description: string;
  value: unknown;
  suffix?: string;
}

interface NumberParameter extends BaseParameter {
  type: 'int' | 'float';
  min: number;
  max: number;
  default: number;
}

interface Choice {
  displayName: string;
  value: unknown;
}

interface ChoiceParameter extends BaseParameter {
  type: 'str';
  choices: Choice[];
  default: string;
}

interface BooleanParameter extends BaseParameter {
  type: 'boolean';
  default: boolean;
}

export type RunTimeParameter = NumberParameter | ChoiceParameter | BooleanParameter;
```

ncdiehl11: Just spoke IRL regarding this. Resolution is to keep explicit `type` for each parameter interface

jerader (approved):
lgtm, thanks for the type change. I guess with that change we can clean up a few areas in the code in `Parameters` and `ParametersTable`

PR #14741, opened 2024-03-27
feat(api-client,app,react-api-client): implement ODD anonymous localization provider

brenthagen:
# Overview

implements a localization provider for the ODD app that substitutes anonymous translation values when the "enableOEMMode" robot setting is on. pushes the localization provider further down the DOM to be within the ODD api host provider for the robot settings request, and splits out a separate instance for the desktop app. 

refactors the `OnDeviceDisplayApp` component a bit to avoid rerenders of providers - currently the entire ODD app rerenders on every route change because of the scroll ref. 

refactors the initial loading screen out of the route tree and up the DOM to block localization provider render until the robot-server api is up

adds api client and react api client functions for robot settings get requests.

all translation keys that reference "opentrons" or "flex" are moved to the new `branded.json` and `anonymous.json` files. Some anonymous translation values have been updated (to confirm that the feature works) but anonymous copy will be finalized in PLAT-243

closes PLAT-265

Screen Shot 2024-03-26 at 5 10 32 PM
Screen Shot 2024-03-26 at 5 11 22 PM

# Test Plan

 - verified substitution of anonymous translation values
 - updates unit tests

# Changelog

 - Adds anonymous localization provider for ODD

# Review requests

evaluate the approach and app changes

# Risk assessment

medium, changes at top level of app


b-cooper (approved):
Went over the code together, changes look good 🗺️

PR #14746, opened 2024-03-28
refactor(api): Split CommandStore

mjhuff:
Closes [EXEC-55](https://opentrons.atlassian.net/browse/EXEC-55) and [EXEC-361](https://opentrons.atlassian.net/browse/EXEC-361)

# Overview

This PR splits the CommandStore. All data access is now O(1). Thanks to @SyntaxColoring for getting this started.

- Run-level concerns are still handled by CommandStore.
- Data-organization and access concerns are handled by the new CommandHistory.  

While this primarily serves as a refactor, there are some new assumptions made about queueing and failing commands. I've outlined them in comments below.

# Test Plan
- If you feel that testing covers this, then it's just a plain old refactor.
- Verify protocols run through various flows (I did do this after edge was fixed). 

# Review requests
Check my comments and double check me on my assumptions. 

# Risk assessment
medium-ish. In theory this is just a refactor. In practice, I'm still learning PE...although the tests are still the same. 

[EXEC-55]: https://opentrons.atlassian.net/browse/EXEC-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EXEC-361]: https://opentrons.atlassian.net/browse/EXEC-361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -513,10 +506,7 @@ def get_all(self) -> List[Command]:
         Replacing a command (to change its status, for example) keeps its place in the
         ordering.
         """
-        return [
-            self._state.commands_by_id[cid].command
-            for cid in self._state.all_command_ids
-        ]
+        return self._state.command_history.get_all_commands()
 
     def get_slice(
```
mjhuff, line 456: I'm not entirely sure why `get_slice` is functionally different from `get_current`, but it is. In practice, do they really need to be, though? 

I know that historically, the app takes a 1 element slice to get the "current command", and `get_current` is not exposed in the HTTP API.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -231,8 +208,6 @@ def __init__(
     def handle_action(self, action: Action) -> None:  # noqa: C901
```
mjhuff, line 208: Initially, I wanted to move this into CommandHistory, but this seems to be a run-level concern, even more-so now with the recent UpdateCommand split.

api/src/opentrons/protocol_engine/state/command_history.py
```diff
@@ -0,0 +1,152 @@
...
+    """The IDs of queued setup commands, in FIFO order"""
+
+    _running_command_id: Optional[str]
+    """The ID of the currently running command, if any"""
+
+    _dequeued_command_id: Optional[str]
+    """ID of the most recent command that was dequeued, if any"""
+
+    def __init__(self) -> None:
+        self._queued_command_ids = OrderedSet()
```
mjhuff, line 44: I got rid of `all_command_ids`. I don't think we need it when we have `queued_command_ids`.

EDIT: Unfortunately I was wrong here - will have to add back the ol' list.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -112,32 +114,11 @@ class CurrentCommand:
...
-
-    queued_setup_command_ids: OrderedSet[str]
-    """The IDs of queued setup commands, in FIFO order"""
-
-    running_command_id: Optional[str]
-    """The ID of the currently running command, if any"""
-
-    commands_by_id: Dict[str, CommandEntry]
-    """All command resources, in insertion order, mapped by their unique IDs."""
+    command_history: CommandHistory
```
SyntaxColoring, line 121: `CommandHistory` is a much better name than `CommandStructure`, thanks.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -591,34 +583,39 @@ def get_error(self) -> Optional[ErrorOccurrence]:
...
+        if running_command is not None:
+            return running_command.command.id
+        else:
+            return None
+
+    def get_queue_ids(self) -> OrderedSet[str]:
+        """Get the IDs of all queued protocol commands, in FIFO order."""
+        return self._state.command_history.get_queue_ids()
 
     def get_current(self) -> Optional[CurrentCommand]:
```
mjhuff, line 541: Before, we were:

- Checking the `running_command`.
- Checking the overall run status.
- Iterating backwards to the `final_command`. 

Now we are: 
- Checking the `running_command`.
- Checking the overall run status.
- Storing the last known completed command, and iterating at most 1 command to determine the final command.

I assume we should do iteration, because it's possible for a command to receive a status update before being moved into/out of a queue. If that's not the case, this can be simplified further.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -726,15 +747,12 @@ def raise_fatal_command_error(self) -> None:
...
-        # TODO(mm, 2024-03-14): This is a slow O(n) scan. When a long run ends and
-        # we reach this loop, it can disrupt the robot server.
-        # https://opentrons.atlassian.net/browse/EXEC-55
-        for command_id in self._state.all_command_ids:
-            command = self._state.commands_by_id[command_id].command
-            if command.error and command.intent != CommandIntent.SETUP:
-                raise ProtocolCommandFailedError(
-                    original_error=command.error, message=command.error.detail
-                )
+        failed_command = self.state.failed_command
```
mjhuff, line 701: Is this sufficient? I believe that if a command fails, it must be set as the failed command. Maybe this just wasn't refactored over, since the change to mark other commands as failed without errors was rather recent.

mjhuff (commented):
Some of the more interesting things that I'd like the PE gurus to check:

api/src/opentrons/protocol_engine/state/command_history.py
```diff
@@ -0,0 +1,152 @@
...
+        try:
+            return self._commands_by_id[command_id]
+        except KeyError:
+            raise CommandDoesNotExistError(f"Command {command_id} does not exist")
+
+    def get_next(self, command_id: str) -> Optional[CommandEntry]:
+        """Get the command which follows the command associated with the given ID, if any."""
+        try:
+            index = self._commands_by_id[command_id].index
+            commands_by_id_list = list(self._commands_by_id.keys())
```
SyntaxColoring: Unfortunately, doing `list(self._commands_by_id.keys())` is going to be a sneaky O(n). `.keys()` returns an iterable that gives you one element at a time, and `list(...)` scans over all of those and builds a list out of them.

I think this is probably why the original `CommandState` had a `all_command_ids: List[str]` attribute. It was redundant with `commands_by_id` in terms of information, but not in terms of performance.

api/src/opentrons/protocol_engine/state/command_history.py
```diff
@@ -0,0 +1,152 @@
...
+        """Clears all commands within the queued setup command ids structure."""
+        self._queued_setup_command_ids.clear()
+
+    def remove_queue_id(self, command_id: str) -> None:
+        """Remove a specific command from the queued command ids structure."""
+        self._queued_command_ids.discard(command_id)
+
+    def remove_setup_queue_id(self, command_id: str) -> None:
+        """Remove a specific command from the queued setup command ids structure."""
+        self._queued_setup_command_ids.discard(command_id)
```
SyntaxColoring: * `set_dequeued_command_id`
* `set_running_command_id`
* etc.

Can we move these setters up a level of abstraction?

For example, when a command transitions from `queued` to `running`, we don't want `CommandStore` to have to remember to call both `command_history.set_running_command_id(id)` and `command_history.remove_setup_queue_id(id)`. We want it to be able to just call something like `command_history.set_command_running(command)`, and then `CommandHistory` takes care of the bookkeeping as an internal implementation detail.

So, in total, the public setters in `CommandHistory` would be roughly like:

* `queue_new_command(new_queued_command: Command)`
    * Validates that `new_queued_command.state == queued`.
    * Validates that `new_queued_command.id` doesn't exist yet.
    * Automatically calculates the index and adds it to the protocol queue or setup queue, as appropriate.
* `set_command_running(new_running_command: Command)`
    * Validates that `new_running_command.id` points to a command that's currently queued.
    * Removes it from the protocol queue or setup queue, as appropriate.
    * Does bookkeeping to allow efficient future retrieval of this as the running command.
* `set_command_succeeded(new_succeeded_command)`/`set_command_failed(new_failed_command)`
    * Validates that the command is currently running.
    * Marks it as no longer running.

That's one of our motivations to extract all of these attributes to the new class.

api/src/opentrons/protocol_engine/state/command_history.py
```diff
@@ -0,0 +1,152 @@
...
+        commands = list(self._commands_by_id.values())
+        return [command.command for command in commands]
+
+    def get_all_ids(self) -> List[str]:
+        """Get all command IDs."""
+        return list(self._commands_by_id.keys())
+
+    def get_slice(self, start: int, stop: int) -> List[Command]:
+        """Get a list of commands between start and stop."""
+        commands = list(self._commands_by_id.values())[start:stop]
```
SyntaxColoring: Same as my other comment: This is a sneaky O(total_num_commands), but we want it to be O(size_of_slice). Bringing back the `all_command_ids: List[str]` attribute would solve this too.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -276,36 +254,42 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
...
+                    index=prev_entry.index,
+                    command=succeeded_command,
+                ),
             )
 
-            assert self._state.running_command_id == action.command.id
-            self._state.running_command_id = None
+            running_command_entry = self._state.command_history.get_running_command()
+            assert running_command_entry is not None
+            assert running_command_entry.command == action.command
```
TamarZanzouri: we should probably assert the command id's

api/src/opentrons/protocol_engine/state/commands.py
TamarZanzouri: we are returning the current command in the `/commands` in the `link.current` prop
https://github.com/Opentrons/opentrons/blob/edge/robot-server/robot_server/runs/router/commands_router.py#L285
I think @shlokamin and I had a convo about this and in some cases we are using this property and in some we use `get_slice`

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -526,36 +516,38 @@ def get_slice(
         """Get a subset of commands around a given cursor.
 
         If the cursor is omitted, a cursor will be selected automatically
-        based on the currently running or most recently executed command."
+        based on the currently running or most recently executed command.
         """
-        # TODO(mc, 2022-01-31): this is not the most performant way to implement
```
TamarZanzouri, line 531: nice!

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: Sorry, this is confusing because of the way the actions are defined—but I think this assert will always fail. `action.command` is the command with `status == succeeded` (what we want the command to transition *to*), whereas `running_command_entry.command` is the command with `status == running` (what the command is *right now*).

api/src/opentrons/protocol_engine/state/command_history.py
SyntaxColoring: This can be in a separate PR if that makes things easier.

SyntaxColoring (commented):
TY!

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: Hmm but somehow the tests are passing. I'm missing something.

mjhuff:
Looking at the tavern test failure, I'm a bit perplexed why this was passing in the first place. I'm sure I'm missing something.

From what I see:
- We add a setup command, which is added as a QueueCommandAction. This means we have no `running_command_id` currently.
- When we `GET` the commands, we add the links via `get_current`, which comes pretty much directly from get_current()
- Invoking through `get_current`, we don't have a `running_command_id`, so we iterate backwards. The tavern test expects this `QUEUED` command to be returned, but that's only true if `run_requested_to_stop`, which isn't true. 

Anyone have any thoughts? I'll dig further into this.

UPDATE: Solved. I missed the `WaitUntilCompletion` bit in the tavern test when the setup command was POSTed. The error is due to not actually invoking `set_dequeued_command_id`.

api/src/opentrons/protocol_engine/state/command_history.py
mjhuff: Ah, yeah, you're right on this. Bummer.

api/src/opentrons/protocol_engine/state/commands.py
mjhuff: Hm, I guess that is more performant, since we can guarantee command ID's are unique. I'll change this.

EDIT: See Max's comment below. This assert will always fail. We do want to compare IDs.

api/src/opentrons/protocol_engine/state/commands.py
TamarZanzouri: we set this property when we receive a `FailedCommandAction` to know what command marked all the other commands as failed so yeah, we probably just didnt refactor this part

api/src/opentrons/protocol_engine/state/command_history.py
mjhuff: Good call. I'll do that here! I kinda need to abstract away adding to the list that I not so cleverly removed, anyway.

TamarZanzouri (dismissed):
Nice and Clean! a few comments but looks great!

api/src/opentrons/protocol_engine/state/commands.py
mjhuff: Ahhhhh, yep. Thanks!

api/src/opentrons/protocol_engine/state/command_history.py
```diff
@@ -0,0 +1,163 @@
...
+        """Get a command entry if present, otherwise raise an exception."""
+        try:
+            return self._commands_by_id[command_id]
+        except KeyError:
+            raise CommandDoesNotExistError(f"Command {command_id} does not exist")
+
+    def get_next(self, command_id: str) -> Optional[CommandEntry]:
+        """Get the command which follows the command associated with the given ID, if any."""
+        try:
+            index = self._commands_by_id[command_id].index
```
sfoster1: nit: if you make this
```
def get_next(...)->...:
    index = self.get(command_id).index
    try:
         return self._commands_by_id[...]
    except:
          ...
```
you get to borrow the command does not exist from `get()`. That's good, because right now there's a technical possibility that you can get a `KeyError` and thus `CommandDoesNotExistError` from _both_ the access onl line 67 and the access on line 68, and you wouldn't be able to tell which easily.

api/src/opentrons/protocol_engine/state/commands.py
sfoster1: this will have to change pretty soon when we can have multiple failed commands and such but I think it's good for now.

sfoster1 (approved):
Looks really solid with some occasional nitpicks!

mjhuff:
Ok, turns out `get_current` was slightly off for two reasons:

1. We don't actually care about dequeued commands, just terminal (`SUCCEEDED` or `FAILED`) commands. I renamed the property to reflect that, and we set the `terminal_id` only in these instances.
2. As an extension of 1), my `get_final_command` logic was slightly off. If `run_requested_to_stop`, we also need to check whether a command is `RUNNING` and handle it appropriately.

TamarZanzouri (approved):
Thank you for the changes! just a question!

api/src/opentrons/protocol_engine/state/command_history.py
```diff
@@ -0,0 +1,256 @@
...
+        self._queued_command_ids.clear()
+
+    def clear_setup_queue(self) -> None:
+        """Clears all commands within the queued setup command ids structure."""
+        self._queued_setup_command_ids.clear()
+
+    def set_command_queued(self, command: Command) -> None:
+        """Validate and mark a command as queued in the command history."""
+        assert command.status == CommandStatus.QUEUED
+        assert not self.has(command.id)
```
TamarZanzouri, line 149: do we need this logic? is there a chance we call this method when a command is already queued?

mjhuff: You're right, we technically don't. I kinda like the extra assertions as I think it makes it easier to look back months from now and understand what's going on. FWIW, we do something similar on `edge` in L294:

`assert succeeded_command.status == CommandStatus.SUCCEEDED`.

SyntaxColoring (approved):
Big improvement, thank you!

PR #14748, opened 2024-03-28
(WIP) Error Recovery Test Protocols

DerekMaggio:
# Overview

Python protocols for testing Error Recovery

PR #14753, opened 2024-03-28
feat(api): Pause when `pick_up_tip()` errors in a Python protocol

SyntaxColoring:
# Overview

Closes EXEC-346.

# Test Plan

## Setup

Enable the hidden error recovery feature flag by manually issuing a `POST /settings` request with `{"id": "enableErrorRecoveryExperiments", "value": true}`.

## Test

Upload and run this protocol.

```python

requirements = {"robotType": "Flex", "apiLevel": "2.17"}


    protocol.load_trash_bin("A3")

    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_50ul", "C2")
    reservoir = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "C3"
    )
        "flex_1channel_50", mount="left", tip_racks=[tip_rack]
    )

    for source, dest in zip(reservoir.columns()[0], reservoir.columns()[1]):
        pipette.pick_up_tip()
        pipette.move_to(source.top())
        pipette.move_to(dest.top())
        pipette.drop_tip()
```

* [x] When it hits a missing spot in the tip rack, it should pause. From there, you should be able to issue a `stop` or `resume-from-recovery` action to `POST /runs/{id}/actions`.
* [x] If you do `resume-from-recovery`, the `pick_up_tip()` calls later on in the protocol should continue from the next wells in the tip rack. It should not repeatedly try to pick up tips from the same well.

## Cleanup

`POST /settings` with `{"id": "enableErrorRecoveryExperiments", "value": false}`.

# Changelog

* We currently have some machinery to send a command to `ProtocolEngine` and block until the command finishes executing. The main part of this PR is adding variants of that to *also* wait for the error recovery to finish. We reimplement `ProtocolContext.pick_up_tip()`, specifically, with that new variant. Other commands will happen later.
* The tip handler in `ProtocolEngine` is modified so that when a `pickUpTip` fails recoverably, it marks the requested tips as used. This is one way of getting the next PAPI `pick_up_tip()` to automatically move on to the next tip. See [discussion below](https://github.com/Opentrons/opentrons/pull/14753#discussion_r1546611986).

# Review requests

See my review comments.

# Risk assessment

Low-risk when the `enableErrorRecoveryExperiments` feature flag is off. Medium-risk if it's on, since this starts venturing further into territory where Protocol Engine's understanding of the protocol history and state is different from PAPI's understanding of the protocol history and state.


DerekMaggio:
@SyntaxColoring, https://github.com/Opentrons/opentrons/pull/14748 contains some error recovery protocols if you need any test cases.

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -242,6 +242,27 @@ async def add_and_execute_command(
         await self.wait_for_command(command.id)
         return self._state_store.commands.get(command.id)
 
+    async def add_and_execute_command_wait_for_recovery(
+        self, request: commands.CommandCreate
+    ) -> None:
```
SyntaxColoring: I'm open to ideas for better names for this.

Or maybe we should make it a `wait_for_recovery: bool` argument on the existing `add_and_execute()` method.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -715,6 +715,26 @@ def get_all_commands_final(self) -> bool:
 
         return no_command_running and no_command_to_execute
 
+    def get_command_recovery_complete(self, command_id: str) -> bool:
```
sfoster1: can we call this like `get_command_recovery_not_needed()` or something? as your docstring calls out this will definitely return true if even there was no recovery to complete, so let's just avoid the confusion

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -408,13 +408,19 @@ def pick_up_tip(
...
             well_location=well_location,
         )
-        self._engine_client.pick_up_tip(
+        self._engine_client.pick_up_tip_wait_for_recovery(
             pipette_id=self._pipette_id,
             labware_id=labware_id,
             well_name=well_name,
             well_location=well_location,
         )
+        # FIX BEFORE MERGE?: If the pickUpTip fails, the tip tracker (which is sort of
```
sfoster1: as discussed in person - I think we are going to have to determine this sort of stuff command by command. we should make our decision based on "what's the most likely physical cause", and then handle the possibility of surprising states by being conservative and defensive about querying PE state explicitly rather than explicitly or implicitly caching.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -408,13 +408,19 @@ def pick_up_tip(
...
+        # kind of in Protocol Engine) doesn't think the tip has been consumed. So,
+        # the next PAPI pick_up_tip() after the error recovery will try to pick up
+        # the same tip, diverging from the protocol's analysis.
 
+        # FIX BEFORE MERGE: We should probably only set_last_location() if the
```
sfoster1: I disagree - I think there are three answers each of which is honestly reasonable, but one which is the best (and also hardest). As a preamble, we should probably push last-location down to the engine.

1. the best and hardest is, base the decision of setting last location on when the error occurred (during which action) and what the error is. For instance, if you got a stall while moving to the tiprack, we should clear the last location since position is now indeterminate; but if we failed our tip-state consistency check after the pickup, we should set the last location to the rack.
2. the simplest and most robust is, always clear the last location on any error
3. and a slightly more optimistic version is to always set it on any error since the most likely causes of failure will in fact leave the gantry in the specified location

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -261,14 +291,26 @@ def estop(
...
 
-        if current_id is not None:
+        if running_or_next_queued_id is not None:
+            assert running_or_next_queued is not None
+
             fail_action = FailCommandAction(
-                command_id=current_id,
+                command_id=running_or_next_queued_id,
+                # FIXME(mm, 2024-04-02): As of https://github.com/Opentrons/opentrons/pull/14726,
+                # this action is only legal if the command is running, not queued.
```
SyntaxColoring, line 323: This is my fault; in #14726, I neglected to account for what this `estop()` method was doing. I'm going to have to fix this in another PR. EXEC-382

SyntaxColoring: I think a lot of this `estop()` method is due for a rethink. Like, I don't really get why it needs to be so different from `stop()`, and why it needs to be messing with things like `FailCommandAction`s itself.

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: Replaced with the negation, `get_recovery_in_progress_for_command()`.

api/src/opentrons/protocol_api/core/engine/instrument.py
SyntaxColoring: After looking at this closer, I think it's appropriate to call this `set_last_location()` unconditionally. See this new comment:

https://github.com/Opentrons/opentrons/blob/6b79625f761910a61eca33dae2c9571bd88ef3e5/api/src/opentrons/protocol_api/core/engine/instrument.py#L412-L423

api/src/opentrons/protocol_engine/clients/transports.py
```diff
@@ -64,21 +75,80 @@ def execute_command(self, request: CommandCreate) -> CommandResult:
...
-        assert command.result is not None, f"Expected Command {command} to have result"
+        if command.result is None:
+            # This can happen with a certain pause timing:
+            #
+            # 1. The engine is paused.
+            # 2. The user's Python script calls this method to start a new command,
+            #    which remains `queued` because of the pause.
+            # 3. The engine is stopped. The returned command will be `queued`
+            #    and won't have a result.
+            raise ProtocolCommandFailedError(
```
sfoster1: let's make this its own unique exception type and code so that we can programmatically differentiate it from other errors

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -261,15 +301,26 @@ def estop(
         """
         if self._state_store.commands.get_is_stopped():
             return
-
-        current_id = (
+        running_or_next_queued_id = (
             self._state_store.commands.get_running_command_id()
             or self._state_store.commands.get_queue_ids().head(None)
+            # TODO(mm, 2024-04-02): Is it possible for the next queued command to
```
sfoster1: sure, this could be a maintenance run

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -708,6 +708,15 @@ def get_all_commands_final(self) -> bool:
 
         return no_command_running and no_command_to_execute
 
+    def get_recovery_in_progress_for_command(self, command_id: str) -> bool:
+        """Return whether the given command failed and its error recovery is in progress."""
+        command_is_most_recent_to_fail = (
```
sfoster1: this is a little gross to me, could we save the command we're currently recovering from explicitly? like, what happens if a fixit command fails, wouldn't it instantly make this `False`?

api/src/opentrons/protocol_engine/state/tips.py
```diff
@@ -124,6 +130,33 @@ def _handle_command(self, command: Command) -> None:
...
+        if (
+            isinstance(action.running_command, PickUpTip)
+            and action.type != ErrorRecoveryType.FAIL_RUN
+        ):
+            self._set_used_tips(
+                pipette_id=action.running_command.params.pipetteId,
+                labware_id=action.running_command.params.labwareId,
+                well_name=action.running_command.params.wellName,
+            )
+            # TODO(mm, 2024-04-01): We're logically removing the tip from the tip rack,
```
sfoster1: It is not really inconsistent and confusing once we're in the world of being aware of possible errors. The tip is not usable; the tip is also not on the pipette. We may want to do various error handling cases where we treat the pipette as if it has a tip, and we'll need the data then; we also may want to do cases where the pipette does not have a tip. We have to be able to handle that, and I think that this is a great way to implement it.

Part of what we get out of the separated states like this _is_ that separation, and we're taking advantage of it here, and I'm glad.

api/src/opentrons/protocol_engine/clients/transports.py
```diff
@@ -30,16 +32,22 @@ def __init__(self, engine: ProtocolEngine, loop: AbstractEventLoop) -> None:
                 want to synchronously access it.
             loop: The event loop that `engine` is running in (in the other thread).
         """
-        self._engine = engine
-        self._loop = loop
+        # We might access these from different threads,
```
TamarZanzouri, line 46: Nice

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -234,14 +234,45 @@ async def add_and_execute_command(
...
+            If the command completed, it will be succeeded or failed. If it failed
+            and then its failure was recovered from, it will still be failed.
+
+            If the engine was stopped before it reached the command,
+            the command will be queued.
+        """
+        queued_command = self.add_command(request)
+        await self.wait_for_command(command_id=queued_command.id)
+        completed_command = self._state_store.commands.get(queued_command.id)
+        await self._state_store.wait_for_not(
```
sfoster1, line 270: it kind of feels like we want to gate this on a synchronous call to whether we're now in recovery for the command, but this in general feels a little race condition-y because of the separation between command state and run state. Specifically,
- `wait_for_command` returns on the asyncio spin after a `FailCommandAction` or a `SucceedCommandAction` for this command (we can neglect the queued-and-stopping part for now)
- but the `get_recovery_in_progress_for_command` predicate is based on the queue status being awaiting-recovery

Do we guarantee mechanically that the queue status will be set before the asyncio spin after the command terminal action is dispatched? Are we sure this won't occasionally race and return early?

sfoster1 (changes requested):
This looks good overall, but a couple points:
- I'm a bit concerned about the "last command to fail" logic in the command store. Can we instead save the command id when it fails recoverably and we enter `AWAITING_RECOVERY` and make that state more explicit?
- Something feels slightly off about the checks in `ProtocolEngine.add_and_execute_command_wait_for_recovery` since they check two different pieces of state in tight sequence

api/src/opentrons/protocol_engine/execution/command_executor.py
```diff
@@ -167,6 +167,7 @@ async def execute(self, command_id: str) -> None:
                 FailCommandAction(
                     error=error,
                     command_id=running_command.id,
+                    running_command=running_command,
```
TamarZanzouri, line 170: I guess I am wondering why we need this in the action? dont we have the failed command stored in PE already?

TamarZanzouri (commented):
overall looks good, added a question and a bit confused about the `wait_for_not` although I get what its doing but the logic around it is a bit confusing to me. besides that looks great!

api/src/opentrons/protocol_engine/protocol_engine.py
SyntaxColoring: Yes, if I understand your concern correctly:

When `await self.wait_for_command(command_id=queued_command.id)` returns, we are guaranteed that the action that finalized the command has already been fully processed, and that `get_recovery_in_progress_for_command()` will see its results on the state.

When we handle an action, we [send it to each of the substores in a loop](https://github.com/Opentrons/opentrons/blob/bd14851425538ebdd10f26dc773a0f898ea59c31/api/src/opentrons/protocol_engine/state/state.py#L208-L211). Only after that's done do we [notify subscribers](https://github.com/Opentrons/opentrons/blob/bd14851425538ebdd10f26dc773a0f898ea59c31/api/src/opentrons/protocol_engine/state/state.py#L355-L357) like this one.

api/src/opentrons/protocol_engine/execution/command_executor.py
SyntaxColoring: It's a hack, we probably don't need it. See this comment:

https://github.com/Opentrons/opentrons/blob/bd14851425538ebdd10f26dc773a0f898ea59c31/api/src/opentrons/protocol_engine/actions/actions.py#L177-L180

api/src/opentrons/protocol_engine/state/commands.py
SyntaxColoring: Changed in a880096260.

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -178,6 +178,9 @@ class CommandState:
     stable. Eventually, we might want this info to be stored directly on each command.
     """
 
+    recovery_target_command_id: Optional[str]
+    """If we're currently recovering from a command failure, which command it was."""
+
```
SyntaxColoring, line 183: Adding this as an orthogonal attribute is quick and dirty. Ideally, this ID would not exist at all when `QueueStatus` is anything other than `AWAITING_RECOVERY`. We could do that by retructuring this state so it's closer to a union of dataclasses instead of a dataclass of unions.

api/src/opentrons/protocol_engine/protocol_engine.py
SyntaxColoring: Ugh, yep, good call.

api/src/opentrons/protocol_engine/clients/transports.py
SyntaxColoring: Done in 97b0ada3dee705bafbc91e22f2f941dbdb4e0523.

api/src/opentrons/protocol_engine/state/tips.py
SyntaxColoring: Discussed in-person:

I think this is inconsistent and confusing today, but it will be much less so if we treated the different tip pickup failure modes (tip consumed vs. not; tip attached vs. not, ...) as distinct, explicit `pickUpTip` results models.

For now, replacing the to-do comment with a neutral note.

sfoster1 (approved):
Looks great, thank you for the changes!

PR #14755, opened 2024-03-28
chore(api): fix typing of task_queue workers

sfoster1:
task_queue has these cleanup and run functions that you pass in with some arguments to bind. None of this was typesafe because it predated ParamSpec. Now that we have ParamSpec, we can make these typesafe.

## changelog
typing only in `task_queue`. `set_run_func` and `set_cleanup_func` get paramspecs. `set_run_func` is trivial; `set_cleanup_func` has a requirement to have an `error` argument (which has to be the first argument, for "paramspec still isn't good enough" reasons, but that was already an implicit requirement).

also, we do a manual closure instead of `partial()` because I'm pretty sure `partial()` isn't actually typesafe.

### DIFF
--- a/api/src/opentrons/protocol_runner/task_queue.py
+++ b/api/src/opentrons/protocol_runner/task_queue.py
@@ -1,25 +1,12 @@
 """Asynchronous task queue to accomplish a protocol run."""
 import asyncio
 import logging
-from functools import partial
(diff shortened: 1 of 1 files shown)


TamarZanzouri (approved):
Nice!

PR #14759, opened 2024-03-29
feat(api): Do not enqueue json commands on protocol load

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/EXEC-352.
first step towards fixit commands. do not enqueue json protocol commands. 

# Test Plan

Tested with Json protocols and Postman:
- Make sure loading a protocol and executing it are happening within order.
- Make sure get run `/commands` returning the list properly with successful commands. 
- Make sure loading a failed protocol should fail the run and fail the command.
- Make sure get run `/commands` for failed runs the list of commands, last command being the failed command. 
- Fixed e2e test to comply with these changes

# Changelog

- Do no enqueue commands in PE for Json command upon load. 
- Execute commands one by one when run get started - same way we do for python protocols.

# Review requests

Changes make sense? 
GET run` /commands` will not return the full list of commands if the run did not start - its a change we are doing to make json protocols run like python protocols. are we ok with this? 

# Risk assessment

Medium. need to do smoke tests for Json protocols and make sure these changes do not affect anything.


sfoster1 (approved):
Looks great to me! Nice work.

robot-server/tests/integration/http_api/runs/test_json_v6_protocol_run.tavern.yaml
SyntaxColoring, line 1: > GET run /commands will not return the full list of commands if the run did not start - its a change we are doing to make json protocols run like python protocols. are we ok with this?

Is the long-term plan for JSON protocols to not have an HTTP-visible queue anymore, or is this just temporary until some future PR adds it back?

api/tests/opentrons/protocol_runner/test_protocol_runner.py
```diff
@@ -393,24 +393,7 @@ async def test_load_json_runner(
...
-        ),
-        protocol_engine.add_command(
-            request=pe_commands.LoadLiquidCreate(
-                params=pe_commands.LoadLiquidParams(
-                    liquidId="water-id", labwareId="labware-id", volumeByWell={"A1": 30}
-                )
-            ),
-        ),
-        task_queue.set_run_func(func=protocol_engine.wait_until_complete),
+        task_queue.set_run_func(func=json_runner_subject._add_command_and_execute),
```
SyntaxColoring: To avoid testing internal methods, I think we want to do something like this:

https://github.com/Opentrons/opentrons/blob/459f14d5f78d064991ba6a195c1de55d750f12e5/api/tests/opentrons/protocol_runner/test_protocol_runner.py#L481-L504

mjhuff:
After pushing all the backend code and restarting the bot, it appears to be failing protocol analysis:

Screenshot 2024-04-01 at 11 19 45 AM

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -355,6 +357,15 @@ async def run(  # noqa: D102
...
 
+    async def _add_command_and_execute(self) -> None:
+        for command in self._queued_commands:
+            result = await self._protocol_engine.add_and_execute_command(command)
+            if result.error:
+                raise ProtocolCommandFailedError(
+                    original_error=result.error,
+                    message=f"{result.error.errorType}: {result.error.detail}",
+                )
+
```
SyntaxColoring: Ignoring queued commands for a moment, this doesn't look like it matches the existing implementation based on `QueueWorker`+`get_next_to_execute()`. This does not have:

* ~~Special handling of `RunStoppedError` to `break` instead of raising an error~~
* ~~Yielding to the event loop on each command~~
* Pausing on recoverable errors (added in #14646)
    * You can see #14753 for how I'm doing this for Python protocols. If we're making JSON behave more like Python now, maybe this should work the same way.
* Prioritization of `"intent": "setup"` commands
    * But according to the tests, this is still working, somehow?

Do we need to handle those things, or are they already handled elsewhere in some way that I'm missing?

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -324,17 +326,17 @@ async def load(self, protocol_source: ProtocolSource) -> None:
...
         initial_home_command = pe_commands.HomeCreate(
             params=pe_commands.HomeParams(axes=None)
         )
         # this command homes all axes, including pipette plugner and gripper jaw
         self._protocol_engine.add_command(request=initial_home_command)
 
         for command in commands:
-            self._protocol_engine.add_command(request=command)
-            await _yield()
```
SyntaxColoring, line 335: Are we intentionally removing this yield?

Actually, wait, do we still need to `.append()` one-by-one? Would it work to just do `self._queued_commands = commands`?

mjhuff (approved):
Disregard my terrifying comments, tested and this does work. App & ODD don't appear to be affected as well!

api/src/opentrons/protocol_runner/protocol_runner.py
TamarZanzouri: woops :-)

robot-server/tests/integration/http_api/runs/test_json_v6_protocol_run.tavern.yaml
TamarZanzouri: The goal is to match python protocols so yes, we will need to document this change

SyntaxColoring: 👍 Discussed with @sfoster1 in-person. Recap:

The downside of this change is that it makes it more difficult for a simple client, uploading simple JSON protocols, to render a run log.

The upsides of this change are that it brings uniformity with Python protocols, and that it *forces* clients to acknowledge the possibility that the queue will change—which, with fixup commands and retries, it will more often.

We'll make this change.

api/src/opentrons/protocol_runner/protocol_runner.py
TamarZanzouri: Setup commands are added to the queue in PE but are being executed after the initial home commands instead of at the end of the run. adding prioritization will happen in a follow up pr. 
> Prioritization of "intent": "setup" commands
But according to the tests, this is still working, somehow?

> Special handling of RunStoppedError to break instead of raising an error
Pausing on recoverable errors (added in https://github.com/Opentrons/opentrons/pull/14646)
You can see https://github.com/Opentrons/opentrons/pull/14753 for how I'm doing this for Python protocols. If we're making JSON behave more like Python now, maybe this should work the same way.

Will add logic in this pr

> Yielding to the event loop on each command

I do not think we need this bc `add_and_execute_command` is async and `add_command` is synch but I will test this to make sure we hare not blocking anything.

SyntaxColoring: > I do not think we need this bc add_and_execute_command is async and add_command is synch but I will test this to make sure we hare not blocking anything.

We do need it, unfortunately, for the reasons described in this comment in the old implementation:

https://github.com/Opentrons/opentrons/blob/93226572668e81d111e49ea99b9a7663b3effb63/api/src/opentrons/protocol_engine/execution/queue_worker.py#L80-L83

[Unlike JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#:~:text=Even%20when%20the%20used%20promise%20is%20already%20fulfilled%2C%20the%20async%20function%27s%20execution%20still%20pauses%20until%20the%20next%20tick.), Python's `await` isn't guaranteed to yield to the event loop. It will only do so when it calls something that goes back to the event loop, like network I/O or an `asyncio.sleep()`.

----

Other than that, that all makes sense, thanks!

TamarZanzouri: I see. I tested this by adding a background task that prints, then `time.sleep(3)` and was able to see the printing sequence while executing the commands. so this test is not efficient? @SyntaxColoring @sfoster1

api/tests/opentrons/protocol_runner/test_protocol_runner.py
TamarZanzouri: So much better!

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -355,6 +356,16 @@ async def run(  # noqa: D102
         commands = self._protocol_engine.state_view.commands.get_all()
         return RunResult(commands=commands, state_summary=run_data, parameters=[])
 
+    async def _add_command_and_execute(self) -> None:
+        print(self._queued_commands)
```
sfoster1: accidentally committed?

sfoster1 (changes requested):

api/src/opentrons/protocol_runner/protocol_runner.py
sfoster1: I don't think we need the await because, crucially, what we're talking about here is in a different async stack than the lines you linked to @SyntaxColoring . This PR does _not_ replace or even touch the execution queue worker.

The way the engine works, just so we agree, is that there's fundamentally at least three async stacks: the execution stack, which is controlled by the execution queue linked above and owned by the engine, and `await`s command implementations; the runner task-queue stack, which is what is changed here and owned by the runner, and (now) dispatches commands via `add_and_execute_command`; and the stack that calls the runner's primary interface, which is I think owned by the server's run data manager.

The execution stack, as you mention, definitely needs to have an explicit yield in case it's running a bunch of commands that have no yield points... so it does, it's in what you linked above, which is still what's used and which this PR does not touch.

The server stack does not need to have an explicit yield as long as the runner uses the task queue, since if the runner uses a task queue it `await self._task_queue.join()`, which awaits a future, which is a yield point.

The task queue stack (which is what's changed here) does not need to have an explicit yield if it's using something that eventually uses `wait_for` because `wait_for` `await (asyncio.Event())`s, which is a sync point.

SyntaxColoring: Ah, that makes sense, thanks. I missed that this was still going through the `QueueWorker` under the hood.

SyntaxColoring: Which means I was also wrong about this:

> Special handling of RunStoppedError to break instead of raising an error

Per the documentation of `ProtocolEngine.add_and_execute()`, if the run is stopped, it actually *doesn't* raise a `RunStoppedError`—it returns the command still `queued`.

api/src/opentrons/protocol_runner/protocol_runner.py
TamarZanzouri: I left this for my testing :-) still working on adding tests

api/src/opentrons/protocol_runner/protocol_runner.py
TamarZanzouri: @SyntaxColoring 
I had a feeling but I wanted to prove that through a test I am trying to add :-) 
> Which means I was also wrong about this:

TamarZanzouri: @SyntaxColoring I added e2e tests to prove that python and json protocols now behave the same when stopping a protocol mid run

sfoster1 (approved):
Looks excellent!

robot-server/tests/integration/http_api/runs/test_play_stop_papi.tavern.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+  - name: Stop the run
+    request:
+      url: '{ot2_server_base_url}/runs/{run_id}/actions'
+      method: POST
+      json:
+        data:
+          actionType: stop
+    response:
+      status_code: 201
```
SyntaxColoring, line 65: We can't be sure where in the protocol this `stop` action will come in. For example, this might stop the run before or after the initial home command. The reason this matters is that we're asserting a specific sequence of commands in the "Get run commands" stage, down below, so this will be flaky depending on timing.

Ideally, we would wait for a specific command (a `waitForResume` or long-running `waitForDuration`) to be active before issuing the `stop` action. This might be easier to do in Python than in Tavern.

robot-server/tests/integration/http_api/runs/test_play_stop_papi.tavern.yaml
```diff
@@ -0,0 +1,114 @@
...
+      status_code: 200
+      strict:
+        - json:off
+      json:
+        data:
+          status: stopped
+
+  - name: Get run commands
+    max_retries: 10
+    delay_after: 0.2
```
SyntaxColoring: I don't think this `delay_after` is doing anything, is it? You might have meant `delay_before`—but even then, the stage above has already made sure that the run is stopped.

SyntaxColoring (approved):
This looks good to me to merge, thanks. Here are some comments on the tests that we've talked about in Slack.

robot-server/tests/integration/http_api/runs/test_play_stop_papi.tavern.yaml
```diff
@@ -39,6 +39,21 @@ stages:
     response:
       status_code: 201
 
+  - name: Wait for the run to complete
```
SyntaxColoring:
```suggestion
  - name: Wait for the command to run
```

SyntaxColoring (approved):
Nice, TY!

SyntaxColoring:
Oh, and I renamed this from `chore: ...` to `feat: ...` because it's a deliberate change in the HTTP API, and we'll have to remember to describe it in the release notes.

PR #14762, opened 2024-04-01
fix(robot-server): maintain correct order of protocol analyses

sanni-t:
Closes AUTH-229

# Overview

Updates the `/protocols` endpoints to always maintain the order of list of analyses as most-recently-started-analysis last, making sure to verify if a new analysis needs to be triggered because of new run-time-parameter values for a previously uploaded protocol.

To achieve that, this PR does the following things:
1. updates the analysis table in the database to include a new column where we save info about the RTP values used in the analysis in question.
2. adds functionality to read and write those values to/ from database
3. adds functionality to compare the new RTP values (if any) sent by the client, with those used in the most recent analysis.
4. updates the protocol upload endpoint to:
   - If the protocol did not exist before: create a new protocol resource, start a new analysis on that resource (existing functionality)
   - If the protocol exists and its most recent analysis used the same parameters' values as the ones in current request, do not create a new protocol resource, and do not trigger a new analysis
   - If the protocol exists and its most recent analysis used *different* parameters' values than the ones in current request, do not create a new protocol resource, but trigger a new analysis using the new RTP override values.

# Test Plan
(Can use test protocol from [this PR](https://github.com/Opentrons/opentrons/pull/14735))

- [ ] Upload a protocol with Run-time parameters defined, but no RTP values specified. Verify that analysis contains correct RTP values (should be the default values)
- [ ] Upload the same protocol, without any RTP values again, verify that there is still only 1 analysis for the protocol and that's the one from before
- [ ] Upload same protocol, with some RTPs explicitly set to default values in the request. Verify that there's still only the one analysis
- [ ] Upload same protocol with some RTPs set to non-default values in the request. Verify that a new analysis is added to the list of analyses and this analysis uses the new RTP values
- [ ] Upload same protocol with no RTP values specified again. Verify that another analysis is appended to the end of the list of analyses and this one uses the default RTP values.

# Review requests

- Make sure code & comments make sense
- See if I've missed any cases

# Risk assessment

Medium. Does database update and fixes the analysis order that was broken by #14688


robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -258,6 +267,94 @@ async def get_by_protocol(self, protocol_id: str) -> List[ProtocolAnalysis]:
...
+        So, we check that the Run Time Parameters in the previous analysis has params
+        with the values provided in the current request, and also verify that rest of the
+        parameters in the analysis use default values.
+        """
+        analyses = self.get_summaries_by_protocol(protocol_id)
+
+        # We only check for matching RTP values if the given protocol ID
+        # (& hence its summary) exists. So this assertion should never be false unless
+        # somehow a protocol resource is created without an analysis record.
+        assert len(analyses) > 0, "This protocol has no analysis associated with it."
```
SyntaxColoring: This *can* happen, can't it?

1. Upload a protocol. This starts an analysis in the background.
2. Shut off the robot before the analysis completes.
3. Power on the robot. The protocol will exist without any analyses.

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -258,6 +267,94 @@ async def get_by_protocol(self, protocol_id: str) -> List[ProtocolAnalysis]:
...
+
+        last_analysis = analyses[-1]
+        assert (
+            last_analysis.status != AnalysisStatus.PENDING
+        ), "Protocol must not already have a pending analysis."
```
SyntaxColoring: This can also trigger, I think?

1. Upload a protocol. This starts an analysis in the background.
2. Before analysis completes, quickly upload the same protocol.

[Edit: Oh, yeah, and this is causing [this](https://github.com/Opentrons/opentrons/pull/14762#discussion_r1547980085).]

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -258,6 +267,94 @@ async def get_by_protocol(self, protocol_id: str) -> List[ProtocolAnalysis]:
...
+            # the client is sending a protocol for the first time or for the nth time.
+            return False
+        for (
+            parameter,
+            prev_value_and_default,
+        ) in rtp_values_and_defaults_in_last_analysis.items():
+            if new_rtp_values.get(parameter) == prev_value_and_default.value or (
+                new_rtp_values.get(parameter) is None
+                and prev_value_and_default.value == prev_value_and_default.default
+            ):
```
SyntaxColoring: Style nit: Is this simpler?

```suggestion
            if new_rtp_values.get(
                parameter, prev_value_and_default.default
            ) == prev_value_and_default.value:
```

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
```diff
@@ -107,3 +107,26 @@ stages:
               status: completed
             - id: !anystr
               status: pending
+
+  - name: Check that uploading the same protocol with run-time parameter values triggers re-analysis
```
SyntaxColoring: Is there a difference between this and the test above that I'm missing, or is it accidentally duplicated?

robot-server/tests/integration/http_api/protocols/test_key.tavern.yaml
```diff
@@ -169,6 +169,8 @@ stages:
             author: engineer@opentrons.com
           key: duplicate_key
   - name: Upload basic_transfer_standalone protocol with same key
+    # add a delay before starting to let previous analysis complete
+    delay_before: 2
```
SyntaxColoring, line 173: This seems like a red flag. Since we're not looking at analysis resources in this test, what happens if we omit this?

robot-server/tests/protocols/test_analysis_store.py
```diff
@@ -6,6 +6,8 @@
 from typing import List, NamedTuple
 
 import pytest
+from decoy import Decoy
+from opentrons.protocol_engine.types import RunTimeParamValuesType
```
SyntaxColoring, line 10: I'm just noticing this, and this is a nitpick, but does `RunTimeParamValuesType` need to be an `opentrons.protocol_engine` concern? It looks like the runner is the deepest that it goes, which makes sense.

SyntaxColoring (commented):
Yay!

robot-server/tests/integration/http_api/protocols/test_key.tavern.yaml
SyntaxColoring: Oh I see, it's the `AssertionError` in https://github.com/Opentrons/opentrons/pull/14762#discussion_r1547901391. Discussing there.

robot-server/robot_server/protocols/analysis_store.py
SyntaxColoring: OK so correct me if I'm wrong, but as far as I can tell, the background is this: You don't want to start a redundant analysis. To determine if an analysis is redundant, you need the overrides supplied to the last analysis, and the protocol's default values when no overrides are supplied. To get the protocol's default values, you need some analysis to have *completed.*

robot-server/tests/protocols/test_analysis_store.py
sanni-t: I've been thinking about that too. These RTP types were added in the very first PR of RTP work so it made most sense to put it in there then. 
We've not needed to have separate types collection for the runner so far but maybe it's time.

robot-server/robot_server/protocols/analysis_store.py
SyntaxColoring: If that's right, then it seems like a solvable problem.

What we want, ideally, is for the server to be able to get a protocol's available parameters and their default values *quickly*—say, within 1–2 seconds—without having to do a full analysis with `aspirate()`s and `dispense()`s and everything.

So when a client does `POST /protocols`:

* If there are no analyses for that protocol yet:
    * Start one.
* If the last analysis is completed:
    * And it has the same runtime parameters as this `POST` request:
        * Keep that one.
    * And it has different runtime parameters from this `POST` request:
        * Start a new one.
* If the last analysis is pending:
    * **Wait until we have details about the protocol's runtime parameters, which should only take 1–2 seconds.** Then, handle as above.

This is all seems readily possible to me. The run-time parameters Python Protocol API was designed for it, with its isolated `add_parameters()` function, which we can call without calling `run()`.

Does that seem doable?

sanni-t: That's a good idea. Maybe we can aim for it in the long term (or even medium term) but it's not doable in the current architecture of the protocol runner and analyzer. Even though `add_parameters()` is separate from `run()`, for a few reasons discussed previously, we decided to make handling the actual parameters parsing and setting as part of protocol execution (inside the simulating runner in case of analysis) as opposed to handling it inside an intermediate layer like the protocol reader. Which means that the RTP information is not available until the analysis is complete.

Let's say we refactor the runner to make the protocol's RTP values accessible before the analysis is completed, we still have the protocol analysis & analysis store working on this binary concept of 'pending' vs 'completed' analysis where a protocol's info is only available in the CompletedAnalysisStore. For the 'pending with RTP values parsed' state, we will need to introduce a new functionality for either the pending store or just redo this concept of the binary states for analysis store.

Once we make even that change so that the server can now successfully cross-check against the most recent RTP values, we will have to decide what happens if the new request used different RTP values than the ones in a pending request; do you start another analysis right away and keep the last one running too? Or do you cancel the previous analysis? 

So there's a lot of changes that we will need to make in order to implement the behavior you are describing.
I agree that it's a better behavior though. I'm going to make a ticket for it. But until then, I think it's totally fine to say that you will need to wait for the current analysis to complete before reuploading the protocol. I believe the app makes it nearly impossible to re-upload a protocol anyway because of how runs are created. So I don't think it's a case you would run into easily.

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
sanni-t: Oops. Definitely duplicate.
Actually triplicate! I meant to remove this test from here since we're testing this behavior in the `test_analyses_with_run_time_parameters.tavern` tests.

robot-server/robot_server/protocols/analysis_store.py
SyntaxColoring: OK, your call!

robot-server/robot_server/protocols/analysis_store.py
sanni-t: That looks right. Previously, that would result in the responses providing an empty list of analyses while now it would error out. I feel like raising an error is slightly better than just returning an empty list, but both approaches make the protocol resource unusable without having a way to force reanalysis.

Hmm.. it might be better to return `False` here so that it triggers a new analysis. There shouldn't be any risk of corrupting the database if we do that, ya?

SyntaxColoring: Yeah, returning `False` and triggering a new analysis seems like a good idea. No, I don't see it causing any database problems.

robot-server/robot_server/protocols/analysis_store.py
SyntaxColoring: In that case, though:

* If you think this is going to be the HTTP-client-facing behavior for the medium term, I would try to get it to return an intentional 503 busy response instead of a 500 internal server error.
* In any case, I would change this to not be an `assert`, since that implies this condition will never happen except by `robot-server` bugs, whereas it really depends on client behavior.
    *   If you're switching the response to HTTP 503 and settling in for this to be the HTTP API, a tri-state "yes"/"no"/"busy" return value seems appropriate.
    * If you're keeping the response as an HTTP 500 and calling this a known bug, I actually think a `NotImplementedError` would be correct?

sanni-t: I like that. I'll go with the 503 busy response.

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -309,21 +317,12 @@ async def matching_rtp_values_in_last_analysis(
...
+        if analysis_summary.status == AnalysisStatus.PENDING:
+            raise AnalysisIsPendingError(analysis_summary.id)
 
         rtp_values_and_defaults_in_last_analysis = (
             await self._completed_store.get_rtp_values_and_defaults_by_analysis_id(
-                last_analysis.id
+                analysis_summary.id
             )
         )
         assert (
```
SyntaxColoring, line 331: Does this assert still make sense? It looks like a second instance of https://github.com/Opentrons/opentrons/pull/14762#discussion_r1547896892?

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
```diff
@@ -26,6 +26,22 @@ stages:
...
+    response:
+      strict:
+        - json:off
+      status_code: 503
+      json:
+        errors:
+          - id: "LastAnalysisPending"
+            title: "Last Analysis Still Pending."
+            errorCode: '4000'
+
```
SyntaxColoring: This will be flaky if the server completes the analysis before it receives this request, right? If so, I would probably not add this test.

SyntaxColoring (commented):
A question on another `assert` and on a Tavern test. Otherwise, this looks good to me. Thanks!

robot-server/robot_server/protocols/analysis_store.py
sanni-t: It's needed for type checking reasons. `completed_store.get_rtp_values_and_defaults_by_analysis_id` returns a `None` if the analysis ID doesn't exist. So, yes, technically it's the same underlying reason. And because this reason is now being handled, that assert shouldn't ever raise unless there's a bug in our code somewhere.

robot-server/tests/integration/http_api/protocols/test_analyses.tavern.yaml
sanni-t: Replaced with stable unit tests instead.

robot-server/robot_server/protocols/analysis_store.py
SyntaxColoring: OK, as long as you can continue to upload protocols to a robot where a protocol is missing an analysis!

SyntaxColoring (approved):
yeet

PR #14764, opened 2024-04-01
feat(protocol-designer): tuberack form warnings & warnings now dismissible

jerader:
closes AUTH-10

# Overview

This PR does 3 things:
1. fixes a very old bug introduced in 4.0.2 where form warnings are not dismissible by wiring them up to dispatch the correct redux action
2. fixes a bug in the warnings related to migrating to the pipette schema v2 and finding the `minVolume` correctly
3. adds a new form warning for pipetting into a tube rack

# Test Plan

Create an ot-2 or flex protocol. Add a tube rack to the deck map. Create a transfer step and try to aspirate and dispense into the tube rack. you should see a timeline warning. They should both be dismissible.

Now create a mix step and try to mix into the tube rack. There should be the timeline warning and it should be dismissible.

# Changelog

- wire up finding the min volume correctly in the warnings
- add the new warnings for asp, dispensing, and mixing into a tube rack
- fix the alert logic for dismissing the form warnings

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/steplist/formLevel/warnings.tsx
+++ b/protocol-designer/src/steplist/formLevel/warnings.tsx
@@ -1,16 +1,19 @@
 import * as React from 'react'
 import { getWellTotalVolume } from '@opentrons/shared-data'
 import { KnowledgeBaseLink } from '../../components/KnowledgeBaseLink'
-import { FormError } from './errors'
+import type { FormError } from './errors'
 /*******************
  ** Warning Messages **
  ********************/
 
 export type FormWarningType =
-  | 'BELOW_PIPETTE_MINIMUM_VOLUME'
-  | 'OVER_MAX_WELL_VOLUME'
-  | 'BELOW_MIN_DISPOSAL_VOLUME'
+  | 'ASPIRATE_TIP_POSITIONED_LOW_IN_TUBE'
   | 'BELOW_MIN_AIR_GAP_VOLUME'
+  | 'BELOW_MIN_DISPOSAL_VOLUME'
+  | 'BELOW_PIPETTE_MINIMUM_VOLUME'
+  | 'DISPENSE_TIP_POSITIONED_LOW_IN_TUBE'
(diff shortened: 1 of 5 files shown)


protocol-designer/src/steplist/formLevel/warnings.tsx
```diff
@@ -64,14 +88,57 @@ export type WarningChecker = (val: unknown) => FormWarning | null
...
   const { pipette, volume } = fields
   if (!(pipette && pipette.spec)) return null
-  return volume < pipette.spec.minVolume
-    ? belowPipetteMinVolumeWarning(pipette.spec.minVolume)
-    : null
+  const liquidSpecs = pipette.spec.liquids
+  const minVolume =
+    'lowVolumeDefault' in liquidSpecs
+      ? liquidSpecs.lowVolumeDefault.minVolume
+      : liquidSpecs.default.minVolume
```
jerader, line 140: I missed migrating these when i migrated to using pipette schema v2. The reason why is because `HydratedFormData` still needs to be typed correctly. We have been putting it off for years 😅

protocol-designer/src/steplist/formLevel/warnings.tsx
```diff
@@ -56,6 +59,27 @@ const belowMinDisposalVolumeWarning = (min: number): FormWarning => ({
   dependentFields: ['disposalVolume_volume', 'pipette'],
 })
 
+const aspirateTipPositionedLowInTube = (): FormWarning => ({
+  type: 'ASPIRATE_TIP_POSITIONED_LOW_IN_TUBE',
+  title:
+    'The default aspirate height is 1mm from the bottom of the well, which could cause liquid overflow or pipette damage. Edit tip position in advanced settings.',
```
jerader, line 65: i'll make a follow up ticket to move these title texts to using i18n

koji (approved):
the changes look good to me.

PR #14772, opened 2024-04-01
feat(app): button to launch quick transfer flow when FF is on

smb2268:
fix PLAT-170

# Overview
When the Quick Transfer FF is on, there should be a purple floating button on the bottom right of the ODD protocols tab. Right now, this just logs that the button has been pressed

# Test Plan
With Quick Transfer FF on, see the button on the bottom right of the protocols tab. With the FF off, see that the button is not displayed
Screen Shot 2024-04-01 at 11 54 12 AM

# Changelog
Add floating action button to protocols tab

# Review requests
Quick code check

# Risk assessment
Low

### DIFF
--- a/app/src/assets/localization/en/protocol_info.json
+++ b/app/src/assets/localization/en/protocol_info.json
@@ -65,6 +65,7 @@
   "protocol_title": "Protocol - {{protocol_name}}",
   "protocol_upload_failed": "Protocol upload failed. Fix the error and try again",
   "protocols": "Protocols",
+  "quick_transfer": "Quick transfer",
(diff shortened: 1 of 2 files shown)


jerader (commented):
question, the designs show that the button is on a new `quick transfer` screen?

Screenshot 2024-04-02 at 09 04 33

smb2268:
> question, the designs show that the button is on a new `quick transfer` screen?

@jerader yep that'll be implemented later on! we're building that screen out last, so in the interim I'm putting the button on the `protocols` tab so that we can launch the flow before building out the new tab.

jerader (approved):
lgtm!

PR #14773, opened 2024-04-01
fix(robot-server): Update status bar to account for `awaiting-recovery` run state

SyntaxColoring:
# Overview

Closes EXEC-363.

# Test Plan

* [x] Run a protocol with error recovery and make sure the status bar responds appropriately.

# Changelog

* Account for `EngineStatus.AWAITING_RECOVERY` in the light bar task.
* Convert the dict lookup to a `match` statement. This is probably slightly slower, but more type-safe, and just as concise.

# Review requests

None in particular.

# Risk assessment

Low.


DerekMaggio (approved):
LGTM, I like the use of the match

PR #14775, opened 2024-04-02
feat(app, shared-data): create odd boolean and choice selection screen

jerader:
closes AUTH-123

# Overview

Creates the screen for boolean and choice params on the odd

# Test Plan

upload a protocol with a boolean and choice param and send it the odd. Or upload the attached protocol and change line 187 of `app/src/organisms/ProtocolSetupParameters/index.tsx` to return the mockData const at the top of the file. 

Click on a boolean param on the setup parameters page. Observe the screen. Try to change the values and reset to default.

Now do the same thing with the choice param and make sure it works as expected.

# Changelog

- create `ChooseEnum` component and test
- create the logic to update the params in the parent `ProtocolSetupParameters`
- fix tests
- rename `formatRunTimeParameterValue` to `formatRunTimeParameterDefaultValue` and create a new `formatRunTimeParameterValue` that returns the value updates.

# Review requests

see test plan

# Risk assessment

low


app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  ? option === true
+                    ? t('true')
+                    : t('false')
+                  : option.displayName
+              }
+              buttonValue={
+                isBoolean
+                  ? option === true
+                    ? 'true'
+                    : 'false'
```
koji, line 92: not on / off?

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
```diff
@@ -0,0 +1,105 @@
...
+          const isBoolean = option === true || option === false
+          return (
+            <RadioButton
+              key={`${isBoolean ? option : option.value}`}
+              data-testid={`${isBoolean ? option : option.value}`}
+              buttonLabel={
+                isBoolean
+                  ? option === true
+                    ? t('true')
+                    : t('false')
```
koji: not on/off?

jerader: ooops you're right, thanks!

app/src/organisms/ProtocolSetupParameters/__tests__/ChooseEnum.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+describe('ChooseEnum', () => {
+  let props: React.ComponentProps<typeof ChooseEnum>
+
+  beforeEach(() => {
+    props = {
+      handleGoBack: vi.fn(),
+      parameter: {
+        displayName: 'Dry Run',
+        variableName: 'DRYRUN',
```
koji: nit
probably snake case would be normal in python. 
```suggestion
        variableName: 'dry_run',
```

app/src/organisms/ProtocolSetupParameters/__tests__/ChooseEnum.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+      rawValue: false,
+    }
+  })
+
+  it('should render the text and buttons work for a boolean param', () => {
+    render(props)
+    screen.getByText('Choose Dry Run')
+    const btn = screen.getByRole('button', { name: 'Restore default values' })
+    screen.getByText('True')
+    screen.getByText('False')
```
koji:
```suggestion
    screen.getByText('On')
    screen.getByText('Off')
```

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
```diff
@@ -0,0 +1,105 @@
...
+        flexDirection={DIRECTION_COLUMN}
+        paddingBottom={SPACING.spacing40}
+      >
+        <StyledText
+          as="h4"
+          textAlign={TYPOGRAPHY.textAlignLeft}
+          marginBottom={SPACING.spacing16}
+        >
+          {parameter.description}
+        </StyledText>
```
koji, line 70: if Sue requests us to keep the space size, we may need to add style.

https://www.figma.com/file/aS5owa8kwmA0uzqzkP1BLJ?type=design&node-id=3312-62720&mode=dev#755884946

app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -220,14 +247,18 @@ export function ProtocolSetupParameters({
         gridGap={SPACING.spacing8}
         paddingX={SPACING.spacing8}
       >
-        {parameters.map((parameter, index) => {
+        {newRunTimeParameters.map((parameter, index) => {
+          console.log(
+            'parameter',
+            parameter.type === 'boolean' ? parameter : undefined
+          )
```
koji: need to keep?

jerader: oops forgot to remove this, thanks!

app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -187,6 +192,30 @@ export function ProtocolSetupParameters({
     runTimeParametersOverrides,
     setRunTimeParametersOverrides,
   ] = React.useState<RunTimeParameter[]>(parameters)
+
+  const updateParameters = (
```
ncdiehl11, line 196: 👍 looks good

app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -230,14 +250,14 @@ export function ProtocolSetupParameters({
...
-        {parameters.map((parameter, index) => {
+        {runTimeParametersOverrides.map((parameter, index) => {
           return (
             <React.Fragment key={`${parameter.displayName}_${index}`}>
               <ProtocolSetupStep
                 hasIcon={!(parameter.type === 'boolean')}
                 status="general"
                 title={parameter.displayName}
-                onClickSetupStep={() => console.log('TODO: wire this up')}
+                onClickSetupStep={() => setChooseValueScreen(parameter)}
```
ncdiehl11: according to designs, I think boolean types should just be a toggle and should not navigate away from the list of parameters, unlike enumeration params and numerical params

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
```diff
@@ -0,0 +1,105 @@
...
+    )
+  }
+  const options = parameter.type === 'str' ? parameter.choices : [true, false]
+  const handleOnClick = (newValue: string | number | boolean): void => {
+    setParameter(newValue, parameter.variableName)
+  }
+
+  React.useEffect(() => {
+    if (rawValue === parameter.default) {
+      makeSnackbar(t('no_custom_values'))
```
ncdiehl11: When you enter the screen to update an enum param, we immediately render the snackbar. According to designs, it like we only want to render this snackbar when the user selects "Restore default value" on the set parameter screen, and the presently set value is equal to the default.

ncdiehl11 (changes requested):

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
jerader: oops good call! i missed that text

app/src/organisms/ProtocolSetupParameters/index.tsx
jerader: oooh actually ya that makes a lot of sense. thanks for calling this out. I guess there's no way to reset a single boolean param, like a user can change the setting but won't know if its the default or not

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  return (
+    <>
+      <ChildNavigation
+        header={t('choose_enum', { displayName: parameter.displayName })}
+        onClickBack={handleGoBack}
+        buttonType="tertiaryLowLight"
+        buttonText={t('restore_default')}
+        onClickButton={() => {
+          setParameter(parameter.default, parameter.variableName)
```
ncdiehl11: I think we want to check here whether the default equals the current value. If we actually are resetting the value, we don't show the snackbar. From figma:

```
Reset default value is disabled when the default value is displayed and enabled once its been changed

If the user taps on the ‘Reset default value’ button when disabled, a snack bar appears informing the user the value has not been edited
```

ncdiehl11 (changes requested):

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
jerader: wow omg thanks so much for pointing this out!

ncdiehl11 (approved):
works well! nice.

app/src/organisms/ProtocolSetupParameters/ChooseEnum.tsx
koji: I think we can work on char limit in a following PR because at this moment we don't have any long case design.

PR #14776, opened 2024-04-02
feat(robot-server): update run creation endpoint to accept runtime parameter values

jbleon95:
# Overview

Closes AUTH-130

This PR adds a new, optional argument `runTimeParameterValues` to the request body of the `POST /runs` endpoint to start a run with new runtime parameter values. It is the same shape as the argument for the `/protocols` endpoint (a dictionary/map of variable names to the parameter values), but since this endpoint does not take form data, it does not need to be parsed with in the body of the run create function.

These values are then fed to the `ProtocolRunner` load method which will bind those to the execute function, so that when the protocol starts it will evaluate and use those values when running the protocol. (assuming there are no errors in setting the parameter values).

Since re-analysis has not been implemented yet, when starting a run with overridden values, the frontend will show an incorrect analysis for the run. This will be addressed in a future PR so that it reflects the values (and therefore potentially labware, pipettes, modules used) used in the actual run.

# Test Plan

Continuing to use this test protocol seen in other PRs to test parameter values. Using postman, I was able to post the protocol with default values, then start a run with new overridden values:

- [x] Posting a run with no `runTimeParameterValues` uses the default values
- [x] Posting a run with valid overridden values uses that in the actual run execution
- [x] Posting a run with invalid overridden values causes the run to fail

```
metadata = {
    'protocolName': 'RTP Test 2 - Default only, all types',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}


    parameters.add_int(
        display_name="Sample count",
        variable_name="sample_count",
        default=6,
        minimum=1,
        maximum=12,
    )
    parameters.add_float(
        variable_name="volume",
        default=20.1,
        choices=[
            {"display_name": "Low Volume", "value": 10.23},
            {"display_name": "Medium Volume", "value": 20.1},
            {"display_name": "High Volume", "value": 50.5},
        ],
        unit="µL"
    )
    parameters.add_bool(
        display_name="Dry Run",
        variable_name="dry_run",
        default=False,
    )
    parameters.add_str(
        display_name="Pipette Name",
        variable_name="pipette",
        choices=[
        ],
        default="flex_1channel_50",
    )



def run(context):

        pipette.pick_up_tip()

        pipette.drop_tip()
```

# Changelog

- added `runTimeParameterValues` argument to `POST /runs`
- connected the runtime parameter values to the engine creation so that runs use those values sent

# Review requests

# Risk assessment

Low

### DIFF
--- a/robot-server/robot_server/runs/run_models.py
+++ b/robot-server/robot_server/runs/run_models.py
@@ -18,6 +18,7 @@ from opentrons.protocol_engine import (
     Liquid,
     CommandNote,
 )
+from opentrons.protocol_engine.types import RunTimeParamValuesType
 from opentrons_shared_data.errors import GeneralError
 from robot_server.service.json_api import ResourceModel
 from robot_server.errors.error_responses import ErrorDetails
@@ -211,6 +212,12 @@ class RunCreate(BaseModel):
     labwareOffsets: List[LabwareOffsetCreate] = Field(
         default_factory=list,
         description="Labware offsets to apply as labware are loaded.",
+    )
+    runTimeParameterValues: Optional[RunTimeParamValuesType] = Field(
+        None,
+        description="Key-value pairs of run-time parameters defined in a protocol."
+        " Also, if this data is included in the request, the server will"
+        " always trigger an analysis (for now).",
     )
 
 
--- a/robot-server/robot_server/runs/run_models.py
+++ b/robot-server/robot_server/runs/run_models.py
@@ -215,9 +215,7 @@ class RunCreate(BaseModel):
     )
     runTimeParameterValues: Optional[RunTimeParamValuesType] = Field(
         None,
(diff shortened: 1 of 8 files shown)


robot-server/robot_server/runs/run_models.py
```diff
@@ -212,6 +213,12 @@ class RunCreate(BaseModel):
         default_factory=list,
         description="Labware offsets to apply as labware are loaded.",
     )
+    runTimeParameterValues: Optional[RunTimeParamValuesType] = Field(
+        None,
+        description="Key-value pairs of run-time parameters defined in a protocol."
+        " Also, if this data is included in the request, the server will"
+        " always trigger an analysis (for now).",
```
sanni-t: I think you can skip this since it'll change very shortly.

sanni-t (approved):
Nice!

PR #14777, opened 2024-04-02
refactor(robot-server): Remove engine polling from RunsPublisher

mjhuff:
Closes [EXEC-310](https://opentrons.atlassian.net/browse/EXEC-310)

# Overview

Removes polling logic from `RunsPublisher`, making use of the new protocol engine event bubbling via `PublisherNotifier`.

There is a minor change to refetch conditions on the app side. In `ProtocolRunHeader`, runs are permanently displayed as `STOP_REQUESTED` if a run is cancelled manually and never progress to `CANCELLED`. After discussing with @shlokamin, it appears that it doesn't make much sense to stop polling during a `STOP_REQUESTED` state, since this is a non-terminal state. Although we _think_ we've had issues with this in the past, I've smoke tested the app-side, and it doesn't appear that there are any problems - notification architecture has changed a lot since we ran into this bug months ago, and the issue we experienced likely to be a more a up-stream problem.

# Test Plan
- General smoke test of running a protocol is sufficient. Verify that notifications still work.

# Risk assessment
low

[EXEC-310]: https://opentrons.atlassian.net/browse/EXEC-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/RunTimeControl/hooks.ts
```diff
@@ -79,11 +79,9 @@ export function useRunStatus(
     refetchInterval: DEFAULT_STATUS_REFETCH_INTERVAL,
     enabled:
       lastRunStatus.current == null ||
-      !([
-        RUN_STATUS_FAILED,
-        RUN_STATUS_SUCCEEDED,
-        RUN_STATUS_STOP_REQUESTED,
-      ] as RunStatus[]).includes(lastRunStatus.current),
+      !([RUN_STATUS_FAILED, RUN_STATUS_SUCCEEDED] as RunStatus[]).includes(
```
sfoster1: `RUN_STATUS_STOPPED` also - `STOPPED` is the terminal state for a protocol that was stopped by user action, e.g. cancelled.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -11,173 +10,104 @@
...
+    ) -> None:
         """Returns a configured Runs Publisher."""
         self._client = client
+        self._publisher_notifier = publisher_notifier
         self._run_data_manager_polling = asyncio.Event()
-        self._previous_current_command: Union[CurrentCommand, None] = None
-        self._previous_state_summary_status: Union[EngineStatus, None] = None
         self._poller: Optional[asyncio.Task[None]] = None
+        #  Variables and callbacks related to PE state changes.
+        self._run_id: Optional[str] = None
```
sfoster1: recommend binding these into an object, a dataclass maybe, that can be initialized or not as a whole. for instance
```
@dataclass
class RunHooks:
    run_id: str
    get_current_command: Callable[[str], Optional[CurrentCommand]]
    get_state_summary: Callable[[str], Optional[StateSummary]]

class RunsPublisher:
    def __init__(...):
        self._run_hooks: Optional[RunHooks] = None
        ...
    ...
    async def initialize(...):
        self._run_hooks = RunHooks(run_id, ...)
```

This means you don't need quite as many assert statements, and you get to express the semantics that you either have all of those three or none of them. 

You could do a similar thing with the the previous-element caches if you want.

sfoster1 (changes requested):
I think a cleanup thing app side and a small refactor server side are needed but as a whole looks great!

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
mjhuff: Good idea, thanks

app/src/organisms/RunTimeControl/hooks.ts
mjhuff: Oh duh, good catch

sfoster1 (approved):
Looks good to me, nice!

PR #14782, opened 2024-04-02
Automated ABR Calibration Data Uploading

rclarke0:
# Overview

Pulls Calibration Data from Robots and Uploads to google_drive/google_sheet

# Test Plan

Tested on ABR robots. Successfully pulls calibration data, uploads to google drive, and saves to google sheet.

# Changelog

- Adds abr_calibration_logs.py 

1. Connects to google drive folder
2. Connects to google sheet
3. Pulls module, instrument, and deck calibration data and compiles into one .json file per robot via http requests
4. Uploads new files to google drive folder
5. adds new rows to instrument, module, and deck calibration sheets if the serial and calibration lastmodified timestamp pairing do not already exist

- Split jira_tool up into a file with just jira_tools and a file that uses the tools with the robots.
- For all scripts uploading to google drive, changed the folder_name argument to folder_id so that the service_account is writing to the correct folder. Adds email as argument to allow for permission sharing by service account.

# Review requests

# Risk assessment

### DIFF
--- a/abr-testing/abr_testing/automation/google_drive_tool.py
+++ b/abr-testing/abr_testing/automation/google_drive_tool.py
@@ -1,6 +1,7 @@
 """Google Drive Tool."""
 import os
-from typing import Set, Any
+from typing import Set, Any, Optional
+import webbrowser
 from oauth2client.service_account import ServiceAccountCredentials  # type: ignore[import]
 from googleapiclient.discovery import build
 from googleapiclient.http import MediaFileUpload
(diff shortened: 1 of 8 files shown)


abr-testing/abr_testing/automation/google_drive_tool.py
```diff
@@ -72,8 +74,8 @@ def upload_file(self, file_path: str) -> str:
         """Upload file to Google Drive."""
         file_metadata = {
             "name": os.path.basename(file_path),
-            "mimeType": "application/vnd.google-apps.folder",
-            "parents": [self.parent_folder] if self.parent_folder else "",
+            "mimeType": "application/json",
```
sfoster1: definitely shouldn't be folder, but should this maybe depend on the kind of file? You can use the standard library [`mimetypes.guess_type()`](https://docs.python.org/3.10/library/mimetypes.html#mimetypes.guess_type) for this.

sfoster1 (approved):
This is great!

abr-testing/abr_testing/automation/google_drive_tool.py
rclarke0: changed to mimetypes.guess_type! Thanks for suggestion

PR #14783, opened 2024-04-02
fix(app): fix excessive /runs network requests

mjhuff:
Closes [EXEC-255](https://opentrons.atlassian.net/browse/EXEC-255)

# Overview

Fixes useNotifyService to utilize hostname instead of the `host` object, preventing a pass by reference issue causing excessive requests sent to `/runs` that often occurs on the RobotDetails page. This fix is sufficient, since this is all we need from `host` anyway. Let's also ensure that if a robot loses connection while a component has passed a callback to `appShellListener`, we ensure we remove the correct callback from the store.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/38bacdbc-b56f-466e-836f-a1f66bb18943

### Fixed Bevahior

https://github.com/Opentrons/opentrons/assets/64858653/0d9513c4-31b4-4004-9c4e-3909cf04a607

# Test Plan
- Verify the fix by going to the Network tab and clicking on a robot in the Devices page.

# Changelog
- Fixed excessive network requests to /runs.

# Risk assessment
low

[EXEC-255]: https://opentrons.atlassian.net/browse/EXEC-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/resources/useNotifyService.ts
+++ b/app/src/resources/useNotifyService.ts
@@ -43,6 +43,7 @@ export function useNotifyService<TData, TError = Error>({
   const doTrackEvent = useTrackEvent()
   const isFlex = useIsFlex(host?.robotName ?? '')
   const hasUsedNotifyService = React.useRef(false)
+  const seenHostname = React.useRef<string | null>(null)
   const { enabled, staleTime, forceHttpPolling } = options
 
   const shouldUseNotifications =
@@ -62,6 +63,7 @@ export function useNotifyService<TData, TError = Error>({
       })
       dispatch(notifySubscribeAction(hostname, topic))
       hasUsedNotifyService.current = true
+      seenHostname.current = hostname
     } else {
       setRefetchUsingHTTP('always')
     }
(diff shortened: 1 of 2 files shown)


app/src/resources/useNotifyService.ts
```diff
@@ -62,21 +63,22 @@ export function useNotifyService<TData, TError = Error>({
...
-          hostname: hostname as string,
+          hostname: seenHostname.current as string,
           topic,
           callback: onDataEvent,
           isDismounting: true,
         })
       }
     }
-  }, [topic, host, shouldUseNotifications])
+  }, [topic, hostname, shouldUseNotifications])
```
DerekMaggio, line 81: Was `host` changing because it was a new object every time this was called? So when it was compared against the previous object React `Object.is` would fail causing a rerender?

mjhuff: Yeah exactly. So the reason this isn't an issue 99% of the time is we use `hostFromProvider` and not `hostOverride`, see L41. `hostFromProvider` is the same object every time, but `hostOverride` is not. That's why this issue only happens in a couple spots.

sfoster1 (approved):
Yup, looks good to me.

PR #14784, opened 2024-04-02
fix(app,components): fix module controls no module connected case

koji:
# Overview
While I was working on Protocol Run page for RTP, I noticed that `Connect modules to see controls` is different from No Parameters even they are in the same page.

Mel updated the design.
https://www.figma.com/file/l7BAJAGICr10ELsSqQD0Sr/Design-System%3A-Desktop?node-id=14096%3A7516&mode=dev

Rename `NoParameters` to `InfoScreen` to utilize one component for both cases.

https://www.figma.com/file/l7BAJAGICr10ELsSqQD0Sr/Design-System%3A-Desktop?node-id=13146%3A50213&mode=dev#757569773

close AUTH-264

[before]
Screenshot 2024-04-02 at 5 15 13 PM

[after]
Screenshot 2024-04-02 at 5 41 33 PM

# Test Plan
0. `make teardown-js && make setup-js`
1. Setup a protocol that requires a module with a robot that doesn't attach any module or remove modules from text-flex.json
2. Click `Module Controls`

# Changelog
- Update NoPrameter component and its test
- Update Module Controls padding

# Review requests
If you have any better name for the component, please let me know.

# Risk assessment
low


ncdiehl11 (approved):
This looks good. Question for the future— how do we plan to handle translations for these kinds of components?

koji:
> This looks good. Question for the future— how do we plan to handle translations for these kinds of components?

Actually, I've been thinking about that.
Ideally just passing text to the component would be good. But for that probably we would need to unify this component for web and odd. 


Also, I think solution would depend on UI components' goal since @opentrons/components is published.
https://www.npmjs.com/package/@opentrons/components

PR #14795, opened 2024-04-03
refactor(protocol-engine): Keep track of failed commands' error recovery types

SyntaxColoring:
# Overview

This goes towards (but does not close) EXEC-346.

# Test Plan

We're sufficiently covered by automated tests.

# Changelog

* When a command fails, and we choose its recovery type (`WAIT_FOR_RECOVERY` or `FAIL_RUN`), store that recovery type for later retrieval.

  We need this for EXEC-346 because when the Python protocol thread sees a `failed` Protocol Engine command, something needs to tell it whether to raise an exception to immediately fail the run, or wait until an HTTP client issues a `resume-from-recovery` action.

  That's not done in this PR. It will probably happen in #14753.

* As discussed a number of times on the RSS team: Deprecate the old unit testing style of testing `CommandView` and `CommandStore` independently. Instead, test them together. We already do this for `TipView` and `TipStore`.

  New tests should be added to the new file `test_command_state.py`. The old test files have been renamed to `*old.py` but are otherwise left in-place. We should gradually shift things to the new style as we continue to work in this area.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -730,6 +730,13 @@ class CommandView(HasState[CommandState]):
                 message=failed_command.command.error.detail,
             )
 
+    def get_error_recovery_type(self, command_id: str) -> ErrorRecoveryType:
+        """Return the error recovery type with which the given command failed.
+
+        The command ID is assumed to point to a failed command.
+        """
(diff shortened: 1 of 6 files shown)


sfoster1 (approved):
Makes sense!

PR #14796, opened 2024-04-03
refactor(api): Rename opentrons.commands to opentrons.legacy_commands

SyntaxColoring:
# Overview

My editor search keeps wanting to open `opentrons.commands`, which is old and vibeless and misleading. The *actual* commands that most of us need to care about are in `opentrons.protocol_engine.commands`.

# Test Plan

- [x] I've looked for public-facing mentions of `opentrons.commands` in the Python Protocol API and could only find one, which looks safe to change.

# Review requests

Is this a good idea?

# Risk assessment

Low.


DerekMaggio (commented):
I think it is a good idea. But I am biased because this exact thing was annoying me beyond belief yesterday. 
Oooooh maybe add a deprecation note with Seth's fancy new Note stuff?

sfoster1 (approved):
📳 📈

SyntaxColoring:
> Oooooh maybe add a deprecation note with Seth's fancy new Note stuff?

Interesting—like if a protocol tries to import `opentrons.commands`, add a warning note somewhere on the run?

I think that's a good idea and something that the system should be able to support. For prioritization reasons, I don't think I'm going to pursue it right now, but I'll keep that use-case in mind as we build out the notes system.

PR #14797, opened 2024-04-03
added functions to count module commands per run

rclarke0:
# Overview

Functions to Count Module commands per run

# Test Plan

- looked at run logs and used cmd f to double check command counts/times

# Changelog

Added a function for the thermocycler, temperature module, and heater shaker to count values of interest for lifetime test comparison
Added those dictionaries to larger dictionary to be included on run sheet
# Review requests

# Risk assessment

- These functions are not set up to handle multiples of the same module in a protocol. It will group total commands together
- some modules do not deactivate at the end of the run. To get total on time, the protocol completedAt timestamp is used.


sfoster1 (approved):
Nice!

PR #14801, opened 2024-04-04
test: Fix 2.17 smoke test

DerekMaggio:
# Overview

I think my logic behind making the 2.17 smoke test blank was wrong.
I think it just needs to be a straight copy of 2.16 smoke test with the api version updated to 2.17

Pulled from protocol
```
# This protocol is exactly the same as 2.16 Smoke Test V3
# The only difference is the API version in the metadata
# There were no new positive test cases for 2.17
# The negative test cases are captured in the 2.17 dispense changes protcol
```

### DIFF
--- a/app-testing/files/protocols/py/OT2_P300M_P20S_TC_HS_TM_2_17_SmokeTestV3.py
+++ b/app-testing/files/protocols/py/OT2_P300M_P20S_TC_HS_TM_2_17_SmokeTestV3.py
@@ -25,6 +25,7 @@ requirements = {"robotType": "OT-2", "apiLevel": "2.17"}
 #### LOOK AT THIS #######
 #########################
 
+
 def run(ctx: protocol_api.ProtocolContext) -> None:
     """This method is run by the protocol engine."""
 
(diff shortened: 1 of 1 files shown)


nusrat813 (approved):
LGTM

app-testing/files/protocols/py/OT2_P300M_P20S_TC_HS_TM_2_17_SmokeTestV3.py
```diff
@@ -3,23 +3,370 @@
...
+
+
+#########################
+#### LOOK AT THIS #######
+#########################
+
+# This protocol is exactly the same as 2.16 Smoke Test V3
+# The only difference is the API version in the metadata
+# There were no new positive test cases for 2.17
+# The negative test cases are captured in the 2.17 dispense changes protcol
```
y3rsh, line 22: I like this note and I think we retain this change log at the top as well as use comments below to show API changes in the code with links to the doc.

y3rsh (approved):
Good catch.

PR #14802, opened 2024-04-04
feat(protocol-designer): add alert if x/y position is too close to edge

jerader:
closes AUTH-252

# Overview

Adds a PD alert banner to the tip position modal if the user adds an x or y position that is too close to the edge of the well

# Test Plan

Create a flex or ot-2 protocol. Add a labware. Add a transfer or mix step. In the advanced settings, select the tip position icon and open the tip position modal. Add a custom offset with an x or y position that is at least 90% of the max/min positions. You should see the PD alert banner pop up. It should go away if you change your value

# Changelog

- add alert banner to pop up if the x or y value is >90% of the min or max
- add test cases

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepEditForm/fields/TipPositionField/TipPositionModal.tsx
+++ b/protocol-designer/src/components/StepEditForm/fields/TipPositionField/TipPositionModal.tsx
@@ -21,6 +21,7 @@ import styles from './TipPositionInput.module.css'
 import * as utils from './utils'
 
 import type { StepFieldName } from '../../../../form-types'
+import { PDAlert } from '../../../alerts/PDAlert'
 
 type Offset = 'x' | 'y' | 'z'
 interface PositionSpec {
@@ -57,7 +58,9 @@ export const TipPositionModal = (
   const { t } = useTranslation(['modal', 'button'])
 
   if (zSpec == null || xSpec == null || ySpec == null) {
-    console.error('expected to find specs for the zPosition but could not')
+    console.error(
+      'expected to find specs for one of the positions but could not'
+    )
   }
 
   const defaultMmFromBottom = utils.getDefaultMmFromBottom({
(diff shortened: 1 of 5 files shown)


protocol-designer/src/components/StepEditForm/fields/TipPositionField/TipPositionModal.tsx
```diff
@@ -218,6 +221,12 @@ export const TipPositionModal = (
   ): void => {
     handleYChange(e.currentTarget.value)
   }
+  const xValueNearEdge =
+    xValue != null &&
+    (parseInt(xValue) > 0.9 * xMaxWidth || parseInt(xValue) < 0.9 * xMinWidth)
```
koji: nit 

maybe define a const for `0.9`?

like
```ts
const EDGE_THRESHOLD = 0.9
```

jerader: oh ya good idea!

protocol-designer/src/components/StepEditForm/fields/TipPositionField/TipPositionModal.tsx
```diff
@@ -218,6 +221,12 @@ export const TipPositionModal = (
   ): void => {
     handleYChange(e.currentTarget.value)
   }
+  const xValueNearEdge =
```
koji: not really good but `isXValueNearEdge` can be easy to understand that the const type is boolean?

koji:
Screenshot 2024-04-04 at 12 49 45 PM

After changing the position, still the banner is staying in the modal. Is this expected behavior?

The changes look good to me but I was wondering, maybe changing the range would be useful since Done button is disabled when use the value that is close to edge.


https://github.com/Opentrons/opentrons/assets/474225/c5fba060-aa88-4299-bf6c-b4c506d9bbe1

jerader:
> After changing the position, still the banner is staying in the modal. Is this expected behavior?

@koji the banner should display if the value inputted is greater than 90% of the max value/min value. But you should be allowed to submit as long as it is below the max value. I think i need to modify the min/max to only display up to 1 decimal value (that is why you could not click done because you added 2 decimals). I will address this, thank you for reviewing!

koji (approved):
thank you for the explanation.
The changes make sense to me.

PR #14803, opened 2024-04-04
feat(app, components): fix ProtocolDetails ParametersTable

ncdiehl11:
# Overview

Update styling of `ParametersTable` component used in desktop `ProtocolDetails`

# Test Plan

- upload protocol containing RTPs 
- click protocol to open protocol details
- verify that you land on 'Parameters' tab
- verify that hovering a parameter's info icon renders a tooltip containing parameter description text
- verify that 'Range' for choice parameter containing 2 options displays "option 1, option 2"
- verify that 'Range' for choice parameter containing >2 options displays "[#options] options"

Screenshot 2024-04-04 at 12 00 03 PM

# Changelog

- refactor `ParametersTable` component to include hoverable icon for tooltip description
- update rendering of choice parameter range depending on number of choices
- add translation key for number of options
- add tests

# Review requests

authorship devs

# Risk assessment

low

### DIFF
--- a/components/src/molecules/ParametersTable/index.tsx
+++ b/components/src/molecules/ParametersTable/index.tsx
@@ -1,9 +1,13 @@
 import * as React from 'react'
 import styled from 'styled-components'
 import { formatRunTimeParameterDefaultValue } from '@opentrons/shared-data'
-import { BORDERS } from '../../helix-design-system'
+import { BORDERS, COLORS } from '../../helix-design-system'
 import { SPACING, TYPOGRAPHY } from '../../ui-style-constants/index'
 import { StyledText } from '../../atoms/StyledText'
+import { Tooltip, useHoverTooltip } from '../../tooltips'
+import { Icon } from '../../icons'
+import { Flex } from '../../primitives'
+import { ALIGN_CENTER } from '../../styles'
 
 import type { RunTimeParameter } from '@opentrons/shared-data'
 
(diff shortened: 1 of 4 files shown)


components/src/molecules/ParametersTable/index.tsx
```diff
@@ -85,6 +94,46 @@ export function ParametersTable({
...
+                name="information"
+                size={SPACING.spacing12}
+                color={COLORS.grey60}
+                data-testid={`Icon_${index}`}
+              />
+            </Flex>
+            <Tooltip
+              {...tooltipProps}
+              backgroundColor={COLORS.black90}
+              width={'8.75rem'}
```
jerader: nit 
```suggestion
              width="8.75rem"
```

components/src/molecules/ParametersTable/index.tsx
```diff
@@ -28,18 +32,20 @@ export function ParametersTable({
       'choices' in runTimeParameter ? runTimeParameter.choices : []
     const count = choices.length
 
+    if (count > 0) {
+      return count > 2
+        ? t != null
+          ? t('num_options', { num: count })
+          : `${count} options`
+        : choices.map(choice => choice.displayName).join(', ')
+    }
```
jerader, line 41: why can't this go in the switch anymore?

jerader (commented):
The `ParametersTable` is gonna be used in PL as well so might be good for @y3rsh to also take a look. But its looking good! just left a few comments

components/src/molecules/ParametersTable/index.tsx
ncdiehl11: I think because we can have choice parameters of any type. If you have an enumerated `int` for example, you need to check for `choices` rather than `min/max`. I saw an issue where I had a protocol with `int` type parameter with choices, and the range displayed as "0-0"

jerader: oh i see,  can we add a default to the switch then? to count for the str type. The default can return an empty ''.

jerader (approved):
lgtm! 🦓

components/src/molecules/ParametersTable/index.tsx
koji: I like ternary but we avoid nested ternary. I would suggest you to use `if-else`.

```ts
if (count > 0) {
  if (count > 2) {
    return t !=null ? t('num_options', { num: count }) : `${count} options`;
  } else {
    return choices.map(choice => choice.displayName).join(', ');
  }
}
```

PR #14805, opened 2024-04-04
chore(pydantic): V1->V2 migration

ahiuchingau:
Superseded by #14871.

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/utils.ts
+++ b/protocol-designer/src/components/modals/CreateFileWizard/utils.ts
@@ -67,12 +67,12 @@ export const getCrashableModuleSelected = (
 
 export const MOVABLE_TRASH_CUTOUTS = [
   {
-    value: 'cutoutA1',
-    slot: 'A1',
-  },
-  {
     value: 'cutoutA3',
     slot: 'A3',
+  },
+  {
+    value: 'cutoutA1',
+    slot: 'A1',
   },
   {
     value: 'cutoutB1',
@@ -240,13 +240,6 @@ export const getTrashSlot = (values: FormState): string => {
   const wasteChuteSlot = Boolean(hasWasteChute)
     ? [WASTE_CHUTE_CUTOUT as string]
     : []
-
-  if (
-    !cutouts.includes(FLEX_TRASH_DEFAULT_SLOT) &&
-    !moduleSlots.includes('A3')
-  ) {
-    return FLEX_TRASH_DEFAULT_SLOT
-  }
 
   const unoccupiedSlot = MOVABLE_TRASH_CUTOUTS.find(
     cutout =>
--- a/protocol-designer/src/step-forms/reducers/index.ts
+++ b/protocol-designer/src/step-forms/reducers/index.ts
@@ -67,8 +67,10 @@ import {
   createPresavedStepForm,
   getDeckItemIdInSlot,
   getIdsInRange,
+  getUnoccupiedSlotForMoveableTrash,
 } from '../utils'
-import {
+
+import type {
   CreateModuleAction,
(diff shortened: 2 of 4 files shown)


shared-data/python/opentrons_shared_data/protocol/models/protocol_schema_v6.py
```diff
@@ -21,44 +21,44 @@
...
-    positionId: Optional[str]
-    temperature: Optional[float]
-    celsius: Optional[float]
-    blockMaxVolumeUl: Optional[float]
-    rpm: Optional[float]
-    height: Optional[float]
-    offset: Optional[OffsetVector]
-    profile: Optional[List[ProfileStep]]
-    radius: Optional[float]
+    slotName: Optional[str] = None
```
sfoster1: we sure about these changes? `slotName: Optional[str]` without the `=None` means "you have to explicitly specify it but you can specify it as `None`/`null`" which I think is what we want

SyntaxColoring: In Pydantic v1, `slotName: Optional[str]` defaulted to `None` when the field was missing from the source JSON:

```python
>>> class M(pydantic.BaseModel):
...     slotName: typing.Optional[str]
...
>>> M.parse_raw("{}")
M(slotName=None)
```

In Pydantic v2, that becomes a validation error. If you want a default, you have to specify a default. This is a good, but breaking, change, so I guess the upgrade script applies this fixup liberally.

In *this particular case,* with the `Params` class, I think it's correct for these to default to `None`. In other cases of `f: Optional[T]`, though, we'll definitely need to be careful to cross-check against our underlying JSON schemas.

shared-data/python/opentrons_shared_data/pipette/pipette_definition.py
```diff
@@ -335,44 +344,53 @@ class PipettePhysicalPropertiesDefinition(BaseModel):
...
+    # TODO[pydantic]: The following keys were removed: `json_encoders`.
+    # Check https://docs.pydantic.dev/dev-v2/migration/#changes-to-config for more information.
+    model_config = ConfigDict(
+        json_encoders={
             pip_types.PipetteChannelType: lambda v: v.value,
             pip_types.PipetteModelType: lambda v: v.value,
             pip_types.PipetteGenerationType: lambda v: v.value,
             pip_types.Quirks: lambda v: v.value,
         }
+    )
```
SyntaxColoring: It looks like this is saying `json_encoders` isn't doing anything anymore; this is dead code.

According to the [migration guide](https://docs.pydantic.dev/2.0/migration/#required-optional-and-nullable-fields:~:text=Due%20to%20performance%20overhead%20and%20implementation%20complexity%2C%20we%20have%20now%20removed%20support%20for%20specifying%20json_encoders%20in%20the%20model%20config), there's some other way to do this now. But it isn't clear to me what this was accomplishing before. These lambdas look like straight passthroughs to me.

shared-data/python/opentrons_shared_data/protocol/models/shared_models.py
```diff
@@ -31,20 +30,20 @@ class WellDefinition(BaseModel):
     x: float
     y: float
     z: float
-    diameter: Optional[float]
-    yDimension: Optional[float]
-    xDimension: Optional[float]
+    diameter: Optional[float] = None
+    yDimension: Optional[float] = None
+    xDimension: Optional[float] = None
```
SyntaxColoring: This whole `WellDefinition` model is apparently unused. I think we can delete it.

SyntaxColoring (commented):
This mostly makes sense to me so far, thanks!

PR #14806, opened 2024-04-04
chore: create performance metrics project

DerekMaggio:
# Overview

Basic setup of Performance Metrics project

Closes https://opentrons.atlassian.net/browse/EXEC-380

# Test Plan

- [x] Verify all makefile commands can run successfully (GH Actions will be in another PR)
- [x] Build wheel, install, and verify it runs


sfoster1 (changes requested):
Let's keep the tooling we use in monorepo python projects consistent, with flake8, mypy, black, and so on at the same versions as the other projects. This eases everybody's ability to context switch.

performance-metrics/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[tool.ruff]
```
sfoster1: need to remove this one too - we're still using `setup.py`

sfoster1 (changes requested):

performance-metrics/pyproject.toml
DerekMaggio: https://github.com/Opentrons/opentrons/pull/14806/commits/71cb58415df045e67a42063338f3080048836b43 should fix it

sfoster1 (approved):
lgtm!

PR #14807, opened 2024-04-04
feat(hardware-testing): liquid sense testing script

ryanthecoder:
# Overview

This PR adds a new testing script that allows us to test all kinds of variations of the liquid-sense routine
it adds some additional features in the hardware control layer to change up output options to during the probe so we can gate using the buffer-on-pipette feature to a firmware version flag, since that feature has to be compiled in separately

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -2584,6 +2584,7 @@ class OT3API(
             probe=probe if probe else InstrumentProbeType.PRIMARY,
         )
         end_pos = await self.gantry_position(mount, refresh=True)
+        await self.move_to(mount, probe_start_pos)
         return end_pos.z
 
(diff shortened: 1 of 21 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2584,7 +2584,6 @@ async def liquid_probe(
             probe=probe if probe else InstrumentProbeType.PRIMARY,
         )
         end_pos = await self.gantry_position(mount, refresh=True)
-        await self.move_to(mount, probe_start_pos)
```
sfoster1: let's note this change a little more prominently in the pr description or better yet make it a separate pr please

sfoster1 (approved):
Looks good but let's make the hwcontrol change a little more prominently documented

PR #14808, opened 2024-04-04
feat(app): orchestration component for new quick transfer flow

smb2268:
fix PLAT-173, PLAT-228, PLAT-271

# Overview
This PR sets up the orchestration component, text file, and state for the new quick transfer flow and adds the first "create new transfer" screen

# Test Plan
1. With the Quick Transfer FF on, navigate to the protocols tab and click the "+ quick transfer" button
2. Add Using the back, continue, and exit buttons, go through the various screens in this wizard
3. Currently only titles are implemented after the first page, continue button and back button move through the wizard, exit takes you back to the protocols page
https://github.com/Opentrons/opentrons/assets/14302493/1288eb55-8b63-440d-9eb5-025f48729883
Screen Shot 2024-04-12 at 4 26 54 PM

# Changelog
1. Add new translation file for quick transfer, add title copy
2. Create orchestration component and setup the 8 steps
3. Add TS type and functions for useReducer storage that we'll use to manage the output of this wizard
4. Make a few tweaks to the ChildNavigation ODD header to allow for designed functionality
5. Adjusted DeckConfigurator to allow for static fixtures in the center column

# Review requests
Look through the orchestration component and reducer types

# Risk assessment
Low, behind FF


app/src/organisms/ChildNavigation/index.tsx
```diff
@@ -29,14 +29,16 @@ import type {
...
+  onClickBack?: React.MouseEventHandler
   buttonText?: React.ReactNode
   inlineNotification?: InlineNotificationProps
   onClickButton?: React.MouseEventHandler
   buttonType?: SmallButtonTypes
+  buttonIsDisabled?: boolean
   iconName?: IconName
   iconPlacement?: IconPlacement
   secondaryButtonProps?: React.ComponentProps<typeof SmallButton>
+  stepMeterPadding?: boolean
```
jerader: might be good to add a story with these optional props added since `ChildNavigation` is a commonly used component

app/src/organisms/ChildNavigation/index.tsx
```diff
@@ -29,14 +29,16 @@ import type {
...
 }
 
 export function ChildNavigation({
```
jerader: also can we add a test case with these new props?

app/src/organisms/QuickTransfer/NewTransferWizard.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const [wizardHeader, setWizardHeader] = React.useState<string>(
+    t('create_new_transfer')
+  )
+  const [continueIsDisabled] = React.useState<boolean>(false)
+  const wizardBody = <Flex>Wizard Body</Flex>
+  // every child component will take state as a prop, an anonymous
+  // dispatch function related to that step (except create new),
+  // and a function to disable the continue button
+
+  React.useEffect(() => {
```
jerader: is it necessary to have a `useEffect` to update the wizard header text?

app/src/organisms/QuickTransfer/NewTransferWizard.tsx
```diff
@@ -0,0 +1,155 @@
...
+    } else if (currentStep === 7) {
+      setWizardHeader(t('select_dest_wells'))
+    } else if (currentStep === 8) {
+      setWizardHeader(t('set_transfer_volume'))
+    }
+  }, [currentStep])
+
+  return (
+    <>
+      <StepMeter totalSteps={8} currentStep={currentStep} />
```
jerader: nit: maybe make 8 a const, like `NEW_TRANSFER_TOTAL_STEPS`

app/src/organisms/QuickTransfer/NewTransferWizard.tsx
```diff
@@ -0,0 +1,155 @@
...
+            if (currentStep === 8) {
+              history.push('protocols')
+            } else {
+              setCurrentStep(prevStep => prevStep + 1)
+            }
+          }}
+          buttonIsDisabled={continueIsDisabled}
+          secondaryButtonProps={{
+            buttonType: 'tertiaryLowLight',
+            buttonText: 'Exit',
```
jerader: can you add this to i18n?

app/src/organisms/QuickTransfer/types.ts
```diff
@@ -0,0 +1,51 @@
+import type { PipetteData } from '@opentrons/api-client'
+import type { LabwareDefinition1 } from '@opentrons/shared-data'
+import { ACTIONS } from './constants'
```
jerader: import can go higher

app/src/pages/ProtocolDashboard/index.tsx
```diff
@@ -280,7 +282,7 @@ export function ProtocolDashboard(): JSX.Element {
           buttonText={t('quick_transfer')}
           iconName="plus"
           onClick={() => {
-            console.log('launch quick transfer flow')
+            history.push('/quick-transfer')
```
jerader, line 288: since the `FloatingActionButton` will be moved to a different component eventually, can you add a comment here stating that? just so others are aware why there is no test coverage for this in `ProtocolDashboard`.

mjhuff (commented):
Should the quick transfer flow follow the general pattern we use for other flows, ie

`organisims/QuickTransferFlows` and then rename `NewTransferWizard.tsx` to `index.tsx`?

app/src/organisms/QuickTransfer/NewTransferWizard.tsx
smb2268: Yeah without this `useEffect` I get an infinite render error -- `Too many re-renders. React limits the number of renders to prevent an infinite loop.`

app/src/organisms/ChildNavigation/index.tsx
smb2268: I think it makes more sense to add `StyleProps` instead of making this specific prop, going to push that refactor! I'll add stories/test cases for the other two changes though

app/src/organisms/ChildNavigation/index.tsx
smb2268: same as above!

app/src/pages/ProtocolDashboard/index.tsx
smb2268: yep will do!

app/src/organisms/QuickTransferFlow/index.tsx
```diff
@@ -0,0 +1,160 @@
...
+    } else if (currentStep === 5) {
+      setWizardHeader(t('select_source_wells'))
+    } else if (currentStep === 6) {
+      setWizardHeader(t('select_dest_labware'))
+    } else if (currentStep === 7) {
+      setWizardHeader(t('select_dest_wells'))
+    } else if (currentStep === 8) {
+      setWizardHeader(t('set_transfer_volume'))
+    }
+  }, [currentStep])
```
b-cooper: This setState within a useEffect is very suspicious.  Why does the wizard header string need to be state at all?  Can't it just be directly dependent on the current step?  (e.g.
```
const ORDERED_STEP_HEADERS = [
      t('create_new_transfer'),
      t('select_attached_pipette'),
      t('select_tip_rack'),
      t('select_source_labware'),
      t('select_source_wells'),
      t('select_dest_labware'),
      t('select_dest_wells'),
      t('set_transfer_volume'),
]

const header = ORDERED_STEP_HEADERS[currentStep +1]
```
)

app/src/organisms/QuickTransferFlow/types.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+interface DestLabwareAction {
+  type: typeof ACTIONS.DEST_LABWARE
+  labware: LabwareDefinition1
+}
+interface DestWellsAction {
+  type: typeof ACTIONS.DEST_WELLS
+  wells: string[]
+}
+interface VolumeAction {
```
b-cooper: The action names prefixed with `Select` have a clear verb.  For consistency, it seems like these other actions could be prefixed with `Set` so it's clear what is being done to the subject noun (i.e. SourceLabware, SourceWells, etc)?

app/src/pages/ProtocolDashboard/index.tsx
```diff
@@ -58,6 +60,9 @@ export function ProtocolDashboard(): JSX.Element {
   const pinnedProtocolIds = useSelector(getPinnedProtocolIds) ?? []
   const pinnedProtocols: ProtocolResource[] = []
 
+  // The quick transfer button is going to be moved to a new quick transfer tab
```
b-cooper: Helpful comment! Maybe it should be marked as a TODO?

app/src/organisms/ChildNavigation/__tests__/ChildNavigation.test.tsx
```diff
@@ -72,4 +72,26 @@ describe('ChildNavigation', () => {
     fireEvent.click(secondaryButton)
     expect(mockOnClickSecondaryButton).toHaveBeenCalled()
   })
+  it.fails(
```
jerader, line 75: wow i didn't know `it.fails` was a thing

smb2268: I didn't either!! found it in the docs looking for a way to test this use case

b-cooper (approved):
✔️

PR #14810, opened 2024-04-04
Fix settings tests

DerekMaggio:
# Overview

When we patch the _read_settings_file function, we are calling to fixtures that exclude flags that have internal_only set to True. When you look at the _read_settings_file function there is no filtering out of flags set to internal_only. 

```python
def _read_settings_file(settings_file: "Path") -> SettingsData:
    """
    Read the settings file, which is a json object with settings IDs as keys
    and boolean values. For each key, look up the `Settings` object with that
    key. If the key is one of the old IDs (kebab case), replace it with the
    new ID and rewrite the settings file

    :param settings_file: the path to the settings file
    :return: a dict with all new settings IDs as the keys, and boolean values
        (the values stored in the settings file, or `False` if the key was not
        found). Along with the version.
    """
    # Read settings from persistent file
    data = _read_json_file(settings_file)
    settings, version = _migrate(data)
    settings = _ensure(settings)

    if data.get("_version") != version:
        _write_settings_file(settings, version, settings_file)

    return SettingsData(settings_map=settings, version=version)
```

Therefore, the patches for the OT-2 and Flex should be calling the fixtures that do not exclude the internal_only flag.

### DIFF
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -250,7 +250,7 @@ settings = [
         ),
         robot_type=[RobotTypeEnum.OT2, RobotTypeEnum.FLEX],
         internal_only=True,
-    )
+    ),
 ]
 
 if (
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
ty, good catch!

PR #14820, opened 2024-04-05
refactor(api): be more permissive with accepting int literals for float parameters

jbleon95:
# Overview

In the course of integrating runtime parameters frontend and backend work, it was discovered that for floating point parameters with a minimum and maximum range, if a literal int or value with only `.0` was entered, the parameter validation would fail as the type `int` did not match the accepted `float`.

While discussing this, it was determined that it would be best to allow those values, while also being less strict in the parameter definition creation `ParameterContext.add_float`. This permissiveness does **not** go the other way for parameter definitions, if a user adds something like `6.0` as a default for `ParameterContext.add_int`, that will still raise an error (value sent over to one of the endpoints however will do this conversion, due to how we defined our Pydantic types). Other values sent over will also not be converted but instead raise the appropriate error elsewhere. This includes values like `bool` which can be type converted in Python, a value of True will also **not** be converted to 1.

# Test Plan

Verified via postman that sending an int literal value for a float parameter override accepts the value and does not error out. Also ensure that this protocol passed analysis.

```
metadata = {
    'protocolName': 'RTP: Float test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}


    parameters.add_float(
        display_name="Volume 1",
        variable_name="volume_1",
        default=20,
        minimum=10,
        maximum=50,
    )
    parameters.add_float(
        display_name="Volume 2",
        variable_name="volume_2",
        default=20.0,
        choices=[
            {"display_name": "Low Volume", "value": 10},
            {"display_name": "Medium Volume", "value": 20},
            {"display_name": "High Volume", "value": 50.0},
        ],
    )


def run(context):
    pipette = context.load_instrument('flex_1channel_50', mount="left", tip_racks=[tip_rack])

    pipette.pick_up_tip()
    pipette.aspirate(context.params.volume_1, source.wells()[0])
    pipette.dispense(context.params.volume_1, destination.wells()[0])

    pipette.aspirate(context.params.volume_2, source.wells()[0])
    pipette.dispense(context.params.volume_2, destination.wells()[0])
    pipette.drop_tip()
```

# Changelog

- Allow int literals to be send as parameter override values for float parameters
- Allow int literals to be used for the `default`, `minimum`, `maximum` and value key of `choices` for `ParameterContext.add_float`

# Review requests
# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/_parameter_context.py
+++ b/api/src/opentrons/protocol_api/_parameter_context.py
@@ -91,10 +91,10 @@ class ParameterContext:
         parameter = parameter_definition.create_float_parameter(
             display_name=display_name,
             variable_name=variable_name,
-            default=default,
-            minimum=minimum,
-            maximum=maximum,
-            choices=choices,
+            default=validation.ensure_float_value(default),
+            minimum=validation.ensure_optional_float_value(minimum),
+            maximum=validation.ensure_optional_float_value(maximum),
+            choices=validation.ensure_float_choices(choices),
             description=description,
             unit=unit,
         )
--- a/api/src/opentrons/protocols/parameters/validation.py
+++ b/api/src/opentrons/protocols/parameters/validation.py
@@ -75,6 +75,36 @@ def ensure_value_type(
     return validated_value
 
 
+def ensure_float_value(value: Union[float, int]) -> float:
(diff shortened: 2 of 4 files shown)


sanni-t (approved):
Looks great!

PR #14821, opened 2024-04-05
fix(app): fallback to current run command on ODD if run diverges from analysis

b-cooper:
# Overview

In the case of a non-deterministic protocol, the running protocol screen of the on device display app will fall back to showing no command text at all, instead it should fall back to showing the current running command

Closes PLAT-274

# Review requests

Run a non-deterministic protocol, the ODD should display the current executing command, even after the divergence point.

e.g.

add the following to a python protocol.
```
if not protocol.is_simulating():
          protocol.comment(message="this should only be shown in the run, not in analysis")
```

Ensure that runtime only the comment and all following resulting commands are still shown in the Running Protocol screen of the on device display. 

# Risk assessment
low


brenthagen (approved):
LGTM !

PR #14828, opened 2024-04-07
feat(robot server): add a POST method on the analyses endpoint

sanni-t:
Closes AUTH-255

# Overview

Adds a POST method to the existing `/protocols/{protocolId}/analyses` endpoint in order to post a new analysis for an existing protocol.

This endpoint will take a request body with two optional fields:
- `runTimeParameterValues`
- `forceReAnalyze`

The new method can affect the analyses in three ways:
1. When the request is sent with `forceReAnalyze=True`, the server will unconditionally start a new analysis for the protocol using any RTP data sent along with it. It will return a 201 CREATED status and respond with a list of analysis summaries of all the analyses (ordered oldest first), including the newly started analysis.
2. When the request is sent without the `forceReAnalyze` field (or with `forceReAnalyze=False`), then the server will check the last analysis of the protocol
    - if the RTP values used for it were **different** from the RTP values sent with the current request, then the server will start a new analysis using the new RTP values. It will return a 201 CREATED status and respond with a list of analysis summaries of all the analyses, including the newly started analysis.
    - if the RTP values used for it were the **same** as the RTP values sent with the current request, then the server will **NOT** start a new analysis. It will return a 200 OK status, and simply return the existing list of analysis summaries.

This request requires the last analysis of the protocol to have been completed before handling this request. If the last analysis is pending, it will return a 503 error.

# Test Plan

Test out the above three cases and anything else you can think might affect the behavior. 
It's pretty well tested in unit & integration tests and it is also the same logic used for handling analyses from `POST /protocols`, so it is expected to work well when used as tested in integration tests.

# Review requests

Usual review for code sanity check.

# Risk assessment

Low. New HTTP API not yet used anywhere.

### DIFF
--- a/robot-server/robot_server/protocols/analysis_models.py
+++ b/robot-server/robot_server/protocols/analysis_models.py
@@ -41,6 +41,8 @@ class AnalysisResult(str, Enum):
 
 
 class AnalysisRequest(BaseModel):
+    """Model for analysis request body."""
+
     runTimeParameterValues: RunTimeParamValuesType = Field(
         default={},
         description="Key-value pairs of run-time parameters defined in a protocol.",
--- a/robot-server/robot_server/protocols/analysis_models.py
+++ b/robot-server/robot_server/protocols/analysis_models.py
@@ -2,7 +2,7 @@
 # TODO(mc, 2021-08-25): add modules to simulation result
 from enum import Enum
 
-from opentrons.protocol_engine.types import RunTimeParameter
+from opentrons.protocol_engine.types import RunTimeParameter, RunTimeParamValuesType
 from opentrons_shared_data.robot.dev_types import RobotType
 from pydantic import BaseModel, Field
 from typing import List, Optional, Union, NamedTuple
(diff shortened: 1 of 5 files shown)


robot-server/robot_server/protocols/router.py
```diff
@@ -519,6 +551,78 @@ async def delete_protocol_by_id(
...
+    values (if any), and appends it to the existing analyses.
+
+    If the last analysis in the existing analyses used the same RTP values, then a new
+    analysis is not created.
+
+    If `forceAnalyze` is True, this will always start a new analysis.
+
+    Returns: List of analysis summaries available for the protocol, ordered as
+             most recently started analysis last.
+    """
```
SyntaxColoring, line 592: I'm not sure if this is intentional, but FYI, none of this docstring will show up in the HTTP API docs—it gets overridden by the `description` in the decorator.

robot-server/robot_server/protocols/analysis_models.py
```diff
@@ -40,6 +40,18 @@ class AnalysisResult(str, Enum):
...
 
 
+class AnalysisRequest(BaseModel):
+    """Model for analysis request body."""
+
+    runTimeParameterValues: RunTimeParamValuesType = Field(
+        default={},
+        description="Key-value pairs of run-time parameters defined in a protocol.",
+    )
+    forceReAnalyze: bool = Field(
```
SyntaxColoring, line 50: FYI this doesn't quite match the name you have in the method docstring. `forceReAnalyze` vs. `forceAnalyze`.

SyntaxColoring (approved):
Code looks great, thanks. 🔬 

No notes other than that we'll need to auto-delete old analyses so they don't accumulate indefinitely, which I think you're already aware of.

PR #14831, opened 2024-04-07
refactor(app): update Divider stories

koji:
# Overview
update Divider stories

close AUTH-297

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


jerader (approved):
this is really similar to `Line`, right? i wonder if we need both?

koji:
> this is really similar to `Line`, right? i wonder if we need both?

yes. divider is used more than line. I guess we can just use divider.

PR #14832, opened 2024-04-08
feat(system-server,robot-server,api): add ability to enable OEM Mode when setting enableOEMMode advanced setting.

vegano1:
# Overview

We need to change the splash screen in user space, this is managed by the systems service `opentrons-loading.service`. This pull request in oe-core [(Opentrons/oe-core#143)](https://github.com/Opentrons/oe-core/pull/143) adds the mechanism to change the splash screen via the `/var/lib/opentrons-system-server/system.env` file. This pull request adds the endpoint and hooks it up to the robot-server to allow the OEM Mode to be enabled/disabled when the `enableOEMMode` feature flag is changed via the `/system/settings` POST request.

Closes: [PLAT-275](https://opentrons.atlassian.net/browse/PLAT-275) [PLAT-276](https://opentrons.atlassian.net/browse/PLAT-276)

# Test Plan

- [x] Build and test oe-core and buildroot builds on the Flex and OT-2
- [x]  Make sure that the robot-server and system-server startup regardless of the value of `enableOEMMode`
- [x] Make sure we can change the `oem_mode_enabled` flag in the `system.env` via PUT `/system/oem_mode/enable`.
- [x] Make sure that we change the `sysem.env` file when the `enableOEMMode` advanced flag is set via POST `/system/settings`.
- [x] Make sure that we return a `non-200` error code if PUT `/system/oem_mode/enable` fails for whatever reason.
- [x] Make sure that we DONT set the `enableOEMMode` advanced setting if PUT `/system/oem_mode/enable` fails.

# Changelog

- add aiohttp==3.8.1 to the robot-server so we can make async HTTP requests
- add `/system/oem_mode/enable` endpoint to enable/disable OEM Mode by changing the `system.env`
- add `oem_mode_enabled` and `oem_mode_splash_custom` system server config flags
- add `oem_mode_enabled` helper function to `feature_flags.py`
- add `save_settings` method to system-server settings so we can save the SystemServerSettings to a `system.env`
- use the `/system/oem_mode/enable` endpoint when the `enableOEMMode` advanced settings flag is changed.

# Review requests

- Does the approach make sense of having the `/system/oem_mode/enable` request get triggered from the robot-server.
- Ideally, the `set_oem_mode_request` is not just some floating function, this should be an `HTTPClient` that encapsulates HTTP requests to local endpoints. This might need to be added at a different time.
- I haven't done HTTP API's in a while, so feedback is appreciated.

# Risk assessment

- low, should only affect `enableOEMMOde` advanced setting for the Flex and OT-2

[PLAT-275]: https://opentrons.atlassian.net/browse/PLAT-275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLAT-276]: https://opentrons.atlassian.net/browse/PLAT-276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/opentrons/config/feature_flags.py
```diff
@@ -76,3 +76,8 @@ def enable_error_recovery_experiments() -> bool:
     return advs.get_setting_with_env_overload(
         "enableErrorRecoveryExperiments", RobotTypeEnum.FLEX
     )
+
+def oem_mode_enabled() -> bool:
```
CaseyBatten, line 85: Looks good but perhaps this should be passed robot type instead of assume it. We've had problems with feature flags before where we assume type in the past and had to correct it, like with `enable_door_safety_switch`. If this is passed an OT-2 robot type I'd expect we'd just raise an error, or do nothing at all?

vegano1: this should do nothing on the OT2, since all the logic that deals with OEM mode is on the client side.

robot-server/robot_server/service/legacy/routers/settings.py
```diff
@@ -78,10 +86,17 @@
...
 ) -> AdvancedSettingsResponse:
     """Update advanced setting (feature flag)"""
     try:
+        # send request to system server if this is the enableOEMMode setting
+        if update.id == "enableOEMMode" and robot_type == RobotTypeEnum.FLEX:
+            resp = await set_oem_mode_request(update.value)
+            if resp != 200:
+                # TODO: raise correct error here
+                raise Exception(f"Something went wrong setting OEM Mode. err: {resp}")
+
```
CaseyBatten, line 101: As discussed, the except handlers for the raised Exception below here are probably sufficient for the error response we'd want, so this should be okay.

api/src/opentrons/config/feature_flags.py
CaseyBatten: If this is unused in the current build (and is not planned to be used) should we simply removed it all together?

system-server/system_server/settings/settings.py
```diff
@@ -56,7 +49,44 @@ class SystemServerSettings(BaseSettings):
...
 
+        env_file = Environment().dot_env_path
         env_prefix = "OT_SYSTEM_SERVER_"
+
+
+def save_settings(settings: SystemServerSettings) -> bool:
+    """Save the settings to the dotenv file."""
+    env_path = Environment().dot_env_path
+    env_path = env_path or f"{settings.persistence_directory}/system.env"
+    prefix = settings.Config.env_prefix
```
CaseyBatten, line 84: As discussed the decision to have the system server responsible for this persistence file write is justified.

CaseyBatten (approved):
Overall everything looks good so far from my perspective. The justification for the odd coupling order (enabled OEM: client -> robot server -> system server, upload image: client -> system server) of the robot server, system server and client for file writing makes sense given actions occurring. However, we may want to note this decision for future divisions of labor between the system and robot servers to more cleanly separate the order of operations if any similar features are developed in the future.

robot-server/robot_server/service/legacy/routers/settings.py
vegano1: correct

PR #14833, opened 2024-04-08
fix(shared-data, components, app): fix runtime parameter min-max range for float

koji:
# Overview
fix runtime parameter min-max range for float.

the change fixes the following case.
range 1.5 - 10.0

current `1.5 -10`
this pr: `1.5-10.0`

I'm thinking about moving formatRange function from parameter table to somewhere `components`' util or the same place in `shared-data` since components should be pure ui components and we should avoid writing a specific function inside a component.

close AUTH-299

# Test Plan

# Changelog
- add an util function to shared-data

# Review requests

# Risk assessment
low

### DIFF
--- a/shared-data/js/helpers/formatRunTimeParameterMinMax.ts
+++ b/shared-data/js/helpers/formatRunTimeParameterMinMax.ts
@@ -1,0 +1,11 @@
+import type { RunTimeParameter } from '../types'
+
+export const formatRunTimeParameterMinMax = (
+  runTimeParameter: RunTimeParameter
+): string => {
+  const min = 'min' in runTimeParameter ? runTimeParameter.min : 0
+  const max = 'max' in runTimeParameter ? runTimeParameter.max : 0
+  return runTimeParameter.type === 'int'
+    ? `${min}-${max}`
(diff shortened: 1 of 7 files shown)


shared-data/js/helpers/formatRunTimeParameterMinMax.ts
```diff
@@ -0,0 +1,11 @@
+import type { RunTimeParameter } from '../types'
+
+export const formatRunTimeParameterMinMax = (
+  runTimeParameter: RunTimeParameter
+): string => {
+  const min = 'min' in runTimeParameter ? runTimeParameter.min : 0
+  const max = 'max' in runTimeParameter ? runTimeParameter.max : 0
+  return runTimeParameter.type === 'int'
+    ? `${min}-${max}`
+    : `${min.toFixed(1)}-${max.toFixed(1)}`
```
jerader, line 10: are floats only limited to 1 significant number?

koji: Actually, I asked about that to Nick when he was working on the same thing. He said at this moment one would be okay.

jerader (approved):
nice!

PR #14839, opened 2024-04-08
feat(protocol-designer): update unused module alert to account for MoaM

jerader:
closes AUTH-23

# Overview

the unused module alert is updated to add slot information.

# Test Plan

Turn on the multiple modules FF!

Create a flex protocol and add 2 temperature modules. Go to the deck map and add a pipetting step but do not add a temperature step.

export the protocol and you should see an unused modules modal that mentions the slot information. 

# Changelog

- update the unused modules alert text to make sense for multiple of the same type of modules
- fix the test

# Revew requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #14841, opened 2024-04-09
fix(protocol-designer): magnetic form correct engage height ranges

jerader:
closes RQA-2490, AUTH-269

# Overview

Adjust the ranges for min and max engage height for a magnetic module gen1 and gen2. This updates also removes the graphic and images in the magnet form.

# Test Plan

Create an ot-2 protocol and add a magnetic module gen1. Go to the deck map view and add a magnet step. Select to add an engage height and see the range from 0 - 45. Try to input ranges that are outside and see that it errors when you click off of the input field.

Now create a new protocol with a gen2. Do the same steps, this time add the Opentrons Tough plate on top of the module. Add a magnet step and see the range from -2.5 - 25 and the recommended text for the labware's recommended height.

# Changelog

- delete now unneeded graphic and images
- add the engage height text to the input field
- add test coverage
- update the constants

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/StepEditForm/forms/MagnetForm.tsx
```diff
@@ -1,40 +1,53 @@
...
+  MIN_ENGAGE_HEIGHT_V2,
+} from '../../../constants'
 import { TextField, RadioGroupField } from '../fields'
-import styles from '../StepEditForm.module.css'
+import type { StepFormProps } from '../types'
 
-import { StepFormProps } from '../types'
+import styles from '../StepEditForm.module.css'
 
 export const MagnetForm = (props: StepFormProps): JSX.Element => {
```
koji:
```suggestion
export function MagnetForm(props: StepFormProps): JSX.Element {
```

protocol-designer/src/components/StepEditForm/forms/MagnetForm.tsx
```diff
@@ -1,40 +1,53 @@
...
   const moduleLabwareOptions = useSelector(
     uiModuleSelectors.getMagneticLabwareOptions
   )
+  const moduleEntities = useSelector(getModuleEntities)
   const { t } = useTranslation(['application', 'form'])
-
-  const moduleEntities = useSelector(stepFormSelectors.getModuleEntities)
   const { magnetAction, moduleId } = props.formData
-  const moduleModel = moduleId ? moduleEntities[moduleId]?.model : null
+  const { propsForFields } = props
```
koji:
```ts
const { formData, propsForFields } = props
const { magnetAction, moduleId } = formData
```

koji:
Question
For V1, engage height is allowed to use `float`? 
I could set `44.5`.

protocol-designer/src/constants.ts
```diff
@@ -65,10 +65,10 @@ export const DEFAULT_MM_BLOWOUT_OFFSET_FROM_TOP = 0
...
 export const DEFAULT_WELL_ORDER_FIRST_OPTION: 't2b' = 't2b'
 export const DEFAULT_WELL_ORDER_SECOND_OPTION: 'l2r' = 'l2r'
-export const MIN_ENGAGE_HEIGHT_V1 = -5
-export const MAX_ENGAGE_HEIGHT_V1 = 40
-export const MIN_ENGAGE_HEIGHT_V2 = -4
-export const MAX_ENGAGE_HEIGHT_V2 = 19
+export const MIN_ENGAGE_HEIGHT_V1 = 0
+export const MAX_ENGAGE_HEIGHT_V1 = 45
+export const MIN_ENGAGE_HEIGHT_V2 = -2.5
+export const MAX_ENGAGE_HEIGHT_V2 = 25
```
koji, line 71: maybe `25.0` for display consistency.

jerader:
> Question For V1, engage height is allowed to use `float`? I could set `44.5`.

@koji , ya that is the current behavior in production so i think we should stick with it.

protocol-designer/src/constants.ts
jerader: oh good call!

koji (approved):
left a few comments
the sandbox worked as expected and the changes make sense to me.

PR #14842, opened 2024-04-09
fix(shared-data): format rtp float and int choices to include suffix

jerader:
closes AUTH-311

# Overview

display suffixes for int and float choices.

# Test Plan

Upload the attached protocol in the ticket and see that the default and updates values in the table display the suffixes for choices that are type `int` and `float`

# Changelog

- modify the 2 utils and fix the test. choices type can be an int and float so i had to remove it from the switch block

# Review requests

see test plan

# Risk assessment

low


koji (approved):
thank you for the quick fix!
lgtm!

PR #14844, opened 2024-04-09
fix(app): display app version again

sfoster1:
When we switched to vite, we had to switch all the stuff we'd been injecting at pack time via webpack environment/define plugins to vite's `define` config functionality. The biggest thing we specify that way is the app version, which is used across the stack for display and for logic.

In the commit that switched to vite, we added that injection for the app-shell vite configs but did not add it for the app vite configs. That meant that at runtime, the version value was undefined, which breaks robot update notifications and causes the app version in the general settings tab to not display (it also makes the logo wrong on internal releases but that's a bit less important).

The fix is to inject the version into the app build again. This is made a little more complicated because if you're doing stuff to the app vite config, it has to work in both the vite devserver and the vite offline packaging environments, and the vite devserver doesn't allow commonjs, and the git-version script that gives us the version is commonjs. For the purposes of vite's devserver, "doesn't work with cjs" actually just means "doesn't support require()", so you can use a hybrid syntax that uses import-statements but still module.export instead of export statements.

Unfortunately, the git-version script is also used in the electron-builder config for the app-shell and the app-shell-odd, and the electron-builder config is run via node, and to import an ESM from a node CJS script - which electron-builder.config.js is - you need to change your import syntax to dynamic import and you need to make the import target explicitly (to node) an ESM, aka change its extension, and you need to use full ESM syntax including exports.

So that means that
- git-version.js becomes git-version.mjs and uses full ESM syntax
- that means that everywhere it's imported we need to import it by full path with extension instead of module name
- also we need to import it dynamically in the electron config
- oh and we need to actually add the define configs so we get the version in the app

And then finally we show the version again.

Closes EXEC-385


app/vite.config.ts
```diff
@@ -1,62 +1,70 @@
 import path from 'path'
-import { defineConfig } from 'vite'
+import { UserConfig, defineConfig } from 'vite'
```
koji:
```suggestion
import {defineConfig } from 'vite'
import type { UserConfig } from 'vite'
```

koji (approved):
The changes look good to me and the versions are shown up in each application.
Thank you for fixing this!

PR #14846, opened 2024-04-09
fix(api): Make labware loads and moves slightly more atomic

SyntaxColoring:
# Overview

This tries to address one of the errors in RESC-227.

Suppose you tried to catch a labware load exception, like this:

```python
hs = protocol.load_module("heaterShakerModuleV1",1)
try:
    # Error: conflicts with the adjacent Heater-Shaker.
    lw = protocol.load_labware("opentrons_24_tuberack_eppendorf_1.5ml_safelock_snapcap", 2)
except Exception:
    # ...
```

When you catch the exception, the system is currently left in an inconsistent state. According to the Protocol Engine layer, the labware load succeeded. But according to the PAPI layer, the labware load failed. This is a [longstanding known problem](https://github.com/Opentrons/opentrons/blob/1819b8c90b8440d6d896f4315380651f1880989c/api/src/opentrons/protocol_api/core/engine/protocol.py#L223-L232) that has caused user-facing confusion before. In RESC-227, a customer tried to clear the slot by doing `del protocol.deck[slot]`, but the inconsistent state caused that to raise an internal `KeyError`.

# Changelog

In `ProtocolContext.load_labware()`, `ProtocolContext.load_adapter()`, and `ProtocolContext.move_labware()`: if we detect a conflict, try to keep the two layers in sync by "undoing" the load in Protocol Engine, by running a Protocol Engine `moveLabware` `offDeck` command.

This is not a real solution because:

* It does not work if the conflict happens in `ProtocolContext.load_module()` (because we can't unload modules).
* The failed labware will still show up in the run and protocol analysis, which means it will still show up in the Opentrons App's setup steps.
* The `offDeck` labware move will show up in the run log.

# Review requests

Is this worth the complexity? Instead of doing this, should we push towards solving this the right way, as described in the `# todo` comment?

https://github.com/Opentrons/opentrons/blob/8cfa337e9ca77d8b52edd3c76334813d9996e2cd/api/src/opentrons/protocol_api/core/engine/protocol.py#L244-L257

# Test Plan

To do.

# Risk assessment

To do.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/protocol.py
+++ b/api/src/opentrons/protocol_api/core/engine/protocol.py
@@ -217,41 +217,59 @@ class ProtocolCore(
             version=version,
             display_name=label,
         )
-        # FIXME(jbl, 2023-08-14) validating after loading the object issue
-        validation.ensure_definition_is_labware(load_result.definition)
 
-        # FIXME(mm, 2023-02-21):
-        #
-        # We're wrongly checking for deck conflicts *after* we've already loaded the
-        # labware into the ProtocolEngine. If it turns out there is a conflict,
-        # and this check raises, it will leave this object and its ProtocolEngine
-        # in a confusing inconsistent state.
-        #
-        # I expect we can get away with this in practice a lot of the time because
-        # exceptions in Python protocols are mostly treated as fatal, anyway.
-        # Users rarely catch them.
-        deck_conflict.check(
-            engine_state=self._engine_client.state,
-            new_labware_id=load_result.labwareId,
-            existing_disposal_locations=self._disposal_locations,
-            # TODO (spp, 2023-11-27): We've been using IDs from _labware_cores_by_id
-            #  and _module_cores_by_id instead of getting the lists directly from engine
-            #  because of the chance of engine carrying labware IDs from LPC too.
-            #  But with https://github.com/Opentrons/opentrons/pull/13943,
-            #  & LPC in maintenance runs, we can now rely on engine state for these IDs too.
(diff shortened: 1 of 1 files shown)


sfoster1 (changes requested):
I don't think this version of it is a good idea. I think you still end up in a generally incorrect state - the labware is loaded but OFF_DECK - and I think you end up with an even more incorrect run log that could confuse clients that rely only on the engine output for protocol state, as we encourage clients to do.

I think the Right Solution to the problem of "we want to enforce certain rules in the python api that we do not want to enforce in the engine" is enforce those rules _before_ involving the engine, and only involve the engine if the enforcement passes. We could do that in two ways:
1. We could reimplement the checks to take `(engine state, proposed change)` instead of just `(engine state with the change already made)`. I think this is more fundamentally sound, because the question of "is this state that exists valid" is a really awkward one because it allows the concept that invalid states can exist. Instead, we should be arranging our code so that a) as many invalid states as possible have invalidity encoded into language semantics (i.e. , "it's invalid for a value to be True and False" is something we encode into language semantics by saying "this value is a bool") and b) the rest of them are checked _before entering the state_.
2. Make a new engine with a copy of the state and an analysis backend, supply the action _to that engine_, and run the check on that engine's state. This sounds like a joke but like, the whole point of the engine is that it's transient, that it can run with and without hardware, that at the end of the day it's just data, and data can be copied.

I'd prefer 1, though.

sanni-t:
Agree with @sfoster1. Also prefer option 1. 
It also does look possible to do the deck conflict check without loading the labware into the engine first.

PR #14848, opened 2024-04-09
feat(protocol-designer): create container for all tipracks

jerader:
closes AUTH-313

# Overview

Now, all tipracks are auto generated on the deck after initial protocol creation. And the tiprack option dropdown is modified to only include tipracks for that specific pipette that are on the deck.

Screenshot 2024-04-09 at 14 11 57

Screenshot 2024-04-09 at 14 11 26

# Test Plan

Create a flex or ot-2 protocol. Perhaps flex might be better to add the max of 6 tipracks to the deck. For the pipette select both p1000s and selecting the first 3 tipracks for one and the other 3 tipracks for the 2nd pipette (this is so you select all 6 different tipracks)

Go to the deck map and see that all 6 tipracks are placed, starting with filling up the center column 2. (this is a Product decision since modules are by default placed in the outside slots with magnetic Block going in slot D2). Then the rest of the tipracks should fill up the next available slots.

Add a pipetting step and see that the dropdown menu for the tipracks shows only the pipette's default tipracks that you selected. Now delete the step and delete a tiprack. Create the step again and one pipette should only have 2 dropdown items now. If you hover over the dropdown field, a tooltip should show up saying that you are missing a tiprack.

# Changelog

- fix logic for auto-generating the tipracks 
- fix which tipracks are displayed in the tiprack options selector
- add a test to `tiprackField`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
+++ b/protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
@@ -35,7 +35,7 @@ export function TiprackField(props: TiprackFieldProps): JSX.Element {
   const pipetteOptions = options.filter(option =>
     defaultTipracks.includes(option.defURI)
   )
-  const missingTiprack = defaultTipracks.length > options.length
+  const missingTiprack = defaultTipracks.length > pipetteOptions.length
 
   return (
     <Box {...targetProps}>
--- a/protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
+++ b/protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
@@ -10,9 +10,9 @@ import {
 } from '@opentrons/components'
 import { selectors as uiLabwareSelectors } from '../../../ui/labware'
 import { getPipetteEntities } from '../../../step-forms/selectors'
-import styles from '../StepEditForm.module.css'
-
 import type { FieldProps } from '../types'
+
+import styles from '../StepEditForm.module.css'
 
 interface TiprackFieldProps extends FieldProps {
   pipetteId?: unknown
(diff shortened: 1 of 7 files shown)


protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
```diff
@@ -1,32 +1,62 @@
...
+  const { t } = useTranslation(['form', 'tooltip'])
+  const [targetProps, tooltipProps] = useHoverTooltip()
+  const pipetteEntities = useSelector(getPipetteEntities)
   const options = useSelector(uiLabwareSelectors.getTiprackOptions)
+  const defaultTipracks =
+    pipetteId != null ? pipetteEntities[pipetteId as string].tiprackDefURI : []
+  const pipetteOptions = options.filter(option =>
+    defaultTipracks.includes(option.defURI)
+  )
+  const missingTiprack = defaultTipracks.length > pipetteOptions.length
```
koji: nit 
maybe adding `is` or `has`?

```ts
  const missingTiprack = defaultTipracks.length > pipetteOptions.length
```

protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
```diff
@@ -1,32 +1,62 @@
...
+  DropdownField,
+  useHoverTooltip,
+  Tooltip,
+  Box,
+} from '@opentrons/components'
 import { selectors as uiLabwareSelectors } from '../../../ui/labware'
+import { getPipetteEntities } from '../../../step-forms/selectors'
 import styles from '../StepEditForm.module.css'
 
 import type { FieldProps } from '../types'
```
koji:
```suggestion
import type { FieldProps } from '../types'

import styles from '../StepEditForm.module.css'
```

koji (approved):
the sandbox worked as expected without the case I sent.

jerader:
> the sandbox worked as expected without the case I sent.

@koji thanks! I made this ticket to address the bug you found https://opentrons.atlassian.net/browse/AUTH-319

PR #14850, opened 2024-04-09
refactor(protocol-designer, components): infoItem to nicely accommoda…

jerader:
…te multiple tipracks

closes AUTH-314

# Overview

Previously, the pipette section in PD was jumbled if you had multiple tipracks selected for the pipette. This PR makes it look prettier 💅 

Screenshot 2024-04-09 at 15 10 16

# Test Plan

Create a flex or ot-2 protocol and add multiple tips onto each pipette. Create protocol and examine the pipette section.

# Changelog

- list tiprack models via bulletpoints and remove now unnecessary component (InfoItem)
- create test for `InstrumentInfo`

# Review requests

see test plan

# Risk assessment

low


components/src/instrument/InstrumentInfo.tsx
```diff
@@ -1,77 +1,101 @@
...
-  /** classes to apply */
-  className?: string
-  /** classes to apply to the info group child */
-  infoClassName?: string
+  /** paired tiprack models */
+  tiprackModels?: string[]
   /** children to display under the info */
   children?: React.ReactNode
+  /** if true, show labels 'LEFT PIPETTE' / 'RIGHT PIPETTE' */
+  showMountLabel?: boolean | null
```
koji, line 27: Question

Does this prop need `null`?

Also, can we set false as the default value?

components/src/instrument/InstrumentInfo.tsx
```diff
@@ -1,77 +1,101 @@
...
-              <InfoItem
-                key={`tiprackModel-${index}`}
-                title={index === 0 ? 'tip racks' : null}
-                value={model}
-              />
-            ))
-          : null}
+        <Flex flexDirection={DIRECTION_COLUMN} gridGap={SPACING.spacing4}>
+          <StyledText as="h3" fontWeight={TYPOGRAPHY.fontWeightSemiBold}>
+            Tip rack
```
koji:
```suggestion
            {'Tip rack'}
```

components/src/instrument/InstrumentInfo.tsx
```diff
@@ -1,77 +1,101 @@
...
+          </StyledText>
+          <ul>
+            {tiprackModels != null ? (
```
koji: from the props type, tiprackModels: string[] so checking its length would be better.

```suggestion
            {tiprackModels != null && tiprackModels.length >0 ? (
```

components/src/instrument/InstrumentInfo.tsx
```diff
@@ -1,77 +1,101 @@
...
+      >
+        <Flex flexDirection={DIRECTION_COLUMN} gridGap={SPACING.spacing4}>
+          <StyledText
+            as="h3"
+            fontWeight={TYPOGRAPHY.fontWeightSemiBold}
+            textTransform={TYPOGRAPHY.textTransformCapitalize}
+          >
+            {showMountLabel && !has96Channel ? `${mount} pipette` : 'pipette'}
+          </StyledText>
+          <StyledText as="p" width="max-content" maxWidth="14rem">
```
koji: maybe define a const for `14rem` since it is used in multiple `StyledText`?

components/src/instrument/InstrumentInfo.tsx
```diff
@@ -1,77 +1,101 @@
...
+              tiprackModels.map((model, index) => (
+                <li key={`${model}_${index}`}>
+                  <StyledText as="p" width="max-content" maxWidth="14rem">
+                    {model}
+                  </StyledText>
+                </li>
+              ))
+            ) : (
+              <StyledText as="p" width="max-content" maxWidth="14rem">
+                None
```
koji:
```suggestion
                {'None'}
```

components/src/instrument/InstrumentInfo.tsx
koji: nit
maybe adding `text-wrap` would be good but completely optional since text's readability is improved well.

https://github.com/Opentrons/opentrons/assets/474225/022832e7-7103-4776-9e38-86b22c29d480

components/src/instrument/InstrumentInfo.tsx
jerader: ah looks like it needs undefined but not null! will remove null

koji: still null is staying

jerader: it turns out showMountLabel in the parent can be null and undefined, figured i'd just keep it for now

koji (approved):
left one comment but it's not a blocker to merge this into edge.
Could you create a ticket to track Storybook implementation for this component?
I guess the design team will need Storybook for DQA in the future.

components/src/instrument/InstrumentInfo.tsx
jerader: these components will be remade when PD is redesigned soon, so we can clean it up then!

PR #14857, opened 2024-04-10
fix(protocol-designer): auto-generate trashBin for flex if no pipetting commands exist

jerader:
closes AUTH-267

# Overview

In PD in production, if you create a Flex protocol with no pipetting steps and reupload it to the app, it does not auto generate a trash bin for you. Technically there is a way around it by adding a trash bin entity. But its a weird state because return tip and no trash entity is not supported in PD. 

Ot-2 does not have this issue because it always ensures a hardcoded fixed trash is generated

So this PR auto-generates a trash bin for a flex if there is none when importing a protocol.

# Test Plan

I created a test case to test all scenarios, so it should work as expected. 

However, you can create a flex protocol and add a pause step. Download the protocol and reimport it back to PD. See that the trash bin is generated.

# Changelog

- create util for getting an unoccupied slot, taking into account modules, labwares, move labware commands, waste chute, and staging area slots
- add test

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/utils.ts
+++ b/protocol-designer/src/components/modals/CreateFileWizard/utils.ts
@@ -67,12 +67,12 @@ export const getCrashableModuleSelected = (
 
 export const MOVABLE_TRASH_CUTOUTS = [
   {
-    value: 'cutoutA1',
-    slot: 'A1',
-  },
-  {
     value: 'cutoutA3',
     slot: 'A3',
+  },
+  {
+    value: 'cutoutA1',
+    slot: 'A1',
   },
   {
     value: 'cutoutB1',
@@ -240,13 +240,6 @@ export const getTrashSlot = (values: FormState): string => {
   const wasteChuteSlot = Boolean(hasWasteChute)
     ? [WASTE_CHUTE_CUTOUT as string]
     : []
-
-  if (
-    !cutouts.includes(FLEX_TRASH_DEFAULT_SLOT) &&
-    !moduleSlots.includes('A3')
-  ) {
-    return FLEX_TRASH_DEFAULT_SLOT
-  }
 
   const unoccupiedSlot = MOVABLE_TRASH_CUTOUTS.find(
     cutout =>
--- a/protocol-designer/src/step-forms/utils/index.ts
+++ b/protocol-designer/src/step-forms/utils/index.ts
@@ -13,6 +13,7 @@ import {
 import { SPAN7_8_10_11_SLOT, TC_SPAN_SLOTS } from '../../constants'
 import { hydrateField } from '../../steplist/fieldLevel'
 import { LabwareDefByDefURI } from '../../labware-defs'
+import { getCutoutIdByAddressableArea } from '../../utils'
 import type {
   AddressableAreaName,
   CutoutId,
@@ -41,7 +42,6 @@ import type {
   FormPipette,
   LabwareOnDeck as LabwareOnDeckType,
 } from '../types'
-import { getCutoutIdByAddressableArea } from '../../utils'
 export { createPresavedStepForm } from './createPresavedStepForm'
 
 const MOVABLE_TRASH_CUTOUTS = [
@@ -297,14 +297,12 @@ export function getHydratedForm(
   return hydratedForm
 }
 
-export function getUnoccupiedSlotForMoveableTrash(
+export const getUnoccupiedSlotForMoveableTrash = (
   file: PDProtocolFile,
   hasWasteChuteCommands: boolean,
   stagingAreaSlotNames: AddressableAreaName[]
-): string {
-  const wasteChuteSlot = hasWasteChuteCommands
-    ? [WASTE_CHUTE_CUTOUT as string]
-    : []
+): string => {
+  const wasteChuteSlot = hasWasteChuteCommands ? [WASTE_CHUTE_CUTOUT] : []
   const stagingAreaCutoutIds = stagingAreaSlotNames.map(slotName =>
     getCutoutIdByAddressableArea(
       slotName,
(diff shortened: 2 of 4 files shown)


protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -241,13 +241,6 @@ export const getTrashSlot = (values: FormState): string => {
     ? [WASTE_CHUTE_CUTOUT as string]
     : []
 
-  if (
-    !cutouts.includes(FLEX_TRASH_DEFAULT_SLOT) &&
-    !moduleSlots.includes('A3')
-  ) {
-    return FLEX_TRASH_DEFAULT_SLOT
-  }
-
```
jerader, line 250: now unnecessary logic since we are listing 'A3' as the first cutout in the array to map through

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -28,9 +30,55 @@ import type {
...
+import type { PDProtocolFile } from '../../file-types'
+import type {
+  AdditionalEquipmentOnDeck,
+  InitialDeckSetup,
+  ModuleOnDeck,
+  FormPipettesByMount,
+  FormPipette,
+  LabwareOnDeck as LabwareOnDeckType,
+} from '../types'
+import { getCutoutIdByAddressableArea } from '../../utils'
```
koji: this can be up

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -248,3 +296,84 @@ export function getHydratedForm(
   // @ts-expect-error(sa, 2021-6-14):type this properly in #3161
   return hydratedForm
 }
+
+export function getUnoccupiedSlotForMoveableTrash(
```
koji: if there is any specific reason, using an arrow function would be good for consistency.

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -248,3 +296,84 @@ export function getHydratedForm(
...
+  file: PDProtocolFile,
+  hasWasteChuteCommands: boolean,
+  stagingAreaSlotNames: AddressableAreaName[]
+): string {
+  const wasteChuteSlot = hasWasteChuteCommands
+    ? [WASTE_CHUTE_CUTOUT as string]
```
koji, line 306:
```suggestion
    ? [WASTE_CHUTE_CUTOUT]
```

protocol-designer/src/step-forms/utils/index.ts
```diff
@@ -248,3 +296,84 @@ export function getHydratedForm(
...
+      !stagingAreaCutoutIds.includes(cutout.value as CutoutId)
+  )
+  if (unoccupiedSlot == null) {
+    console.error(
+      'Expected to find an unoccupied slot for auto-generating a trash bin but could not'
+    )
+    return ''
+  }
+
+  return unoccupiedSlot?.slot
```
koji: this line is after null-check so `?` is not required.

```suggestion
  return unoccupiedSlot.slot
```

koji:
question

when I added waste chute, export and import the protocol into pd, trash bin is generated but waste chute is gone.
Is this expected behavior?

jerader:
> question
> 
> when I added waste chute, export and import the protocol, trash bin is generated but waste chute is gone. Is this expected behavior?

@koji oops that is definitely a bug! Thanks for catching it.

jerader:
> > question
> > when I added waste chute, export and import the protocol, trash bin is generated but waste chute is gone. Is this expected behavior?
> 
> @koji oops that is definitely a bug! Thanks for catching it.

@koji actually, if you did not add any waste chute commands, it is not a bug since we remove any staging area slot and waste chute that is not used in the protocol. But if you did add waste chute commands, it is a bug.

koji:
> > > question
> > > when I added waste chute, export and import the protocol, trash bin is generated but waste chute is gone. Is this expected behavior?
> > 
> > 
> > @koji oops that is definitely a bug! Thanks for catching it.
> 
> @koji actually, if you did not add any waste chute commands, it is not a bug since we remove any staging area slot and waste chute that is not used in the protocol. But if you did add waste chute commands, it is a bug.

@jerader I added waste chute from the initial setup wizard.

koji:
one more question

I compared two json.
left: exported json - no trash bin
right: import the left json and re-export (trash bin with auto generated)

making pipetteTiprackAssignments item nested (adding `[]`) is an expected behavior?

Screenshot 2024-04-10 at 1 57 51 PM

left
[no-trashbin.json](https://github.com/Opentrons/opentrons/files/14935992/no-trashbin.json)

right
[no-trashbin (1).json](https://github.com/Opentrons/opentrons/files/14935993/no-trashbin.1.json)

jerader:
> one more question
> 
> I compared two json. left: exported json - no trash bin right: import the left json and re-export (trash bin with auto generated)
> 
> making pipetteTiprackAssignments item nested (adding `[]`) is an expected behavior?
> 
> left [no-trashbin.json](https://github.com/Opentrons/opentrons/files/14935992/no-trashbin.json)
> 
> right [no-trashbin (1).json](https://github.com/Opentrons/opentrons/files/14935993/no-trashbin.1.json)

@koji  oh i think what's happening is when you reimport, it goes through the 8.1 migration because the version when you export is `8.0.3` and not `8.1.0` - this is a current limitation with PD where the versioning doesn't update until i push a release tag 😦 so for now we have to manually change the version before we reimport.

This is a great catch though! i forgot that we have to do this.

jerader:
> > > > question
> > > > when I added waste chute, export and import the protocol, trash bin is generated but waste chute is gone. Is this expected behavior?
> > > 
> > > 
> > > @koji oops that is definitely a bug! Thanks for catching it.
> > 
> > 
> > @koji actually, if you did not add any waste chute commands, it is not a bug since we remove any staging area slot and waste chute that is not used in the protocol. But if you did add waste chute commands, it is a bug.
> 
> @jerader I added waste chute from the initial setup wizard.

@koji got it, this sounds like expected behavior. if you didn't add any step using the waste chute then it isn't being used in the protocol and won't generate upon reimport or even if you upload it to the app.

koji (approved):
thank you for answering to all my questions.
LGTM!

PR #14859, opened 2024-04-10
feat(shared-data, api): add uiMaxFlowRate key to pipette definitions

jerader:
closes AUTH-328

# Overview

This PR add max flow rate values to each GEN2 and GEN3 pipette model definition to be used for FE applications, based off of the calculations [here](https://github.com/Opentrons/opentrons/blob/chore-calculate-max-flow-rate-per-pipette-model/hardware-testing/hardware_testing/scripts/calculate_plunger_flow_rates.py )

Rather than having a max flow rate for every pipette's supported tip's volume for aspirate and dispense, we have only 1 max flow rate for every pipette's supported tip that is. That value is based off of the lowest volume minus 2% to account for safety. 

For example:
`p1000_single_v3.6` has a `minVolume` of 5uL. The `t50` tips has a max flow rate at volume 5uL of `801.3`. So the `uiMaxFlowRate` value is 98% of that which is `785.2` 

The new key in the definition is called `uiMaxFlowRate` because it is limited to the lowest volume's max flow rate, which will be used for FE applications only (Quick transfer and PD)

🔈  Additionally, the `lowVolumeDefault` default flow rates for `aspirate`, `dispense`, and `blowout` have been changed to match the `uiMaxFlowRate`. This DOES NOT result in a physical change behavior and the change is needed for ui purposes.

# Test Plan

Review the additions to the definitions
Review the pytest

# Changelog

- add `uiMaxFlowRate` key to each pipette liquid definitions's supported tips
- change the `lowVolumeDefault` flowrates to match the `uiMaxFlowRate`
- add to JS type and add to schema
- make a pytest that asserts that the ui max flow rate value is less than all the max flow rate values for aspirate and dispense for that pipette's supported tip
- create new `ul_per_mm` python file and copy over the `piecewise_volume_conversion` fn from the api hardware directory, removing the old version
# Review requests

see test plan

# Risk assessment

low-ish, nothing is using these new values yet


jerader:
Spoke with @sanni-t , I am going to add the max values to the pipette definitions in a `maxFlowRate` key and will modify my util to just read the information from the definitions.

shared-data/pipette/definitions/2/liquid/eight_channel/p1000/default/3_3.json
```diff
@@ -2,6 +2,54 @@
   "$otSharedSchema": "#/pipette/schemas/2/pipetteLiquidPropertiesSchema.json",
   "supportedTips": {
     "t50": {
+      "maxFlowRate": {
+        "5": 824.8,
+        "6": 849.8,
+        "7": 872.5,
+        "8": 885.8,
+        "9": 898,
```
koji: question

`898` instead of `898.0`, is this intend?

jerader: these max flow rates are all a result from the equation [here](https://github.com/Opentrons/opentrons/blob/chore-calculate-max-flow-rate-per-pipette-model/hardware-testing/hardware_testing/scripts/calculate_plunger_flow_rates.py)! We are rounding to 
1 decimal value if any decimal values exist. I think if there are no decimals values, it won't show it. Since this data is backend, i think for the frontend, we can expose another decimal value to make everything match.

shared-data/js/__tests__/pipettes.test.ts
```diff
@@ -248,4 +328,69 @@ describe('pipette data accessors', () => {
       mockLiquids
     )
   })
+
+  describe('getMaxFlowRateByVolume', () => {
+    const mockMaxFlowRate: Record<string, number> = {
```
koji: creating a fixture would be good?

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_6.json
```diff
@@ -7,20 +7,8 @@
...
-        "5": 10.0,
-        "6": 10.0,
-        "7": 10.0,
-        "8": 10.0,
-        "12": 10.0,
-        "16": 10.0,
-        "24": 10.0,
-        "48": 10.0
-      },
+      "speed": 10.0,
```
sfoster1: this is going to cause issues when merged and is because this is based on an old edge - may want to merge recent edge into this

shared-data/pipette/definitions/2/liquid/eight_channel/p50/lowVolumeDefault/3_4.json
```diff
@@ -2,17 +2,18 @@
   "$otSharedSchema": "#/pipette/schemas/2/pipetteLiquidPropertiesSchema.json",
   "supportedTips": {
     "t50": {
+      "uiMaxFlowRate": 32.6,
       "defaultAspirateFlowRate": {
-        "default": 35,
-        "valuesByApiLevel": { "2.14": 35 }
+        "default": 32.6,
```
sfoster1, line 7: Let's call this out in the PR description and thus the eventual commit message a little  more please

shared-data/python/opentrons_shared_data/pipette/ul_per_mm.py
```diff
@@ -0,0 +1,33 @@
+from typing import List, Tuple
```
sfoster1: We prooobably should change the code that imports this to import it from here also

sfoster1 (changes requested):
Looks great overall, but a couple things
- Definitely need to call out the actual `lowVolumeDefault` changes in the description a bit more; as we discussed, they're fine because they won't result in actual changed physical behavior, but they should be highlighted
- This is less significant of an opinion - it could also be a followup - but we need to switch the python over to importing `ul_per_mm` from the new location at some point

shared-data/pipette/definitions/2/general/ninety_six_channel/p1000/3_6.json
jerader: oops i did not catch this when i merged edge in, thank you!

shared-data/pipette/schemas/2/pipetteLiquidPropertiesSchema.json
```diff
@@ -92,6 +92,11 @@
               "dispense": {
                 "type": "array",
                 "items": { "$ref": "#/definitions/liquidHandlingSpecs" }
+              },
+              "uiMaxFlowRate": {
+                "type": "string",
```
sanni-t:
```suggestion
                "type": "number",
```
or probably better is to use a reference to the existing `positiveNumber`

sfoster1 (approved):
Looks good to me!

PR #14862, opened 2024-04-10
feat(performance-metrics): add RobotContextTracker

DerekMaggio:
# Overview

Create RobotContextTracker class.

This class provides a `track` method. 
The track method should be used as a decorator on either a function/method. It takes a RobotContextState enum value to label what state the RobotContextTracker is tracking. Uses `FunctionTimer` to measure execution time and stores results in a list. 

RobotContextTracker is defaulted not to track anything at all. To turn on tracking, instantiate the class with `should_track=True`. When not tracking, the `track` method calls the underlying wrapped function as quickly as possible.

# Test Plan

- See test_robot_contest_tracker.py

# Changelog

- Add RobotContextTracker class
- Add test_robot_context_tracker.py

# Review requests

None

# Risk assessment

Low, not being used on any production code


performance-metrics/src/performance_metrics/robot_context_tracker.py
```diff
@@ -0,0 +1,58 @@
...
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+class RobotContextTracker:
+    """Tracks and stores robot context and execution duration for different operations."""
+
+    def __init__(self, should_track: bool = False) -> None:
+        """Initializes the RobotContextTracker with an empty storage list."""
+        self._storage: List[RawContextData] = []
```
sfoster1: use a `deque` for this. python lists are dynamically-allocated memory arrays and any array append may cause a reallocation, which will cause an uneven perf impact. the deque will have more bounded performance, and I don't believe we're ever planning to access items in here by-index anyway.

sfoster1 (changes requested):
One small change but looking great, love the decorator approach

sfoster1 (approved):
Looks good to me!

PR #14865, opened 2024-04-11
feat(system-server): add /system/oem_mode/upload_splash endpoint to change boot splash

vegano1:
# Overview

Now that we can enable/disable OEM mode, we need an endpoint to upload an image file which will become the splash screen. This pull request adds the endpoint, which takes a file and validates it to make sure it's a PNG image, < 5 MB in size, and is 1024x600 in dimension. Once validated, the file is saved to its final location in `/var/lib/opentrons-system-server/oem_mode/{filename}` then we update `OT_SYSTEM_SERVER_oem_mode_splash_custom` dotenv file with the same location.

Closes: [PLAT-277](https://opentrons.atlassian.net/browse/PLAT-277)

# Test Plan

- [x] Make sure we only process the request if OEM mode has been enabled
- [x] Make sure we only handle valid images, png, <5mb
- [x] Make sure that we can upload a PNG image that gets stored in `/var/lib/opentrons-system-server/oem_mode/{filename}`
- [x] Make sure that we set the `OT_SYSTEM_SERVER_oem_mode_splash_custom` to `/var/lib/opentrons-system-server/oem_mode/{filename}` in the `system.env` file
- [x] Make sure that the splash screen changes after rebooting the system

# Changelog

- added `/system/oem_mode/upload_splash` endpoint to the system server

# Review requests

# Risk assessment

Low, unreleased

[PLAT-277]: https://opentrons.atlassian.net/browse/PLAT-277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


system-server/system_server/system/oem_mode/router.py
```diff
@@ -36,3 +44,84 @@ async def enable_oem_mode_endpoint(
...
+    ):
+        raise HTTPException(
+            status_code=status.HTTP_415_UNSUPPORTED_MEDIA_TYPE,
+            detail="Unsupported file type",
+        )
+
+    file_size = 0
+    for chunk in file.file:
+        file_size += len(chunk)
+        if file_size > 5242880:  # 5MB
```
sfoster1: nit: can we make this the literal sequence `5 * 1024 * 1024`

system-server/system_server/system/oem_mode/router.py
```diff
@@ -35,3 +46,87 @@ async def enable_oem_mode_endpoint(
...
+    file_size = 0
+    for chunk in file.file:
+        file_size += len(chunk)
+        if file_size > 5 * 1024 * 1024:  # 5MB
+            raise HTTPException(
+                status_code=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+                detail="File is larger than 5mb.",
+            )
+
+    # TODO: Validate image dimensions
```
CaseyBatten, line 108: This TODO can be removed since we will no longer be validating dimensions on this side.

system-server/system_server/system/oem_mode/router.py
```diff
@@ -35,3 +46,87 @@ async def enable_oem_mode_endpoint(
...
+    summary="Upload an image to be used as the boot up splash screen.",
+    responses={
+        status.HTTP_201_CREATED: {"message": "OEM Mode splash screen uploaded"},
+        status.HTTP_400_BAD_REQUEST: {"message": "OEM Mode splash screen not set"},
+        status.HTTP_413_REQUEST_ENTITY_TOO_LARGE: {
+            "message": "File is larger than 5mb"
+        },
+        status.HTTP_415_UNSUPPORTED_MEDIA_TYPE: {"message": "Invalid file type"},
+        status.HTTP_500_INTERNAL_SERVER_ERROR: {
+            "message": "OEM Mode splash unhandled exception."
```
CaseyBatten, line 62: Status responses all make sense and are error codes in line with what would reasonably be expected on receiving end.

CaseyBatten (approved):
Looks good and links together as expected based on robot-server side changes as well, flow for client->system server endpoint relationship makes sense from discussions on how the behavior is split between the two.

PR #14866, opened 2024-04-11
feat(robot-server): add runtime parameter definitions to run summary

jbleon95:
# Overview

Closes AUTH-105

This PR adds the runtime parameter definitions to the run summary for both current and non current runs, accessible via the `GET` `/runs` and `/runs/{run_id}` endpoints. These definitions contain all parameter settings, including the value used during the run itself. An additional column, `run_time_parameters` was added to the run table in the database, along with a migration step for it in the existing v3 to v4 database schema migration added in PR #14762

Because we don't actually parse a run's parameters until the run has first started, when a run is first created, it will not have any runtime parameters. This matches the pattern for other values in the run summary, like `labware` and `pipettes` that don't get populated until it reaches that step of the protocol (typically the beginning by convention). This means if a protocol is cancelled before it starts, the RTP values will be empty for that summary. Otherwise, it will reflect the parameters used.

# Test Plan

On robot testing using postman to ensure that current and non current runs returned the correct parameters in the response to the `GET` runs endpoints. Also tested on robot that the migration worked and that older runs from v3 schema were properly migrated to the v4 schema. Unit and integration test coverage along with this as well.

# Changelog

- Added `runTimeParameters` field to the `Run` `ResourceModel` returned via the `GET` `/runs` and `/runs/{run_id}` endpoints.
- Added new column `run_time_parameters` to the run table for v4 database schema, along with serialization and deserialization methods for a list of pydantic objects to and from json

# Review requests

# Risk assessment

Lowish-medium, adds new data to an existing endpoint but changes are small in scope and have been tested and covered by automated tests.

### DIFF
--- a/robot-server/robot_server/persistence/_migrations/v3_to_v4.py
+++ b/robot-server/robot_server/persistence/_migrations/v3_to_v4.py
@@ -50,3 +50,8 @@ class Migration3to4(Migration):  # noqa: D101
                 schema_4.analysis_table.name,
                 schema_4.analysis_table.c.run_time_parameter_values_and_defaults,
             )
+            add_column(
+                dest_engine,
+                schema_4.run_table.name,
+                schema_4.run_table.c.run_time_parameters,
+            )
--- a/robot-server/robot_server/persistence/_migrations/v3_to_v4.py
+++ b/robot-server/robot_server/persistence/_migrations/v3_to_v4.py
@@ -3,6 +3,7 @@
 Summary of changes from schema 3:
 
 - Adds a new "run_time_parameter_values_and_defaults" column to analysis table
+- Adds a new "run_time_parameters" column to run table
 """
 
 from pathlib import Path
--- a/robot-server/robot_server/runs/run_models.py
+++ b/robot-server/robot_server/runs/run_models.py
@@ -124,10 +124,9 @@ class Run(ResourceModel):
     runTimeParameters: List[RunTimeParameter] = Field(
         default_factory=list,
         description=(
-            "Run time parameters used during analysis."
+            "Run time parameters used during the run."
             " These are the parameters that are defined in the protocol, with values"
-            " specified either in the protocol creation request or reanalysis request"
-            " (whichever started this analysis), or default values from the protocol"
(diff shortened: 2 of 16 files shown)


robot-server/robot_server/persistence/_migrations/v3_to_v4.py
```diff
@@ -50,3 +50,8 @@ def add_column(
                 schema_4.analysis_table.name,
                 schema_4.analysis_table.c.run_time_parameter_values_and_defaults,
             )
+            add_column(
+                dest_engine,
+                schema_4.run_table.name,
+                schema_4.run_table.c.run_time_parameters,
+            )
```
SyntaxColoring, line 58: Can we add a note for this in the docstring at the top of this file?

jbleon95:
All in favor of the rename, though I'd personally prefer `request_stop` since I generally prefer verb_noun as function names rather than verb_adverb

robot-server/robot_server/runs/run_models.py
```diff
@@ -121,6 +121,16 @@ class Run(ResourceModel):
...
         description="Labware offsets to apply as labware are loaded.",
     )
+    runTimeParameters: List[RunTimeParameter] = Field(
+        default_factory=list,
+        description=(
+            "Run time parameters used during analysis."
+            " These are the parameters that are defined in the protocol, with values"
+            " specified either in the protocol creation request or reanalysis request"
+            " (whichever started this analysis), or default values from the protocol"
+            " if none are specified in the request."
```
sanni-t:
```suggestion
            "Run time parameters used during the run."
            " These are the parameters that are defined in the protocol, with values"
            " specified either in the run creation request or default values from the protocol"
            " if none are specified in the request."
```

robot-server/robot_server/runs/run_models.py
```diff
@@ -185,6 +195,16 @@ class BadRun(ResourceModel):
...
+        ),
```
sanni-t: Same as above

sanni-t (approved):
Looks great! 
Just the model field description change request.

PR #14867, opened 2024-04-11
chore(api,shared-data): Require Python >=3.10, not >=3.8

SyntaxColoring:
# Overview

Closes EXEC-391.

# Test Plan

Not sure. Is there a way to do something like  a private PyPI deploy to test this?
# Changelog

* Remove Python 3.8 and 3.9 from `CLASSIFIERS`. As far as I know, these are just human-readable tags, and don't have any behavioral effect.
* Change `python_requires` from `>=3.8` to `>=3.10` for `api` and `shared-data`.

Mercifully, I think we can do this without having to do another `pipenv lock`, which would conflict #14805. `pipenv verify` doesn't complain.

# Review requests

Other than `api` and `shared-data`, are there any other user-facing projects that this needs to happen in?

# Risk assessment

Low.

### DIFF
--- a/api/setup.py
+++ b/api/setup.py
@@ -46,8 +46,6 @@ CLASSIFIERS = [
     "Intended Audience :: Science/Research",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Topic :: Scientific/Engineering",
 ]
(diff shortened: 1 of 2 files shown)


SyntaxColoring:
The "shared-data test/lint/deploy" workflow is failing for reasons unrelated to this PR; it's broken in `edge` too.

sfoster1 (approved):
This looks correct to me, but also check `hardware` (I think it's fine though)

SyntaxColoring:
Looks like `hardware` is already narrowed to 3.10.

PR #14868, opened 2024-04-11
Module ramp rate to google sheet

rclarke0:
# Overview

Calculates module ramp rates based on run log and uploads to google sheet.

# Test Plan

Ramp rate script tested on all three modules with different robots. 

# Changelog

Created module ramp rate script to find ramp rate runs in run log folder and upload ramp rates to abr-run-data sheet
Also changed IP address in error recording to a user input rather than an input in order to allow the command to be created into a desktop shortcut.

# Review requests

# Risk assessment

### DIFF
--- a/abr-testing/abr_testing/data_collection/abr_google_drive.py
+++ b/abr-testing/abr_testing/data_collection/abr_google_drive.py
@@ -43,77 +43,87 @@ def create_data_dictionary(
                 file_results = json.load(file)
         else:
             continue
-        run_id = file_results.get("run_id", "NaN")
-        if run_id in runs_to_save:
-            robot = file_results.get("robot_name")
-            protocol_name = file_results["protocol"]["metadata"].get("protocolName", "")
-            software_version = file_results.get("API_Version", "")
-            left_pipette = file_results.get("left", "")
-            right_pipette = file_results.get("right", "")
-            extension = file_results.get("extension", "")
-            (
-                num_of_errors,
-                error_type,
-                error_code,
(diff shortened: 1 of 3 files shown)


abr-testing/abr_testing/data_collection/abr_google_drive.py
```diff
@@ -43,77 +43,87 @@ def create_data_dictionary(
...
-            right_pipette = file_results.get("right", "")
-            extension = file_results.get("extension", "")
-            (
-                num_of_errors,
-                error_type,
-                error_code,
-                error_instrument,
-                error_level,
-            ) = read_robot_logs.get_error_info(file_results)
+        if isinstance(file_results, dict):
```
sfoster1: with something like this, it's usually a good idea to split out these checks and the code inside them by creating some boundary where 
- we do some validity checks before entering the boundary, and therefore
- once inside the boundary, the data is known to be valid

A common way to do it would be to have a function that takes a dict and only a dict - and doesn't check - and returns some data; and before calling the function you check that it's a dict.

Here though a good intermediate step would be to invert the logic to
```
if not isinstance(file_results, dict):
    continue
run_id = file_results.get("run_id", "NaN")
...
```

sfoster1 (approved):

abr-testing/abr_testing/data_collection/abr_google_drive.py
rclarke0: Gotcha that makes sense.

PR #14869, opened 2024-04-11
fix(shared-data): adapt 96 3.6 to new schema

sfoster1:
The schema changes in edge weren't in release and need to be manually merged.

Closes RQA-2558


mjhuff (approved):
Thank you!

Issue #13983, opened 2023-11-15
feat: Upgrade pydantic to v2


adrijanik:
### Overview

Hi team, 

Do you have any plans to upgrade pydantic to v2 from v1.8.2? Is there maybe any branch that I could use if such transition is happening? It becomes a problem when you want to use opentrons library in a project where packages dependencies requires pydantic v2. I couldn't find any issue that address that.

Thanks!

### Implementation details

I haven't tested that, but this could be a simple fix by using v1 apis from within v2 version, see below:
`from pydantic.v1 import BaseModel` 

There is also a helper package `bump-pydantic` which may help with that. Here is also a [migration guide](https://docs.pydantic.dev/latest/migration/).

### Acceptance criteria

Lift the dependency from pydantic 1.8.2 to pydantic >=2.

SyntaxColoring:
Yeah, we definitely want to upgrade Pydantic. (Internal reference: RSS-227)

Unfortunately we haven't started that work yet, so there's no branch you could use. We don't have a timeline for it, since we're focusing on the new Opentrons Flex.

> There is also a helper package bump-pydantic which may help with that. Here is also a [migration guide](https://docs.pydantic.dev/latest/migration/).

This is helpful. Thank you!

ChrisTwig:
Hello @SyntaxColoring is there any proposed plan for this?
You could even use the new namespace pydantic.v1

SyntaxColoring:
Sorry for the lack of updates here. Yes, we're working on this in PR #14871. It's not our first priority, though, so I can't promise it will be completed soon.

silvioscience:
Hello @SyntaxColoring, any updates on this?

SyntaxColoring:
Latest update: https://github.com/Opentrons/opentrons/pull/14871#issuecomment-2294227559

strangemonad:
please please please prioritize this. Pydantic should have fixed many import perf issues. It's a major pain that we can't easily interoperate between core models in parts of our codebase.

SyntaxColoring:
Done in PR #14871.

PR #14871, opened 2024-04-11
refactor: Migrate Python projects from Pydantic v1 to v2

ahiuchingau:
[PR originally by @ahiuchingau, current description by @SyntaxColoring]

# Overview

This updates our robot-stack Python projects from Pydantic v1 to v2.

Closes PLAT-326 and GitHub issue #13983.

Affected projects that run on the robot:

 - api
 - robot-server
 - shared-data
 - hardware
 - server-utils
 - system-server
 - performance-metrics (re-lock only, depends on shared-data)

Affected projects that only run in CI and on our laptops:

 - g-code-testing
 - hardware-testing
 - abr-testing

# Test Plan

When the https://github.com/Opentrons/buildroot and https://github.com/Opentrons/oe-core changes are ready:

* [x] OT-2
    * [x] Make sure all servers still boot without errors (`systemctl status`)
    * [ ] No unexpected mismatches reported by `pip check`
        * I get "robot-server 8.2.0 has requirement pydantic==2.9.0, but you have pydantic 2.9.2." I think we can fix this in a follow-up.
    * [x] Try running a protocol or something and make sure the server doesn't return an error or take an obscenely long time to respond.
* [x] Flex
    * [x] Make sure all servers still boot on a Flex without errors (`systemctl status`)
    * [x] No unexpected mismatches reported by `pip check`
    * [x] Try running a protocol or something and make sure the server doesn't return an error or take an obscenely long time to respond.

Beyond that, this PR touches a million little things in a million little ways, so it's difficult to test. We should try to merge it early in the release cycle to give us time to shake things out.

# Changelog

See https://docs.pydantic.dev/latest/migration/#migration-guide for everything that has changed from Pydantic v1 to v2.

The basic methodology of this PR is:

* Update `setup.py`, `Pipfile`, and `Pipfile.lock` files to a new Pydantic version, trying to follow  Unfortunately, FastAPI is tightly coupled to Pydantic, so we need to update it too. The FastAPI bump is kept minimal.
* Run [bump-pydantic](https://github.com/pydantic/bump-pydantic) on all projects. This automatically does a lot of the grunt work, but it does need manual follow-up.
* Manually fix up lots of little things
* Do global find+replaces for some [trivial renames](https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel). I moved some of this to [a separate PR, #17123](https://github.com/Opentrons/opentrons/pull/17123), because the GitHub web UI was struggling with the big diff.

# Review requests

- Do the setup.py, Pipfile, and Pipfile.lock files look good? We're trying to align on a definition of "good" [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/4671602797/Python+dependency+management).
- Do all Pydantic migrations look correct? Reference: https://docs.pydantic.dev/latest/migration/#migration-guide
    - Do the rewritten validators look correct? Some of these needed manual intervention.
    - Do the `= None` additions look correct? `bump-pydantic` added these automatically to match prior parse behavior—see https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields. As far as I can tell, these additions are always safe. But defaulting to `None` may not be what we actually want, e.g. it may not match the underlying JSON schema.
- This was a long-lived PR that changed hands several times, so there are definitely vestigial things left over from earlier attempts. If something looks unexplained or out of place to you, please speak up.

# Risk assessment

## Performance

This *will,* at least in the short term, make robot-server take much longer to start up, and make the tests much slower. This is a known Pydantic v2 problem (https://github.com/pydantic/pydantic/issues/6768 etc.). Earlier testing on a Flex found it slowed from 46s to 1m54s (2.5x). I don't think we'll be able to get it back down to Pydantic v1 times, but some proofs of concept suggest that we can mitigate it to only ~1.6x slower. There are some ideas in EXEC-1060.

## Correctness

High-risk due to the breadth of changes: storage reads and writes, HTTP requests and responses, communication between the `opentrons.protocol_api` and `opentrons.protocol_engine`, ...

* [Pydantic's type coercion behavior has changed](https://docs.pydantic.dev/latest/migration/#validator-behavior-changes), trending in the direction of doing less type coercion. This is a good direction, but it is basically impossible to audit affected one Python protocol in the snapshot tests—see the inline review comments below.


api/src/opentrons/calibration_storage/ot2/models/v1.py
```diff
@@ -1,19 +1,24 @@
...
 from opentrons_shared_data.pipette.dev_types import LabwareUri
 
 from opentrons.types import Point
 from opentrons.calibration_storage import types
 
+DatetimeType = typing.Annotated[
+    datetime,
+    PlainSerializer(lambda x: x.isoformat(), when_used="json"),
+]
+
```
SyntaxColoring, line 16: I'm inferring that this is replacing the old `json_encoders` config. Does something need to replace the old `json_decoders` config, to match?

api/src/opentrons/calibration_storage/ot3/models/v1.py
```diff
@@ -12,15 +12,21 @@
 from opentrons.calibration_storage import types
 
 
+DatetimeType = typing.Annotated[
+    datetime,
+    PlainSerializer(lambda x: x.isoformat(), when_used="json"),
+]
```
SyntaxColoring, line 18: Ditto.

api/src/opentrons/protocol_engine/clients/sync_client.py
SyntaxColoring, line 1: This file should continue to have `cast()`s, not `model_validate()`s.

The `result` objects in this case are coming from a trusted source—our own `ProtocolEngine` code—so revalidating them would just add overhead and slow down protocol simulation.

api/src/opentrons/protocol_engine/commands/command_unions.py
```diff
@@ -542,6 +542,11 @@
     Field(discriminator="commandType"),
 ]
 
+# Each time a TypeAdapter is instantiated, it will construct a new validator and
+# serializer. To improve performance, TypeAdapters are instantiated once.
+# See https://docs.pydantic.dev/latest/concepts/performance/#typeadapter-instantiated-once
+CommandCreateAdatper: TypeAdapter[CommandCreate] = TypeAdapter(CommandCreate)  # type: ignore[arg-type]
```
SyntaxColoring: 1. Typo: `Adatper` instead of `Adapter`
2. What is this `TypeAdapter` doing?

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -230,7 +230,7 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
             # request > command mapping, figure out how to type precisely
             # (or wait for a future mypy version that can figure it out).
             # For now, unit tests cover mapping every request type
-            queued_command = action.request._CommandCls.construct(
```
SyntaxColoring: Can this be `.model_construct`, for performance reasons?

shared-data/python/opentrons_shared_data/types.py
```diff
@@ -0,0 +1,24 @@
...
+
+NumberType = TypeVar("NumberType")
+
+
+class Vec3f(BaseModel, Generic[NumberType]):
+    """A 3D Vector."""
+
+    x: NumberType
+    y: NumberType
+    z: NumberType
```
SyntaxColoring: 1. The name `Vec3f` comes from "vector of 3 `float`s." If this is generic across other kinds of `NumberType`, let's rename it to just `Vec3`.
2. Since we're really pushing this deep down into common shared code, I'm worried that someone will come along later and add an `= 0` default for something without realizing that it will affect a million things. Let's spell out that this is specifically for vectors where all elements are required, to hopefully prompt people to use a different type if that's not the behavior that they want.

```suggestion
class Vec3f(BaseModel, Generic[NumberType]):
    """A 3D vector where all elements are required."""

    x: NumberType
    y: NumberType
    z: NumberType
```

shared-data/python/tests/pipette/test_validate_schema.py
```diff
@@ -68,3 +70,48 @@ def test_pick_up_configs_tip_count_keys() -> None:
...
+    quirk_0 = types.Quirks.pickupTipShake
+    quirk_1 = types.Quirks.dropTipShake
+    loaded_model.quirks = [quirk_0, quirk_1]
+
+    assert loaded_model.pipette_type == types.PipetteModelType.p1000
+    assert loaded_model.display_category == types.PipetteGenerationType.FLEX
+    assert loaded_model.channels == types.PipetteChannelType.NINETY_SIX_CHANNEL
+
+    model_dict = loaded_model.model_dump()
+    # each field should be the value of the enum class
```
SyntaxColoring, line 167: Nice.

SyntaxColoring (commented):
I'm about 60% through the changes so far. Mostly looks great, only a few comments and questions:

api/src/opentrons/protocol_runner/legacy_command_mapper.py
SyntaxColoring, line 1: Same as the other comment—for performance reasons, can we change the `.construct()`s to `.model_construct()`s instead of changing them to fully-validating `__init__()` calls?

robot-server/robot_server/log.txt
SyntaxColoring, line 1: File left over from debugging?

robot-server/setup.py
```diff
@@ -56,10 +56,10 @@ def get_version():
     f"opentrons-shared-data=={VERSION}",
     f"server-utils=={VERSION}",
     "anyio==3.7.1",
-    "fastapi==0.99.1",
+    "fastapi>=0.100.0",
     "python-dotenv==1.0.1",
     "python-multipart==0.0.6",
-    "pydantic==1.10.12",
+    "pydantic>=2.0.0,<3",
```
SyntaxColoring: Do these need to be pinned to a specific version like they were before?

SyntaxColoring (commented):
I'm caught up!

api/src/opentrons/protocol_engine/clients/sync_client.py
SyntaxColoring: Made this change.

robot-server/robot_server/log.txt
SyntaxColoring: Removed.

robot-server/setup.py
ecederstrand: Please don't pin to exact versions unless strictly necessary. See e.g. https://github.com/Opentrons/opentrons/issues/11905

SyntaxColoring: This is `robot-server`, which is not published on PyPI, so it should not affect what gets pulled in when you do `pip install opentrons`.

I'm a bit shaky on our packaging mechanisms, but I think this specific setup.py may not even affect what we install on robots. I'm sure we can make this less confusing somehow. But for the sake of not changing too many things in this PR, I think we'll keep the `==`.

[Edit: Opened a discussion [here](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/4671602797/Python+dependency+management) for Opentrons people and added this to the review requests.]

SyntaxColoring:
An update on where this stands at the time of writing:

* The big problem to solve is that there's a pretty severe hit to `robot-server` startup time. On Flex, it goes from ~46s to ~1m54s, and it's probably worse on OT-2.
    * The Pydantic people are [aware of the problem](https://github.com/pydantic/pydantic/discussions/6748#discussioncomment-9619776), but it doesn't look like there will be an imminent fix.
    * [`defer_build`](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.defer_build) might help, but some quick experiments didn't show it making a huge difference. I wouldn't expect it to shave more than 20 seconds, and that's being optimistic.
    * [There is a large FastAPI component to the problem](https://github.com/pydantic/pydantic/issues/6768#issuecomment-1644532429), having to do with the way it nests routers within each other. We might be able to work around it by flattening our routers. I don't know exactly what that would look like yet. A proof of concept would be a good next step.
* There are still various tests that are failing for one reason or another. So far, I haven't seen anything that fundamentally concerns me. Just a lot of little things to work through.
* `edge` is a moving target and there are always going to be more merge conflicts. One thing that would help is ignoring the trivial rename `DeprecationWarning`s for now—maybe even remove the fixes that we have so far from this PR—and fix them in incremental follow-ups.

sfoster1:
Per https://github.com/pydantic/pydantic/discussions/6748#discussioncomment-10396844 now may be the time to evaluate the latest pydantic2 release

strangemonad:
please please please prioritize this. Pydantic should have fixed many import perf issues. It's a major pain that we can't easily interoperate between core models in parts of our codebase.

api/src/opentrons/calibration_storage/ot2/models/v1.py
SyntaxColoring: Resolving my own question:

I'm not sure why this code needed to customize datetime encoding in the first place, but it is *probably* safe to omit customization of decoding. I expect Pydantic's default decoding to be able to decode whatever `datetime.isoformat()` encodes.

api/src/opentrons/protocol_runner/legacy_command_mapper.py
SyntaxColoring: Did this.

shared-data/command/schemas/8.json
SyntaxColoring, line 1: We probably want to revert changes to schema 8. Schema 11 is current at the time of writing.

shared-data/python/opentrons_shared_data/types.py
SyntaxColoring: Reverted these vector refactors in fe5005e. We can un-revert them in another PR.

analyses-snapshot-testing/tests/__snapshots__/analyses_snapshot_test/test_analysis_snapshot[2bc6830494][pl_langone_ribo_pt1_ramp].json
Comment (SyntaxColoring) [inline analyses-snapshot-testing/tests/__snapshots__/analyses_snapshot_test/test_analysis_snapshot[2bc6830494][pl_langone_ribo_pt1_ramp].json:1]: This is the one analysis snapshot change that is an actual change.

This Python protocol did something like `module.load_labware("load_name", 8)`. That's a bug in the protocol: the second argument is supposed to be a display name string.

Pydantic v1 coerced the `int` to a `str`. Pydantic v2 no longer does that particular coercion (though it still does most other coercions), so now the protocol raises an error.

If we think it's important to preserve the prior behavior here (be bug-for-bug compatible), I think we need to do that in a separate PR. It would probably entail some larger audit of arguments in `opentrons.protocol_api`; it's very unlikely that this is the only place where this could happen.

analyses-snapshot-testing/tests/__snapshots__/analyses_snapshot_test/test_analysis_snapshot[004ebb2b82][OT2_S_v2_11_P10S_P300M_MM_TC1_TM_Swift].json
Comment (SyntaxColoring) [inline analyses-snapshot-testing/tests/__snapshots__/analyses_snapshot_test/test_analysis_snapshot[004ebb2b82][OT2_S_v2_11_P10S_P300M_MM_TC1_TM_Swift].json:1]: All of the diffs in `analyses-snapshot-testing/` are the same. Some floats used to get serialized like `0` and now they're getting serialized like `0.0`. I've already looked through these files, so I don't think you have to.

There is one exception: https://github.com/Opentrons/opentrons/pull/14871/files#r1884999286

shared-data/command/schemas/11.json
SyntaxColoring, line 1: Reminder to self: Probably want to repeat the fixups in 74606e673b0307dfb6807dba685f862591be9aa5 on 1b9542140e1c8ac6d6b4e91b1e2320b28614d9bd.

robot-server/robot_server/runs/run_store.py
```diff
@@ -433,6 +434,7 @@ def get_run_time_parameters(self, run_id: str) -> List[RunTimeParameter]:
 
         try:
             return (
+                # TODO
```
SyntaxColoring: Reminder to self to look into whatever this was and clean up the `# TODO`.

robot-server/robot_server/persistence/_migrations/_up_to_3_worker.py
```diff
@@ -85,12 +85,9 @@ def migrate_commands_for_run(
             _legacy_pickle.loads(old_commands_bytes) if old_commands_bytes else []
         )
 
+        # TODO Fold the list into this?
```
SyntaxColoring: Reminder to near-future self to discern what mid-past self was getting at here so that slightly-less-near-future self can clean up this `# TODO`

robot-server/robot_server/service/json_api/response.py
```diff
@@ -41,29 +40,38 @@ class BaseResponseBody(BaseModel):
...
+        With Pydantic v1, the OpenAPI spec described `Optional`(i.e., possibly
+        `None`-valued) fields as omittable, but not nullable. This did not match
+        Pydantic's actual serialization behavior, which serialized Python `None` to
+        JSON `null` by default. This method override fixed the mismatch by making
+        Pydantic omit the field from serialization instead.
+
+        With Pydantic v2, the OpenAPI spec does describe `Optional` fields as nullable,
+        matching Pydantic's serialization behavior. We therefore no longer need this
+        override to make them match. However, removing this override and changing
+        serialization behavior at this point would risk breaking things on the client.
```
SyntaxColoring, line 56: Heads up for this.

sfoster1 (approved):
AAAhhhhhhh

vegano1 (approved):
thank you.

PR #14872, opened 2024-04-11
fix(protocol-designer, components): discarding vs delete step form button text

jerader:
closes AUTH-319

# Overview

@koji noticed this bug initially where if you delete a step via the close modal for a form, it was different from deleting a step from the delete button in the form. Upon investigating, the actions were actually correct so I modified the button text to be more clear. When closing from the close modal, you are basically discarding of an unsaved step. So you aren't deleting anything, you are just closing/discarding of it.

Additionally, the `started deck state` and `final deck state` now is highlighted again when it is active (along with the active color in the liquids tab and app in the settings tab), to match legacy behavior that was broken starting in PD 6.0.0 (see that it is not broken here: https://sandbox.designer.opentrons.com/protocol-designer@5.0.0/)

# Test Plan

Create a flex or ot-2 protocol. Go to deck map and see that the starting deck state is blue. Add a step and before saving it, delete it via the the "delete step" button and see that it navigates you to the starting deck state. Now create a new step and delete it via the "cancel" button and see that it just discards of the step and does not navigate you to the starting deck state.

# Changelog

- fix the classname shenanigans issue with the text not being highlighted blue
- update button text 

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/components/src/lists/TitledList.tsx
+++ b/components/src/lists/TitledList.tsx
@@ -101,7 +101,7 @@ export function TitledList(props: TitledListProps): JSX.Element {
     iconProps && iconProps.className
   )
 
-  let textColor = 'auto'
+  let textColor = ''
   if (disabled) {
     //  the below hex code is for our legacy --c-font-disabled to match other text colors
     textColor = '#9c9c9c'
--- a/components/src/lists/TitledList.tsx
+++ b/components/src/lists/TitledList.tsx
@@ -2,9 +2,12 @@
 import * as React from 'react'
 import cx from 'classnames'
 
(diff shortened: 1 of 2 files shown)


components/src/lists/TitledList.tsx
```diff
@@ -98,6 +101,15 @@ export function TitledList(props: TitledListProps): JSX.Element {
     iconProps && iconProps.className
   )
 
+  let textColor = 'auto'
```
koji: need to set `auto` instead of '' or `undefined`?

koji (approved):
left a comment.

PR #14873, opened 2024-04-11
feat(app): add generic run paused splash screen

mjhuff:
Closes [EXEC-387](https://opentrons.atlassian.net/browse/EXEC-387)

# Overview

Add a fallback run paused splash screen that will show when no specific errorType is handled explicitly. The exact logic used to display the splash message will most likely change (and we'll add specific error type subtext later). 

Screenshot 2024-04-11 at 1 41 31 PM

# Test Plan
- Make sure the ODD enable run notes FF is on.
- `curl --location '<robot ip>:31950/settings' \
--header 'opentrons-version: *' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--data '{
  "id": "enableErrorRecoveryExperiments",
  "value": true
}'` (Please set this to false when you're done testing).
- Run anything that will throw a pick up tip failed error (make sure `edge` is pushed). Verify that the splash screen appears (note that cancelling the run after this point doesn't work currently).

# Changelog
- Added "Run paused" splash screen.

# Risk assessment
low

[EXEC-387]: https://opentrons.atlassian.net/browse/EXEC-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { useTranslation } from 'react-i18next'
 
 import {
   Flex,
@@ -27,7 +28,13 @@ export function RunPausedSplash({
   errorType,
   protocolName,
 }: RunPausedSplashProps): JSX.Element {
-  const subText = protocolName ?? null
+  const { t } = useTranslation('error_recovery')
+
+  let subText: string | null
+  switch (errorType) {
+    default:
+      subText = protocolName ?? null
+  }
 
   return (
     <Btn
@@ -45,7 +52,7 @@ export function RunPausedSplash({
       <SplashFrame>
         <Flex gridGap={SPACING.spacing32} alignItems={ALIGN_CENTER}>
           <Icon name={'ot-alert'} size="4.5rem" color={COLORS.white} />
-          <SplashHeader>{'Run paused'}</SplashHeader>
+          <SplashHeader>{t('run_paused')}</SplashHeader>
         </Flex>
         <Flex width="49rem" justifyContent={JUSTIFY_CENTER}>
           <SplashBody>{subText}</SplashBody>
--- a/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
@@ -1,0 +1,87 @@
+import * as React from 'react'
+
+import {
+  Flex,
+  Btn,
+  Icon,
+  JUSTIFY_CENTER,
+  ALIGN_CENTER,
+  SPACING,
+  COLORS,
+  DIRECTION_COLUMN,
+  POSITION_ABSOLUTE,
+  TYPOGRAPHY,
+  BORDERS,
+  OVERFLOW_WRAP_BREAK_WORD,
+} from '@opentrons/components'
+import styled from 'styled-components'
+
+interface RunPausedSplashProps {
+  onClose: () => void
+  errorType?: string
+  protocolName?: string
+}
+
+export function RunPausedSplash({
+  onClose,
+  errorType,
+  protocolName,
+}: RunPausedSplashProps): JSX.Element {
+  const subText = protocolName ?? null
+
+  return (
+    <Btn
+      height="100vh"
+      width="100%"
+      justifyContent={JUSTIFY_CENTER}
+      alignItems={ALIGN_CENTER}
+      position={POSITION_ABSOLUTE}
+      flexDirection={DIRECTION_COLUMN}
+      gridGap={SPACING.spacing40}
+      padding={SPACING.spacing40}
+      backgroundColor={COLORS.grey50}
+      onClick={onClose}
+    >
+      <SplashFrame>
+        <Flex gridGap={SPACING.spacing32} alignItems={ALIGN_CENTER}>
+          <Icon name={'ot-alert'} size="4.5rem" color={COLORS.white} />
+          <SplashHeader>{'Run paused'}</SplashHeader>
+        </Flex>
+        <Flex width="49rem" justifyContent={JUSTIFY_CENTER}>
+          <SplashBody>{subText}</SplashBody>
+        </Flex>
+      </SplashFrame>
+    </Btn>
+  )
+}
+
+const SplashHeader = styled.h1`
+  font-weight: ${TYPOGRAPHY.fontWeightBold};
+  text-align: ${TYPOGRAPHY.textAlignLeft};
+  font-size: 80px;
+  line-height: 94px;
+  color: ${COLORS.white};
+`
+const SplashBody = styled.h4`
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 4;
+  overflow: hidden;
+  overflow-wrap: ${OVERFLOW_WRAP_BREAK_WORD};
+  font-weight: ${TYPOGRAPHY.fontWeightSemiBold};
+  text-align: ${TYPOGRAPHY.textAlignCenter};
+  text-transform: ${TYPOGRAPHY.textTransformCapitalize};
+  font-size: ${TYPOGRAPHY.fontSize32};
+  line-height: ${TYPOGRAPHY.lineHeight42};
+  color: ${COLORS.white};
+`
+
+const SplashFrame = styled(Flex)`
+  width: 100%;
+  height: 100%;
+  flex-direction: ${DIRECTION_COLUMN};
+  justify-content: ${JUSTIFY_CENTER};
+  align-items: ${ALIGN_CENTER};
+  grid-gap: ${SPACING.spacing40};
+  border-radius: ${BORDERS.borderRadius8};
+`
--- a/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import styled from 'styled-components'
 import { useTranslation } from 'react-i18next'
 
 import {
@@ -13,8 +14,8 @@ import {
   POSITION_ABSOLUTE,
   TYPOGRAPHY,
   OVERFLOW_WRAP_BREAK_WORD,
+  DISPLAY_FLEX,
 } from '@opentrons/components'
-import styled from 'styled-components'
 
 interface RunPausedSplashProps {
   onClose: () => void
@@ -37,6 +38,7 @@ export function RunPausedSplash({
 
   return (
     <Btn
+      display={DISPLAY_FLEX}
       height="100vh"
       width="100%"
       justifyContent={JUSTIFY_CENTER}
@@ -64,7 +66,7 @@ export function RunPausedSplash({
 const SplashHeader = styled.h1`
   font-weight: ${TYPOGRAPHY.fontWeightBold};
   text-align: ${TYPOGRAPHY.textAlignLeft};
-  font-size: ${TYPOGRAPHY.fontSize80}
+  font-size: ${TYPOGRAPHY.fontSize80};
   line-height: ${TYPOGRAPHY.lineHeight96};
   color: ${COLORS.white};
 `
--- a/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
@@ -12,7 +12,6 @@ import {
   DIRECTION_COLUMN,
   POSITION_ABSOLUTE,
   TYPOGRAPHY,
-  BORDERS,
   OVERFLOW_WRAP_BREAK_WORD,
 } from '@opentrons/components'
 import styled from 'styled-components'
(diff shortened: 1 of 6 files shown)


app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+  errorType,
+  protocolName,
+}: RunPausedSplashProps): JSX.Element {
+  const { t } = useTranslation('error_recovery')
+
+  let subText: string | null
+  switch (errorType) {
+    default:
+      subText = protocolName ?? null
+  }
```
jerader, line 37: can't this be simplified to just `const subText = protooclName ?? null`?

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+      position={POSITION_ABSOLUTE}
+      flexDirection={DIRECTION_COLUMN}
+      gridGap={SPACING.spacing40}
+      padding={SPACING.spacing40}
+      backgroundColor={COLORS.grey50}
+      onClick={onClose}
+    >
+      <SplashFrame>
+        <Flex gridGap={SPACING.spacing32} alignItems={ALIGN_CENTER}>
+          <Icon name={'ot-alert'} size="4.5rem" color={COLORS.white} />
```
jerader: nit

```suggestion
          <Icon name="ot-alert" size="4.5rem" color={COLORS.white} />
```

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+      </SplashFrame>
+    </Btn>
+  )
+}
+
+const SplashHeader = styled.h1`
+  font-weight: ${TYPOGRAPHY.fontWeightBold};
+  text-align: ${TYPOGRAPHY.textAlignLeft};
+  font-size: 80px;
+  line-height: 94px;
```
jerader: can these be in rem?

app/src/organisms/OnDeviceDisplay/RunningProtocol/__tests__/RunPausedSplash.test.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react'
+import { MemoryRouter } from 'react-router-dom'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { fireEvent, screen } from '@testing-library/react'
+
+import { COLORS } from '@opentrons/components'
+
+import { RunPausedSplash } from '../RunPausedSplash'
+import { renderWithProviders } from '../../../../__testing-utils__'
+import { i18n } from '../../../../i18n'
```
jerader: imports can go higher

app/src/pages/RunningProtocol/__tests__/RunningProtocol.test.tsx
```diff
@@ -30,12 +31,14 @@ import { getLocalRobot } from '../../../redux/discovery'
...
 import { useTrackProtocolRunEvent } from '../../../organisms/Devices/hooks'
 import { useMostRecentCompletedAnalysis } from '../../../organisms/LabwarePositionCheck/useMostRecentCompletedAnalysis'
+import { RunPausedSplash } from '../../../organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash'
 import { OpenDoorAlertModal } from '../../../organisms/OpenDoorAlertModal'
 import { RunningProtocol } from '..'
 import {
   useNotifyLastRunCommandKey,
   useNotifyRunQuery,
 } from '../../../resources/runs'
+import { useFeatureFlag } from '../../../redux/config'
```
jerader, line 41: import can go higher

app/src/pages/RunningProtocol/index.tsx
```diff
@@ -160,114 +168,137 @@ export function RunningProtocol(): JSX.Element {
...
+                      marginTop="9.25rem"
+                      alignSelf={ALIGN_FLEX_END}
+                    />
+                  </>
+                )
+              ) : (
+                <RunningProtocolSkeleton currentOption={currentOption} />
+              )}
+              <Flex
+                marginTop="2rem"
```
jerader: theres a typography constant for this i think

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
mjhuff: The intention is we will very shortly add more `errorType`s to the list.

koji: question

what is the purpose of using switch here?
in the future displaying something here?

app/src/organisms/OnDeviceDisplay/RunningProtocol/__tests__/RunPausedSplash.test.tsx
mjhuff: One of these days I will remember this 😄

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <Btn
+      height="100vh"
+      width="100%"
+      justifyContent={JUSTIFY_CENTER}
+      alignItems={ALIGN_CENTER}
+      position={POSITION_ABSOLUTE}
+      flexDirection={DIRECTION_COLUMN}
+      gridGap={SPACING.spacing40}
+      padding={SPACING.spacing40}
```
koji: from Figma

Screenshot 2024-04-11 at 2 07 12 PM


```suggestion
      padding={SPACING.spacing120}
```

app/src/pages/RunningProtocol/index.tsx
mjhuff: I don't love using `TYPOGRAPHY` constants for margins offsets. I think we can make the case for margin constants, though.

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
mjhuff: Ah this must have been updated, thanks.

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
koji: I checked line-height and it says `normal` I think that would be 96px that we have in typography.ts. But not 100% sure, could you check the line-height value to a designer?
 
```suggestion
  font-size: ${TYPOGRAPY.fontSize80};
  line-height:${TYPOGRAPY.lineHeight96};
```

mjhuff: Sue is still fleshing the modal out. I'll tweak these as she updates them, but this makes sense, thanks!

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+`
+
+const SplashFrame = styled(Flex)`
+  width: 100%;
+  height: 100%;
+  flex-direction: ${DIRECTION_COLUMN};
+  justify-content: ${JUSTIFY_CENTER};
+  align-items: ${ALIGN_CENTER};
+  grid-gap: ${SPACING.spacing40};
+  border-radius: ${BORDERS.borderRadius8};
```
koji: need border-radius?

mjhuff: Ah yep, we did remove the white border from the splash pages, so it's unneeded now.

app/src/pages/RunningProtocol/index.tsx
koji:
```suggestion
                marginTop={SPACING.spacing32}
```

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
mjhuff: Yep, exactly. See PR description.

app/src/pages/RunningProtocol/index.tsx
mjhuff: Ah, yep that it. Thanks!

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,92 @@
...
+  JUSTIFY_CENTER,
+  ALIGN_CENTER,
+  SPACING,
+  COLORS,
+  DIRECTION_COLUMN,
+  POSITION_ABSOLUTE,
+  TYPOGRAPHY,
+  OVERFLOW_WRAP_BREAK_WORD,
+} from '@opentrons/components'
+import styled from 'styled-components'
```
koji: this import can be higher.

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  let subText: string | null
+  switch (errorType) {
+    default:
+      subText = protocolName ?? null
+  }
+
+  return (
+    <Btn
+      height="100vh"
```
koji, line 42:
```suggestion
      display={DISPLAY_FLEX}
      height="100vh"
```

koji (approved):
LGTM!

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
mjhuff: Hmmmm, let me double check this one on a robot. I want to say this might not work with a `Btn`, but if it does, I'll change it 👍

mjhuff:
@jerader @koji Thank you both for reviewing!

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
sfoster1: Reviving this - would love to see this factored into an independent, pure, exported-for-testability function e.g.
```
export /* for testability */ function subtext(props: RunPausedSplashProps, /* may require future extra args */): string | null {
   switch(...) return (...)
   return null
}

export function RunPausedSplash(...): ... {
    return (.. <SplashBody>{subtext(props)}</SplashBody>...)
}
```

This seems minor, but I know that we use the open-coded let+switch in a lot of places and then test everything with rendered-dom inspection later, and in my view it makes the component functions really big and hard to follow. I think the pattern of factoring out functions and testing them as independent units can be nicer.

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+      position={POSITION_ABSOLUTE}
+      flexDirection={DIRECTION_COLUMN}
+      gridGap={SPACING.spacing40}
+      padding={SPACING.spacing120}
+      backgroundColor={COLORS.grey50}
+      onClick={onClose}
+    >
+      <SplashFrame>
+        <Flex gridGap={SPACING.spacing32} alignItems={ALIGN_CENTER}>
+          <Icon name="ot-alert" size="4.5rem" color={COLORS.white} />
```
sfoster1, line 55: do we have a design system sizing element?

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
```diff
@@ -0,0 +1,94 @@
...
+          <SplashHeader>{t('run_paused')}</SplashHeader>
+        </Flex>
+        <Flex width="49rem" justifyContent={JUSTIFY_CENTER}>
```
sfoster1, line 58: do we have a sizing element for this? do we want to have absolute sizing here at all, or just specify padding or however you do that in css?

app/src/pages/RunningProtocol/index.tsx
```diff
@@ -160,114 +168,137 @@ export function RunningProtocol(): JSX.Element {
...
-                robotAnalyticsData={robotAnalyticsData}
-                protocolName={protocolName}
-                runStatus={runStatus}
-                currentRunCommandIndex={currentRunCommandIndex}
-                robotSideAnalysis={robotSideAnalysis}
-                runTimerInfo={{ runStatus, startedAt, stoppedAt, completedAt }}
-                lastAnimatedCommand={lastAnimatedCommand.current}
-                updateLastAnimatedCommand={(newCommandKey: string) =>
-                  (lastAnimatedCommand.current = newCommandKey)
+      {enableSplash &&
```
sfoster1, line 171: this optionality is getting kind of tough to follow. could we 
1) factor the common parts into a common function
2) put the unique parts, I guess with uniqueness defined by the feature flag, in independent functions that don't in and of themselves check the feature flag
3) have this component check the feature flag and delegate?

sfoster1 (commented):
Have some code-style stuff.

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
mjhuff: Nope

app/src/pages/RunningProtocol/index.tsx
mjhuff: Yeah sounds good!

app/src/organisms/OnDeviceDisplay/RunningProtocol/RunPausedSplash.tsx
mjhuff: Referring to the icon `size` prop? If by absolute sizing you mean a defined "4.5rem" then yeah, we need that to achieve the proper width/height of the SVG.

mjhuff:
https://github.com/Opentrons/opentrons/pull/14873#discussion_r1562619576

Can't respond to this in-line for some reason. I agree with you - this isn't a pattern we have used extensively, and I think that's historically because of how React used to work with the stack reconciler, back when I'm guessing we first used React. I think we can make the case pretty easily for it if we pull this out for use in the same file, and I'm fine doing this if the same-file approach works with you.

PR #14879, opened 2024-04-11
refactor(protocol-engine): Rename stop() and pause() -> request_stop() and request_pause()

SyntaxColoring:
# Overview

This fixes something that keeps confusing me as I work on EXEC-382.

Various things state that `ProtocolEngine.stop()` takes effect immediately—meaning, to me, that the robot's motion is stopped immediately, the protocol exits immediately, and the HTTP run is marked as `stopped` immediately. This does not seem true. It merely puts the run into a `stop-requested` state, which only later settles into a `stopped` state.

This PR adjusts some docstrings and renames `stop()` to ~~`stop_soon()`~~ `request_stop()`. ~~The name `stop_soon()` is inspired by asyncio and anyio's `call_soon()`.~~ `pause()` has the same caveat, so it's renamed to `request_pause()` for consistency.

# Test plan

None needed.

# Review requests

* ~~Taking for granted, for a moment, that the `ProtocolEngine` interface has to work like this: is `stop_soon()` a good name? Maybe `request_stop()` would be better?~~ Done.
* ~~`pause()` has the same caveat. Do we want to rename that too, for consistency?~~ Done.

# Risk assessment

No risk.

### DIFF
--- a/api/src/opentrons/protocol_engine/actions/actions.py
+++ b/api/src/opentrons/protocol_engine/actions/actions.py
@@ -55,10 +55,7 @@ class PauseAction:
 
 @dataclass(frozen=True)
 class StopAction:
-    """Stop the current engine execution.
-
-    After a StopAction, the engine status will be marked as stopped.
-    """
+    """Request engine execution to stop soon."""
 
     from_estop: bool = False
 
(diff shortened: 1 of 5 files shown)


jbleon95:
All in favor of the rename, though I'd personally prefer request_stop since I generally prefer verb_noun as function names rather than verb_adverb

mjhuff (commented):
I'm all for renaming `pause` to whatever is decided!

sfoster1 (approved):

PR #14882, opened 2024-04-11
feat(performance-metrics): store to file

DerekMaggio:
# Overview

Add functionality to store CSV data in a file

# Test Plan

- Added test cases to verify storing to a file correctly

# Changelog

- Added `store` method to RobotContextTracker
- Added data shaping to RawContextData
- Tests

# Review requests

None

# Risk assessment

Low

### DIFF
--- a/performance-metrics/src/performance_metrics/robot_context_tracker.py
+++ b/performance-metrics/src/performance_metrics/robot_context_tracker.py
@@ -24,6 +24,7 @@ class RobotContextTracker:
     def __init__(self, storage_file_path: Path, should_track: bool = False) -> None:
         """Initializes the RobotContextTracker with an empty storage list."""
         self._storage: deque[RawContextData] = deque()
+        self._storage_file_path = storage_file_path
         self._should_track = should_track
 
     def track(self, state: RobotContextState) -> Callable:  # type: ignore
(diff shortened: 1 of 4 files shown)


performance-metrics/src/performance_metrics/robot_context_tracker.py
```diff
@@ -56,3 +61,14 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
             return wrapper
 
         return inner_decorator
+
+    def store(self) -> None:
+        """Returns the stored context data and clears the storage list."""
+        stored_data = self._storage.copy()
+        self._storage.clear()
+        rows_to_write = [context_data.csv_row() for context_data in stored_data]
```
sfoster1, line 70: we should probably sort this by start time before writing just in case - once we add various queues or whatever reordering might become possible

sfoster1 (approved):
LGTM!

performance-metrics/src/performance_metrics/robot_context_tracker.py
DerekMaggio: Won't it always be sortable by start time regardless? My thought is ONLY robot context data would go in this file. Other metrics like CPU/Memory/Flash would be in a separate file. That way all this does is just push data to a file as fast as possible. 

Then when those files are pulled for analysis they can be sorted at that time.

PR #14883, opened 2024-04-11
feat(hardware-testing): enable multi sensor processing in liquid probe

ryanthecoder:
# Overview
This allows the liquid level detection script to tell the pipette to buffer the data from both pipettes and fetch them afterwards,
it will now spit out seprate CSVs for each sensor. post processing not yet updated so the final report just grabs one from each trial, will implement in EXEC-268

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/config/defaults_ot3.py
+++ b/api/src/opentrons/config/defaults_ot3.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, cast, List, Iterable, Tuple
+from typing import Any, Dict, cast, List, Iterable, Tuple, Optional
 from typing_extensions import Final
 from dataclasses import asdict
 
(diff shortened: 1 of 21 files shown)


hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -114,82 +114,177 @@ def _build_pass_step(
...
-        mount=head_node,
-        data_file=log_file,
-        file_heading=pressure_output_file_heading,
-        sensor_metadata=sensor_metadata,
-    )
-    async with sensor_capturer:
-        print("starting move group runner")
-        positions = await move_group.run(can_messenger=messenger)
-        messenger.add_listener(sensor_capturer, None)
+    print("starting move group runner")
```
sfoster1: accidental checkin?

sfoster1 (approved):
Looks good, we should remove that print though

api/src/opentrons/config/defaults_ot3.py
```diff
@@ -278,6 +321,18 @@ def _build_default_cap_pass(
 def _build_default_liquid_probe(
     from_conf: Any, default: LiquidProbeSettings
 ) -> LiquidProbeSettings:
+    output_option = _build_output_option_with_default(
+        from_conf.get("output_option", None), default.output_option
+    )
+    print(f"output_option {output_option}")
```
caila-marashaj: got a print left in here

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1372,6 +1373,16 @@ async def liquid_probe(
         can_bus_only_output = bool(
             output_option.value & OutputOptions.can_bus_only.value
         )
+        IPT_TO_SID: Dict[InstrumentProbeType, SensorId] = {
```
caila-marashaj: why not just use `sensor_id_for_instrument` here?

caila-marashaj (approved):
Couple very small things, and otherwise looks great!

PR #14888, opened 2024-04-12
fix(api): raise an error if protocol's defined parameters have duplicate variable names

jbleon95:
# Overview

Closes AUTH-333.

An unintended behavior was found where protocol analysis would let you have two parameters share the same variable name. Not only would this cause issues reporting them since we key them by variable name, the protocol itself would only use the latter value, leading to unintended behavior. I thought this case would be covered once the parameters would be set, but since they are all stored in a dictionary, that was actually never being encountered since the latest one would overwrite the previous one.

Now we have an explicit check for uniqueness, raising an appropriate error message if you have two (or more, though it'll raise on the second) variable names shared.

# Test Plan

Along with unit test coverage, ensured this protocol from the ticket now correctly raises.

```
metadata = {
    "protocolName": "Multiple RTP Variables with Same Name",
}
requirements = {"robotType": "OT-2", "apiLevel": "2.18"}
just_right: str = "This is a description"
    parameters.add_int(
        display_name="int 1",
        variable_name="variable_a",
        default=1,
        minimum=1,
        maximum=3,
        description=just_right,
    )
    parameters.add_int(
        display_name="int 2",
        variable_name="variable_b",
        default=1,
        minimum=1,
        maximum=3,
        description=just_right,
    )
    parameters.add_int(
        display_name="int 3",
        variable_name="variable_a", # duplicate variable name
        default=1,
        minimum=1,
        maximum=3,
        description=just_right,
    )
def run(context):
    for variable_name, value in context.params.get_all().items():
        context.comment(f"variable {variable_name} has value {value}")
```

# Review requests
# Risk assessment

Low, simple bug fix.

### DIFF
--- a/api/src/opentrons/protocol_api/_parameter_context.py
+++ b/api/src/opentrons/protocol_api/_parameter_context.py
@@ -52,6 +52,7 @@ class ParameterContext:
             description: A description of the parameter as it will show up on the frontend.
             unit: An optional unit to be appended to the end of the integer as it shown on the frontend.
         """
+        validation.validate_variable_name_unique(variable_name, set(self._parameters))
         parameter = parameter_definition.create_int_parameter(
             display_name=display_name,
             variable_name=variable_name,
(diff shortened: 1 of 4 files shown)


sanni-t (approved):
LGTM!

ecormany (approved):
Error message looks good. Tested with a sample protocol and it works.

PR #14890, opened 2024-04-12
refactor(protocol-designer): export modal to require app 7.3.0 or higher

jerader:
closes AUTH-340

# Overview

Basically, as outlined in the ticket, in the past we have generally always required users to update their robot to the latest version that is out at the time of PD's release mainly because there are usually new features in the app that PD depends upon. This time, there are no new features in the app 7.3.0 that PD depends upon. But we want to require 7.3.0 to minimize the app builds that QA needs to test on.

# Test Plan

in PD,  type `localStorage.clear()` in console and refresh, you should see the announcement modal. Make sure it says app 7.3.0 is required. Make a protocol and export and see that the export modal says that app 7.3.0 is required.

# Changelog

- change from 7.2.0 -> 7.3.0

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/cypress/integration/migrations.spec.js
+++ b/protocol-designer/cypress/integration/migrations.spec.js
@@ -127,7 +127,7 @@ describe('Protocol fixtures migrate and match snapshots', () => {
 
           cy.get('div')
             .contains(
-              'This protocol can only run on app and robot server version 7.2.0 or higher'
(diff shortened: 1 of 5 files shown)


koji (approved):
lgtm
tested the sandbox and it worked as expected.

PR #14891, opened 2024-04-12
Exec 285 add fixit commands

TamarZanzouri:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/.github/workflows/performance-metrics-test-lint.yaml
+++ b/.github/workflows/performance-metrics-test-lint.yaml
@@ -1,0 +1,54 @@
+# This workflow runs lint on pull requests that touch anything in the performance-metrics directory
+
+name: 'performance-metrics test & lint'
+
+on:
+  pull_request:
+    paths:
+      - 'performance-metrics/**'
+      - '.github/workflows/performance-metrics-test-lint.yaml'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  lint:
+    name: 'performance-metrics test & lint'
+    timeout-minutes: 5
+    runs-on: 'ubuntu-latest'
+    steps:
+      - name: Checkout opentrons repo
+        uses: 'actions/checkout@v4'
+
+      - name: Setup Python
+        uses: 'actions/setup-python@v5'
+        with:
+          python-version: '3.10'
+          cache: 'pipenv'
+          cache-dependency-path: performance-metrics/Pipfile.lock
+
+      - name: "Install Python deps"
+        uses: './.github/actions/python/setup'
+        with:
+          project: 'performance-metrics'
+
+      - name: Setup
+        id: install
+        working-directory: ./performance-metrics
+        run: make setup
+
+      - name: Test
+        if: always() && steps.install.outcome == 'success' || steps.install.outcome == 'skipped'
+        working-directory: ./performance-metrics
+        run: make test
+
+      - name: Lint
+        if: always() && steps.install.outcome == 'success' || steps.install.outcome == 'skipped'
+        working-directory: ./performance-metrics
+        run: make lint
--- a/performance-metrics/Makefile
+++ b/performance-metrics/Makefile
@@ -26,3 +26,7 @@ clean:
 wheel:
 	$(python) setup.py $(wheel_opts) bdist_wheel
 	rm -rf build
+
+.PHONY: test
+test:
+	$(pytest) tests
--- a/performance-metrics/Pipfile
+++ b/performance-metrics/Pipfile
@@ -5,15 +5,17 @@ name        = "pypi"
 
 [packages]
 opentrons-shared-data = {file = "../shared-data/python", editable = true}
+performance-metrics = {file = ".", editable = true}
 
 [dev-packages]
-pytest  = "==7.2.2"
+pytest  = "==7.4.4"
 mypy    = "==1.8.0"
 flake8 = "==7.0.0"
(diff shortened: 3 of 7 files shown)


api/src/opentrons/protocol_engine/commands/pick_up_tip.py
```diff
@@ -83,6 +84,8 @@ async def execute(self, params: PickUpTipParams) -> PickUpTipResult:
             well_name=well_name,
         )
 
+        raise TipNotAttachedError()
```
TamarZanzouri: This is to test it locally! no worries!

api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -804,12 +812,36 @@ def validate_action_allowed(  # noqa: C901
                 raise SetupCommandNotAllowedError(
                     "Setup commands are not allowed after run has started."
                 )
+            elif action.request.intent == CommandIntent.FIXIT:
+                if self._state.queue_status != QueueStatus.AWAITING_RECOVERY:
+                    raise FixitCommandNotAllowedError(
+                        "Fixit commands are not allowed when the run is not in a recoverable state."
+                    )
+                elif not self._state.failed_command or (
```
TamarZanzouri: should probably make sure its PROTOCOL intent?

sfoster1: The command linking is ultimately _for_ the client, so it's up to them IMO

robot-server/robot_server/runs/router/commands_router.py
```diff
@@ -137,6 +138,11 @@ async def get_current_run_engine_from_url(
         You can create a protocol purely over HTTP using protocol commands.
         If you are running a protocol from a file(s), then you will likely
         not need to enqueue protocol commands using this endpoint.
+        
+        Fixit commands may be enqueued anytime using this endpoint.
```
sfoster1: not true that it's "any time", right? only in `AWAITING_RECOVERY`?

robot-server/robot_server/runs/router/commands_router.py
```diff
@@ -137,6 +138,11 @@ async def get_current_run_engine_from_url(
...
+        These commands are intended to fix a failed command. 
+        They will be executed right after the failed command 
+        and only if the run is in a paused state.
```
sfoster1: `AWAITING_RECOVERY` for now, right?

api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -94,6 +95,12 @@ class BaseCommandCreate(GenericModel, Generic[CommandParamsT]):
             " If a value is not provided, one will be generated."
         ),
     )
+    failed_command_id: Optional[str] = Field(
```
sfoster1: Putting this into `BaseCommandCreate` means you need to regenerate the command schema. I wonder if it's better if, instead of putting this in command create, it's an optional addition to just the http endpoint to make it clear that this is for interactive and not canned-protocol use?

api/src/opentrons/protocol_engine/errors/exceptions.py
```diff
@@ -518,6 +531,19 @@ def __init__(
         super().__init__(ErrorCodes.GENERAL_ERROR, message, details, wrapping)
 
 
+class ResumeForRecoveryNotAllowedError(ProtocolEngineError):
```
sfoster1:
```suggestion
class ResumeFromRecoveryNotAllowedError(ProtocolEngineError):
```

?

api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -195,10 +195,11 @@ def add_command(self, request: commands.CommandCreate) -> commands.Command:
         )
 
         command_id = self._model_utils.generate_id()
-        request_hash = commands.hash_command_params(
-            create=request,
-            last_hash=self._state_store.commands.get_latest_command_hash(),
-        )
+        if request.intent != commands.CommandIntent.FIXIT:
```
sfoster1: I have two opinions about this.

Strong opinion: I think that rather than doing this here we should alter `protocol_engines.commands.hash_command_params.hash_command_params` to return `None` for `FIXIT` intents as it does for `SETUP` intents. Then we don't have to alter this code. 

Medium opinion: I think we need a clarity refactor that can mostly amount to renaming things here. `protocol_engine.state.commands.CommandState.latest_command_hash`  and `CommandState.get_latest_command_hash()` actually implicitly only store and retrieve the hash of the last-submitted `PROTOCOL` intent parameter, not the last-submitted general command. I think this is the correct behavior, but it's very confusing to read because the way it works is
- `latest_command_hash` and `get_latest_command_hash` are explicitly documented as holding "the latest hash value received in a `QueueCommandAction`
- but `latest_command_hash`, which is updated in the command state `handle_action`, is _only_ updated if the action's `request_hash` is not `None`
- the `request_hash` is generated from `hash_command_params`  as stated above, which will return `None` if the command's intent is `SETUP`
- so effectively `SETUP` commands are ignored in the hashing mechanism
This is as it should be but it's spread to a couple too many places and requires a little too much implicit behavior for my taste. I'd love to see us make this a bit more explicit by 
- renaming the command state cache and accessor to `latest_protocol_command_hash` and (or maybe only, if we consider these API) altering their docstrings appropriately
- moving the `None`-setting logic up to `add_command()`

TamarZanzouri: sure! thats probably better

api/src/opentrons/protocol_engine/commands/command.py
TamarZanzouri: that is not a bad idea. I will look into this.

PR #14904, opened 2024-04-15
refactor(robot-server): consolidate DB transactions, fix a max analyses length bug

sanni-t:
Closes AUTH-347

# Overview

#14885 added the feature to limit number of analyses we store in DB. In [this](https://github.com/Opentrons/opentrons/pull/14885#discussion_r1563058134) comment, @SyntaxColoring pointed out that we should consolidate the DB transactions for better performance, so that's what this PR does.

Also fixes a bug where if the existing number of analyses in the DB was 3 and we were to add another analysis, then the formula for getting the analysis IDs to delete would result in `analysis_ids[:-1]` and it would delete all analyses except last one.

# Test Plan

- Tested the cases mentioned in #14885 
- Tested the bug case

# Risk assessment

Low. Refactor + small bug fix

### DIFF
--- a/robot-server/robot_server/protocols/analysis_store.py
+++ b/robot-server/robot_server/protocols/analysis_store.py
@@ -196,7 +196,7 @@ class AnalysisStore:
                 completed_analysis
             ),
         )
-        await self._completed_store.add(
+        await self._completed_store.make_room_and_add(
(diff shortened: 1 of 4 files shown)


SyntaxColoring (approved):
TY!

jbleon95 (approved):
LGTM!

PR #14906, opened 2024-04-15
feat(app): push recent RTP run card click to protocol details

ncdiehl11:
# Overview

Because we don't yet have designs displaying details of a recent protocol run card with specific runtime parameter values, we want to allow the user to reset parameter values afer clicking a recent run card for an RTP-containing protocol. Behavior for non-RTP protocols will not change and will push the user directly to protocol run setup.

In the future, ideally, we can communicate RTP overrides on the recent run card and produce unique cards for different parameter selections on the same protocol ID. In this way, we can clone the run and avoid having to re-select parameters.

# Test Plan

- push this branch to Flex ODD
- send an RTP protocol and a non-RTP protocol to the Flex
- setup run both protocols to produce recent run cards on the robot dashboard recent run carousel
- click the non-RTP recent run card and verify that you are pushed directly to protocol run setup
- click the RTP recent run card and verify that you are pushed to protocol details to have the ability to reset parameter values

https://github.com/Opentrons/opentrons/assets/47604184/280e3095-f83e-4cc8-b944-5ad0cf81ebf0

# Changelog

- check for RTPs on recent run card
- on click, push to protocol details if RTPs exist
- add test for RTP recent run click

# Review requests

auth

# Risk assessment

medium

### DIFF
--- a/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
+++ b/app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
@@ -3,6 +3,7 @@ import { css } from 'styled-components'
 import { useTranslation } from 'react-i18next'
 import { useHistory } from 'react-router-dom'
 import { formatDistance } from 'date-fns'
+import { last } from 'lodash'
 
 import {
   BORDERS,
@@ -17,14 +18,14 @@ import {
   StyledText,
   TYPOGRAPHY,
 } from '@opentrons/components'
-import { useProtocolQuery } from '@opentrons/react-api-client'
+import {
+  useProtocolAnalysisAsDocumentQuery,
+  useProtocolQuery,
+} from '@opentrons/react-api-client'
 import {
   RUN_STATUS_FAILED,
   RUN_STATUS_STOPPED,
   RUN_STATUS_SUCCEEDED,
-  Run,
-  RunData,
-  RunStatus,
 } from '@opentrons/api-client'
 
 import { ODD_FOCUS_VISIBLE } from '../../../atoms/buttons//constants'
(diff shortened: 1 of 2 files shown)


app/src/organisms/OnDeviceDisplay/RobotDashboard/RecentRunProtocolCard.tsx
```diff
@@ -3,6 +3,7 @@ import { css } from 'styled-components'
 import { useTranslation } from 'react-i18next'
 import { useHistory } from 'react-router-dom'
 import { formatDistance } from 'date-fns'
+import { last } from 'lodash'
```
jerader: nit: importing this way is better so you don't import the whole lodash library

```suggestion
import last from 'lodash/last'
```

app/src/organisms/OnDeviceDisplay/RobotDashboard/__tests__/RecentRunProtocolCard.test.tsx
```diff
@@ -252,4 +281,14 @@ describe('RecentRunProtocolCard', () => {
     const [{ getByText }] = render(props)
     getByText('mock Skeleton')
   })
+
+  it.only('should push to protocol details if protocol contains runtime parameters', () => {
```
jerader:
```suggestion
  it('should push to protocol details if protocol contains runtime parameters', () => {
```

jerader (approved):
left a few minor comments but good job! lgtm!

app/src/organisms/OnDeviceDisplay/RobotDashboard/__tests__/RecentRunProtocolCard.test.tsx
ncdiehl11: 😓 my bad, thanks!

PR #14907, opened 2024-04-15
fix(app): properly disable proceed button if no available robots

ncdiehl11:
closes [RQA-2517](https://opentrons.atlassian.net/browse/RQA-2517)

# Overview

Fixes bug where ChooseRobotSlideout rendered by ChooseRobotToRunProtocolSlideout sets the selected robot to the first _connectable_ robot, even if there is no _available_ robot, when all connectable robots are busy (in runs). Checks for healthy reachable AND available robots when setting the selected robot on the slideout.

Leave current behavior for slideout when called by SendProtocolToFlexSlideout, since we can still send protocols to Flexes in active runs.

# Test Plan

- limit connectable robots to one or a few (one way is to disconnect from the network and only run dev robot server)
- begin protocol setup on all connectable robots
- navigate to `Protocols` tab and open the slideout by selecting "Start setup"
- confirm that no available robots options are shown and that the footer button is disabled
- on the protocol details page, select "Send to Opentrons Flex" from the overflow menu
- confirm that all connectable Flexes show, even if they are in runs

https://github.com/Opentrons/opentrons/assets/47604184/34a4f8e9-00e1-465c-932f-5e47fef788de

# Changelog

- if starting protocol setup, map AvailableRobotOption to only available robots
- fix test

# Review requests

@mjhuff per working on bug fix together

# Risk assessment

low

[RQA-2517]: https://opentrons.atlassian.net/browse/RQA-2517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/ChooseRobotSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotSlideout/index.tsx
@@ -184,10 +184,9 @@ export function ChooseRobotSlideout(
     {}
   )
 
-  const reducerAvailable = healthyReachableRobots.filter(
-    robot => !robotBusyStatusByName[robot.name]
+  const reducerAvailable = healthyReachableRobots.filter(robot =>
+    showIdleOnly ? !robotBusyStatusByName[robot.name] : robot
   )
-
   const reducerBusyCount = healthyReachableRobots.filter(
     robot => robotBusyStatusByName[robot.name]
   ).length
--- a/app/src/organisms/ChooseRobotSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotSlideout/index.tsx
@@ -79,7 +79,7 @@ interface RobotBusyStatusByName {
 
 export type RobotBusyStatusAction = RobotIsBusyAction | RobotIsIdleAction
 
-export function robotBusyStatusByNameReducer(
+function robotBusyStatusByNameReducer(
   state: RobotBusyStatusByName,
   action: RobotBusyStatusAction
 ): RobotBusyStatusByName {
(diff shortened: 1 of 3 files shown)


app/src/organisms/ChooseRobotSlideout/index.tsx
```diff
@@ -79,7 +79,7 @@ interface RobotBusyStatusByName {
 
 export type RobotBusyStatusAction = RobotIsBusyAction | RobotIsIdleAction
 
-function robotBusyStatusByNameReducer(
+export function robotBusyStatusByNameReducer(
```
mjhuff: Maybe I'm missing something, but do we need to export this?

app/src/organisms/ChooseRobotSlideout/index.tsx
```diff
@@ -184,15 +184,18 @@ export function ChooseRobotSlideout(
     {}
   )
 
+  const reducerAvailable = healthyReachableRobots.filter(robot =>
```
mjhuff:
```suggestion
  const reducerAvailableRobots = healthyReachableRobots.filter(robot =>
```
Would be a bit easier to tell what this does if we specify "robots"

mjhuff (approved):
Couple nits, but looks good with passing CI. Thank you for fixing this and adding tests!

app/src/organisms/ChooseRobotSlideout/index.tsx
ncdiehl11: absolutely not, I forgot to remove

PR #14910, opened 2024-04-15
fix(app): Handle Unsafe Move to Plunger during Drop-Tip

mjhuff:
Closes [EXEC-186](https://opentrons.atlassian.net/browse/EXEC-186)

# Overview

If the gantry is not homed and a powercycle occurs, drop-tip wizard cannot proceed with flows. An error is raised during the flow, and ultimately a home command is dispatched that has the side effect of potentially aspirating liquid into the pipette, damaging it. We need to special case home errors to prevent this. See ticket for designs.

The primary functional difference is now that any time an error occurs, exiting the wizard via the header should not home the gantry. Homing as a result of an error should only occur when the "Confirm removal and home" button is presented and clicked.

There's some minor refactor work done in this PR as well. More refactoring is definitely needed, but that should be handled in a separate PR (and take into account the wizard's use in the new error recovery flows). 

Note that I'm doing something a bit different for testing purposes, as a lot of our large wizard flows lack testing and are hard to appropriately approach with one big DOM driven integration test. Let's create some local utility functions and unit test those. Let me know what ya'll think.

### ODD
Screenshot 2024-04-15 at 1 42 25 PM

### Desktop
Screenshot 2024-04-15 at 1 00 33 PM

# Test Plan
- Do something that causes the gantry not to be homed. Power cycle the robot.
- Launch drop tip wizard, and proceed through the flow until the `MustHomeError` occurs.
- Confirm error recovery functionality works as intended based on designs.

# Changelog
- Fixed unsafe move plunger commands that may occur drop tip wizard. 

# Risk assessment
low

[EXEC-186]: https://opentrons.atlassian.net/browse/EXEC-186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/DropTipWizard/BeforeBeginning.tsx
```diff
@@ -57,16 +47,8 @@ export const BeforeBeginning = (
...
-    return (
-      <InProgressModal
-        description={
-          createdMaintenanceRunId == null
-            ? t('getting_ready')
-            : t('stand_back_exiting')
-        }
-      />
-    )
+  if (createdMaintenanceRunId == null) {
```
sfoster1, line 50: i was confused about this at first but looking at the later changes to index I'm a big fan

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -141,38 +153,44 @@ export function DropTipWizard(props: MaintenanceRunManagerProps): JSX.Element {
...
-        true
-      )
-        .then(() => {
-          deleteMaintenanceRun(maintenanceRunData?.data.id)
-        })
-        .catch(error => {
-          console.error(error.message)
-          deleteMaintenanceRun(maintenanceRunData?.data.id)
-        })
+      if (homeOnExit) {
```
sfoster1: nit: if we're going to live in promise-monad land let's just live fully in promise-monad land, i.e.
```
(homeOnExit 
   ? chainRunCommands(...) 
  : new Promise((resolve, reject) => resolve()))
.catch(error => console.error(error.message))
.finally(() => deleteMaintenanceRun(...)
```
rather than mixing if-statements and the promise chaining

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -326,41 +363,45 @@ export const DropTipWizardComponent = (
...
     } else {
-      setErrorMessage(`error moving to position: invalid addressable area.`)
+      setSpecificErrorDetails({
+        message: `Error moving to position: invalid addressable area.`,
+      })
       return Promise.resolve(null)
     }
   }
 
   let modalContent: JSX.Element = <div>UNASSIGNED STEP</div>
```
sfoster1: another great place for one of those let-switch to `(...) => JSX.Element` refactors...

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -537,3 +578,153 @@ export const DropTipWizardComponent = (
     getTopPortalEl()
   )
 }
+
```
sfoster1: imo if we're doing full interface object defs for hooks like this we should drop them in their own files

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -537,3 +578,153 @@ export const DropTipWizardComponent = (
...
+}: UseDropTipErrorComponentsProps): DropTipErrorComponents {
+  const genericSubHeader = (
+    <>
+      {t('drop_tip_failed')}
+      <br />
+      {errorDetails?.message}
+    </>
+  )
+
+  const result: DropTipErrorComponents = {
```
sfoster1: this is effectively a let-switch - could we maybe factor this into an arrow func for the complex case, letting us do something like a ternary return, e.g.
```
const buildHandleMustHome= () => { 
    const handleOnClick = ...
    ...
    return {
       button: isOnDevice ? ...,
       subHeader: <>{errorDetails.message}</>
    }
}

return errorDetails?.type === DROP_TIP_SPECIAL_ERROR_TYPES.MUST_HOME_ERROR
    ? buildHandleMustHome() 
    : {button: null, subHeader: (<>{t('drop_tip_failed')}<br />{errorDetails?.message}</>)}
```

I personally think it's a little clearer (continues to live fully in functional immutable data land)

app/src/organisms/DropTipWizard/index.tsx
```diff
@@ -537,3 +578,153 @@ export const DropTipWizardComponent = (
...
+ * most one time.
+ */
+export function useWizardExitHeader({
+  isFinalStep,
+  hasInitiatedExit,
+  errorDetails,
+  handleCleanUpAndClose,
+  confirmExit,
+}: UseWizardExitHeaderProps): () => void {
+  let handleExit: () => void = () => null
```
sfoster1: Well, to be fair, it's a very small function...

sfoster1 (approved):
I think this is great and obviously I'm very into the testing approach. I wonder if we could take it a bit further and drop these hooks into their own file though. Plus some nits

mjhuff:
@ecormany is OOO until Thursday. I'll merge this and just open a copy PR later if necessary.

PR #14915, opened 2024-04-16
fix(app): set max height for desktop modal shell

ncdiehl11:
closes [RQA-2279](https://opentrons.atlassian.net/browse/RQA-2279)

# Overview

We special-cased the size of the modal shell to accommodate a large body for the 96-channel pipette attachment BeforeBeginning step to avoid scrollable content. The size was set to 70% of the screen, but that resulted in a lot of modal whitespace when the app was resized to larger than default dimensions. Here, I add a max height of 31rem to accommodate the content of the largest pipette wizard flow tile content without growing arbitrarily large with a resized desktop app. I also ensure minimum padding between tile content and footer buttons, and fix padding on the ChoosePipette tile.

# Test Plan

- launch 96-channel attachment wizard flow
- resize desktop app to larger than default in the vertical direction
- observe that modal shell does not resize to larger than the content

I also smoketested various wizard flows that use LegacyModalShell.

Screenshot 2024-04-16 at 10 11 31 AM

# Changelog

- set special-cased LegacyModalShell height to constant rather than percentage to not grow with resized app
- add padding between body and bottom buttons of GenericWizardTile
- set explicit max height to 31rem for LegacyModalShell if not full screen
- fix container padding for ChoosePipette

# Review requests

frontend friends

# Risk assessment

low

[RQA-2279]: https://opentrons.atlassian.net/browse/RQA-2279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/molecules/GenericWizardTile/index.tsx
+++ b/app/src/molecules/GenericWizardTile/index.tsx
@@ -131,7 +131,11 @@ export function GenericWizardTile(props: GenericWizardTileProps): JSX.Element {
 
   return (
     <Flex css={TILE_CONTAINER_STYLE}>
-      <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing24}>
+      <Flex
+        flexDirection={DIRECTION_ROW}
+        gridGap={SPACING.spacing24}
+        paddingBottom={SPACING.spacing24}
+      >
         <Flex
           flexDirection={DIRECTION_COLUMN}
           flex="1"
--- a/app/src/molecules/GenericWizardTile/index.tsx
+++ b/app/src/molecules/GenericWizardTile/index.tsx
@@ -131,11 +131,7 @@ export function GenericWizardTile(props: GenericWizardTileProps): JSX.Element {
 
   return (
     <Flex css={TILE_CONTAINER_STYLE}>
-      <Flex
-        flexDirection={DIRECTION_ROW}
-        gridGap={SPACING.spacing24}
-        paddingBottom={SPACING.spacing24}
-      >
+      <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing24}>
         <Flex
           flexDirection={DIRECTION_COLUMN}
           flex="1"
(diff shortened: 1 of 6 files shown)


app/src/molecules/LegacyModal/LegacyModalShell.tsx
```diff
@@ -105,7 +105,7 @@ const ModalArea = styled.div<
 >`
   position: ${POSITION_RELATIVE};
   overflow-y: ${OVERFLOW_AUTO};
-  max-height: 100%;
+  max-height: ${({ isFullPage }) => (isFullPage ? '100%' : '31rem')};
```
jerader: doesn't this change affect every component that uses `LegacyModalShell` since the default is now `31rem` instead of `100%`?

mjhuff: Agree with @jerader. I think `LegacyModal` is due for a bit of a re-think, but I'm all for avoiding changing the base component if we can help it. If we are special-casing the attach flows due to the 96 graphic, let's pass in custom CSS.

app/src/molecules/GenericWizardTile/index.tsx
```diff
@@ -131,7 +131,11 @@ export function GenericWizardTile(props: GenericWizardTileProps): JSX.Element {
 
   return (
     <Flex css={TILE_CONTAINER_STYLE}>
-      <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing24}>
+      <Flex
+        flexDirection={DIRECTION_ROW}
+        gridGap={SPACING.spacing24}
+        paddingBottom={SPACING.spacing24}
```
mjhuff: See other comment. I don't think we want to change base components if we can help it.

ncdiehl11: Fair point. I will leave this alone.

app/src/molecules/LegacyModal/LegacyModalShell.tsx
ncdiehl11: For the time being, I'll leave alone the max height here til we can give some more thought to standardizing our modal styling.

mjhuff (commented):
Looking good! Would you mind fixing an unrelated copy error here: L35 of `pipette_wizard_flows.json` should have a capital A for `and Attach 96-Channel pipette`.

mjhuff (approved):
LGTM with passing CI. Thanks for fixing!

PR #14919, opened 2024-04-16
fix: fix timing function cross-platform bug

DerekMaggio:
# Overview

Closes https://opentrons.atlassian.net/browse/EXEC-406

Use a timing function universal to Windows, Mac, and Linux when running on Windows and Mac
This will allow performance metrics tests to run locally for devs
The production implementation will always use Linux

# Test Plan

- Create test that patches _get_timing_function to return universal timing function and make sure that works  
- Realized I missed timing synchronous functions so I added that too

# Changelog

- In robot_context_tracker.py create a _get_timing_function which imports the correct package based on os

# Review requests

None
# Risk assessment

low

### DIFF
--- a/performance-metrics/src/performance_metrics/robot_context_tracker.py
+++ b/performance-metrics/src/performance_metrics/robot_context_tracker.py
@@ -34,6 +34,7 @@ def _get_timing_function() -> Callable[[], int]:
 
     return time_function
 
+
 timing_function = _get_timing_function()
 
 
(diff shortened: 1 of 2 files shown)


performance-metrics/src/performance_metrics/robot_context_tracker.py
```diff
@@ -2,22 +2,39 @@
...
-    RawContextData,
-    RobotContextState,
-)
+from performance_metrics.datashapes import RawContextData, RobotContextState
 
 P = ParamSpec("P")
 R = TypeVar("R")
 
 
+def _get_timing_function() -> Callable[[], int]:
```
sfoster1, line 21: let's refactor this so that it only does the logic check once at import time rather than every time it's called

sfoster1 (changes requested):

sfoster1 (approved):
LGTM

PR #14922, opened 2024-04-16
chore(api): handle performance-metrics package not existing

DerekMaggio:
# Overview

Closes https://opentrons.atlassian.net/browse/EXEC-407

We do not want performance-metrics to be a required dependency for the publicly available opentrons package on PyPI. But we want to utilize performance-metrics when running the opentrons package on the robot itself.

The issue is, that since the performance-metrics package uses decorators to track running code, the decorators will always need to exist. This PR handles the case where the performance-metrics package does not exist and injects stubs where necessary. 

# Changelog

- Created the `SupportsTracking` mypy Protocol to define an interface both `RobotContextTracker` and the stubbed class, `StubbedTracker` implement
- Added performance-metrics as a dev dependency so tests using performance-metrics can still run 
- Created `performance_helpers.py`
  - Contains `StubbedTracker` defininition
  - Handles loading `StubberTracker` if performance-metrics library fails to load
  - Provides `_get_robot_context_tracker` private singleton function for eventual public-facing functions to use.

# Test Plan

- Testing to ensure stubbed `track` decorator returns the decorated function unchanged
- Validate singleton logic of _get_robot_context_tracker

# Risk assessment

Low, still not actually being used anywhere


api/src/opentrons/util/performance_helpers.py
```diff
@@ -0,0 +1,56 @@
...
+    """Handle the import of the performance_metrics package.
+
+    If the package is not available, return a stubbed tracker.
+    """
+    try:
+        from performance_metrics import RobotContextTracker
+
+        return RobotContextTracker
+    except ImportError:
+        return StubbedTracker
```
DerekMaggio, line 33: Should I check opentrons.config.IS_ROBOT here? 
```python
if IS_ROBOT:
    try:
       from performance_metrics import RobotContextTracker
    except 
        <raise or log here>
else:
    return StubbedTracker
```

sfoster1: nah I think this is totally fine

sfoster1 (approved):

PR #14926, opened 2024-04-16
fix(app): fix various install and version issues

sfoster1:
Fixes various ongoing issues building versions into the system. This is a follow-on to #14844 (61b137132a3130e8cea170c7cf3d4c1931735942)

- vite `define` config the way we do it does not hang the defined values off of props of `global` explicitly (or maybe us injecting `'globalThis'` into `global` breaks it) so use them as true globals, altering the way they're accessed and the way they're declared in the typings.
- i guess you don't actually have to do type imports in the top level typings? removing the type import of the ipc bridge in the app-shell and app-shell-odd global.d.ts fixed that issue. don't know why
- the ESM import for the script that updates the releases.json was wrong which breaks some update stuff

## Testing
This is a bit annoying to test.

You _must_ test this on a compiled app package. You _cannot_ test this on a dev build.
On a compiled app package,
- [x] the version should display in the settings tab of the app
- [x] you shouldn't have warnings about `include` on undefined in your app logs
- [ ] you should get robot update prompts when you use an internal-release build and connect to a robot running 1.3 or previous; you should get robot update prompts when you use a release build and connect to a robot running 7.2.1 or previous (note: couldn't test this in time but the rest of it works)
- [x] the help menu should have a bugs url that works (the "report an issue" button; it should pop a browser tab)
- [x] the help menu should say "View Opentrons App Logs" or "View Opentrons OT-3 App Logs" as the variant demands
   - [x] it should NOT say "View App Logs". that means the package name wasn't properly interpolated.

I haven't dev-tested the second part because on my home setup making a full compiled app package is broken for some reason, and you can't actually run the node side in dev
This once more,
Closes EXEC-385 
Closes RQA-2579

### DIFF
--- a/app-shell-odd/src/config/__fixtures__/index.ts
+++ b/app-shell-odd/src/config/__fixtures__/index.ts
@@ -11,11 +11,13 @@ import type {
   ConfigV21,
 } from '@opentrons/app/src/redux/config/types'
 
+const PKG_VERSION: string = _PKG_VERSION_
+
 export const MOCK_CONFIG_V12: ConfigV12 = {
   version: 12,
   devtools: false,
   reinstallDevtools: false,
-  update: { channel: _PKG_VERSION_.includes('beta') ? 'beta' : 'latest' },
+  update: { channel: PKG_VERSION.includes('beta') ? 'beta' : 'latest' },
   log: { level: { file: 'debug', console: 'info' } },
   ui: {
     width: 1024,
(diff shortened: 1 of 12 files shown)


mjhuff (approved):

sfoster1:
note - the tests were cancelled because i needed a build and the branch isn't named right. here's a build of a correctly-named version of the branch (with no code changes) in which the tests succeed https://github.com/Opentrons/opentrons/actions/runs/8713134684

PR #14928, opened 2024-04-16
feat(api): add analysis context tracking

DerekMaggio:
# Overview

Closes https://opentrons.atlassian.net/browse/EXEC-399

Evaluate robot context as `In Analysis` when _analyze is running in `opentrons.cli.analyze`

# Changelog

- Add `performance_metrics_data` directory 
- Inside of performance_helpers.py
    - Add logic to determine if tracking should be enabled (based off FF)
    - Add `track_analysis` factory decorator (I think that is the right name for it lol)
-  Wrap `opentrons.cli.analyze._analyze` with `track_analysis` decorator
- Change attribute naming on SupportsTracking to `storage_location`. If the passed Path wants to be a dir that needs to have a file defined or a direct file path, is an implementation detail.
- Have `RobotContextTracker` define its own file name

# Test Plan

- Verifying storage with tests in test_cli.py. A follow-up PR will implement storing to a file on the robot. At that point, testing will be done on the robot. 
- Did verify that `performance_metrics_data` directory is getting created

# Review requests

None

# Risk assessment

Low, although this is wrapping production code, it is gated by a feature flag that defaults to false

### DIFF
--- a/api/src/opentrons/cli/analyze.py
+++ b/api/src/opentrons/cli/analyze.py
@@ -28,6 +28,7 @@ from opentrons.protocol_engine import (
 )
 
 from opentrons_shared_data.robot.dev_types import RobotType
+from opentrons.util.performance_helpers import track_analysis
 
 
 @click.command()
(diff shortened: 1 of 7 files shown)


sfoster1 (approved):
lgtm!

PR #14929, opened 2024-04-16
fix(api): Various E-stop fixes

SyntaxColoring:
# Overview

Fixes various issues with `ProtocolEngine`'s handling of E-stops—some that go back to when the code was initially added, and some that I accidentally introduced while working on error recovery.

Closes EXEC-382.

# Test Plan

All of the following should behave reasonably.

* [x] Runs:
  * [x] Hitting the E-stop before anything has been done with the run, e.g. on the Opentrons App's run setup page.
  * [x] Hitting the E-stop in the middle of a hardware command.
  * [x] Hitting the E-stop in the middle of a non-hardware command (like `waitForResume`).
  * [x] Hitting the E-stop in between commands.
    * EXEC-382 has a protocol to help test this. 
* [x] Maintenance runs (like Labware Position Check):
  * [x] Hitting the E-stop before anything has been done with the maintenance run.
  * [x] Hitting the E-stop in the middle of a hardware command.
  * [x] Hitting the E-stop in the middle of a non-hardware command.
  * [x] Hitting the E-stop in between commands.

Given that we're not sure how much of this was working properly in v7.2, I think we can be flexible with our interpretation of "behaves reasonably." For example, it's now theoretically possible that the UI briefly shows the run as "stop requested," where it did not before. I think that's fine as long as nothing functionally bad happens, like the run hanging.

# Changelog

* `ProtocolEngine.estop()` was doing low-level state manipulation for individual commands. I accidentally broke this in #14726 when I made the command lifecycle stricter; I made it start raising `AssertionError`s.

  To fix this, `.estop()` is rewritten to do nothing more complicated than what `.stop()` does. This leaves the command state manipulation to `CommandStore`, where it canonically belongs and where it can be unit-tested easily. This is the change that fixes EXEC-382.

* `ProtocolEngine.estop()` was eagerly dispatching a `HardwareStoppedAction`. This is normally reserved for after the protocol has exited and the hardware has settled. But it was issuing it right away, basically lying to higher layers about the engine being "finished."

  This is fixed by removing all of that from `.estop()`, and relying on the existing teardown code in `.finish()` instead. This makes E-stop handling work basically the same way as other fatal run errors.

* We were naively calling `ProtocolEngine.estop()` from a hardware event callback. This did not look thread-safe to me: the hardware event callback presumably runs in the hardware API thread, and the `ProtocolEngine` lives in the main server thread. See [`ProtocolEngine`'s concurrency requirements](https://github.com/Opentrons/opentrons/blob/0e79a4086a1176b58b9bc728d32c5284fdb3d970/api/src/opentrons/protocol_engine/protocol_engine.py#L79-L84).

  This is fixed by wrapping the call in `loop.call_soon_threadsafe()`.

# Review requests

None in particular.

# Risk assessment

Medium. This can only be tested manually, and it's dealing with concurrency, and touching some confusing parts of our code.


sfoster1 (approved):
This looks great to me, as discussed

DerekMaggio:
Could any of this affect other flows that are part of the ProtocolEngine? Namely, LPC using PE, and the PE calibration command.
We should do a sanity check and push the e-stop during these flows to ensure it functions as expected.

SyntaxColoring:
> Could any of this affect other flows that are part of the ProtocolEngine? Namely, LPC using PE, and the PE calibration command. We should do a sanity check and push the e-stop during these flows to ensure it functions as expected.

It definitely can affect those, and we definitely should do that testing. I will test at least the things mentioned in the test plan. Feel free to add any more that you think of.

TamarZanzouri (approved):
Great work! TY!

SyntaxColoring:
~~I'm looking into issues where, after an E-stop, the run acts as if it was cancelled in the normal way, and tries to home.~~ Fixed, see the messages in the commits below.

PR #14932, opened 2024-04-17
fix(hardware): remove message ignored by filter spammy log

vegano1:


sfoster1 (approved):

sanni-t:
Thank you for the prompt fix!

PR #14940, opened 2024-04-17
feat(api): implement labware.set_offset in 2.18

sfoster1:
The python protocol api's labware.set_offset() command lets you set a labware position check offset programmatically. This is useful in protocols run outside the app (i.e. through Jupyter, through opentrons_execute).

When we moved to the protocol engine, we didn't support running those protocols outside the app, and therefore we didn't implement this method on those new API versions.

Now, we allow that again, so implement that method again.

This relies on dispatching AddLabwareOffset actions at arbitrary times, because it will set an offset for the location in which the labware is _currently_ present. This allows the user to set offsets without having to wrangle with the engine's internal definitions of where something is, at the cost of the user having to spread their offset calls around their protocol if they have protocols that load things in one place and later move them to another.

## Warning
~This won't work the way you think it will right now. Labware offsets are applied to labware instances by the `LoadLabware` and `MoveLabware` commands. So if you do an `AddLabwareOffsetAction` after a `LoadLabware`, the offset will never apply to the labware. We need to add a new `ReloadLabware` command, but I'm going to do that in a separate PR and then put this PR on top of that one.~
Fully implemented yay

Closes RSQ-29

## To Leave Draft
- [x] tests
- [x] general agreement that this is the right path
- [x] autogen doc updates (i.e. docstrings for `set_offset`
- [x] rebase on top of PR that adds `ReloadLabware`
- [x] also dispatch reload labware

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/labware.py
+++ b/api/src/opentrons/protocol_api/core/engine/labware.py
@@ -117,6 +117,9 @@ class LabwareCore(AbstractLabware[WellCore]):
             vector=LabwareOffsetVector(x=delta.x, y=delta.y, z=delta.z),
         )
         self._engine_client.add_labware_offset(request)
+        self._engine_client.reload_labware(
+            labware_id=self._labware_id,
+        )
 
     def get_calibrated_offset(self) -> Point:
         return self._engine_client.state.geometry.get_labware_position(self._labware_id)
--- a/api/src/opentrons/protocol_api/core/engine/labware.py
+++ b/api/src/opentrons/protocol_api/core/engine/labware.py
@@ -10,8 +10,17 @@ from opentrons_shared_data.labware.labware_definition import LabwareRole
 
 from opentrons.protocol_engine.errors import LabwareNotOnDeckError, ModuleNotOnDeckError
 from opentrons.protocol_engine.clients import SyncClient as ProtocolEngineClient
+from opentrons.protocol_engine.types import (
+    LabwareOffsetCreate,
+    LabwareOffsetLocation,
+    LabwareOffsetVector,
+    DeckSlotLocation,
+    ModuleLocation,
+    OnLabwareLocation,
+)
 from opentrons.types import DeckSlotName, Point
 from opentrons.hardware_control.nozzle_manager import NozzleMap
+from .exceptions import CannotSetOffsetAtThisLocationError
 
 from ..labware import AbstractLabware, LabwareLoadParams
 from .well import WellCore
@@ -91,10 +100,65 @@ class LabwareCore(AbstractLabware[WellCore]):
     def get_quirks(self) -> List[str]:
         return self._definition.parameters.quirks or []
 
-    def set_calibration(self, delta: Point) -> None:
-        raise NotImplementedError(
-            "Setting a labware's calibration after it's been loaded is not supported."
+    def _get_offset_location(self) -> LabwareOffsetLocation:
+        slot = self.get_deck_slot()
+        if None is slot:
+            raise CannotSetOffsetAtThisLocationError(
+                message=f"Cannot set offset for {self.get_name()} as it is not currently in a deck slot.",
+                detail={"kind": "labware-not-in-slot"},
+            )
+        parent_location = self._engine_client.state.labware.get_location(
+            self._labware_id
         )
+
+        if isinstance(parent_location, DeckSlotLocation):
+            return LabwareOffsetLocation(
+                slotName=slot, moduleModel=None, definitionUri=None
+            )
+        elif isinstance(parent_location, ModuleLocation):
+            module_model = self._engine_client.state.modules.get_requested_model(
+                parent_location.moduleId
+            )
+            return LabwareOffsetLocation(
+                slotName=slot, moduleModel=module_model, definitionUri=None
+            )
+        elif isinstance(parent_location, OnLabwareLocation):
+            non_labware_parent_location = (
+                self._engine_client.state.labware.get_parent_location(self._labware_id)
+            )
+            parent_uri = self._engine_client.state.labware.get_definition_uri(
+                parent_location.labwareId
+            )
+            if isinstance(non_labware_parent_location, DeckSlotLocation):
+                return LabwareOffsetLocation(
+                    slotName=slot, moduleModel=None, definitionUri=parent_uri
+                )
+            elif isinstance(non_labware_parent_location, ModuleLocation):
+                module_model = self._engine_client.state.modules.get_requested_model(
+                    non_labware_parent_location.moduleId
+                )
+                return LabwareOffsetLocation(
+                    slotName=slot, moduleModel=module_model, definitionUri=parent_uri
+                )
+
+            else:
+                raise CannotSetOffsetAtThisLocationError(
+                    message=f"Cannot set offset for {self.get_name()} as it is not currently in a deck slot.",
+                    detail={"kind": "labware-not-in-slot"},
+                )
+        else:
+            raise CannotSetOffsetAtThisLocationError(
+                message=f"Cannot set offset for {self.get_name()} as it is not currently in a deck slot.",
+                detail={"kind": "labware-not-in-slot"},
+            )
+
+    def set_calibration(self, delta: Point) -> None:
+        request = LabwareOffsetCreate.construct(
+            definitionUri=self.get_uri(),
+            location=self._get_offset_location(),
+            vector=LabwareOffsetVector(x=delta.x, y=delta.y, z=delta.z),
+        )
+        self._engine_client.add_labware_offset(request)
 
     def get_calibrated_offset(self) -> Point:
         return self._engine_client.state.geometry.get_labware_position(self._labware_id)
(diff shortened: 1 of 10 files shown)


DerekMaggio:
How are we going to communicate to users where/when they need to call `set_offset`? 
Seems to me that documenting that might be harder than documenting how to use the engines internal definitions of where something is.

Don't each one of the internal definitions of the locations correspond to, "an offset we calculate in LPC"? 

Can you maybe provide an example of what users would have to define if we went the other route?

sfoster1:
> How are we going to communicate to users where/when they need to call `set_offset`? Seems to me that documenting that might be harder than documenting how to use the engines internal definitions of where something is.
I really disagree with this. I think it's very documentable to say "use set_offset by calling it right after you move a labware"
> Don't each one of the internal definitions of the locations correspond to, "an offset we calculate in LPC"?
yes, but it's annoying (see below) to actually wrangle the various function calls, unions, isinstance checks, and type imports to do it correctly. 
> Can you maybe provide an example of what users would have to define if we went the other route?

if you went the other way, they'd have to do this: https://github.com/Opentrons/opentrons/pull/14940/files#diff-6f733f7119a96d280a77b7a2c54d99087fa07c7759299c4ce029c48ec0c1c668R103-R154 or we'd have to provide some function to look it up

api/src/opentrons/protocol_api/core/engine/labware.py
```diff
@@ -91,10 +100,65 @@ def get_parameters(self) -> LabwareParametersDict:
     def get_quirks(self) -> List[str]:
         return self._definition.parameters.quirks or []
 
+    def _get_offset_location(self) -> LabwareOffsetLocation:
+        slot = self.get_deck_slot()
+        if None is slot:
```
TamarZanzouri: nip but can we do this instead?
```suggestion
        if slot is None:
```

TamarZanzouri (commented):
LGTM! just a few comments

api/src/opentrons/protocol_api/core/engine/labware.py
```diff
@@ -91,10 +100,65 @@ def get_parameters(self) -> LabwareParametersDict:
     def get_quirks(self) -> List[str]:
         return self._definition.parameters.quirks or []
 
+    def _get_offset_location(self) -> LabwareOffsetLocation:
```
TamarZanzouri: I wonder if we have a method already to get parent location and then we can lose most of this logic.. maybe `get_labware_location`? its very similar to what you did

DerekMaggio:
@sfoster1, your response to my earlier comment helped me understand. 
I thought users would get confused about keeping track of where the labware was at. But they have to do that regardless. 

My mistake, you're correct, the downside of having to do all those imports and describing to the user how to use said imports, is much more difficult than just telling them to call set offset. 

This looks good to me so far

api/src/opentrons/protocol_api/core/engine/labware.py
SyntaxColoring: Ideally, I think we'd just make this easier at the `ProtocolEngine` level.

Right now, to set labware offsets, `ProtocolEngine` matches based on labware location and URI. But it doesn't fundamentally have to be that way. It could match based on labware ID, too.

SyntaxColoring (commented):
Cool, TY!

I think we'll need to document on the public `set_offset()` method that identical labware in the same location will share the same offset. This is the opposite of what the PAPIv≤2.13 behavior was, and what is currently documented.

api/src/opentrons/protocol_api/core/engine/labware.py
sfoster1: No, I think it fundamentally has to be that way, because labware offsets are on purpose distinguished by "what labware am I using and where am I using it". If you load two corning 96 plates, and you load an offset for a corning 96 plate in slot 5, and you have a protocol where one plate starts on slot 5 and then gets moved out and the other starts off deck and later gets moved in, they have the same offset. I don't think ID enters into it at all.

sfoster1:
> Cool, TY!
> 
> I think we'll need to document on the public `set_offset()` method that identical labware in the same location will share the same offset. This is the opposite of what the PAPIv≤2.13 behavior was, and what is currently documented.

Right, definitely - it's now adding a labware offset. Can do.

api/src/opentrons/protocol_api/core/engine/labware.py
SyntaxColoring: Yeah, if we take that underlying model for granted, I agree that this is as good as it gets.

`set_offset()` is a method on *a loaded `Labware` instance.* That strongly implies a conceptual model where it's one offset per instance, not one offset per location+URI.

You could view this as: The `set_offset()` interface is a relic from our older conceptual models of labware offsets. If we were designing it for our newer philosophy, it would look more like `ProtocolContext.add_offset(location, load_name, namespace, version)`. The mismatch between PAPI interface and underlying foundation will cause quirky behavior, and we'll have to accept that as long as we have a `Labware.set_offset()` method.

You could also view this as: The engine's conceptual model of labware offsets 1:1 matches the one used by the Opentrons App and by the HTTP API, only because that's what we originally built it for. It happens to not match the conceptual model implied by `Labware.set_offset()`, only because we didn't consider it at the time. Wanting a labware instance to have an offset is just as valid as wanting a URI+location to have an offset. The engine can support both if we let it.

I lean towards the second view. All models are wrong, but some are useful; matching by ID is more useful here, and, in my view, no more wrong than location+URI.

sfoster1: I lean toward the first view, and the reason is that _physically_, the thing that we want labware offsets to be is a patch over a slightly mistaken or misimplemented understanding of the labware geometry; the parent geometry; and their interaction. We touch all the labware, even multiple instances of the same definition, in LPC because they are in different places. We actively _want_ the behavior where if you have labwares A and B with the same definition, and A starts in C2 and later B moves to C2, the same offset applies in both places.

sfoster1:
@ecormany Would love your opinion on the documentation strategy here

sfoster1:
> Cool, TY!
> 
> I think we'll need to document on the public `set_offset()` method that identical labware in the same location will share the same offset. This is the opposite of what the PAPIv≤2.13 behavior was, and what is currently documented.

https://github.com/Opentrons/opentrons/pull/14967

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -1090,3 +1091,48 @@ def _labware_gripper_offsets(
...
+        (for instance, OFF_DECK) then return None.
+        """
+        parent_location = self._labware.get_location(labware_id)
+
+        if isinstance(parent_location, DeckSlotLocation):
+            return LabwareOffsetLocation(
+                slotName=parent_location.slotName, moduleModel=None, definitionUri=None
+            )
+        elif isinstance(parent_location, ModuleLocation):
+            module_model = self._modules.get_connected_model(parent_location.moduleId)
```
SyntaxColoring: I think this needs to be `get_requested_model()`, not `get_connected_model()`. See [this docstring](https://github.com/Opentrons/opentrons/blob/ee6e9fe94a64f5cb7c027faa5e014f0fc846bf6f/api/src/opentrons/protocol_engine/types.py#L556-L569).

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -1090,3 +1091,48 @@ def _labware_gripper_offsets(
...
+            if isinstance(non_labware_parent_location, DeckSlotLocation):
+                return LabwareOffsetLocation(
+                    slotName=non_labware_parent_location.slotName,
+                    moduleModel=None,
+                    definitionUri=parent_uri,
+                )
+            elif isinstance(non_labware_parent_location, ModuleLocation):
+                module_model = self._modules.get_connected_model(
+                    non_labware_parent_location.moduleId
+                )
```
SyntaxColoring: Ditto.

api/tests/opentrons/protocol_engine/state/test_geometry_view.py
```diff
@@ -99,54 +111,145 @@ def patch_mock_move_types(decoy: Decoy, monkeypatch: pytest.MonkeyPatch) -> None
...
+    By default, this will return a view with those dependencies as mocked. If you add a
+    parameter to your test of use_mocks that returns false, i.e.
+
+    @pytest.mark.parametrize('use_mocks', [False])
+    def my_cool_test(subject: GeometryView) -> None:
+        pass
+
+    then the provided subject will use actual states. Over time, we should get more and more
+    things using use_mocks=True, and then flip the default
+    """
```
SyntaxColoring, line 237: I'd be down to have `use_mocks` default to ~~`True`~~ [edit: oops, meant `False`] so it's harder to accidentally write new tests in the old way. Your call, though.

api/tests/opentrons/protocol_api/__init__.py
SyntaxColoring, line 1: Neat.

SyntaxColoring (commented):
One note about requested vs. connected module model, but this LGTM other than that. Thank you!

SyntaxColoring (approved):
Hooray

api/tests/opentrons/protocol_engine/state/test_geometry_view.py
sfoster1: I think i'm not going to because it's such a pain to change haha

PR #14942, opened 2024-04-17
fix(app): replace remark with react-markdown

mjhuff:
Closes [RQA-2587](https://opentrons.atlassian.net/browse/RQA-2587) and [RQA-2566](https://opentrons.atlassian.net/browse/RQA-2566)

# Overview

After the vite migration, the app would crash any time the release notes markdown file needed parsing, because remark decides to [invoke process.cwd() for some reason.](https://github.com/remarkjs/react-markdown/issues/189) Webpack can mock "process" out, but there's no 1:1 analogous solution with Vite. 

Although there's probably some hacky way do accomplish the same thing, remark-react has been deprecated for over three years, so that's more incentive to migrate. 

After testing a few options, react-markdown is relatively lightweight and properly overrides the markdown as desired (ie, no additional changes need to be made to the markdown for the refactor). It also doesn't have the same Vfile lib dependency that is the source of the "process" issue, so no more crashing when parsing release notes. It's also well-maintained, etc. 

Screenshot 2024-04-17 at 5 23 41 PM

# Test Plan

- Pull down the branch and build the app locally. You may fail key signing, and that's ok. It will still work.
- You should see robot update banners. Click on update and verify that the release notes appear instead of the app white screening.

# Changelog
- Updating the app or a robot no longer causes a whitescreen when the release sh

# Review requests
- Did I add this package correctly? Are there any other places I'm missing?

# Risk assessment
low

[RQA-2587]: https://opentrons.atlassian.net/browse/RQA-2587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RQA-2566]: https://opentrons.atlassian.net/browse/RQA-2566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Looks good and correct and not missing anything!

PR #14943, opened 2024-04-17
fix(hardware): remove can messenger `motor_enabled` listener

ahiuchingau:
# Overview
When #14479 was added, the motor enable message listener was never cleaned up properly and thus causing the robot to slow down significantly as it runs more and more move commands.


vegano1 (approved):
Nice,
The listener count was at 64 and incrementing before this change, now it's about 2-3 and not incrementing. The bot is also not super slow anymore.

sfoster1 (approved):
ah, nice

PR #14945, opened 2024-04-18
feat(shared-data, protocol-designer): return latest pipette model def f…

jerader:
…rom pipetteName and add max flow rates

closes AUTH-243 AUTH-245

# Overview

Sorry this should really be 2 PRs but realized I needed the shared-data changes in order to properly test the protocol-designer changes.

This PR does 2 things:
1. refactor the shared-data pipette v2 specs util to return the latest pipette model version definition that exists if it is given a pipette name. Spoke with Seth and Andy about this and they said that the first model version (`3.0`, `2.0`, etc) are prototypes so we shouldn't use them. 
2. wire up max flow rates in PD using the `uiMaxFlowRate` key

# Test Plan

- create a flex or ot-2 protocol. Make sure you choose a flex or GEN2 pipette. Add a transfer or mix step and click on the flow rate input field. It should render the flow rate modal. See that the max flow rate is properly wired up.
- review the util in shared-data and make sure the logic makes sense

# Changelog

- modify the flow rate field to use i18n strings and add a test for it and use the max flow rate from the pipette def key
- modify the logic in the pipette spec v2 util to return the latest pipette model version given a pipette name

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepEditForm/fields/FlowRateField/__tests__/FlowRateField.test.tsx
+++ b/protocol-designer/src/components/StepEditForm/fields/FlowRateField/__tests__/FlowRateField.test.tsx
@@ -54,9 +54,9 @@ describe('FlowRateField', () => {
   it('renders the flowRateInput and clicking on it opens the modal with all the text', () => {
     render(props)
     screen.getByText('Flow Rate')
-    fireEvent.click(screen.getByRole('textbox', { name: '' }))
+    fireEvent.click(screen.getByRole('textbox'))
     screen.getByText(
-      'Our default aspirate speed is optimal for a mockPipDisplayName aspirating liquids with a viscosity similar to water'
+      'The default mockPipDisplayName flow rate is optimal for handling aqueous liquids'
     )
     screen.getByText('aspirate speed')
     screen.getByText('160 μL/s (default)')
--- a/protocol-designer/src/components/StepEditForm/fields/FlowRateField/__tests__/FlowRateField.test.tsx
+++ b/protocol-designer/src/components/StepEditForm/fields/FlowRateField/__tests__/FlowRateField.test.tsx
@@ -1,0 +1,68 @@
+import * as React from 'react'
+import { describe, it, vi, beforeEach } from 'vitest'
+import { fireEvent, screen } from '@testing-library/react'
+import { fixtureP100096V2Specs } from '@opentrons/shared-data'
+import { renderWithProviders } from '../../../../../__testing-utils__'
+import { i18n } from '../../../../../localization'
+import { getPipetteEntities } from '../../../../../step-forms/selectors'
+import { FlowRateField } from '../index'
+
+vi.mock('../../../../../step-forms/selectors')
+const render = (props: React.ComponentProps<typeof FlowRateField>) => {
+  return renderWithProviders(<FlowRateField {...props} />, {
+    i18nInstance: i18n,
+  })[0]
+}
+const mockMockId = 'mockId'
+describe('FlowRateField', () => {
+  let props: React.ComponentProps<typeof FlowRateField>
+
+  beforeEach(() => {
+    props = {
+      disabled: false,
+      flowRateType: 'aspirate',
+      volume: 100,
+      value: null,
+      name: 'flowRate',
+      tiprack: 'tipRack:opentrons_flex_96_tiprack_1000ul',
+      updateValue: vi.fn(),
+      onFieldBlur: vi.fn(),
+      onFieldFocus: vi.fn(),
+      pipetteId: mockMockId,
+    }
+    vi.mocked(getPipetteEntities).mockReturnValue({
+      [mockMockId]: {
+        name: 'p50_single_flex',
+        spec: {
+          liquids: fixtureP100096V2Specs.liquids,
+          displayName: 'mockPipDisplayName',
+        } as any,
+        id: mockMockId,
+        tiprackLabwareDef: [
+          {
+            parameters: {
+              loadName: 'opentrons_flex_96_tiprack_1000ul',
+              tipLength: 1000,
+            },
+            metadata: { displayName: 'mockDisplayName' },
+          } as any,
+        ],
+        tiprackDefURI: ['mockDefURI1', 'mockDefURI2'],
+      },
+    })
+  })
+  it('renders the flowRateInput and clicking on it opens the modal with all the text', () => {
+    render(props)
+    screen.getByText('Flow Rate')
+    fireEvent.click(screen.getByRole('textbox', { name: '' }))
+    screen.getByText(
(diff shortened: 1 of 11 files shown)


protocol-designer/src/components/StepEditForm/fields/FlowRateField/__tests__/FlowRateField.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+          } as any,
+        ],
+        tiprackDefURI: ['mockDefURI1', 'mockDefURI2'],
+      },
+    })
+  })
+  it('renders the flowRateInput and clicking on it opens the modal with all the text', () => {
+    render(props)
+    screen.getByText('Flow Rate')
+    fireEvent.click(screen.getByRole('textbox', { name: '' }))
```
koji: I think we don't need to use name here.
```suggestion
    fireEvent.click(screen.getByRole('textbox'))
```

protocol-designer/src/components/StepEditForm/fields/FlowRateField/index.tsx
```diff
@@ -27,7 +27,7 @@ export function FlowRateField(props: OP): JSX.Element {
   const pipetteEntities = useSelector(stepFormSelectors.getPipetteEntities)
   const pipette = pipetteId != null ? pipetteEntities[pipetteId] : null
   const pipetteDisplayName = pipette ? pipette.spec.displayName : 'pipette'
-  const innerKey = `${name}:${String(value || 0)}`
+  const innerKey = `${props.name}:${String(value || 0)}`
```
koji:
```suggestion
 const {
    pipetteId,
    flowRateType,
    name,
    value,
    volume,
    tiprack,
    ...passThruProps
  } = props

  const innerKey = `${name}:${String(value || 0)}`
```

shared-data/js/pipettes.ts
```diff
@@ -139,24 +139,51 @@ const getChannelsFromString = (
...
+const getHighestVersion = (
+  wholeVersion: string,
+  path: string,
+  pipetteModel: string,
+  channels: Channels | null,
+  majorVersion: number,
+  highestVersion: string
+): string => {
+  const versionComponents = wholeVersion.split('_')
+  const majorPathVersion = parseInt(versionComponents[0])
```
koji, line 168: need to parseInt ?

I think the following would work with numeric char.
if(majorPathVersion === majorVersion)

shared-data/js/pipettes.ts
```diff
@@ -139,24 +139,51 @@ const getChannelsFromString = (
     }
   }
 }
-const getVersionFromGen = (gen: Gen): string | null => {
+const getVersionFromGen = (gen: Gen): number | null => {
```
koji: nit

maybe passing 0 for default and make return type number?
then we don't need to do null-check.

shared-data/js/pipettes.ts
```diff
@@ -174,13 +201,18 @@ export const getPipetteSpecsV2 = (
   const pipetteModel = nameSplit[0] // ex: p300
   const channels = getChannelsFromString(nameSplit[1] as PipChannelString) //  ex: single -> single_channel
   const gen = getVersionFromGen(nameSplit[2] as Gen)
```
koji: nit
gen is vague a little bit. maybe `pipetteGen`?

shared-data/js/pipettes.ts
```diff
@@ -219,8 +261,9 @@ export const getPipetteSpecsV2 = (
       liquidTypes.push(type)
     }
     if (
-      `../pipette/definitions/2/liquid/${channels}/${pipetteModel}/${type}/${version}.json` ===
-      path
+      `../pipette/definitions/2/liquid/${channels}/${pipetteModel}/${type}/${
```
koji, line 264: nit
maybe make a const for `../pipette/definitions/2/liquid/${channels}/${pipetteModel}`

then
filePath/${version === '' ? highestVersion : version}.json
filePath/${type}/${version === '' ? highestVersion : version}.json

koji:
Screenshot 2024-04-22 at 12 57 14 PM

not an issue but it's a little bit un-user-friendly since the default is 3.78 but we don't allow users to use 1 decimal place.

For the error message, `A max of 1 decimal places is allowed`

maybe `One decimal place is allowed` is enough?

@ecormany any thought on the above error message?

jerader:
> For the error message, `A max of 1 decimal places is allowed`
> 

@koji good catch, this is legacy behavior haha, we can see what @ecormany  says!

ecormany:
I don't think we should restrict the number of decimals entered, since the Python API doesn't. Any valid float between the mix and max for the pipette–tip combo should be fine. I just tested this on 2.17:

```
    p50.flow_rate.aspirate = 1.23456
    p50.aspirate(50, plate['A1'])
```
which simulates as `Aspirating 50.0 uL from A1 of NEST 96 Well Plate 200 µL Flat on slot A1 at 1.23456 uL/sec`

protocol-designer/src/localization/en/form.json
```diff
@@ -83,7 +83,13 @@
         "label": "delay"
       },
       "tip_position": { "label": "tip position" },
-      "flow_rate": { "label": "Flow Rate" },
+      "flow_rate": {
+        "label": "Flow Rate",
+        "range": "between {{min}} and {{max}}",
+        "error_decimals": "A max of {{decimals}} decimal places is allowed",
+        "error_out_of_bounds": "accepted range is {{min}} to {{max}}",
+        "default_text": "Our default aspirate speed is optimal for a {{displayName}} aspirating liquids with a viscosity similar to water"
```
ecormany: We should either 
- customize this text depending on whether the user is editing an aspirate or dispense rate
- make it generic and use the term "flow rate"

i think i prefer the latter. it's more consistent with PAPI docs. so, maybe something like this:
```suggestion
        "default_text": "The default {{displayName}} flow rate is optimal for handling aqueous liquids"
```

shared-data/js/pipettes.ts
jerader: hmm i kind of prefer having it inline like this instead of a constant. I think the constant makes it harder to look up immediately what the file path is since you'd have to navigate to the constant 🤔 i might leave as is. but thanks for the suggestion!

protocol-designer/src/localization/en/form.json
jerader: thanks Ed! i copied over the text that was there into i18n. This is a lot better!

koji:
@jerader 
https://github.com/Opentrons/opentrons/pull/14993

koji (approved):
There are a few e2e test errors.
But the changes look to me.

jerader:
> There are a few e2e test errors. But the changes look to me.

@koji oops good catch! i didn't notice

PR #14949, opened 2024-04-18
fix(app, api-client): fix choose protocol slideout issue

koji:
# Overview
Slideout only shows protocols that have the same robotType as well as the selected robot.
As a result, the issue Nus faced won't happen. 

I noticed that the data we use for RTP/non-RTP is wrong.
We used something like below but runtimeParameter is under mostRecentAnalisys and the following doesn't work as non-runtime parameter protocol.

I will update the test in a following pr. 

```
...dataFixture,
runtimeParameter: []
```

close RQA-2607

# Test Plan

# Changelog
- add filter to storedProtocol to hide unnecessary protocols(different from the selected robot's robotType)

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ChooseProtocolSlideout/__tests__/ChooseProtocolSlideout.test.tsx
+++ b/app/src/organisms/ChooseProtocolSlideout/__tests__/ChooseProtocolSlideout.test.tsx
@@ -7,7 +7,10 @@ import { renderWithProviders } from '../../../__testing-utils__'
 import { i18n } from '../../../i18n'
 import { getStoredProtocols } from '../../../redux/protocol-storage'
 import { mockConnectableRobot } from '../../../redux/discovery/__fixtures__'
-import { storedProtocolData as storedProtocolDataFixture } from '../../../redux/protocol-storage/__fixtures__'
+import {
+  storedProtocolData as storedProtocolDataFixture,
+  storedProtocolDataWithoutRunTimeParameters,
+} from '../../../redux/protocol-storage/__fixtures__'
 import { useTrackCreateProtocolRunEvent } from '../../../organisms/Devices/hooks'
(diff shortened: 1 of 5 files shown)


app/src/organisms/ChooseProtocolSlideout/index.tsx
```diff
@@ -492,11 +492,13 @@ function StoredProtocolList(props: StoredProtocolListProps): JSX.Element {
...
     runCreationError,
     runCreationErrorCode,
-    robotName,
+    robot,
   } = props
   const { t } = useTranslation(['device_details', 'protocol_details', 'shared'])
   const storedProtocols = useSelector((state: State) =>
     getStoredProtocols(state)
+  ).filter(
+    protocol => protocol.mostRecentAnalysis?.robotType === robot.robotModel
```
jerader, line 501: nice. I had to double check `robotModel` type 😅 , didn't realize `robotType` and `robotModel` are the same thing, very confusing

app/src/organisms/ChooseProtocolSlideout/__tests__/ChooseProtocolSlideout.test.tsx
```diff
@@ -86,34 +106,32 @@ describe('ChooseProtocolSlideout', () => {
...
-  //     name: 'Proceed to setup',
-  //   })
-  //   fireEvent.click(proceedButton)
-  //   expect(mockCreateRunFromProtocol).toHaveBeenCalledWith({
-  //     files: [expect.any(File)],
-  //     protocolKey: storedProtocolDataFixture.protocolKey,
-  //   })
-  //   expect(mockTrackCreateProtocolRunEvent).toHaveBeenCalled()
-  // })
+  it('calls createRunFromProtocolSource if CTA clicked', () => {
```
jerader, line 109: nice!

jerader (approved):
lgtm!

PR #14950, opened 2024-04-18
feat(app): add tiprack selection step to quick transfer flow

smb2268:
fix PLAT-290

# Overview
Implement the tip rack selection screen for the quick transfer wizard flow

# Test Plan
1. Launch quick transfer flow on ODD or ODD dev app
2. Select a pipette and proceed to step 3
3. You should see the pipette's default tip racks as options
4. The continue button should be disabled if no tiprack is selected
5. You should also see that if you go back to the selected pipette screen, the previously selected tiprack should be wiped
Screen Shot 2024-04-18 at 4 15 24 PM
Screen Shot 2024-04-18 at 4 15 16 PM

# Changelog
1. Added tip rack selection screen and test
2. Adjusted some styling to be in line with designs

# Review requests
Quick look through the changes - I've tested this out a bit

# Risk assessment
Low, behind FF


jerader (approved):
nice! lgtm! Will custom tipracks be supported in quick transfer?

smb2268:
> nice! lgtm! Will custom tipracks be supported in quick transfer?

@jerader MVP isn't going to support any custom labware but we may add that down the line!

PR #14951, opened 2024-04-18
refactor(protocol-designer): assign module slot in createFileWizard instead of modal

jerader:
closes AUTH-355 AUTH-22

# Overview

It was a bit wonky when the module slots were being assigned previously so i refactored a lot of the logic that i made a few weeks ago. Now, the slot is assigned at the end of the create file wizard (only for MoaM) instead of during the `module and other` modal of the wizard. singular module types still get assigned slots like before - so Ot-2 behavior for instance should be the same

# Test Plan

create a flex protocol and mess around with adding modules, including multiple temps, and see that they get rendered properly with nothing overlapping on the same slot

# Changelog

- refactor logic in `ModulesAndOtherTile` to now just disable when there is no slot available in terms of space on the deck
- make a redux thunk for creating modules with no slot defined
- create a few new utils and write tests
- wire up the new thunk in `CreateFileWizard`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
@@ -30,7 +30,6 @@ import {
   getModuleDisplayName,
   getModuleType,
   FLEX_ROBOT_TYPE,
-  THERMOCYCLER_MODULE_TYPE,
   MAGNETIC_BLOCK_TYPE,
 } from '@opentrons/shared-data'
 import { getIsCrashablePipetteSelected } from '../../../step-forms'
@@ -45,9 +44,8 @@ import { ModuleFields } from '../FilePipettesModal/ModuleFields'
 import { GoBack } from './GoBack'
 import {
   getCrashableModuleSelected,
-  getDisabledEquipment,
-  getNextAvailableModuleSlot,
-  getTrashBinOptionDisabled,
+  getIsSlotAvailable,
+  getTrashOptionDisabled,
 } from './utils'
 import { EquipmentOption } from './EquipmentOption'
 import { HandleEnter } from './HandleEnter'
@@ -197,10 +195,6 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
   const additionalEquipment = watch('additionalEquipment')
   const moduleTypesOnDeck =
     modules != null ? Object.values(modules).map(module => module.type) : []
-  const trashBinDisabled = getTrashBinOptionDisabled({
-    additionalEquipment,
-    modules,
-  })
 
   const handleSetEquipmentOption = (equipment: AdditionalEquipment): void => {
     if (additionalEquipment.includes(equipment)) {
@@ -209,6 +203,11 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
       setValue('additionalEquipment', [...additionalEquipment, equipment])
     }
   }
+  const trashBinDisabled = getTrashOptionDisabled({
+    additionalEquipment,
+    modules,
+    trashType: 'trashBin',
+  })
 
   React.useEffect(() => {
     if (trashBinDisabled) {
@@ -222,19 +221,8 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
         const moduleType = getModuleType(moduleModel)
         const moduleOnDeck = moduleTypesOnDeck.includes(moduleType)
 
-        let defaultSlot = getNextAvailableModuleSlot(
-          modules,
-          additionalEquipment
-        )
-        if (moduleType === THERMOCYCLER_MODULE_TYPE) {
-          defaultSlot = 'B1'
-        } else if (moduleType === MAGNETIC_BLOCK_TYPE) {
-          defaultSlot = 'D2'
-        }
-        const isDisabled = getDisabledEquipment({
-          additionalEquipment,
-          modules,
-        })?.includes(moduleType)
+        const isDisabled = !getIsSlotAvailable(modules, additionalEquipment)
+
         const handleMultiplesClick = (num: number): void => {
           const temperatureModules =
             modules != null
@@ -250,10 +238,7 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
                 [uuid()]: {
                   model: moduleModel,
                   type: moduleType,
-                  slot: getNextAvailableModuleSlot(
-                    modules,
-                    additionalEquipment
-                  ),
+                  slot: null,
                 },
               })
             }
@@ -290,7 +275,7 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
                 [uuid()]: {
                   model: moduleModel,
                   type: moduleType,
-                  slot: defaultSlot,
+                  slot: DEFAULT_SLOT_MAP[moduleModel],
                 },
               })
             }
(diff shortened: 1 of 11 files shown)


protocol-designer/src/components/modals/CreateFileWizard/ModulesAndOtherTile.tsx
```diff
@@ -222,19 +221,8 @@ function FlexModuleFields(props: WizardTileProps): JSX.Element {
         const moduleType = getModuleType(moduleModel)
         const moduleOnDeck = moduleTypesOnDeck.includes(moduleType)
```
koji: nit

`isModuleOnDeck` might be easier to understand its type.

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
```diff
@@ -229,9 +230,31 @@ export function CreateFileWizard(): JSX.Element | null {
...
+      // then modules without a slot are generated in remaining available slots
+      modules.sort((a, b) => {
+        if (a.slot == null && b.slot != null) {
+          return 1
+        }
+        if (b.slot == null && a.slot != null) {
+          return -1
+        }
+        return 0
+      })
```
koji, line 243: maybe

```suggestion
modules.sort((a, b) => {
  return (b.slot == null) - (a.slot == null);
});
```

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
```diff
@@ -229,9 +230,31 @@ export function CreateFileWizard(): JSX.Element | null {
...
+          return dispatch(stepFormActions.createModule(moduleArgs))
+        } else {
+          return dispatch(
+            createModuleWithNoSlot({
+              model: moduleArgs.model,
+              type: moduleArgs.type,
+            })
+          )
+        }
+      })
```
koji: nit

```suggestion
      modules.forEach(moduleArgs => {
       return moduleArgs.slot != null ? dispatch(stepFormActions.createModule(moduleArgs))
       : dispatch(
            createModuleWithNoSlot({
              model: moduleArgs.model,
              type: moduleArgs.type,
            })
          )
      })
```

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -120,102 +83,59 @@ export const getUnoccupiedStagingAreaSlots = (
...
-  )
+): boolean => {
+  const moduleLength = modules != null ? Object.keys(modules).length : 0
+  const additionalEquipmentLength = additionalEquipment.length
   const hasTC =
     modules != null
       ? Object.values(modules).some(
           module => module.type === THERMOCYCLER_MODULE_TYPE
         )
       : false
```

```suggestion
const hasTC = Object.values(modules || {}).some(module => module.type === THERMOCYCLER_MODULE_TYPE);
```

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -120,102 +83,59 @@ export const getUnoccupiedStagingAreaSlots = (
...
-    removeSlotForTrash = MODULES_SLOTS_FLEX.slice(0, -1)
-  }
-  const unoccupiedSlot = removeSlotForTrash.find(
-    cutout =>
-      !stagingAreaCutouts.includes(cutout.value) &&
-      !moduleSlots.includes(cutout.slot) &&
-      !wasteChuteSlot.includes(cutout.value)
+  const filteredModuleLength = hasTC ? moduleLength + 1 : moduleLength
+  const hasWasteChute = additionalEquipment.find(equipment =>
+    equipment.includes('wasteChute')
```
koji: maybe? (assuming boolean?)

```suggestion
  const hasWasteChute = additionalEquipment.some(equipment =>
    equipment.includes('wasteChute')
```

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -120,102 +83,59 @@ export const getUnoccupiedStagingAreaSlots = (
...
   )
-  if (unoccupiedSlot == null) {
-    return ''
+  const isStagingAreaInD3 = additionalEquipment
+    .filter(equipment => equipment.includes('stagingArea'))
+    .find(stagingArea => stagingArea.split('_')[1] === 'cutoutD3')
+  const hasGripper = additionalEquipment.find(equipment =>
```
koji: maybe

```suggestion
  const hasGripper = additionalEquipment.some(equipment =>
```

protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -120,102 +83,59 @@ export const getUnoccupiedStagingAreaSlots = (
...
-  const nextAvailableSlot = getNextAvailableModuleSlot(
-    modules,
-    additionalEquipment
-  )
-  const hasTrashBinAlready = additionalEquipment.includes('trashBin')
-  return nextAvailableSlot === '' && !hasTrashBinAlready
+  const { additionalEquipment, modules, trashType } = props
+  const isSlotAvailable = getIsSlotAvailable(modules, additionalEquipment)
+  const hasTrashBinAlready = additionalEquipment.includes(trashType)
+  return !isSlotAvailable && !hasTrashBinAlready
```

maybe just `!getIsSlotAvailable() && !additionalEquipment.includes()` since both of them are very simple.

protocol-designer/src/components/modules/MultipleModulesRow.tsx
```diff
@@ -25,7 +25,9 @@ interface MultipleModulesRowProps {
   moduleOnDeck?: ModuleOnDeck[]
 }
 
-export function MultipleModuleRow(props: MultipleModulesRowProps): JSX.Element {
+export function MultipleModulesRow(
```
koji, line 28: 👍

protocol-designer/src/modules/__tests__/moduleData.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+        trashBinId: {
+          name: 'trashBin',
+          id: 'trashBinId',
+          location: 'D1',
+        },
+      },
+    }
+    const result = getNextAvailableModuleSlot(mockInitialDeckSetup)
+    expect(result).toBe('C1')
+  })
```
koji, line 36: can you add a case that returns `undefined`?

protocol-designer/src/modules/moduleData.ts
```diff
@@ -7,8 +8,17 @@ import {
   ModuleType,
   MAGNETIC_BLOCK_TYPE,
   RobotType,
+  CutoutId,
+  MOVABLE_TRASH_ADDRESSABLE_AREAS,
+  WASTE_CHUTE_ADDRESSABLE_AREAS,
+  FIXED_TRASH_ID,
+  AddressableAreaName,
 } from '@opentrons/shared-data'
```
koji: nit

```ts
import {
  MAGNETIC_MODULE_TYPE,
  TEMPERATURE_MODULE_TYPE,
  THERMOCYCLER_MODULE_TYPE,
  HEATERSHAKER_MODULE_TYPE,
  MAGNETIC_BLOCK_TYPE,
  MOVABLE_TRASH_ADDRESSABLE_AREAS,
  WASTE_CHUTE_ADDRESSABLE_AREAS,
  FIXED_TRASH_ID,
import { getStagingAreaAddressableAreas } from '../utils'
import { getSlotIsEmpty } from '../step-forms'
import type { DropdownOption } from '@opentrons/components'
import type {
  ModuleType,
  RobotType,
  CutoutId,
  AddressableAreaName,
import type { InitialDeckSetup } from '../step-forms'
import type { DeckSlot } from '../types'
```

protocol-designer/src/components/modals/CreateFileWizard/index.tsx
jerader: hmm i get a bunch of ts errors that way. i think i'll keep it as is 🤔 thanks for the suggestion though!

koji: interesting

koji (approved):
the changes look good to me and the sandbox worked as expected.

PR #14952, opened 2024-04-18
refactor(robot-server, api): refactor ChangeNotifier to a message-capable async generator

mjhuff:
Closes [EXEC-404](https://opentrons.atlassian.net/browse/EXEC-404)

# Overview

ChangeNotifier is now EventNotifier and can support passing messages to callbacks via notify(). It's also an async generator, supporting communication between two threads. Although this doesn't change anything currently, we'll need an async generator for handling notes and mqtt notifications from hardware control to robot-server/PE in the future. 

Note that I refactored the PE callback to be an EventNotifier.notify(). I think keeping things simple with as few notification interfaces is a plus, but if others feel that it's not a great idea to use an async generator here, I'm happy to bring back ChangeNotifier.

I've renamed a few functions in preparation for `hardware` equivalents, since we'll be bubbling from more than just PE in the near future.

# Test Plan

- Run a protocol while connected to the MQTT server on Postman. Verify that /runs events are published, demonstrating that EventNotifier successfully invokes the robot-server callback within PE. 

# Risk assessment
low - ultimately not too much is different here

[EXEC-404]: https://opentrons.atlassian.net/browse/EXEC-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,147 @@
...
+
+        self._initialize()
+
+    def notify(self, message: Any = None) -> None:
+        """Notify all `subscribers` of a change."""
+        if self._task is None:
+            raise NotInitializedError()
+
+        try:
+            self._queue.put_nowait(message)
```
sfoster1, line 76: this seems like a mistake, or at least it seems like a mistake to _only_ have this interface. maybe call it `notify_lossy()` and also have an `async def notify_reliable()`?

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,147 @@
...
+    Optional,
+    Generator,
+    ContextManager,
+    Any,
+)
+
+_CallbackT = Callable[..., Awaitable[None]]
+
+
+class ReadOnlyEventNotifier(ABC):
```
sfoster1: We should make this interface and class type safe by making them `Generic` across a typevar that can contain a union of possible messages, ideally a heterogenous struct.

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,147 @@
...
+    """The read-only subset of `EventNotifier`.
+
+    Only allow registering callbacks for an instantiated `EventNotifier`.
+    """
+
+    @abstractmethod
+    def subscribed(self, callback: _CallbackT) -> ContextManager[None]:
```
sfoster1: this should also have an `@contextmanager` decorator

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,147 @@
...
+        def unsubscribe() -> None:
+            self._callbacks.remove(callback)
+
+        return unsubscribe
+
+    def subscribe_many(self, callbacks: List[_CallbackT]) -> Callable[[], None]:
+        """Register a list of callbacks to be called on each notify().
+
+        Returns:
+            A function that you can call to unsubscribe ``callback``.
```
sfoster1:
```suggestion
            A function that you can call to unsubscribe all ``callbacks``.
```
?

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,147 @@
...
+
+    async def _wait(self) -> AsyncIterable[Optional[Any]]:
+        """Wait until the next change notification."""
+        while True:
+            try:
+                yield await self._queue.get()
+            except asyncio.CancelledError:
+                raise StopAsyncIteration
+
+    async def _wait_for_event(self) -> None:
```
sfoster1, line 160: these are private? what uses them?

sfoster1 (changes requested):
I like the structure, but
- We should have the await-a-message interface be public
- Right now this erases all the message types; we should use generic types to make it type-safe.

api/src/opentrons/util/event_notifier.py
mjhuff: Ah yep, definitely a mistake to only have this interface. Thanks for the catch.

api/src/opentrons/util/event_notifier.py
mjhuff: `_wait` is utilized by `_wait_for_event`, which is utilized by `_initialize`.  I think we want these methods to be private, since we want callbacks to be the way in which clients interact with messages/events.

api/src/opentrons/util/event_notifier.py
```diff
@@ -9,36 +9,45 @@
...
-    ContextManager,
-    Any,
+    TypeVar,
+    Union,
+    Generic,
 )
 
-_CallbackT = Callable[..., Awaitable[None]]
+_MessageT = TypeVar("_MessageT")
+_CallbackT = Union[
```
mjhuff, line 18: Is there an alternative way of typing `_CallbackT` in which I don't have to include `# type: ignore` in some spots in this file and more annoyingly, in `publisher_notifier.py`?

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,174 @@
...
+        """See `Broker.subscribe_many()`."""
+        pass
+
+
+class EventNotifier(Generic[_MessageT], ReadOnlyEventNotifier[_MessageT]):
+    """Asynchronously manages a queue of events and notifying subscribed callbacks when a change occurs.
+
+    It is safe to use an `EventNotifier` for communication between two threads if:
+     Thread #1 acts as the sole notifier via notify().
+     Thread #2 acts as the sole callback registrar via subscribe() or subscribe_many().
```
SyntaxColoring, line 55: I don't believe this is actually thread-safe, is it? The central mechanism here is an `asyncio.Queue`, and, from https://docs.python.org/3/library/asyncio-queue.html:

> asyncio queues are designed to be similar to classes of the [queue](https://docs.python.org/3/library/queue.html#module-queue) module. Although asyncio queues are not thread-safe, they are designed to be used specifically in async/await code.

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,174 @@
...
+                self._callbacks.remove(callback)
+
+        return unsubscribe
+
+    def _initialize(self) -> None:
+        """Creates an asyncio task that indefinitely waits for an event to be notified.
+
+        The task will continue running until it is explicitly cancelled or the program exits.
+        """
+        self._task = asyncio.create_task(self._wait_for_event())
```
SyntaxColoring, line 150: Nothing ever awaits `self._task`, as far as I can tell? If that's true, this is a kind of resource leak. It may or may not matter for this specific code, but as a general practice, we want to ensure every `create_task()` is paired with an `await`, for reasons including:

* Exceptions in the background task will propagate somewhere.
* Any resources that someone opens inside the task will be closed promptly. (I don't think you're opening any right now, but future refactors might.)
* Any side-effects that the background task does will be guaranteed to have stopped by some deterministic time.

The avoidance of orphaned tasks like this is an idea of "structured concurrency." This is a best-practice that the Python community discovered *after* the introduction of `asyncio`, so the `asyncio` API doesn't always obey it by default.

I think [this blog post](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/#go-statement-considered-harmful) popularized the idea of spawning tasks from a `with` block. It's a worthwhile read. [Trio](https://trio.readthedocs.io/en/stable/) [implemented it](https://trio.readthedocs.io/en/stable/reference-core.html#nurseries-and-spawning), and then [AnyIO](https://anyio.readthedocs.io/en/stable/) also [implemented it](https://anyio.readthedocs.io/en/stable/tasks.html#creating-and-managing-tasks), and now Python 3.11 [finally has it in the standard library](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup).

api/src/opentrons/util/event_notifier.py
```diff
@@ -0,0 +1,174 @@
...
+
+    async def _wait(self) -> AsyncIterable[Optional[_MessageT]]:
+        """Wait until the next change notification."""
+        while True:
+            try:
+                yield await self._queue.get()
+            except asyncio.CancelledError:
+                raise StopAsyncIteration
```
SyntaxColoring, line 158: Here, [and as a general rule](https://docs.python.org/3/library/asyncio-task.html#:~:text=In%20case%20asyncio.CancelledError%20is%20explicitly%20caught%2C%20it%20should%20generally%20be%20propagated%20when%20clean%2Dup%20is%20complete.), I don't think we want to swallow `CancelledError`s like this.

Suppose you have some code like this:

```python
await do_stuff_before_events()
async for event in get_events():
    await do_stuff_with_event(event)
await do_stuff_after_events()
```

Then suppose you cancel this code while it's waiting for a new `event`, on the `async for event in events` line.

If the `get_events()` implementation replaces `CancelledError` with `StopAsyncIteration`, it will make our code exit the `for` loop, but continue on to `do_stuff_after_events()`, which we do not want. We want the entire code block to be cancelled, and only execute additional things if they were in something like a `finally` block.

api/src/opentrons/util/event_notifier.py
SyntaxColoring: I'm a bit confused about the motivation here, then. This is on me for misinterpreting, but when talking about "refactoring the notifier to be an async generator," I thought that was intended to support an API roughly like this:

```python
event_notifier = ...
with event_notifier.listen() as events:  # .listen() returns an async generator
    async for event in events:
        print("Received event", event)
```

But it sounds here like the async generator is just meant to be a private implementation detail?

If we aren't pursuing any API changes, I think we should...not make this change? The problem is that if `ChangeNotifier` has an internal async generator, then it needs exactly one `asyncio.Task` to pull from it in the background, and then managing the lifetime of that task adds complexity of its own (as described in my other comment).

api/src/opentrons/util/event_notifier.py
SyntaxColoring: So one way of fixing this would be to turn this class into a context manager, and have its `__exit__()` method do the `await self._task()`.

Another way—and I think this is better—is to avoid the background task completely. I can imagine a couple ways of doing this, but I think I need to better understand the introduction of the async generator—see my other comment.

api/src/opentrons/util/event_notifier.py
mjhuff: Thanks for taking the time for the thoughtful feedback and the links! I'll work through this slowly in between other tickets and reading, so bear with me. 

To this point - yes, this queue isn't inherently thread-safe, but if one queue thread is only appending to the queue while the other dequeues, to the best of my knowledge (and please correct me if I'm wrong), this is never a potential threading issue, hence the docstring.

api/src/opentrons/util/event_notifier.py
mjhuff: Ah yep, this makes total sense, thanks. I should have read the docs here.

api/src/opentrons/util/event_notifier.py
SyntaxColoring: Of course, no rush. We can chat about it in-person, too.

> To this point - yes, this queue isn't inherently thread-safe, but if one queue thread is only appending to the queue while the other dequeues, to the best of my knowledge, this is never a potential thread issue, hence the docstring.

Unfortunately no—"not thread-safe" means it's unsafe to do *anything with the object* from concurrent threads. Not merely that it's unsafe to call `put()` or `get()` from concurrent threads. Which makes sense, since `put()` and `get()` must be implemented internally by sharing mutable state.

[Here's a demo](https://gist.github.com/SyntaxColoring/2c86b70f1c8755b2a40e2f614142e1c5). On my machine, it hangs ~50% of the time: the consuming thread does not see the insertions from the producing thread. You might need to fiddle with the constants at the top to reproduce that.

mjhuff: Ah yeah, that does make sense, thanks for the demo. 

Alright, I'll have a think about this more. I see what you and Seth mean about exposing the async generator now.

mjhuff:
This is due for some more thinking, and I'll circle back to it once higher priority work quiets down. Per discussion with @SyntaxColoring , the async generator needs to be exposed publicly, and the queue needs to be at least thread safe if that is the correct route.

PR #14955, opened 2024-04-19
fix(api): engage axis to enable the motor before attempting to move the axis.

vegano1:
# Overview

There seems to be some issue on the firmware where the motor is enabled but does not seem to get enabled, causing the axis we are attempting to move to throw a `finished movement with a condition not met` error. The firmware should be enabling the motor, and if we query the motor enable status with get_status_request = 0x01 we get the correct response where the motor is enabled. For whatever reason sending an explicit enable_motor_request = 0x06 before moving the axis fixes the problem.

Closes: [RQA-2591](https://opentrons.atlassian.net/browse/RQA-2591)

# Test Plan

- [x] Make sure that we can attach/detatch single-channel pipettes
- [x] Make sure that we can attach/detatch multi-channel pipettes
- [x] Make sure that we can attach/detatch the 96-channel pipette
- [x] Make sure that we can attach/detatch the gripper
- [x] Make sure we can calibrate all instruments without failing on `movement not met` once the flow is done.

# Changelog

- explicitly enable the motor before attempting to move an axis during `home` and `retract_axis`

# Review requests

- Please test this with all the permutations of pipettes attached/detached/etc

# Risk assessment

Med, needs testing so we understand all the implications.

[RQA-2591]: https://opentrons.atlassian.net/browse/RQA-2591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Okay, glad we can work around it.

PR #14959, opened 2024-04-19
fix(app): clone run with RTPs from HistoricalProtocolRun

ncdiehl11:
closes [RQA-2601](https://opentrons.atlassian.net/browse/RQA-2601)

# Overview

Here, I extend the useCloneRun hook to optionally create a new analysis for a protocol with RTP
override values that will be referenced at protocol setup. This provides functionality for cloning a
HistoricalProtocolRun with its RTP values preserved.

# Test Plan

1. Run an RTP protocol with any parameter set
2. Rerun that same protocol with different parameters on the same robot
3. Navigate to device details run history
4. Click the overflow menu > "Rerun protocol now"
5. Verify that the Parameters tab shows the RTP for the selected run and not the most recent run
6. Repeat 4 and 5, but clicking the overflow menu > "View protocol run record"
7. Repeat 4 and 5, but clicking the date of the run record rather than "Rerun now" 

# Changelog

- update `useCloneRun` to optionally trigger a new protocol analysis, properly invalidate queries to `/runs` and `/protocols`
- use run record rather than most recent analysis able on `Parameters` and `RunPreview`
- add and export constant for terminal run statuses
- fix `runTimeParameters` key on `Run` type
- add ProtocolRun > Parameters `InfoScreen` for run canceled before beginning

# Review requests

auth, @shlokamin per pairing

# Risk assessment

medium

[RQA-2601]: https://opentrons.atlassian.net/browse/RQA-2601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mjhuff (commented):
I like the new terminal constant!

I'd probably spend some time grepping for the terminal run statuses in the app individually, because there are a few places we use them with slightly different names. Ex, in `ProtocolRunHeader` we have `RUN_OVER_STATUSES`.  These should be refactored over.

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
```diff
@@ -34,7 +40,15 @@ export function ProtocolRunRuntimeParameters({
...
-  const runTimeParameters = mostRecentAnalysis?.runTimeParameters ?? []
+  const runStatus = useRunStatus(runId)
+  const isRunTerminal =
+    // @ts-expect-error: expected that runStatus may not be a terminal status
+    runStatus == null ? false : RUN_STATUSES_TERMINAL.includes(runStatus)
+  const run = useNotifyRunQuery(runId).data
+  const runTimeParameters =
+    (isRunTerminal
+      ? run?.data?.runTimeParameters
+      : mostRecentAnalysis?.runTimeParameters) ?? []
```
shlokamin, line 57: i'd add a little comment above here explaining why we're doing this — i can see myself forgetting a year from now 😛

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
```diff
@@ -79,7 +93,11 @@ export function ProtocolRunRuntimeParameters({
       </Flex>
       {!hasParameter ? (
         <Flex padding={SPACING.spacing16}>
-          <InfoScreen contentType="parameters" />
+          <InfoScreen
+            contentType={
+              runStatus === RUN_STATUS_STOPPED ? 'runNotStarted' : 'parameters'
```
shlokamin: this isn't quite right — a run could be stopped half way through a run. i think what we want to check here is whether a play action has been dispatched. so we'd do something like this:


```js
  const runActions = runRecord?.data.actions
  const hasRunStarted = runActions?.some(
    action => action.actionType === RUN_ACTION_TYPE_PLAY
  )
```

app/src/organisms/RunPreview/index.tsx
```diff
@@ -41,6 +45,13 @@ export const RunPreviewComponent = (
 ): JSX.Element | null => {
   const { t } = useTranslation('run_details')
   const robotSideAnalysis = useMostRecentCompletedAnalysis(runId)
+  const runStatus = useRunStatus(runId)
+  // @ts-expect-error: expected that runStatus may not be a terminal status
+  const isRunTerminal = RUN_STATUSES_TERMINAL.includes(runStatus)
+  const commandsFromQuery = useAllCommandsQuery(runId, null, {
+    staleTime: Infinity,
+    cacheTime: Infinity,
+  }).data?.data
```
shlokamin: id add a comment here explaining that we're doing this because we only ever want one request done because its so heavy. i'd also disable this query whenever the run is NOT terminal so we don't unnecessarily take the performance hit

shlokamin (commented):
looks good to me, just a few small things

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
ncdiehl11: That makes more sense— thanks for the catch

app/src/organisms/RunPreview/index.tsx
ncdiehl11: makes sense to me as well

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
```diff
@@ -25,6 +30,8 @@ import { Tooltip } from '../../../atoms/Tooltip'
 import { useMostRecentCompletedAnalysis } from '../../LabwarePositionCheck/useMostRecentCompletedAnalysis'
 
 import type { RunTimeParameter } from '@opentrons/shared-data'
+import { useRunStatus } from '../../RunTimeControl/hooks'
+import { useNotifyRunQuery } from '../../../resources/runs'
```
jerader, line 32: these imports can go a bit higher

app/src/organisms/Devices/ProtocolRun/__tests__/ProtocolRunRuntimeParameters.test.tsx
```diff
@@ -7,13 +7,18 @@ import { InfoScreen } from '@opentrons/components'
...
+import { mockSucceededRun } from '../../../RunTimeControl/__fixtures__'
 
 import { ProtocolRunRuntimeParameters } from '../ProtocolRunRunTimeParameters'
 
 import type {
   CompletedProtocolAnalysis,
   RunTimeParameter,
 } from '@opentrons/shared-data'
+import { Run } from '@opentrons/api-client'
+import { UseQueryResult } from 'react-query'
```
jerader: same

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
```diff
@@ -34,13 +41,28 @@ export function ProtocolRunRuntimeParameters({
...
   const { t } = useTranslation('protocol_setup')
   const mostRecentAnalysis = useMostRecentCompletedAnalysis(runId)
-  const runTimeParameters = mostRecentAnalysis?.runTimeParameters ?? []
-  const hasParameter = runTimeParameters.length > 0
-
-  const hasCustomValues = runTimeParameters.some(
+  const runStatus = useRunStatus(runId)
+  const isRunTerminal =
+    // @ts-expect-error: expected that runStatus may not be a terminal status
+    runStatus == null ? false : RUN_STATUSES_TERMINAL.includes(runStatus)
```
jerader: instead of ts-expect-erroring all of these, you can type `RUN_STATUSES_TERMINAL` as `RunStatus[]`

app/src/organisms/Devices/hooks/useRunStatuses.ts
```diff
@@ -29,9 +27,8 @@ export function useRunStatuses(): RunStatusesInfo {
     runStatus === RUN_STATUS_RUNNING ||
     runStatus === RUN_STATUS_AWAITING_RECOVERY
   const isRunTerminal =
-    runStatus === RUN_STATUS_SUCCEEDED ||
-    runStatus === RUN_STATUS_STOPPED ||
-    runStatus === RUN_STATUS_FAILED
+    // @ts-expect-error: runStatus expected to not necessarily be in RUN_STATUSES_TERMINAL
```
jerader: once you type `RUN_STATUSES_TERMINAL` you can remove this ts-expect-error

app/src/organisms/RunPreview/index.tsx
```diff
@@ -41,6 +45,15 @@ export const RunPreviewComponent = (
 ): JSX.Element | null => {
   const { t } = useTranslation('run_details')
   const robotSideAnalysis = useMostRecentCompletedAnalysis(runId)
+  const runStatus = useRunStatus(runId)
+  // @ts-expect-error: expected that runStatus may not be a terminal status
```
jerader: same as above

app/src/organisms/RunTimeControl/hooks.ts
```diff
@@ -78,11 +80,7 @@ export function useRunStatus(
     refetchInterval: DEFAULT_STATUS_REFETCH_INTERVAL,
     enabled:
       lastRunStatus.current == null ||
-      !([
-        RUN_STATUS_FAILED,
-        RUN_STATUS_SUCCEEDED,
-        RUN_STATUS_STOPPED,
-      ] as RunStatus[]).includes(lastRunStatus.current),
+      !(RUN_STATUSES_TERMINAL as RunStatus[]).includes(lastRunStatus.current),
```
jerader, line 83: should be able to remove this typing when `RUN_STATUSES_TERMINAL` is typed

components/src/molecules/ParametersTable/InfoScreen.tsx
```diff
@@ -8,14 +8,25 @@ import { Flex } from '../../primitives'
...
+  let bodyText: string = ''
+  switch (contentType) {
+    case 'parameters':
+      bodyText = 'No parameters specified in this protocol'
+      break
+    case 'moduleControls':
+      bodyText = 'Connect modules to see controls'
+      break
+    case 'runNotStarted':
+      bodyText = 'Run was never started'
```
jerader: can we add these to i18n? you'll have to pass it down as a prop since `components` doesn't have i18n

jerader (approved):
lgtm, i don't wanna block this from merging so all the comments i left can be addressed in a follow up

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
ncdiehl11: wow super helpful thank you!

components/src/molecules/ParametersTable/InfoScreen.tsx
ncdiehl11: I can, but am I wrong in thinking we removed that translation object prop somewhere along the way? Do we remember the reason for this?

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
```diff
@@ -77,9 +103,15 @@ export function ProtocolRunRuntimeParameters({
           </Banner>
         ) : null}
       </Flex>
-      {!hasParameter ? (
+      {!hasRunTimeParameters ? (
         <Flex padding={SPACING.spacing16}>
-          <InfoScreen contentType="parameters" />
+          <InfoScreen
+            contentType={
+              !hasRunStarted && runStatus === RUN_STATUS_STOPPED
```
shlokamin: i dont think we care whether the run has been stopped right? we just care if its has been started

components/src/molecules/ParametersTable/InfoScreen.tsx
shlokamin: yeah lets use i18n here  — you can use the `useTranslation` hook directly from this component

shlokamin (approved):

app/src/organisms/Devices/ProtocolRun/ProtocolRunRunTimeParameters.tsx
ncdiehl11: I think we do need to check whether the run has been stopped because otherwise we will show "run never started" immediately when starting protocol setup for a non-RTP protocol, and we actually want to show "no parameters" in that case

ncdiehl11: example:
https://github.com/Opentrons/opentrons/assets/47604184/20d97382-2e6c-4bfa-a8dd-bdd9bd0dc20e

PR #14962, opened 2024-04-19
feat(test-data-generation): first pass at generating deck configurations

DerekMaggio:
# Overview

With the large code change in https://github.com/Opentrons/opentrons/pull/14684 we want to test it as thoroughly as possible. This PR adds generation of test cases with [hypothesis](https://hypothesis.readthedocs.io/en/latest/index.html) for deck configuration

The idea is that hypothesis will generate DeckConfiguration objects that represent what a user defines in their protocol or their deck configuration in the UI. These objects will then end up being used to auto-generate Python protocols to pipe through analysis to exercise our deck configuration validation logic

# Test Plan

- I went through some of the generated deck configurations to verify they were being created correctly

# Changelog

- Create datashapes.py
  - This defines a simplified deck configuration model and all of its contents
- Create helper_strategies.py
  - This file provides the building block strategies that are utilized to make a final strategy that makes a DeckConfiguration object
- Create final_strategies.py
  - This contains the logic for generating the final DeckConfiguration objects

# Review requests

- Should I add some tests to confirm that DeckConfiguration objects are being generated as expected? 

# Risk assessment

None.... yet


test-data-generation/src/test_data_generation/deck_configuration/datashapes.py
```diff
@@ -0,0 +1,82 @@
...
+    """Possible contents of a slot on a Flex."""
+
+    EMPTY = enum.auto()
+    THERMOCYCLER_MODULE = enum.auto()
+    MAGNETIC_BLOCK_MODULE = enum.auto()
+    TEMPERATURE_MODULE = enum.auto()
+    HEATER_SHAKER_MODULE = enum.auto()
+    TRASH_BIN = enum.auto()
+    WASTE_CHUTE = enum.auto()
+    LABWARE = enum.auto()  # This will need to be here for generating protocols
```
SyntaxColoring: > The idea is that the created DeckConfiguration objects represent **what a user defines either in their protocol** or in their deck configuration in the UI.

I see the appeal of aiming for both of these birds with one stone, but I suspect we'll want to keep deck configuration generation separate from protocol generation. They're two different inputs to our system that have two different uses and obey two different sets of constraints.

But we can keep this for now and see how it turns out.

test-data-generation/src/test_data_generation/deck_configuration/datashapes.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+@dataclasses.dataclass
+class DeckConfiguration:
+    """The deck on a Flex."""
+
+    A: Row
+    B: Row
+    C: Row
+    D: Row
```
SyntaxColoring: Style nit, totally optional: I would use lowercase `a`/`b`/`c`/`d` for these attribute names to match the usual Python conventions, even though that makes them not exactly match how they're labeled on the robot.

test-data-generation/src/test_data_generation/deck_configuration/assumptions.py
```diff
@@ -0,0 +1,45 @@
...
+    """Return True if the row has a staging area."""
+    return row.col4 is not None
+
+
+class EvaluationFunction(typing.Protocol):
+    """A method that validates a deck configuration."""
+
+    def __call__(self, deck: DeckConfiguration) -> bool:
+        """Evaluates the deck configuration and returns True if it is valid."""
+        ...
```
SyntaxColoring: > What is the best way to determine if my created valid and invalid deck configurations meet our internal requirements of being valid and invalid? I am thinking of just starting to throw them at the analysis engine and the deck_configuration endpoint and see what happens.

Our "official" validation lives in [robot_server/deck_configuration/validation.py](https://github.com/Opentrons/opentrons/blob/26929a2d92715ab23d7a4832f2548a2bab565d11/robot-server/robot_server/deck_configuration/validation.py) and [robot_server/deck_configuration/validation_mapping.py](https://github.com/Opentrons/opentrons/blob/26929a2d92715ab23d7a4832f2548a2bab565d11/robot-server/robot_server/deck_configuration/validation_mapping.py). I don't have an informed opinion about whether it's a good idea to just `import` that directly, or go through `python -m opentrons.cli analyze`, or what.

As far as the machine-readable schema is concerned, it's currently pretty weakly-defined. For example, the fixture names are typed as `str`s instead of `enum`s, and we validate that them "manually," in our application layer, instead of in our model layer. The idea was to avoid duplicating things from the deck definition. A downside to this is that you can't just feed Hypothesis our Pydantic models and have it automatically generate all the valid deck definitions, because it will generate bogus values for those strings.

On the other hand, maybe that's a good thing? Maybe we *should* be generating a bunch of invalid deck configurations to test with, as long as they don't drown out the valid ones? I'm curious what you think about this, philosophically.

There's also https://pypi.org/project/hypothesis-jsonschema/. I suppose we could use that to generate deck configurations from robot-server's OpenAPI spec. Or add a deck configuration schema to shared-data, and generate it from that. Those options would have the same `str` vs. `enum` problem, but we might have more options for dealing with that at the JSON schema level because JSON schemas can reference and include each other.

test-data-generation/src/test_data_generation/deck_configuration/strategies.py
```diff
@@ -0,0 +1,66 @@
...
+    return deck
+
+
+@st.composite
+def an_invalid_deck(draw: st.DrawFn) -> DeckConfiguration:
+    """Generate an invalid deck."""
+    deck = draw(a_deck_configuration())
+
+    for evaluation_function in VALID_DECK_CONFIGURATION_EVALUATION_FUNCTIONS:
+        assume(not evaluation_function(deck))
```
SyntaxColoring: I don't know Hypothesis so I might be missing something, but does this do what we want? This reads to me like it will generate decks that are invalid according to *every single evaluation function,* whereas I would have thought we'd want decks that are invalid according to *at least one evaluation function.*

test-data-generation/tests/test_data_generation/deck_configuration/test_deck_configuration.py
```diff
@@ -0,0 +1,15 @@
+from hypothesis import given, settings
+from test_data_generation.deck_configuration.datashapes import DeckConfiguration
+from test_data_generation.deck_configuration.strategies import a_valid_deck
+
+@given(deck=a_valid_deck())
+@settings(max_examples=200)
+def test_a_valid_deck(deck: DeckConfiguration) -> None:
```
SyntaxColoring: I'm a little confused about what this is doing. This is not yet testing our production code, right? It's just sanity-checking that the Hypothesis machinery for generating "a valid deck" is actually generating stuff that looks like "a valid deck"?

SyntaxColoring (approved):
This all looks reasonable to me so far. Thank you!

test-data-generation/tests/test_data_generation/deck_configuration/test_deck_configuration.py
DerekMaggio: correct

test-data-generation/src/test_data_generation/deck_configuration/datashapes.py
DerekMaggio: Maybe have separate logic that looks the DeckConfiguration logic and then figures out what labware to place in a protocol?

test-data-generation/src/test_data_generation/deck_configuration/strategies.py
DerekMaggio: ah. That makes sense. You are correct.

test-data-generation/src/test_data_generation/deck_configuration/assumptions.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def _no_modules_in_staging_area_col_4(deck: DeckConfiguration) -> bool:
+    """Return True if there are no modules in the staging area of col4."""
+    return [
+        row
+        for row in deck.rows
+        if _row_has_staging_area(row) and row.col4.is_a_module()
+    ] == []
```
CaseyBatten: I don't believe this is necessary for a few reasons. Modules for the deck are now loaded the same as fixtures, into _cutouts_. Fundamentally deck configuration items can only be loaded if they have a cutout associated with them. There are no cutouts for column 4, there really shouldn't even be a column 4 in the deck configuration. Instead, you'd have a fixture/module in column 3 of the deck config that provides an addressable area claiming to _be_ column 4. Ex: Staging area slot in cutoutD3 provides addressable areas: "D3" and "D4".

CaseyBatten (commented):
Overall looks like a good start but I have some questions below regarding a few validations. Also just to clear up some of the theory behind this PR, I think it's important not to mix up the notion of literal "areas" (or addressable areas as it were) on the deck versus the "fixtures" in a deck's configuration file. The deck configuration represents 12 exact physical cutouts in the deck, and the physical hardware mounted within them. The engine interprets that configuration file's information and loads the definitions for the fixtures provided in those 12 cutouts. From those definitions we then find out what "spaces" on the deck are provided by each of our fixtures. A generic `singleRightSlot` fixture may only provide a "slot" called "D3" on the deck, but a `stagingAreaRightSlot` fixture would provide a "D3" slot and a "D4" slot. The deck config doesn't see 4 columns, it only sees that some fixture was loaded into cutout D3, what that fixture "creates" is an engine concern.

test-data-generation/src/test_data_generation/deck_configuration/assumptions.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def _no_modules_in_staging_area_col_3(deck: DeckConfiguration) -> bool:
+    """Return True if there are no modules in the staging area of col3."""
+    return [
+        row
+        for row in deck.rows
+        if _row_has_staging_area(row) and row.col3.is_a_module()
+    ] == []
```
CaseyBatten: This is probably incorrect, as currently we have at least one module fixture which provides staging area slots, the `stagingAreaSlotWithMagneticBlockV1`. What this looks like is a Physical staging areas slot with a magnetic block in the column 3 position and an empty staging area to the right, awaiting labware. 
Also as a terminology thing, I don't think we refer to the column 3 positions as staging areas, maybe a more generic name.

test-data-generation/tests/test_data_generation/deck_configuration/test_deck_configuration.py
```diff
@@ -0,0 +1,27 @@
+"""Tests to ensure that the deck configuration is generated correctly."""
+
+import pytest
+from hypothesis import given, settings, HealthCheck
+from test_data_generation.deck_configuration.datashapes import DeckConfiguration
+from test_data_generation.deck_configuration.strategies import an_invalid_deck
+
+unique_values = set()
```
sfoster1: you can probably define this in the fixture and yield it, since python is by reference you'll be able to get it after. e.g.
```
@pytest.fixture
def print_unique_values():
    unique_values = set()
    yield unique_values
    print(unique_values)
```

then it doesn't have to be global

test-data-generation/src/test_data_generation/deck_configuration/strategies.py
```diff
@@ -0,0 +1,128 @@
...
+            col3=a_slot(row=row, col="3"),
+        )
+    )
+
+
+@st.composite
+def a_deck_configuration(draw: st.DrawFn) -> DeckConfiguration:
+    """Generate a deck with random rows."""
+    row_a_contents = draw(a_row("A"))
+    row_b_contents = draw(a_row("B"))
```
sfoster1: any reason for the asymmetry here between row_a/row_b and the c and d rows being inline to the call?

test-data-generation/src/test_data_generation/deck_configuration/deck_evaluation.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@mark_invalid
+def has_invalid_fixture_in_slot_2(deck: DeckConfiguration) -> bool:
+    """Return True if the deck has an invalid fixture in slot 2."""
+    valid_fixtures = [PSC.LABWARE_SLOT, PSC.MAGNETIC_BLOCK_MODULE]
+    for slot in deck.slots:
+        if slot.col == "2" and slot.contents not in valid_fixtures:
+            return True
+    return False
```
CaseyBatten: Totally correct and this test should provide true decks for the current set of fixtures/modules. Maybe in the future if we invent more fixtures that occupy col 2 to we can automatically pull this data from the `cutoutFixtures` section of `shared-data/deck/definitions/5/ot3_standard.json` by whichever `mayMountTo` lists overlap with column 2, but thats not really necessary.

test-data-generation/src/test_data_generation/deck_configuration/datashapes.py
```diff
@@ -0,0 +1,193 @@
...
+SlotName = typing.Literal[
+    "A1", "A2", "A3", "B1", "B2", "B3", "C1", "C2", "C3", "D1", "D2", "D3"
+]
+
+
+class PossibleSlotContents(enum.Enum):
+    """Possible contents of a slot on a Flex."""
+
+    # Implicitly defined fixtures
+    THERMOCYCLER_MODULE = enum.auto()
```
CaseyBatten, line 18: Should not be a problem for generated decks, but depending on how we test and use them going forward it's important to note there are actually **two** thermocycler fixtures defined in the v5 `ot3_standard.json` deck definitions file. They are `thermocyclerModuleV2Rear` and `thermocyclerModuleV2Front`. These fixtures belong to the same `fixtureGroup`, which means when one or the other is loaded into the deck both are expected to be present. Going forward, any module/fixture that occupies multiple cutouts or "slots" will have a fixtureGroup. The important thing this means for actual runs is that only one of these two "fixtures" provides an addressable area, so when calculating geometry and movement only one of them provides the "origin" for our math calculations. I do not think this will be an issue for the tests you have here, however if we ever do more sophisticated deck generation and validation these `fixtureGroups` might come into play.

CaseyBatten (approved):
This looks great and the changes so far seem comprehensive. I've got two comments below regarding some possible future behavior to keep in mind, but nothing blocking.

sfoster1 (approved):
This looks great, love the explicit hypotheses and the approach towards generating the data.

test-data-generation/tests/test_data_generation/deck_configuration/test_deck_configuration.py
DerekMaggio: I have since removed

test-data-generation/src/test_data_generation/deck_configuration/strategies.py
DerekMaggio: removed this too

test-data-generation/tests/test_data_generation/deck_configuration/test_deck_configuration.py
```diff
@@ -0,0 +1,41 @@
...
+
+@given(
+    deck_config=a_deck_configuration_with_a_module_or_trash_slot_above_or_below_a_heater_shaker()
+)
+@settings(
+    max_examples=NUM_EXAMPLES,
+    suppress_health_check=[HealthCheck.filter_too_much, HealthCheck.too_slow],
+)
+def test_above_below_heater_shaker(deck_config: DeckConfiguration) -> None:
+    """I hypothesize, that any deck configuration with a non-labware slot fixture above or below a heater-shaker is invalid."""
```
y3rsh, line 21: helpful

test-data-generation/tests/test_data_generation/deck_configuration/test_deck_configuration.py
```diff
@@ -0,0 +1,41 @@
...
+    # assert e.exception == "Some statement about the deck configuration being invalid because of the labware above or below the Heater-Shaker"
+
+
+@given(deck_config=a_deck_configuration_with_invalid_fixture_in_col_2())
+@settings(
+    max_examples=NUM_EXAMPLES,
+    suppress_health_check=[HealthCheck.filter_too_much, HealthCheck.too_slow],
+)
+def test_invalid_fixture_in_col_2(deck_config: DeckConfiguration) -> None:
+    """I hypothesize, that any deck configuration that contains at least one, Heater-Shaker, Trash Bin, or Temperature module, in column 2 is invalid."""
```
y3rsh, line 38: ❤️ 😊

y3rsh (approved):
let me have the protocols :shipit:

PR #14963, opened 2024-04-19
feat(api): add a reload-labware command

sfoster1:
Adds a new command ReloadLabware, which allows dispatchers to change all the details of a loaded labware except for the location. This is primarily intended to allow getting a new labware offset that was not added to the engine by the time this labware was loaded (though it can technically do more, for symmetry).

## Testing
- [x] Put it in a json protocol and run it, I suppose

This doesn't really change a whole lot of behavior and is well-supported with testing. It's a prerequisite for #14940 

Closes RSQ-29

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/reload_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/reload_labware.py
@@ -56,7 +56,7 @@ class ReloadLabwareResult(BaseModel):
 class ReloadLabwareImplementation(
     AbstractCommandImpl[ReloadLabwareParams, ReloadLabwareResult]
 ):
-    """Load labware command implementation."""
+    """Reload labware command implementation."""
 
     def __init__(
         self, equipment: EquipmentHandler, state_view: StateView, **kwargs: object
@@ -77,7 +77,7 @@ class ReloadLabwareImplementation(
 
 
 class ReloadLabware(BaseCommand[ReloadLabwareParams, ReloadLabwareResult]):
-    """Load labware command resource model."""
+    """Reload labware command resource model."""
 
     commandType: ReloadLabwareCommandType = "reloadLabware"
     params: ReloadLabwareParams
@@ -87,7 +87,7 @@ class ReloadLabware(BaseCommand[ReloadLabwareParams, ReloadLabwareResult]):
 
 
 class ReloadLabwareCreate(BaseCommandCreate[ReloadLabwareParams]):
-    """Load labware command creation request."""
+    """Reload labware command creation request."""
 
     commandType: ReloadLabwareCommandType = "reloadLabware"
     params: ReloadLabwareParams
--- a/api/src/opentrons/protocol_engine/commands/reload_labware.py
+++ b/api/src/opentrons/protocol_engine/commands/reload_labware.py
@@ -4,15 +4,6 @@ from pydantic import BaseModel, Field
 from typing import TYPE_CHECKING, Optional, Type
 from typing_extensions import Literal
 
-from opentrons_shared_data.labware.labware_definition import LabwareDefinition
-from ..resources import labware_validation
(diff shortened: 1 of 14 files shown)


api/src/opentrons/protocol_engine/commands/reload_labware.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+    version: int = Field(
+        ...,
+        description="The labware definition version.",
+    )
+    displayName: Optional[str] = Field(
+        None,
+        description="An optional user-specified display name "
+        "or label for this labware.",
+    )
```
SyntaxColoring: I would have expected this to only take an `id`. What do `loadName`/`namespace`/`version` do if they're different from the original? Replace the labware with a different one, but keep the ID?

sfoster1: Correct! Honestly I don't know why someone would want to do this, but it does work, and it leads to better symmetry with the load command

SyntaxColoring: Gotcha. Changing the `loadName`/`namespace`/`version` makes me nervous. No concrete reason, but it feels pretty good to me that a labware ID immutably points to a single thing. How do you feel about leaving this out until we have a use case for it? 

I think `loadLabware` does let you do this, but it's always kind of been by accident.

TamarZanzouri: I agree with @SyntaxColoring. maybe we should leave this out for now?

sfoster1: Hmm. Okay, though I think it's going to do so little at that point that it'll look weird. I'm not so nervous about the changing-the-definition thing, I think this part of the code is really well tested and generally capable.

SyntaxColoring (approved):
Looks like there are some trivial linting errors, but otherwise this LGTM. Thank you!

api/src/opentrons/protocol_engine/commands/reload_labware.py
```diff
@@ -0,0 +1,95 @@
...
+    from ..execution import EquipmentHandler
+
+
+ReloadLabwareCommandType = Literal["reloadLabware"]
+
+
+class ReloadLabwareParams(BaseModel):
+    """Payload required to load a labware into a slot."""
+
+    labwareId: str = Field(
```
TamarZanzouri, line 20: Ty!

api/src/opentrons/protocol_engine/commands/reload_labware.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+        return ReloadLabwareResult(
+            labwareId=params.labwareId,
+            offsetId=reloaded_labware.offsetId,
+        )
+
+
+class ReloadLabware(BaseCommand[ReloadLabwareParams, ReloadLabwareResult]):
+    """Load labware command resource model."""
```
TamarZanzouri: Reload?

api/src/opentrons/protocol_engine/commands/reload_labware.py
```diff
@@ -0,0 +1,95 @@
...
+
+    commandType: ReloadLabwareCommandType = "reloadLabware"
+    params: ReloadLabwareParams
+    result: Optional[ReloadLabwareResult]
+
+    _ImplementationCls: Type[ReloadLabwareImplementation] = ReloadLabwareImplementation
+
+
+class ReloadLabwareCreate(BaseCommandCreate[ReloadLabwareParams]):
+    """Load labware command creation request."""
```
TamarZanzouri: Reload?

TamarZanzouri (approved):
Lgtm! Some minor things but looks good!

PR #14964, opened 2024-04-19
fix(api): add case correction to module_context.load_labware

sanni-t:
Closes RESC-235 and RQA-2610

# Overview

Fixes the `module_context.load_labware()` method by ensuring that labware names are converted into lower case before using them in any part of the code. 

The escalations issue above was caused by a cascade of things happening because the protocol had a module labware loaded with the labware name written in mixed case. There are a few places where we do string comparison of the labware names, including when trying to find new versions of a labware and when looking up LPC offsets for a labware. These string comparisons would fail because all labware definitions have lowercase names, and hence would lead to unexpected behavior. It should no longer create such a problem 

# Test Plan

Tested that this protocol's run uses version 2 of the Nest PCR plate and applies the correct LPC offset to it:

```
requirements = {"robotType": "OT-2", "apiLevel": "2.15"}

def run(protocol_context):

	p300racks = [protocol_context.load_labware('opentrons_96_tiprack_300ul', 2)]
	p300 = protocol_context.load_instrument('p300_single', 'right', tip_racks=p300racks)
	tc = protocol_context.load_module(module_name="thermocyclerModuleV1")
	reaction_plate1 = protocol_context.load_labware(
	    'Nest_96_wellplate_100uL_pcr_full_skirt', 1)
	reaction_plate3 = tc.load_labware('Nest_96_wellplate_100uL_pcr_full_skirt')

	tc.open_lid()
	p300.pick_up_tip()
	p300.aspirate(300, reaction_plate1['A1'])
	p300.dispense(10, reaction_plate3['A1'])
	p300.drop_tip()
```

# Risk assessment

Very low. Tiny bug fix

### DIFF
--- a/api/src/opentrons/protocol_api/module_contexts.py
+++ b/api/src/opentrons/protocol_api/module_contexts.py
@@ -151,7 +151,7 @@ class ModuleContext(CommandPublisher):
             load_location = loaded_adapter._core
         else:
             load_location = self._core
-
+        name = validation.ensure_lowercase_name(name)
         labware_core = self._protocol_core.load_labware(
             load_name=name,
             label=label,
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
classic

PR #14967, opened 2024-04-19
docs(api): new behavior for `Labware.set_offset()` in PAPI 2.18

ecormany:
# Overview

Documents [updated behavior](#14940) for `Labware.set_offset()` in Python API version 2.18.

Addresses RTC-234

# Test Plan

Sandbox:
- [API reference entry](http://sandbox.docs.opentrons.com/docs-set_offset-updates/v2/new_protocol_api.html#opentrons.protocol_api.Labware.set_offset)
- [new section](http://sandbox.docs.opentrons.com/docs-set_offset-updates/v2/new_advanced_running.html#labware-offset-behavior) of Advanced Control
 
# Changelog

- adapted new docstring draft written by @sfoster1 
- adapt sample code and move it to Advanced Control

# Review requests

have we accurately captured the three phases of behavior (2.12–13, 14–17, 18+)?

# Risk assessment

none

### DIFF
--- a/api/docs/v2/new_advanced_running.rst
+++ b/api/docs/v2/new_advanced_running.rst
@@ -134,24 +134,9 @@ Keep in mind that ``set_offset()`` commands will override any labware offsets se
 Labware Offset Behavior
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-The effects of ``set_offset()`` vary depending on the API level of your protocol.
+How the API applies labware offsets varies depending on the API level of your protocol. This section describes the latest behavior. For details on how offsets work in earlier API versions, see the API reference entry for :py:meth:`.set_offset`.
 
-.. list-table::
-    :header-rows: 1
-    
-    * - API version 
-      - Offset behavior
-    * - 2.12–2.13
-      - Offsets only apply to the exact :py:class:`.Labware` instance.
-    * - 2.14–2.17
-      - ``set_offset()`` is not available, and the API raises an error.
-    * - 2.18 and newer
-      - 
-        - Offsets apply to any labware of the same type, in the same on-deck location. 
-        - Offsets can't be set on labware that is currently off-deck.
-        - Offsets do not follow a labware instance when using :py:meth:`.move_labware`.
-
-For example, in the latest API version, if you use ``set_offset()`` on a tip rack, use all the tips, and replace the rack with a fresh one of the same type in the same location, the offsets will apply to the fresh tip rack::
+In the latest API version, offsets apply to labware type–location combinations. For example,  if you use ``set_offset()`` on a tip rack, use all the tips, and replace the rack with a fresh one of the same type in the same location, the offsets will apply to the fresh tip rack::
 
     tiprack = protocol.load_labware(
         load_name="opentrons_flex_96_tiprack_1000ul", location="D3"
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_api/labware.py
```diff
@@ -577,22 +577,42 @@ def set_offset(self, x: float, y: float, z: float) -> None:
...
-            because it will produce undefined behavior.
-            Instead, use Labware Position Check in the app or on the touchscreen.
+            Setting offsets with this method will override any labware offsets set
+            by running Labware Position Check in the Opentrons App.
+
+            This method is designed for use with mechanisms like
+            :obj:`opentrons.execute.get_protocol_api`, which lack an interactive way
+            to adjust labware offsets. (See :ref:`advanced-control`.)
+
+        .. versionchanged:: 2.14
```
sfoster1, line 608: swap the order since "new in..." is autogenerated at the bottom?

ecormany: yeah…other Python references go in ascending order. but they also manage to get the "new in" notices above them. ours come from `@requires_version` rather than `..versionadded::`.

so this is consistent with other entries in our ref, if not ideal. we can revisit another time.

api/docs/v2/new_advanced_running.rst
```diff
@@ -118,11 +123,65 @@ This automatically generated code uses generic names for the loaded labware. If
...
+	Improperly reusing offset data may cause your robot to move to an unexpected position or crash against labware, which can lead to incorrect protocol execution or damage your equipment. When in doubt: run Labware Position Check again and update your code!
+
+.. _labware-offset-behavior:
+
+Labware Offset Behavior
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The effects of ``set_offset()`` vary depending on the API level of your protocol.
+
+.. list-table::
```
jwwojak: Great use of a table here.

api/src/opentrons/protocol_api/labware.py
```diff
@@ -577,22 +577,34 @@ def set_offset(self, x: float, y: float, z: float) -> None:
...
-        to adjust labware offsets. (See :ref:`advanced-control`.)
+            - In API level 2.13 and earlier, the offset applies only to the labware
+              instance that ``set_offset()`` is called on.
+            - In API levels 2.14–2.17, ``set_offset()`` is not available, and the API
+              raises an error.
+            - In API level 2.18 and newer, the offset applies to any labware of the
+              same type, in the same on-deck location. Labware type is defined as the
+              combination of the ``loadName``, ``namespace``, and ``version`` parameters
+              of :py:meth:`~.ProtocolContext.load_labware`.
+              See :ref:`labware-offset-behavior` for examples.
```
jwwojak: I wonder if a table would work here, similar to above. You might get rid of some repetition via the column headers:

Col 1 header: API Level
Col 2 header: Description

Eliminates the need to repeat "In API level x.x ...". In the first column, just write "API 2.13 and earlier"

In the first and 3rd bullets, the text uses "earlier" and "newer" respectively. It might be stronger or more clear to use "older" and "newer." Something about "earlier" sort of threw me off. I had to think for a moment, which probably explains the burning smell.

jwwojak (commented):
Very nice, clear. Left a few comments for your consideration.

api/src/opentrons/protocol_api/labware.py
ecormany: I don't think we can embed a table in a docstring, although I'll give it a try.

You're right about the "earlier/newer" mismatch. Our standard is "earlier/later", and I'll update it.

ecormany: actually this works and looks good! 
image


gonna swap things around. we don't need the table twice.

PR #14972, opened 2024-04-22
refactor(protocol-designer): tip position modal max values round down

jerader:
closes AUTH-352

# Overview

Since we are rounding the ranges for the x/y values to the nearest 1 decimals, they were being rounded up for the max value which resulted in the highest value in the description to be out of range. This PR rounds down instead of up.

# Test Plan

Just review the code, the test cases tests this.

# Changelog

- use lodash `floor` to round down if needed
- fix test

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepEditForm/fields/TipPositionField/utils.ts
+++ b/protocol-designer/src/components/StepEditForm/fields/TipPositionField/utils.ts
@@ -61,7 +61,6 @@ export const roundValue = (
       return floor(Number(value), DECIMALS_ALLOWED)
     }
   }
-  return value === null ? 0 : round(Number(value), DECIMALS_ALLOWED)
 }
 
 const OUT_OF_BOUNDS: 'OUT_OF_BOUNDS' = 'OUT_OF_BOUNDS'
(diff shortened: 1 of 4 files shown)


protocol-designer/src/components/StepEditForm/fields/TipPositionField/utils.ts
```diff
@@ -46,7 +47,20 @@ export function getDefaultMmFromBottom(args: {
...
+
+  switch (direction) {
+    case 'up': {
+      return round(Number(value), DECIMALS_ALLOWED)
+    }
+    case 'down': {
+      return floor(Number(value), DECIMALS_ALLOWED)
+    }
+  }
   return value === null ? 0 : round(Number(value), DECIMALS_ALLOWED)
```
koji: I think this line is not needed anymore since the line itself is unreachable and the above switch statement covers the `value !== null`

koji (approved):
left one comment but the changes look good to me.

protocol-designer/src/components/StepEditForm/fields/TipPositionField/utils.ts
jerader: oops thanks, i forgot to delete it

PR #14976, opened 2024-04-22
fix(app): add robotSerialNumber to proceedToRun event

koji:
# Overview
Add robotSerialNumber to `proceedToRun`

A few of `proceedToRun` events don't have robotSerialNumber because the events are sent before selecting a robot.
fix RQA-2386

I think we would need to create a custom hook that can handle trackEvents efficiently and we would need to update the package version. We use `2.22.1` but the latest version is `2.49.0`. 

I think we would need to organize all information about trackEvents to test all cases properly.

# Test Plan

# Changelog
- pass robotSerialNumber to cases that can retrieve robot serial number
- modify tests for the above update

# Review requests

# Risk assessment
low


mjhuff (approved):
Yeah, this makes sense. Thank you for updating!

PR #14977, opened 2024-04-23
refactor(monorepo): update lint-js command

koji:
# Overview
update lint-js command to speed up the linting process

ESLint and Prettier can be run in parallel to reduce processing time.  Using `pid` to handle commands parallelly. 

Add `--cache`
https://eslint.org/blog/2015/09/eslint-v1.4.0-released/#cached-results

close AUTH-363

# Test Plan
`make lint-js`

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,11 +214,17 @@ lint-py: $(PYTHON_LINT_TARGETS)
 %-py-lint:
 	$(MAKE) -C $* lint
 
-lint-js:
-	@pids=""; \
-	yarn eslint --quiet=$(quiet) --ignore-pattern "node_modules/" --cache ".*.@(js|ts|tsx)" "**/*.@(js|ts|tsx)" & pids="$$pids $$!"; \
-	yarn prettier --ignore-path .eslintignore --check $(FORMAT_FILE_GLOB) & pids="$$pids $$!"; \
-	for pid in $$pids; do wait $$pid; done
+.PHONY: lint-js
+lint-js: lint-js-eslint lint-js-prettier
+
+.PHONY: lint-js-eslint
+lint-js-eslint:
+	yarn eslint --quiet=$(quiet) --ignore-pattern "node_modules/" --cache ".*.@(js|ts|tsx)" "**/*.@(js|ts|tsx)"
+
+.PHONY: lint-js-prettier
+lint-js-prettier:
+	yarn prettier --ignore-path .eslintignore --check $(FORMAT_FILE_GLOB)
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -214,10 +214,11 @@ lint-py: $(PYTHON_LINT_TARGETS)
 %-py-lint:
 	$(MAKE) -C $* lint
 
-.PHONY: lint-js
```
SyntaxColoring, line 217: We should keep this `.PHONY` marker so `make` doesn't mistakenly think `lint-js` is supposed to be a filename.

Makefile
```diff
@@ -214,10 +214,11 @@ lint-py: $(PYTHON_LINT_TARGETS)
...
 lint-js:
-	yarn eslint --quiet=$(quiet) ".*.@(js|ts|tsx)" "**/*.@(js|ts|tsx)"
-	yarn prettier --ignore-path .eslintignore --check $(FORMAT_FILE_GLOB)
+	@pids=""; \
+	yarn eslint --quiet=$(quiet) --ignore-pattern "node_modules/" --cache ".*.@(js|ts|tsx)" "**/*.@(js|ts|tsx)" & pids="$$pids $$!"; \
+	yarn prettier --ignore-path .eslintignore --check $(FORMAT_FILE_GLOB) & pids="$$pids $$!"; \
+	for pid in $$pids; do wait $$pid; done
```
SyntaxColoring: This looks like it'll work, but I think we can do this without the PID management and manual `wait`. `make` is already designed for parallelism if we structure the dependencies for it. We could do this like:

```Makefile
.PHONY: lint-js
lint-js: lint-js-eslint lint-js-prettier

.PHONY: lint-js-eslint
lint-js-eslint:
	yarn eslint [...]

.PHONY: lint-js-prettier:
lint-js-prettier:
	yarn prettier [...]
```

And then running `make -j lint-js` will run `lint-js-eslint` and `lint-js-prettier` in parallel.

(And adding `--output-sync` will group the output more nicely, but it's not available on Apple's default `make`. You can install GNU `make` from [Homebrew](https://formulae.brew.sh/formula/make) instead.)

SyntaxColoring (approved):
TY!

sfoster1 (approved):
That's a lot better for `eslint` (first run is like 3-5 minutes, subsequent runs are 25 seconds) - enough that `prettier --check` becomes the limiting factor. Let's merge this IMO

PR #14980, opened 2024-04-23
feat(api): add option to ignore different tip presence states

caila-marashaj:
## Overview
This code adds an argument called `ht_operational_sensor` to `get_tip_presence_status`, that when used tells the api to only return the tip presence state of the instrument probe type specified. This allows calibration and partial tip flows to execute and check against their expected tip status without failing. 

## TODO
A follow-up pr will go up using this parameter for the `get_tip_presence` call in the calibration flow.

## Review Requests
I'll most likely address any non-blocking change requests in a follow-up pr so we can cut the internal release as fast as possible, but let me know if:

- `ht_operational_sensor` makes sense or if we can think of a better name
- we should otherwise go about anything differently here.

### DIFF
--- a/api/src/opentrons/hardware_control/backends/tip_presence_manager.py
+++ b/api/src/opentrons/hardware_control/backends/tip_presence_manager.py
@@ -3,7 +3,7 @@ from functools import partial
 from typing import cast, Callable, Optional, List, Set
 from typing_extensions import TypedDict, Literal
 
-from opentrons.hardware_control.types import TipStateType, OT3Mount
+from opentrons.hardware_control.types import TipStateType, OT3Mount, InstrumentProbeType
 
 from opentrons_hardware.drivers.can_bus import CanMessenger
 from opentrons_hardware.firmware_bindings.constants import NodeId
@@ -14,7 +14,10 @@ from opentrons_hardware.hardware_control.tip_presence import (
 from opentrons_shared_data.errors.exceptions import (
     TipDetectorNotFound,
     UnmatchedTipPresenceStates,
+    GeneralError,
 )
+
+from .ot3utils import sensor_id_for_instrument
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


api/src/opentrons/hardware_control/backends/tip_presence_manager.py
```diff
@@ -111,17 +114,40 @@ def current_tip_state(self, mount: OT3Mount) -> Optional[bool]:
...
+        results: List[tip_types.TipNotification],
+        ht_operational_sensor: Optional[InstrumentProbeType] = None,
+    ) -> TipStateType:
+        """
+        We can use ht_operational_sensor used to specify that we only care
+        about the status of one tip presence sensor on a high throughput
+        pipette, and the other is allowed to be different.
+        """
+        if ht_operational_sensor:
+            target_sensor_id = sensor_id_for_instrument(ht_operational_sensor)
```
CaseyBatten, line 127: In the case that an OT2 has somehow reached this stage, can it be expected that any sensors it has will have properly mapped IDs (primary/secondary), or do we assume an OT2 in this state would deflect to the raise case?

CaseyBatten (approved):
This makes sense and provides a solution path for problems we were running into with partial tip configurations when going beyond one column/row on a given side. Next steps on that particular front would be to identify the sensors associated with a given map, which may be abstract-able based on the nozzle maps primary nozzle and the nozzle ids covered within a given map. This is especially finite now given what we are learning about maximum map sizes available to us based on hardware limitations, so we may be able to add this to definitions somewhere or as an available parameter within the NozzleMap class whenever a given map is configured.

api/src/opentrons/hardware_control/backends/tip_presence_manager.py
vegano1: This should not happen, the OT2 does not have any `tip sensing` capabilities and does not use the tip presence manager.

y3rsh:
@caila-marashaj with the approval I am merging so I may build

PR #14981, opened 2024-04-23
fix(app): fix infinitely re-rendering/never rendering firmware success toasts

mjhuff:
Closes [RQA-2588](https://opentrons.atlassian.net/browse/RQA-2588)

# Overview

There are two issues with rendering firmware update toasts. 

First, the toasts depend on a request id stored as state within the ModuleCard, but ModuleCards most often unrender during the firmware update process, so this state is lost. This causes the toast never to render. 

The second issue is that sometimes, given the timing of the polling for attached modules, the module is always attached during the firmware update, thereby causing the module card not to unrender. When this happens, the useEffect hook responsible for making the success toast has conditional logic that is always true after an update, causing the toast to render infinitely. 

The solution to both these problems is to lift the request id state out of the module card itself and then abstract away the storage/retrieval via a utility hook, which is utilized by all parent components of ModuleCard.

### Current Behavior

https://github.com/Opentrons/opentrons/assets/64858653/3056ed9f-0c62-4248-a0db-90fde6e8a5d4

### Fixed Behavior

https://github.com/Opentrons/opentrons/assets/64858653/0f5261b9-4d68-462a-a7e2-5924efdb9383

# Test Plan
- Update a module's firmware then verify the update successful toast renders once. Let me know if you need help downgrading a module to test!

# Changelog
- Firmware update toasts now work.

# Risk assessment
low

[RQA-2588]: https://opentrons.atlassian.net/browse/RQA-2588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/assets/localization/en/device_details.json
+++ b/app/src/assets/localization/en/device_details.json
@@ -58,7 +58,7 @@
   "firmware_update_needed": "Instrument firmware update needed. Start the update on the robot's touchscreen.",
   "firmware_update_available": "Firmware update available.",
   "firmware_update_failed": "Failed to update module firmware",
-  "firmware_update_installation_successful": "Installation successful",
+  "firmware_updated_successfully": "Firmware updated successfully",
   "firmware_update_occurring": "Firmware update in progress...",
   "fixture": "Fixture",
   "have_not_run_description": "After you run some protocols, they will appear here.",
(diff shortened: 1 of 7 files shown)


koji (approved):
lgtm

thank you for teaching me the way to downgrade module version.

PR #14982, opened 2024-04-23
refactor(performance-metrics): separate storage logic from tracker

DerekMaggio:
# Overview

Closes https://opentrons.atlassian.net/browse/EXEC-415

- Pulls out storage logic from tracker into MetricsStore object
- Defines MetricsMetadata to store information about the MetricsStore object

# Test Plan

- Modified file storage test in performance-metrics to use the new classes. 
  - Loads file content and validates it 
- Modified test_track_analysis test in api/test_cli 
  - Loads file content and validates it 

# Changelog

- Create MetricsMetadata in shared data which defines where to store data, what to call it, and the headers 
- Create MetricsStore which handles creating any necessary directories or files and then storing data to the files
- Remove storage logic from RobotContextTracker then instantiate MetricsStore and call to its methods to handle data storage

# Review requests

None

# Risk assessment

Low

### DIFF
--- a/api/tests/opentrons/cli/test_cli.py
+++ b/api/tests/opentrons/cli/test_cli.py
@@ -299,6 +299,7 @@ def test_track_analysis(tmp_path: Path) -> None:
 
     with open(store.metadata.data_file_location, "r") as f:
         stored_data = f.readlines()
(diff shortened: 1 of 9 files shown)


sfoster1 (approved):
This all looks good to me, nice use of a typevar!

PR #14985, opened 2024-04-23
fix(api): Filter out `air_gap()` calls as higher-order commands

SyntaxColoring:
# Overview

This fixes RQA-2621.

# Test Plan

* [x] Follow the steps to reproduce in RQA-2621.

This PR adds unit tests and a "mid-level" integration test. @y3rsh mentioned he's working on adding an analysis snapshot test.

# Details

Before ~v5.0, Python protocols reported their activity as a nested tree, roughly like this:

1. Transfer
    1. Aspirate
    2. Dispense
2. Mix
    1. Aspirate
    2. Dispense

But the Protocol-Engine–based HTTP API introduced in v5.0 could only represent a flat list of the innermost commands. So our mapping logic in `LegacyCommandMapper` had to filter out the "higher-order" things like transfers and mixes, leaving only the "atomic" things like aspirates and dispenses.

We apparently forgot to filter out `air_gap()`, which looks like this:

1. Air gap
    1. Aspirate

The effect of forgetting to filter it out was that the engine would have two commands running at the same time: the outer air gap *and* the inner aspirate. This was a dormant bug; Protocol Engine is certainly not meant for that to work.

In #14726, we started being stricter about things like not having more than one command running at a time, so this started to raise an `AssertionError` (correctly, but confusingly).

The fix is just to filter out air gaps, like we filter out transfers, mixes, etc.

# Review requests

We should be on the lookout for any other higher-order commands that aren't being filtered out properly.

We should think about ways to make this inherently safer. Like, instead of mapping every command by default and filtering out the ones that we know to be higher-order, we might want to omit commands by default and only map the ones that we know to be at the deepest level of nesting. Or, we might want to defensively code `LegacyCommandMapper` to automatically close the last command before starting a new one.

# Risk assessment

Low risk to the fix, but there's definitely some risk that we're missing similar bugs.

### DIFF
--- a/api/tests/opentrons/protocol_runner/smoke_tests/test_legacy_command_mapper.py
+++ b/api/tests/opentrons/protocol_runner/smoke_tests/test_legacy_command_mapper.py
@@ -5,6 +5,7 @@ and mock in an isolated unit test environment.
 """
 from datetime import datetime
 from pathlib import Path
+from textwrap import dedent
 from typing import List
 
 import pytest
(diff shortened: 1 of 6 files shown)


api/tests/opentrons/protocol_runner/smoke_tests/test_legacy_command_mapper.py
```diff
@@ -753,3 +754,46 @@ async def test_zero_volume_dispense_commands(
...
+        pick_up_tip,
+        move_to_well,
+        air_gap_aspirate,
+    ] = result_commands
+    assert isinstance(initial_home, commands.Home)
+    assert isinstance(load_tip_rack, commands.LoadLabware)
+    assert isinstance(load_well_plate, commands.LoadLabware)
+    assert isinstance(load_pipette, commands.LoadPipette)
+    assert isinstance(pick_up_tip, commands.PickUpTip)
+    # TODO(mm, 2024-04-23): This commands.Custom looks wrong. This should be a commands.MoveToWell.
```
sanni-t, line 797: 👀 
Might be worth ticketing out if it looks like another bug.

sanni-t (approved):
LGTM!

PR #14986, opened 2024-04-23
feat(robot-server): add deck_configuration notification publisher

mjhuff:
Closes [EXEC-168](https://opentrons.atlassian.net/browse/EXEC-168)

# Overview

Add a deck configuration publisher to the robot server. This will enable the app to avoid polling for updated deck config state. An app-sided PR will follow this one.

# Test Plan
- On the desktop app/ODD, mess around with the deck config. Any time you update the deck state, you should see an MQTT refetch flag published on the `robot-server/deck_configuration` topic. Postman is the easiest way to visualize this.

# Changelog
- Add a deck configuration publisher to the robot server.

# Review requests
- Double check me on the `store.py` changes. I don't think there's anything weird/wrong here, but it would be nice for a python dev to confirm. We just publish a refetch flag any time the store is updated, and I believe I've covered the only spots updates happen.

# Risk assessment
low - this is about as vanilla as you can get for a notification publisher

[EXEC-168]: https://opentrons.atlassian.net/browse/EXEC-168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/deck_configuration/fastapi_dependencies.py
+++ b/robot-server/robot_server/deck_configuration/fastapi_dependencies.py
@@ -19,6 +19,10 @@ from robot_server.persistence.fastapi_dependencies import (
     get_active_persistence_directory,
     get_active_persistence_directory_failsafe,
 )
+from robot_server.service.notifications import (
+    DeckConfigurationPublisher,
+    get_deck_configuration_publisher,
+)
 
 
 # This needs to be kept in sync with opentrons.execute, which reads this file.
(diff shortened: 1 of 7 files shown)


TamarZanzouri (approved):
I was not part of the initial notifications but changes make sense to me and match other notifications

PR #14989, opened 2024-04-23
feat(step-generation, shared-data): pipette collision warnings

jerader:
closes AUTH-19

# Overview

Update the pipette collision warnings in step-generation for the 96-channel column pick up only for now by checking all adjacent slots to the labware of interest and checking to see if the pipette will go into the bounding boxes. The logic is sort of complicated but should match python.

# Test Plan

Create a flex protocol with a 96-channel attached. Add a heater-shaker module with an adapter and labware, and add a labware. Move the tiprack off of the tiprack adapter. Make sure the labware is next to the module. Create a move liquid step and make sure its a column pick up. See that the possible collision error occurs. Move the labware around and see that it disappears when its not next to the module.

I added a bunch of test cases to test all the options to make sure it works as expected.

NOTE: the thermocycler collision stuff isn't working yet, i'll investigate more and fix it in a followup PR. I left a comment to add test coverage for it.

# Changelog

- create `getFlexSurroundingSlots` in shared-data that returns all surrounding slots for a slot of interest only for the flex
- create `getIsSafePipetteMovement` with a bunch of helper utils to use in step-generation
- update the old collision warning error creator and make it more generic and delete the old 96-channel collision util that was being used
- add thorough test coverage
- fix all instances of the step-generation util in the compound pipetting commands

# Review requests

I added most of the utils to step-generation directly since I think only step-generation needs them. Lots of the utils rely on having `RobotState` and `InvariantContext` as props which is exclusive to step-generation. I added 1 util to shared-data and it is the util that returns all surrounding slots. Let me know if you have thoughts on that!

# Risk assessment

medium-ish, there is a lot of logic here but i tried to thoroughly test.


jerader:
TODO:

- [x] add some more test coverage

step-generation/src/__tests__/getIsSafePipetteMovement.test.ts
```diff
@@ -0,0 +1,167 @@
...
+      mockRobotState,
+      mockInvariantProperties,
+      mockPipId,
+      mockLabwareId,
+      mockTiprackId,
+      { x: 0, y: 0, z: 0 }
+    )
+    expect(result).toEqual(false)
+  })
+  //    todo(jr, 4/23/24): add more test cases, test thermocycler collision - i'll do this in a follow up
```
jerader, line 166: here is the follow up ticket https://opentrons.atlassian.net/browse/AUTH-367

step-generation/src/utils/safePipetteMovements.ts
```diff
@@ -0,0 +1,393 @@
...
+      //  labware on module
+      deckSlot = modules[labwareSlot].slot
+    } else if (labwareState[labwareSlot] != null) {
+      const adapterSlot = labwareState[labwareSlot].slot
+      if (modules[adapterSlot] != null) {
+        //  labware on adapter on module
+        deckSlot = modules[adapterSlot].slot
+      } else {
+        //  labware on adapter on deck
+        // eslint-disable-next-line no-unused-vars
```
jerader: i cannot seem to figure out why lint thinks line 359 isn't being used??? i'm so confused. If anyone has advice, that'd be great!

ncdiehl11 (commented):
I seem to be getting an unexpected message for following this test plan
Screenshot 2024-04-24 at 11 36 28 AM

jerader:
> I seem to be getting an unexpected message for following this test plan

@ncdiehl11 , oh you need to move the tiprack off of the adapter. my bad, let me add that to the test plan!

ncdiehl11:
> > I seem to be getting an unexpected message for following this test plan
> 
> @ncdiehl11 , oh you need to move the tiprack off of the adapter. my bad, let me add that to the test plan!

I did! Before the transfer step
Screenshot 2024-04-24 at 11 51 42 AM

ncdiehl11 (approved):
logic looks solid and I tested as expected with the caveat in the comment thread above

PR #14994, opened 2024-04-24
fix(app-shell,app): Send labware files and runtime parameters over USB

SyntaxColoring:
# Overview

Fixes RESC-247 and RQA-2627.

# Test Plan

* [x] Following the steps in RESC-247, make sure protocols with custom labware work over USB now.
* [x] Following the steps in RQA-2627, make sure runtime parameters work over USB now.
* [x] Make sure there's no discernable app freeze or hiccup that wasn't there before when you click the "Send" button, especially with a big protocol.
* [x] Make sure other requests (like toggling the lights) are unaffected.
* [x] Make sure requests over Wi-Fi are unaffected.

# Changelog

When the Electron rendering process wants to send a USB message, it needs to do so through the Electron main process. This is a problem because some JavaScript types for those messages—`File`s and `Blob`s and `FormData`s—can't be sent over IPC. We've worked around this for `POST /protocols` requests by substituting a `protocolKey: string` in place of these problematic types, in the IPC message. The main process then finds that and reconstructs the `File`s and `FormData`s from it.

That implementation didn't account for labware files external to the protocol (RESC-247), nor did it account for new multipart fields for things like run-time parameters. (RQA-2627).

With this PR, the rendering process slurps all the file contents—protocol and labware—into memory, and then sends them over IPC. We still need to use a substitute structure for the `FormData`, and the main process still needs to reconstruct all of this back into `FormData`s and `File`s. But the reconstruction code no longer needs to be concerned with application-level ideas like "labware" and "protocols" and where those things live on the filesystem, which is intended to make all of this a bit more robust.

# Review requests

See my review comments.

# Risk assessment

Medium.

### DIFF
--- a/app-shell/src/usb.ts
+++ b/app-shell/src/usb.ts
@@ -17,7 +17,7 @@ import {
   USB_DEVICE_REMOVED,
 } from './constants'
 
-import type { StructuredCloneableFormData } from '@opentrons/app/src/redux/shell/types'
+import type { IPCSafeFormData } from '@opentrons/app/src/redux/shell/types'
 import type { UsbDevice } from '@opentrons/app/src/redux/system-info/types'
 import type { PortInfo } from '@opentrons/usb-bridge/node-client'
 import type { Action, Dispatch } from './types'
@@ -82,11 +82,9 @@ function isUsbDeviceOt3(device: UsbDevice): boolean {
   )
 }
 
-function reconstructFormData(
-  structuredCloneableFormData: StructuredCloneableFormData
-): FormData {
+function reconstructFormData(ipcSafeFormData: IPCSafeFormData): FormData {
   const result = new FormData()
-  structuredCloneableFormData.forEach(entry => {
+  ipcSafeFormData.forEach(entry => {
     entry.type === 'file'
       ? result.append(entry.name, Buffer.from(entry.value), entry.filename)
       : result.append(entry.name, entry.value)
--- a/app-shell/src/usb.ts
+++ b/app-shell/src/usb.ts
@@ -84,10 +84,18 @@ function isUsbDeviceOt3(device: UsbDevice): boolean {
   )
 }
 
-interface StructuredCloneableFormDataEntry {
-  name: string
-  value: string | Blob | File
-}
+type StructuredCloneableFormDataEntry =
+  | {
+      type: 'string'
+      name: string
+      value: string
+    }
+  | {
+      type: 'file'
+      name: string
+      value: ArrayBuffer
+      filename: string
+    }
 
 type StructuredCloneableFormData = StructuredCloneableFormDataEntry[]
 
@@ -96,7 +104,9 @@ function reconstructFormData(
 ): FormData {
   const result = new FormData()
   structuredCloneableFormData.forEach(entry => {
-    result.append(entry.name, entry.value)
+    entry.type === 'file'
+      ? result.append(entry.name, Buffer.from(entry.value), entry.filename)
+      : result.append(entry.name, entry.value)
   })
   return result
 }
--- a/app-shell/src/usb.ts
+++ b/app-shell/src/usb.ts
@@ -1,8 +1,6 @@
 import { ipcMain, IpcMainInvokeEvent } from 'electron'
 import axios, { AxiosRequestConfig } from 'axios'
 import FormData from 'form-data'
-import fs from 'fs'
-import path from 'path'
 
 import {
   fetchSerialPortList,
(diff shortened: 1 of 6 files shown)


app-shell/src/usb.ts
```diff
@@ -83,6 +83,34 @@ function isUsbDeviceOt3(device: UsbDevice): boolean {
...
+      value: string
+    }
+  | {
+      type: 'file'
+      name: string
+      value: ArrayBuffer
+      filename: string
+    }
+
+type StructuredCloneableFormData = StructuredCloneableFormDataEntry[]
```
SyntaxColoring: I would love to share these type definitions between the `app` and `app-shell` projects, but I wasn't sure exactly how to organize that. Which direction should the dependency point? I see some imports at the top of this file that directly import some `app` types here into `app-shell`. Is that a good pattern to extend?

app/src/redux/shell/remote.ts
```diff
@@ -20,18 +19,55 @@ export const remote: Remote = new Proxy(emptyRemote, {
...
+): Promise<AxiosResponse<Data>> {
   const { data } = config
-  // special case: protocol files and form data cannot be sent through invoke. proxy by protocolKey and handle in app-shell
+  // Special case: FormData objects can't be sent through invoke().
+  // Convert it to a structured-cloneable object so it can be.
+  // app-shell will convert it back.
   const formDataProxy =
     data instanceof FormData
-      ? { formDataProxy: { protocolKey: data.get('key') } }
+      ? { proxiedFormData: await proxyFormData(data) }
```
SyntaxColoring, line 56: The downsides of this new approach are:

* It uses more RAM.
* It (maybe?) blocks both processes while they process the gigantic IPC message.

From [our performance stress-test cases](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3563028488/Performance+stress+test+cases), ~8MB is a decent idea of how big this message can get. https://github.com/electron/electron/issues/1948#issuecomment-864191345 suggests the IPC delay for that would be ~25ms.

If we want to improve this, @shlokamin and I think we'd want to send file *paths* across IPC instead of file contents. I think that's possible, but I got scared away from doing it in this PR by code like [this](https://github.com/Opentrons/opentrons/blob/2e1104293dd2cd6a23f2683ce0ae504b17e589d6/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx#L148) and [this](https://github.com/Opentrons/opentrons/blob/2e1104293dd2cd6a23f2683ce0ae504b17e589d6/app/src/organisms/ChooseProtocolSlideout/index.tsx#L129). It's removing the path parts of the filenames, and we'd need to get those back, but I didn't really understand all of the surrounding code and I was scared to touch it.

I can keep working on it if anyone thinks it's worthwhile.

app-shell/src/usb.ts
shlokamin: app-shell can import from app (app is a dep of app-shell), but app should NOT import from app-shell

app/src/redux/shell/remote.ts
shlokamin: if this works fine as it lets keep it — its more robust despite the performance hit

app/src/redux/shell/remote.ts
```diff
@@ -20,18 +19,55 @@ export const remote: Remote = new Proxy(emptyRemote, {
...
       : data
   const configProxy = { ...config, data: formDataProxy }
 
-  return remote.ipcRenderer.invoke('usb:request', configProxy)
+  return await remote.ipcRenderer.invoke('usb:request', configProxy)
```
shlokamin, line 60: why are we awaiting this? before this was an async non blocking call which i believe is what we want right?

shlokamin: lol woa: https://typescript-eslint.io/rules/return-await/

SyntaxColoring: If I understand correctly, it's doing the same thing as before. The difference comes from `appShellRequestor()` now being an `async function` instead of just a regular `function`.

SyntaxColoring: In other words, before, the caller of `appShellRequestor()` got a promise—the one returned by the inner `invoke()`—that resolved when the `invoke()` completed. Now, it's getting a different promise—the one created out of all the `await`s—that still resolves when the `invoke()` completes.

shlokamin: yeah i think you're right — just went down a rabbit hole with @sfoster1 exploring this lol

shlokamin (approved):
nice work, the `value.arrayBuffer()` conversion was very clever thinking

PR #14995, opened 2024-04-24
refactor(protocol-designer): tweak when module cards are disabled in …

jerader:
…createFileWizard

closes AUTH-369

# Overview

There were 2 things that the `getIsSlotAvailable` logic did not account for and it is 
1) the TC taking up 2 slots
2) the magnetic block not taking up any slots

This pr fixes that and adds test coverage

# Test Plan

I have good unit testing for this so main priority is reviewing the logic

If you want, turn on the MoaM feature flag and create a flex protocol On the "Additional Items" page, add 6 temperature modules and see that the TC is disabled since there is only 1 slot available. Add another module and see that all modules + waste chute slot cards are disabled except for the magnetic block (this is because it is placed in an inner deck slot)

# Changelog

- rename util and refactor to return number of available slots 
- add support for TC and magnetic block

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/utils.ts
+++ b/protocol-designer/src/components/modals/CreateFileWizard/utils.ts
@@ -1,4 +1,5 @@
 import {
+  MAGNETIC_BLOCK_TYPE,
   THERMOCYCLER_MODULE_TYPE,
   WASTE_CHUTE_CUTOUT,
 } from '@opentrons/shared-data'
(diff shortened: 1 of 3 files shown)


protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -85,17 +86,26 @@ export const getUnoccupiedStagingAreaSlots = (
...
+): number => {
   const moduleLength = modules != null ? Object.keys(modules).length : 0
   const additionalEquipmentLength = additionalEquipment.length
   const hasTC = Object.values(modules || {}).some(
     module => module.type === THERMOCYCLER_MODULE_TYPE
   )
+  const hasMagneticBlock = Object.values(modules || {}).some(
+    module => module.type === MAGNETIC_BLOCK_TYPE
+  )
+  let filteredModuleLength = moduleLength
```
moduleLength is not used by anything except `filteredModuleLength`

let filteredModuleLength = modules != null ? Object.keys(modules).length : 0

koji (approved):
left a comment but it is nit.
The logic looks good to me.

PR #14996, opened 2024-04-24
fix(shared-data): remove sort from choices

koji:
# Overview
remove sort from choices.

close AUTH-368

# Test Plan

# Changelog
- remove sort from choices
- update a test case

# Review requests

# Risk assessment
low


jerader (approved):
🚀

PR #14997, opened 2024-04-24
fix(app): specify component for markdown uls

sfoster1:
When these are undefined, if react-markdown tries to render them it will throw and that will crash the app. With these implemented with the same StyledText passthrough as other text elements, it won't crash anymore - though our styling rules aren't applied to the actual text in the ul.

Closes RQA-2566
Closes RQA-2587


mjhuff:
I've thrown this branch on CI to test it. For some reason, locally building isn't sufficient for reproing the initial error, and the modal renders locally without issue despite being `undefined`.

https://github.com/Opentrons/opentrons/actions/runs/8820595504

Will leave a review after that's done.

sfoster1:
> I've thrown this branch on CI to test it. For some reason, locally building isn't sufficient for reproing the initial error, and the modal renders locally without issue despite being `undefined`.
> 
> https://github.com/Opentrons/opentrons/actions/runs/8820595504
> 
> Will leave a review after that's done.

It reprod on dev mode for me, _but_ you have to be actually trying to render the elements that are used here, at least for me. So there have to be `ul`s in the release notes.

mjhuff (approved):
Ah that makes sense, thanks for explaining.

This looks good. We probably will need to follow up on this to get it matching designs (or get Ed to change the way we do the .md), but this is definitely better than the whitescreening! Thanks for fixing!

PR #14998, opened 2024-04-24
refactor(protocol-designer): minimize and extend position and tuberac…

jerader:
…k warnings

closes AUTH-370 AUTH-371

# Overview

This PR addresses 2 issues that were documents by Anothony from testing PD

1. the tuberack aspirating/dispensing too low into the well warnings were overbearing since it showed 2 warnings in one step. so this PR combines them
2. the x/y position warning banner only existed for x/y and not the z position, so this PR includes the z position as well

# Test Plan

Create a flex or ot-2 protocol and go to the deck map. Add a tube rack. Add an aspirate/dispense into the tube rack and see that a form warning shows up saying that the default aspirate/dispense height is close to the bottom of the well. NOTE: in order to trigger this warning, you need to add both an aspirate and dispense labware location and click out of the dropdown, this is the legacy behavior for all form warnings in PD where you have to click out of the field to trigger it. 

Now, adjust the tip position field and add a custom z position that = 0, see that the warning banner shows up

Screenshot 2024-04-24 at 13 36 47

Screenshot 2024-04-24 at 13 34 46

# Changelog

- add banner logic for z position
- refactor the aspirate/dispense in tuberack warning to be combined

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/steplist/formLevel/warnings.tsx
+++ b/protocol-designer/src/steplist/formLevel/warnings.tsx
@@ -2,18 +2,18 @@ import * as React from 'react'
 import { getWellTotalVolume } from '@opentrons/shared-data'
 import { KnowledgeBaseLink } from '../../components/KnowledgeBaseLink'
 import type { FormError } from './errors'
+
 /*******************
  ** Warning Messages **
  ********************/
 
 export type FormWarningType =
-  | 'ASPIRATE_TIP_POSITIONED_LOW_IN_TUBE'
   | 'BELOW_MIN_AIR_GAP_VOLUME'
   | 'BELOW_MIN_DISPOSAL_VOLUME'
   | 'BELOW_PIPETTE_MINIMUM_VOLUME'
-  | 'DISPENSE_TIP_POSITIONED_LOW_IN_TUBE'
   | 'OVER_MAX_WELL_VOLUME'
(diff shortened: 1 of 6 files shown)


protocol-designer/src/steplist/formLevel/warnings.tsx
```diff
@@ -88,34 +81,38 @@ export type WarningChecker = (val: unknown) => FormWarning | null
...
-    ? dispenseTipPositionedLowInTube()
-    : null
+
+  if (isAspirateTubeRack && aspirate_mmFromBottom === null) {
+    return tipPositionedLowInTube()
+  } else if (isDispenseTubeRack && dispense_mmFromBottom === null) {
+    return tipPositionedLowInTube()
+  } else {
+    return null
+  }
```
koji:
```suggestion
  if ((isAspirateTubeRack && aspirate_mmFromBottom === null) || (isDispenseTubeRack && dispense_mmFromBottom === null)) {
    return tipPositionedLowInTube()
  } else {
    return null
  }
```

koji:
nit
seems that the padding is not needed?
Screenshot 2024-04-24 at 4 56 41 PM

koji (approved):
left a couple of comments, but the sandbox worked as expected.
the changes look good to me.

PR #14999, opened 2024-04-24
fix(api): disengage Z motors during 96-channel pipettes bracket detach/attach flows

ahiuchingau:
# Overview
This PR makes sure we are engaging ebrakes when we're attaching/detaching the 96-channel pipette and mount bracket.
On top of that, we can now use retract instead of using a movement Point just to move the z up. 
Fixes [RQA-2631](https://opentrons.atlassian.net/browse/RQA-2631).

[RQA-2631]: https://opentrons.atlassian.net/browse/RQA-2631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Nice!

PR #15000, opened 2024-04-24
fix(app): add robot serial number to BuildrootAnalyticsData

koji:
# Overview
add robot serial number to `BuildrootAnalyticsData` and modify the condition to send existing data.
If robot is not null, sending `currentSystem`, `currnetVersion`, and `robotSerialNumber` would be good even if `updateVersion` is null since those properties are reflected the current user's robot situation.

close RQA-2386

# Test Plan
- make -C app dev OT_APP_MIXPANEL_ID=project_token 
project_token can be found in Project Settings page
- update robot system from robot settings page

# Changelog

# Review requests
 

# Risk assessment


app/src/redux/analytics/selectors.ts
```diff
@@ -32,16 +32,20 @@ export function getBuildrootAnalyticsData(
       ? getRobotUpdateRobot(state)
       : getViewableRobots(state).find(r => r.name === robotName) ?? null
 
-  if (updateVersion === null || robot === null) return null
```
koji, line 35: `updateVersion === null` is causing the following issue.
The app sends events that are related to `robotUpdate` but most cases cannot send data(currentVersion, currentSystem, updateVersion).

Screenshot 2024-04-24 at 4 21 29 PM

Screenshot 2024-04-24 at 4 22 17 PM

app/src/redux/analytics/constants.ts
```diff
@@ -45,3 +45,10 @@ export const ANALYTICS_ODD_APP_ERROR = 'oddError'
 export const ANALYTICS_NOTIFICATION_PORT_BLOCK_ERROR =
   'notificationPortBlockError'
 export const ANALYTICS_DESKTOP_APP_ERROR = 'desktopAppError'
+
+export const ANALYTICS_ROBOT_UPDATE_VIEW = 'robotUpdateView'
+export const ANALYTICS_ROBOT_UPDATE_CHANGE_LOG_VIEW = 'robotUpdateChangeLogView'
+export const ANALYTICS_ROBOT_UPDATE_IGNORE = 'robotUpdateIgnore'
+export const ANALYTICS_ROBOT_UPDATE_INITIATE = 'robotUpdateInitiate'
+export const ANALYTICS_ROBOT_UPDATE_ERROR = 'robotUpdateError'
+export const ANALYTICS_ROBOT_UPDATE_COMPLETE = 'robotUpdateComplete'
```
mjhuff: Nit, I think we could benefit from doing an object literal here like the following:

```suggestion
export const ANALYTICS_ROBOT_UPDATE_VIEW = 'robotUpdateView'
export const ANALYTICS_ROBOT_UPDATE_CHANGE_LOG_VIEW = 'robotUpdateChangeLogView'

const ANALYTICS_STATE_ROBOT_UPDATE = {
  IGNORE: 'robotUpdateIgnore',
  INITIATE: 'robotUpdateInitiate',
  ERROR: 'robotUpdateError',
  COMPLETE: 'robotUpdateComplete',
} as const;
```
This gives us the added benefit of not importing the entire `constants` module and enforces a clear relationship between these constants with the end goal of encouraging devs to use the other states if they're attempting to use only one.

mjhuff (approved):
One nit that's your discretion. This makes sense to me. Thank you!

app/src/redux/analytics/selectors.ts
mjhuff: Yeah, the way you're handling this seems better!

app/src/redux/analytics/constants.ts
koji: thank you @mjhuff, that looks good.
I will update them.

koji: @mjhuff 

Do you think we should do the same change to run events?

```ts
export const ANALYTICS_PROTOCOL_RUN_AGAIN = 'runAgain'
export const ANALYTICS_PROTOCOL_RUN_FINISH = 'runFinish'
export const ANALYTICS_PROTOCOL_RUN_PAUSE = 'runPause'
export const ANALYTICS_PROTOCOL_RUN_START = 'runStart'
export const ANALYTICS_PROTOCOL_RUN_RESUME = 'runResume'
export const ANALYTICS_PROTOCOL_RUN_CANCEL = 'runCancel'

const ANALYTICS_PROTOCOL_RUN_ACTION = {
  AGAIN: 'runAgain',
  FINISH: 'runFinish',
  PAUSE: 'runPause',
  START: 'runStart',
  RESUME: 'runResume',
  CANCEL: 'runCancel',
} as const


```

mjhuff: Good catch - yes!

PR #15001, opened 2024-04-24
fix(hardware): add hepa uv to tool detector

caila-marashaj:
Right now, the hepa/uv device is sending responses after an `InstrumentInfoRequest`, but the tool detector code isn't set up to parse these messages, meaning that the queue doesn't get properly emptied, and the instrument info responses keep coming in

### DIFF
--- a/api/tests/opentrons/hardware_control/backends/test_ot3_subsystem_manager.py
+++ b/api/tests/opentrons/hardware_control/backends/test_ot3_subsystem_manager.py
@@ -226,12 +226,13 @@ class ToolDetectionController:
             if device.subidentifier != 0
             else default_name
         )
-        return tools.types.PipetteInformation(
-            pipette_name.name,
-            f"dummyserial{pipette_name.name}",
-            pipette_name,
-            pipette_name.value,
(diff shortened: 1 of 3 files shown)


api/tests/opentrons/hardware_control/backends/test_ot3_subsystem_manager.py
```diff
@@ -227,10 +227,10 @@ def _pipette_info_from_network(
             else default_name
         )
         return tools.types.PipetteInformation(
-            pipette_name,
-            pipette_name.value,
             pipette_name.name,
             f"dummyserial{pipette_name.name}",
+            pipette_name,
```
sfoster1: oof. can we change this to use kwargs like `pipette_name=` instead of positional?

sfoster1 (approved):
Nice, but maybe use kw args

vegano1 (approved):
The changes look good on my end,

I just tested this on Flexen and the HepaUVInfoResponse is being handled correctly.
I checked the robot server GET `/instruments` routes and we are NOT exposing the Hepa/UV as expected.

Thank you!

PR #15002, opened 2024-04-24
feat(shared-data, api): return latest pipette version from pipetteName

jerader:
closes AUTH-357

# Overview

Previously, if you have access to only the `pipetteName`, the pipette definitions you get back are the first pipette version for the generation. For example, if you have a `p1000_single_flex`, you would get the pipette definitions for `3_0`. After speaking with Seth, Andy, Ed, and Sanniti, we decided that this is not great because 1) version `.0`s are prototypes and aren't in the wild and 2) the latest versions have the most up to date specs for things such as `flowRate`s. ([slack](https://opentrons.slack.com/archives/C06M6R72UDS/p1713454131088679) conversation about this if interested)

We thought about making the prototype versions the "perfect definitions" but that requires maintenance and it is much easier in the long run just to return the latest version that exists.

This change is already implemented in the frontend (see `getPipetteV2Specs`), so Protocol Designer (et al) will have the latest flow rates 🚀 .

But, we needed to align in the api as well so `opentrons-simulate` and analysis also have the latest data. This PR updates the `version_from_generation` def in `shared-data` to return the latest version. 

Note: the downfall to the current approach is there are a few api tests that will need to be updated whenever the specific pipette has a new version - see the PR for which tests those are. 

# Test Plan

upload this to the app on edge and look at the run log - the aspirate/dispense flow rate should be 160 ul/sec, which is the 1000uL tip default flow rate for the `3_0` model.

then upload this to the app on this branch and look at the run log  - the aspirate/dispense flow rate should be 716 ul/sec, which is the 1000uL tip default flow rate for the `3_6` model.

```
requirements = {
	"robotType": "Flex",
	"apiLevel": "2.16"
}

def run(ctx):
  tip_rack2 = ctx.load_labware("opentrons_flex_96_tiprack_1000ul", "D3")
  instrument = ctx.load_instrument('flex_1channel_1000', tip_racks=[tip_rack2], mount="left")
  
  my_pcr_plate = ctx.load_labware('nest_96_wellplate_200ul_flat', "C2")
  my_other_plate = ctx.load_labware('nest_96_wellplate_200ul_flat', "C1")
  ctx.load_trash_bin("A3")
    
  instrument.pick_up_tip()  
  instrument.dispense(20, my_other_plate.wells_by_name()["A2"])

  instrument.drop_tip()
```

# Changelog

- refactor `version_from_generation`
- update the `test_convert_pipette_name`
- update all affected tests that now return the latest pipette version from the `pipetteName`
- update the low volume default max flow rate for the `p50_single_3.6` pipette definition to match the ui max flow rate (see https://github.com/Opentrons/opentrons/pull/14859 for details on why)

# Review requests

Please review the code and logic and let me know if there is anything else I need to change?

# Risk assessment

low


shared-data/pipette/definitions/2/liquid/single_channel/p50/lowVolumeDefault/3_6.json
```diff
@@ -5,15 +5,15 @@
       "uiMaxFlowRate": 26.7,
       "defaultAspirateFlowRate": {
         "default": 35,
-        "valuesByApiLevel": { "2.14": 35 }
+        "valuesByApiLevel": { "2.14": 26.7 }
```
jerader, line 8: i updated these to match the `uiMaxFlowRate` - see https://github.com/Opentrons/opentrons/pull/14859 for more details

ecormany (commented):
It worrrrrrrrrks! 🎉 
image

Users ask about this all the time — so glad this will be accurate now.

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
```diff
@@ -106,26 +109,85 @@ def version_from_string(version: str) -> PipetteVersionType:
...
+        or "96" in pipette_name_list
+    ):
+        return 3
+    elif "gen2" in pipette_name_list:
+        return 2
+    else:
+        return 1
+
+
 def version_from_generation(pipette_name_list: List[str]) -> PipetteVersionType:
```
sfoster1: let's drop an `lru_cache` on this, it's a perfect candidate:
- it will often be called with the same thing when something tries to load data
- it is pretty expensive to call since it does a whole ton of filesystem interactions

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
```diff
@@ -106,26 +109,85 @@ def version_from_string(version: str) -> PipetteVersionType:
...
+        pipette_name_list
+    )
+    model_from_pipette_name = pipette_name_list[0]
+    channel_from_pipette_name = get_channel_from_pipette_name(pipette_name_list)
+
+    paths_to_validate = (
+        get_shared_data_root() / "pipette" / "definitions" / "2" / "general"
+    )
+
+    highest_minor_version: Literal[0, 1, 2, 3, 4, 5, 6] = 0
```
sfoster1: we'd have to update this whenever we add a new minor version, right? can we weaken its type to `int`?

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
```diff
@@ -106,26 +109,85 @@ def version_from_string(version: str) -> PipetteVersionType:
...
+                    minor_version_int = int(minor_version)
+                    minor_version_lit: PipetteModelMinorVersionType = cast(
+                        PipetteModelMinorVersionType, minor_version_int
+                    )
+
+                    # Update the highest minor version if this version is higher
+                    if highest_minor_version < minor_version_lit:
+                        highest_minor_version = minor_version_lit
+
+    if highest_minor_version == 0:
```
sfoster1: i don't think we need this, right? `0` is a valid value there, and the major version is the same, so we can unconditionally `return PipetteVersionType(major_version_from_pipette_name, highest_minor_version)` and it's fine if `highest_minor_version` is 0

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
```diff
@@ -106,26 +109,85 @@ def version_from_string(version: str) -> PipetteVersionType:
...
+    model_from_pipette_name = pipette_name_list[0]
+    channel_from_pipette_name = get_channel_from_pipette_name(pipette_name_list)
+
+    paths_to_validate = (
+        get_shared_data_root() / "pipette" / "definitions" / "2" / "general"
+    )
+
+    highest_minor_version: Literal[0, 1, 2, 3, 4, 5, 6] = 0
+
+    for channel_dir in os.listdir(paths_to_validate):
```
sfoster1: The outer two `for` loops don't seem to be doing much, I think. we're iterating through them and checking for exact equality to the values we already have, so let's just construct the path at the beginning:
```
versions_path = (paths_to_validate
             / channel_from_pipette_name
             / model_from_pipette_name)

for version_file in version_paths.iterdir():
   ...
```
[Path.iterdir](https://docs.python.org/3/library/pathlib.html#pathlib.Path.iterdir)

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
```diff
@@ -106,26 +109,85 @@ def version_from_string(version: str) -> PipetteVersionType:
...
+        if channel_dir != channel_from_pipette_name:
+            continue
+
+        for model_dir in os.listdir(paths_to_validate / channel_dir):
+            if model_dir != model_from_pipette_name:
+                continue
+
+            for version_file in os.listdir(paths_to_validate / channel_dir / model_dir):
+                version_list = version_file.split(".json")[0].split("_")
```
sfoster1:
```suggestion
                version_list = version_file.stem.split("_")
```
[PurePath.stem](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem)

(as long as we're using e.g. `(paths_to_validate/channel_dir/model_dir).iterdir()` rather than `os.listdir` so we get path objects)

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
```diff
@@ -106,26 +109,85 @@ def version_from_string(version: str) -> PipetteVersionType:
...
+                major_version = version_list[0]
+                minor_version = version_list[1]
+
+                # Check if the major version matches the expected major version
+                if major_version == str(major_version_from_pipette_name):
```
sfoster1: maybe invert logic and continue to save some indentation, i.e.
```
if major_version != str(major_version_from_pipette_name):
    continue
```

sfoster1 (changes requested):
Some inline things, I think we can rework that main function a bit.

It's weird that the test asks for a kind of pipette that does not exist... maybe we can add a couple more cases to it?

shared-data/python/opentrons_shared_data/pipette/pipette_load_name_conversions.py
jerader: yeah, so the return type of `version_from_generation` is `PipetteVersionType`, which is a Literal. I guess i could weaken it so the return type is just an `int`. i agree it should be weakened so that we don't have to always update it

jerader: nevermind, this is giving me a hard time updating because all the other utils in the file use `PipetteVersionType` as well 🤔 maybe it can be refactored in a followup? Plus there is this comment in `types.py`

```
# TODO Literals are only good for writing down
# exact values. Is there a better typing mechanism
# so we don't need to keep track of versions in two
# different places?
PipetteModelMajorVersionType = Literal[1, 2, 3]
PipetteModelMinorVersionType = Literal[0, 1, 2, 3, 4, 5, 6]
```

sfoster1: yeah, a followup is fine i suppose. could we maybe have this be `highest_minor_version: PipetteModelMinorVersionType = 0` then? that way we wouldn't have to change it in two places

sfoster1 (approved):
Looks great!

PR #15003, opened 2024-04-24
docs(api): update descriptions of partial tip pickup behavior for 7.3

ecormany:
# Overview

Update partial tip pickup docs to reflect the latest behavior, as added in #15009 and to be released in robot software 7.3.

Addresses RTC-440.

# Test Plan

[Sandbox](http://sandbox.docs.opentrons.com/docs-partial-pickup-channel-1/v2/)

# Changelog

- Removed "Column 12" and "Column 1" headers from Partial Tip Pickup page. Removed tip-tracking code for Column 1 as no longer needed..
- Reworded note in `configure_nozzle_layout` API reference entry.

# Review requests

Already spoke with @CaseyBatten to get an idea of what needed to be included here and what will go in separate documentation.

- <s>Should we actually put this straight into `edge` and deploy ASAP, or can it wait a month and roll out with 2.18 docs (current plan)?</s> PAPI 2.18 deploy will be timed with robot stack 7.3 release, which this aligns with.

# Risk assessment

none

### DIFF
--- a/api/docs/v2/pipettes/partial_tip_pickup.rst
+++ b/api/docs/v2/pipettes/partial_tip_pickup.rst
@@ -23,7 +23,7 @@ For greater convenience, also import the individual layout constants that you pl
 
     from opentrons.protocol_api import COLUMN, ALL
 
-Then when you call ``configure_nozzle_layout`` later in your protocol, you can set ``style=COLUMN``. 
+Then when you call ``configure_nozzle_layout`` later in your protocol, you can set ``style=COLUMN``.
 
 Here is the start of a protocol that performs both imports, loads a 96-channel pipette, and sets it to pick up a single column of tips.
 
@@ -106,6 +106,10 @@ When switching between full and partial pickup, you may want to organize your ti
     partial_tip_racks = [tips_1, tips_2]
     full_tip_racks = [tips_3, tips_4]
 
+.. Tip::
+
+    It's also good practice to keep separate lists of tip racks when using multiple partial tip pickup configurations (i.e., using both column 1 and column 12 in the same protocol). This improves positional accuracy when picking up tips. Additionally, use Labware Position Check in the Opentrons App to ensure that the partial configuration is well-aligned to the rack.
+
 Now, when you configure the nozzle layout, you can reference the appropriate list as the value of ``tip_racks``::
 
     pipette.configure_nozzle_layout(
@@ -120,7 +124,7 @@ Now, when you configure the nozzle layout, you can reference the appropriate lis
         tip_racks=full_tip_racks
     )
     pipette.pick_up_tip()  # picks up full rack in C1
-    
+
 This keeps tip tracking consistent across each type of pickup. And it reduces the risk of errors due to the incorrect presence or absence of a tip rack adapter.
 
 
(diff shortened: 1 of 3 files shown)


api/docs/v2/pipettes/partial_tip_pickup.rst
```diff
@@ -164,26 +159,10 @@ You would get a similar error trying to aspirate from or dispense into a well pl
...
-    pipette.configure_nozzle_layout(style=COLUMN, start="A1")
-    row_a = tip_rack.rows()[0]
-    pipette.pick_up_tip(row_a.pop())  # pick up A12-H12
-    pipette.drop_tip()
-    pipette.pick_up_tip(row_a.pop())  # pick up A11-H11
-    pipette.drop_tip()
+If your application can't accommodate a deck layout that works well with column 12, you can configure the 96-channel pipette to pick up tips with column 1, but this is not recommended. Column 1 pickup has several drawbacks:
 
-This code first constructs a list of all the wells in row A of the tip rack. Then, when picking up a tip, instead of referencing one of those wells directly, the ``location`` is set to ``row_a.pop()``. This uses the `built-in pop method <https://docs.python.org/3/tutorial/datastructures.html#more-on-lists>`_ to get the last item from the list and remove it from the list. If you keep using this approach to pick up tips, you'll get an error once the tip rack is empty — not from the API, but from Python itself, since you're trying to ``pop`` an item from an empty list.
+- Tip pickup force may be incorrect, leading to poor tip fit and loss of positional accuracy.
```
CaseyBatten: The decision was made to not update the pick up tip current values, so technically the force used on A1 is no worse than the force used on A12 currently (they're equally "wrong").

api/docs/v2/pipettes/partial_tip_pickup.rst
```diff
@@ -164,26 +159,10 @@ You would get a similar error trying to aspirate from or dispense into a well pl
...
+- Automatic tip tracking is not available
```
CaseyBatten: This has been untrue since 7.2, we have dynamic automatic tip tracking for all possible configurations now.

CaseyBatten (commented):
Good so far, but some feedback on inaccuracies below.

api/docs/v2/pipettes/partial_tip_pickup.rst
```diff
@@ -106,6 +106,10 @@ When switching between full and partial pickup, you may want to organize your ti
     partial_tip_racks = [tips_1, tips_2]
     full_tip_racks = [tips_3, tips_4]
 
+.. Tip::
+
+    It's also good practice to keep separate lists of tip racks when using multiple partial tip pickup configurations (i.e., using both column 1 and column 12 in the same protocol). This improves positional accuracy when picking up tips. Additionally, use Labware Position Check in the Opentrons App to ensure that the partial configuration is well-aligned to the rack.
```
CaseyBatten, line 111: This looks good and communicates what we want. Regarding the last sentence, I think we need to confirm what exactly LPC looks like for partial configuration (I personally have never encountered it) and communicate that if possible.

CaseyBatten (approved):
These changes look great and remove any ambiguity we had during the transition from 7.1 to 7.2.

api/docs/v2/pipettes/partial_tip_pickup.rst
CaseyBatten: Just spoke with Sanniti, looks like LPC for partial tips is just standard LPC. You'd account for the tip overlap offset during said LPC, then presumably only use the partial configuration you LPC'd for on that piece of labware. So I think this documentation is acceptable on that front, we don't want to over-explain and end up causing confusion once this problem is addressed in full.

PR #15005, opened 2024-04-24
Correct HS Time Calculation and Print Statements

rclarke0:
# Overview

Concise print statements and HS on time calculation correction.

# Test Plan

Tested on ABR robots.

# Changelog

Previously, the heater shaker temperature on time was linked to the shaker on time. This has been corrected to reference the correct command string. If the heatershaker is not deactivated, the on time is calculated with the protocol end time stamp.
Changed print statements for get robot logs and abr_google_drive to make it more obvious to the user if there is an error or not.

# Review requests

# Risk assessment


sfoster1 (approved):
Nice!

PR #15009, opened 2024-04-25
fix(api): Raise cases for unsupported nozzle layouts

CaseyBatten:
# Overview
See RTC-440

This PR seeks to limit user access to unapproved configurations through the PAPI and raise errors when configuring potentially unsafe layouts for the nozzle map.

UPDATE:
After discussion, doc side changes and API docstrings have been moved to alternative PR.

# Test Plan
Ensure that protocols with that attempt to utilize unsupported single/row/quadrants API commands are rejected. Ensure that if engine commands are used, no more than 24 tips can be added to partial configuration.

# Changelog

- Added disabled layouts filter to PAPI to prevent configuration in unapproved layouts (these layouts can still be achieved through backend commands, which is desired for hardware testing or advanced users)
- Limit the size of configurations that are not `FULL` to 24 nozzles, a size limit found by hardware testing.

# Review requests

# Risk assessment
Relatively low as we were not exposing these configurations publicly yet anyways, however some use cases involving non-public facing API may be effected.

### DIFF
--- a/api/docs/v2/pipettes/partial_tip_pickup.rst
+++ b/api/docs/v2/pipettes/partial_tip_pickup.rst
@@ -25,10 +25,6 @@ For greater convenience, also import the individual layout constants that you pl
 
 Then when you call ``configure_nozzle_layout`` later in your protocol, you can set ``style=COLUMN``. 
 
-It is important to note that for versions <= 7.3, when configuring for COLUMN layout, there may be a noticeable tip overlap offset that will need to be accounted for through Labware Position Check.
-
-Along that same line of logic, it would be advisable to determine a configuration to be used for a specific labware, and only interact with that labware when in said configuration.
-
 Here is the start of a protocol that performs both imports, loads a 96-channel pipette, and sets it to pick up a single column of tips.
 
 .. code-block:: python
--- a/api/docs/v2/pipettes/partial_tip_pickup.rst
+++ b/api/docs/v2/pipettes/partial_tip_pickup.rst
@@ -25,6 +25,10 @@ For greater convenience, also import the individual layout constants that you pl
 
 Then when you call ``configure_nozzle_layout`` later in your protocol, you can set ``style=COLUMN``. 
 
+It is important to note that for versions <= 7.3, when configuring for COLUMN layout, there may be a noticeable tip overlap offset that will need to be accounted for through Labware Position Check.
+
+Along that same line of logic, it would be advisable to determine a configuration to be used for a specific labware, and only interact with that labware when in said configuration.
+
 Here is the start of a protocol that performs both imports, loads a 96-channel pipette, and sets it to pick up a single column of tips.
 
 .. code-block:: python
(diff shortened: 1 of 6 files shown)


api/docs/v2/pipettes/partial_tip_pickup.rst
```diff
@@ -25,6 +25,10 @@ For greater convenience, also import the individual layout constants that you pl
 
 Then when you call ``configure_nozzle_layout`` later in your protocol, you can set ``style=COLUMN``. 
 
+It is important to note that for versions <= 7.3, when configuring for COLUMN layout, there may be a noticeable tip overlap offset that will need to be accounted for through Labware Position Check.
```
ecormany: If there isn't an API change here, I think this is best addressed in the 7.3 release notes, probably under the "Improved features" header.

api/docs/v2/pipettes/partial_tip_pickup.rst
```diff
@@ -25,6 +25,10 @@ For greater convenience, also import the individual layout constants that you pl
...
+
+Along that same line of logic, it would be advisable to determine a configuration to be used for a specific labware, and only interact with that labware when in said configuration.
```
ecormany: I'm not sure what this means.

api/docs/v2/pipettes/partial_tip_pickup.rst
ecormany: Another possible location is as a note in the API Reference entry. Regardless, I think I'd want to keep it out of the main description of the feature in this article — especially if we've improved the behavior to the point that we don't have to be issuing warnings about how the latest version works.

api/src/opentrons/hardware_control/nozzle_manager.py
```diff
@@ -267,6 +269,17 @@ def build(
...
 
+        if (
+            NozzleConfigurationType.determine_nozzle_configuration(
+                physical_rows, rows, physical_columns, columns
+            )
+            != NozzleConfigurationType.FULL
+        ):
+            if len(rows) * len(columns) > MAXIMUM_NOZZLE_COUNT:
+                raise IncompatibleNozzleConfiguration(
+                    f"Partial Nozzle Layouts may not be configured to contain more than {MAXIMUM_NOZZLE_COUNT} channels."
```
ecormany, line 280: copy suggestion
```suggestion
                    f"Partial nozzle layouts can contain at most {MAXIMUM_NOZZLE_COUNT} channels."
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1947,6 +1947,20 @@ def configure_nozzle_layout(
...
+        #       have appropriate data regarding tip-type to map current values added to the
+        #       pipette definitions.
+        disabled_layouts = [
+            NozzleLayout.ROW,
+            NozzleLayout.SINGLE,
+            NozzleLayout.QUADRANT,
+        ]
+        if style in disabled_layouts:
+            raise ValueError(
+                f"Nozzle layout configuration of style {style.value} is currently unsupported."
```
ecormany, line 1954: copy suggestion
```suggestion
                f"{style.value} nozzle layouts are currently not supported."
```

ecormany (commented):
Thanks for the draft doc suggestions! I think we should communicate almost all of this info to users in one venue or another. I'll set up a time for us to discuss the best placement and wording.

vegano1 (approved):
changes look fine to me.

PR #15010, opened 2024-04-25
chore(mono): Enforce notification hooks in the app

mjhuff:
Closes [EXEC-419](https://opentrons.atlassian.net/browse/EXEC-419)

# Overview

Enforces the usage of the notification wrapper for the equivalent HTTP hook if the wrapper exists. We need this, so we don't accidentally poll in the app/ODD when we can use notifications instead.

I went with the lint approach over the "global mock and fail test" approach for a couple reasons:

1. We don't have 100% test coverage, and making sure we find a way to enforce this across the codebase is important.
2. It's nice having the immediate IDE feedback that the import is wrong instead of running into the error with the CI suite.

Also, I've intentionally excluded test files from the lint rule for a few reasons:

1. Test files are for devs, not the end user.
2. A lot of times, we import all of `react-api-client` in a test file, and the lint rule will catch these spots even when the rule doesn't apply.
3. The reason hooks are imported in the test file is to mock them out from the respective file they are test. Enforcing lint on the test files is redundant. 

# Test Plan
- In any file within the app, try importing any vanilla HTTP that has a `useNotify` equivalent, such as `useCurrentMaintenanceRun`. Verify that lint fails.
- Alternatively, look at the first commit CI failure errors.

# Risk assessment
low

[EXEC-419]: https://opentrons.atlassian.net/browse/EXEC-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


koji (approved):
This is awesome!
eslint shows the error with the message.

PR #15011, opened 2024-04-25
fix(app): check /instruments for 96 channel attachment

smb2268:
Fix RQA-2633

# Overview
I disabled the `/pipettes` query for Flex in favor of using `/instruments` but missed switching the reference for whether a 96 channel is attached to this data. 

# Test Plan
Look at the desktop app with a 96 channel attached, no Right Pipette card should be visible

# Changelog
Check left pipette data from `/instruments` to see if it is a 96 channel

# Review requests
Look over changes

# Risk assessment
Low


mjhuff (approved):
lgtm!

koji (approved):
the changes look good to me.
I don't see any empty pipette card anymore!

PR #15013, opened 2024-04-25
fix(app): liquid detail fixed precision

sfoster1:
There were a couple places remaining where we presented liquid volumes but didn't have fixed precision, so we could have a pretty excessive number of decimals. Now those are all toFixed(1)'d.

Also, there was an extra "total volume" label in some little grey-background-roundrect text objs in the liquid detail map view that would cause the background to clip behind the map. the label's not in the designs, so I pulled it.

Closes RSQ-78

#### new:
Screenshot 2024-04-25 at 2 24 42 PM
#### old:
Screenshot 2024-04-25 at 2 27 48 PM

#### new:
Screenshot 2024-04-25 at 2 26 06 PM
#### old:
Screenshot 2024-04-25 at 2 28 25 PM

#### new:
Screenshot 2024-04-25 at 2 26 54 PM

#### old:

Screenshot 2024-04-25 at 2 29 54 PM

Closes RSQ-78


koji (approved):
lgtm. 
thank you for fixing this.

PR #15016, opened 2024-04-25
fix(app): add grey square for stale and error protocol map

smb2268:
fix RQA-2634

# Overview
Implement designs for stale protocol state deck map - now a grey square with rounded edges

# Test Plan
Look at choose a protocol slideout and desktop protocols tab to view failed and stale protocols
Screen Shot 2024-04-30 at 1 44 18 PM
Screen Shot 2024-04-30 at 1 43 09 PM
Screen Shot 2024-04-30 at 1 43 04 PM

# Changelog
Updated Choose a Protocol Slideout and Protocol Card to show grey square (rounded edges)

# Review requests
Test this out

# Risk assessment
Low

### DIFF
--- a/app/src/organisms/ChooseProtocolSlideout/index.tsx
+++ b/app/src/organisms/ChooseProtocolSlideout/index.tsx
@@ -526,19 +526,35 @@ function StoredProtocolList(props: StoredProtocolListProps): JSX.Element {
                 onClick={() => handleSelectProtocol(storedProtocol)}
               >
                 <Box display="grid" gridTemplateColumns="1fr 3fr">
-                  <Box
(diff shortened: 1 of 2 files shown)


ecormany:
These kind of look like close or delete buttons. Like, I would expect if I clicked on the "x" the stale protocol would be removed from the list. (I know this isn't the case.)

smb2268:
> These kind of look like close or delete buttons. Like, I would expect if I clicked on the "x" the stale protocol would be removed from the list. (I know this isn't the case.)

@mmencarelli I've gotten this feedback from a few devs as well - do we still want to move forward with the x icon?

brenthagen (approved):
code LGTM !

PR #15017, opened 2024-04-25
fix(app): fix restore default values button behavior

koji:
# Overview
currently restore default values button is tappable when tapping confirm values button.
This PR makes restore default values button is disabled during confirming values.

close RQA-2639

# Test Plan
set up a RTP protocol, change a runtime parameter and tap confirm values.
then try to tap restore default values button and check that the restore default values modal doesn't show up.

# Changelog
- add disabled to secondary button in protocol setup parameters
- add a test case for this 

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ProtocolSetupParameters/index.tsx
+++ b/app/src/organisms/ProtocolSetupParameters/index.tsx
@@ -151,6 +151,7 @@ export function ProtocolSetupParameters({
         secondaryButtonProps={{
           buttonType: 'tertiaryLowLight',
           buttonText: t('restore_defaults'),
+          disabled: isLoading || startSetup,
           onClick: () => showResetValuesModal(true),
         }}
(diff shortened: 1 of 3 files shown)


app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -151,6 +151,7 @@ export function ProtocolSetupParameters({
         secondaryButtonProps={{
           buttonType: 'tertiaryLowLight',
           buttonText: t('restore_defaults'),
+          disabled: isLoading || startSetup,
```
jerader, line 154: random nit but line 138 could be a `console.error` and have a more descriptive log, such as "should not hit here, this is a bad param"

jerader (approved):
lgtm!

PR #15018, opened 2024-04-25
fix(app,api,robot-server): remove old sim flag

sfoster1:
Remove the disableFastProtocolUpload flag.

This flag makes the simulation stack, when simulating older protocols, use a hardware controller built in simulation mode and a full standard protocol context and/or engine. This is... not something we've significantly tested and has a lot of surprising problems, and at this point it doesn't bring much value. So remove the feature flag.

The old behavior can still be accessed via opentrons_simulate, which will only use the simulating hardware controller when simulating pre-engine python and json protocols.

Closes EXEC-360

## To come out of draft
- is this a good idea?


DerekMaggio (approved):
It makes sense to me to remove this. 
If customers are relying on it because the fast protocol upload stuff fails, then we should be fixing fast protocol upload stuff, not exposing unmaintained and basically deprecated functionality. 

The only caveat is, have we checked with support to see if they have an idea of if there is a large amount of customers this will cause issues for?

All the Python side of things look good to me

TamarZanzouri (approved):
not sure about the FE changes but backend looks great

app/src/organisms/Devices/RobotSettings/AdvancedTab/UseOlderProtocol.tsx
```diff
@@ -1,69 +0,0 @@
...
-          id="AdvancedSettings_showLink"
-        >
-          {t('use_older_protocol_analysis_method')}
-        </StyledText>
-        <StyledText as="p">
-          {t('branded:use_older_protocol_analysis_method_description')}
-        </StyledText>
-      </Box>
-      <ToggleButton
-        label="use_older_protocol_analysis_method"
```
SyntaxColoring, line 61: We can delete these translation strings, too: `use_older_protocol_analysis_method` and `use_older_protocl_analysis_method_description` in assets/localization/en/anonymous.json, .../branded.json, and .../device_settings.json.

SyntaxColoring (approved):
(rhythmic chanting) red diff red diff red diff

Frontend and backend both look good to me.

caila-marashaj (approved):
Seems like this makes sense

PR #15020, opened 2024-04-26
fix(api): skip command key hash generation only if command intent is SETUP

sanni-t:
Fixes RQA-2640

# Overview

In a previous PR we made a change from skipping hash generation if `CommandCreate.intent == CommandIntent.SETUP` to if `CommandCreate.intent != CommandIntent.PROTOCOL`. But a command can have a null intent when it is a protocol command; we only expect setup commands to have an explicit SETUP intent. So we were accidentally skipping key hash generation for python protocols, which was resulting in the app not being able to match run commands with their analysis counterparts.

This PR fixes that and also adds an integration test so we don't accidentally break this functionality again.

# Review requests

- want to make sure that there wasn't an important reason for making the original change that I'm missing. But since all tests are still passing I'm guessing there wasn't any functional change that required a change in the original `if` statement.

# Risk assessment

Low. Bug fix.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/hash_command_params.py
+++ b/api/src/opentrons/protocol_engine/commands/hash_command_params.py
@@ -28,7 +28,7 @@ def hash_protocol_command_params(
         The command hash, if the command is a protocol command.
         `None` if the command is a setup command.
     """
-    if create.intent != CommandIntent.PROTOCOL:
+    if create.intent == CommandIntent.SETUP:
         return None
     # We avoid Python's built-in hash() function because it's not stable across
     # runs of the Python interpreter. (Jira RSS-215.)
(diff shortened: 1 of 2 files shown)


y3rsh (approved):
Good test.  
I verified this works on the robot.
ODD and app steps now displaying commands as expected.
image

I made a copy of this to cut an alpha
#15022

sfoster1 (approved):
oh, great. in a followup we should make this always be present I think

api/src/opentrons/protocol_engine/commands/hash_command_params.py
```diff
@@ -28,7 +28,7 @@ def hash_protocol_command_params(
         The command hash, if the command is a protocol command.
         `None` if the command is a setup command.
     """
-    if create.intent != CommandIntent.PROTOCOL:
+    if create.intent == CommandIntent.SETUP:
```
TamarZanzouri, line 31: @sfoster1 this should check for FIXIT commands as well

TamarZanzouri, line 31:
```suggestion
    if create.intent in [CommandIntent.SETUP, CommandIntent.FIXIT]:
```

PR #15023, opened 2024-04-26
fix(docs): fix release notes

y3rsh:
## Keep experimenting with the formatting of the release notes
- `<br>` was no good
- trying headers


DerekMaggio (approved):
> Everything looks good to me. Is the random link at the end supposed to be there? It produces a diff with no changes

Talked to Josh, the tag needs to be created for it to kick off a release for the link to show any changes So 100% all good now

PR #15024, opened 2024-04-26
chore(app-shell): no hard links when building the app locally

mjhuff:
# Overview

Fixes the symlink errors that occur when building the app via "make -C app-shell dist-XYZ". Looks like newer versions of electron-builder change the symlinking, and locally trying to build the app will fail with symlink issues (at least for some OS's). This is the workaround. 

[We already do this when we run CI](https://github.com/Opentrons/opentrons/blob/edge/.github/workflows/app-test-build-deploy.yaml#L299), so this just makes this not an issue locally.

# Test Plan
- Run `make -C app-shell dist-XYZ`. Verify that the command does not fail with a symlinking error (I've tested on all dist-XYZ variations, but feel free to double-check me).
- NOTE: You'll often fail with a key signing error when building a Mac version of the app. That's ok, it still works, and this isn't a regression. 

# Risk assessment
low


shlokamin (approved):
gahhh thank you i really should have done this during the vite migration and saved you the headache 🙈

PR #14908, opened 2024-04-15
feature(api, robot-server): Allow fixit commands to recover from an error

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/EXEC-285.
Allow FIXIT command intent when waiting for recovery and allow resume from recovery. 

# Test Plan

Tested on dev server works as expected!

- set error recovery ff to true `OT_API_FF_enableErrorRecoveryExperiments=1 `
- upload the following protocol - it should fail:
```
from opentrons import protocol_api

requirements = {"robotType": "Flex", "apiLevel": "2.17"}


def run(protocol: protocol_api.ProtocolContext) -> None:
    protocol.load_trash_bin("A3")

    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_50ul", "C2")
    reservoir = protocol.load_labware(
        "opentrons_96_wellplate_200ul_pcr_full_skirt", "C3"
    )
    pipette = protocol.load_instrument(
        "flex_1channel_50", mount="left", tip_racks=[tip_rack]
    )

    for source, dest in zip(reservoir.columns()[0], reservoir.columns()[1]):
        pipette.pick_up_tip()
        pipette.move_to(source.top())
        pipette.move_to(dest.top())
        pipette.drop_tip()
```
- GET `/run` and make sure it is in `AWAITING_RECOVERY` mode. 
- POST a few commands with `intent=FIXIT`
- make sure commands are executed.
- when done, issue a resumeFromRecovery action.  # need to check if this is true

# Review requests

Changes make sense? did I miss anything?

# Risk assessment

Medium. changing `add_command` in pe and `create_command` in router.

PR #15026, opened 2024-04-26
fix(api): disallow hashing we using a fixit command

TamarZanzouri:
# Overview

follow up for https://github.com/Opentrons/opentrons/pull/14908.
disallow hashing for fixit commands.

# Test Plan

POST a fixit command and make sure the hash is `None`.

# Risk assessment

low. set hash to `None` when using a fixit command

### DIFF
--- a/api/tests/opentrons/protocol_engine/commands/test_hash_command_params.py
+++ b/api/tests/opentrons/protocol_engine/commands/test_hash_command_params.py
@@ -1,4 +1,5 @@
 """Tests for hash_command_params."""
+import pytest
 
 from opentrons.protocol_engine import CommandIntent
 from opentrons.protocol_engine import commands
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
ok, cool, nice fix!

api/tests/opentrons/protocol_engine/commands/test_hash_command_params.py
```diff
@@ -66,10 +67,11 @@ def test_repeated_commands() -> None:
     assert a_hash != b_hash
 
 
-def test_setup_command() -> None:
+@pytest.mark.parametrize("command_intent", [CommandIntent.SETUP, CommandIntent.FIXIT])
+def test_setup_command(command_intent: CommandIntent) -> None:
     """Setup commands should always hash to None."""
```
sanni-t:
```suggestion
def test_setup_and_fixit_command(command_intent: CommandIntent) -> None:
    """Setup and fixit commands should always skip hashing."""
```

sanni-t (approved):
LGTM!

Discussed with @TamarZanzouri- As @sfoster1 said in the other PR, we should look into making command intent be always required because the nullability can keep causing confusion. But it is quite an important change that might affect HTTP API (& probably JSON protocols) and would need dedicated testing so we should avoid making such a change this late into the QA process. 

So the approach in this PR looks good and we will add a ticket to make the intent be required for post-release.

PR #15027, opened 2024-04-26
feat(api): improve analysis cli

sfoster1:
- adds a human-json-output flag that will output formatted json that is readable
- adds a log-output and log-level flag to pipe logs around (default is stderr)
- adds a check flag to control return code values
- allows specifying stdout streams with - for file paths

Closes EXEC-425

### DIFF
--- a/api/src/opentrons/cli/analyze.py
+++ b/api/src/opentrons/cli/analyze.py
@@ -1,12 +1,27 @@
 """Opentrons analyze CLI."""
 import click
 
-from anyio import run, Path as AsyncPath
+from anyio import run
+from contextlib import contextmanager
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
 from pydantic import BaseModel
-from typing import Any, Dict, List, Optional, Sequence, Union
-from typing_extensions import Literal
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Union,
+    Literal,
+    Callable,
+    IO,
+    TypeVar,
+    Iterator,
+)
+import logging
+import sys
 
 from opentrons.protocol_engine.types import RunTimeParameter
 from opentrons.protocols.api_support.types import APIVersion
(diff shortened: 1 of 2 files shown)


DerekMaggio (commented):
Do we care that if they specify `--json-output - --human-json-output -` that it will be a whole lot of output piped to stdout? I think the answer is no, but wanted to check.

Also should add some tests for the changes in this PR

sfoster1:
> Do we care that if they specify `--json-output - --human-json-output -` that it will be a whole lot of output piped to stdout? I think the answer is no, but wanted to check.

nope! if you want to do that, you can.

> Also should add some tests for the changes in this PR

the changes in this pr are all interface-level changes that we don't really test for most of our code (including this one - there is no explicit testing for argument handling behavior. i'll make sure that the human-output stuff is parsable, but I don't intend to add exhaustive argument handling tests.

sfoster1:
okay I did add some though

mjhuff (approved):
Nice, lgtm!

api/src/opentrons/cli/analyze.py
```diff
@@ -41,15 +65,107 @@
 @click.option(
     "--json-output",
     help="Return analysis results as machine-readable JSON.",
-    type=click.Path(path_type=AsyncPath),
+    type=click.File(mode="wb"),
```
SyntaxColoring: FYI, using stdout for machine-readable output is unreliable because if we're analyzing a Python protocol, it can *also* print to stdout, and that will corrupt the JSON stream.

Should we `print()` a warning if `-` is specified, so it will fail early and obviously if somebody tries to rely on this?

SyntaxColoring (approved):
Nice.

api/src/opentrons/cli/analyze.py
sfoster1: IMO no, but we should put a warning in the help text. Somebody will only know to use that flag by actually reading the help text, so that should be good enough.

That's a good secondary improvement too - there's no reason we can't capture stdout and save it somewhere separate.

PR #15028, opened 2024-04-26
added script to read individual files

rclarke0:
# Overview

Local Run Log Reader

# Test Plan

Tested on a folder of multiple run logs. 
-->

# Changelog

Added functions in google drive tool to download and search for files
Added a script to look for files based on a local .json file with strings and download
Added a script to allow for local run log reading without a google sheets upload.

# Review requests

# Risk assessment


sfoster1 (approved):
Nice!

PR #15030, opened 2024-04-26
refactor(api): Use a dataclass, not a BaseModel, for PipetteNozleLayoutResultMixin

SyntaxColoring:
# Overview

This class is only used internally, in-memory, and never serialized to JSON. So, make it a `@dataclass` instead of a `pydantic.BaseModel` for clarity and performance.

# Test Plan

Just CI.

# Review requests

None in particular.

# Risk assessment

lo


sfoster1 (approved):
love it

PR #15032, opened 2024-04-26
fix(api): pass the calibrated jaw max offset value when we resetting instrument

ahiuchingau:
# Overview
We really only need to calibrate the gripper's max jaw offset once every time we attach a new gripper. This means that when we call `ot3api.reset()`, which is pretty often because it's called by `ot3api.stop()`, we should just reuse the calibrated value so we don't see the jaw opening and closing all the time.

### DIFF
--- a/api/src/opentrons/hardware_control/instruments/ot3/gripper.py
+++ b/api/src/opentrons/hardware_control/instruments/ot3/gripper.py
@@ -331,9 +331,7 @@ def _reload_gripper(
                     new_config,
                     cal_offset,
                     attached_instr._gripper_id,
-                    attached_instr._jaw_max_offset
-                    if attached_instr.has_jaw_width_calibration
-                    else None,
+                    attached_instr._jaw_max_offset,
                 ),
                 False,
             )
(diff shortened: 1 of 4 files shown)


sfoster1 (changes requested):
Okay but let's get some tests please.

api/src/opentrons/hardware_control/instruments/ot3/gripper.py
```diff
@@ -330,6 +331,9 @@ def _reload_gripper(
                     new_config,
                     cal_offset,
                     attached_instr._gripper_id,
+                    attached_instr._jaw_max_offset
+                    if attached_instr.has_jaw_width_calibration
```
vegano1: `has_jaw_width_calibration` just checks if `_jaw_max_offset` is not None
so we can just pass in `_jaw_max_offset` directly without checking.

api/src/opentrons/hardware_control/instruments/ot3/gripper_handler.py
```diff
@@ -51,6 +51,9 @@ def reset_gripper(self) -> None:
             og_gripper.config,
             load_gripper_calibration_offset(og_gripper.gripper_id),
             og_gripper.gripper_id,
+            og_gripper._jaw_max_offset
+            if og_gripper.has_jaw_width_calibration
```
vegano1: same as above

sfoster1 (approved):
lgtm!

PR #15035, opened 2024-04-29
fix(app): fix input field behavior on blur

koji:
# Overview
fix input field behavior on blur.
The error message will show up when deleting a value in an input field and fired `onBlur` event.

This PR is for the first one of RQA-2646 since the second one will need some time to fix and should be fixed in a following PR.

close RQA-2646 first one

[before]

https://github.com/Opentrons/opentrons/assets/474225/c5acadf7-6765-47ab-a0cd-34f30bbe6410

[after]

https://github.com/Opentrons/opentrons/assets/474225/c14861c1-8fea-450e-a15e-05186f2937c6

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


app/src/organisms/ChooseProtocolSlideout/index.tsx
```diff
@@ -228,8 +229,9 @@ export function ChooseProtocolSlideoutComponent(
       } else if (runtimeParam.type === 'int' || runtimeParam.type === 'float') {
         const value = runtimeParam.value as number
         const id = `InputField_${runtimeParam.variableName}_${index.toString()}`
+        console.log()
```
jerader:
```suggestion
```

jerader (approved):
lgtm

app/src/organisms/ChooseProtocolSlideout/index.tsx
koji: thanks!

PR #15041, opened 2024-04-29
fix(app,components): fix add a new case to info screen for labware

koji:
fix add a new case to info screen for

close RQA-2648

# Overview
fix add a new case to info screen for Labware.

For InfoScreen, we are passing `t`, but in my understanding it never works since there is no information that we can retrieve by a key.

I think we should remove `t` since ui components should not include i18n.
What do you think about this? @ncdiehl11  @jerader 

### [before]
missing labware empty state

### [after]
Screenshot 2024-04-29 at 5 43 55 PM

close RQA-2648

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/app/src/organisms/ProtocolDetails/ProtocolLabwareDetails.tsx
+++ b/app/src/organisms/ProtocolDetails/ProtocolLabwareDetails.tsx
@@ -8,6 +8,7 @@ import {
   DIRECTION_ROW,
   Flex,
   Icon,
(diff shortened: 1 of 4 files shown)


mjhuff (approved):
Looks good. Thanks for adding the test, too!

koji:
@mjhuff thank you for reviewing!

PR #15047, opened 2024-04-30
fix(app-shell): monitor usb only once

sfoster1:
We've observed a crash in the app node side that occurs inside usb-detection likely to do with some surprisingly-shaped descriptors, and those descriptors are polled only (a) on request or (b) when you initialize a webusb device. So let's remove the webusb device initialization and see if that helps.

Also, the webusb connect listeners are as far as I can tell completely duplicative of the listeners in system-info/usb-devices, I wonder if this was a merge conflict?

Closes RQA-2641

### DIFF
--- a/app-shell/src/main.ts
+++ b/app-shell/src/main.ts
@@ -4,7 +4,6 @@ import electronDebug from 'electron-debug'
 import dns from 'dns'
 import contextMenu from 'electron-context-menu'
 import * as electronDevtoolsInstaller from 'electron-devtools-installer'
-import { webusb } from 'usb'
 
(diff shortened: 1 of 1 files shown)


koji (approved):
thanks!

TamarZanzouri (approved):
LGTM!

y3rsh (approved):
Cool so now just handled in
`app-shell/src/system-info/index.ts`

sfoster1:
> Cool so now just handled in `app-shell/src/system-info/index.ts`

and always was. I think this code would have been leaking listeners and causing connect events to fire a bunch of times.

PR #15051, opened 2024-04-30
refactor(api): Make sure command implementations return something compatible with the command's result type

SyntaxColoring:
# Overview

This is a step towards EXEC-427.

Every command declares a `result` type, e.g `Aspirate` has `AspirateResult`. But it turns out that nothing enforces that the command's implementation returns that type—`AspirateImplementation` could return a `DispenseResult`. This has been okay in practice so far because each implementation only returns one thing, and it's not easy for them to fall out of sync. But you can imagine how it will get messy when we start introducing more structured error types. My rough plan is for `execute() -> AspirateResult` to become `execute() -> AspirateResult | AspirateErrorA | AspirateErrorB | ...`, and we will want to enforce that `AspirateErrorA | AspirateErrorB` matches the type of the `Aspirate.error` field.

# Changelog

See my inline notes.

# Test plan

* [x] Make sure CI keeps passing.
* [x] Make sure the "new field" doesn't show up in robot-server's HTTP API documentation.

# Review requests

Does this make sense?

# Risk assessment

Very low.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/command.py
+++ b/api/src/opentrons/protocol_engine/commands/command.py
@@ -31,14 +31,11 @@ if TYPE_CHECKING:
     from ..state import StateView
 
 
-CommandParamsT = TypeVar("CommandParamsT", bound=BaseModel)
-CommandParamsT_contra = TypeVar(
-    "CommandParamsT_contra", bound=BaseModel, contravariant=True
-)
-CommandResultT = TypeVar("CommandResultT", bound=BaseModel)
-CommandResultT_co = TypeVar("CommandResultT_co", bound=BaseModel, covariant=True)
-CommandPrivateResultT = TypeVar("CommandPrivateResultT")
-CommandPrivateResultT_co = TypeVar("CommandPrivateResultT_co", covariant=True)
+_ParamsT = TypeVar("_ParamsT", bound=BaseModel)
+_ParamsT_contra = TypeVar("_ParamsT_contra", bound=BaseModel, contravariant=True)
+_ResultT = TypeVar("_ResultT", bound=BaseModel)
+_ResultT_co = TypeVar("_ResultT_co", bound=BaseModel, covariant=True)
+_PrivateResultT_co = TypeVar("_PrivateResultT_co", covariant=True)
 
 
 class CommandStatus(str, Enum):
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -167,10 +169,15 @@ class BaseCommand(GenericModel, Generic[CommandParamsT, CommandResultT]):
         ),
     )
 
+    _ImplementationCls: Union[
+        Type[AbstractCommandImpl[_ParamsT, _ResultT]],
+        Type[AbstractCommandWithPrivateResultImpl[_ParamsT, _ResultT, object]],
+    ]
+
```
SyntaxColoring, line 176: By convention, each command links itself to its implementation via an `_ImplementationCls` class attribute.

Here, I'm declaring that attribute on the base class. This enforces that each command's `._ImplementationCls` matches its `.params` and `.result`. In the near future, this will also make sure it matches its `.error`.

api/src/opentrons/protocol_engine/commands/command.py
```diff
@@ -167,10 +169,15 @@ class BaseCommand(GenericModel, Generic[CommandParamsT, CommandResultT]):
...
 
 class AbstractCommandImpl(
     ABC,
-    Generic[CommandParamsT, CommandResultT],
+    Generic[_ParamsT_contra, _ResultT_co],
```
SyntaxColoring, line 180: In English:

* An implementation that accepts a broader parameter type can be placed in a variable where it's annotated to accept a narrower parameter type. It's contravariant in its parameters.
* An implementation that returns a narrower result type can be placed in a variable where it's annotated to return a broader parameter type. It's covariant in its result.

[This is the usual behavior for callables](https://peps.python.org/pep-0483/#covariance-and-contravariance:~:text=One%20of%20the%20best%20examples%20to%20illustrate%20(somewhat%20counterintuitive)%20contravariant%20behavior%20is%20the%20callable%20type.), which is basically what a command implementation is. I think we need to spell it out manually because we're going through a wrapper class; the type-checker would understand this automatically if, e.g., the `AspirateImplementation.execute()` method were just a free-standing `execute()` function.

api/src/opentrons/protocol_engine/commands/command.py
SyntaxColoring, line 176: If you're using VS Code or another editor with Pyright integration, you'll get a red squiggly on every command's override of `_ImplementationCls`. (Mypy does not complain.)

I think it has to do with the fact that `_ImplementationCls` looks writable, so it's unsound for a subclass to have a more restrictive type than the base class. It looks like `commandType` has this problem, too. I don't have a solution right now. Maybe it'll get easier with Pydantic v2.

jbleon95 (approved):
Changes make sense to me.

PR #15053, opened 2024-04-30
chore(app-shell): make-dev fixes

sfoster1:
make dev for the robot-stack release kind would print errors in its logs and not prompt updates because the dev app update was malformed. Also, we weren't switching the dev-app-update correctly because it wasn't a phony target.

The log is the "undefined trying to access includes" one.

After this, `make dev` should prompt updates whether you're on `OPENTRONS_PROJECT=robot-stack` or `OPENTRONS_PROJECT=ot3` (before, it only would on `=ot3`)

### DIFF
--- a/app-shell/Makefile
+++ b/app-shell/Makefile
@@ -171,18 +171,16 @@ _dist-collect-artifacts:
 # development
 #####################################################################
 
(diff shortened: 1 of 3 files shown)


mjhuff (approved):
sweet!

PR #15057, opened 2024-05-01
feat(protocol-designer): custom blowout flowRates

jerader:
closes AUTH-4 AUTH-6

# Overview

This PR wires up a custom blowout flowrate field in protocol designer for the mix and transfer steps. Step-generation was already wired up! So just had to fix up PD. Then the migration field was updated to include the new `blowout_flowRate` field. 

Additionally, a new cypress migration test was added that adds MoaM for temp and the new advanced settings

# Test Plan

Create a flex or ot-2 protocol. Add a transfer step. in the advanced settings, see that when the blowout checkbox is checked in the dispense area, that there are 3 fields for location, flow rate and position. The flow rate field should have a tooltip describing that it is the blowout flow rate and clicking on it should open the flow rate modal to use default or add a custom.

Now create a mix step and see that the blowout checkbox also reveals the flow rate input field. Should work as expected.

The migration should work since there are cypress tests to test it!

# Changelog

- add a new field `blowout_flowRate` and wire it up in mix and move liquid
- add a new flow rate field for blowout
- add to the migrations - including a new util for returning the correct liquid specs
- update cypress tests
- update unit tests
- add new cypress test that tests the phase 0 new features

# Review requests

see test plan

# Risk assessment

low


ncdiehl11 (approved):
tested and works as expected, including error handling for volumes outside of the given ranges for both transfer and mix 🥇

PR #15060, opened 2024-05-01
feat(protocol-designer): default dispense height is 1mm from the bottom

jerader:
closes AUTH-332 AUTH-377

# Overview

Several people from FAS have mentioned that users don't like that the default dispense height in PD is 0.5mm from the bottom. This PR updates it to be 1 mm from the bottom. So if a user does not change the default height at all it will say 1mm. 

Additionally, the migration modal is updated to include info about this change in behavior.

screenshot_2024-05-01_at_11 34 56_720

# Test Plan

Create a flex or ot-2 protocol and add a transfer step and a mix step. Export the protocol. See that the z-offset for the dispense steps is 1. 

# Changelog

- update the constant
- add new migration modal and text
- update cypress tests

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/cypress/integration/migrations.spec.js
+++ b/protocol-designer/cypress/integration/migrations.spec.js
@@ -105,7 +105,9 @@ describe('Protocol fixtures migrate and match snapshots', () => {
         if (migrationModal) {
           if (migrationModal === 'v8.1') {
             cy.get('div')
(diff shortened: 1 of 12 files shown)


ncdiehl11 (approved):
tested and verified for OT-2 and Flex for mix and transfer defaults 🚀

PR #15064, opened 2024-05-01
fix(protocol-designer): color picker chooses  new color for defaults

jerader:
closes AUTH-375

# Overview

Fixes bug in production that was introduced when i migrated to react-hook-form

# Test Plan

create a flex or ot-2 protocol and add liquids. see that the liquid colors automatically select the next one on the list instead of staying the same color

# Changelog

- don't hard set a default display color and clean up from types and imports and deconstructing props

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/LiquidsPage/LiquidEditForm.tsx
+++ b/protocol-designer/src/components/LiquidsPage/LiquidEditForm.tsx
@@ -4,7 +4,6 @@ import { Controller, useForm } from 'react-hook-form'
 import { yupResolver } from '@hookform/resolvers/yup'
 import { useSelector } from 'react-redux'
 import * as Yup from 'yup'
-import { swatchColors } from '../swatchColors'
 import {
   Card,
   DeprecatedCheckboxField,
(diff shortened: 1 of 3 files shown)


protocol-designer/src/components/LiquidsPage/LiquidEditForm.tsx
```diff
@@ -18,18 +17,23 @@ import {
...
-import { LiquidGroup } from '../../labware-ingred/types'
-import { ColorPicker } from '../ColorPicker'
-import { ColorResult } from 'react-color'
+import type { ColorResult } from 'react-color'
+import type { LiquidGroup } from '../../labware-ingred/types'
 
-type Props = LiquidGroup & {
+interface LiquidEditFormProps {
+  name: string | null | undefined
+  description: string | null | undefined
```
koji:
```suggestion
  name?: string | null
  description?: string | null
```

koji:
@jerader
one quick question
When editing an existing, the app watches liquid name and description's change, but doesn't watch color.
Is that expected behavior? (I don't remember before the app checks color change)


https://github.com/Opentrons/opentrons/assets/474225/3352efee-04a6-4217-9d3b-018624aadcb8

jerader:
> @jerader one quick question When editing an existing, the app watches liquid name and description's change, but doesn't watch color. Is that expected behavior? (I don't remember before the app checks color change)
> 

@koji great catch! this bug probably exists in production too. i will address it

koji (approved):
the sandbox worked as expected!

PR #15065, opened 2024-05-01
refactor(app): generify MQTT Analytics

mjhuff:
# Overview

Now that the OT-2 supports MQTT, we don't need to special-case analytics for only the Flex.

# Risk assessment
low


koji (approved):
Nice! 🤖

PR #15066, opened 2024-05-01
fix(api): stop-requested hanging on stop when awaiting-recovery

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/EXEC-403.
fix bug related to stop-requested hanging when issuing a StopAction when in AwaitingRecovery state.

# Test Plan

- use ff for recoverable commands
- upload a protocol that will fail on a recoverable command.
- issue a stop action while in `awaiting-recovery`
- run should stop. 

# Review requests

is test sufficient ?

# Risk assessment

low. bug fix.


sfoster1 (approved):
Ahm that'd be it, nice find!

SyntaxColoring (approved):
Nice find. Thank you!

PR #15070, opened 2024-05-02
fix(protocol-designer): clean up trash modal and new labware dropdown logic

jerader:
closes RQA-2501, RQA-2143, RQA-1156, AUTH-378

# Overview

Addresses a bunch of bugs and a quality of life improvement:
1. the adapter text when adding an adapter to the deck is capitalized (Auth-378)
2. changing a pipette and tiprack updates the tiprack on the deck (RQA-1156)
3. the trash modal does not show cannot place banner when the trash is already in selected slot (RQA-2143)
4. when `use gripper` is selected, the waste chute option will show in labware location dropdown but when not selected, the waste chute + off-deck locations will show in dropdown (RQA-2501)

# Test Plan

Create a flex protocol and add a gripper and trash bin and waste chute. Open the trash edit modal and see that the `cannot place` banner is not there since the trash is the thing occupying slot A3. Go to the deck map and add an adapter. Add a labware on top of the adapter and see that the text is capitalized in the modal. Add a move labware step and if the gripper is selected, see that the waste chute location + all slots are included in the new location dropdown. If the gripper is not selected, see that the waste chute, all slots, and off-deck labware is included. Then go back to the edit file page and change the pipette and tiprack. Go back to the deckmap and see that the tiprack is udpated to the correctly selected one.

# Changelog

- add text to i18n and capitalize 
- refactor logic to delete and add new containers for the new tipracks that are selected
- add additional logic to when slot is available in the trash modal
- clean up logic for the new location dropdown for when gripper is selected and not selected

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/FilePipettesModal/index.tsx
+++ b/protocol-designer/src/components/modals/FilePipettesModal/index.tsx
@@ -23,14 +23,7 @@ import {
   PipetteName,
   OT2_ROBOT_TYPE,
   getPipetteSpecsV2,
-  FLEX_ROBOT_TYPE,
-  RobotType,
 } from '@opentrons/shared-data'
-import { StepChangesConfirmModal } from '../EditPipettesModal/StepChangesConfirmModal'
-import { PipetteFields } from './PipetteFields'
-import { CrashInfoBox } from '../../modules'
-import styles from './FilePipettesModal.module.css'
-import modalStyles from '../modal.module.css'
 import {
   actions as stepFormActions,
   selectors as stepFormSelectors,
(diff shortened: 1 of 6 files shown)


protocol-designer/src/components/modals/FilePipettesModal/index.tsx
```diff
@@ -159,6 +170,38 @@ const makeUpdatePipettes = (
...
+  // Find tipracks to delete (old tipracks not in new pipettes)
+  previousTiprackLabwares
+    .filter(labware => !newTiprackUris.has(labware.labwareDefURI))
+    .forEach(labware => dispatch(deleteContainer({ labwareId: labware.id })))
+
+  // Create new tipracks that are not in previous tiprackURIs
+  newTiprackUris.forEach(tiprackDefUri => {
+    if (!previousTiprackUris.has(tiprackDefUri)) {
+      const adapterUnderLabwareDefURI = newPipetteArray.some(
+        pip => pip.name === 'p1000_96'
```
koji: let's avoid using `pip` 😁 

```suggestion
        pipette => pipette.name === 'p1000_96'
```

jerader: i got lazy 😆 😅

protocol-designer/src/components/modals/FilePipettesModal/index.tsx
```diff
@@ -159,6 +170,38 @@ const makeUpdatePipettes = (
...
+      )
+        ? adapter96ChannelDefUri
+        : undefined
```
koji, line 196: Do you think we can use `null` instead of `undefined`?

koji (approved):
leave 2 comments but the changes make sense to me and the sandbox worked as expected.

PR #15071, opened 2024-05-02
fix(app): fix tooltip behavior for disabled Restore default value on RTP slideout

ncdiehl11:
# Overview

When all parameter values in the RTP slideout match their respective defaults, the link to restore all default values should be disabled. Designs specify that clicking this button will render a tooltip explaining its disabled reason for 2 seconds, but hovering over it will not.

# Test Plan

- begin setup on Desktop App for RTP protocol from protocol details
- select any robot and continue to parameters
- observe that "Restore default values" link is disabled because all RTP values match defaults
- click restore values link
- observe that a tooltip renders with the info "No custom values specified"
- repeat all of the above starting from device details rather than from protocol details

# Changelog

- modify tooltip behavior to show on click and disappear after 2 seconds, only if link is disabled
- add tests

# Review requests

auth js

# Risk assessment

low


koji (approved):
both cases worked as expected
the changes look good to me.

the target might be `7.3.0 branch`?

PR #15072, opened 2024-05-02
chore(mypy): fix cache cleanup

ahiuchingau:
# Overview
We haven't been properly clearing the mypy cache for some python directories when we do `make clean-py` because we were using the wrong paths. 

It seems like we don't clear pytest caches in most directories in the cleanup rules, is that intentional?

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -61,7 +61,7 @@ twine_repository_url ?= $(pypi_test_upload_url)
 
 # Defined separately than the clean target so the wheel file doesn’t have to
 # depend on a PHONY target
-clean_cmd = $(SHX) rm -rf build .coverage coverage.xml '*.egg-info' '**/__pycache__' '**/*.pyc' 'src/**/.mypy_cache'
+clean_cmd = $(SHX) rm -rf build .coverage coverage.xml '*.egg-info' '**/__pycache__' '**/*.pyc' '**/.mypy_cache'
 clean_wheel_cmd = $(clean_cmd) dist/*.whl
 clean_sdist_cmd = $(clean_cmd) dist/*.tar.gz
 clean_all_cmd = $(clean_cmd) dist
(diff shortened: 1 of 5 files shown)


sfoster1 (commented):
Do we have to? It does tend to slow things down. Maybe we can make `clean-extra` targets if we really want to?

ahiuchingau:
> Do we have to? It does tend to slow things down. Maybe we can make `clean-extra` targets if we really want to?

Do you mean the pytest cache or mypy or both? Do you propose that I move the mypy cache and add pytest cache to a different target (makes sense and I'm very ok with that)? 

For context, I noticed this only because I was switching between branches that have different pipenv requirements and for some reason mypy was not behaving unless I manually removed the cache first. So I thought not clearing the cache properly was the problem. But now I can't seem to reproduce the same issue I was having so I'm stumped.

sfoster1:
> > Do we have to? It does tend to slow things down. Maybe we can make `clean-extra` targets if we really want to?
> 
> Do you mean the pytest cache or mypy or both? Do you propose that I move the mypy cache and add pytest cache to a different target (makes sense and I'm very ok with that)?
> 
> For context, I noticed this only because I was switching between branches that have different pipenv requirements and for some reason mypy was not behaving unless I manually removed the cache first. So I thought not clearing the cache properly was the problem. But now I can't seem to reproduce the same issue I was having so I'm stumped.

Yeah the thing is that running mypy over and over with no cache is slower, and we always clean when we build, so you'd end up with slow behavior when you do normal dev processes. 

That said we definitely shouldn't have wrong paths in there, i wonder if having like `clean-build` removing `build/` and such and `clean:` removing all caches would be a good idea. then the build targets depend on `clean-build` and `clean` is only ever run by humans.

sfoster1 (approved):
looks good, ty!

PR #15073, opened 2024-05-02
fix(protocol-designer): air-gap in substep-highlight

jerader:
closes AUTH-298

# Overview

This is a gnarly bug that i'm pretty sure has existed since the conception of air gap. Basically, before the vitest migration, `Well.tsx` is `components` directory had a null protection where if well was null then it returned null. When we removed that null protection in the vitest migration, users began to notice the whitescreen outlined in the ticket.

This issue was that when you hover over the air gap step,  the highlighted step wasn't working correctly (highlighting only the aspirate labware and not the dispense labware). The protocol that causes the whitescreen is unique in that the aspirate labware is a 24 well aluminum block and the dispense labware is a 96 well plate. When the air gap after dispense was happening in well F1, the whitescreen occurred because the 24 well aluminum block does not have well F1.

The work around for now (just a temporary fix) is to filter out the air gap substeps. We can add them back if customers want it. BUT in order to add the air gap substeps back, we would need knowledge of if the air gap is happening in the aspirate or dispense labware which it currently does not know, so it would be a pretty hefty refactor I think.

# Test Plan

First on edge, upload the attached protocol in the ticket and open the 4th step (a transfer step). Expand the substeps and highlight through each of the items. Once you reach the air gap in F1, you should get a white screen (Also well. G1 and H1 should white screen too)

Now on this branch, upload the same protocol and do the same steps, you shouldn't get a whitescreen

# Changelog

- add isAirGap to substep type and filter out the air gap aspirate commands for not as a temp fix

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/top-selectors/substep-highlight.ts
+++ b/protocol-designer/src/top-selectors/substep-highlight.ts
@@ -204,7 +204,7 @@ function _getSelectedWellsForSubstep(
 
   if (
     'sourceLabware' in stepArgs &&
-    stepArgs.sourceLabware &&
+    stepArgs.sourceLabware != null &&
     stepArgs.sourceLabware === labwareId
   ) {
     wells.push(...getWells('source'))
@@ -212,7 +212,7 @@ function _getSelectedWellsForSubstep(
 
   if (
     'destLabware' in stepArgs &&
-    stepArgs.destLabware &&
+    stepArgs.destLabware != null &&
     stepArgs.destLabware === labwareId
   ) {
     wells.push(...getWells('dest'))
--- a/protocol-designer/src/top-selectors/substep-highlight.ts
+++ b/protocol-designer/src/top-selectors/substep-highlight.ts
@@ -202,13 +202,19 @@ function _getSelectedWellsForSubstep(
 
   // source + dest steps
 
-  // @ts-expect-error(sa, 2021-6-22): `sourceLabware` is missing in `MixArgs`
-  if (stepArgs.sourceLabware && stepArgs.sourceLabware === labwareId) {
+  if (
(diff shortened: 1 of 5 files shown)


protocol-designer/src/top-selectors/substep-highlight.ts
```diff
@@ -202,13 +202,19 @@ function _getSelectedWellsForSubstep(
 
   // source + dest steps
 
-  // @ts-expect-error(sa, 2021-6-22): `sourceLabware` is missing in `MixArgs`
-  if (stepArgs.sourceLabware && stepArgs.sourceLabware === labwareId) {
+  if (
+    'sourceLabware' in stepArgs &&
+    stepArgs.sourceLabware &&
```
koji:
```suggestion
    stepArgs.sourceLabware != null &&
```

protocol-designer/src/top-selectors/substep-highlight.ts
```diff
@@ -202,13 +202,19 @@ function _getSelectedWellsForSubstep(
...
+    stepArgs.sourceLabware === labwareId
+  ) {
     wells.push(...getWells('source'))
   }
 
-  // @ts-expect-error(sa, 2021-6-22): property `destLabware` is missing in `MixArgs`
-  if (stepArgs.destLabware && stepArgs.destLabware === labwareId) {
+  if (
+    'destLabware' in stepArgs &&
+    stepArgs.destLabware &&
```
koji:
```suggestion
    stepArgs.destLabware != null &&
```

koji (approved):
the changes look good to me.
the sandbox didn't get the white screen.
(I didn't know the expand ui)

PR #15074, opened 2024-05-02
feat(app): add labware selection and volume entry screens

smb2268:
fix PLAT-176, PLAT-177, PLAT-237

# Overview
This PR adds all functionality necessary for the Source Labware Selection, Destination Labware Selection, and Volume screens of the quick transfer flow

# Test Plan
With FF turned on, navigate to the protocols tab and launch the quick transfer flow. All steps of the initial wizard now exist, so you can navigate through the whole flow. After volume entry, since summary screen is not yet implemented the flow will close and final state of information gathered from the flow will be printed in the console. 
Video not uploading but I posted it here: https://opentrons.slack.com/archives/C06LPQN1YHM/p1714685498743729

# Changelog
1. Add and implement SelectSourceLabware, SelectDestLabware, VolumeEntry components as well as shells with TODO for SelectSourceWells and SelectDestWells
2. Add util for calculating the acceptable volume range for the final screen based on all previous selections
3. Add util for determining what labware to display for each pipette, including a util thats output is now stored in the `constants.ts` file for this repo so that we don't have to do the well geometry comparison at run time
4. Add exit prevention screen
5. Add test coverage for all of these new additions

# Review requests
Look through the logic of the utils - I tried to explain functionality in comments in the test cases but let me know if you have questions or if anything doesn't look right!

# Risk assessment
Low, behind FF


app/src/atoms/SoftwareKeyboard/NumericalKeyboard/index.css
```diff
@@ -10,6 +10,7 @@
 .simple-keyboard.oddTheme1.hg-theme-default {
   width: 100%;
   height: 100%;
+  border-radius: 0;
```
smb2268, line 13: Designs for this component show no rounded edges, the default for the keyboard component we use is border-radius 5

app/src/organisms/QuickTransferFlow/types.ts
```diff
@@ -8,7 +8,7 @@ export interface QuickTransferSetupState {
   tipRack?: LabwareDefinition2
   source?: LabwareDefinition2
   sourceWells?: string[]
-  destination?: LabwareDefinition2
+  destination?: LabwareDefinition2 | 'source'
```
smb2268, line 11: We allow users to select either a new labware or use the same labware as they selected for the source

app/src/organisms/QuickTransferFlow/index.tsx
```diff
@@ -82,52 +81,74 @@ export const QuickTransferFlow = (): JSX.Element => {
         exitButtonProps={exitButtonProps}
       />
     )
+  } else if (currentStep === 4) {
```
b-cooper: Nit: this looks fine, though I wonder if it would be clearer as an array of JSX that you index into?

e.g. 

```
const stepsInOrder = [
  <...>,
  <...>,
  ...
]

return (
  <>
    {stepsInOrder[currentStepIndex]}
  </>
)
```

b-cooper (approved):
Code looks great!  I like the constant generation util solution too.  Super clear naming and comments ☄️

app/src/organisms/QuickTransferFlow/index.tsx
smb2268: Sure, that works! I also made a shared props object to minimize lines a bit

PR #15075, opened 2024-05-02
refactor(api): remove legacy from names of variables and classes that are not actually legacy

jbleon95:
# Overview

Sometime ago, a number of classes in our `protocol_runner` folder were marked as legacy with the thought that we'd move to a new major API version and those would no longer be the actively developed parts of our protocol reading and execution.

However this proved not to be the cases, and for the past couple years these have continued to be marked as legacy despite the fact that they are being actively used now and for the foreseeable future. Furthermore, there are a number of folders, classes and variables that are marked legacy which _are_ actually legacy parts of our system, further increasing confusion of what is current and what is deprecated/no longer in active development.

This PR renames everything in the `protocol_runner` that was prematurely marked as legacy, and additionally gets rid of a bunch of re-exports that were further naming things that are no longer legacy.

# Test Plan

Pure refactor PR, so existing unit and integration tests cover everything.

# Changelog

- a bunch of internal renaming, but the most outward facing changes are renaming `legacy_wrappers.py` to `python_protocol_wrappers.py`, `LegacyFileReader` to `PythonProtocolFileReader`, `LegacyContextCreator` to `ProtocolContextCreator`, and `LegacyExecutor` to `PythonProtocolExecutor`

# Review requests

Is there anything I either missed or incorrectly labelled as no longer legacy but which in fact is. I also removed the export renaming of `LoadInfo` since the fact that it's legacy is shown by the fact its in the correctly named `legacy` folder of `protocol_api`, but if this is a step too far I can revert those changes.

# Risk assessment

Low, refactor only PR.

### DIFF
--- a/api/src/opentrons/protocol_runner/legacy_command_mapper.py
+++ b/api/src/opentrons/protocol_runner/legacy_command_mapper.py
@@ -4,9 +4,24 @@ from collections import defaultdict
 from datetime import datetime
 from typing import Dict, List, Optional, Tuple, Union
 
+from opentrons.hardware_control.modules.types import (
+    ModuleModel as HardwareModuleModel,
+    TemperatureModuleModel,
+    MagneticModuleModel,
+    ThermocyclerModuleModel,
+    HeaterShakerModuleModel,
+)
 from opentrons_shared_data.pipette.dev_types import PipetteNameType
 from opentrons.types import MountType, DeckSlotName, Location
 from opentrons.legacy_commands import types as legacy_command_types
+from opentrons.protocol_api import InstrumentContext
+from opentrons.protocol_api.core.legacy.deck import FIXED_TRASH_ID
+from opentrons.protocol_api.core.legacy.load_info import (
+    LoadInfo,
+    LabwareLoadInfo,
+    InstrumentLoadInfo,
+    ModuleLoadInfo,
+)
 from opentrons.protocol_engine import (
     ProtocolEngineError,
     actions as pe_actions,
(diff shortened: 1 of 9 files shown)


api/src/opentrons/protocol_runner/legacy_command_mapper.py
```diff
@@ -4,9 +4,24 @@
...
 from opentrons_shared_data.pipette.dev_types import PipetteNameType
 from opentrons.types import MountType, DeckSlotName, Location
 from opentrons.legacy_commands import types as legacy_command_types
+from opentrons.protocol_api import InstrumentContext
+from opentrons.protocol_api.core.legacy.deck import FIXED_TRASH_ID
+from opentrons.protocol_api.core.legacy.load_info import (
+    LoadInfo,
+    LabwareLoadInfo,
+    InstrumentLoadInfo,
+    ModuleLoadInfo,
```
sanni-t: I do prefer these to be aliased as `LegacyLoadInfo`, etc because they actually are classes from/for the legacy system.

api/src/opentrons/protocol_runner/python_protocol_wrappers.py
```diff
@@ -63,24 +44,24 @@
 LEGACY_JSON_SCHEMA_VERSION_CUTOFF = 6
 
 
-class LegacyFileReader:
+class PythonProtocolFileReader:
```
sanni-t: How about `PythonAndLegacyFileReader` to be consistent with the `PythonAndLegacyRunner`, and also since it actually does include the legacy JSON protocol files.

api/src/opentrons/protocol_runner/legacy_context_plugin.py
```diff
@@ -129,7 +129,7 @@ def _handle_legacy_command(self, command: LegacyCommand) -> None:
         pe_actions = self._legacy_command_mapper.map_command(command=command)
         self._actions_to_dispatch.put(pe_actions)
 
-    def _handle_equipment_loaded(self, load_info: LegacyLoadInfo) -> None:
+    def _handle_equipment_loaded(self, load_info: LoadInfo) -> None:
         """Handle an equipment load reported by the APIv2 protocol.
```
sanni-t: Maybe we can update these docstrings to be clearer too.
```suggestion
        """Handle an equipment load reported by the legacy APIv2 protocol.
```

sanni-t (commented):
So glad this is happening! Added some soft suggestions.

sanni-t (approved):
LGTM!

PR #15076, opened 2024-05-02
fix(step-generation): liquidState tracking for 96-channel column pickup

jerader:
closes RESC-246 AUTH-373

# Overview

This PR fixes a bug in the liquid state tracking specifically for using a 96-channel column partial tip configuration. The bug was that I forgot to consider partial tip for the 96-channel in the liquid state tracking all together so the protocol in the RESC ticket was tracking liquids as if all 96 nozzles were being used instead of 1 column

# Test Plan

Upload the protocol attached to the RESC ticket on edge. The last transfer step should have a "not enough liquid in wells" warning. Upload the same protocol on this branch. The last transfer step should NOT have a "not enough liquid in wells" warning.

# Changelog

- update the dispense update liquid state component to grab the pipette's nozzle configuration. If nozzles is column then the channels is 8 instead of 96
- fix the test
- extend props to include the robot state and warnings and update all affected components

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/__tests__/dispenseUpdateLiquidState.test.ts
+++ b/step-generation/src/__tests__/dispenseUpdateLiquidState.test.ts
@@ -8,7 +8,12 @@ import merge from 'lodash/merge'
 import omit from 'lodash/omit'
 import produce from 'immer'
 import { createEmptyLiquidState, createTipLiquidState } from '../utils'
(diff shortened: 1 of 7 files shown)


koji (approved):
the sandbox worked as expected!

PR #15077, opened 2024-05-02
feat(app): Add Error Recovery orchestrator

mjhuff:
Closes [EXEC-389](https://opentrons.atlassian.net/browse/EXEC-389)

# Overview

Error Recovery Flows are "special and unique"™ compared to other wizard flows. This PR addresses the underlying structure of ER and how to best solve for that uniqueness. I've outlined some of that uniqueness below and appreciate feedback/thoughts!

### We May (or Not) Support Keeping Apps in Sync 

Because there may be a world down the road in which we want to keep users across multiple desktop apps/the ODD in sync, we need some sort of way for the client to report where it is in ER. To that end, I propose breaking all content within ER flows into **recovery routes** and **recovery steps**. A recovery route is a logically-related collection of recovery steps or a single step if unrelated to any existing recovery route. A recovery step is analogous to a "step" in other wizard flows - a deterministic ordering of content for a given route. 

While it would be great if we could use React Router to create our own mini-router, in practice this doesn't work. There is only one route at any one time (as far as React Router is concerned), and we can't simply use a subset of routes within a modal (React router will redirect the page). While this could work if all apps were the ODD, where we could effectively make the modal a page), we can't do this with desktop apps, so this option is out.

My solution to the above has been to recreate react-router without using react-router. We navigate around the modal when it's opened using a route, and the step applies changes to the view when needed. What this means is that when we decide to keep clients in sync across apps, we just need to create some hook that creates a URL-esque string from the route and step. 

The major drawback of this approach is that we don't really do this anywhere else on the FE, so it's a new thing to learn. I've tried to compensate for the newness by enforcing strict type checking, so as soon as someone tries adding a new route somewhere, lint should be helpful.

### ER Flows are Non-Deterministic

Unlike other flows, ER does not use a step counter. This is because a user may select from one or more recovery options, each with a variable number of steps. These recovery options all branch from a central node, the `SelectRecoveryOptions` route. The major utility to help navigate these flows is `useRouteUpdateActions`. 

Despite being non-determinstic include a lot of "Go back." The way we handle this in other flows is to return the previous step within a series of steps, which I've also emulated when we are in a particular route. The drawback of this is that there are a bunch of routes, so that requires us to always account for the ordering of steps for each route, even if the route has one step (which is pretty common at the moment, but I don't want to make assumptions with how things will turn out when designs are still changing). When we can't "go back" anymore, we return to the `SelectRecoveryOptions` by default.

### ER Should be DRY

One issue I've noticed in other wizards in the past is that we break the wizard down into steps, but then each step is independent from other steps entirely in terms of the components that comprise the step. This means that if the same button that's used in every page needs a change, every single component needs to change: we create a relatively large bug surface. To attempt to address this, I've introduced a `shared` directory within `RecoveryOptions`. I may include a `shared` in the broader `ErrorRecoveryFlows` dir if needed. The point is that things that get reused a lot should probably be grouped. I've tried encapsulating exports as well so only the things that should be used by higher-order components are exported.

This being said, I'm open to other ideas. I admit that these footer buttons require some props for overriding behavior, but I still think this is better than doing nothing...it also has the added benefit of keeping things more testable/readable.

https://github.com/Opentrons/opentrons/assets/64858653/cd390fcb-415a-402b-b82f-c85120e348fb

# Test Plan
- Enable both the BE/FE FFs for ER (let me know if you want an explanation on how to do this). 
- Try to pick up a tip but fail.
- Follow the flow on the video!

# Changelog
- Add the error recovery orchestrator (hidden behind feature flags)

# Review requests
While the focus of this PR is more on overall structure and approach rather than emulating wireframe designs (these designs aren't finalized, and that's why I'm using crazy colors), high-level design feedback is welcome! 

# Risk assessment
low

[EXEC-389]: https://opentrons.atlassian.net/browse/EXEC-389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/assets/localization/en/error_recovery.json
+++ b/app/src/assets/localization/en/error_recovery.json
@@ -1,3 +1,19 @@
 {
-  "run_paused": "Run paused"
+  "are_you_sure_you_want_to_resume": "Are you sure you want to resume?",
+  "before_you_begin": "Before you begin",
+  "cancel_run": "Cancel run",
+  "confirm": "Confirm",
+  "continue": "Continue",
+  "general_error": "General error",
+  "general_error_message": "<Placeholder>",
+  "go_back": "Go back",
+  "how_do_you_want_to_proceed": "How do you want to proceed?",
+  "recovery_mode": "Recovery Mode",
+  "recovery_mode_explanation": "<block>Recovery Mode provides you with guided and manual controls for handling errors at runtime.</block><br/><block>You can make changes to ensure the step in progress when the error occurred can be completed or choose to cancel the protocol. When changes are made and no subsequent errors are detected, the method completes. Depending on the conditions that caused the error, you will only be provided with appropriate options.</block>",
+  "resume": "Resume",
+  "run_paused": "Run paused",
+  "run_will_resume": "The run will resume from the point at which the error occurred. Take any necessary actions to correct the problem first. If the step is completed successfully, the protocol continues.",
+  "stand_back": "Stand back, robot is in motion",
+  "stand_back_resuming": "Stand back, resuming current step",
+  "view_recovery_options": "View recovery options"
 }
--- a/app/src/organisms/ErrorRecoveryFlows/RecoveryOptions/ResumeRun.tsx
+++ b/app/src/organisms/ErrorRecoveryFlows/RecoveryOptions/ResumeRun.tsx
@@ -1,0 +1,62 @@
+import * as React from 'react'
+import { useTranslation } from 'react-i18next'
+
+import {
+  ALIGN_CENTER,
+  DIRECTION_COLUMN,
+  Flex,
+  Icon,
+  JUSTIFY_SPACE_BETWEEN,
+  SPACING,
+  StyledText,
+} from '@opentrons/components'
+
+import { RecoveryFooterButtons } from './shared'
+
+import type { RecoveryContentProps } from '../types'
+
+export function ResumeRun({
+  isOnDevice,
+  onComplete,
+  routeUpdateActions,
+}: RecoveryContentProps): JSX.Element | null {
+  const { t } = useTranslation('error_recovery')
+  const { goBackPrevStep } = routeUpdateActions
+
+  if (isOnDevice) {
+    return (
+      <Flex
+        padding={SPACING.spacing32}
+        gridGap={SPACING.spacing24}
+        flexDirection={DIRECTION_COLUMN}
+        justifyContent={JUSTIFY_SPACE_BETWEEN}
+        alignItems={ALIGN_CENTER}
+        height="100%"
+      >
+        <Flex
+          flexDirection={DIRECTION_COLUMN}
+          alignItems={ALIGN_CENTER}
+          gridGap={SPACING.spacing24}
+          height="100%"
+          width="848px"
+        >
+          <Icon name="ot-alert" size="3.75rem" marginTop={SPACING.spacing24} />
+          <StyledText as="h3Bold">
+            {t('are_you_sure_you_want_to_resume')}
+          </StyledText>
+          <StyledText as="h4" textAlign={ALIGN_CENTER}>
+            {t('run_will_resume')}
+          </StyledText>
+        </Flex>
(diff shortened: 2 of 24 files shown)


app/src/assets/localization/en/error_recovery.json
```diff
@@ -1,3 +1,19 @@
 {
-  "run_paused": "Run paused"
+  "are_you_sure_you_want_to_resume": "Are you sure you want to resume?",
```
sfoster1, line 2: I think we need an anonymized version of this, right @brenthagen

app/src/organisms/ErrorRecoveryFlows/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+      routeUpdateActions={routeUpdateActions}
+    />
+  )
+}
+
+function ErrorRecoveryComponent(props: RecoveryContentProps): JSX.Element {
+  return createPortal(
+    <Flex
+      flexDirection={DIRECTION_COLUMN}
+      width="992px"
```
sfoster1, line 64: do we have a place to derive these? i guess we use them in very few places

app/src/organisms/ErrorRecoveryFlows/RecoveryOptions/ResumeRun.tsx
```diff
@@ -0,0 +1,62 @@
...
+        justifyContent={JUSTIFY_SPACE_BETWEEN}
+        alignItems={ALIGN_CENTER}
+        height="100%"
+      >
+        <Flex
+          flexDirection={DIRECTION_COLUMN}
+          alignItems={ALIGN_CENTER}
+          gridGap={SPACING.spacing24}
+          height="100%"
+          width="848px"
```
sfoster1, line 41: can we derive this from somewhere?

sfoster1 (approved):
This is generally good, but depending on how often we want to reuse step visualization I wonder if we might want a separate per-route orchestrator or something. I'm a little worried about encoding the paths between steps as an array because I could certainly see a case where that flow is not linear even within a route, which we'd only be able to model by breaking down routes more and more pathologically.

If instead, the top-level orchestrator hosted some route object (one for each route), and each route object hosted the appropriate step object, we might be able to admit more complex step sequencing behavior in code in the route objects while keeping that behavior localizxed just to the route objects.

app/src/organisms/ErrorRecoveryFlows/index.tsx
mjhuff: I should have left a comment about this - we don't have a unified component for this currently, but it's becoming common enough that we need one IMO. I have a running list of components that should be unified and this one is on it. I plan on giving that list to Mel for feedback when we formally start unification work.

app/src/organisms/ErrorRecoveryFlows/RecoveryOptions/ResumeRun.tsx
mjhuff: I don't think so. This is special-cased by design. My guess is Sue didn't like the way the copy looks in the modal if we use the default width, so there's a hardcoded one-off adjustment to it.

app/src/assets/localization/en/error_recovery.json
mjhuff: Yeah @brenthagen let me know what I need to do for localization/OEM stuff in general, since I'm pretty out of the loop on that.

mjhuff:
> If instead, the top-level orchestrator hosted some route object (one for each route), and each route object hosted the appropriate step object, we might be able to admit more complex step sequencing behavior in code in the route objects while keeping that behavior localizxed just to the route objects.

Thanks for the feedback. I'll mull this over for a bit.

mjhuff:
After in person convo with Seth, we've decided we'll keep this for now and most likely refactor the route/step state as a directed graph when the time is right. For now though, this is sufficient.

PR #15078, opened 2024-05-02
docs(api): describe new `drop_tip()` behavior introduced in PAPI 2.18

ecormany:
# Overview

Describe the new `apiLevel`-gated behavior for `drop_tip()` in Python API 2.18.

Addresses RTC-448

# Test Plan

[Sandbox API ref entry](http://sandbox.docs.opentrons.com/docstring-drop-tip-alternating-2.18/v2/new_protocol_api.html#opentrons.protocol_api.InstrumentContext.drop_tip)

# Changelog

Move ¶ about change in 2.15 down, and add another ¶ about 2.18.

# Review requests

This is how it works, right?

# Risk assessment

None, docstring only.

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1039,14 +1039,14 @@ class InstrumentContext(publisher.CommandPublisher):
               the ``WasteChute`` object. For example,
               ``pipette.drop_tip(location=waste_chute)``.
 
-        Starting with API version 2.15, if ``location`` is a trash container or not
+        In API versions 2.15 to 2.17, if ``location`` is a ``TrashBin`` or not
         specified, the API will instruct the pipette to drop tips in different locations
-        within the trash container. Varying the tip drop location helps prevent tips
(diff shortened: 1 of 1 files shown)


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1044,6 +1039,15 @@ def drop_tip(
               the ``WasteChute`` object. For example,
               ``pipette.drop_tip(location=waste_chute)``.
 
+        In API versions 2.15 to 2.17, if ``location`` is a ``TrashBin`` or not
+        specified, the API will instruct the pipette to drop tips in different locations
```
jbleon95, line 1043: Not sure if we want to be more specific that each mount will alternate between two different left/right locations. Your call.

jbleon95 (approved):
Looks good and describes the behavior accurately. Small comment on specificity but otherwise ✅

api/src/opentrons/protocol_api/instrument_context.py
ecormany: Probably more in the weeds than necessary. Thanks tho!

PR #15080, opened 2024-05-02
User input instead of args to allow for easier bat file creation

rclarke0:
Converted arguments to user inputs to allow for easier bat file creation for scripts that are used often.


sfoster1 (approved):
If it's more convenient, then great!

PR #15082, opened 2024-05-02
fix(app): fix the space between icon and banner outline

koji:
# Overview
fix the space between icon and banner outline
Actually, I don't like this fix personally since using margin isn't a good way to make a layout but Banner is used in many components, and we shouldn't touch the component itself since updating the component may cause other bugs.
We will need to update the component to align with the latest design.

design
https://www.figma.com/file/l7BAJAGICr10ELsSqQD0Sr/Design-System%3A-Desktop?node-id=7602-143972&mode=dev

### [before]
padding bug on success inline notifcation

### [after]
Screenshot 2024-05-02 at 5 25 53 PM

fix RQA-2649

# Test Plan

# Changelog

# Review requests

# Risk assessment


mjhuff (commented):
If we are merging this into `edge`, and we'll have the time to address broader DQA feedback before next release, do you think it's worthwhile just adding this offset directly to the banner component? 

Just looking over all of its usages, I _think_ this margin offset is probably safe to do everywhere. I imagine setting the icon/text a bit more left from the banner's edge will look fine, but if you or others disagree I'm fine with this change for now.

koji:
> If we are merging this into `edge`, and we'll have the time to address broader DQA feedback before next release, do you think it's worthwhile just adding this offset directly to the banner component?
> 
> Just looking over all of its usages, I _think_ this margin offset is probably safe to do everywhere. I imagine setting the icon/text a bit more left from the banner's edge will look fine, but if you or others disagree I'm fine with this change for now.

Actually, this PR should go to `chore_release-7.3.0`

mjhuff:
> Actually, this PR should go to `chore_release-7.3.0`

Think you may need to rebase the branch. I think it's bringing all the `edge` changes into `chore-release`.

koji:
have failed rebase I will re-open a pr that is based on 7.3.0 since that will be faster than fixing rebase issue.

PR #15083, opened 2024-05-02
chore(api-client): run api-client tests in ci

brenthagen:
# Overview

we haven't been running (the minimal) api-client tests in CI. this changes that. `make -C api-client test` runs both api-client and react-api-client tests.

# Test Plan

verified that api-client tests are running in CI

# Changelog

 - Changes workflow to run api-client tests in CI

# Review requests

check that api-client tests are running in CI

# Risk assessment

low

### DIFF
--- a/.github/workflows/react-api-client-test.yaml
+++ b/.github/workflows/react-api-client-test.yaml
@@ -59,8 +59,8 @@ jobs:
           npm config set cache ./.npm-cache
           yarn config set cache-folder ./.yarn-cache
           make setup-js
-      - name: 'run react-api-client unit tests'
-        run: make -C react-api-client test-cov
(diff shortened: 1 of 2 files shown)


.github/workflows/react-api-client-test.yaml
```diff
@@ -59,8 +59,8 @@ jobs:
           npm config set cache ./.npm-cache
           yarn config set cache-folder ./.yarn-cache
           make setup-js
-      - name: 'run react-api-client unit tests'
-        run: make -C react-api-client test-cov
+      - name: 'run api-client unit tests'
```
sfoster1: both maybe?

sfoster1 (approved):

PR #15086, opened 2024-05-03
RDEVOPS-84 Opetrons AI Server foundation

y3rsh:
# Overview

This is the base of the tooling for the backend of the Opentrons AI project

This code is successfully deployed on the dev and sandbox environment using the deploy script here.

## Test Plan

- [x] validate deployment using the `deploy.py` of `dev` and `sandbox`


.github/workflows/ai-server-lint-test.yaml
```diff
@@ -0,0 +1,52 @@
+# This workflow runs lint on pull requests that touch anything in the app-testing directory
```
y3rsh:
```suggestion
# This workflow runs lint on pull requests that touch anything in the opentrons-ai-server directory
```

opentrons-ai-server/Makefile
```diff
@@ -1,2 +1,107 @@
...
+	mkdir -p package
+	cp -r api/. package/
+	@echo "Creating the deployment zip file..."
+	cd package && zip -r ../function.zip . && cd ..
+	@echo "Package created successfully."
+
+.PHONY: package
+package: clean-package install-deps package-lambda
+
+ALLOWED_ENVS := sandbox-opentrons-ai-api dev-opentrons-ai-api
```
y3rsh: need to change this.  switch to just `sandbox` and `dev` for now

opentrons-ai-server/api/models/todo.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Todo(BaseModel):
+    userId: int
+    id: int
+    title: str
+    completed: bool
+
+
+class Todos(List[Todo]):
+    pass
```
y3rsh, line 14: I don't actually know if this works 🤣

opentrons-ai-server/requirements.txt
y3rsh, line 1: I like checking this in too, this is only the prod requirements.  In infra we use the powertools layer, that is why it is a dev dependency.

opentrons-ai-server/README.md
```diff
@@ -6,34 +6,45 @@ The Opentrons AI application's server.
...
-## Stack and structure
+1. clone the repository `gh repo clone Opentrons/opentrons`
+1. `cd opentrons/opentrons-ai-server`
+1. Have pyenv installed per [DEV_SETUP.md](../DEV_SETUP.md)
+1. Use pyenv to install python `pyenv install 3.12.3` or latest 3.12.\*
+1. Have nodejs and yarn installed per [DEV_SETUP.md](../DEV_SETUP.md)
+   1. This allows formatting of of `.md` and `.json` files
+1. select the python version `pyenv local 3.12.3`
+   1. This will create a `.python-version` file in this directory
+1. select the node version `nvs` currently 18.\*
```
koji:
```suggestion
1. select the node version `nvs` currently 18.19\*
```

opentrons-ai-server/README.md
```diff
@@ -6,34 +6,45 @@ The Opentrons AI application's server.
...
+1. Install pipenv and python dependencies `make setup`
+
+## Install a dev dependency
 
-The UI stack is built using:
+`python -m pipenv install pytest==8.2.0 --dev`
```
koji, line 26: Seems other python projects are using v7. Is there any big diff between v7 and v8?

opentrons-ai-server/api/domain/todo.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any, List
+
+from httpx import Response
+
+from api.integration.jsonplaceholder import JSONPlaceholder
+from api.models.todo import Todo, Todos
```
koji, line 6: Question is there any rule for import statements' order?

in frontend we use the following order rule.
1. 3rd party packages
2. internal packages (components, shared-data)
3. local packages/components/functions
4. types (types follows 1-3)
5. css

opentrons-ai-server/README.md
y3rsh: There are, lots of bug fixes and breaking changes so want to use the latest since we are not locked into 7 with the rest of the stack.
<https://docs.pytest.org/en/8.0.x/changelog.html>

opentrons-ai-server/api/domain/todo.py
y3rsh: We are using [ruff ](https://docs.astral.sh/ruff/settings/#lint_isort_order-by-type) for this, I like going with their default but we could investigate the settings to see if we could change it.

opentrons-ai-server/README.md
sfoster1: v8 wasn't released when we did dep upgrades last. it might actually be fine, I don't think we've evaluated it

opentrons-ai-server/api/settings.py
```diff
@@ -0,0 +1,62 @@
...
+    SecretStr,
+)
+
+ENV_PATH: Path = Path(Path(__file__).parent.parent, ".env")
+
+
+class Settings:
+    def __init__(self) -> None:
+        # Load environment variables from .env file if it exists
+        # These map to the the environment variables defined and set in teh
```
sfoster1: unfinished sentence?

opentrons-ai-server/requirements.txt
```diff
@@ -0,0 +1,16 @@
+-i https://pypi.org/simple
```
sfoster1: is this for the lambda package? it's a bit odd to have both this and pipenv in use at the same time, could/should we generate this and gitignore it?

opentrons-ai-server/deploy.py
```diff
@@ -0,0 +1,149 @@
...
+@dataclass(frozen=True)
+class DeploymentConfig:
+    SANDBOX_S3_BUCKET: str = "sandbox-opentrons-ai-api"
+    SANDBOX_FUNCTION_NAME: str = "sandbox-api-function"
+    DEV_S3_BUCKET: str = "dev-opentrons-ai-api"
+    DEV_FUNCTION_NAME: str = "dev-api-function"
+    S3_KEY = "function.zip"
+    S3_ZIP_PATH = Path(Path(__file__).parent, S3_KEY)
+    HEALTH_EVENT = Path(Path(__file__).parent, "test_events", "health.json")
+    DEPLOYMENT_TIMEOUT = 60  # seconds
```
sfoster1: nit: imo this could be `DEPLOYMENT_TIMEOUT_S` and then the comment isn't necessary

opentrons-ai-server/deploy.py
```diff
@@ -0,0 +1,149 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+
+import boto3
+from botocore.exceptions import NoCredentialsError, PartialCredentialsError
+from rich import print
+
+
+@dataclass(frozen=True)
+class DeploymentConfig:
```
sfoster1: sort of nit, sort of more - with this arrangement where everything is in the same bag-of-data dataclass, it's easy and not statically analyzable to accidentally return the wrong one of e.g. `DEV_S3_BUCKET` and `SANDBOX_S3_BUCKET`. if instead there's a `DeploymentConfig` dataclass that has just e.g. `S3_BUCKET`, `FUNCTION_NAME`, and `ZIP_PATH` and no defaults, then you create three of them with the appropriate defaults and pass it around as needed

opentrons-ai-server/deploy.py
```diff
@@ -0,0 +1,149 @@
...
+            response = self.lambda_client.update_function_code(
+                FunctionName=self.function_name, S3Bucket=self.s3_bucket, S3Key=self.s3_key, Publish=True
+            )
+            print("Updated Lambda function:")
+            print(response)
+            version = str(response["Version"])
+            print(f"New version: {version}")
+            return version
+        except Exception as e:
+            print(f"Failed to update Lambda function: {e}")
```
sfoster1: nit: can't we just let this expression raise and print something out at a higher level? then this function doesn't have to return `| None`, and we don't have to check its return at the callsite

opentrons-ai-server/deploy.py
```diff
@@ -0,0 +1,149 @@
...
+        return None
+
+    def wait_for_lambda_status(self, version: str) -> None:
```
sfoster1: this function really needs to indicate that an error occurred somehow. right now you'd have to look at what it prints since if it times out it just returns and it swallows all exceptions. Let's let this raise, and let's raise on a timeout, and have a higher level format nice error prints and `sys.exit(1)` at least.

opentrons-ai-server/deploy.py
```diff
@@ -0,0 +1,149 @@
...
+                    break
+                else:
+                    print("Status still 'Pending'. Checking again in 3 seconds...")
+                    time.sleep(3)  # Wait for 3 seconds before checking again
+
+            except Exception as e:
+                print(f"Error retrieving function status: {e}")
+                break  # Exit loop if there is an error retrieving the function status
+
+    def invoke_lambda(self, version: str) -> None:
```
sfoster1: same here - if this is for a gut-check that the lambda is actually healthy, it really needs to be able to fail in a way that's machine checkable. It should raise if the invocation fails.

opentrons-ai-server/api/settings.py
y3rsh: finish this comment

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+        :param endpoint: Endpoint URL (to be concatenated with base_url)
+        :param kwargs: Additional arguments to pass to httpx request (like json, headers)
+        :return: httpx.Response object or error message as a string
+        """
+        try:
+            url = f"{self.base_url}{endpoint}"
+            response = self.client.request(method, url, **kwargs)
+            return response
+        except httpx.RequestError as e:
+            logger.error(f"An error occurred while requesting {e.request.url!r}.")
```
sfoster1: do we need to log things twice like this? can't we just do the exception logger? the way it works in python is if you `logger.exception('some message')` it logs the message and if there's an active exception logs the exception info. it's basically exactly for this.

opentrons-ai-server/api/settings.py
```diff
@@ -0,0 +1,62 @@
...
+            # Fetch secrets from AWS Secrets manager using AWS Lambda Powertools
+            self.openai_api_key: SecretStr = self.fetch_secret(f"{self.environment}-openai-api-key")
+            self.huggingface_api_key: SecretStr = self.fetch_secret(f"{self.environment}-huggingface-api-key")
+        else:
+            # Use values from .env or defaults if not set
+            self.openai_api_key = SecretStr(os.getenv("OPENAI_API_KEY", "default-openai-secret"))  # can change to throw
+            self.huggingface_api_key = SecretStr(os.getenv("HUGGINGFACE_API_KEY", "default-huggingface-secret"))  # can change to throw
+
+    @staticmethod
+    def fetch_secret(secret_name: str) -> SecretStr:
```
y3rsh, line 35: think about this more, should this be in integrations? Probably

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+            logger.exception(e)
+        except httpx.HTTPStatusError as e:
```
sfoster1: As far as I can tell, `httpx` will only raise an `HTTPStatusError` if you call `response.raise_for_status()`:
> The HTTPStatusError class is raised by response.raise_for_status() on responses which are not a 2xx success code. These exceptions include both a .request and a .response attribute.

(per [their quickstart](https://www.python-httpx.org/quickstart/) which unfortunately doesn't allow deeplinks). So I don't think this is possible to do.

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+        :param kwargs: Additional arguments to pass to httpx request (like json, headers)
+        :return: httpx.Response object or error message as a string
+        """
+        try:
+            url = f"{self.base_url}{endpoint}"
+            response = self.client.request(method, url, **kwargs)
+            return response
+        except httpx.RequestError as e:
+            logger.error(f"An error occurred while requesting {e.request.url!r}.")
+            logger.exception(e)
```
sfoster1: Handling this exception like this, with an `except (e):` that does nothing but log, swallows the error fully and won't raise it further. In both exception cases (though as below, I don't think the second one can happen) that means that this function will return `None` implicitly because the exception will happen before the return clause, which is both not as annotated and probably not what you want. I think these exception clauses should reraise.

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+        :param method: HTTP method as a string ('GET', 'POST', etc.)
+        :param endpoint: Endpoint URL (to be concatenated with base_url)
+        :param kwargs: Additional arguments to pass to httpx request (like json, headers)
+        :return: Coroutine that yields an httpx.Response object or error message as a string
+        """
+        async with httpx.AsyncClient(base_url=self.base_url, timeout=self.timeout) as client:
+            try:
+                response = await client.request(method, endpoint, **kwargs)
+                return response
+            except httpx.RequestError as e:
```
sfoster1: Same comments as for the synchronous method.

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+            response = self.client.request(method, url, **kwargs)
+            return response
+        except httpx.RequestError as e:
+            logger.error(f"An error occurred while requesting {e.request.url!r}.")
+            logger.exception(e)
+        except httpx.HTTPStatusError as e:
+            logger.error(f"Non-success status code received: {e.response.status_code} for URL: {e.request.url!r}")
+            logger.exception(e)
+        finally:
+            self.client.close()
```
sfoster1: is there a solid reason to close after every request? in theory, one of the hooks of httpx is that it has implicit connection pooling.

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+import httpx
+from aws_lambda_powertools import Logger
+
+from api.settings import Settings
+
+settings: Settings = Settings()
+logger: Logger = Logger(service=settings.service_name)
+
+
+class HTTPClient:
```
sfoster1: I think this class may be on the wrong side of the balance of effort vs usefulness. It's capturing three parameters so you don't have to specify them every time, which is useful, but in return you have to go reimplement every httpx client interaction method that you want to use elsewhere and either tediously type out the same function signatures because mypy doesn't have `typeof` or do as you did an `**kwargs: Any` type-erase the result. I'd love to see a bit more of a balance in terms of preserving the information `httpx` provides, maybe via annotating these methods with `functools.wrap` on the client methods or something.

opentrons-ai-server/api/integration/http_client.py
```diff
@@ -0,0 +1,80 @@
...
+    def __init__(
+        self,
+        base_url: str,
+        timeout: float = 10.0,
```
sfoster1: nit: could be `timeout_s` and `connect_timeout_s` and need a little less docs

opentrons-ai-server/api/handler/function.py
```diff
@@ -0,0 +1,49 @@
...
+@tracer.capture_method
+def get_health() -> Response[Any]:
+    logger.info("GET health")
+    return Response(status_code=HTTPStatus.OK, content_type=content_types.APPLICATION_JSON, body={"version": "0.0.1"})
+
+
+# You can continue to use other utilities just as before
+@logger.inject_lambda_context(correlation_id_path=correlation_paths.API_GATEWAY_HTTP, log_event=True)
+@tracer.capture_lambda_handler
+def handler(event: dict[Any, Any], context: LambdaContext) -> dict[Any, Any]:
```
sfoster1, line 48: wait do we really not get python types of lambda events out of using this lambda powertool thing? jeez

opentrons-ai-server/api/handler/function.py
```diff
@@ -0,0 +1,49 @@
...
+            content_type=content_types.APPLICATION_JSON,
+            body={"error": str(e), "message": "An error occurred while retrieving todos"},
+        )
+
+
+@app.get("/health")  # type: ignore[misc]
+@tracer.capture_method
+def get_health() -> Response[Any]:
+    logger.info("GET health")
+    return Response(status_code=HTTPStatus.OK, content_type=content_types.APPLICATION_JSON, body={"version": "0.0.1"})
```
sfoster1, line 42: obv haven't implemented the full versioning thing yet, but maybe have a `__version__.py` that starts out handwritten and will be generated in the future and import that here for this response?

sfoster1 (changes requested):
The structure looks good and I agree with the handler/domain/integration separation, but I think we can tune up a lot about the more full-featured stuff in here like the http client wrapper and the deploy script.

opentrons-ai-server/api/settings.py
y3rsh: yes, fixed

opentrons-ai-server/requirements.txt
y3rsh: yes... it felt awkward, made it part of the build, removed the tracking

opentrons-ai-server/deploy.py
y3rsh: yes this is better

opentrons-ai-server/deploy.py
y3rsh: :heart yes.

opentrons-ai-server/deploy.py
y3rsh: like much better.

opentrons-ai-server/deploy.py
y3rsh: yes.

opentrons-ai-server/deploy.py
y3rsh: fixed

opentrons-ai-server/api/integration/http_client.py
y3rsh: yup not needed.

opentrons-ai-server/api/integration/http_client.py
y3rsh: no

opentrons-ai-server/api/integration/http_client.py
y3rsh: removed

opentrons-ai-server/api/integration/http_client.py
y3rsh: ya again not needed

opentrons-ai-server/api/integration/http_client.py
y3rsh: got it.

opentrons-ai-server/api/integration/http_client.py
y3rsh: Ya, I was trying to get some reasonable defaults.  This will need to be used for huggingface unless they have a client like openai, I have not looked.  See if the new version is a little better.

opentrons-ai-server/api/integration/http_client.py
y3rsh: fixed

opentrons-ai-server/api/integration/http_client.py
y3rsh: This was totally unnecessary.  Removed it. httpx typing is great.

opentrons-ai-server/api/handler/function.py
y3rsh: yes for sure will be dynamic and able to be tied to code, will do in a separate PR

opentrons-ai-server/api/handler/function.py
y3rsh: I think we can, I did not dig in enough on the event.  Ours will always be 1 type of event as things are now but all their examples allow events from multiple services and I left it.

.github/workflows/ai-server-lint-test.yaml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          python-version: '3.12'
+          cache: 'pipenv'
+          cache-dependency-path: opentrons-ai-server/Pipfile.lock
+      - name: Setup
+        id: install
+        working-directory: ./opentrons-ai-server
+        run: make setup
+      - name: black-check
+        if: always() && steps.install.outcome == 'success' || steps.install.outcome == 'skipped'
```
sfoster1, line 38: what's going on with this logic?

sfoster1 (approved):
Looks good! What's up with the logic in the `if` blocks in the workflow? I can't quite see like what the purpose of the `always()` is and why we'd want to run the subsequent steps if the install was skipped?

.github/workflows/ai-server-lint-test.yaml
y3rsh: The use of always() combined with the outcomes of the install step ensures that the step will always run if install was either successful or skipped, but not if install failed. 

So the step runs even if a previous step failed except if the install step fails.

PR #15089, opened 2024-05-03
refactor(robot-sever,api): make ChangeNotifier thread safe

mjhuff:
Closes [EXEC-404](https://opentrons.atlassian.net/browse/EXEC-404)

# Overview

We use ChangeNotifier as a wrapper around asyncio.Event, which lets the robot-server be notified of changes within PE. This is sufficient when the notifier and the waiter operate in the same thread, but is not inherently multi-thread safe. We can extend Event to make it thread safe by utilizing asyncio's internal event loop call_soon_threadsafe method.

`RunsPublisher` now takes a `ChangeNotifier` or a `ChangeNotifier_ts`. `ChangeNotifier` is moved to a better home in `api/utils`, and the redundant interfaces/tests are removed from robot-server and api. There's some renaming to reflect the upcoming `hardware_notify_publisher` equivalent. 

This PR is a follow-up to #14952. After discussing with @SyntaxColoring in person, we don't really need a full-blown message-capable thread-safe notification interface for communicating between HWC and robot-server. That opens up to problems such as backpressure and potential longer-than-expected resource locking. Most importantly, we can likely bubble HWC notes to PE via a callback that can be architected in such a way that we don't have to worry about resource locking. If this really turns out not to be the case, notifications don't need a message-capable interface, anyway, so we probably want a different interface to reflect that.

# Test Plan
- Verified PE events are emitted as expected to robot-server. 
- Although this PR does not implement HWC notifications, I have a POC of this `ChangeNotifier_ts` that demonstrates this working with toggling the lights. You can find it [here](https://github.com/Opentrons/opentrons/tree/HWC-BUBBLE) if you're curious. 

# Risk assessment
low

[EXEC-404]: https://opentrons.atlassian.net/browse/EXEC-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Yeah, that looks like a good change and a good new place for it.

PR #15091, opened 2024-05-04
fix(app): Fix dropdown menu open direction issue

koji:
# Overview
add open direction to dropdown menu
close RQA-2662

# Test Plan
run RTP protocol and try to open dropdown menus

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/atoms/MenuList/DropdownMenu.tsx
+++ b/app/src/atoms/MenuList/DropdownMenu.tsx
@@ -64,6 +64,46 @@ export function DropdownMenu(props: DropdownMenuProps): JSX.Element {
   } = props
   const [targetProps, tooltipProps] = useHoverTooltip()
   const [showDropdownMenu, setShowDropdownMenu] = React.useState<boolean>(false)
+
+  const [dropdownPosition, setDropdownPosition] = React.useState<
+    'top' | 'bottom'
+  >('bottom')
+
+  React.useEffect(() => {
+    const handlePositionCalculation = () => {
+      const dropdownRect = dropDownMenuWrapperRef.current?.getBoundingClientRect()
+      if (dropdownRect) {
+        const parentElement = dropDownMenuWrapperRef.current.parentElement
+        const grandParentElement = parentElement?.parentElement
+        let availableHeight = window.innerHeight
+        let scrollOffset = 0
+
+        if (grandParentElement) {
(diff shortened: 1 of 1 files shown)


app/src/atoms/MenuList/DropdownMenu.tsx
```diff
@@ -64,6 +64,46 @@ export function DropdownMenu(props: DropdownMenuProps): JSX.Element {
...
+
+    window.addEventListener('resize', handlePositionCalculation)
+    window.addEventListener('scroll', handlePositionCalculation)
+    handlePositionCalculation()
+
+    return () => {
+      window.removeEventListener('resize', handlePositionCalculation)
+      window.removeEventListener('scroll', handlePositionCalculation)
+    }
+  }, [filterOptions.length])
```
ncdiehl11: Looks good but I think we may want to change the dependency array to include the available height? In testing, I zoomed such that the menu should pop up on top and it looks good. But then, if you zoom out such that there is room for the menu to pop up on bottom, it still pops up on top.

koji: added `window.innerHeight` since availableHeight is a local var but windw.innerHeight change has an impact on the calc.

ncdiehl11: I think the intuition about grabbing the grandparent element is correct, but the grandparent implementation in the use of the dropdown menu is not the whole slideout
Screenshot 2024-05-07 at 2 22 32 PM
Maybe we can have an optional container reference to pass into `DropdownMenu` and default to that if it is not undefined?

koji: What do you think about this?
```ts
 setDropdownPosition(
          dropdownBottom > availableHeight &&
            window.innerHeight < availableHeight
            ? 'top'
            : 'bottom'
        )
```

ncdiehl11: I think this is almost perfect except we are not accounting for the footer 
Screenshot 2024-05-07 at 5 24 12 PM

koji: @ncdiehl11 
updated. thank you for your help!

ncdiehl11 (approved):
working as expected 🤝

PR #15094, opened 2024-05-06
fix(app, app-shell, app-shell-odd): block initial HTTP request until successful MQTT subscription

mjhuff:
Closes [EXEC-429](https://opentrons.atlassian.net/browse/EXEC-429)

# Overview

Currently, the desktop app/ODD attempt to address the ["missed update problem"](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/917110846/Old+proposal#Nuance%3A-Avoiding-missed-first-updates) in the following way: while we subscribe to a topic, we simultaneously GET whatever equivalent HTTP resource we just subscribed to. However, there's definitely a world (albeit a very small one) in which we receive the HTTP response, a server update occurs, the server publishes, and _then_ we successfully subscribe to a topic. In this world, we've missed the update event. 

Solve for this by simply blocking the initial HTTP GET until we subscribe. While the subscribe handshake could theoretically take a maximum of 2 seconds (at which point we forcefully timeout the subscribe action and fallback to polling), in practice it's more like <~250ms (wall-clock request to response). We already handle failed connections and don't go through this handshake if we can't connect to the client to begin with, so the "wait 2 second until sub failure" scenario shouldn't realistically happen. 

### Edge vs. 7.3.0?

After discussing with @SyntaxColoring, we believe it's best not to include the fix in 7.3.0, since there's always the possibility we cause a worse problem than we solve (although I feel good that this isn't the case). The worst case scenario is the client very very rarely doesn't receive an update notification as expected. 

In practice, this problem has yet to be reported by QA or identified by a dev, so I think it's better to hold off. I'm willing to be convinced otherwise, though. 

# Test Plan
- Verified behavior works as expected. We block until the subscription ACK, then we send the initial request. 

# Changelog
- Fixed an edge-case in which a client could not receive MQTT updates. 

# Risk assessment
Medium(ish). See subsection in overview.

[EXEC-429]: https://opentrons.atlassian.net/browse/EXEC-429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell-odd/src/notifications/deserialize.ts
+++ b/app-shell-odd/src/notifications/deserialize.ts
@@ -31,6 +31,10 @@ export function sendDeserialized(
   } catch {} // Prevents shell erroring during app shutdown event.
 }
 
+export function sendDeserializedRefetch(topic: NotifyTopic): void {
+  sendDeserialized(topic, { refetch: true })
+}
+
 export function sendDeserializedGenericError(topic: NotifyTopic): void {
   sendDeserialized(topic, FAILURE_STATUSES.ECONNFAILED)
 }
--- a/app-shell-odd/src/notifications/subscribe.ts
+++ b/app-shell-odd/src/notifications/subscribe.ts
@@ -1,7 +1,11 @@
 import mqtt from 'mqtt'
 
 import { connectionStore } from './store'
-import { sendDeserialized, sendDeserializedGenericError } from './deserialize'
+import {
+  sendDeserialized,
+  sendDeserializedGenericError,
+  sendDeserializedRefetch,
+} from './deserialize'
 import { notifyLog } from './notifyLog'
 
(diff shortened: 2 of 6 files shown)


sfoster1 (changes requested):
I'm a little hesitant to introduce more latency here. What if instead we send the synthetic refresh _again_ after we establish the connection? We have an extra refresh, but we don't extend the so-called time to first light, and we're still way better off than when we were just polling.

mjhuff:
> I'm a little hesitant to introduce more latency here. What if instead we send the synthetic refresh _again_ after we establish the connection?

Yeah that makes a lot of sense.

sfoster1 (approved):
Very good, looks good to me!

PR #15096, opened 2024-05-06
refactor(protocol-designer, step-generation): multi tiprack uses defURI instead of id

jerader:
closes AUTH-12

# Overview

If you import a protocol from an older version of PD that uses several of the same tipracks, the migration just assigned the same tiprack to each pipetting step even though the tiprack might not have tips. Step-generation was smart enough to decide which tip for the commands though but the step form would be wrong. 

This PR fixes this issue by simplifying the tiprack in stepForm's value from the tiprack ID to the tiprack Def uri. It required a bit of refactoring but wasn't too bad. I think it is stronger code this way.

# Test Plan

Upload the attached protocol into PD and see that it goes through the migration (should see the migration modal) Then click on the timeline and hover over both steps. The correct tiprack should be highlighted and there should be no errors.

[multipleTipracksForOnePipette.json](https://github.com/Opentrons/opentrons/files/15223630/multipleTipracksForOnePipette.json)

# Changelog

- fix PD components so the step form populates the dropdown and values with the tiprack def uri instead of the tiprack id
- refactor replaceTip and getNextTip logic slightly to account for the tiprack uri
- refactor the logic to tell if there is a pipette collision to use tiprack def uri instead of id
- fix the migration for `8_1_0` to populate the tiprack value
- fix up all affected unit tests and cypress tests

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
+++ b/protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
@@ -30,13 +30,14 @@ export function TiprackField(props: TiprackFieldProps): JSX.Element {
   const [targetProps, tooltipProps] = useHoverTooltip()
   const pipetteEntities = useSelector(getPipetteEntities)
   const options = useSelector(uiLabwareSelectors.getTiprackOptions)
-  const defaultTipracks =
+  const defaultTiprackUris =
     pipetteId != null ? pipetteEntities[pipetteId as string].tiprackDefURI : []
-  const pipetteOptions = options.filter(option =>
-    defaultTipracks.includes(option.defURI)
+  const tiprackOptions = options.filter(option =>
+    defaultTiprackUris.includes(option.value)
   )
-  const hasMissingTiprack = defaultTipracks.length > pipetteOptions.length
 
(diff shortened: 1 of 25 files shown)


protocol-designer/src/components/StepEditForm/fields/TiprackField.tsx
```diff
@@ -30,21 +30,22 @@ export function TiprackField(props: TiprackFieldProps): JSX.Element {
...
     pipetteId != null ? pipetteEntities[pipetteId as string].tiprackDefURI : []
-  const pipetteOptions = options.filter(option =>
-    defaultTipracks.includes(option.defURI)
+  const tiprackOptions = options.filter(option =>
+    defaultTiprackUris.includes(option.value)
   )
-  const hasMissingTiprack = defaultTipracks.length > pipetteOptions.length
 
+  const hasMissingTiprack = defaultTiprackUris.length > tiprackOptions.length
+  console.log('tiprackOptions', tiprackOptions)
```
koji: forgot to remove?

koji (approved):
the changes look good to me.

PR #15099, opened 2024-05-06
fix(app): fix choose number input field behavior

koji:
# Overview
fix choose number input field behavior

https://github.com/Opentrons/opentrons/assets/474225/c29f1a4b-d289-4e02-ad97-f3d126193bcb

close RQA-2651

# Test Plan

# Changelog

# Review requests

# Risk assessment


ncdiehl11 (approved):
tested and looks good, including an error toast if I try to go back with an empty value. nice

PR #15100, opened 2024-05-06
fix(protocol-designer): default tiprack selection refreshing

jerader:
closes AUTH-382 AUTH-383

# Overview

Fixes 2 bugs related to tipracks:
1. if you select a bunch of tipracks then go back and change the pipette, the tipracks wouldn't be reset so this PR resets the tipracks selected whenever you rerender the tiprack selection page
2. if you have 3 or more default tipracks with the thermocycler for the OT-2, 2 tipracks would be placed in the TC's slots 8 and 11. this PR fixes that

# Test Plan

Create an ot-2 protocol and select 2 tipracks for the pipette. Then go back and select a different pipette and see that the tipracks selected is reset to the 1st one. Then select 2 more tipracks so you have a total of 3 selected. On the modules page of the wizard, add a Thermocycler GEN1 or GEN2. Create the protocol. Go to the deck map and see that slots 2, 4 and 1 are occupied for the tipracks.

# Changelog

- add branching logic to accommodate tiprack slots when the TC is selected for ot-2
- reset the tipracks selected in the pipette tip tile

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/PipetteTipsTile.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/PipetteTipsTile.tsx
@@ -182,12 +182,10 @@ function PipetteTipsField(props: PipetteTipsFieldProps): JSX.Element | null {
   const selectedValues = pipettesByMount[mount].tiprackDefURI ?? []
 
   React.useEffect(() => {
-    if (selectedValues.length === 0) {
-      setValue(`pipettesByMount.${mount}.tiprackDefURI`, [
-        tiprackOptions[0]?.value ?? '',
-      ])
-    }
-  }, [selectedValues, setValue, tiprackOptions])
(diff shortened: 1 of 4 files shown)


koji:
@jerader 
>see that slots 2, 4 and 1 are occupied for the tipracks.


actually, slots 1,2, and 5 are occupied on the sandbox,

Screenshot 2024-05-06 at 2 48 37 PM

jerader:
> @jerader
> 
> > see that slots 2, 4 and 1 are occupied for the tipracks.
> 
> from the code, they would be 1, 2, 5? actually, they are occupied on the sandbox,
> 

@koji this is my mistake! it should be 2, 5 and 1! Your screenshot is expected 😄

koji (approved):
lgtm!

PR #15103, opened 2024-05-06
fix(robot-server): clear ChangeNotifier's internal event immediately after waiting

mjhuff:
Closes [RQA-2677](https://opentrons.atlassian.net/browse/RQA-2677)

# Overview

ChangeNotifier is the interface for alerting robot-server of notification updates deriving from Protocol Engine. Protocol Engine `notify()`ies, and the PublisherNotifier `wait()`s. After the event flag is set() in `notify()`, the PublisherNotifier no longer waits (correctly). However, the event flag isn't cleared immediately on receiving a new event...it's cleared _after_  PublisherNotifer fires all its registered callbacks. This causes some events to be dropped if Protocol Engine tries setting the flag before PublisherNotifier has gotten around to clearing it... woops. 

The reason we haven't seen this problem earlier is because we currently don't have too many immediately back-to-back events firing with publishers that would be interested in those events, but in the specific scenario mentioned in the ticket, the following is happening:

- A play action hits PE, and PE sets the event flag via `notify()`.
- PublisherNotifier starts firing through its registered callbacks.
- The "command succeeded" & "next running command" actions hit PE very shortly after the play action, and the set event flag is set from `True` to `True` redundantly.
- PublisherNotifier finishes its callbacks then sets the event flag to `False`...dropping these new events.

The fix: simply clear the event flag immediately after waiting.

# Test Plan
- You can test this with any protocol containing a manual move labware command + a command afterwards. Here's the [protocol I used.](https://github.com/Opentrons/opentrons/files/15226308/GripperMoveTestingWithPause.py.zip)
- Verify that the Intervention Modal derenders on the ODD immediately after the command is completed (so like 1 second after pressing the button), NOT a whole command after the command proceeding the manual move command. 

# Changelog
- Fixed a bug in which commands after manual move commands were slow to update on the ODD.

# Risk assessment
low

[RQA-2677]: https://opentrons.atlassian.net/browse/RQA-2677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/robot-server/robot_server/service/notifications/change_notifier.py
+++ b/robot-server/robot_server/service/notifications/change_notifier.py
@@ -15,9 +15,5 @@ class ChangeNotifier:
 
     async def wait(self) -> None:
         """Wait until the next change notification."""
-        self._event.clear()
         await self._event.wait()
-
-    def clear(self) -> None:
-        """Reset the internal event flag."""
         self._event.clear()
--- a/robot-server/tests/service/notifications/test_change_notifier.py
+++ b/robot-server/tests/service/notifications/test_change_notifier.py
@@ -54,3 +54,30 @@ async def test_multiple_subscribers(_test_repetition: int) -> None:
     await asyncio.gather(task_1, task_2, task_3)
 
     assert results == [1, 2, 3]
+
+
+async def test_notify_while_busy() -> None:
+    """Test that waiters process a new notify() after they are done being busy."""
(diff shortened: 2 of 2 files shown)


sfoster1 (approved):
Nice, good catch

PR #15104, opened 2024-05-06
chore(api): allow specifying module models to sim

sfoster1:
The simulator setup files let you specify modules, but you can only do it by specifying their types (i.e. magdeck, tempdeck) and not their modules (i.e. magneticModuleV1, temperatureModuleV2). This hasn't really mattered up til now because nothing cares, but RSQ-6 will have both desktop and ODD looking pretty hard at which generation of module is present on a flex, so we need a way to specify.

This adds a model key to the module section of the simulator setup (and sets the test-flex.json and test.json up with it) so that you can specify module models.

Also, while I was at it, improve the type checking of the init function of the threadmanager.

Closes RSQ-6


api/src/opentrons/hardware_control/api.py
```diff
@@ -255,7 +257,7 @@ async def build_hardware_simulator(
         attached_instruments: Optional[
             Dict[top_types.Mount, Dict[str, Optional[str]]]
         ] = None,
-        attached_modules: Optional[Dict[str, List[str]]] = None,
+        attached_modules: Optional[Dict[str, List[Tuple[str, Optional[str]]]]] = None,
```
TamarZanzouri: Is this the proper data structure for this?

TamarZanzouri (approved):
Overall looks good! The only thing I don't like is the list of tuples.

api/src/opentrons/hardware_control/api.py
sfoster1: well, no. i guess this is the point where the dataclasses should get persisted through.

api/src/opentrons/hardware_control/api.py
```diff
@@ -255,7 +257,7 @@ async def build_hardware_simulator(
         attached_instruments: Optional[
             Dict[top_types.Mount, Dict[str, Optional[str]]]
         ] = None,
-        attached_modules: Optional[Dict[str, List[str]]] = None,
+        attached_modules: Optional[Dict[str, List[modules.SimulatingModule]]] = None,
```
TamarZanzouri, line 260: Much better 😀

api/src/opentrons/hardware_control/backends/simulator.py
```diff
@@ -71,8 +71,9 @@ async def build(
                                      This dict should map mounts to either
                                      empty dicts or to dicts containing
                                      'model' and 'id' keys.
-        :param attached_modules: A list of module model names (e.g.
-                                 `'tempdeck'` or `'magdeck'`) representing
+        :param attached_modules: A map of module type names (e.g.
+                                 `'tempdeck'` or `'magdeck'`) to lists of (serial, model)
+                                 tuples representing
```
TamarZanzouri: It's not tuples anymore

api/src/opentrons/hardware_control/scripts/repl.py
```diff
@@ -114,8 +114,6 @@ def synchronizer(*args: Any, **kwargs: Any) -> Any:
     def build_thread_manager() -> ThreadManager[Union[API, OT3API]]:
         return ThreadManager(
             API.build_hardware_controller,
-            use_usb_bus=ff.rear_panel_integration(),
```
TamarZanzouri, line 117: Mmm is this removal ok?

api/src/opentrons/hardware_control/__init__.py
```diff
@@ -38,7 +38,8 @@
 ]
 HardwareControlAPI = Union[OT2HardwareControlAPI, OT3HardwareControlAPI]
 
-ThreadManagedHardware = ThreadManager[HardwareControlAPI]
+# this type ignore is because of https://github.com/python/mypy/issues/13437
+ThreadManagedHardware = ThreadManager[HardwareControlAPI]  # type: ignore[misc]
```
TamarZanzouri, line 42: Just for me why do we need this type: ignore now? Looks like the code is the same?

TamarZanzouri (approved):
Amazing! Looks great! A few comments but love this

api/src/opentrons/hardware_control/scripts/repl.py
sfoster1: Yup! This was revealed by `ThreadManager.__init__` actually forwarding the types. This code is building an OT2 hardware controller, it doesn't have that argument this code would have crashed (I don't think anybody uses this repl on the OT-2 or indeed the flex anymore, it was made during flex development)

api/src/opentrons/hardware_control/__init__.py
sfoster1: It's this bug in mypy where if you have 

- a class that is `Generic`
- and an `__init__` method that takes a typevar

then you get an error when you try to bind to it.

PR #15107, opened 2024-05-06
feat(opentrons-ai-client): add primary floating button

koji:
# Overview
add primary floating button

close AUTH-384

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/opentrons-ai-client/src/molecules/PrimaryFloatingButton/PrimaryFloatingButton.stories.tsx
+++ b/opentrons-ai-client/src/molecules/PrimaryFloatingButton/PrimaryFloatingButton.stories.tsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import { PrimaryFloatingButton as PrimaryFloatingButtonComponent } from './index'
+
+import type { Meta, StoryObj } from '@storybook/react'
+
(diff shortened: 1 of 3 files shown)


ncdiehl11 (approved):
LGTM, couple of questions: do we need to account for no icon? Do we need to set a minimum width?

koji:
> LGTM, couple of questions: do we need to account for no icon? Do we need to set a minimum width?

Actually, this component is existing ui components but it has been implemented yet for Desktop app.
There is the button type without icon at this moment. (Seems that still designing it)
https://www.figma.com/file/l7BAJAGICr10ELsSqQD0Sr/Design-System%3A-Desktop?type=design&node-id=13031-84732&mode=design&t=hwuihVyEdnKK9Bjc-0

Sue asked me to create Primary Floating Button for `copy code` button.
https://www.figma.com/file/6de3DHeR9o4q56Hd4S3YTb/Playground%3A-OpentronsAI?type=design&node-id=1229-4656&mode=design&t=N6wGjjI8lo3ufjKZ-0


For the second question, having min width might be good but the thing is that I don't know the other usage cases except opentrons-ai. I think we can add min width as a prop when we need to use the component for Desktop app or other web application.

PR #15111, opened 2024-05-07
feat(test-data-generation): generate deck configuration protocols

DerekMaggio:
# Overview

Generate Python protocols from Hypothesis-generated Deck Configurations.

Closes [RQA-2617](https://opentrons.atlassian.net/browse/RQA-2617
)

# Test Plan

- [x] Run analysis on a few of the generated Python protocols and ensure they are valid Python

# Changelog

- Create `ast_helpers.py` which is an abstraction layer that handles creating [Python ast nodes](https://docs.python.org/3.10/library/ast.html#node-classes). 
- Create `generation_phases` dir. Logic inside the directory, utilizes `ast_helpers.py` to map the test data logic to ast nodes
  - Create `setup_phase.py` which defines the boilerplate code that every protocol needs: importing the opentrons, package, creating requirements dictionary, creating run function
  - Create `load_phase.py` which evaluated the Deck Configuration and generates the `load_*` calls to load in labware, modules, and fixtures to the protocol
  - Create `call_phase.py` which makes calls to loaded entities to ensure analysis recognizes them as being on the deck

- Create `python_protocol_generator.py` which calls `generate_ast` on all of the `ast_helpers` and wires up the created nodes. Also handles calling into the `astor` package to generate the python code
- Added different Makefile targets to run generated tests in a more verbose mode

# Review requests

Is there a better way, than the `phase` logic, to handle figuring out what python statements to generate?

# Risk assessment

Low

[RQA-2617]: https://opentrons.atlassian.net/browse/RQA-2617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test-data-generation/src/test_data_generation/deck_configuration/strategy/helper_strategies.py
+++ b/test-data-generation/src/test_data_generation/deck_configuration/strategy/helper_strategies.py
@@ -6,14 +6,16 @@ from test_data_generation.deck_configuration.datashapes import (
     Row,
     Slot,
     PossibleSlotContents as PSC,
+    RowName,
+    ColumnName,
 )
 
 
 @st.composite
 def a_slot(
     draw: st.DrawFn,
-    row: str,
-    col: str,
+    row: RowName,
+    col: ColumnName,
     content_options: List[PSC] = PSC.all(),
 ) -> Slot:
     """Generate a slot with a random content.
@@ -39,7 +41,7 @@ def a_slot(
         if not content.is_a_staging_area()
     ]
 
-    if col == "1" and (row == "A" or row == "B"):
+    if col == "1" and (row == "a" or row == "b"):
         return draw(
             st.builds(
                 Slot,
@@ -49,7 +51,7 @@ def a_slot(
             )
         )
     elif col == "3":
-        if row == "D":
+        if row == "d":
             return draw(
                 st.builds(
                     Slot,
@@ -83,7 +85,7 @@ def a_slot(
 @st.composite
 def a_row(
     draw: st.DrawFn,
-    row: str,
+    row: RowName,
     content_options: List[PSC] = PSC.all(),
 ) -> Row:
     """Generate a row with random slots."""
@@ -101,7 +103,7 @@ def a_row(
 @st.composite
 def a_column(
     draw: st.DrawFn,
-    col: str,
+    col: ColumnName,
     content_options: List[PSC] = PSC.all(),
 ) -> Column:
     """Generate a column with random slots."""
--- a/test-data-generation/src/test_data_generation/deck_configuration/strategy/helper_strategies.py
+++ b/test-data-generation/src/test_data_generation/deck_configuration/strategy/helper_strategies.py
@@ -156,12 +156,11 @@ def a_column(
 def a_deck_by_columns(
     draw: st.DrawFn,
     thermocycler_on_deck: bool | None = None,
-    col_1_contents=PSC.all(),
-    col_2_contents=PSC.all(),
-    col_3_contents=PSC.all(),
+    col_1_contents: typing.List[PSC] = PSC.all(),
+    col_2_contents: typing.List[PSC] = PSC.all(),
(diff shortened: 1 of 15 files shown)


caila-marashaj (approved):
Awesome!

DerekMaggio:
Pushing this back to draft for a bit to make some fixes.

test-data-generation/src/test_data_generation/deck_configuration/strategy/helper_strategies.py
```diff
@@ -39,17 +37,17 @@ def a_slot(
         if not content.is_a_staging_area()
     ]
 
-    if col == "1" and (row == "A" or row == "B"):
+    if thermocycler_on_deck and col == "1" and (row == "a" or row == "b"):
```
y3rsh: I'd like a comment here of the why.

y3rsh (approved):
Sweet action.

test-data-generation/src/test_data_generation/python_protocol_generation/generation_phases/load_phase.py
```diff
@@ -0,0 +1,248 @@
...
+        ),
+    )
+    # well_plate_{slot}.is_tiprack
+
+
+def create_deck_slot_load_statement(
+    slot: Slot,
+) -> ast_h.AssignStatement | typing.List[ast_h.AssignStatement]:
+    """Maps the contents of a slot to the correct assign statement."""
+    match slot.contents:
```
sfoster1, line 168: 👀 👀 👀 👀 👀 👀 👀 👀

test-data-generation/src/test_data_generation/python_protocol_generation/ast_helpers.py
```diff
@@ -0,0 +1,137 @@
...
+class CanGenerateAST(typing.Protocol):
+    """Protocol for objects that can generate an AST node."""
+
+    def generate_ast(self) -> ast.AST:
+        """Generate an AST node."""
+        ...
+
+
+@dataclass
+class ImportStatement:
```
sfoster1, line 21: IMO it would be nice to cut down the number of things to override here - for instance, if you defined constants like `ProtocolContextMethods` as `Literal[...]` instead of enums, then you wouldn't need the string conversions

test-data-generation/src/test_data_generation/python_protocol_generation/python_protocol_generator.py
```diff
@@ -0,0 +1,87 @@
...
+            pipette_load_statements
+            + deck_slot_load_statements
+            + calls_to_loaded_entities
+        )
+
+        for statement in statements_to_make:
+            run_function.body.append(statement.generate_ast())
+        module.body.append(run_function)
+
+        return str(astor.to_source(module))
```
sfoster1, line 87: is this the only thing we use `astor` for? we're not building something that complex (in generated source code terms), so could we just use [ast.unparse](https://docs.python.org/3/library/ast.html#ast.unparse)

sfoster1 (approved):
This is extremely awesome, nice work!

SyntaxColoring (approved):
> Is there a better way, than the phase logic, to handle figuring out what python statements to generate?

This all looks reasonable to me, yeah. Thank you!

test-data-generation/src/test_data_generation/python_protocol_generation/ast_helpers.py
DerekMaggio: I have that in code I am working on now. 
So I will include those changes as part of a followup PR.

test-data-generation/src/test_data_generation/python_protocol_generation/python_protocol_generator.py
DerekMaggio: There was some issue with that, but I can't remember what it was. 
I will investigate just using `ast.unparse` again.

PR #15113, opened 2024-05-07
Saves robot logs as .log when reporting error

rclarke0:
# Overview

Saves robot logs as .log when reporting error

# Test Plan

Ran script on robot and compared log format to logs that are downloaded through the app.

# Changelog

In the get_robot_logs function I changed the files to be saved as .log and removed json.dumps() 

# Review requests

# Risk assessment

### DIFF
--- a/abr-testing/abr_testing/automation/jira_tool.py
+++ b/abr-testing/abr_testing/automation/jira_tool.py
@@ -84,9 +84,10 @@ class JiraTicket:
                 # Include other required fields as needed
             }
         }
+        print(data)
         try:
             response = requests.post(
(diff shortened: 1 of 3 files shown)


abr-testing/abr_testing/automation/jira_tool.py
```diff
@@ -84,9 +84,10 @@ def create_ticket(
                 # Include other required fields as needed
             }
         }
+        print(data)
```
vegano1: debug print?

vegano1 (approved):
This is great, thank you!

PR #15118, opened 2024-05-07
feat(api): add thermocycler control script

caila-marashaj:
## Overview
This is a debugging script that allows us to easily communicate with a thermocycler in a similar fashion to `can_control` or `ot3repl`.  It's just a control loop where the user is prompted to send a thermocycler command, and the script waits for and sends back the thermocycler's response. 

You can pass in any gcode command the thermocycler will recognize, and alternatively, there are a few shortcuts for commonly used debugging commands. For example, you can enter `ol` or `cl` to open or close the lid.

If there's a need for it in the future, this can likely also be modified to include other modules without that much change to the overall structure. 

## Review Requests

- Is this an appropriate file location for this?

## Update Notes
This is only meant for use on a robot. Once this is merged I'll write a doc explaining what this is and how to use it.


api/src/opentrons/hardware_control/scripts/tc_control.py
```diff
@@ -0,0 +1,86 @@
...
+            dev.write(f"{command}\n")
+            print(await message_return(dev))
+        except TypeError:
+            print("Invalid input.")
+    return _exit
+
+
+async def _main():
+    tc_name = subprocess.check_output(
+        ["find", "/dev/", "-name", "*thermocycler*"]
```
sfoster1: fyi this will only work on a robot. we get these `/dev/` things that are named after modules because we specify udev rules; it won't weven work on a non-robot linux computer.

caila-marashaj: I might want to do a `glob` instead right

sfoster1: or just document that it's only going to work on a robot if that's all it's intended to do. To make it work off of a robot, you'll have to use e.g. [`pyserial.tools.list_ports.comports`](https://pyserial.readthedocs.io/en/latest/tools.html) rather than relying on files in particular directories

sfoster1 (approved):

PR #15119, opened 2024-05-07
refactor(app, api-client, shared-data): Set command intent to "fixit"

mjhuff:
Closes [EXEC-431](https://opentrons.atlassian.net/browse/EXEC-431)

# Overview

Before we can POST fixit commands during Error Recovery flows, we need to refactor `useChainRunCommands` to support sending the params necessary to POST a fixit command. In addition to a normal `CreateCommand`, this requires sending:
-  A `failedCommandId`
- An explicit `fixit` intent

Because the FE never explicitly sets `intent` in the app until now, it feels correct to keep the `intent: fixit` logic abstracted when creating a new command from the app. Open to thoughts/comments/concerns. 

# Test Plan
- We're covered by new tests and there are no effective changes. However, [here is a branch](https://github.com/Opentrons/opentrons/tree/CHAIN-FIXIT) with a working 'home' command during ER when you click "resume". You'll have to enable the backend & frontend FFs for this to work.

# Review requests
Are the `shared-data` types changes correct?

# Risk assessment
low

[EXEC-431]: https://opentrons.atlassian.net/browse/EXEC-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api-client/src/runs/commands/types.ts
+++ b/api-client/src/runs/commands/types.ts
@@ -43,6 +43,7 @@ export interface CommandsAsPreSerializedListData {
 export interface CreateCommandParams {
   waitUntilComplete?: boolean
   timeout?: number
+  failedCommandId?: string
 }
 
 export interface RunCommandError {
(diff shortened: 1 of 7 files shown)


sfoster1 (approved):
I think this is the right way to go, yeah. I'd be open to having a second function or something but I think this is a lot better.

PR #15122, opened 2024-05-07
feat(hardware): implement multipass lld

ryanthecoder:
# Overview
Due to limits on how far the z stage can move depending on a combination of factors including z speed, plunger speed and labware depth we may need to run the probe multiple times in order to find the liquid. 

This PR lets the liquid probe script handle when the liquid is not found and start a new pass that starts at the z height the the previous pass stopped at.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -140,6 +140,7 @@ from opentrons.hardware_control.types import (
     EstopState,
     HardwareEventHandler,
     HardwareEventUnsubscriber,
+    LiquidNotFound,
 )
 from opentrons.hardware_control.errors import (
     InvalidPipetteName,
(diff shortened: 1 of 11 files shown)


api/src/opentrons/hardware_control/types.py
```diff
@@ -697,14 +697,10 @@ def __init__(
 class LiquidNotFound(RuntimeError):
```
sfoster1: Definitely needs to be an `EnumeratedError` with its own code in shared-data

sfoster1 (changes requested):
Let's begin as we mean to go on and get that error as a proper enumerated error defined in shared data; otherwise, lgtm

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1400,7 +1400,12 @@ async def liquid_probe(
         for node, point in positions.items():
             self._position.update({node: point.motor_position})
             self._encoder_position.update({node: point.encoder_position})
-        if positions[head_node].move_ack == MoveCompleteAck.complete_without_condition:
+        print(f"positions {positions}")
```
sfoster1: stray

sfoster1 (changes requested):

api/src/opentrons/hardware_control/types.py
ryanthecoder: Alright I think I did this right?

sfoster1 (approved):
Nice, thanks! I could see arguments for this either being 2-series or 3-series but I'm not super bothered about it.

sfoster1:
@ryanthecoder I think you have to run format-js to format the error codes json file

PR #15124, opened 2024-05-07
feat(app): add summary screen to quick transfer flow

smb2268:
fix PLAT-180

# Overview
This adds the orchestration component that will house the overview, advanced settings, and tip management for quick transfer. It also completes the overview tab.

# Test Plan
Run through a quick transfer flow. After the volume entry screen, see that you now get an overview of entered information. Advanced settings and Tip management tabs are clickable but don't have any data. Create transfer button behaves how the volume entry continue button did prior to this PR, logging the entire state.
Screen Shot 2024-05-07 at 2 55 40 PM
Screen Shot 2024-05-07 at 2 49 06 PM

# Changelog
1. Create `SummaryAndSettings` component which will be the 9th and final step of the wizard flow and manage the input of the remaining data
2. Add `Overview` component which displays the required information
3. Change copy on deck map of first screen of the flow to say "source" and "destination" instead of "labware" as requested by design
4. Fixed issue with the indeces of the wizard steps not matching up with the StepMeter
5. Added & updated tests as needed

# Review requests
Quick code check

# Risk assessment
Low, behind FF


app/src/organisms/QuickTransferFlow/utils.ts
```diff
@@ -67,6 +67,18 @@ export function quickTransferReducer(
       }
     }
     case 'SET_DEST_WELLS': {
+      let transferType = 'transfer'
+      if (
+        state.sourceWells != null &&
+        state.sourceWells.length > action.wells.length
+      ) {
+        transferType = 'distribue'
```
mjhuff: typo?

app/src/organisms/QuickTransferFlow/index.tsx
```diff
@@ -60,19 +52,27 @@ export const QuickTransferFlow = (): JSX.Element => {
...
+    <SelectSourceLabware key={3} {...sharedMiddleStepProps} />,
+    <SelectSourceWells key={4} {...sharedMiddleStepProps} />,
+    <SelectDestLabware key={5} {...sharedMiddleStepProps} />,
+    <SelectDestWells key={6} {...sharedMiddleStepProps} />,
+    <VolumeEntry key={7} {...sharedMiddleStepProps} />,
+    <SummaryAndSettings
+      key={8}
+      {...sharedMiddleStepProps}
+      onNext={() => {
+        console.log('final quick transfer flow state:', state)
```
mjhuff, line 72: can't tell if stray log or intentional given pr description

app/src/organisms/QuickTransferFlow/Overview.tsx
```diff
@@ -0,0 +1,78 @@
...
+      value:
+        state.destination === 'source'
+          ? state.source?.metadata.displayName
+          : state.destination?.metadata.displayName,
+    },
+    {
+      option: transferCopy,
+      value: `${state.volume}µL`,
+    },
+  ]
```
mjhuff, line 53: Can several of these `value` values actually be undefined here? If so, we probably want some sort of fallback value if that's true. If they an actually not be undefined, maybe we should update the types?

mjhuff (commented):
Pretty easy to reason about and straightforward. I'll give it a test tomorrow. Couple nits but looks good!

app/src/organisms/QuickTransferFlow/utils.ts
mjhuff: I think creating a `type TransferType = 'transfer' | 'distribute' | 'consolidate'` would be worth it, especially considering this variable gets passed around a few files.

mjhuff (approved):
Looks good. See comments above, but I think it's ready with those considerations!

app/src/organisms/QuickTransferFlow/index.tsx
smb2268: I actually left this one in on purpose - this logs the state when the flow closes so you can see all of the selections that were made. I was planning to leave it in until the protocol creation step is completed

app/src/organisms/QuickTransferFlow/Overview.tsx
smb2268: They will not be undefined at this step in the flow. All of the state values are undefined at step 1, and then get filled out as the flow progresses. Since this is the final step everything will have a value!

PR #15129, opened 2024-05-08
fix(protocol-designer): fix dropdown icon to match legacy icon

jerader:
closes AUTH-392

# Overview

The `menu-down` icon was changed but it affected PD's icons in the dropdown input so this makes a new icon and uses that instead

# Test Plan

verify that the icons in the step form dropdowns match the icons in the same area from PD in production

# Changelog

- add new icon that matches the old `menu-down` path and size
- use new icon in component

# Review requests

see test plan

# Risk assessment

low


koji (approved):
thank you for fixing!

PR #15131, opened 2024-05-08
fix(step-generation): timeline error if multiple entities in same slotName

jerader:
closes RQA-2695

# Overview

This bug exists in production. Basically if you move a labware to a new slot that is available and then go back to starting deck state and add a labware to that slot, there wasn't an error so you have overlapping labware in the same slot name or module slot.

# Test Plan

Upload the attached protocol and see that the move labware step errors until you change the new location

[5_7 Test (1).json](https://github.com/Opentrons/opentrons/files/15252764/5_7.Test.1.json)

# Changelog

- add new error creator
- add test case to move labware atomic command
- add logic for if there are multiple thigns in the same slot or module slot

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #15135, opened 2024-05-08
fix(api): Automatic tip tracking index out of range fix

CaseyBatten:
# Overview

Addresses RQA-2700

Automatic tip tracking when iterating through multiple tip racks did not take into account progression range across those tipracks resulting in the potential for an index out of range error to manifest under certain conditions. These changes seek to fix that.

# Test Plan
Ensure that API 2.18 protocols consume tipracks without encountering an index out of range error manifesting as an out of tips error when iterating through multiple tipracks using tip-cluster based automatic tip tracking. Protocols such as the one below should execute without issue.

```

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18",
}



def run(ctx: protocol_api.ProtocolContext):

    tiprack1000_1 = ctx.load_labware("opentrons_flex_96_tiprack_1000ul", "B2")
    tiprack1000_2 = ctx.load_labware("opentrons_flex_96_tiprack_1000ul", "C2")

    pipette = ctx.load_instrument(
        "flex_96channel_1000", mount="left", tip_racks=[tiprack1000_1, tiprack1000_2]
    )

    trashA3 = ctx.load_trash_bin("A3")

    pipette.configure_nozzle_layout(style=COLUMN, start="A12", tip_racks=[tiprack1000_1, tiprack1000_2])

    # pickup and drop all 12 columns of the first tiprack and the first column of the second tiprack using Column A12 of the pipette (Should progress across the tiprack left to right)
    for x in range(13):
        pipette.pick_up_tip()
        pipette.drop_tip()
    
    pipette.configure_nozzle_layout(style=COLUMN, start="A1", tip_racks=[tiprack1000_1, tiprack1000_2])
    # pickup and drop the remaining 11 columns of the second tiprack using column A1 of the pipette (should progress right to left)
    for x in range(11):
        pipette.pick_up_tip()
        pipette.drop_tip()
```

# Changelog

- Updated the tip cluster logic under tips.py to return None in situations where the iterator would extend beyond the bounds of a tiprack before the iteration logic has been able to evaluate the next upcoming step.

# Review requests

Do these changes appear comprehensive enough given existing safegaurds within the tiprack iteration logic?

# Risk assessment
Low, fixes oversight.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/tips.py
+++ b/api/src/opentrons/protocol_engine/state/tips.py
@@ -312,7 +312,9 @@ class TipView(HasState[TipState]):
                 if critical_row + active_rows < len(columns[0]):
                     critical_row = critical_row + active_rows
                 else:
-                    critical_column = critical_column + 1
+                    critical_column += 1
+                    if critical_column >= len(columns):
+                        return None
                     critical_row = active_rows - 1
             return None
 
@@ -336,7 +338,9 @@ class TipView(HasState[TipState]):
                 if critical_row + active_rows < len(columns[0]):
                     critical_row = critical_row + active_rows
                 else:
-                    critical_column = critical_column - 1
+                    critical_column -= 1
+                    if critical_column < 0:
+                        return None
                     critical_row = active_rows - 1
             return None
 
@@ -360,7 +364,9 @@ class TipView(HasState[TipState]):
                 if critical_row - active_rows >= 0:
                     critical_row = critical_row - active_rows
                 else:
-                    critical_column = critical_column + 1
+                    critical_column += 1
+                    if critical_column >= len(columns):
+                        return None
                     critical_row = len(columns[critical_column]) - active_rows
             return None
 
@@ -384,7 +390,9 @@ class TipView(HasState[TipState]):
                 if critical_row - active_rows >= 0:
                     critical_row = critical_row - active_rows
                 else:
-                    critical_column = critical_column - 1
+                    critical_column -= 1
+                    if critical_column < 0:
+                        return None
                     critical_row = len(columns[critical_column]) - active_rows
             return None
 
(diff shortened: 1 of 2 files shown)


vegano1 (approved):
Logically this makes sense.

api/src/opentrons/protocol_engine/state/tips.py
```diff
@@ -312,7 +312,9 @@ def _cluster_search_A1(active_columns: int, active_rows: int) -> Optional[str]:
                 if critical_row + active_rows < len(columns[0]):
                     critical_row = critical_row + active_rows
                 else:
-                    critical_column = critical_column + 1
+                    critical_column += 1
+                    if critical_column >= len(columns):
```
jbleon95, line 316: This if statement could probably be removed and the `while` check be changed to `while critical_column < len(columns)` if I understand this correctly, since we're not doing any indexing immediately after this.

jbleon95: This if statement could probably be removed and the while check be changed to while critical_column < len(columns) if I understand this correctly, since we're not doing any indexing immediately after this.

api/src/opentrons/protocol_engine/state/tips.py
```diff
@@ -336,7 +338,9 @@ def _cluster_search_A12(active_columns: int, active_rows: int) -> Optional[str]:
                 if critical_row + active_rows < len(columns[0]):
                     critical_row = critical_row + active_rows
                 else:
-                    critical_column = critical_column - 1
+                    critical_column -= 1
+                    if critical_column < 0:
```
jbleon95: Same thing as above, except I don't think you'd need to change anything in the while conditional

jbleon95 (approved):
These changes look like they'll solve the issue with tip tracking. Made a couple suggestions to potentially simplify these changes. If time allows I think a further refactor could simplify a lot of this logic, as well as going further since our whole tip tracking behavior is very spaghetti code to begin with.

PR #15141, opened 2024-05-09
refactor(protocol-designer): tiprack option redesign to not have scroll

jerader:
closes RQA-2681

# Overview

Pipette selection modal now has a better UI for the tiprack selection where the text doesn't require a left/right scroll

Screenshot 2024-05-09 at 10 27 37

# Test Plan

Create a flex or ot-2 protocol and open the edit pipettes section and observe the UI of the tipracks

# Changelog

- change the ui to be similar to selecting tipracks/pipettes/modules in the wizard
- fix the test
- change an icon

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/components/src/forms/Select.tsx
+++ b/components/src/forms/Select.tsx
@@ -118,7 +118,10 @@ function DropdownIndicator(
           [styles.flipped]: props.selectProps.menuIsOpen,
         })}
       >
-        <Icon name="menu-down" className={cx(styles.dropdown_indicator_icon)} />
+        <Icon
+          name="menu-down-pd"
+          className={cx(styles.dropdown_indicator_icon)}
+        />
       </div>
     </reactSelectComponents.DropdownIndicator>
   )
--- a/protocol-designer/src/components/modals/FilePipettesModal/TiprackOption.tsx
+++ b/protocol-designer/src/components/modals/FilePipettesModal/TiprackOption.tsx
@@ -74,29 +74,23 @@ export function TiprackOption(props: TiprackOptionProps): JSX.Element {
     <>
       <Flex
         aria-label={`TiprackOption_flex_${text}`}
-        cursor="pointer"
         onClick={isDisabled ? undefined : onClick}
         flexDirection={DIRECTION_ROW}
         alignItems={ALIGN_CENTER}
+        width="13.5rem"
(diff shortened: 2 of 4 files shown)


components/src/forms/Select.tsx
```diff
@@ -118,7 +118,10 @@ function DropdownIndicator(
           [styles.flipped]: props.selectProps.menuIsOpen,
         })}
       >
-        <Icon name="menu-down" className={cx(styles.dropdown_indicator_icon)} />
+        <Icon
+          name="menu-down-pd"
```
jerader, line 122: i have a feeling this change might affect a few app components - maybe we can come back to find a better fix for this in both the App and PD after 7.3 is out since 7.3 is contained to a release branch

koji:
the scrollbars are shown up on the sandbox and the look is different from your screenshot.
is this expected?
 
Screenshot 2024-05-09 at 1 56 18 PM

jerader:
> the scrollbars are shown up on the sandbox and the look is different from your screenshot. is this expected?
> 

@koji thanks for posting the screenshot! I thought there were no scrollbars 😢 maybe i need to just hide any scrollbars that could appear

protocol-designer/src/components/modals/FilePipettesModal/TiprackOption.tsx
```diff
@@ -1,41 +1,106 @@
...
+          alignItems={ALIGN_CENTER}
+          padding={SPACING.spacing8}
+          border={
+            isSelected && !isDisabled
+              ? BORDERS.activeLineBorder
+              : BORDERS.lineBorder
+          }
+          borderRadius={BORDERS.borderRadius8}
+          cursor={isDisabled ? 'auto' : 'pointer'}
+          backgroundColor={COLORS.transparent}
```
koji: I guess probably adding box or flex for these properties would be good since generally we pass font-size, line-height, color, and font-weight to StyledText.

jerader: ya you're right! i cleaned it up a bit

koji (approved):
left a couple of comments but the changes look good to me.
pipette selection modal looks great!

PR #15144, opened 2024-05-09
fix(step-generation): h-S collision warning for Flex protocols

jerader:
closes RQA-2596

# Overview

Apparently we do need H-S collision warnings for Flex protocols but only if you are pipetting next to a heater-shaker that is in motion. So This PR adds that logic in to trigger the existing timeline error

# Test Plan

Upload the attached protocol and see that the there is a timeline error. If you create a heater-shaker command to stop shaking and drag it before the timeline error step, it should go away.

[1.4.0 alpha1 test.json](https://github.com/Opentrons/opentrons/files/15264651/1.4.0.alpha1.test.json)

# Changelog

- create util for getting if flex slots are adjacent to each other and plugging the logic into the `pipetteIntoHeaterShakerWhileShaking`
- extend prop to include robot type and add it to all atomic commands that have `pipetteIntoHeaterShakerWhileShaking`
- fix tests
- add test for the new util `getAreFlexSlotsAdjacent`

# Review requests

see test plan

# Risk assessment

low


koji (approved):
the changes look good to me.

PR #15145, opened 2024-05-09
fix(app): fix password button rendering issue and update input field component

koji:
# Overview
the issue is that when tapping the eye icon button, the entire screen is re-rendered and the caret is move to the left from the right. 
use React.memo for this.

Introducing forwardRef to control `focus()` instead of using React event.
I think using ref would bring more flexibility to us.

remove secondary caption  since it seems that secondary caption is just taken over from the original component which is `library/inputfield` and Mel asked `why do we need that`?

InputField component is existing in `app` right now but this should be moved into `components`.
Mel is requesting us to update Storybook. I think I will move input field into components after 7.3.0 and update stories and test cases.

close RQA-2697

# Test Plan

# Changelog

# Review requests

# Risk assessment


mjhuff (approved):
Nice fix! Lgtm with passing CI.

PR #15155, opened 2024-05-10
fix(protocol-designer, step-generation): labware only removed if prev…

jerader:
…iously been discarded

closes RESC-258, RQA-2703

# Overview

This is a weird one, basically a customer got confused that their labware did not appear from the dropdown list to move the labware and that is because they discarded it later in the protocol. So this PR changes the logic to only remove it from the dropdown list if it is past the step where it was discarded. Along with that, i created a new error creator to let users know if a labware has been previously discarded.

# Test Plan

Create a flex protocol and add the waste chute and gripper.
On the deck setup page, add another labware.
First, create a move labware step to move the tiprack to the waste chute with the gripper
Second, create another move labware step to move the other labware to the waste chute with the gripper
Third, go back to step 1 and change the move labware to move the other labware to the waste chute. 
You should then receive a timeline error saying that the labware has been discarded 

# Changelog

- tweak logic in the labware location selector
- create new error creator for discarded labware and add a test case in the move labware atomic command
- fix logic for when waste chute is selectable because you must use the gripper to move it

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/StepEditForm/fields/LabwareLocationField/index.tsx
```diff
@@ -34,6 +42,12 @@ export function LabwareLocationField(
     )
   }
 
+  if (!useGripper && getHasWasteChute(additionalEquipmentEntities)) {
+    unoccupiedLabwareLocationsOptions = unoccupiedLabwareLocationsOptions.filter(
+      option => option.value !== WASTE_CHUTE_CUTOUT
+    )
+  }
```
jerader, line 49: i previously got rid of this filter a month or so ago per Product and QA's request. But i realized that we have no addressable area name for manually moving something to the waste chute. So I am pretty sure that behavior is not allowed and you MUST move a labware to the waste chute WITH the gripper.

koji (approved):
lgtm

PR #15157, opened 2024-05-10
fix(protocol-designer): properly disable up arrow for MoaM

jerader:
closes RQA-2704

# Overview

As described in the bug ticket, even when the arrow was disabled, the CTA was not disabled. This PR fixes it

# Test Plan

create a flex protocol and add 2 staging area slots. Then add as many temperature moduels as you can. It should be disabled after 5 temperature modules.

# Changelog

- fix logic for disabling the CTA
- add test coverage

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/EquipmentOption.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/EquipmentOption.tsx
@@ -163,7 +163,7 @@ export function EquipmentOption(props: EquipmentOptionProps): JSX.Element {
           {...tempTargetProps}
           data-testid="EquipmentOption_upArrow"
           onClick={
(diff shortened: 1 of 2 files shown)


koji:
https://github.com/Opentrons/opentrons/assets/474225/d82641b3-7a75-45c9-83eb-eec824060d6c


extra `,`
Screenshot 2024-05-10 at 10 47 20 AM

jerader:
> Screen.Recording.2024-05-10.at.10.45.42.AM.mov 


@koji oof thank you, good catch!! i should have checked! I might fix in a followup PR to keep things organized.

koji (approved):
left a comment but I think it won't block anything since the error Rob reported has been solved correctly.

PR #15158, opened 2024-05-10
fix(hardware): fix sensor moves that aren't using the buffer

ryanthecoder:
# Overview
When we were working on the sensor buffer based moves we accidently broke other sensor based moves by changing the move stop condition. 
In the buffer moves we trigger the sync differently because we're doing it from the motor-interrupt and then clearing it from the python side
but the other that are normally expecting to check against the move stop condition of the sync line were never finishing because the sync line wasn't getting sent.

this PR fixes the issue by only changing the move stop condition when we're using the buffer based moves

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (changes requested):
Good but please add tests

sfoster1 (approved):
cool cool

PR #15159, opened 2024-05-10
refactor(protocol-designer): logic for when slots are occupied for staging areas

jerader:
closes RQA-2705

# Overview

This addresses the bug that Koji found. I forgot to update the staging area slot unoccupied logic to account for the modules not being assigned a slot initially.

# Test Plan

Create a flex protocol and add a bunch of modules then go back to the staging area slots tile. See that it only allows you to add a staging area if there are enough slots.

NOTE: there is nothing notifying you why there are no available slots. I guess i could add something there, like a banner or something but I think i will leave as is unless QA or design think otherwise. This will be redesigned in PD 8.2 anyway
# Changelog

- refactor logic for getting unoccupied staging area slots
- fix test

# Review requests

see test plan

# Risk assessment

low


koji (approved):
Nice!

PR #15162, opened 2024-05-10
refactor(api): Allow commands to return errors instead of raising them

SyntaxColoring:
# Overview

This reworks the `opentrons.protocol_engine` command boilerplate to prepare for EXEC-427.

# Test plan

CI is sufficient.

# Changelog

Formerly, command implementations had to either return a `BaseModel`, which would get put in the command's `result` field, or raise an exception, which would get converted into a `BaseModel` and then put in the command's `error` field.

This PR adds a third option: to return an error `BaseModel` directly. We want returned error `BaseModel`s to represent [*defined* errors](https://opentrons.atlassian.net/wiki/spaces/PER/pages/4126703622/Proposal+Structured+representation+of+recoverable+errors), and raised exceptions to represent *undefined* errors. This will let us enforce some parity between a command's publicly declared interface and what it will actually return.

Implementations can associate private data with an error—they return a `DefinedErrorData(public=SomeErrorBaseModel(...), private=...)` dataclass. For symmetry with this, implementations now also represent success by returning a `SuccessData(public=FooResult(...), private=...)` dataclass, instead of a `(public, private)` tuple or just returning `public` directly. Having separate dataclasses makes it cleaner for the `CommandExecutor` to distinguish between success and failure.

No commands take advantage of this yet.

# Review requests

This PR has a lot of noise because it changes the command boilerplate. The most worthwhile things to look at are:

* `command.py` and `command_executor.py` have the main changes.
* `configure_for_volume.py`/`configure_nozzle_layout.py`/`load_pipette.py` are the three commands that return a private result. It's worth looking at one of them to see how that translates into the new system.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/aspirate.py
+++ b/api/src/opentrons/protocol_engine/commands/aspirate.py
@@ -1,7 +1,7 @@
 """Aspirate command request, result, and implementation models."""
 from __future__ import annotations
 from typing import TYPE_CHECKING, Optional, Type
-from typing_extensions import Literal
+from typing_extensions import Literal, Never
 
 from .pipetting_common import (
     PipetteIdMixin,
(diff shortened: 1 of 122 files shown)


sfoster1 (approved):
This looks good. I wonder if having all these tests instead be
```
assert result.public = SomeResult(...)
assert result.private = SomePrivateResult(...)
```
might have saved some time but now you've already done it so it's immaterial.

PR #15167, opened 2024-05-10
feat(shared-data, api): add absorbance plate reader definition and module control

ahiuchingau:
# Overview
This PR adds support for the Absorbance Plate Reader in the hardware controller API. We can now see the plate reader as an attached module on the backend and use the AbsorbanceReaderDriver to communicate with the plate reader like we would with the other modules.

# Caveat
The driver is not fully finished - there are some functions that still need to be implemented and I expect the format of the response will change depending on how we implement the engine commands. 

We still need to add some changes on oe-core side to make sure this runs on all robot - however i'm happy to show you the plate reader in action at this time.


api/src/opentrons/drivers/absorbance_reader/hid_protocol.py
```diff
@@ -0,0 +1,89 @@
...
+        self, device_handle: int
+    ) -> Tuple[ErrorCode, SlotState]:
+        ...
+
+    def byonoy_get_device_parts_aligned(
+        self, device_handle: int
+    ) -> Tuple[ErrorCode, bool]:
+        ...
+
+    def byonoy_abs96_get_available_wavelengths(
```
caila-marashaj, line 99: are these all commands you access through the hid interface?

caila-marashaj (approved):
Nice! Congrats on getting this up and running

api/src/opentrons/drivers/absorbance_reader/hid_protocol.py
ahiuchingau: Yes they are! So we'd have to update this file and the AsyncByonoy whenever the byonoy library gets updated.

sfoster1 (approved):
This looks pretty good, this seems like a very annoying interface to deal with.

Let's get those error strings defined as a `Literal` enum or something

api/Pipfile.lock
```diff
@@ -280,6 +280,15 @@
             ],
             "version": "==4.2.2"
         },
+        "pyusb": {
```
vegano1, line 283: Do we want to include all these changes to the lock file or just whats required to include pyusb?

vegano1 (approved):
This looks good, 
Just have a comment on whether we want to update all the Pipfile.lock files, which will update the version of a few other modules or just add whats required for `pyusb `.

PR #15168, opened 2024-05-10
fix(app, components, protocol-designer, shared-data): import type lint rule to error, fix occurrences

b-cooper:
# Overview

Removes ~1000 lint warnings and standardizes the separation of type imports from module imports

# Risk assessment
medium, no code should be effected, but touches many files


brenthagen (approved):
LGTM !

PR #15169, opened 2024-05-11
OpenAI API available in the opentrons-ai-server

y3rsh:
# Overview

This PR introduces a structured pattern for handling requests and responses within the lambda. It demonstrates actual communication with OpenAI, starting from the lambda and extending back to the client. Additionally, the endpoint is secured, requiring a JWT bearer token for access.
___

**Details:**
- **Client for Testing:** Established a dedicated live client to facilitate testing.
- **Testing:** Developed tests that exercise the client and all endpoints.

## Status

- [x] Deployed to dev and live tested `make build deploy test-live ENV=dev AWS_PROFILE=the-profile`
- [x] Deployed to sandbox  and live tested `make build deploy test-live ENV=sandbox AWS_PROFILE=the-profile`
- [x] `make live-client` with fake=false works on sandbox
- [x] `make live client` with fake=false works on dev

> [!NOTE]
> to exercise the client reach out to @y3rsh for the `test.env` you need


opentrons-ai-server/Makefile
```diff
@@ -71,13 +71,13 @@ clean-package:
...
 	@echo "Generating requirements.txt from Pipfile.lock..."
-	python -m pipenv requirements --hash > requirements.txt
+	python -m pipenv requirements > requirements.txt
 
 .PHONY: install-deps
 install-deps:
 	@echo "Installing dependencies to package/ directory..."
 	mkdir -p package
-	python -m pipenv run pip install -r requirements.txt --target ./package --upgrade
+	docker run --rm -v "$$PWD":/var/task "public.ecr.aws/sam/build-python3.12" /bin/sh -c "pip install -r requirements.txt -t ./package"
```
sfoster1, line 80: why the double expansion?

opentrons-ai-server/api/domain/openai_predict.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+
+        assistant_message: ChatCompletionMessage = response.choices[0].message
+        return assistant_message.content
+
+
+def main() -> None:
+    settings = Settings()
+    openai = OpenAIPredict(settings)
+    prompt = Prompt.ask("Type a prompt to send to the OpenAI API:")
```
sfoster1, line 49: nit: there's nothing mechanically making sure that `Prompt` was imported, though it's implicitly always going to be present assuming the lambda doesn't run this script as main. still, might be better to dynamically import it here or something

opentrons-ai-server/api/settings.py
```diff
@@ -1,54 +1,77 @@
...
-            self.huggingface_api_key: SecretStr = self.fetch_secret(f"{self.environment}-huggingface-api-key")
+            self.openai_api_key: SecretStr = fetch_secret(f"{self.ENVIRONMENT}-openai-api-key")
+            self.huggingface_api_key: SecretStr = fetch_secret(f"{self.ENVIRONMENT}-huggingface-api-key")
         else:
             # Use values from .env or defaults if not set
             self.openai_api_key = SecretStr(os.getenv("OPENAI_API_KEY", "default-openai-secret"))  # can change to throw
             self.huggingface_api_key = SecretStr(os.getenv("HUGGINGFACE_API_KEY", "default-huggingface-secret"))  # can change to throw
+        self._is_initializing = False  # End initialization phase
+
+    def __setattr__(self, name: Any, value: Any) -> None:
```
sfoster1: Instead of doing this, let's 
1) make `Settings` a [frozen dataclass](https://docs.python.org/3/library/dataclasses.html#frozen-instances)
2) make a `build()` classmethod that returns an instance and does this work by running the code in the body and then returning a constructed dataclass

All that's doing is a library-code version of what's written here, but it lets you define the values contained in the class and thus its shape separately from the logic fo building it.

sfoster1 (approved):
Looks good to me aside from the dataclass thing - let's change that IMO

opentrons-ai-server/README.md
```diff
@@ -46,10 +46,13 @@ The Opentrons AI application's server.
...
+1. Make your changes
+1. Fix what can be automatically then lent and unit test like CI will `make pre-commit`
+1. `make pre-commit` passes
+1. deploy to sandbox `make build deploy test-live ENV=sandbox AWS_PROFILE=the-profile`
 
-1. build the package `make build`
-1. deploy the package `make deploy ENV=sandbox AWS_PROFILE=robotics_ai_sandbox`
+## TODO
+
+- llama-index is gigantic. Have to figure out how to get it in the lambda
```
Elyorcv, line 58: you mean storage? or library itself?

opentrons-ai-server/Pipfile
koji, line 1: installing two instead of the entire llama_index is still too big?
```
llama-index-core
llama-index-llms-openai
```

opentrons-ai-server/api/settings.py
```diff
@@ -1,54 +1,77 @@
...
 
+from api.integration.aws_secrets_manager import fetch_secret
+
 ENV_PATH: Path = Path(Path(__file__).parent.parent, ".env")
 
+# TODO:y3rsh:2024-05-10: use https://github.com/pydantic/pydantic-settings
+# once it does not version conflict with AWS Lambda Powertools
+
 
 class Settings:
```
Elyorcv, line 20: nit: Could we give different name, we noticed that llama-index lib uses this `Settings` names as well?

opentrons-ai-server/README.md
Elyorcv: - Could we add clarification in terms of all commands for running locally? 
- Now we have two folders: this is for BE and another for FE. What I would like to understand is that how do we know when we run the front end it is connected to BE?

Elyorcv (commented):
LGTM. Thanks!

opentrons-ai-server/api/settings.py
y3rsh: on it

opentrons-ai-server/Makefile
y3rsh: You use double dollar signs ($$) to escape the first dollar sign from being interpreted by the make process, ensuring that it is passed as a literal dollar sign to the shell invoked by Docker.

Single doesn't work.

opentrons-ai-server/api/domain/openai_predict.py
y3rsh: Oh, ya for sure, meant to make it dynamic.

opentrons-ai-server/api/settings.py
y3rsh: yes, I think when we import llama-index Settings rename that one with `as`.
our package get to win the name war :D

y3rsh: yes, I think when we import llama-index Settings rename that one with as.
our package get to win the name war :D

opentrons-ai-server/Pipfile
y3rsh: Got closer trying this.  Will resolve in another PR.

opentrons-ai-server/README.md
y3rsh: All of the python dependencies of llama-index make our package to run on the lambda >200MB
We will address in another PR

- All of the commands locally means that every command wired up in the `Makefile` works.
- Once we get the frontend wired up we will make it very clear which is all wired up.  We will likely add version and service name to the API responses.

y3rsh:
Latest commit is deployed on sandbox and dev.  All live tests pass including the new ones on both environments.
Addressed all feedback and made the frozen data class conversion.
Merging once CI finishes.

PR #15139, opened 2024-05-09
refactor(app): migrate /runs/:runId/commands to notifications

mjhuff:
Closes [EXEC-442](https://opentrons.atlassian.net/browse/EXEC-442)

# Overview

Migrate `useAllCommandsQuery` to notifications. This was missed from the initial MQTT migration audit. Fortunately, we have all the backend publishing infrastructure in place for this already, so it's just an app only refactor.

We can also refactor `useNotifyLastRunCommand` -> `useLastRunCommand`, since this is a compound hook that doesn't live in api-client, and all the HTTP hooks it utilizes are now on notifications.  

# Test Plan
- Verified that `useAllCommandsQuery` doesn't poll. An easy way to test this without extensive logging is to watch run command text update immediately.

# Risk assessment
low

[EXEC-442]: https://opentrons.atlassian.net/browse/EXEC-442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #15170, opened 2024-05-13
feat(app, react-api-client): add "retry failed command" during Error Recovery

mjhuff:
Closes [EXEC-432](https://opentrons.atlassian.net/browse/EXEC-432)

# Overview

Adds the ability to retry the command that caused the run to enter Error Recovery, resuming the run after the issued command completes. Implementing the retry logic requires laying out (hopefully) most of the remaining app-side foundational work, therefore that is included here as well.

There is a good bit of noise in this PR. A few specifics are mentioned via comments. Below are the major highlights and discussion points:

### Chaining Fixit Commands

In order to chain (or even create) commands with a `fixit` intent, we utilize `useRecoveryCommands`. I currently make the assumption that we do not want to handle fixit commands that fail during error recovery, so we always `continuePastCommandFailure`. Whether we actually want this behavior or not (or make this dependent on the exact fixit command) is an involved conversation and must also ensure error recovery can't possibly cause an infinite loop (such as a faulty sensor always detecting no tip attached after the user dispatches additional fixit pick up tip commands). 

### Identifying the Failed Run Command

Designs utilize splash pages that require knowledge of the failed command ([example](https://www.figma.com/file/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?type=design&node-id=324-157288&mode=design&t=rOQCfaqiD13cbARh-4)). While identifying and retaining the failed command is simple on the ODD, the complexity increases when we add desktop apps into the mix. A desktop user may be anywhere in the app and then may navigate to Error Recovery while a user engages (or has engaged) with Error Recovery on the ODD. In order to determine which command is the failed command, we must navigate backwards through the active run's list of commands, finding the most recent failed command with a `protocol` intent. And that's the issue; because `fixit` commands are added to the run command list, the following scenario may occur:

- User A on the ODD creates a significantly large number of `fixit` commands. These commands are appended to the protocol run commands list.
- User B on the desktop app navigates to the run, and the app attempts to find the last failed `protocol` command. No failed command is found when querying the server, which occurs due to the failed command being farther back in history than the `pageLength` sent to GET /runs/:runId/commands returned. [See this discussion](https://opentrons.slack.com/archives/C06M9F26YGH/p1715349271043819) for some background as to why this is the case.

@sanni-t's recent PR https://github.com/Opentrons/opentrons/pull/15031 solves the serialization issue for _completed_ runs. We could consider extending this functionality for commands during an on-going run (and maybe add a rate limiter?!?). @sfoster1 's thoughts are highlighted in the slack convo linked above after discussion in-person.

### Robot Motion State

If you look at [designs](https://www.figma.com/file/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?type=design&node-id=458-38190&mode=design&t=rOQCfaqiD13cbARh-4), there are a lot of different "robot in motion" screens. Sometimes we only want the motion screen when the robot does a command, and sometimes we want to persist it longer or indefinitely. I think the best way to handle motion is to explicitly control it by setting the motion screen before we do a command and then control it afterwards. This is a departure from the implicit `RobotInMotion` component takeover that all the other wizard flows use, but I don't think that's going to work for us, and even if it is possible, the control flow would be very hard to follow.

# Test Plan
- Turn on the BE/FE FFs for ER.
- Run any protocol with a Pick up tip command and don't pick up the tip. Proceed through the ER flows. Verify the initial Z-home occurs when launching ER, and clicking "resume" (see Review requests - designs are quite different than current flows and that's ok) causes a pick up tip command.
- NOTE: The retry failed command pick up tip will actually fail until https://github.com/Opentrons/opentrons/pull/15143 merges.
- Verify the run resumes. There may be some oddness with the command text on the desktop app/odd, but that's ok for now.

# Changelog
- Added "retry failed command" option in Error Recovery.

# Review requests
Designs have changed quite a bit in the past couple of weeks. This PR is just focused on functionality and structure!

# Risk assessment
low -- behind a ff

[EXEC-432]: https://opentrons.atlassian.net/browse/EXEC-432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
mjhuff, line 1: Diff this against the old `index.tsx` to see changes.

The differences are the new route for the "retrying command" view, the `useInitialPipetteHome`, and the new name, `ErrorRecoveryWizard`.

app/src/organisms/ErrorRecoveryFlows/utils.ts
```diff
@@ -1,18 +1,66 @@
...
+): FailedCommand | null {
+  const [
+    recentFailedCommand,
+    setRecentFailedCommand,
+  ] = React.useState<FailedCommand | null>(null)
+  // The most recently failed protocol command causes the run to enter "awaiting-recovery", therefore only check
+  // for a newly failed command when the run first enters "awaiting-recovery."
+  const isRunStatusAwaitingRecovery = runStatus === RUN_STATUS_AWAITING_RECOVERY
+
+  const { data: allCommandsQueryData } = useAllCommandsQuery(runId, null, {
```
mjhuff: https://github.com/Opentrons/opentrons/pull/15139 migrates this to notifications.

app/src/organisms/ErrorRecoveryFlows/index.tsx
```diff
@@ -1,109 +1,51 @@
...
-
-  const buildSelectRecoveryOption = (): JSX.Element => {
-    return <SelectRecoveryOption {...props} />
-  }
+export function useErrorRecoveryFlows(
+  runId: string,
+  runStatus: RunStatus | null
+): UseErrorRecoveryResult {
+  const [isEREnabled, setIsEREnabled] = React.useState(false)
+  const failedCommand = useCurrentlyFailedRunCommand(runId, runStatus)
```
mjhuff, line 22: Not directly exporting `useCurrentlyFailedRunCommand` may be something we have to refactor out later, as this implies that we only would ever need a `failedCommand` for reasons related to Error Recovery.

mjhuff (commented):
Some explanatory comments below.

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
```diff
@@ -0,0 +1,133 @@
...
+export function ErrorRecoveryWizard({
+  runId,
+  failedCommand,
+}: ErrorRecoveryFlowsProps): JSX.Element {
+  /**
+   * Recovery Route: A logically-related collection of recovery steps or a single step if unrelated to any existing recovery route.
+   * Recovery Step: Analogous to a "step" in other wizard flows.
+   */
+  const [recoveryMap, setRecoveryMap] = React.useState<IRecoveryMap>({
+    route: RECOVERY_MAP.ROBOT_IN_MOTION.ROUTE, // Initialize the route to "in motion", since wizard always start with a home command.
```
DerekMaggio: Do we expect the robot to already be in motion when we start this wizard? 
Or does the robot not start homing until the useInitialPipetteHome call like 15 lines down? 

If it is already in motion, then this makes sense, if it is not, should `useInitialPipetteHome` be the thing that adds the ROBOT_IN_MOTION step to the route?

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
```diff
@@ -0,0 +1,133 @@
...
+  recoveryCommands: ReturnType<typeof useRecoveryCommands>,
+  routeUpdateActions: ReturnType<typeof useRouteUpdateActions>
+): void {
+  const { homePipetteZAxes } = recoveryCommands
+  const { setRobotInMotion } = routeUpdateActions
+
+  React.useEffect(() => {
+    void setRobotInMotion(true)
+      .then(() => homePipetteZAxes())
+      .then(() => setRobotInMotion(false))
```
sfoster1: probably wants to be `.finally()`

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
```diff
@@ -0,0 +1,133 @@
...
+      left="14.5px"
+      top="16px"
+      borderRadius={BORDERS.borderRadius12}
+      position={POSITION_ABSOLUTE}
+      backgroundColor={COLORS.white}
+    >
+      <ErrorRecoveryHeader errorKind={props.errorKind} />
+      <ErrorRecoveryContent {...props} />
+    </Flex>,
+    getTopPortalEl()
```
sfoster1: probably wants to be `getModalPortalEl` if this is on the desktop

app/src/organisms/ErrorRecoveryFlows/index.tsx
```diff
@@ -1,109 +1,51 @@
...
-      position={POSITION_ABSOLUTE}
-      backgroundColor={COLORS.white}
-    >
-      <ErrorRecoveryHeader errorKind={props.errorKind} />
-      <ErrorRecoveryContent {...props} />
-    </Flex>,
-    getTopPortalEl()
-  )
+interface UseErrorRecoveryResult {
+  isEREnabled: boolean
```
sfoster1: kind of a nit, but the use of "enabled" here makes you think this is a wrapper for the feature flag at first glance. can we call it like `isERActive` and so on?

app/src/organisms/ErrorRecoveryFlows/utils.ts
```diff
@@ -1,18 +1,70 @@
...
+// TODO(jh, 05-09-24): Migrate utils, useRecoveryCommands.ts, and respective tests to a utils dir, and make each util a separate file.
+
+// TODO(jh, 05-09-24): It is possible for a desktop app not to detect the most recently failed run command if there are
+// too many FIXIT commands. See PR associated with TODO for explanation.
+
+// While the run is "awaiting-recovery", return the most recently failed run command with a protocol intent.
+// Otherwise, returns null.
+const ALL_COMMANDS_POLL_MS = 5000
+
+export function useCurrentlyFailedRunCommand(
```
sfoster1, line 29: credit to @SyntaxColoring for this one - what if we added a cursor to `GET /runs/:runid/commands` that was "last failed command" so we didn't have to parse the command list at all?

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
mjhuff: Good Q - doing it this way prevents us initially rendering some other screen for one render cycle before the in-motion screen. This problem is really only noticeable if you test it on the physical ODD, and it looks bad enough IMO that I'd rather eagerly set the robot to 'in motion' immediately and not have the screen immediately changing as soon as we launch ER.

app/src/organisms/ErrorRecoveryFlows/utils.ts
mjhuff: Are you talking about a server side change? If app side, how would this work with a desktop app that navigates to the run after 100+ fixit commands? Wouldn't it need knowledge of which command failed to know which cursor to use? CC @SyntaxColoring

sfoster1: Yeah, a server side change.

SyntaxColoring: EXEC-458

I'm working on the server part of this now.

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
mjhuff: @DerekMaggio After further investigation, I found a better React way of doing this while initializing the route to something more sensible, so thank you for calling this out!

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
mjhuff: Ah yep, great point. Given hi-fi designs, it looks like we may have some modal-within-modals as well, so we'd want the lower `zIndex` for ODD as well.

sfoster1 (approved):
Looks good to me!

PR #15171, opened 2024-05-13
style(api): Automatically sort imports

SyntaxColoring:
# Overview

The ordering of `import` statements is a boring style concern, so let's automate it like we automate formatting with Black.

Among other things, this is helpful for refactors like #15162, where we want to add a new import to ~60 files.

This goes towards EXEC-456.

**To avoid horrific merge conflicts, we should wait to merge this until we merge v7.3.0 back into `edge`.**

# Test plan

None needed.

# Changelog

Via [isort](https://pycqa.github.io/isort/index.html):

* During `make -C api lint`, check that imports are sorted.
* During `make -C api format`, automatically sort imports.

# Review requests

Review commit-by-commit to separate my manual changes from the automated changes from the initial run of `isort`.

* Any reason we shouldn't do this?
* Any reason to adjust isort's configuration from the default?
* Any files that need special manual attention?

If we like this, we can do the same thing in our other Python projects.

# Risk assessment

Low.


sfoster1 (approved):
I think this is a good idea and I think that actually we don't really need to wait to merge it; 7.3.0 has gotten some regular mergebacks (and could use another) and only changes to imports should conflict - there's not that many of them.

sfoster1:
That said, looking at the failed tests we're gonna have to do some followups lol

SyntaxColoring:
> I think this is a good idea and I think that actually we don't really need to wait to merge it; 7.3.0 has gotten some regular mergebacks (and could use another) and only changes to imports should conflict - there's not that many of them.

It looks like import reordering can bring up latent circular dependency problems. I'll hold off on this until after the final mergeback, on the off chance that it happens there. We don't want whoever does the mergeback to have to deal with that kind of thing.

SyntaxColoring:
Closing this for now. It'll probably be a little bit before I get to it.

At the time of writing, we "just" need to resolve the circular dependency problems described in #15234 to unblock this.

I also suspect `robot-server` and `shared-data` will go way smoother, so maybe we'll do those first.

PR #15174, opened 2024-05-13
refactor(app): split InterventionModal to molecule

sfoster1:
The implementation of the intervention modal (i.e., manual move labware) in the desktop app was sort of manually open-coded in the intervention modal organism. We're going to need modals styled in this way elsewhere, so split it out into an app molecule with some overridable styles.

Also, make its background wrapper position: absolute instead of position: fixed, for the same reasons as #15166 - when this modal is hung off of the modal portal rather than the top portal, it should allow interaction with the navbar and the breadcrumbs. This should not affect the modal's use in the actual InterventionRequired organism, since in the organism the modal is hung off of the top portal.

Closes RSQ-6

## Testing
- [x] The new storybook component should render
- [x] Move labware and pause modals on desktop should still render as in ~figma~ the shipping app
- [x] Move labware and pause modals on desktop should continue to ~(a) visually overlap and (b)~ inhibit interaction with the navbar and breadcrumbs 

the shipping app has these modals confined to the route component area, while figma has them overlapping everything with a small padding to the viewport boundary. this pr doesn't change that. the navbar and route breadcrumbs are still non interactive.

### DIFF
--- a/app/src/molecules/InterventionModal/InterventionModal.stories.tsx
+++ b/app/src/molecules/InterventionModal/InterventionModal.stories.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react'
+
+import { StyledText } from '@opentrons/components'
+import { InterventionModal as InterventionModalComponent } from './'
+import type { Story, Meta } from '@storybook/react'
+
+export default {
+  title: 'App/Molecules/InterventionModal',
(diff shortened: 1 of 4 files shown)


mjhuff (approved):
Nice, thanks for the refactor. We'll definitely put this to good use soon!

PR #15176, opened 2024-05-13
feat(api): Return a defined `tipPhysicallyMissing` error from `pickUpTip` commands

SyntaxColoring:
# Overview

This implements our first real-world instance of a Protocol Engine command returning a defined error, instead of raising an undefined error. Closes EXEC-427.

# Changelog

* Define a `tipPhysicallyMissing` error and make the `pickUpTip` command return it, when appropriate.
* Some small remaining Protocol Engine updates to handle errors that are returned in addition to errors that are raised.

# Test plan

* [x] Run a protocol with some missing tips. When it fails, the command should fail with the new `tipPhysicallyMissing` error.
    * Confirmed by looking at the raw HTTP responses.
    * The Flex ODD, however, is still showing the underlying (wrapped) error. I suspect it's considering it higher-priority because its code is numerically higher. I'm going to look at this in a separate PR. EXEC-454
* [x] Make sure the new `tipPhysicallyMissing` error shows up in the `pickUpTip` command's OpenAPI spec.
    * This *sort of* works. It does show up as a new kind of `error` value, separate from the default `ErrorOccurrence`. But it's not showing the class docstring, which is where I've tried to document the error's semantics. So we'll need to figure that out. I'd like to upgrade to Pydantic v2 before worrying about this any further.

# Review requests

See my inline GitHub comments.

# Risk assessment

Low.


api/src/opentrons/protocol_engine/commands/pick_up_tip.py
```diff
@@ -50,25 +61,61 @@ class PickUpTipResult(DestinationPositionResult):
...
+
+@dataclass
+class TipPhysicallyMissingErrorInternalData:
+    """Internal-to-ProtocolEngine data about a TipPhysicallyMissingError."""
+
+    pipette_id: str
+    labware_id: str
+    well_name: str
+
+
```
sfoster1, line 88: this is kind of a nit, but what we do here is going to be what devs pattern match when adding things to subsequent commands. Can we establish the pattern of
- create an alias for the return type: `PickUpTipReturn = Union[SuccessData[...], DefinedErrorData[...], ...]`
- use that alias in the implementation generic type argument
- use that alias as the return annotation

api/src/opentrons/protocol_engine/commands/pick_up_tip.py
```diff
@@ -50,25 +61,61 @@ class PickUpTipResult(DestinationPositionResult):
...
+    detail: str = "No tip detected."
+
+
+@dataclass
+class TipPhysicallyMissingErrorInternalData:
+    """Internal-to-ProtocolEngine data about a TipPhysicallyMissingError."""
+
+    pipette_id: str
+    labware_id: str
+    well_name: str
```
SyntaxColoring, line 86: `tips.py` will need this info to logically remove the tip from the tip rack. It's currently getting this info from the command's `params`, which was [quick hack](https://github.com/Opentrons/opentrons/blob/76360e51dd263e1a11e2f5f7168e549bc8c132ac/api/src/opentrons/protocol_engine/actions/actions.py#L175-L180) that probably won't scale to other error types.

For the sake of being incremental, I'll undo that hack in a separate PR, not this one. So, as of this PR, this data is just in anticipation of that, and is not currently consumed by anything.

api/src/opentrons/protocol_engine/commands/pick_up_tip.py
```diff
@@ -50,25 +61,65 @@ class PickUpTipResult(DestinationPositionResult):
...
+class TipPhysicallyMissingError(ErrorOccurrence):
+    """Returned when sensors determine that no tip was physically picked up.
+
+    That space in the tip rack is marked internally as not having any tip,
+    as if the tip were consumed by a pickup.
+
+    The pipette will act as if no tip was picked up. So, you won't be able to aspirate
+    anything, and movement commands will assume there is no tip hanging off the bottom
+    of the pipette.
+    """
```
SyntaxColoring, line 73: > movement commands will assume there is no tip hanging off the bottom of the pipette.

Is this a good idea? Theoretically, we could assume that there *is* a tip hanging off the bottom of the pipette.

That might be a bit harder to implement and maintain, since there "is a tip" for movement purposes but "there is not a tip" for liquid handling purposes. But it would be safer for the client. Imagine this error triggers falsely because of a flaky sensor, so physically, there actually is a tip attached. Then, if a client commands a `moveToWell` as part of its recovery flow, we probably want the robot to go high enough to clear all the labware on the deck, right?

SyntaxColoring: On the other hand, if a client commands a `moveToWell` to be followed by a `pickUpTipInPlace`, we want the robot to put the pipette nozzle in the proper tip pickup position, which means acting as if there is no tip attached...

api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -158,13 +163,25 @@ class FailCommandAction:
     """An ID to assign to the command's error.
 
     Must be unique to this occurrence of the error.
+
+    todo(mm, 2024-05-13): This is redundant with `error` when it's a defined error.
```
TamarZanzouri, line 167: sorry for the question but is this part of the docstring?

api/src/opentrons/protocol_engine/actions/actions.py
```diff
@@ -158,13 +163,25 @@ class FailCommandAction:
...
     """
 
     failed_at: datetime
-    """When the command failed."""
+    """When the command failed.
+
+    todo(mm, 2024-05-13): This is redundant with `error` when it's a defined error.
```
TamarZanzouri, line 173: same

api/src/opentrons/protocol_engine/actions/actions.py
SyntaxColoring: I figured, if you're in a different file and hovering over this field in your IDE trying to figure out what the difference is between it and `error.public.id`, this would be helpful to see.

SyntaxColoring:
Robot server test failures are unrelated, caused by AUTH-401.

api/src/opentrons/protocol_engine/commands/pick_up_tip.py
```diff
@@ -82,24 +124,46 @@ async def execute(
...
+        try:
+            tip_geometry = await self._tip_handler.pick_up_tip(
+                pipette_id=pipette_id,
+                labware_id=labware_id,
+                well_name=well_name,
+            )
+        except TipNotAttachedError as e:
+            return DefinedErrorData(
+                public=TipPhysicallyMissingError(
+                    id=self._model_utils.generate_id(),
```
TamarZanzouri, line 136: maybe a stupid question but why do we need 2 different id's?

api/src/opentrons/protocol_engine/commands/save_position.py
```diff
@@ -53,11 +53,11 @@ class SavePositionImplementation(
     def __init__(
         self,
         gantry_mover: GantryMover,
-        model_utils: Optional[ModelUtils] = None,
+        model_utils: ModelUtils,
```
TamarZanzouri, line 56: why do we need this change?

TamarZanzouri (approved):
Looks great!!! just a couple of questions but overall I like this change

api/src/opentrons/protocol_engine/commands/save_position.py
SyntaxColoring: It’s not necessary. But this is how dependency injection works in general for command implementations, and I didn’t see a reason for `ModelUtils` to be different.

api/src/opentrons/protocol_engine/commands/pick_up_tip.py
SyntaxColoring: Not a stupid question. One error wraps another error, and each error “has a unique ID.” I interpret that as being unique to the individual error object, not unique to the overall error event. I’m not 100% sure this is correct.

Separately: the error `id` field is useless these days, as far as I can tell, and I would like to delete it.

TamarZanzouri: Thank you! yeah that was another question I had. I didn't see we use the `id` anywhere so I wondered why we have it. I guess I just figured that the wrapped error already had an id.

PR #15179, opened 2024-05-14
fix(app): show InfoScreen under RunPreview for a run canceled before start

ncdiehl11:
Closes RQA-2717

# Overview

Show a 'Run was never started' InfoScreen under RunPreview for a terminal run that was never started (has 0 commands)

# Test Plan

- Begin protocol setup
- Cancel run before starting
- Verify that 'Run was never started' info screen shows
Screenshot 2024-05-14 at 11 52 24 AM

# Changelog

- show appropriate info screen if no commands exist on the run (consistent with parameters tab)
- fallback to analysis commands while commands from `runs/{runId}/commandsAsPreSerializedList` are loading so we don't get weird json as a step

# Review requests

auth js

# Risk assessment

low

### DIFF
--- a/app/src/assets/localization/en/run_details.json
+++ b/app/src/assets/localization/en/run_details.json
@@ -59,6 +59,7 @@
   "load_module_protocol_setup": "Load {{module}} in Slot {{slot_name}}",
   "load_pipette_protocol_setup": "Load {{pipette_name}} in {{mount_name}} Mount",
   "loading_protocol": "Loading Protocol",
+  "loading_data": "Loading data...",
   "location": "location",
   "module_controls": "Module Controls",
   "module_slot_number": "Slot {{slot_number}}",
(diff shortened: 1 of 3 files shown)


koji (approved):
The changes look good to me.
Thank you for fixing this issue!

shlokamin (approved):
code looks good to me!

ncdiehl11:
@shlokamin @koji I just pushed a commit to better address fallback logic. There was an issue that after a run is made terminal, we read from the _run_ commands rather than analysis commands, but for a moment, those run commands are `null`. Rather than fall back to analysis commands during the null reading, I think we should fallback to "Loading data..." as we do at protocol setup for consistency and reduced confusion.

https://github.com/Opentrons/opentrons/assets/47604184/727046ed-8c0f-4ef8-818e-e59c13be00fb

koji (approved):
lgtm!

sfoster1 (approved):
Approving the logic stuff - it's gross and we should look at it later, but good for now.

ncdiehl11:
updated behavior. TODO: fix the run canceled banner flickering. unrelated to this component.
https://github.com/Opentrons/opentrons/assets/47604184/6f6050da-1545-44b3-aff7-4860791bbce5

PR #15181, opened 2024-05-14
feat(app): Add incompatible module notification modals

sfoster1:
We meant to add this for the launch of the Flex, but it got cut for scope; in the time since, people connecting modules like first-generation temperature modules to Flexes that don't support them has been a pretty consistent thorn in the side of support. This PR implements the modals designed by design for telling the user when there's incompatible modules.

This PR is pretty big and might be best viewed by commit, since it has both the backend and the frontend, and both the ODD and the desktop - and a couple ancillary support PRs that I'll rebase out of here as they get merged. Specifically,

## Overview

90dc748fbb - Python, adds data to module configs to say what kinds of deck they're compatible with, and adds code to the api and robot server to expose a bool about whether the pipette's compatible or not in the HTTP API
788e022bac - JS, adds the above to the api clients
2e49f6160c - JS, adds a blocking modal to the ODD that pops up whenever modules are attached per [figma](https://www.figma.com/file/0kuPeCi1t2Auu2GPMnqRb2/Primary%3A-Flex?type=design&node-id=19961-155919&mode=design&t=pkbLPA9150bwQQNw-4)
c5021ca145 - JS, adds a modal that blocks interaction with a robot but _not_ the navbar or breadcrumbs whenever modules are attached per [figma](https://www.figma.com/design/Nx7ORMnfyJNP4FQYxN4e5x/Primary%3A-Desktop-App?node-id=2032-304206&t=iat6IH7YQD4pMijc-4)

## Notes and Review requests

90dc748fbb - I'm pretty sure these are the module compatibility rules, but we'll need to make sure. This will get exposed via HTTP as soon as the module is plugged in. Review requests:
- [ ] bool should be correct for plugged-in modules basically as soon as they appear in the app

The above, and 788e022bac - This is set up to support polling. Let me know if you think this should be notifications instead, will be another couple commits though. 

2e49f6160c - ODD modal. Is this according to design, is there something I'm missing? Is the testing approach OK? You can test by editing `robot-server/simulator/test-flex.env` to specify a `ThermocyclerModuleV1` instead of a `ThermocyclerModuleV2` and run with `make -C robot-server dev-flex`, as well as actually plugging stuff in to a real robot.

c5021ca145 - Desktop modal, basically same questions and same testing approach. Note that this is the first ish time we've added a modal that doesn't inhibit interaction with navbar or breadcrumbs.


robot-server/robot_server/modules/module_data_mapper.py
```diff
@@ -131,6 +140,9 @@ def map_data(
             firmwareVersion=module_identity.firmware_version,
             hardwareRevision=module_identity.hardware_revision,
             hasAvailableUpdate=has_available_update,
+            compatibleWithRobot=(
+                not (self.deck_type.value in module_definition["incompatibleWithDecks"])
```
TamarZanzouri, line 144: not a big deal but why not `compatibleWithDecks` instead of `incompatibleWithDecks`? and lose the not? this way it matches `compatibleWith`

TamarZanzouri: @sfoster1 answered privately, this way we dont need to update for every deck we add.

TamarZanzouri (approved):
Backend changes look good! glanced on fe changes but someone else can probably confirm better than me :-)

app/src/organisms/IncompatibleModule/IncompatibleModuleDesktopModalBody.tsx
```diff
@@ -0,0 +1,92 @@
...
+    >
+      <Flex flexDirection={DIRECTION_COLUMN} padding={SPACING.spacing32}>
+        <Flex
+          overflowY={OVERFLOW_SCROLL}
+          flexDirection={DIRECTION_COLUMN}
+          gridGap={SPACING.spacing12}
+          maxHeight="196px"
+          as="ul"
+        >
+          {...modules.map(module => (
```
mjhuff: I believe we don't need to spread here.

app/src/organisms/IncompatibleModule/IncompatibleModuleDesktopModalBody.tsx
```diff
@@ -0,0 +1,92 @@
...
+            <li key={module.id}>
```
mjhuff, line 54: Kind of surprised we don't have a component wrapper for `li`, but the pattern we use in these modal + list spots is to nix the `li` and use `Flex` as the top level component. Using `li` here (or a component that wraps it) is better practice, though, so I think we should keep it.

app/src/organisms/IncompatibleModule/IncompatibleModuleODDModalBody.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <StyledText as="p" marginBottom={SPACING.spacing32}>
+          <Trans t={t} i18nKey="remove_before_running_protocol" />
+        </StyledText>
+        <Flex
+          overflowY={OVERFLOW_SCROLL}
+          flexDirection={DIRECTION_COLUMN}
+          gridGap={SPACING.spacing8}
+          maxHeight="196px"
+        >
+          {...modules.map(module => (
```
mjhuff: Ditto

app/src/organisms/IncompatibleModule/IncompatibleModuleTakeover.tsx
```diff
@@ -0,0 +1,40 @@
...
+          : createPortal(
+              <IncompatibleModuleDesktopModalBody
+                modules={incompatibleModules}
+                robotName={robotName ?? ''}
+              />,
+              getModalPortalEl()
+            )
+        : null}
+    </>
+  )
```
mjhuff: Sweet, the logic here makes sense. Can we avoid the nested ternaries here? It's a bit hard to read.

mjhuff (approved):
This looks good! Just a few nits.

Both of the modals and their respective testing lgtm. 

It would be good to support module incompatibility notifications at some point, but I think that would be more involved than we probably think, and the poll interval is reasonable. 

Thank you for tackling all of this!

app/src/organisms/IncompatibleModule/IncompatibleModuleDesktopModalBody.tsx
sfoster1: discussed in person, keeping these

PR #15184, opened 2024-05-14
feat(ai-server): switch to custom runtime

y3rsh:
## Overview

- Switch to custom runtime.  Due to the size of packages we need and our own data a custom runtime is required.  This also allows us to HTTP stream.  
- It is likely that using llama-index in the lambda will not be performant but this gets us started until we may extract it into its own service.

### Test - deployed and make live-test passing
- [x] crt
- [x] sandbox
- [x] dev

### DIFF
--- a/opentrons-ai-server/README.md
+++ b/opentrons-ai-server/README.md
@@ -20,6 +20,9 @@ The Opentrons AI application's server.
    1. This will create a `.python-version` file in this directory
 1. select the node version with `nvs` or `nvm` currently 18.19\*
 1. Install pipenv and python dependencies `make setup`
+1. to build and deploy you must have
+1. AWS credentials and the right roles
+1. docker installed
 
 ## Install a dev dependency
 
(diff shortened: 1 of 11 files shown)


opentrons-ai-server/README.md
```diff
@@ -51,8 +46,16 @@ The Opentrons AI application's server.
 1. Make your changes
 1. Fix what can be automatically then lent and unit test like CI will `make pre-commit`
 1. `make pre-commit` passes
-1. deploy to sandbox `make build deploy test-live ENV=sandbox AWS_PROFILE=the-profile`
-
-## TODO
-
-- llama-index is gigantic. Have to figure out how to get it in the lambda
+1. deploy to sandbox `make deploy test-live ENV=sandbox AWS_PROFILE=the-profile`
```
Elyorcv, line 49: how would I get `the-profile` please?

Elyorcv (commented):
Looking good

opentrons-ai-server/README.md
y3rsh: getting you setup today

Elyorcv (approved):

PR #15187, opened 2024-05-15
fix(api): Parse all RTP fields strictly to fix flakiness

SyntaxColoring:
# Overview

Closes AUTH-401.

# Changelog

As described in AUTH-401, there were two problems:

1. Our Pydantic models sometimes parse run-time parameter values as the wrong types. For example, parsing `true` into `Union[float, bool, str]` could return `1.0` instead of `True`. This is because of [Pydantic's type coercion](https://docs.pydantic.dev/1.10/usage/models/#data-conversion).
2. Exactly which values get wrongly parsed into exactly which wrong types is unstable; it changes depending on things like import statements in files far away. This is apparently because, although Pydantic parses unions “left to right,” the order of “left to right” can change depending on Python’s caching of generic types.
    * https://bugs.python.org/issue42317
    * https://github.com/pydantic/pydantic/issues/7022

The solution to both of these problems is to disable Pydantic's type coercion. That solves the first problem for obvious reasons. It solves the second problem because, when parsing something like a `Union[float, bool, str]`, there will no longer be any "ties" that Pydantic needs to break by using `Union` order.

(In my opinion, this is also a good practice in general, and we should do it for everything going forward.)

For all fields in all run-time parameter models, change `int` to `StrictInt`, `str` to `StrictStr`, `bool` to `StrictBool`, and `float` to `Union[StrictInt, StrictFloat]`. That last `Union` is because numbers like `123`, without a decimal point, do not parse into `StrictFloat`.

# Test plan

* [x] In robot-server, run `pipenv run pytest tests/protocols/test_protocols_router.py::test_create_protocol_analyses_with_same_rtp_values` on commits c90840de0bfe1650b363971de81561251da8f372 and 65a57e840eb5ffcaad7fd2fb0140e9d0927fc8c1. It will fail on the first one (before the fix) and succeed on the second one (after the fix).
* [ ] Continued overall testing of run-time parameters? See risk assessment. 

# Review requests

Are there any fields or models that I missed?

Also see risk assessment.

# Risk assessment

Medium?

Changes like this can break the database if we've been accidentally storing stuff wrongly, like if we've been storing `"1.23"` where we meant to store `1.23` and this PR changes that field from `float` to `StrictFloat`.

There's also some risk that the frontend has had bugs where it sends the wrong types, but the server-side type coercion has been masking them. The server will return `422` errors now.

I'll defer to AUTH people for their assessment of how likely we are to have problems like these, and how much manual testing we need to do to be confident.

### DIFF
--- a/api/src/opentrons/protocol_engine/types.py
+++ b/api/src/opentrons/protocol_engine/types.py
@@ -885,12 +885,14 @@ class TipPresenceStatus(str, Enum):
 class RTPBase(BaseModel):
     """Parameters defined in a protocol."""
 
-    displayName: str = Field(..., description="Display string for the parameter.")
-    variableName: str = Field(..., description="Python variable name of the parameter.")
-    description: Optional[str] = Field(
+    displayName: StrictStr = Field(..., description="Display string for the parameter.")
+    variableName: StrictStr = Field(
+        ..., description="Python variable name of the parameter."
+    )
+    description: Optional[StrictStr] = Field(
         None, description="Detailed description of the parameter."
     )
-    suffix: Optional[str] = Field(
+    suffix: Optional[StrictStr] = Field(
         None,
         description="Units (like mL, mm/sec, etc) or a custom suffix for the parameter.",
     )
--- a/api/src/opentrons/protocol_engine/types.py
+++ b/api/src/opentrons/protocol_engine/types.py
@@ -4,7 +4,15 @@ import re
 from datetime import datetime
 from enum import Enum
 from dataclasses import dataclass
-from pydantic import BaseModel, Field, validator
+from pydantic import (
+    BaseModel,
+    Field,
+    StrictBool,
+    StrictFloat,
+    StrictInt,
+    StrictStr,
+    validator,
+)
 from typing import Optional, Union, List, Dict, Any, NamedTuple, Tuple, FrozenSet
 from typing_extensions import Literal, TypeGuard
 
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_engine/types.py
sanni-t, line 1: I don't think it's necessary to do strict typing of things that our code generates after type-checking and validating user/ client input data. So strict typing `RTPBase` and its derivatives feels very unnecessary.

SyntaxColoring: Sorry, I don't think I follow?

Take `EnumParameter`, for example. Its JSON-serialized form is stored in the database (`RunStore.update_run_state()`) and then, later, read from the database (`RunStore.get_run_time_parameters()`). When it's read from the database, its `value` and `default` are parsed by Pydantic. (Please correct me if I'm wrong.) When that happens, it will suffer the same `Union`-of-coerced-types hazards described in the PR description.

In other words, this is a hazard *any time we use Pydantic to parse JSON values,* regardless of whether the underlying data was previously type-checked and validated.

sanni-t: Ah! I forgot there's a `RunStore.get_run_time_parameters()` and that it uses pydantic parsing. Ya that makes sense then.

sanni-t (approved):
LGTM! Thanks for fixing this

sanni-t:
@ncdiehl11 verified that there are no app-side changes for this

SyntaxColoring:
Thank you both for the quick reviews and tests! This was an adventure. 🗺️ 

[Edit] Oh, is it correct for this to merge into `edge`? Discussing in Slack.

y3rsh:
Edge please.  Not enough time to test before the release.  Especially with possible DB side effects.

I need to add type assertions on all the RTP at runtime in my snapshot test.  On my list now.

PR #15188, opened 2024-05-15
refactor(react-api-client): fix react query key object hashing for useAllCommandsAsPreSerializedList

ncdiehl11:
# Overview

React query does not properly handle objects passed in useQuery's queryKey array argument that contain keys with `undefined` values. Here, in `useAllCommandsAsPreSerializedList`, I map undefined values to null so that they are properly cached and do not trigger a refetch when the host object does not change.

_In the future, I will provide a utility for all `react-api-client` hooks that will sanitize objects with possible undefined values to be included in the queryKey._

[tanstack reference](https://github.com/TanStack/query/issues/3741)

# Test Plan

- start up protocol setup
- cancel the run before beginning
- verify that the run preview does not flicker after the run becomes terminal and the 'Run was never started' info screen shows

# Changelog

- map `host` object `undefined` values to `null`

# Review requests

@b-cooper per collab

# Risk assessment

low


sfoster1 (approved):
So gross

PR #15189, opened 2024-05-15
refactor(react-api-client): sanitize host object undefined values in all react-api-client hooks

ncdiehl11:
# Overview

React query does not properly handle objects passed in useQuery's queryKey array argument that contain keys with undefined values. Here, I introduce a new utility in React API client that replaces undefined values of an object with null values. This allows for proper caching of objects in queryKey arrays.

[reference](https://github.com/Opentrons/opentrons/pull/15188)

# Test Plan

- [x] smoke test throughout app
- [ ] inspect number of requests over time in the app with an without sanitized host included in queryKey 

# Changelog
- create utility for converting `undefined` object values to `null`
- implement in every react api client hook using a `host` object in queryKey

# Review requests

# Risk assessment

medium, on the basis of touching nearly the entirety of the react api client


react-api-client/src/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Replace undefined values of an object with null values for use in useQuery query key argument
+ * @example
+ * getSanizedQueryKeyObject({key1: 'value1', key2: null, key3: undefined}):
+ * {key1: 'value1', key2: null, key3: null}
+ *
+ * @param {Object | null} obj - object to replace undefined values
+ *
+ * @returns {Object | null} Returns value-updated object
+ */
+export function getSanitizedQueryKeyObject(obj: Object | null): Object | null {
```
sfoster1, line 13: If this is specific to sanitizing things, then let’s not have it be (Object | null): Object, let’s have it be either <T>(T | null): T or have an explicit instance that’s ergonomic for the most common use case of host config, which is (Host): HostConfig. Doing that last one (Host might not be the right type name, I mean whatever useHost returns) means we can drop all those as HostConfig clauses at every point of use, which means that if you forget to use this then tsc will yell at you.

PR #15190, opened 2024-05-15
chore(api, robot-server): Manage runners via run orchestrator 

TamarZanzouri:
# Overview

part of https://opentrons.atlassian.net/browse/EXEC-418. 
create RunOrchestrator provider to manage, engine->runners scope.
this tickets is just to introduce the term and redirect calls via the orchestrator.

# Test Plan

- uploading protocols and running them should not change
- issuing commands without runs should not change 

# Changelog

- Added RunOrchestrator provider to manage multiples runners with PE. 
- change engine_store runners/engine props to use RunOrchestrator instead. 
- create RunOrchestrator when creating a run. 

# Review requests

do we feel good about these changes? 

# Risk assessment

low. redirecting calls via orchestrator but structure stayed the same.


api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,80 @@
...
+
+from . import protocol_runner
+from ..hardware_control import HardwareControlAPI
+from ..protocol_engine import ProtocolEngine
+from ..protocol_engine.types import PostRunHardwareState
+from ..protocol_reader import JsonProtocolConfig, PythonProtocolConfig
+
+
+class RunOrchestrator:
+    _json_or_python_runner: Optional[protocol_runner.AnyRunner]
```
sfoster1: i think we should think about these runners as filling semantic purposes, so this is the `_protocol_runner`, and maybe is `Union[protocol_runner.PythonAndLegacyRunner, protocol_runner.JSONRunner, None]`

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,80 @@
...
+    _setup_runner: protocol_runner.AnyRunner
```
sfoster1: and then these are `: protocol_runner.LiveRunner`

api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -411,9 +411,9 @@ async def run(  # noqa: D102
 
 
 def create_protocol_runner(
-    protocol_config: Optional[Union[JsonProtocolConfig, PythonProtocolConfig]],
     protocol_engine: ProtocolEngine,
     hardware_api: HardwareControlAPI,
+    protocol_config: Optional[Union[JsonProtocolConfig, PythonProtocolConfig]] = None,
```
SyntaxColoring: Nit: do we really want this to default to `None`?

This is not like the other things that are defaulting to `None`, because `protocol_config=None` means "use a live runner," which is not a safe default choice, in general. I think the caller really does need to think about this param, which suggests to me that it should be explicit.

robot-server/robot_server/runs/engine_store.py
```diff
@@ -126,32 +120,32 @@ def __init__(
         self._robot_type = robot_type
         self._deck_type = deck_type
         self._default_engine: Optional[ProtocolEngine] = None
-        self._runner_engine_pair: Optional[RunnerEnginePair] = None
         hardware_api.register_callback(_get_estop_listener(self))
 
     @property
     def engine(self) -> ProtocolEngine:
```
TamarZanzouri, line 126: wanted to change this to a get method instead of a property now that its a property of the orchestrator but bc its a store it dosent really match?

robot-server/robot_server/runs/engine_store.py
```diff
@@ -214,6 +206,7 @@ async def create(
         else:
             load_fixed_trash = False
 
+        # move this to the orchestrator
```
TamarZanzouri:
```suggestion
```

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,103 @@
...
+            hardware_api=hardware_api,
+            # post_run_hardware_state=post_run_hardware_state,
+            # drop_tips_after_run=drop_tips_after_run,
+        )
+        json_or_python_runner = None
+        if protocol_config:
+            json_or_python_runner = cast(  # or we can use type ignore
+                Union[
+                    protocol_runner.JsonRunner, protocol_runner.PythonAndLegacyRunner
+                ],
```
SyntaxColoring: Is this because we know we're passing a JSON or Python, non-`None`, `protocol_config` to `create_protocol_runner()`?

If so, I think the "right" solution is to define `@typing.overload`s on `create_protocol_runner()` so mypy can see that this particular call will never return a `LiveRunner`.

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @property
+    def engine(self) -> ProtocolEngine:
+        """Get the "current" persisted ProtocolEngine."""
+        return self._protocol_engine
+
+    @property
+    def runner(self) -> AnyRunner:
+        """Get the "current" persisted ProtocolRunner."""
+        return self._protocol_runner or self._setup_runner
```
SyntaxColoring, line 62: Will this property's return value need to change intelligently between runners depending on the run state? Like, it returns `self._setup_runner` while the run is in the setup phase, and `self._fixit_runner` while error recovery is active, and `self._protocol_runner` otherwise? Or am I misunderstanding the plan?

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,103 @@
...
+
+from . import protocol_runner, AnyRunner
+from ..hardware_control import HardwareControlAPI
+from ..protocol_engine import ProtocolEngine
+from ..protocol_engine.types import PostRunHardwareState
+from ..protocol_reader import JsonProtocolConfig, PythonProtocolConfig
+
+
+class RunOrchestrator:
+    """Provider for runners and associated protocol engine."""
```
SyntaxColoring: Nitpick, I know this is all preliminary: I'd find it very helpful to have a detailed docstring here to describe the responsibilities and non-responsibilities of the `RunOrchestrator`.

api/src/opentrons/protocol_runner/run_orchestrator.py
DerekMaggio, line 1: Am I correct in assuming that we will only have one `active` runner at a time, as in, only one runner making changes at a time? 

If so, should we add some logic over the top of the 3 possible runners so that when you activate one, you deactivate the other 2?

Also, is it always inherently true, that you will never have a `json_or_python_protocol_runner` set on the object until after setup_runner has run through all of its commands?

robot-server/robot_server/runs/engine_store.py
```diff
@@ -126,32 +120,32 @@ def __init__(
...
         self._default_engine: Optional[ProtocolEngine] = None
-        self._runner_engine_pair: Optional[RunnerEnginePair] = None
         hardware_api.register_callback(_get_estop_listener(self))
 
     @property
     def engine(self) -> ProtocolEngine:
         """Get the "current" persisted ProtocolEngine."""
-        if self._runner_engine_pair is None:
+        if self._run_orchestrator is None:
             raise NoRunnerEnginePairError()
```
DerekMaggio: Should we change the name of the error here?

api/src/opentrons/protocol_runner/run_orchestrator.py
TamarZanzouri: thats not a bad idea. I was thinking about this more like:
if there is a protocol runner always return that, if there is no protocol runner then fixit_runner is not active and we are only able to use setup runner but! this property will probably vanish bc we are going to remove access directly with the runner and let the orchestrator handle the logic. let me know if this makes sense.

api/src/opentrons/protocol_runner/run_orchestrator.py
TamarZanzouri: what do you mean by active? all 3 are able to add commands at the same time (active at the same time) the one that will manage what command will get executed at a given moment will be the run orchestrator.

robot-server/robot_server/runs/engine_store.py
TamarZanzouri: I was debating about it bc technically it still means the same thing. there is no engine/runner available. I will change the name a bit. thank you!

api/src/opentrons/protocol_runner/run_orchestrator.py
DerekMaggio: Ignore this question. I found the run orchestrator proposal. 
I had a fundamental misunderstanding of what the run orchestrator is supposed to be doing

api/src/opentrons/protocol_runner/run_orchestrator.py
sfoster1: Yeah I think this property goes away forever and the orchestrator itself offers semantically relevant api, like a method called run_setup_command and a method called run_fixit_command

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,105 @@
...
+        ] = None,
+        post_run_hardware_state: PostRunHardwareState = PostRunHardwareState.HOME_AND_STAY_ENGAGED,
+        drop_tips_after_run: bool = True,
+        run_id: Optional[str] = None,
+    ) -> "RunOrchestrator":
+        """Build a RunOrchestrator provider."""
+        setup_runner = protocol_runner.LiveRunner(
+            protocol_engine=protocol_engine,
+            hardware_api=hardware_api,
+            # post_run_hardware_state=post_run_hardware_state,
```
sfoster1: Checked in comments

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -0,0 +1,105 @@
...
+            # drop_tips_after_run=drop_tips_after_run,
+        )
+        fixit_runner = protocol_runner.LiveRunner(
+            protocol_engine=protocol_engine,
+            hardware_api=hardware_api,
+            # post_run_hardware_state=post_run_hardware_state,
```
sfoster1: Checked in comments

robot-server/robot_server/runs/engine_store.py
sfoster1: I think it’s fine to keep it as a property since in theory this will go away in favor of something that returns the orchestrator

robot-server/robot_server/runs/engine_store.py
```diff
@@ -126,32 +120,32 @@ def __init__(
...
         """Get the run identifier associated with the current engine/runner pair."""
         return (
-            self._runner_engine_pair.run_id
-            if self._runner_engine_pair is not None
+            self._run_orchestrator.run_id
+            if self._run_orchestrator is not None
             else None
         )
 
+    # TODO(tz, 2024-5-14): probably dont need this once its all redirected via orchestrator
```
sfoster1: Definitely will - there will need to be a different default_orchestrator that has a different engine and probably only setup and fixit runners

robot-server/robot_server/runs/engine_store.py
sfoster1: Let’s either get rid of this or format it as a todo

sfoster1 (approved):
Functionality looks good but let’s get those comments cleaned up! Great first step.

sfoster1 (approved):
Nice!

PR #15191, opened 2024-05-15
refactor(api): Move opentrons.initialize() to its own file

SyntaxColoring:
# Overview

PR #15171 runs into a circular import problem. We can fix it by paring down the contents of `api`'s top-level `__init__.py`, which might *also* have the benefit of improving performance (see explanation in PR #14480), so this does that. This goes towards EXEC-456.

# Test plan

None needed.

# Changelog

Move `opentrons.initialize()` to `opentrons.hardware_control.initialization.initialize()`, a new file.

# Review requests

* Does the new location make sense?
* Are there any obvious ways to fix the various things that are importing the private `_find_smoothie_file()` function?
* Is there an obvious place where it would make sense to move `opentrons.should_use_ot3()`, which currently remains in the top-level `__init__.py`?

# Risk assessment

Low.


api/src/opentrons/_resources_path.py
SyntaxColoring, line 1: I put this in its own file to keep it close to the `resources/` directory that it's referencing. Otherwise we'd be trying to access it from inside `hardware_control/` with some fragile relative path.

api/src/opentrons/__init__.py
```diff
@@ -1,39 +1,19 @@
...
-    name,
-    robot_configs,
-    IS_ROBOT,
-    ROBOT_FIRMWARE_DIR,
-)
-from opentrons.util import logging_config
+from typing import List
+
+from opentrons.config import feature_flags as ff
 from opentrons.protocols.types import ApiDeprecationError
```
sfoster1: you could even move these down to be dynamic imports inside the module getattr, that'd be really awesome. `opentrons.protocols.types` and `opentrons.protocols.api_support.types` are pretty big!

sfoster1 (approved):
This is awesome and could even take it a step farther

api/src/opentrons/__init__.py
SyntaxColoring: Great idea.

api/src/opentrons/__init__.py
```diff
@@ -1,47 +1,33 @@
...
+__all__ = ["version", "__version__", "config"]
 
 
 def __getattr__(attrname: str) -> None:
-    """
-    Prevent import of legacy modules from global to officially
-    deprecate Python API Version 1.0.
+    """Prevent import of legacy modules from global.
+
+    This is to officially deprecate deprecate Python API Version 1.0.
```
vegano1: Dup "deprecate"

vegano1 (approved):
This makes sense 👍

api/src/opentrons/__init__.py
SyntaxColoring: TY!

PR #15192, opened 2024-05-15
feat(app): setup advanced settings/tip management state for quick transfer

smb2268:
fix PLAT-227

# Overview
This is a setup PR that doesn't add functionality. It adds the state and reducers needed for implementing the advanced settings and tip management portions of the quick transfer flow.

# Test Plan
This doesn't impact functionality so no need to test

# Changelog
1. Rename `QuickTransferSetupState` to `QuickTransferWizardState`
2. Create `QuickTransferSummaryState` type and `getInitialSummaryState` util that will generate the starting point for the state including default values based on what was selected in the wizard flow. There are a few `TODO` comments in here that will be addressed when those fields' screens are implemented.
3. Add `quickTransferSummaryReducer` that will allow us to amend the state based on selections made in the advanced settings and tip management fields

# Review requests
Look through utils for clarity

# Risk assessment
Low, behind FF


app/src/organisms/QuickTransferFlow/utils.ts
```diff
@@ -110,6 +115,172 @@ export function quickTransferReducer(
...
+  tipRack: LabwareDefinition2
+  source: LabwareDefinition2
+  sourceWells: string[]
+  destination: LabwareDefinition2 | 'source'
+  destinationWells: string[]
+  transferType: TransferType
+  volume: number
+}): QuickTransferSummaryState {
+  const tipVolume = Object.values(props.tipRack.wells)[0].totalLiquidVolume
+  const tipType = `t${tipVolume}`
```
smb2268: Is there a better way to derive `tipType` from the tiprack labware definition? @CaseyBatten maybe you know?

app/src/organisms/QuickTransferFlow/constants.ts
```diff
@@ -6,6 +6,24 @@ export const ACTIONS = {
   SET_DEST_LABWARE: 'SET_DEST_LABWARE',
   SET_DEST_WELLS: 'SET_DEST_WELLS',
   SET_VOLUME: 'SET_VOLUME',
+  SUMMARY_SETUP: 'SUMMARY_SETUP',
```
brenthagen: what is `SUMMARY_SETUP`?

brenthagen (approved):
LGTM !

app/src/organisms/QuickTransferFlow/constants.ts
smb2268: I'll remove - initially had the `getInitialSummaryState` as a reducer action and left this in by mistake

app/src/organisms/QuickTransferFlow/utils.ts
smb2268: Talked with Brian and confirmed this is the right method, but going to move this into a shared-data helper

PR #15193, opened 2024-05-15
refactor(opentrons-ai-client): refactoring api call and inputprompt component

koji:
# Overview
refactoring api call and input prompt component
add a new package for testing axios

close AUTH-402

# Test Plan
1. `make teardown-js && make setup-js`
2. `make -C opentrons-ai-client dev`
3. open localhost:5173 (the port must be 5173 since the backend allows us to use only 5173 port)
4. log in (if you don't have an account, you will need to sign up)
5. send a prompt
check the app displays your prompt and gets the return from OpenAI API

# Changelog
- add custom hook for callApi and get access token
- export a function as a util function
- update paragraph text style to align the font styles

# Review requests

# Risk assessment
low


koji:
- [x] add a custom hook for data fetch
- [x] add a custom hook for token retrieving
- [x] update tests
- [x] add test for `useGetAccessToken`
- [x] fix fetch functionality
- [x] merge edge

opentrons-ai-client/src/resources/utils/utils.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Calculates the number of lines in a given string.
+ * @param input - The string to calculate the number of lines for.
+ * @returns The number of lines in the input string.
+ */
+
+export const calcTextAreaHeight = (input: string): number => {
+  const rowsNum = input.split('\n').length
+  return rowsNum
```
ncdiehl11: nit, maybe just return the expression instead of assigning a const?

ncdiehl11 (commented):
Code all looks good, just having trouble testing the response after posting a prompt. Maybe we can chat IRL about how to test?

opentrons-ai-client/src/resources/utils/utils.ts
koji: good catch, thanks!

Elyorcv:
@koji please rebase

y3rsh (approved):
Works for me against #15287

PR #15194, opened 2024-05-16
chore: add known issue for losing LPC offsets when swapping labware type via RTP

ecormany:
# Overview

Letting people know they can get chicken-and-egg'd if they apply labware offsets, set RTPs that change their labware, and then reanalysis gets rid of their original offsets.

# Review requests

Is this clear? This is kind of a tricky one to wrap up in a single sentence.

# Risk assessment

zip


nusrat813 (approved):
LGTM

PR #15198, opened 2024-05-16
feat(robot-server): Return a `currentlyRecoveringFrom` command pointer

SyntaxColoring:
# Overview

Goes towards EXEC-458.

# Test plan

I'm unable to push this to a robot because of problems that appear unrelated (`ModuleNotFoundError: No module named 'usb'`). I think this is low-risk enough that we can get away with skipping it.

# Changelog

This adds a `currentlyRecoveringFrom` command pointer to `GET /runs` and `GET /maintenance_runs`, adjacent to `current`, and wires it up to the underlying `ProtocolEngine`.

This does not yet send MQTT notifications for it, nor does it update the frontend code to take advantage of the new pointer. Doing those in follow-up PRs.

# Review requests

~~API design. Is this helpful to the client? Does it have simple and maintainable semantics? The term "current command" has historically been the subject of much confusion, and I wouldn't want the same thing to happen here. Would it be better or worse to define this more literally, like `mostRecentFailedProtocolCommand`?~~

# Risk assessment

Low.


robot-server/robot_server/runs/command_models.py
```diff
@@ -0,0 +1,66 @@
...
+
+class CommandCollectionLinks(BaseModel):
+    """Links returned along with a collection of commands."""
+
+    current: Optional[CommandLink] = Field(
+        None,
+        description="Information about the currently running or next queued command.",
+    )
+
+    recoveryTarget: Optional[CommandLink] = Field(
```
DerekMaggio: Could we call this something like `currentlyRecoveringFrom`
I have always been a fan of adding a bit of verbosity in the name of clarity

sfoster1 (commented):
I think that “recovery target” is kind of a misleading name for this, because you might not actually be recovering from the command that gets pointed to there - it might not be recoverable, for instance, or it might already have been recovered from: if command C fails, is recovered, and the protocol succeeds, command C will continue to be the thing that gets returned.

I think this should be “lastFailedProtocolCommand” or something specific like that.

SyntaxColoring:
> I think that “recovery target” is kind of a misleading name for this, because you might not actually be recovering from the command that gets pointed to there - it might not be recoverable, for instance, or it might already have been recovered from: if command C fails, is recovered, and the protocol succeeds, command C will continue to be the thing that gets returned.
>
> I think this should be “lastFailedProtocolCommand” or something specific like that.

I spoke with @mjhuff about this, and it turns out we do *not* actually want command C to continue to be the thing that gets returned in this case. It's easier for the client if the return value goes back to being `null` after a command is recovered from.

api/tests/opentrons/protocol_engine/state/test_command_state.py
```diff
@@ -356,8 +356,8 @@ def test_error_recovery_type_tracking() -> None:
     assert view.get_error_recovery_type("c2") == ErrorRecoveryType.FAIL_RUN
 
 
-def test_get_recovery_in_progress_for_command() -> None:
-    """It should return whether error recovery is in progress for the given command."""
+def test_recovery_target_tracking() -> None:
```
DerekMaggio, line 359: What about the case where the command failed recoverably, but we are not yet trying to recover from it?

api/tests/opentrons/protocol_engine/state/test_command_state.py
```diff
@@ -439,7 +442,8 @@ def test_get_recovery_in_progress_for_command() -> None:
     subject.handle_action(fail_3)
 
     # c3 failed, but not recoverably.
-    assert not subject_view.get_recovery_in_progress_for_command("c2")
+    assert subject_view.get_recovery_target_id() is None
+    assert not subject_view.get_recovery_in_progress_for_command("c3")
```
DerekMaggio: What about the case where the command failed recoverably, but we are not yet trying to recover from it?

SyntaxColoring: Good question. There is no such state. If a command failed recoverably, we're trying to recover from it sort of by definition.

robot-server/robot_server/runs/command_models.py
SyntaxColoring: Renamed to `currentlyRecoveringFrom`.

mjhuff (approved):
This makes sense to me. As discussed in person, we might think about extending this to other routes in the future, but /commands is as much as we need right now. Thank you for implementing this!

EDIT: I can help test tomorrow if you'd like, but code LGTM.

api/tests/opentrons/protocol_engine/state/test_command_state.py
DerekMaggio: Is ProtocolEngine already in a recovery state before the customer clicks the "Begin Recovery" modal?

SyntaxColoring: Yes.

PR #15200, opened 2024-05-16
chore(api): clean up lld hardware layer

ryanthecoder:
# Overview

This PR removes some uneeded arguments and fixes a few #FIXME  that were hanging around

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -2580,13 +2580,22 @@ class OT3API(
         if probe_settings.aspirate_while_sensing:
             await self._move_to_plunger_bottom(mount, rate=1.0)
         else:
-            # TODO: shorten this distance by only moving just far enough
-            #       to account for the specified "max-z-distance"
-            target_pos = target_position_from_plunger(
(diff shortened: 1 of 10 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2580,13 +2580,22 @@ async def liquid_probe(
...
-            #       to account for the specified "max-z-distance"
+            # find the ideal travel distance by multiplying the plunger speed
+            # by the time it will take to complete the z move.
+            ideal_travel = probe_settings.plunger_speed * (
+                probe_settings.max_z_distance / probe_settings.mount_speed
+            )
+
+            # TODO limit the z distance to the max  allowed by the plunger travel at this speed.
+            # or here is probably the ideal place to implement multi-probe
+            assert (
```
sfoster1, line 2591: This probably wants to be an actual enumerated error rather  than just an assert. I know we plan to change this code further before ship, but having a nice error here means even if we don’t change it we’ll be ok.

sfoster1 (approved):
Looks good but let’s make it a real error instead of an assert.

PR #15202, opened 2024-05-16
refactor(app): make on device display deck configurator updates live

b-cooper:
# Overview

Instead of the previous method of staging a batch of changes to deck configuration before commiting
them with the confirm button, to avoid synchronization issues, make the ODD deck configurator behave
like the desktop version. all updates immediately permeate to the robot's source of truth

Re: https://opentrons.atlassian.net/browse/RQA-2669

# Review requests

Updating deck configuration from the 3-dot menu on the ODD should be visible on every client as soon as any fixture is added or removed. The same should go for updates that originate from a desktop client editing the same robot's deck config.

# Risk assessment
low


brenthagen (approved):
tested on ODD/desktop, changes make sense to me

PR #15206, opened 2024-05-16
build(fix): add the pyusb to dev-packages in api/Pipfile.

vegano1:
# Overview

Commit [#15167](https://github.com/Opentrons/opentrons/pull/15167) added support for the plate reader, which required a `pyusb` a Python dependency. The device works as expected on the robot, but we also use parts of the Python stack for robot analysis on the app and ci, which uses API dev-packages. So let's add `pyusb` to the dev-packages section in the api/Pipfile file.

Note: Make sure you run `make -C api teardown && make -C api setup` when you pull in this change. 

# Test Plan

- [x] Make sure that protocol analysis passes in CI
- [x] Make sure that protocol analysis passes in the desktop app

# Changelog

- add `pyusb` dep to dev-packages for api project

# Review requests
# Risk assessment
low


sfoster1 (changes requested):
If this dep is required for analysis then it also needs to be in the setup.py install_requires so that people installing from pypi won’t have broken installs.

vegano1:
@sfoster1 

Analyses Snapshot Test is no longer failing because of the missing pyusb dependency.
It's not failing because the snapshots are no longer valid and need to be updated.

https://github.com/Opentrons/opentrons/actions/runs/9197098594/job/25297114720

app-testing/Pipfile
```diff
@@ -18,6 +18,7 @@ ruff = "==0.3.4"
 docker = "==7.0.0"
 syrupy = "==4.6.1"
 pytest-html = "==4.1.1"
+pyusb = "==1.2.1"
```
y3rsh: ### This should not be needed.
The packages here are not a part of the environment where analysis is run.
A docker container
https://github.com/Opentrons/opentrons/blob/edge/app-testing/citools/Dockerfile
Mimics the build of the app. 
https://github.com/Opentrons/opentrons/blob/edge/app-shell/scripts/before-pack.js
> [!NOTE]
> If the analysis does not run in this container then I don't think it would in the app.
> Somehow `pyusb` must be a part of the packages that get bundled with the app.
> Adding it to /api should do that.

> Let us remove just this change to `app-testing/Pipfile` and run the analysis snapshot again.  It did fix the issue https://github.com/Opentrons/opentrons/pull/15212

sfoster1 (approved):
Shouldn't be needed as much after https://github.com/Opentrons/opentrons/commit/ccc944efb238a0f9a5d32f04aeeb308018993c97

ahiuchingau:
Do we still need this?

PR #15209, opened 2024-05-16
fix(app): clear intervention modal if run is finishing

ncdiehl11:
Closes RQA-2730

# Overview

Remove InterventionModal if run is finishing. Implement useNotifyLastRunCommand in desktop RunProgressMeter

# Test Plan

- run a protocol in which the last command is an intervention command (pause, etc.)
- resume the protocol on desktop app
- verify that the modal clears when the run is finishing
- repeat the above steps on ODD

# Changelog

- introduce `useNotifyLastRunCommand` hook to `RunProgressMeter`
- check for run finishing status to remove intervention modal

# Review requests

@b-cooper per work together

# Risk assessment

low


b-cooper (approved):
We both agreed that you can revert the RunProgressMeter change, but after that reversion, this looks good to go

PR #15215, opened 2024-05-17
fix(api): restore empty error blocks on cancelled runs

sfoster1:
I think this would be a subtle side effect of a previous pr trying to improve PE end behavior. 

The reason this fix is required is that when a client cancels a protocol by stopping the engine, the `StopAction` sets the run result (good) and doesn't set a run error (good, correct, there wasn't one, this is a cancel). In 7.2, with the behavior this PR now restores, a `FinishAction` that might contain an error wouldn't set that error in the run if it already had a _result_, whether or not it had an _error_. In 7.3, it would set the error if the engine had no error, which it wouldn't, because there is no error when you stop.

## Testing
- [x] With this PR, cancelling a run on a real robot causes the ODD to show an inactive error details button
- [x] With this PR, stopping a run by hitting an estop still causes the ODD and desktop to display that an estop error cancelled the run

Closes RQA-2737

### DIFF
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -390,6 +390,15 @@ class CommandStore(HasState[CommandState], HandlesActions):
                         action.error_details.created_at,
                         action.error_details.error,
                     )
+            else:
+                # HACK(sf): There needs to be a better way to set
+                # an estop error than this else clause
+                if self._state.stopped_by_estop and action.error_details:
+                    self._state.run_error = self._map_run_exception_to_error_occurrence(
+                        action.error_details.error_id,
+                        action.error_details.created_at,
+                        action.error_details.error,
+                    )
 
         elif isinstance(action, HardwareStoppedAction):
             self._state.queue_status = QueueStatus.PAUSED
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -384,12 +384,12 @@ class CommandStore(HasState[CommandState], HandlesActions):
                 else:
                     self._state.run_result = RunResult.STOPPED
 
-            if not self._state.run_error and action.error_details:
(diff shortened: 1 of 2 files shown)


SyntaxColoring (commented):
Whoof, nice find, sorry about that.

The tests show this failing some E-stop-related test that I added in https://github.com/Opentrons/opentrons/pull/14929. I'm trying to remind myself of its implications.

There's certainly a way to solve both problems at once, but I'm a little reluctant to do more PE state work this late in v7.3. What do you think?

vegano1:
Wow subtle indeed, this fixed the issue.
Tested this on appui bot and cannot repro anymore.
Thank you!

sfoster1:
> Whoof, nice find, sorry about that.
> 
> The tests show this failing some E-stop-related test that I added in #14929. I'm trying to remind myself of its implications.
> 
> There's certainly a way to solve both problems at once, but I'm a little reluctant to do more PE state work this late in v7.3. What do you think?

The implication would be that, conversely, an estop error _would not_ get reported. 

https://github.com/Opentrons/opentrons/pull/15215/commits/4c30bf5d0fb833762cffb0bace7add9b01236004 i'm pretty sure restores the exact thing your PR was meant to remove but such is life on a timeline, and I think it will work.

SyntaxColoring (approved):
OK, sounds like a plan, presuming it passes the on-robot tests. (Thank you @vegano1 for testing!)

vegano1:
The change is working as expected, see attached images.
- The error modal is no longer shown when the run is manually canceled
- The error modal shows the appropriate error when the e-stop is pressed

Screen Shot 2024-05-17 at 5 43 03 PM
Screen Shot 2024-05-17 at 5 38 07 PM
Screen Shot 2024-05-17 at 5 40 19 PM

vegano1 (approved):
Tested on app ui bot, working as expected.

sfoster1:
> Tested on app ui bot, working as expected.

Thank you!

PR #15217, opened 2024-05-20
refactor(api): extract RTP definitions and set values during runner load

sanni-t:
Addresses AUTH-282

# Overview

First part of the two-part changes required to make RTP values be available earlier during analysis.
This PR moves extraction of RTP definitions and setting the values of the params to run load stage. This makes sure that the final RTP values are available to the runner before the run is played.

# Test Plan

Mainly need to make sure that current protocol execution hasn't changed from any client perspective. Unit and integration tests passing should be enough.

# Review requests

- usual code review and make sure I am not missing anything related to the python protocol execution pipeline

# Risk assessment

Low. Refactor


TamarZanzouri (approved):
LGTM!

PR #15218, opened 2024-05-20
fix(protocol-designer): do not duplicate into adjacent h-s slot for o…

jerader:
…t-2 tall labware

closes RQA-2716

# Overview

This bug has certainly existed in production since PD 6.0.0 i think. Shout out to Derek for finding it. Basically, the heater-shaker on the ot-2 has many collision warnings and any tall labware greater than 53mm can hit the heater-shaker when it is in motion. So this PR filters out the adjacent slots of the heater-shaker when duplicating the labware on the deck and the initial set up when the tipracks are auto generated.

# Test Plan

Follow the steps outlined in the ticket. Basically, you don't want the tiprack auto generating or duplicating into slot 2.

# Changelog

- add logic to special case the heater-shaker tall labware on ot-2 for create file wizard and get next available slot util

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/modals/CreateFileWizard/index.tsx
+++ b/protocol-designer/src/components/modals/CreateFileWizard/index.tsx
@@ -17,6 +17,7 @@ import {
   THERMOCYCLER_MODULE_TYPE,
   FLEX_ROBOT_TYPE,
   WASTE_CHUTE_CUTOUT,
+  getAreSlotsAdjacent,
 } from '@opentrons/shared-data'
 import { actions as stepFormActions } from '../../../step-forms'
 import { INITIAL_DECK_SETUP_STEP_ID } from '../../../constants'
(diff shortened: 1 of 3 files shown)


koji:
This PR can handle the case of [RQA-2716](https://opentrons.atlassian.net/browse/RQA-2716), but seems that this PR still cannot handle the following case.

https://github.com/Opentrons/opentrons/assets/474225/6a98b3da-9b39-4c03-81ce-655897bf04bd

I guess this would be a kind of an edge case so I'm not sure that we would need to handle that case. (But probably showing warning would be useful for users.)

[RQA-2716]: https://opentrons.atlassian.net/browse/RQA-2716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

koji (approved):
I left a comment but this bug has been existing from v6.0 so the case I mentioned wouldn't be a big issue at this moment.

jerader:
> I guess this would be a kind of an edge case so I'm not sure that we would need to handle that case. (But probably showing warning would be useful for users.)

@koji thank you for catching this! I am going to make a new ticket and put it in the backlog. In PD 8.2, i think we won't run into that issue since we are redesigning everything but i still want to document it at least.

PR #15219, opened 2024-05-20
fix(protocol-designer): enable arrow when reaching max slots 

jerader:
closes RQA-2718

# Overview

When reaching max capacity on the deck, the down arrow for the multiple temp modules was disabled. This PR fixes that

# Test Plan

add 7 temperature modules to reach max capacity and then press the down arrow to delete them

# Changelog

- modify on click cta 
- fix bug with the TC disabled

# Review requests

see test plan

# Risk assessment

low


koji (approved):

PR #15221, opened 2024-05-20
fix(protocol-designer): disabled color logic and pause step item ui

jerader:
closes RQA-2710 RQA-2720

# Overview

Fixes a few bugs found in RC-B:
1. submit button disabled reason for adding a color 
2. removing the empty quotes from the pause step item

# Test Plan

- create a flex or ot-2 protocol.
- Add a color and first add a description followed by a name. see that the submit button is disabled until a name is added
- go to the deck map and add a pause with no message. create the pause step and see on the timeline that the pause step item does not show empty quotes

# Changelog

- refine logic for showing the message for a pause step item
- refine disabled logic for when the submit button is disabled for the liquid

# Review requests

see test plan

# Risk assessment

low


ncdiehl11 (approved):
🚀

PR #15223, opened 2024-05-20
docs(robot-server): Correct and unify the definition of "current"

SyntaxColoring:
# Overview

This tries to resolve some confusion around the `cursor` query parameter and `links.current` field in the `GET /runs/{id}/commands` and `GET /maintenance_runs/{id}/commands` endpoints.

# Test plan

None needed.

# Changelog

`links.current` is a bit confusing:

* The public OpenAPI spec says that it returns the command that's currently running or **next queued.**
* The docstring on the internal low-level Python code says that it returns the command that's currently running or **most recently completed.**
* The Python implementation appears to actually return the command that's currently running or **most recently completed.**

Compare with the `cursor` *query param* of `GET /runs/{id}/commands`, which appears consistent about returning the currently running or **most recently completed** command.

This PR changes the OpenAPI spec of `links.current` to say it's the currently running or **most recently completed** command. This unifies everything. I'm working under the premise that this reflects what we have now, and it's working fine, and we have no reason to change it.

# Review requests

Am I misreading the `StateView.get_slice()` and `StateView.get_current()` implementation logic and misunderstanding the current behavior?

If this is the current behavior, is my premise correct that it's fine and we have no reason to change it?

# Risk assessment

No risk.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -520,6 +520,8 @@ class CommandView(HasState[CommandState]):
         queued_command_ids = self._state.command_history.get_queue_ids()
         total_length = self._state.command_history.length()
 
+        # TODO(mm, 2024-05-17): This looks like it's attempting to do the same thing
+        # as self.get_current(), but in a different way. Can we unify them?
         if cursor is None:
             if running_command is not None:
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Looks good to me, I agree that the implementation is to return the most recently executed command.

PR #15224, opened 2024-05-20
fix(protocol-designer): default to correct tiprack uri

jerader:
closes RQA-2709

# Overview

PD now selects correct default tiprack

paired with @ncdiehl11 

# Test Plan

Create an ot-2 protocol with a p10 single and 10uL tiprack and p300 single and 300uL tiprack. 
add a labware to the deck
create a transfer step for the p10 single pipette
create a transfer step for the p300 single pipette
see that the highlighted wells are in the correct labware

# Changelog

- remove the presaved tiprack logic
- replace that presaved tiprack logic to a use effect in the tiprack field

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #15226, opened 2024-05-20
feat(opentrons-ai-client): send chat history instead user prompt

koji:
# Overview
send chat history to the backend instead only user prompt.

The currently opentrons-ai-client sends userPrompt (string) to the backend but we need to send the entire chat history to align with the application on HuggingFace.
https://huggingface.co/spaces/Opentrons/OpentronsAI-v0.0.5/blob/main/flexchatinterface.py
The sample data is in the ticket.

close AUTH-407 AUTH-414

# Test Plan

# Changelog
- add a new atom chatHistoryAtom to store chatHistory

# Review requests

# Risk assessment
low


koji:
- [ ] need to deploy the latest backend
- [ ] need to check backend log

opentrons-ai-client/src/molecules/InputPrompt/index.tsx
```diff
@@ -63,8 +68,12 @@ export function InputPrompt(): JSX.Element {
             audience: 'sandbox-ai-api',
           },
         })
+        const message: Chat = {
+          role: 'user',
+          content: prompt,
+        }
         const postData = {
-          message: prompt,
+          messages: [...chatHistory, message],
```
Elyorcv, line 76: why three dots in front of `chatHistory`?

Elyorcv:
@koji I think you also need to modify server function such that it accepts list rather currently string?

opentrons-ai-client/src/molecules/InputPrompt/index.tsx
koji: destructuring

koji:
> Reference

wait, server is not updated yet?
the format you shared in the Jira ticket might be different?

Elyorcv:
> > Reference
> 
> wait, server is not updated yet? the format you shared in the Jira ticket might be different?

If you see lines 10-11 at `function.py` you see `ChatRequest`. I assume that we need to change this one as well to account for the changes you are making. @y3rsh may comment on this.

koji:
This will be resolved by https://github.com/Opentrons/opentrons/pull/15288

PR #15228, opened 2024-05-20
fix(app-shell): reduce latency when switching between MQTT clients

mjhuff:
Closes [RQA-2748](https://opentrons.atlassian.net/browse/RQA-2748)

# Overview

Each network interface is tied to a specific IP, and the app-shell uses one IP to communicate with MQTT. If a client manually disconnects from a network interface that is connected to MQTT, some network interfaces on some robots (such as the USB on the OT-2) do not fire the MQTT 'error' or 'disconnect' events immediately. This means that the app-shell thinks it's still connected to MQTT until the keepAlive packet receives no response. 

The MQTT standard keepalive (and what we currently use) is 60 seconds. After the packet is sent, there is a connectTimeout that the server has to respond before the client fires the 'disconnect' event (we use 30 seconds, the standard). After a client disconnects, the app-shell will attempt to connect to the same robot on a different IP if available. 

This problem scenario shouldn't affect most users in most circumstances, but there is one rather simple improvement we can make that _sometimes_ reduces this latency - MQTT has an 'offline' event. Some network interfaces cause the 'offline' event to fire, and we can immediately end the client (and reestablish a new MQTT connection on a different network interface if available). 

# Test Plan
- The offline event _sometimes_ fires on the OT-2 when you disconnect USB/link-local from OT-2 (don't pull the connection out of your laptop). This seems to depend on the dongle or whatever other peripherals you use to connect the OT-2 to your laptop. 
- You can verify the issue mentioned in the ticket is related to keepalive by following the steps in the ticket, then after connecting to wifi and disconnecting USB, wait 90 seconds before starting a protocol run.

# Changelog
- The desktop app will sometimes be more responsive to robot notifications after manually disconnecting from a network interface.

# Review requests

Do we want to adjust the connectTimeout interval? I think the default is fine, but this would reduce latency. Open to thoughts/suggestions.

# Risk assessment
low - this just adds a new event listener that we weren't using before

[RQA-2748]: https://opentrons.atlassian.net/browse/RQA-2748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-shell/src/notifications/connect.ts
+++ b/app-shell/src/notifications/connect.ts
@@ -68,6 +68,7 @@ export function establishConnections(
 ): Promise<RobotData[]> {
   return new Promise((resolve, reject) => {
     const newConnections = healthyRobots.filter(({ ip, robotName }) => {
+      // Only attempt a new connection if the current broker connection is bad.
       if (connectionStore.isConnectedToBroker(robotName)) {
         return false
       } else {
@@ -171,7 +172,7 @@ function establishListeners(
   client.on('reconnect', () => {
     notifyLog.debug(`Attempting to reconnect to ${robotName} on ${ip}`)
   })
-  // handles transport layer errors only
+  // Handles transport layer errors only
   client.on('error', error => {
     notifyLog.warn(`Error - ${error.name}: ${error.message}`)
     sendDeserializedGenericError(ip, 'ALL_TOPICS')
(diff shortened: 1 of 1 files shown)


DerekMaggio (approved):
This LGTM, as for changing the default timeout, I don't think it is necessarily required, if there is any risk associated with changing it. This is the edgiest of edge cases, that only hangs around for a short period of time.

sfoster1 (approved):
Whatever works. Hard problem in general. We might need to address this piecemeal by
- detecting in the app that a usb device went away and bouncing the connection if there was one
- robot network control routes emit a new about-to-disconnect notification when you switch what network the robot is connected to
- system integration with network connection for when you switch networks on your laptop or connect/disconnect ethernet cables if possile

PR #15229, opened 2024-05-20
feat(robot-server): show run time parameters in pending analysis

sanni-t:
Closes AUTH-282

# Overview

Follow-up to #15217 

This PR changes a *pending analysis'* `AnalysisSummary` to include the run time parameters (with overrides).
In order to do this, we will now be parsing the RTPs before analysis response is created. This is a very quick process and shouldn't cause any perceptible delays in the response of neither `POST /protocols` nor `POST /runs/.../analyses` endpoints, as long as recommended python protocol guidelines are followed.

# Test Plan

- [ ] Upload a protocol with run time parameters, verify that the response contains run time parameter definitions along with the values specified in the request
- [ ] Send a `POST` request on `/protocols/{protocolId}/analyses` endpoint with different RTP values, or by specifying the `forceReanalysis` flag, and verify that the newly pending analysis also contains (updated) RTPs
- [ ] Verify that pending summaries in `GET` endpoints also contain RTPs
- [ ] Verify that summaries of completed analyses don't expose RTPs, while RTPs are available in the full analyses.

# Changelog

- Main change is the introduction of `AnalysesManager`. The analyses manager is responsible for creating a protocol analyzer per analysis, extracting the run time parameters from the runner and starting the analysis.
- To facilitate this extraction of run time parameters, the `ProtocolAnalyzer` has been divided into two functions- one that loads the runner and other that runs the runner.
- Updated the router to use `AnalysesManager` for starting analyses instead of directly interacting with `ProtocolAnalyzer`.

# Review requests

- Does the new architecture of `AnalysesManager` & updated `ProtocolAnalyzer` makes sense. 
- This PR doesn't expose the run time parameters in analysis summaries of completed analyses because of two reasons- (1) as opposed to the RTPs of a pending analysis, the completed analysis isn't guaranteed to be cached in memory. So we would need to read the completed analysis from database, de-serialize it and then return it. This would be made easier when we start storing RTPs in their own database table, but we're not there yet. (2) I'm not sure if it is needed. A completed analysis already contains the RTPs.
Would like to get others' opinions on this.
- Usual code review

# Risk assessment

- Medium. It changes a core part of protocol analysis but the end effect is a small change to the API. As long as tests are passing, it shouldn't cause any problems.
- Small risk of analysis behavior change for badly written protocols as mentioned in [this](https://github.com/Opentrons/opentrons/pull/15229#discussion_r1619075927) comment.


robot-server/robot_server/protocols/router.py
```diff
@@ -311,25 +311,43 @@ async def create_protocol(
...
+            protocol_id=protocol_id,
+            analysis_id=analysis_id,
+            run_time_parameters=[],
+        )
+        await protocol_analyzer.update_to_failed_analysis(
+            analysis_id=analysis_id,
+            protocol_robot_type=protocol_resource.source.robot_type,
+            error=error,
+            run_time_parameters=[],
+        )
```
sanni-t: This need to add to pending store and then updating to a failed analysis is weird. Making changes so that we won't have to go through a temporary intermediate state that will never be exposed to the client anyway. 
Suggestions are welcome!

SyntaxColoring: It is a bit weird, but I think it's okay. In my experience, allowing states to be "skipped over" really opens up the surface area for bugs to grow onto, so fixing the weirdness doesn't feel like a good tradeoff in this case.

-----

But I also think we might want to actually expose this intermediate state? Because it's not fundamentally instantaneous. A badly-written Python protocol (maybe copy-pasted from Jupyter Notebook) could have something like `time.sleep(60)` at the top level. I don't think we want to block `POST /protocols` for that long; I think we want to immediately return a pending analysis *without* run-time parameters, and then update the analysis to have the run-time parameters once they're available. And yeah, we should put a timeout around our `exec()`s to protect against things like this, but even then, we'd make the timeout something like 5 seconds, not instant.

In other words, before this PR, we have these resources, going across time:

> no protocol or analysis → protocol and pending analysis → protocol and completed analysis, including RTPs

And right now, this PR is doing:

> no protocol or analysis → protocol and pending analysis, including RTPs → protocol and completed analysis

But I think we actually want:

> no protocol or analysis → protocol and pending analysis, *without* RTPs → protocol and pending analysis, *with* RTPs → protocol and completed analysis

robot-server/robot_server/protocols/router.py
```diff
@@ -311,25 +311,43 @@ async def create_protocol(
...
-    pending_analysis = analysis_store.add_pending(
-        protocol_id=protocol_id,
-        analysis_id=analysis_id,
-    )
+    try:
+        run_time_parameters = await protocol_analyzer.load_runner(
+            protocol_resource=protocol_resource,
+            run_time_param_values=parsed_rtp,
+        )
+    except BaseException as error:
```
SyntaxColoring: I think we want to catch `Exception` instead of `BaseException`. As a heuristic: if something raises `KeyboardInterrupt` or `asyncio.CancelledError`, we do not want to swallow it.

I see that the `ProtocolAnalyzer` internals are also currently catching `BaseException`. I think those should be catching `Exception`, too.

robot-server/robot_server/protocols/router.py
```diff
@@ -311,25 +311,43 @@ async def create_protocol(
...
-        protocol_resource=protocol_resource,
-        analysis_id=analysis_id,
-        run_time_param_values=parsed_rtp,
-    )
-    pending_analysis = analysis_store.add_pending(
-        protocol_id=protocol_id,
-        analysis_id=analysis_id,
-    )
+    try:
+        run_time_parameters = await protocol_analyzer.load_runner(
```
SyntaxColoring: Can we encapsulate all of this `try`/`except` stuff, and the `.add_pending()`→`.update_to_failed_analysis()` thing, inside `ProtocolAnalyzer`? Because that's really what I view as `ProtocolAnalyzer`'s job. It relieves the router of this nuanced and off-topic analysis stuff.

So, ideally, I think  this router would look roughly like:

```python
protocol_store.insert(protocol_resource)

# Stores a pending analysis in-memory. Creates a background task
# for completing the analysis. When the background task completes, stores
# the results in the database.
pending_analysis_summary = protocol_analyzer.start_analysis(protocol_resource)

return Protocol(
    analysis_summaries=[pending_analysis_summary],
    ...
)
```

To accomplish this, I guess we'd need `ProtocolAnalyzer` to call `TaskRunner` itself, instead of this router calling `ProtocolAnalyzer` inside `TaskRunner`.

SyntaxColoring (commented):
Thank you! I know this is all a work in progress, but some suggestions so far:

robot-server/robot_server/protocols/router.py
sanni-t: > no protocol or analysis → protocol and pending analysis, without RTPs → protocol and pending analysis, with RTPs → protocol and completed analysis

Yep, you have a good point about badly written protocols but for this iteration, we are not concerning ourselves with it.

We've talked about this in authorship and this flow with an intermediate analysis state that can change anytime is not useful and complicates a couple of things for the app so we want to avoid it. Even on the backend, it creates a big problem for the current architecture because you are essentially creating a new state. The ProtocolAnalyzer is not a singleton but rather a setup, start and forget object. So monitoring for an intermediate state change (i.e. availability of run time parameters) within the analyzer requires quite a big change that I am trying to avoid in this work since it doesn't strictly need it.

There's a bigger architecture change being discussed where we add an AnalysesManager that can track and manage multiple analyzers and such intermediate state could be possible with that manager.

robot-server/robot_server/protocols/router.py
sanni-t: Yep, this is what the upcoming `AnalysesManager` mentioned in the [above comment](https://github.com/Opentrons/opentrons/pull/15229#discussion_r1622926744) will be doing too.

robot-server/robot_server/protocols/router.py
SyntaxColoring: > The ProtocolAnalyzer is not a singleton but rather a setup, start and forget object. So monitoring for an intermediate state change (i.e. availability of run time parameters) within the analyzer requires quite a big change that I am trying to avoid in this work since it doesn't strictly need it.

~~Can you elaborate on the big change that you're trying to avoid?~~

~~Abstractly, the robot needs to do this:~~

1. ~~Statically validate the input files and parse them into something convenient, like `ast` objects or whatever.~~
2. ~~`eval()` that one way to harvest runtime parameter info.~~
3. ~~`eval()` that a different way to harvest command and equipment info.~~

~~I don't see why that's not easy to do inside `ProtocolAnalyzer`? It doesn't have to monitor for an intermediate state change, per se—it just has to call two different functions and store their results in sequence, right? Am I missing something?~~

[EDIT] This is better described in https://github.com/Opentrons/opentrons/pull/15229#discussion_r1619120583, and we discussed that on a call.

~~I ask because it would be good to avoid having to layer on additional `Manager`s, if we can.~~

[EDIT] Discussed on a call. `AnalysesManager` is going to be useful for cancellation, so it's not avoidable. I think, then, that we can merge `ProtocolAnalyzer`'s duties into `AnalysesManager`  and delete `ProtocolAnalyzer`.

SyntaxColoring: Separate question:

> We've talked about this in authorship and this flow with an intermediate analysis state that can change anytime is not useful and complicates a couple of things for the app so we want to avoid it.

Makes sense. But just checking: When you say this "complicates a couple things for the app," you don't happen to mean that the app will expect this RTP info to always be available for any uploaded protocol at any time, do you? Because I don't think that assumption will hold with this PR, if the robot is power-cycled mid-analysis.

robot-server/robot_server/protocols/router.py
SyntaxColoring: Discussed on a call:

We might want to try this in this PR depending on how far off `AnalysesManager` is, but yes, `AnalysesManager` would cover this too.

jbleon95 (approved):
Changes look good and make sense to me, and will make adding the "new file needed" analysis result much easier

SyntaxColoring (approved):
yay ty

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -21,57 +25,62 @@ class ProtocolAnalyzer:
     def __init__(
         self,
         analysis_store: AnalysisStore,
+        protocol_resource: ProtocolResource,
     ) -> None:
         """Initialize the analyzer and its dependencies."""
         self._analysis_store = analysis_store
+        self._protocol_resource = protocol_resource
+
+    async def load_runner(
```
sanni-t, line 34: from discussion w/ @SyntaxColoring: making runner loading as a part of the analyzer initialization will help prevent misuse of `load_runner()`, for example, calling `load_runner()` multiple times.

I like this suggestion. Will take it into account when making the upcoming changes to the analysis arch.

PR #15230, opened 2024-05-21
fix(protocol-designer): submit and don't show again fixes

jerader:
closes RQA-2715 RQA-2719

# Overview

Addresses 2 bugs:
1. fixes the logic for don't show again checkbox in the modals
2. fixes the update button for editing the metadata

# Test Plan

-Go to settings and click on the `Restore all hints and tips notifications` restore button (if you can't click it then its restored so no worries)
- create a flex or ot-2 protocol
- edit the metadata (author, name, etc) and see that the update button works as expected and disables when there are no new changes
- go to the deck map and see the hint modal for `Setting up your protocol`. Click on the don't show again checkbox and make sure it works as expected

# Changelog

- add a use callback for the don't show again checkbox logic
- add a manual `isDirty` state for keeping track of when the update button is disabled/not disabled 

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/FilePage.tsx
+++ b/protocol-designer/src/components/FilePage.tsx
@@ -83,7 +83,7 @@ export const FilePage = (): JSX.Element => {
   const saveFileMetadata = (nextFormValues: FileMetadataFields): void => {
     dispatch(actions.saveFileMetadata(nextFormValues))
   }
-
+  const [isManualDirty, setManualDirty] = React.useState<boolean>(false)
   const {
     handleSubmit,
     watch,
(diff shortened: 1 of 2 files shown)


ncdiehl11 (approved):
works as expected and code looks good!

protocol-designer/src/components/FilePage.tsx
```diff
@@ -155,6 +154,7 @@ export const FilePage = (): JSX.Element => {
                     name="protocolName"
                     value={protocolName}
                     onChange={field.onChange}
+                    onClick={() => setManualDirty(true)}
```
koji, line 157: update button is activated without any change.
I guess activating the button when a user changes the value might be better.

PR #15232, opened 2024-05-21
fix(app): instead filter magnetic blocks (not magnetic modules) out of the required modules

b-cooper:
# Overview

Fix bug where we were accidentally filtering magnetic modules instead of magnetic blocks from the
required modules list from analysis.

Closes [RQA-2754](https://opentrons.atlassian.net/browse/RQA-2754)

# Review requests

send a protocol that requires the magnetic block to a Flex. From the Protocol detail page of the ODD, there should only be one listing for that required magnetic block in the hardware list. There should also be no incorrect location conflicts.

# Risk assessment
low

[RQA-2754]: https://opentrons.atlassian.net/browse/RQA-2754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/pages/ProtocolDetails/Hardware.tsx
+++ b/app/src/pages/ProtocolDetails/Hardware.tsx
@@ -19,6 +19,8 @@ import {
   getModuleType,
   getFixtureDisplayName,
   GRIPPER_V1_2,
+  MAGNETIC_BLOCK_FIXTURES,
+  MAGNETIC_BLOCK_TYPE,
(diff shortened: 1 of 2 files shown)


CaseyBatten (approved):
Successfully removes redundant magnetic block in hardware listing while maintaining module graphic for the magnetic block.

PR #15236, opened 2024-05-21
fix(docker): need to specify platform

y3rsh:
### Forgot to specify the platform explicitly.  

> When it was built on a mac, we got the wrong platform.


koji (approved):
The changes make sense to me.
Thank you for fixing this!

PR #15237, opened 2024-05-21
fix(protocol-designer): set default z value for TipPositionModal

ncdiehl11:
Closes RQA-2750

# Overview

If selecting a custom offset for TipPositionModal on a transfer step, the Z field should populate with the default offset from bottom rather than empty.

# Test Plan

- Setup a PD protocol on Flex or OT-2
- Create a transfer step
- Select a custom tip position for aspirate/dispense
- Click 'custom'
- Verify that the z field is the expected default (1mm) rather than empty

# Changelog

- set initial z value in useState on TipPositionModal

# Risk assessment

low


koji (approved):
lgtm

PR #15238, opened 2024-05-21
fix(shared-data): raise uiMaxFlowRate values to match the default blowout and dispense

jerader:
closes RQA-2722

# Overview

A few of the `uiMaxFlowRate` values were not high enough to account for the default `blowout` and `dispense` flow rates. Rather than changing those flow rates, I raised the `uiMaxFlowRate`s for those pipette models. This required me to modify the test to expect the failures for those models. 

Those models include: ` "p50_single_v3.4",
                    "p50_single_v3.5",
                    "p50_single_v3.6",
                    "p50_multi_v3.5",
                    "p50_multi_v3.4"`

# Test Plan

verify that my test changes and comment make sense

# Changelog

- raised the ui max flow rate to match the default blowout and dispense flow rates
- modified the pytest to account for the failed asserts and left a comment explaining why we are expecting failed asserts

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/shared-data/js/__tests__/pipettes.test.ts
+++ b/shared-data/js/__tests__/pipettes.test.ts
@@ -158,7 +158,7 @@ describe('pipette data accessors', () => {
       minVolume: 5,
       supportedTips: {
         t50: {
-          uiMaxFlowRate: 47,
+          uiMaxFlowRate: 57,
           aspirate: {
             default: {
               1: expect.anything(),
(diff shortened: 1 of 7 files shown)


sfoster1 (commented):
sadly, yeah we have to. really at some point we're going to have to change the way we do the flow rate limits to not be based on a single value.

sfoster1 (approved):

PR #15239, opened 2024-05-21
fix(app): fix slot validation on ODD location conflict check to account for thermocycler occupying 2 slots

CaseyBatten:
# Overview

Slot conflict checking did not account for the fact that the thermocycler (and any future multi-slot fixture) was composed of multiple fixtures. This means that the expected fixture (thermocycler front) was not returned, because thermocycler rear occupies the first slot in the list.

# Test Plan
Upload a protocol requiring a thermocycler to a Flex, ensure that from the main protocol list view page and the protocol dashboard do not list a location conflict.

# Risk assessment
low


sfoster1 (approved):
Looks correct to me!

b-cooper (approved):
🏁

PR #15240, opened 2024-05-21
feat(app): "Cancel run" during Error Recovery

mjhuff:
Closes [EXEC-462](https://opentrons.atlassian.net/browse/EXEC-462)

# Overview

This PR adds the ability to cancel a run during Error Recovery. 

https://github.com/Opentrons/opentrons/assets/64858653/16cd4608-dedb-4cfb-a375-e67f233e1514

# Test Plan
- Follow the procedure outlined the video. As always, make sure both FFs are enabled.

# Changelog
- Added the ability to cancel a run during Error Recovery.

# Review requests
This will not implement designs, and that's intentional, since they're still somewhat up in the air. We'll eventually get around to it. This PR just adds the underlying logic and matches designs as reasonably as possible for the moment.

# Risk assessment
low

[EXEC-462]: https://opentrons.atlassian.net/browse/EXEC-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
```diff
@@ -80,7 +80,7 @@ function ErrorRecoveryComponent(props: RecoveryContentProps): JSX.Element {
       <ErrorRecoveryHeader errorKind={props.errorKind} />
       <ErrorRecoveryContent {...props} />
     </Flex>,
-    getModalPortalEl()
+    getTopPortalEl()
```
mjhuff, line 83: @sfoster1 Turns out this has to be `topPortalEl`, at least as things stand. If it's the modal equivalent there are a few issues:

1. The running protocol control buttons (pause and play) appear above error recovery. It looks great.
2. The run timer appears above error recovery.
3. The canceling run modal is appears above the ER modal while the run is cancelled.

All of these are solvable, but I think it points to us probably wanting to think about modals in general, as we discussed earlier.

sfoster1: oh gross. are those ones on the top portal?

the distinction between top and modal portals is a little weaker on the ODD I think.

sfoster1 (approved):
Looks really good!

app/src/organisms/ErrorRecoveryFlows/ErrorRecoveryWizard.tsx
mjhuff: Nope...random things have styled components with z-index values. Here's an example in `RunningProtocol`:

Screenshot 2024-05-22 at 8 52 42 AM

Maybe there is a reason for it, but IDK what that is.

PR #15241, opened 2024-05-21
chore(api): raise EnumeratedError for failed status check

sfoster1:
If you did a motor_engaged() query and the node didn't respond, you'd get a KeyError wrapped in a 4000 error, which isn't correct or useful. Make that a 1000 error instead.


SyntaxColoring (approved):
Nice.

vegano1 (approved):
Thank you!

PR #15242, opened 2024-05-21
feat(app,robot-server): Subscribe to `currentlyRecoveringFrom` notifications

SyntaxColoring:
# Overview

Closes EXEC-458.

# Test plan

1. Push robot-server and ODD to a Flex
2. Induce a `pickUpTip` failure
3. Make sure the ODD goes into recovery mode when the `pickUpTip` failure happens
4. Make sure both recovery options work

And maybe check the app or server logs to make sure it doesn't spam network requests, especially when notifications are unavailable.

# Changelog

* Generalize the `...current_command` MQTT topic, which was "the equivalent of `GET /runs/:runId/commands?cursor=null&pageLength=1`, to `...command_links`, which is the equivalent of the `links` field of any `GET /runs/:runId/commands` query.
* Update the frontend to take advantage of that. When it sees `links` update with a new `currentlyRecoveringFrom` command ID, it will do a separate `GET /runs/{id}/commands/{id}` query to fetch the full details of the command.

# Review requests

Several frontend questions because I don't know what I'm doing there. See my GitHub comments below.

# Risk assessment

Probably at least a medium risk of breaking recent unreleased work (#15170). Although we can cover these notifications hooks with tests, it doesn't look like those tests are in a good position to make sure the UI is actually doing the correct thing overall.


app/src/organisms/ErrorRecoveryFlows/utils.ts
```diff
@@ -21,46 +20,35 @@ import type {
...
+
+  // One-shot query for the full details of the currentlyRecoveringFrom command.
+  //
+  // TODO(mm, 2024-05-21): When the server supports fetching the
+  // currentlyRecoveringFrom command in one step, do that instead of this chained query.
+  const { data: commandQueryData } = useCommandQuery(
+    currentlyRecoveringFromLink?.meta.runId ?? null,
+    currentlyRecoveringFromLink?.meta.commandId ?? null,
+    { enabled: currentlyRecoveringFromLink != null, staleTime: Infinity }
+  )
```
SyntaxColoring: Is this the correct way to do a one-shot query chained behind an earlier query? With `enabled` and `staleTime`?

app/src/organisms/ErrorRecoveryFlows/utils.ts
```diff
@@ -21,46 +20,35 @@ import type {
...
-          command => command.status === 'failed' && command.intent !== 'fixit'
-        ) ?? null
-      setRecentFailedCommand(failedCommand)
-    } else if (!isRunStatusAwaitingRecovery && recentFailedCommand != null) {
-      setRecentFailedCommand(null)
-    }
-  }, [isRunStatusAwaitingRecovery, recentFailedCommand, allCommandsQueryData])
-
-  return recentFailedCommand
+  return commandQueryData?.data ?? null
```
SyntaxColoring: This was the case before, but I'm nervous about the way this code makes these two cases indistinguishable from each other:

1. there is no `currentlyRecoveringFrom` link
2. there was an error fetching the `currentlyRecoveringFrom` link

Should we do something about this, or is this standard practice and I'm worrying over nothing?

app/src/organisms/ErrorRecoveryFlows/__tests__/utils.test.ts
```diff
@@ -145,87 +142,45 @@ describe('useRouteUpdateActions', () => {
...
+    const { result } = renderHook(() => useCurrentlyRecoveringFrom(MOCK_RUN_ID))
 
-    expect(result.current).toEqual({
-      status: 'failed',
-      intent: 'protocol',
-      id: '123',
-    })
+    expect(result.current).toStrictEqual(null)
+    // TODO(mm, 2024-05-22): Figure out how to assert that useCommandQuery was
+    // called with `enabled: false`.
```
SyntaxColoring: I tried asserting the result of `vi.mocked(useCommandQuery).calls`, based on https://vitest.dev/api/mock.html#mock-calls, but TS complains that `.calls` does not exist on that object, which I don't understand because it's definitely a mock and TS definitely knows that. Am I doing something fundamentally wrong?

app/src/resources/runs/useNotifyAllCommandsQuery.ts
```diff
@@ -11,8 +11,14 @@ export function useNotifyAllCommandsQuery<TError = Error>(
   params?: GetCommandsParams | null,
   options: QueryOptionsWithPolling<CommandsData, TError> = {}
 ): UseQueryResult<CommandsData, TError> {
+  // Assume the useAllCommandsQuery() response can only change when the command links change.
+  //
+  // TODO(mm, 2024-05-21): I don't think this is correct. If a command goes from
+  // running to succeeded, that may change the useAllCommandsQuery response, but it
+  // will not necessarily change the command links. We might need an MQTT topic
+  // covering "any change in `GET /runs/{id}/commands`".
```
SyntaxColoring, line 19: Does this make sense?

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
```diff
@@ -40,21 +41,23 @@ def __init__(
...
+        publisher_notifier.register_publish_callbacks(
+            [
+                self._handle_current_command_change,
+                self._handle_recovery_target_command_change,
+                self._handle_engine_status_change,
+            ]
         )
 
-    async def initialize(
+    async def start_publishing_for_run(
```
mjhuff, line 56: Nice, I do like this better.

robot-server/robot_server/service/notifications/publishers/runs_publisher.py
mjhuff, line 1: Thank you for the cleanup here!

app/src/organisms/ErrorRecoveryFlows/utils.ts
```diff
@@ -21,46 +19,34 @@ import type {
...
-  // for a newly failed command when the run first enters "awaiting-recovery."
-  const isRunStatusAwaitingRecovery = runStatus === RUN_STATUS_AWAITING_RECOVERY
-
   const { data: allCommandsQueryData } = useNotifyAllCommandsQuery(
     runId,
-    null,
+    { cursor: null, pageLength: 0 }, // pageLength 0 because we only care about the links.
     {
-      enabled: isRunStatusAwaitingRecovery && recentFailedCommand == null,
       refetchInterval: ALL_COMMANDS_POLL_MS,
```
mjhuff, line 41: Discussed in person - we do want to keep the enabled condition here.

app/src/organisms/ErrorRecoveryFlows/utils.ts
mjhuff: Discussed in person, we do not need `staleTime`.

app/src/organisms/ErrorRecoveryFlows/utils.ts
mjhuff: Discussed in person, no we're ok. Most likely React Query or Axios (IDK which) is solving this for us

mjhuff (commented):
Sorry, I forgot to submit these comments!

This looks great overall. Just a few things to fix.

app/src/organisms/ErrorRecoveryFlows/__tests__/utils.test.ts
mjhuff: Also discussed in person, but it sounds like there was just a typo in your approach.

mjhuff (approved):
Nice job, thank you for implementing this!

SyntaxColoring:
I'll probably test this on Monday and merge it then, assuming all goes well.

PR #15244, opened 2024-05-22
refactor(protocol-designer): dismissed warning now dismisses for duration of protocol

jerader:
closes RQA-2756

# Overview

The current behavior for a dismissed warnings for both form warnings and timeline warnings is that it is tied to the step id. This means that if you dismiss a warning then duplicate the step or create a new step, the warning appears again for that step. This has been the legacy behavior for dismissed warnings.

This new behavior is instead of tying each dismissed warning to a specific step id, it is tied to the protocol itself. So if you dismiss a warning, it is dismissed forever in the protocol, even if you reimport your protocol. But if you make a new protocol, the warnings are all visible again.

# Test Plan

Create a flex or OT-2 protocol. Do not add any liquids but add a tuberack to the deck. Create a transfer step where you are aspirating and/or dispensing into the tube. See that the tuberack warning is there. DIsmiss it.

Next, duplicate that step by right clicking on the step in the timeline and selecting the duplicate button. See that the only warnings on both steps is the "no liquid" warning. But the tuberack warning should not be visible. Then, delete the "no liquid" warning and it should delete for both steps.

Create one final transfer step with the tuberack and see that both the tuberack and "no liquid" warnings are not visible.

Finally, export your protocol and reimport. go to the deckmap and timeline and see that there are no warnings.

# Changelog

- refactored the dismissed warnings reducer and selector to remove the record, keyed by the step id. 
- refactor the types to match the new shape
- refactor affected components
- fix all affected tests
- add to the migration
- fix cypress tests

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/fixtures/protocol/8/doItAllV3MigratedToV8.json
+++ b/protocol-designer/fixtures/protocol/8/doItAllV3MigratedToV8.json
@@ -27,7 +27,7 @@
           "opentrons/opentrons_96_tiprack_300ul/1"
         ]
       },
-      "dismissedWarnings": { "form": {}, "timeline": {} },
+      "dismissedWarnings": { "form": [], "timeline": [] },
       "ingredients": {
         "0": {
           "name": "Water",
--- a/protocol-designer/fixtures/protocol/8/doItAllV4MigratedToV8.json
+++ b/protocol-designer/fixtures/protocol/8/doItAllV4MigratedToV8.json
@@ -27,7 +27,7 @@
           "opentrons/opentrons_96_tiprack_300ul/1"
         ]
       },
-      "dismissedWarnings": { "form": {}, "timeline": {} },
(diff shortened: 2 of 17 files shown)


koji:
@jerader 

when import the protocol that followed test plan, PD shows the following error. Is this an expected behavior?

2024-05-22 09_54_16-Opentrons Protocol Designer

koji (approved):
leave one question.
The changes are consistent with what you’re referring to.

jerader:
> @jerader
> 
> when import the protocol that followed test plan, PD shows the following error. Is this an expected behavior?
> 
> 2024-05-22 09_54_16-Opentrons Protocol Designer

@koji oof thanks for catching this! looks like a bug, I'll investigate.

jerader:
> @koji oof thanks for catching this! looks like a bug, I'll investigate.

@koji oh its the import bug that QA documented 👍

PR #15246, opened 2024-05-22
fix(api): monitor the pressure sensor more

ryanthecoder:
# Overview
This will allow us to monitor the pressure sensor more, specifically it should help us catch and differentiate cases where the tip collides with the labware by monitoring it during the z movement in addition to the plunger movement. 
Should help diagnose:
RABR-319
RABR-305
RABR-304

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/backends/ot3controller.py
+++ b/api/src/opentrons/hardware_control/backends/ot3controller.py
@@ -45,7 +45,7 @@ from .ot3utils import (
     motor_nodes,
     LIMIT_SWITCH_OVERTRAVEL_DISTANCE,
     map_pipette_type_to_sensor_id,
-    moving_axes_in_move_group,
+    moving_pipettes_in_move_group,
     gripper_jaw_state_from_fw,
     get_system_constraints,
     get_system_constraints_for_calibration,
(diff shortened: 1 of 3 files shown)


api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -667,7 +667,8 @@ async def move(
         mounts_moving = [
             k
             for k in moving_axes_in_move_group(move_group)
-            if k in [NodeId.pipette_left, NodeId.pipette_right]
+            if k
+            in [NodeId.pipette_left, NodeId.pipette_right, NodeId.head_l, NodeId.head_r]
         ]
         async with self._monitor_overpressure(mounts_moving):
```
sfoster1: i don't think this can handle mount axis nodes

sfoster1 (changes requested):

api/src/opentrons/hardware_control/backends/ot3controller.py
ryanthecoder: Yup, added a new util function and fixed that

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -664,11 +664,8 @@ async def move(
             else False,
         )
 
-        mounts_moving = [
```
caila-marashaj, line 670: nitpick here, but we can maybe rename this variable now to `pipettes_to_monitor` or something to that effect

caila-marashaj (approved):
nice. I like this.

sfoster1 (approved):
Okay, but let's add tests at least for `moving_pipettes_in_move_group` please

PR #15247, opened 2024-05-22
refactor(api): conditionally import usb dep

sfoster1:
I think the intent was always that AsyncByonoy and thus pyusb would only be imported when actually building an AsyncByonoy instance, but at some point it got imported globally for a typecheck. Make that typecheck typecheck-only with a different name, and the problem goes away.
 Other side of #15206


SyntaxColoring (approved):
Makes sense. Might be worth a comment explaining why the import needs to be conditional.

ahiuchingau (approved):
Thank you

PR #15251, opened 2024-05-23
Add name field to pipette v2 general specs

smb2268:
# Overview
We need a to add `name` to v2 of pipette specs under general since it isn't otherwise represented in this version of the schema. The `name` for each pipette is pulled from the key of `pipetteNameSpecs` and the `name` field of `pipetteModelSpecs`

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/shared-data/js/__tests__/__snapshots__/pipettes.test.ts.snap
+++ b/shared-data/js/__tests__/__snapshots__/pipettes.test.ts.snap
@@ -5967,3 +5967,57 @@ exports[`pipette data accessors > getPipetteNameSpecs > name p1000_single snapsh
   },
 }
 `;
+
+exports[`pipette data accessors > getPipetteNameSpecs > name p1000_single_flex snapshot 1`] = `
+{
+  "channels": 1,
+  "defaultAspirateFlowRate": {
+    "max": 812,
+    "min": 3,
+    "value": 478,
+    "valuesByApiLevel": {
+      "2.0": 137.35,
+      "2.14": 478,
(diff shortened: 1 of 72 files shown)


jerader (commented):
nice, lgtm! Might be good to add a quick pytest for this though - maybe similarly to `test_model_enum` in `shared-data/python/tests/pipette/test_pipette_definition.py` but instead of testing a model, it tests that each name equals a `PipetteNameType` (just a suggestion, i'd defer to @sfoster1's recommendations on test coverage 😄)

sfoster1:
> nice, lgtm! Might be good to add a quick pytest for this though - maybe similarly to `test_model_enum` in `shared-data/python/tests/pipette/test_pipette_definition.py` but instead of testing a model, it tests that each name equals a `PipetteNameType` (just a suggestion, i'd defer to @sfoster1's recommendations on test coverage 😄)

Yeah, I agree

sfoster1 (commented):
I agree with @jerader 's point about adding a test, but looking at this in practice I'm a little worried that we're still implicitly relying on data from v1 of the pipette definitions - specifically the `PipetteName` type. 

Can tsc load type data from jsonschema? If so, maybe the right thing to do here is actually for this to be an enum in the schema so tsc can pick it up there. That ability would be enough for me to want to encode it in the schema instead of implicitly, even though I'd previously made the point that we should do that.

y3rsh:
I think the plan for this is a retarget at edge, it will not be in 8.0.0

PR #15255, opened 2024-05-23
docs(api): fix type specification for `add_parameters()`

ecormany:
# Overview

I got the type specification for the object of `add_parameters()` wrong. 

# Test Plan

Simulated a protocol that uses the `def add_parameters(…` line verbatim on 7.3.0-b1.

# Changelog

1-line diff

# Risk assessment

none


SyntaxColoring (approved):
TY!

PR #15256, opened 2024-05-23
fix(react-api-client): filter out unknown modules in the /modules endpoint.

vegano1:
# Overview

The desktop app is blocked when it gets an unknown module in the `GET /modules` response, which happens when the user clicks on the `Devices` side panel in the desktop app. Since we query modules for all discovered robots, it can be confusing to know which robot the unknown module is connected to. There will sometimes be a lag between developing and supporting new modules in the desktop app. So instead of blocking the desktop app, we should filter out unknown modules in the `GET /response`.

Closes: [RABR-316](https://opentrons.atlassian.net/browse/RABR-316)

# Test Plan

- [x] Connect an unknown module (plate reader) to a robot and make sure we no longer see the "invalid module" modal.
- [x] Make sure we still handle known modules like before.

# Changelog

- Use the `select` option of react `UseQuery` to filter out unknown modules

# Review requests
# Risk assessment
low

[RABR-316]: https://opentrons.atlassian.net/browse/RABR-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react-api-client/src/modules/useModulesQuery.ts
+++ b/react-api-client/src/modules/useModulesQuery.ts
@@ -3,6 +3,7 @@ import { getModules } from '@opentrons/api-client'
 import { useHost } from '../api'
 import type { UseQueryResult, UseQueryOptions } from 'react-query'
 import type { HostConfig, Modules } from '@opentrons/api-client'
+import { MODULE_MODELS } from '@opentrons/shared-data'
 
 export type UseModulesQueryOptions = UseQueryOptions<Modules>
 
@@ -26,7 +27,20 @@ export function useModulesQuery(
           }
         }
       }),
-    { enabled: host !== null, ...options }
+    {
+      enabled: host !== null,
+      // Filter unknown modules so we don't block the app, which
+      // can happen when developing new devices not yet known to the client.
+      select: resp => {
+        return {
+          ...resp,
+          data: resp.data.filter(module =>
+            MODULE_MODELS.includes(module.moduleModel)
+          ),
+        }
+      },
+      ...options,
+    }
   )
 
   return query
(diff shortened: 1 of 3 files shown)


react-api-client/src/modules/useModulesQuery.ts
```diff
@@ -26,7 +27,18 @@ export function useModulesQuery(
...
       }),
-    { enabled: host !== null, ...options }
+    {
+      enabled: host !== null,
+      select: resp => {
+        return {
+          ...resp,
+          data: resp.data.filter(module => {
+            MODULE_MODELS.includes(module.moduleModel)
+          }),
```
brenthagen: filter callback should return a boolean, so we want to remove the braces here:
```
          data: resp.data.filter(module => 
            MODULE_MODELS.includes(module.moduleModel)
          ),
```

react-api-client/src/modules/useModulesQuery.ts
```diff
@@ -26,7 +27,18 @@ export function useModulesQuery(
           }
         }
       }),
-    { enabled: host !== null, ...options }
+    {
+      enabled: host !== null,
+      select: resp => {
```
brenthagen, line 34: probably worth dropping a short comment explaining why we are doing this

brenthagen (approved):
LGTM, guards us a bit against modules outside the type system

react-api-client/src/modules/useModulesQuery.ts
vegano1: good call.

sfoster1 (approved):
nice!

PR #15257, opened 2024-05-23
refactor(api, robot-server): redirect router level calls to PE and protocol_runners via run orchestrator

TamarZanzouri:
# Overview

second phase of run orchestrator. redirect all router level calls via `run_orchestrator`.

# Test Plan

- All e2e tests are passing.
- through it on a robot and make sure nothing is broken. -> tests this with dev server. 
  - [x] test json protocols
  - [x] test python protocols 
  - [x] post commands to a run
  - [x] post commands with default router

# Changelog

- added method calls from `run_data_manager` and routers to `engine_store`.
- all calls get through `run_orchestrator` to PE and to runners. 

# Review requests

changes make sense? will something break? do we like this change? should I add more tests?

# Risk assessment

Medium. tested with dev server and all tests are passing but still a fundamental change.

### DIFF
--- a/api/src/opentrons/protocol_runner/run_orchestrator.py
+++ b/api/src/opentrons/protocol_runner/run_orchestrator.py
@@ -102,7 +102,7 @@ class RunOrchestrator:
             protocol_engine=protocol_engine,
         )
 
-    def add_command_and_wait_for_interval(self, command: CommandCreate, wait_until_complete: bool = False,
+    async def add_command_and_wait_for_interval(self, command: CommandCreate, wait_until_complete: bool = False,
                                           timeout: Optional[int] = None, failed_command_id: Optional[str] = None) -> Command:
         added_command = self._protocol_engine.add_command(request=command, failed_command_id=failed_command_id)
         if wait_until_complete:
--- a/api/src/opentrons/protocol_runner/run_orchestrator.py
+++ b/api/src/opentrons/protocol_runner/run_orchestrator.py
@@ -7,7 +7,7 @@ from anyio import move_on_after
 from . import protocol_runner, AnyRunner
 from ..hardware_control import HardwareControlAPI
 from ..protocol_engine import ProtocolEngine, CommandCreate, Command
-from ..protocol_engine.types import PostRunHardwareState
+from ..protocol_engine.types import PostRunHardwareState, EngineStatus
 from ..protocol_reader import JsonProtocolConfig, PythonProtocolConfig
 
 
@@ -22,6 +22,7 @@ class RunOrchestrator:
     ]
     _setup_runner: protocol_runner.LiveRunner
     _fixit_runner: protocol_runner.LiveRunner
+    _protocol_live_runner: protocol_runner.LiveRunner
     _hardware_api: HardwareControlAPI
     _protocol_engine: ProtocolEngine
 
@@ -53,15 +54,15 @@ class RunOrchestrator:
         self._setup_runner = setup_runner
         self._fixit_runner = fixit_runner
 
-    @property
-    def engine(self) -> ProtocolEngine:
-        """Get the "current" persisted ProtocolEngine."""
-        return self._protocol_engine
-
-    @property
-    def runner(self) -> AnyRunner:
-        """Get the "current" persisted ProtocolRunner."""
-        return self._protocol_runner or self._setup_runner
+    # @property
+    # def engine(self) -> ProtocolEngine:
+    #     """Get the "current" persisted ProtocolEngine."""
+    #     return self._protocol_engine
+    #
+    # @property
+    # def runner(self) -> AnyRunner:
+    #     """Get the "current" persisted ProtocolRunner."""
+    #     return self._protocol_runner or self._setup_runner
 
     @classmethod
     def build_orchestrator(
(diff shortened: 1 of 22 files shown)


api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -41,25 +68,24 @@ def __init__(
...
+    def run_id(self) -> str:
         """Get the "current" persisted ProtocolEngine."""
-        return self._protocol_engine
-
-    @property
-    def runner(self) -> AnyRunner:
-        """Get the "current" persisted ProtocolRunner."""
-        return self._protocol_runner or self._setup_runner
+        if not self._run_id:
+            raise NotImplementedError("default orchestrator.")
```
TamarZanzouri: change this to a designated error

sfoster1: also what about it is not implemented? having a default orchestrator?

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -98,4 +128,196 @@ def build_orchestrator(
...
             protocol_engine=protocol_engine,
+            protocol_live_runner=protocol_live_runner,
+        )
+
+    def play(self, deck_configuration: Optional[DeckConfigurationType] = None) -> None:
+        """Start or resume the run."""
+        self._protocol_engine.play(deck_configuration=deck_configuration)
+
+    async def run(self, deck_configuration: DeckConfigurationType) -> RunResult:
+        """Start or resume the run."""
```
sfoster1: Fix up copy/pasted docstrings

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -98,4 +128,196 @@ def build_orchestrator(
...
+
+    def get_command_slice(
+        self,
+        cursor: Optional[int],
+        length: int,
+    ) -> CommandSlice:
+        """Get a slice of run commands.
+
+        Args:
+            run_id: ID of the run.
```
sfoster1: fix up docstring, doesn't take a run id

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -98,4 +128,196 @@ def build_orchestrator(
...
             hardware_api=hardware_api,
             protocol_engine=protocol_engine,
+            protocol_live_runner=protocol_live_runner,
+        )
+
+    def play(self, deck_configuration: Optional[DeckConfigurationType] = None) -> None:
+        """Start or resume the run."""
+        self._protocol_engine.play(deck_configuration=deck_configuration)
+
+    async def run(self, deck_configuration: DeckConfigurationType) -> RunResult:
```
TamarZanzouri, line 148: add test if I haven't already

robot-server/robot_server/runs/engine_store.py
```diff
@@ -235,34 +235,35 @@ async def create(
...
                 python_parse_mode=PythonParseMode.ALLOW_LEGACY_METADATA_AND_REQUIREMENTS,
                 run_time_param_values=run_time_param_values,
             )
-        elif isinstance(self.runner, JsonRunner):
+        elif isinstance(runner, JsonRunner):
             assert (
                 protocol is not None
-            ), "A JSON protocol should have a protocol source file."
-            await self.runner.load(protocol.source)
+            ), "A JSON protocol shouZld have a protocol source file."
```
sfoster1:
```suggestion
            ), "A JSON protocol should have a protocol source file."
```

robot-server/robot_server/runs/engine_store.py
```diff
@@ -271,23 +272,121 @@ async def clear(self) -> RunResult:
...
         return RunResult(
             state_summary=run_data, commands=commands, parameters=run_time_parameters
         )
+
+    def play(self, deck_configuration: Optional[DeckConfigurationType] = None) -> None:
+        """Start or resume the run."""
+        self.run_orchestrator.play(deck_configuration=deck_configuration)
+
+    async def run(self, deck_configuration: DeckConfigurationType) -> RunResult:
+        """Start or resume the run."""
```
sfoster1: docstrings

sfoster1 (changes requested):
This looks really good, thanks for taking it on! Let's fix up some docstring copypasta and test it, but the structure looks great!

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -98,4 +134,192 @@ def build_orchestrator(
...
+
+    def get_is_okay_to_clear(self) -> bool:
+        """Get whether the engine is stopped or sitting idly so it could be removed."""
+        return self._protocol_engine.state_view.commands.get_is_okay_to_clear()
+
+    def prepare(self) -> None:
+        """Prepare live runner for a run."""
+        self._setup_runner.prepare()
+        self._fixit_runner.prepare()
+        self._protocol_live_runner.prepare()
```
SyntaxColoring: It looks a little questionable that `self._setup_runner.prepare()` and `self._fixit_runner.prepare()` *are* called for live HTTP runs (because that's the only thing calling `RunOrchestrator.prepare()`), and are *not* called for JSON or Python protocols. In all cases, the setup runner and fixit runner behave the same, right? So shouldn't they consistently always be `.prepare()`'d or never be `.prepare()`'d?

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -1,12 +1,43 @@
...
+
+class NoProtocolRunAvailable(RuntimeError):
+    """An error raised if there is no protocol run available."""
+
+
+class RunNotFound(GeneralError):
+    """An error raised if there is no run associated."""
 
 
 class RunOrchestrator:
```
SyntaxColoring, line 44: This does not yet intend to close EXEC-418, right? It's just a step towards it? Since that ticket says we want to move the command queues to the runner, and it doesn't look like that change has been made yet.

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -98,4 +134,192 @@ def build_orchestrator(
...
+            )
+        elif self._protocol_live_runner:
+            return await self._protocol_live_runner.run(
+                deck_configuration=deck_configuration
+            )
+        else:
+            return await self._setup_runner.run(deck_configuration=deck_configuration)
+
+    def pause(self) -> None:
+        """Start or resume the run."""
```
SyntaxColoring: Docstring.

api/src/opentrons/protocol_runner/run_orchestrator.py
```diff
@@ -98,4 +134,192 @@ def build_orchestrator(
...
+        self._protocol_engine.request_pause()
+
+    async def stop(self) -> None:
+        """Start or resume the run."""
```
SyntaxColoring: Docstring.

robot-server/tests/runs/router/test_commands_router.py
```diff
@@ -22,78 +21,77 @@
...
-    with pytest.raises(ApiError) as exc_info:
-        await get_current_run_engine_from_url(
-            runId="run-id",
-            engine_store=mock_engine_store,
-            run_store=mock_run_store,
-        )
-
-    assert exc_info.value.status_code == 409
-    assert exc_info.value.content["errors"][0]["id"] == "RunStopped"
+# async def test_get_current_run_engine_from_url(
```
SyntaxColoring: Should these commented-out tests be removed or uncommented?

robot-server/tests/runs/test_engine_store.py
```diff
@@ -215,48 +185,45 @@ async def test_clear_idle_engine(subject: EngineStore) -> None:
...
-        deck_type=deck_type,
-    )
-
-    result = await subject.get_default_engine()
-
-    assert result.state_view.config.robot_type == robot_type
-
-
-async def test_get_default_engine_current_unstarted(subject: EngineStore) -> None:
+# @pytest.mark.parametrize("robot_type", ["OT-2 Standard", "OT-3 Standard"])
```
SyntaxColoring: Commented-out tests

SyntaxColoring (commented):
Thank you! This mostly makes sense to me so far and I think we should keep going. A few questions:

SyntaxColoring:
Oh and this should probably be `refactor` instead of `chore` since it's meaningfully affecting production code.

api/src/opentrons/protocol_runner/run_orchestrator.py
TamarZanzouri: correct. this is part of the change getting there. I will update the ticket. actually I should have split it but might be too late?

sfoster1 (approved):
I agree with @SyntaxColoring about the prepare thing, we should have an explicit reason why we're not doing it, but other than that looks good!

api/src/opentrons/protocol_runner/run_orchestrator.py
TamarZanzouri: I agree. I will fix it.

robot-server/tests/runs/test_engine_store.py
SyntaxColoring: These tests are still commented out. Other than that, LGTM. Thanks!

SyntaxColoring (approved):

PR #15258, opened 2024-05-23
fix(protocol-designer): candidate-C bug addressing and don't unnecess…

jerader:
…arily migrate

closes RQA-2772 RQA-2773 RQA-2776 RQA-2777

# Overview

Addresses several things:
1. no longer unnecessarily migrate protocols when the `candidate-` string is at the end of the version
2. correctly generate a waste chute when only a move labware step is involved and not a pipetting step
3. don't create a pipette entity when the pipette name is an empty string
4. fix the cta for changing the protocol name

# Test Plan

First, for fix 3: create a flex protocol and add a pipette, add a tiprack, add the 2nd pipette then add a tiprack, then go back and select "none" for the 2nd pipette. create the protocol and see that it does no white screen

Next, edit the name of the protocol and see that it updates in the banner when you hit submit (for fix 4).

Then for fix 2, add a waste chute and a gripper and create a move labware step to move the tiprack into the waste chute with the gripper. Export the protocol and see that the "unused waste chute" modal DOES NOT show up

Import the protocol back and there SHOULD NOT be a migration modal that pops up (for fix 1). see that the waste chute is still selected in the protocol.

# Changelog

- check for move labware into the waste chute and add the fix to the `getUnusedTrash` util and the reducer
- add an extra protection when generating pipette entities
- add some logic to not migrate if the version has the main version string in it
- change the onClick to happen inside the `saveFileMetadata`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/FilePage.tsx
+++ b/protocol-designer/src/components/FilePage.tsx
@@ -115,7 +115,6 @@ export const FilePage = (): JSX.Element => {
     'author',
     'description',
   ])
-  console.log('!isDirty || !isManualDirty', !isDirty || !isManualDirty)
   return (
     <div className={styles.file_page}>
       <Card title={t('application:information')}>
--- a/protocol-designer/src/components/FilePage.tsx
+++ b/protocol-designer/src/components/FilePage.tsx
@@ -82,6 +82,7 @@ export const FilePage = (): JSX.Element => {
 
   const saveFileMetadata = (nextFormValues: FileMetadataFields): void => {
     dispatch(actions.saveFileMetadata(nextFormValues))
+    setManualDirty(false)
   }
   const [isManualDirty, setManualDirty] = React.useState<boolean>(false)
   const {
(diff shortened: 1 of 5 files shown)


koji:
this is kind of an edge case.

The form doesn't allow to change protocol name from something to `test test` because `test test` is saved data (browser's side) 

https://github.com/Opentrons/opentrons/assets/474225/3c1bed1f-264f-4c43-8214-11f9dc9d8797

koji (approved):
however, seems that the form part would need some changes. please check my comment.

jerader:
> this is kind of an edge case.
> 
> The form doesn't allow to change protocol name from something to `test test` because `test test` is saved data (browser's side)
> 
>  Screen.Recording.2024-05-23.at.5.08.23.PM.mov

@koji thank you for finding this, i will keep this in mind. for some reason i'm unable to replicate it on this branch? but i was able to replicate it in PD in production 🤔 Since its inconsistent i'm not sure how to fix it but i'll note this in the backlog since as you said it is an edge case. Thank you!

PR #15260, opened 2024-05-24
fix(protocol-designer): well order input field cta logic refinement

jerader:
closes RQA-2779

# Overview

This bug exists in production currently and is a bug as a result of when i refactored out the class components back in january. But basically the well order modal was not working properly when selecting different dropdown options for the 1st and 2nd fields

# Test Plan

Create an ot-2 or flex protocol and add a transfer step. Open advanced settings and click on the well order field. Open the modal and play around with changing the 2 settings. Make sure they populate as expected.

# Changelog

- fix the logic for selecting the first and second field and clean up the component

# Review requests

see test plan

# Risk assessment

low


shlokamin (approved):
thanks! changed around some of the fields and downloaded a protocol. when inspecting it looks like the well order is correct

PR #15262, opened 2024-05-24
refactor(app, components): Update "run paused" splash

mjhuff:
Closes [EXEC-398](https://opentrons.atlassian.net/browse/EXEC-398) and [EXEC-447](https://opentrons.atlassian.net/browse/EXEC-447)

# Overview

https://github.com/Opentrons/opentrons/assets/64858653/dac0f27a-4cda-4f04-acb1-d150f16c2231

NOTE: There is a ton of noise in this PR. See full details below. The main changes are in only a few places:

1. `components/src/icons/icon-data.ts`. This is the cool recovery icon!
2. `app/src/atoms/buttons/LargeButton.tsx`
3. `ErrorRecoveryFlows/index.tsx`
4. `ErrorRecoveryFlows/RunPausedSplash.tsx`
5. `useRouteUpdateActions` in `utils.ts`.
6. Designs got rid of the "Go back" button in `SelectRecoveryOption.tsx`.

### Run Paused Refactors

Refactors the Splash page, changing functionality and aligning it more with current Hi-Fi designs. The refactored splash page now [utilizes part of the error recovery wizard flow, specifically the cancel flow](https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=1212-154340&t=Ppidm2FwooTwhusq-4), which creates an interesting scenario: how do we properly refactor the splash page to best utilize error recovery wizard flows? Should it be part of error recovery, or should it be an entirely separate entity? Should it be a route in error recovery wizard flows?

There are a few things to think about that should inform this decision:
- Confirmed with Sue, we want the ER wizard to be a modal _above_ the splash screen. We need to show render the splash separate from the wizard.
- The desktop app won't have a splash screen. If we make this a wizard route and eventually decide to keep desktop and ODD wizard flows in sync, special-casing this will not be pretty.
-  We still need some way to hook into the cancel flow of ER, but pressing "Go back" enough should land us on the splash screen. 
- Is there actually any gain by making the splash screen a route? I don't think so. In fact, if we think of all ER routing as a graph and the `SelectRecoveryOptions` screen as the central node, we now either add a new central node or we have the `Cancel Run` route with two series of steps. I think the second option is reasonable, granted, but if cancel run flows route into something like a drop tip flow (which it will in the future), pressing "go back" and ending up where we expect, the splash screen, becomes more challenging.

I think the best approach is to consider the splash screen as a part of the error recovery flow, but NOT part of the error recovery **wizard** flows. This gets us:
- Render the wizard flows above the splash screen.
- Not have to handle the splash screen as a route.
- Keep "I have entered the wizard officially" behavior separate from the "I have not entered the wizard but am using a couple of the flows" behavior separate. Ex, when we launch error recovery, we z-home the pipettes. We don't do this when we press cancel run.
- Any future changes we make to the cancel flow in the ER wizard are applied to the splash page cancel flow for free.

The downside is we do internalize the rendering of the splash screen, and the public API no longer can control it. That being said, it's not a hard refactor to make if we do end up launching ER from other places in the desktop app/ODD to make this more manageable.

There's a lot of noise in this PR (again), as some design stuff needed to change, because copy no longer reflected the behavior of what actually happens (ex, clicking resume actually retries the failed command). It's still a mess, but I'd rather do as much of that work in a separate PR as possible and do the bare minimum here. I updated some typing as well, since it was out of date. Some fixtures were added as well, because they were sorely needed.

### Large Button

Design still sees [these new buttons](https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=1374-161944&t=Ppidm2FwooTwhusq-4) as `LargeButton`, so that means we should, too. The component didn't expose a way to adjust the color of icons (or anything about icons) for that matter, so I added just a mechanism for adjusting the color. I'm sure this is worth revisiting again, but I intentionally did not create a prop for adjusting `Icon` css, because SVG CSS styling doesn't work like component styling in most cases (try setting a `color` or `background-color` on an svg - it doesn't appear as you might expect!).

### Misc things

I've left a few explanatory comments below.

# Test Plan
- Enable both FE/BE FFs.
- Follow the flow in the video.

# Review requests
- Am I overlooking a very easy out that fulfills the "this gets us" bullet points section above? 

# Risk assessment
low - I type a lot but this matters pretty little. 

[EXEC-398]: https://opentrons.atlassian.net/browse/EXEC-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EXEC-447]: https://opentrons.atlassian.net/browse/EXEC-447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api-client/src/runs/commands/types.ts
mjhuff, line 1: Yeah this appears just to be a mistake. We have `commands` and `command`, and this was a copy past of the actual interface everything uses. It was only used in `DropTipWizard`, so I changed the import path and deleted this one.

app/src/pages/RunningProtocol/index.tsx
mjhuff, line 1: The diff is confusing. I just moved out the `RunPausedSplash`.

shared-data/command/types/index.ts
mjhuff, line 1: Updated the type here for ER reasons.

mjhuff (commented):
Some explanatory comments that didn't fit well above.

mjhuff:
Also worth mentioning - currently there is no pressed state on the splash screen buttons in design, and the subtext is very very likely to change (the stuff that says placeholder). Other than that, it's open for DQA!

app/src/atoms/buttons/LargeButton.tsx
```diff
@@ -22,6 +22,7 @@ interface LargeButtonProps extends StyleProps {
   buttonType?: LargeButtonTypes
```
sfoster1, line 22: the way i read the design spec, these are new fundamental button kinds. I see the big css overrides at the point of use - I wonder if what we want here is to add `'on-color'` and `'alert-on-color'` as `LargeButtonTypes` so we can at least get those big css overrides in here instead of at the point of use

app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
```diff
@@ -0,0 +1,177 @@
...
+      proceedToRoute(RECOVERY_MAP.CANCEL_RUN.ROUTE)
+    )
+  }
+
+  const onLaunchERClick = (): Promise<void> => toggleERWiz(true)
+
+  // TODO(jh 05-22-24): The hardcoded Z-indexing is non-ideal but must be done to keep the splash page above
+  // several components in the RunningProtocol page. Investigate why these components have seemingly arbitrary zIndex values
+  // and devise a better solution to layering modals.
+  return (
```
sfoster1, line 62: Most of this component is styling, and logic and routing is passed in props. That makes me feel like this should be implemented as a molecule or even atom

sfoster1 (approved):
The functionality and the styling looks good, but I think we should either now or in an immediate followup move the significant button style css overrides into the component as a new button type and consider moving the splash to its own atom or molecule.

I agree with the comment in the description about not making the splash a page. I think we'll end up being really annoyed at ourselves either way whenever we make this sync with the desktop app, so keeping the flexibility now seems ideal.

app/src/atoms/buttons/LargeButton.tsx
mjhuff: Doh you're right. After slack convo, I'll make these button types in a separate PR just so I can more easily tag in the other FE devs to review it.

app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
mjhuff: Yeah, good point. I think a splash page component would be quite useful and could easily unify all our splash pages. 

Sustaining ticket: 
[EXEC-469](https://opentrons.atlassian.net/browse/EXEC-469)

app/src/atoms/buttons/LargeButton.tsx
mjhuff: Ticket: [EXEC-470](https://opentrons.atlassian.net/browse/EXEC-470)

PR #15264, opened 2024-05-24
refactor(app): isNotifyEnabled -> shouldRefetch

mjhuff:
# Overview

`isNotifyEnabled` does not accurately represent what `useNotifyService` is doing, which is determining whether or not the HTTP request should be executed given notifications, polling status, etc. Notifications may be disabled, but this value will return 'always' - that's a bit confusing. "shouldRefetch" is a better name here.

# Review requests

If someone has a better name for `useNotifyService`, I'm open to ideas, too!

# Risk assessment
low


sfoster1 (approved):
Looks good to me. Something like `useNotifyDataReady` seems like a pretty good hook name.

PR #15267, opened 2024-05-28
feat(scripts): migrate AWS-SDK from v2 to v3

jerader:
closes [RDEVOPS-96](https://opentrons.atlassian.net/browse/RDEVOPS-96)

# Overview

this PR migrates from using AWS-SDK v2 to v3 and updates all our web deployment scripts (`promote-to-staging`, `promote-to-production`, `rollback`) and the utils used

NOTE: v3 does support typescript and i tried to update the scripts to using typescript but ran into a bunch of TS compiler issues 🫠. Seemed much easier to just stick with JS. Additionally, v2 is supposed to be deprecated as early as September 8th this year so figured it was better to update to v3 and still use JS than update to v3 and use TS (for now).

# Test Plan

Only someone with the proper aws creds can test this. But should be tested for docs, PD, and LL with promote to staging, promote to production, and rollback dry runs.

for promote to staging: `node ./scripts/deploy/promote-to-staging project tag` (PD tag is `protocol-designer@8.0.0`)

for promote to prod: `node ./scripts/deploy/promote-to-production project`

rollback: `node ./scripts/deploy/rollback project environment`

# Changelog

- install appropriate dependencies (aws-sdk v3 uses modules now)
- update all the deploy utils
- update staging, prod, and rollback scripts

# Review requests

see test plan

# Risk assessment

medium - we have no more releases that will use these anytime soon. The dry runs all seem to work! 

[RDEVOPS-96]: https://opentrons.atlassian.net/browse/RDEVOPS-96?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/deploy/create-invalidation.js
+++ b/scripts/deploy/create-invalidation.js
@@ -1,8 +1,10 @@
-const AWS = require('aws-sdk')
+const {
+  CloudFrontClient,
+  CreateInvalidationCommand,
+} = require('@aws-sdk/client-cloudfront')
 
-function getCreateInvalidation(productionCredentials, cloudfrontArn) {
-  const cloudfront = new AWS.CloudFront({
-    apiVersion: '2019-03-26',
+async function getCreateInvalidation(productionCredentials, cloudfrontArn) {
+  const client = new CloudFrontClient({
     region: 'us-east-1',
     credentials: productionCredentials,
   })
@@ -17,7 +19,11 @@ function getCreateInvalidation(productionCredentials, cloudfrontArn) {
       },
     },
   }
-  return cloudfront.createInvalidation(cloudFrontParams).promise()
+
+  const data = await client.send(
+    new CreateInvalidationCommand(cloudFrontParams)
+  )
+  return data.Invalidation.Id
 }
 
 module.exports = { getCreateInvalidation }
--- a/scripts/deploy/create-invalidation.js
+++ b/scripts/deploy/create-invalidation.js
@@ -3,7 +3,11 @@ const {
   CreateInvalidationCommand,
 } = require('@aws-sdk/client-cloudfront')
 
-async function getCreateInvalidation(productionCredentials, cloudfrontArn, dryrun) {
+async function getCreateInvalidation(
+  productionCredentials,
+  cloudfrontArn,
+  dryrun
+) {
   const client = new CloudFrontClient({
     region: 'us-east-1',
     credentials: productionCredentials,
--- a/scripts/deploy/create-invalidation.js
+++ b/scripts/deploy/create-invalidation.js
@@ -3,7 +3,7 @@ const {
   CreateInvalidationCommand,
 } = require('@aws-sdk/client-cloudfront')
 
-async function getCreateInvalidation(productionCredentials, cloudfrontArn) {
+async function getCreateInvalidation(productionCredentials, cloudfrontArn, dryrun) {
   const client = new CloudFrontClient({
     region: 'us-east-1',
     credentials: productionCredentials,
@@ -19,6 +19,8 @@ async function getCreateInvalidation(productionCredentials, cloudfrontArn) {
       },
     },
   }
+
+  if (dryrun) return Promise.resolve()
 
   const data = await client.send(
     new CreateInvalidationCommand(cloudFrontParams)
--- a/scripts/deploy/create-invalidation.js
+++ b/scripts/deploy/create-invalidation.js
@@ -24,7 +24,7 @@ async function getCreateInvalidation(
     },
   }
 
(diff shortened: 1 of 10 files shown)


scripts/deploy/promote-to-production.js
```diff
@@ -58,53 +58,37 @@ async function runPromoteToProduction() {
...
+    const s3Client = new S3Client({
+      apiVersion: '2006-03-01',
+      region: 'us-east-1',
+      credentials: productionCredentials,
+    })
+
+    const stagingBucket = `staging.${projectDomain}`
+    const productionBucket = projectDomain
+    console.log(`Promoting ${projectDomain} from staging to production\n`)
+    await getDeployMetadata(s3Client, stagingBucket)
```
sfoster1: it seems like `getDeployMetadata` used to be here to get something called `current` that was then logged, and not really any other purpose. This PR moves the log above `getDeployMetadata` and removes the use of its return, so maybe we should just get rid of `getDeployMetadata`

scripts/deploy/promote-to-production.js
```diff
@@ -58,53 +58,37 @@ async function runPromoteToProduction() {
...
+    await syncBuckets(
+      s3Client,
+      { bucket: stagingBucket },
+      { bucket: productionBucket },
+      dryrun
+    )
+
+    console.log('Promotion to production done\n')
+
+    getCreateInvalidation(productionCredentials, cloudfrontArn)
```
sfoster1: needs to be `await`-ed i think

scripts/deploy/promote-to-production.js
```diff
@@ -58,53 +58,37 @@ async function runPromoteToProduction() {
...
+
+    console.log('Cache invalidation initiated for production\n')
+    process.exit(0).catch(err => {
```
sfoster1: I don't think we need this `catch`, if it's the equivalent of the `catch` on the old line 100 then that functionality is covered by the function-wide `try/catch`

sfoster1 (changes requested):
Couple small things to change but thanks for doign this!

scripts/deploy/promote-to-production.js
jerader: oops yes this is not needed there. thanks!

scripts/deploy/promote-to-production.js
```diff
@@ -2,11 +2,10 @@
...
 const assert = require('assert')
-const AWS = require('aws-sdk')
+const { S3Client } = require('@aws-sdk/client-s3')
 
 const parseArgs = require('./lib/parseArgs')
 const syncBuckets = require('./lib/syncBuckets')
-const { getDeployMetadata } = require('./lib/deploy-metadata')
 const { getAssumeRole } = require('./assume-role')
 const { getCreateInvalidation } = require('./create-invalidation')
 const { checkCurrentAWSProfile } = require('./check-current-profile')
```
y3rsh, line 11: Cannot comment on the not edited collapsed lines but all the arns for all CloudFront distributions in this file and in the promote-staging.js are incorrect.  This is why we have had to do them manually?

y3rsh (changes requested):
Let us try and fix the invalidations.

scripts/deploy/promote-to-production.js
jerader: ooh, thanks Josh. i thought we haven't had to do them manually since i added the cloud invalidations to these scripts. But did the cloudfront arns change when we added the new administrator role arn last year?

y3rsh: Sorry PD CloudFront ARNs are correct.  ~~I think LL, static, docs may be off~~ but are not relevant to this PR.
No they are still in old account we think.

y3rsh (approved):

scripts/deploy/create-invalidation.js
```diff
@@ -17,7 +23,13 @@ function getCreateInvalidation(productionCredentials, cloudfrontArn) {
       },
     },
   }
-  return cloudfront.createInvalidation(cloudFrontParams).promise()
+
+  if (dryrun) return Promise.resolve()
```
sfoster1: I think you can just `return` here since it's `async function` not `function(...): Promise`

sfoster1 (approved):
Looks good, I think. There is that one weird `Promise.resolve()` that I don't think needs to be there but good otherwise

scripts/deploy/create-invalidation.js
jerader: ooooh ya good point thank you!

PR #15269, opened 2024-05-28
fix(step-generation): properly emit configureNozzleLayout command

jerader:
closes RQA-2781

# Overview

Step-generation was incorrectly assuming that the configure nozzle layout command would be emitted always before a pick up tip but that logic needed to be moved to the `replaceTip` instead.

But the bug was that the `configureNozzleLayout` command was emitting with a tip still on the pipette which caused the protocol to fail analysis

# Test Plan

first examine the attached protocol and see that there are 6 `configureNozzleLayout` commands. upload the file to PD and re-export. Confirm that there is now only 1 `configureNozzleLayout` command

[CELL LYSIS PROTOCOL - FLEX - HS One plate -v2.json](https://github.com/Opentrons/opentrons/files/15472666/CELL.LYSIS.PROTOCOL.-.FLEX.-.HS.One.plate.-v2.json)

# Changelog

- remove emitting the command from `transfer`, `mix`, `consolidate` and `distribute` and add the logic to `replaceTip`
- add a test case

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/commandCreators/atomic/replaceTip.ts
+++ b/step-generation/src/commandCreators/atomic/replaceTip.ts
@@ -22,6 +22,7 @@ import {
   getWasteChuteAddressableAreaNamePip,
 } from '../../utils'
 import { dropTip } from './dropTip'
+import { configureNozzleLayout } from './configureNozzleLayout'
 
 import type { NozzleConfigurationStyle } from '@opentrons/shared-data'
 import type {
@@ -29,6 +30,7 @@ import type {
   CommandCreatorError,
   CurriedCommandCreator,
 } from '../../types'
+
 interface PickUpTipArgs {
   pipette: string
   tiprack: string
@@ -91,7 +93,7 @@ export const replaceTip: CommandCreator<ReplaceTipArgs> = (
   prevRobotState
 ) => {
   const { pipette, dropTipLocation, nozzles, tipRack } = args
-
+  const stateNozzles = prevRobotState.pipettes[pipette].nozzles
   if (tipRack == null) {
     return {
       errors: [errorCreators.noTipSelected()],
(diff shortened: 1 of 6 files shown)


step-generation/src/commandCreators/atomic/replaceTip.ts
```diff
@@ -241,11 +244,24 @@ export const replaceTip: CommandCreator<ReplaceTipArgs> = (
   const addressableAreaNameWasteChute = getWasteChuteAddressableAreaNamePip(
     channels
   )
+
+  const configureNozzleLayoutCommand: CurriedCommandCreator[] =
```
koji, line 248: nit

```suggestion
  const configureNozzleLayoutCommands: CurriedCommandCreator[] =
```

jerader: ooh good catch!

step-generation/src/commandCreators/atomic/replaceTip.ts
koji, line 1: seems that L42, `invariantContext` isn't used

koji:
@jerader 
When importing the protocol you attached to the ticket, the app goes the white screen.

Screenshot 2024-05-28 at 12 08 12 PM


The protocol you attached to this PR worked as expected.

Are they different?

jerader:
> @jerader When importing the protocol you attached to the ticket, the app goes the white screen.
> 
> Screenshot 2024-05-28 at 12 08 12 PM .

@koji oops i attached the wrong one to the ticket. i just updated it. the one in this PR should work too. I accidentally added a "z" to the moduleId string in one location in the json file that whitescreens 😅

koji (approved):
The changes look good to me, and the sandbox worked as expected with the protocol you attached to this PR.

PR #15270, opened 2024-05-28
feat(app): generate quick transfer file from summary state

smb2268:
# Overview
This PR adds the logic to convert the quick transfer summary state to a JSON protocol file using `step-generation` utilities. It also creates the modal for users to determine whether to save a protocol for later or run it immediately, and wires this up.

# Test Plan
1. Generate a few different protocols using the quick transfer flow, ideally one of each path type. 
2. Look at the logged protocol file and compare to the output of a similar PD protocol
3. Select each "save" and "run" and ensure the protocol is stored or you are directed to run setup as expected
4. Protocols should pass analysis and you could proceed to run with the correct setup

https://github.com/Opentrons/opentrons/assets/14302493/4384e491-35d9-4156-97a0-71755e1e97b6

# Changelog
1. Add `generateQuickTransferArgs` util which takes a `QuickTransferSummaryState` object as well as the deck config and creates everything step-generation needs to make a protocol - `MoveLiquidStepArgs`, `RobotState`, and `InvariantContext`
2. Make `createQuickTransferFile` util which utilizes the `generateQuickTransferArgs` output to assemble load commands and calls the appropriate step generation function to get the rest of the commands. The function then appends these commands with other necessary protocol data into a JSON file
3. Add `SaveOrRunModal` which pops up upon pressing "Create transfer" and generates the protocol file. Depending on whether the user presses save or run - the file will be generated, sent to the robot server, and then either used to create a run or displayed on the protocols dashboard

# Review requests
I've tested this a fair amount and will continue to do so as we add more functionality. It'd be great for reviewers to look through the utils and understand the flow of data and get familiar with all of this to help with continued implementation!

# Risk assessment
Low, behind FF


app/src/assets/localization/en/quick_transfer.json
```diff
@@ -18,6 +18,10 @@
   "quick_transfer_volume": "Quick Transfer {{volume}}µL",
   "right_mount": "Right Mount",
   "reservoir": "Reservoirs",
+  "run_quick_transfer_now": "Do you want to run your quick transfer now?",
+  "run_now": "Run now",
```
jerader: i think alphabetically these should be swapped

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
```diff
@@ -0,0 +1,159 @@
...
+export function createQuickTransferFile(
+  quickTransferState: QuickTransferSummaryState,
+  deckConfig: DeckConfiguration
+): File {
+  const {
+    stepArgs,
+    invariantContext,
+    initialRobotState,
+  } = generateQuickTransferArgs(quickTransferState, deckConfig)
+  const pipetteEntity = Object.values(invariantContext.pipetteEntities)[0]
```
jerader, line 42: you are always going to have only 1 pipette entity?

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
```diff
@@ -0,0 +1,159 @@
...
+    acc.push({
+      key: uuid(),
+      commandType: 'loadLabware' as const,
+      params: {
+        displayName: entity.def.metadata.displayName,
+        labwareId: entity.id,
+        loadName: entity.def.parameters.loadName,
+        namespace: entity.def.namespace,
+        version: entity.def.version,
+        location: { slotName: initialRobotState.labware[entity.id].slot },
```
jerader: just a small nit here, you could deconstruct `entity` above so you won't have to call it so many times

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
```diff
@@ -0,0 +1,159 @@
...
+      },
+    })
+    return acc
+  }, [])
+
+  let nonLoadCommandCreator
```
jerader: nit: you can type this as `CommandCreator<args>`

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
smb2268: yep! Quick Transfer will only allow use of one pipette

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
```diff
@@ -0,0 +1,159 @@
...
+      invariantContext,
+      initialRobotState
+    )
+  } else if (stepArgs?.commandCreatorFnName === 'consolidate') {
+    nonLoadCommandCreator = consolidate(
+      stepArgs as ConsolidateArgs,
+      invariantContext,
+      initialRobotState
+    )
+  } else if (stepArgs?.commandCreatorFnName === 'distribute') {
```
jerader, line 117: i noticed in the protocol that i generated that a final `dropTip` command isn't generated. You might need to put in a rogue `dropTip` command creator in here, similarly to how step-generation has one in `protocol-designer/src/timelineMiddleware/generateRobotStateTimeline.ts`. Basically, the `transfer`, `consolidate` and `distribute` compound commands only generate drop tip if it is followed by a pick up tip (from the `replaceTip` logic). So if its the final pick up tip, a drop tip won't be associated with it.

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
```diff
@@ -0,0 +1,368 @@
...
+  const allWellsOrdered = orderWells(labwareDef.ordering, 't2b', 'l2r')
+  return intersection(allWellsOrdered, unorderedWells)
+}
+
+function getInvariantContextAndRobotState(
+  quickTransferState: QuickTransferSummaryState,
+  deckConfig: DeckConfiguration
+): { invariantContext: InvariantContext; robotState: RobotState } {
+  const pipetteId = uuid()
+  const tipRackId = uuid()
```
jerader: nit: this is fine! just curious why you didn't want to add a human readable identifier to the id? like `${uuid()}_${pipetteName}` and `${uuid()}_${defURI}`. I guess i'm biased since i like having it in PD, makes it easier imo to know what the id is in reference of.

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
```diff
@@ -0,0 +1,368 @@
...
+      },
+    }
+  }
+
+  const invariantContext = {
+    labwareEntities,
+    moduleEntities: {},
+    pipetteEntities,
+    additionalEquipmentEntities: additionalEquipmentEntities,
+    config: { OT_PD_DISABLE_MODULE_RESTRICTIONS: false },
```
jerader, line 181: do you need this config key?

app/src/organisms/QuickTransferFlow/SummaryAndSettings.tsx
```diff
@@ -9,27 +11,39 @@ import {
...
 export function SummaryAndSettings(
   props: SummaryAndSettingsProps
 ): JSX.Element | null {
-  const { onNext, exitButtonProps, state: wizardFlowState } = props
+  const { exitButtonProps, state: wizardFlowState } = props
+  const history = useHistory()
+  const queryClient = useQueryClient()
+  const host = useHost()
   const { t } = useTranslation(['quick_transfer', 'shared'])
+  const [showSaveOrRunModal, setShowSaveOrRunModal] = React.useState(false)
```
jerader: nit 
```suggestion
  const [showSaveOrRunModal, setShowSaveOrRunModal] = React.useState<boolean>(false)
```
```suggestion
  const [showSaveOrRunModal, setShowSaveOrRunModal] = React.useState(false)
```

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
smb2268: To appease typescript, yes. Or we could make that field optional on the `config` type!

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
smb2268: Yeah I could change this! I didn't think about making it human readable since quick transfer protocols can't be downloaded so the actual JSON won't be read except for our internal testing purposes. The user will only have access to what we display on the summary screen or protocol run!

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
jerader: oh ya good point, i think down the line we can make it optional. i think it is fine like this for now.

jerader (commented):
The data is looking awesome! nice work! 🦖 really cool to see step-generation utils being used outside of PD. I left a few comments and nits. 

Also, I looked at the data from a protocol and have a comment. Your `source` and `destination` key values are objects that look to be the labware definition associated with the source and destination. But then you have the `labwareDefinitions` key where the value is a record of all the used labware definitions. This results in the labware definition info duplicated. Seems unnecessary and that they can be consolidated somehow?

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
smb2268: I think this is actually `CommandCreatorResult` - I'll type it!

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
smb2268: ah good catch - thanks!

smb2268:
> Also, I looked at the data from a protocol and have a comment. Your `source` and `destination` key values are objects that look to be the labware definition associated with the source and destination. But then you have the `labwareDefinitions` key where the value is a record of all the used labware definitions. This results in the labware definition info duplicated. Seems unnecessary and that they can be consolidated somehow?

@jerader yeah that's a good point! I have a TODO to formalize the structure for `designerApplicationData` and figure out exactly what data I want to include. Planning to address that once more of the functionality is implemented!

app/src/organisms/QuickTransferFlow/utils/createQuickTransferFile.ts
```diff
@@ -0,0 +1,250 @@
...
+          : { slotName: location },
+      },
+    })
+    return acc
+  }, [])
+
+  let nonLoadCommandCreator: CommandCreatorResult | null = null
+  if (stepArgs?.commandCreatorFnName === 'transfer') {
+    nonLoadCommandCreator = transfer(
+      stepArgs as TransferArgs,
```
brenthagen: don't need these `stepArgs` type assertions

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
```diff
@@ -0,0 +1,418 @@
...
+    }
+  }
+  if (
+    quickTransferState.dropTipLocation === 'wasteChute' ||
+    quickTransferState.blowOut === 'wasteChute'
+  ) {
+    const wasteChuteLocation = deckConfig.find(
+      configCutout =>
+        configCutout.cutoutFixtureId.includes('WasteChute') ||
+        configCutout.cutoutFixtureId.includes('wasteChute')
```
brenthagen: prefer `WASTE_CHUTE_FIXTURES.includes(configCutout.cutoutFixtureId)`

brenthagen (approved):
tested both save and run, looks good to me !

jerader (approved):
this is awesome! Great job! 🌷 

Just a few notes but i don't think it should block this from merging:
1. probably some error handling is needed so you don't select an amount of wells to aspirate/dispense into that doesn't work (i selected 6 wells for aspirate and 4 for dispense and got this error)
Screenshot 2024-05-29 at 15 34 30

2. probably would be good to add some test coverage throughout these components

smb2268:
@jerader  thank you!!
> 1. probably some error handling is needed so you don't select an amount of wells to aspirate/dispense into that doesn't work (i selected 6 wells for aspirate and 4 for dispense and got this error)

We have a ticket for this, coming soon! https://opentrons.atlassian.net/browse/PLAT-181

> 2. probably would be good to add some test coverage throughout these components

And yes I agree! Going to make tests for the two new utils once I'm back from vacation

b-cooper (approved):
Looks great!  I agree with the comments from previous reviews. 🎆

PR #15271, opened 2024-05-28
test(robot-server): Fix datetime regexes

SyntaxColoring:
# Overview

This updates some regexes in `robot-server` integration tests to prepare for a Pydantic upgrade (PLAT-326).

# Changelog

The `robot-server` integration tests widely copy-paste this regex for datetimes:

```regex
\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+\+\d{2}:\d{2}
```

But that only accounts for timezones of the form `+HH:MM`. It's [also valid](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) for them to be `-HH:MM`, or `Z`, and it looks like Pydantic v2 does `Z`.

So this PR fixes it by updating the regexes to:

```regex
\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+(Z|([+-]\d{2}:\d{2}))
```

I'm also taking the opportunity to consistently use this regex for all `createdAt`, `startedAt`, etc. datetime fields, instead of `!anystr`.

# Test plan

# Review requests

None in particular.

# Risk assessment

No risk.

### DIFF
--- a/robot-server/tests/integration/http_api/protocols/test_deck_coordinate_load.tavern.yaml
+++ b/robot-server/tests/integration/http_api/protocols/test_deck_coordinate_load.tavern.yaml
@@ -24,7 +24,7 @@ stages:
           files:
             - name: deck_coordinate_load.py
               role: main
-          createdAt: !re_search "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+(Z|([+-]\\d{2}:\\d{2}))$"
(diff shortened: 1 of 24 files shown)


TamarZanzouri (approved):
Thank you! I see some files we just changed from `"` to `'` still fine by me :-)

SyntaxColoring:
Yeah, sorry for the extraneous changes from my editor's autoformatter. The good news is that it should be consistent with `eslint` now.

PR #15275, opened 2024-05-28
feat(api): add new pick up tip function

caila-marashaj:
## Overview
This is a follow-up to https://github.com/Opentrons/opentrons/pull/15143.

In order to support error recovery, we should restructure some of our api functions so that they _only_ do the motor routines necessary to do what they say to do, without any additional movements or information caching. This creates a new function that makes that change to `pick_up_tip`, while still leaving the original function in place.

## Changelog
- create an `execute_pick_up_tip` function that leaves out both of the pipette plunger moves at the beginning and end, and does not add the tip
- add some tests
- have `pick_up_tip` call this new function

## Review Requests
Is giving this new function a different signature between `api` and `ot3api` a no-no?

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1796,12 +1796,14 @@ class OT3API(
         self._gripper_handler.check_ready_for_jaw_move("hold_jaw_width")
         await self._hold_jaw_width(jaw_width_mm)
 
-    async def execute_pick_up_tip(
+    async def tip_pickup_moves(
         self,
         mount: OT3Mount,
         presses: Optional[int] = None,
         increment: Optional[float] = None,
     ) -> None:
+        """This is a slightly more barebones variation of pick_up_tip. This is only the motor routine
+        directly involved in tip pickup, and leaves any state updates and plunger moves to the caller."""
         realmount = OT3Mount.from_mount(mount)
         instrument = self._pipette_handler.get_pipette(realmount)
         if (
@@ -2204,7 +2206,7 @@ class OT3API(
 
         await self._move_to_plunger_bottom(realmount, rate=1.0)
 
-        await self.execute_pick_up_tip(realmount, presses, increment)
+        await self.tip_pickup_moves(realmount, presses, increment)
 
         add_tip_to_instr()
 
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1796,6 +1796,49 @@ class OT3API(
         self._gripper_handler.check_ready_for_jaw_move("hold_jaw_width")
         await self._hold_jaw_width(jaw_width_mm)
 
+    async def execute_pick_up_tip(
+        self,
+        mount: OT3Mount,
+        presses: Optional[int] = None,
+        increment: Optional[float] = None,
+    ) -> None:
+        realmount = OT3Mount.from_mount(mount)
+        instrument = self._pipette_handler.get_pipette(realmount)
+        if (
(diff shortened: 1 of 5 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1796,6 +1796,49 @@ async def hold_jaw_width(self, jaw_width_mm: int) -> None:
         self._gripper_handler.check_ready_for_jaw_move("hold_jaw_width")
         await self._hold_jaw_width(jaw_width_mm)
 
+    async def execute_pick_up_tip(
+        self,
+        mount: OT3Mount,
+        presses: Optional[int] = None,
+        increment: Optional[float] = None,
+    ) -> None:
```
SyntaxColoring: 1. Does this also need to exist on the OT-2 `API` class so the two classes can be substituted for each other?
2. Could we have docstrings defining what this does and how it's different from `pick_up_tip()`?
3. I feel like there could be a more specific name for this, but I don't know what. `pick_up_tip_bare()` or `just_pick_up_tip()` or `pick_up_tip_motions()` or something else?

SyntaxColoring (approved):
Makes sense to me, TY. Just a few suggestions.

api/src/opentrons/hardware_control/ot3api.py
caila-marashaj: Yeah thats a good point, I'll make another commit with some changes.

sfoster1: `pick_up_tip_in_place`? I think at some poitn we're going to want to get it out of this file anyway

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -1796,6 +1796,49 @@ async def hold_jaw_width(self, jaw_width_mm: int) -> None:
...
+        # This extra shake ensures those tips are removed
+        for rel_point, speed in spec.shake_off_moves:
+            await self.move_rel(realmount, rel_point, speed=speed)
+
+        # fixme: really only need this during labware position check so user
+        # can verify if a tip is properly attached
+        if spec.ending_z_retract_distance:
+            await self.move_rel(
+                realmount, top_types.Point(z=spec.ending_z_retract_distance)
+            )
```
sfoster1, line 1846: we need to do the simulator state update thing that's on line 2159 too, this:
```python
if isinstance(self._backend, OT3Simulator):
     self._backend._update_tip_state(realmount, True)
```

sfoster1 (changes requested):
Looks good overall, just a couple little things!

api/src/opentrons/hardware_control/ot3api.py
SyntaxColoring: I wouldn't use `_in_place` because in Protocol Engine, `_in_place` is used to roughly mean "do the operation without moving X/Y," whereas here, *neither* function moves in X/Y, and the difference instead has to do with tip-state and volume-state setting.

api/src/opentrons/hardware_control/ot3api.py
caila-marashaj: Why do we want this? My impression was that we want to leave any kind of state updates to the caller

sfoster1: This specifically is a hack to get the simulated tip presence sensors returning the right thing - that's why it's checking the backend. You're right in general, but this is an exception

sfoster1 (commented):
Looks good to me once we remove the simulator change from the old function

sfoster1 (approved):
last review was meant to be an approval

SyntaxColoring (approved):
🛻

PR #15278, opened 2024-05-29
refactor(components): remove close-icon icon

mjhuff:
# Overview

Per design (Mel), we no longer need the close-icon icon, since it is similar enough to the remove icon that we would use remove in any place we'd use close-icon. Close-icon is currently not used in the codebase, so no other refactor is required.

# Review requests

- Just quickly grep in the codebase to make sure I didn't overlook anything, but I'm pretty sure we don't use 'close-icon' anywhere that would require a refactor to 'remove'. 

# Risk assessment
low


koji (approved):
lgtm

PR #15280, opened 2024-05-29
fix(shared-data): defaults should match valuesByApiLevel for p50 3_6 lowVolume

jerader:
closes RQA-2785

# Overview

Somehow i forgot to change the `defaults` for the p50 3_6 pipette version to match the `uiMaxFlowRate` and `valuesByApiLevel`. I double checked and this is the only model with the issue.

# Test Plan

Just confirm that the `defaults` matches the `valuesByApiLevel`

# Changelog

- change pipette definition

# Review requests

see test plan

# Risk assessment

low


DerekMaggio (approved):
The award for fastest ticket report time to fix time of 3 minutes goes to @jerader. Compared to other pipettes and the pattern of default == valuesByAPILevel is used on other pipettes

sfoster1 (approved):

PR #15147, opened 2024-05-09
fix(shared-data) Update tip overlap values for Flex Pipettes

Carlos-fernandez:
# Overview
After conducting hardware testing, we've formulated updated tip overlap values for every tip size and pipette model. These adjustments apply to 1-channel, 8-channel, and 96-channel pipettes. These revised values aim to mitigate overpressure occurrences in ABR. 

[https://docs.google.com/spreadsheets/d/16dEmqulO4SIKU6E0aYTfVQEnZRTn3MwNF_WB8K1KmRM/edit#gid=0](url)

# Test Plan
None

# Changelog
### P1KS Single Channel
- Updated V3_5 Tip Overlap values for 50uL, 200uL, and 1000uL tips.
- Updated V3_6 Tip Overlap values for 50uL 200uL, and 1000uL tips.
### P50S Single Channel Pipette
- Updated V3_5 Tip Overlap values for 50uL, 200uL, and 1000uL tips.
- Updated V3_6 Tip Overlap values for 50uL 200uL, and 1000uL tips.
### P1KM 8 Channel Pipette
- Updated V3_5 Tip Overlap values for 50uL, 200uL, and 1000uL tips.
- Updated V3_6 Tip Overlap values for 50uL 200uL, and 1000uL tips.
### P50M 8 Channel Pipette
- Updated V3_5 Tip Overlap values for 50uL, 200uL, and 1000uL tips.
- Updated V3_6 Tip Overlap values for 50uL 200uL, and 1000uL tips
### P1H 95-Channel Pipette
- Updated V3_5 Tip Overlap values for 50uL, 200uL, and 1000uL tips.
- Updated V3_6 Tip Overlap values for 50uL 200uL, and 1000uL tips

# Review requests

# Risk assessment
Low

PR #15284, opened 2024-05-29
fix(shared-data): Re-apply update tip overlap values for Flex Pipettes (#15147)

sfoster1:
This is #15147 again but this time outside the context of an extremely imminent release. 

While this code needs to be merged, we should carefully consider how we do so. We can't ship this until we can separate this behavior by API level so that customers who may have tuned in offsets  based on the old incorrect tip offsets won't have their protocols invalidated - this data needs to be loaded by API level.

Closes EXEC-452


SyntaxColoring (approved):
Approved previously in #15147, looks good if and when the versioning considerations are solved.

Carlos-fernandez (approved):
All looks correct! Good to go!

api/tests/opentrons/protocol_api/core/engine/test_overlap_versions.py
```diff
@@ -0,0 +1,26 @@
...
+    """Versions below 2.19 should use v0."""
+    assert overlap_for_api_version(api_version) == "v0"
+
+
+@pytest.mark.parametrize("api_version", versions_at_or_above(APIVersion(2, 19)))
+def test_all_above_219_use_v1(api_version: APIVersion) -> None:
+    """Versions above 2.19 should use v1."""
+    assert overlap_for_api_version(api_version) == "v1"
+
+
```
DerekMaggio, line 23: Can you add a test to verify that a bump to the major version is evaluated correctly, please? So something like `APIVersion(3, 12)` 
I can imagine this being something missed because a 3.x would likely be years from now and subsequently breaking everything.

DerekMaggio (changes requested):

api/tests/opentrons/protocol_api/core/engine/test_overlap_versions.py
DerekMaggio: Actually I might be incorrect, it looks like the underlying versions_between would throw an assertion error https://github.com/Opentrons/opentrons/blob/3643682c0c213286b8c31666f5aabfefeadab705/api/tests/opentrons/protocol_api/__init__.py#L102C1-L104C75
If that is true, I will change to approval

api/src/opentrons/protocol_api/core/engine/__init__.py
```diff
@@ -12,6 +12,7 @@
 ENGINE_CORE_API_VERSION: Final = APIVersion(2, 14)
 SET_OFFSET_RESTORED_API_VERSION: Final = APIVersion(2, 18)
 
+
```
TamarZanzouri: snuck in there

TamarZanzouri (approved):
Nice! LGTM!

skowalski08 (approved):
tested 2.18,2.19, and json -lgtm

api/tests/opentrons/protocol_api/core/engine/test_overlap_versions.py
sfoster1: Yes, that's true @DerekMaggio

DerekMaggio (approved):

PR #15286, opened 2024-05-29
fix(protocol-designer): set numMultiples to numItems instead of 0

jerader:
closes RQA-2787

# Overview

This was a good catch by @DerekMaggio! Basically, the `numMultiples` was being set to 0 on render instead of remembering the actual number. So if you go back to change something, the `numMultiples` would be reset to 0, disallowing the user to select the down arrow.

# Test Plan

follow the test plan outlined in the attached ticket!

# Changelog

- set the `useState` correctly

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm
tested all cases and the button worked as expected.

great job, @DerekMaggio!

PR #15290, opened 2024-05-30
feat(api): Add `isDefined` field to `ErrorOccurrence` and update docs

SyntaxColoring:
# Overview

Closes EXEC-453.

# Test plan

* [x] Check the server's `/redoc` page to make sure this renders reasonably in the documentation for e.g. the `GET /runs/{id}/commands/{id}` endpoint.

# Changelog

As described in [the proposal](https://opentrons.atlassian.net/wiki/spaces/PER/pages/4126703622/Proposal+Structured+representation+of+recoverable+errors#JSON-representation):

* Add an `isDefined: bool` field to command errors. Default it to `false` so we can still read existing commands in the database that don't have this field. Set it to `true` for `tipPhysicallyMissing`, which is the one defined error that currently exists.
* Update the documentation of the error fields, especially to describe the new proper use of `errorType`.

# Review requests

Does this make sense to people other than me?

# Risk assessment

Very low.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/pick_up_tip.py
+++ b/api/src/opentrons/protocol_engine/commands/pick_up_tip.py
@@ -72,9 +72,10 @@ class TipPhysicallyMissingError(ErrorOccurrence):
     of the pipette.
     """
 
+    isDefined = True
     errorType: Literal["tipPhysicallyMissing"] = "tipPhysicallyMissing"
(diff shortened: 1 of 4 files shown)


sfoster1 (approved):
Nice, makes sense to me!

TamarZanzouri (approved):
LGTM! you have a test failing :-)

mjhuff (approved):
Makes sense! Thank you!

PR #15291, opened 2024-05-30
feat(app): add tip management tab and items

smb2268:
# Overview
This PR adds the tip management tab with change tip and drop tip location settings
Screen Shot 2024-05-30 at 5 08 13 PM
Screen Shot 2024-05-30 at 5 08 00 PM
Screen Shot 2024-05-30 at 5 07 51 PM

# Test Plan
Play around with the tip management options at the end of a quick transfer flow

# Changelog
1. Added TipManagement, ChangeTip, and TipDropLocation components along with test cases for each
2. Change the type of dropTipLocation to CutoutConfig so that we can differentiate between multiple trash bins or waste chutes and adjust functions accordingly

# Review requests
Look over components, logic to determine what inputs are acceptable, and type change for errors

I'll be out for the review process so @b-cooper or @brenthagen can take over for updates and merging!

# Risk assessment
Low, behind FF


app/src/organisms/QuickTransferFlow/TipManagement/TipDropLocation.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { t } = useTranslation('quick_transfer')
+  const deckConfig = useNotifyDeckConfigurationQuery().data ?? []
+
+  const tipDropLocationOptions = deckConfig.reduce<CutoutConfig[]>(
+    (acc, cutoutConfig) => {
+      if (WASTE_CHUTE_FIXTURES.includes(cutoutConfig.cutoutFixtureId)) {
+        acc.push(cutoutConfig)
+      } else if (TRASH_BIN_ADAPTER_FIXTURE === cutoutConfig.cutoutFixtureId) {
+        acc.push(cutoutConfig)
+      }
```
b-cooper: looks like this could be collapse into a single `if` with an `||` ?

app/src/organisms/QuickTransferFlow/TipManagement/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+    {
+      option: t('change_tip'),
+      value: t(`${state.changeTip}`),
+      onClick: () => setSelectedSetting('change_tip'),
+    },
+    {
+      option: t('tip_drop_location'),
+      value: t(
+        `${
+          state.dropTipLocation.cutoutFixtureId === TRASH_BIN_ADAPTER_FIXTURE
```
brenthagen, line 48: i'm intermittently getting `Cannot read properties of undefined (reading 'cutoutFixtureId')` here, looking into why `dropTipLocation` would be undefined

app/src/organisms/QuickTransferFlow/utils/getInitialSummaryState.ts
```diff
@@ -10,62 +18,75 @@ import type {
...
     path: path,
     tipPositionAspirate: 1,
     preWetTip: false,
     tipPositionDispense: 1,
-    // TODO expand default logic for change tip depending on path, transfer type, number of tips
     changeTip,
-    // TODO add default logic for drop tip location depending on deck config
-    dropTipLocation: 'trashBin',
+    // @ts-expect-error this will not be undefined if there is a trash in deck config
+    dropTipLocation: trashConfigCutout,
```
brenthagen: this is why i'm getting `Cannot read properties of undefined (reading 'cutoutFixtureId')` and app crash when the deck config doesn't have a trash. we can drop a TODO, but do we have a plan/ticket to prompt the user to update the deck config at some point in the quick transfer flow if no trash or waste chute is configured?

brenthagen: pushed a commit to default the drop tip location to a trash bin in A3 if no trash or waste chute is configured. i think this is the best option - this should be an edge case, and the user will be able to address the trash in A3 deck configuration on run setup.

brenthagen (approved):
tested the tip management options with various configurations, code looks good to me.

app/src/organisms/QuickTransferFlow/utils/getInitialSummaryState.ts
smb2268: We should to talk to Felix about this, we could treat this similarly to how we treat needing to have a pipette attached - with an error modal when you try to launch the flow letting the user know they need to have a trash configured before creating a quick transfer. Or we can leave this defaulted if they don't have one attached!

PR #15298, opened 2024-05-31
refactor(api-client,shared-data): Update TS types for new `isDefined` field on errors

SyntaxColoring:
# Overview

I forgot to do the TypeScript part in #15290, sowwwyyyyy.

# Test Plan

# Changelog

Update TypeScript bindings for #15290—add an `isDefined` field to command and run errors.

# Review requests

Does it make sense to alias `RunError` with `RunCommandError` like this?

# Risk assessment

No risk.

### DIFF
--- a/api-client/src/runs/types.ts
+++ b/api-client/src/runs/types.ts
@@ -4,6 +4,7 @@ import type {
   LoadedModule,
   LoadedPipette,
   ModuleModel,
+  RunCommandError,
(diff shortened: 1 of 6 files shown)


mjhuff (approved):
✅ 

Yeah this is totally fine as long as we keep the comment IMO.

EDIT: Looks like there are a couple more lints that need fixing but it shouldn't be any real issue!

PR #15299, opened 2024-05-31
fix(opentrons-ai-client): fix example buttons and clear text after submitting prompt

shlokamin:
# Overview

closes AUTH-434 and AUTH-437

This PR fixes the example buttons and clears the input prompt text after clicking the submit button.

I also removed the unnecessary jotai state variable that was holding onto prompt button text because the prompt button can access the form directly. This allows us to get rid of an unnecessary useEffect which was just listening for changes to the state variable and setting form state with the updated state variable value.

Also, testing react-hook-form is a massive pain. sometimes we need to mock internals, and sometimes we want to test full integration. I introuced patterns for both in this PR. Though honestly I'm not even convinced we need react hook form. Ideally we can get rid of both that and jotai

# Changelog

- Clear prompt text after submitting
- Enable same button to be pressed after initial press

# Review requests

- Check to see that the input prompt text clears after pressing the submit button
- Check to see that you can press the same button twice after submitting

# Risk assessment

Low


opentrons-ai-client/src/App.tsx
```diff
@@ -64,8 +73,12 @@ export function App(): JSX.Element | null {
...
         </LinkButton>
       </Flex>
-      <SidePanel />
-      <ChatContainer />
+      <>
+        <FormProvider {...methods}>
+          <SidePanel />
+          <ChatContainer />
+        </FormProvider>
+      </>
```
koji: I think we don't need Fragment here. Is this intentional?

shlokamin: good catch, i had it originally because i was wrapping two sibling components but i dont need it anymore. thanks!

opentrons-ai-client/src/organisms/PromptButton/__tests__/PromptButton.test.tsx
```diff
@@ -1,36 +1,49 @@
...
+      setValue: mockSetValue,
+    })),
+  }
+})
+
 const render = (props: React.ComponentProps<typeof PromptButton>) => {
   return renderWithProviders(<PromptButton {...props} />)
 }
 
+let mockSetValue = vi.fn()
```
koji, line 25:
```suggestion
const mockSetValue = vi.fn()
```

opentrons-ai-client/src/organisms/PromptButton/__tests__/PromptButton.test.tsx
```diff
@@ -1,36 +1,49 @@
...
+
 describe('PromptButton', () => {
   let props: React.ComponentProps<typeof PromptButton>
+
   beforeEach(() => {
     props = {
       buttonText: 'Reagent Transfer',
     }
+    mockSetValue = vi.fn()
```
koji, line 34:
```suggestion

```

koji (commented):
lgtm!
thank you for fixing this!

koji (approved):

PR #15300, opened 2024-05-31
chore(api): depend on packaging

sfoster1:
We recently added an explicit runtime dependency on packaging but didn't add it to our deps. Add it now.

Picked 22.0 as the version to depend on because it introduced type stubs.

Closes RDEVOPS-98


SyntaxColoring (approved):
Do we also need an `==` pin for `packaging` in `Pipfile` matching what's installed on robots, to match how we do `pydantic` etc.?

sfoster1:
> Do we also need an `==` pin for `packaging` in `Pipfile` matching what's installed on robots, to match how we do `pydantic` etc.?

very good point, updated to include it

SyntaxColoring (approved):
📦

PR #15302, opened 2024-05-31
chore(api): add API version 2.19

sfoster1:
This carves out a 2.19 API version that will contain updated tip overlap data.

Closes EXEC-452


ecormany (approved):
Fine by me. This value doesn't affect anything in docs anymore — we severed that link and update the current version manually in the docs `conf.py` file. So it's also fine for this to go all the way into `edge` at any point, from my perspective.

y3rsh (approved):

PR #15307, opened 2024-05-31
fix(opentrons-ai-client) fix textarea overlapping issue

koji:
# Overview
fix Textarea overlapping issue and fix auto scrolling.

attached the video to the ticket since I cannot upload the big video to Github.

close AUTH-435

# Test Plan
- run the app
- send a few prompts

# Changelog
- update height to fix the overlapping issue
- rename component name (ChatContainer -> MainContentContainer)

# Review requests

# Risk assessment
low


opentrons-ai-client/src/molecules/ChatDisplay/index.tsx
```diff
@@ -65,19 +65,31 @@ export function ChatDisplay({ chat, chatId }: ChatDisplayProps): JSX.Element {
...
-            li: ListItemText,
-            p: ParagraphText,
-            a: ExternalLink,
-            code: CodeText,
-          }}
-        >
-          {reply}
-        </Markdown>
+        {isUser ? (
+          <Markdown
```
shlokamin: sorry if this is obvious, but why do we conditionally change styles if the role is user?

koji: the reason we need to use conditional here is that we don't display code hight lighter in user prompt.

oops, this shouldn't be in the PR.
https://github.com/Opentrons/opentrons/pull/15308

I think we can do something like this.

```ts
<Markdown
            components={{
              div: undefined,
              ul: UnnumberedListText,
              h2: HeaderText,
              li: ListItemText,
              p: ParagraphText,
              a: ExternalLink,
              code: isUser ? undeifned : CodeText,
            }}
          >
```

koji: https://github.com/Opentrons/opentrons/pull/15308/files



I'm going to remove this conditional statement from the PR.

shlokamin (approved):

PR #15309, opened 2024-05-31
test(app, components, protocol-designer): testing library lint rules to error, fix all occurrences

b-cooper:
# Overview

Promote the two previously demoted testing library eslint rules back to error and correct all
outstanding occurrences.

# Risk assessment
low, only test files are effected


shlokamin (approved):
thank you!

PR #15310, opened 2024-05-31
chore(ai-server): continuous deployment to staging

y3rsh:
I am not certain that the OIDC role has all the right `ECS` permissions but this will let us know if anything is missing if it fails.

On merge to edge of any changes in `opentrons-ai-server` deploy the change to the staging environment.

Also note that this is another reason why storing the environment variables in a secret as a map is useful.  Those values are needed in the deploy.py to define the task and we can give the OIDC role the permission to retrieve them as I already have.


shlokamin (approved):
lgtm

PR #15312, opened 2024-06-03
fix(protocol-designer): duplicate labware thunk refinement for offdeck

jerader:
closes RQA-2792

# Overview

the duplicate labware logic didn't work for offdeck labware sometimes (when there is no available slot on the deck)

# Test Plan

Create a flex or ot-2 protocol and fill the deck up with labware. Then add an off-deck labware and make sure the duplicate button works as expected

# Changelog

- fix the thunk logic for duplicating labware

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/labware-ingred/actions/thunks.ts
+++ b/protocol-designer/src/labware-ingred/actions/thunks.ts
@@ -131,7 +131,7 @@ export const duplicateLabware: (
     robotType,
     labwareDef
   )
-  if (duplicateSlot == null) {
+  if (duplicateSlot == null && !templateLabwareIdIsOffDeck) {
     console.error('no slots available, cannot duplicate labware')
   }
   const allNicknamesById = uiLabwareSelectors.getLabwareNicknamesById(state)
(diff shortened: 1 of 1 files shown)


protocol-designer/src/labware-ingred/actions/thunks.ts
```diff
@@ -140,15 +140,29 @@ export const duplicateLabware: (
     Object.keys(allNicknamesById).map((id: string) => allNicknamesById[id]), // NOTE: flow won't do Object.values here >:(
     templateNickname
   )
+  const duplicateLabwareId = uuid() + ':' + templateLabwareDefURI
```
koji, line 143: nit

```suggestion
  const duplicateLabwareId = `${uuid()}:${templateLabwareDefURI}`
```

jerader: good call out Koji! I might keep this as is for now since i need to merge this PR in asap. But I'll note it as a clean up item post 8.1 release

ncdiehl11 (approved):
tested and looks good

PR #15314, opened 2024-06-03
fix(protocol-designer): allow wasteChute selection when staging area in d3

jerader:
closes RQA-2790

# Overview

Fixes a bug where the waste chute option was incorrectly disabled

# Test Plan

follow the steps in the ticket but basically:

create flex protocol and add a staging area in slot D3. Then add a TC, a heater-shaker, and temperature modules until all the modules are disabled (except for the magnetic block). See that the waste chute is still enabled and is clickable. 

# Changelog

- refined logic for when the slot is available and special cased the waste chute
- added a test case

# Review requests

see test plan

# Risk assessment

low


ncdiehl11 (approved):
tested and code LGTM ✅

PR #15316, opened 2024-06-03
feat: configure performance-metrics to work on robot

DerekMaggio:
# Overview

Adds packaging scripts and configuration to enable pushing and running performance metrics on a robot. 

Currently running analysis and getting a cached analysis are the only things that are tracked on the robot 

Closes https://opentrons.atlassian.net/browse/EXEC-497

# Test Plan

- Follow the steps in [README](https://github.com/Opentrons/opentrons/blob/b7274f123ee5cbdd93430a52668d07b0c90b6ae8/performance-metrics/README.md) to setup the env
- Run robot analysis for 2-3 different protocols by going to the setup screen for the protocols
- Repeat running analysis for each protocol
- SSH into the robot and go to /data/performance_metrics_data/
- There should be 2 files 
  - robot_context_data
  - robot_context_data_headers
- Check their content
  - robot_context_data should have lines that look like this, 
    ```
    ANALYZING_PROTOCOL,1718739269730290312,14028043893
    GETTING_CACHED_ANALYSIS,1718739323578647179,47381526
    
    ...
    ```
  - robot_context_data_headers should have a single line 
    ```
    state_name,function_start_time,duration
    ```

- After you are finished run `make unset-performance-metrics-ff`
- Run a protocol and make sure performance metrics are no longer being recorded. Also, make sure you can still run the analysis and get a cached analysis

- [x] Verify analysis and cached analysis on the OT-2 dev server
- [x] Verify analysis and cached analysis on the Flex dev server
- [x] Verify analysis and cached analysis on OT-2
- [x] Verify analysis and cached analysis on Flex Dev Kit
- [x] Verify analysis and cached analysis on Flex @skowalski08 

# Changelog

## `performance_helpers.py`

- Add TrackingFunctions class which contains functions intended to be used as decorators 
- Generalize wrapper creation function
- Make all other entities in the file private

## `performance-metrics/Makefile`

- Add push targets for OT-2 and Flex
- Add helper target `update-robot-version` to hack the robot into thinking it is on the same version as the app
- Add `set-performance-metrics-ff` and `unset-performance-metrics-ff`
- Add a get-logs target 

## `performance-metrics/setup.py`

Remove shared-data as a dep for performance-metrics as it does not use it 

## `robot-server/Pipfile`

Remove performance-metrics as a dependency as robot_server is not using it directly

## `robot-server/robot_server/protocols/protocol_analyzer.py`

Wrap `analyze` and label it as `ANALYZING_PROTOCOL`

## `robot-server/robot_server/protocols/router.py`

Wrap logic that gets cached_analysis and label it as `GETTING_CACHED_PROTOCOL_ANALYSIS`

# Review requests

- Take a look at my test plan, is there anything else that should be added to reduce the risk?
- Am I installing the performance-metrics package correctly on the flex by just unpacking it in `/opt/opentrons-robot-server`? 
- 

# Risk assessment

Medium? `performance-metrics` is now wrapping production code. But the package is not being built and installed on robots, it has to be pushed by a dev. Furthermore, nothing will be run without setting the feature flag

### DIFF
--- a/performance-metrics/Makefile
+++ b/performance-metrics/Makefile
@@ -23,7 +23,7 @@ ssh_helper = $(if $(ssh_key),-i $(ssh_key)) $(ssh_opts)
 wheel_file = dist/$(call python_get_wheelname,performance-metrics,$(project_rs_default),performance_metrics,$(BUILD_NUMBER))
 
 # Find the version of the sdist file from git using a helper script.
-sdist_file = dist/$(call python_get_sdistname,performance_metrics,$(project_rs_default),performance_metrics)
+sdist_file = dist/$(call python_get_sdistname,performance-metrics,$(project_rs_default),performance_metrics)
 
 
 clean_cmd = $(SHX) rm -rf 'build' '**/*.egg-info' '**/__pycache__' **/*.pyc '.mypy_cache' '.pytest_cache'
--- a/performance-metrics/Makefile
+++ b/performance-metrics/Makefile
@@ -84,6 +84,29 @@ push-no-restart-ot3: sdist
 push-ot3: push-no-restart-ot3
 	$(call restart-server,$(host),$(ssh_key),$(ssh_opts),"opentrons-robot-server")
 
+.PHONY: update-robot-version
+update-robot-version:
+	$(eval update_dict := '{"opentrons_api_version": "$(version)", "update_server_version": "$(version)", "robot_server_version": "$(version)", "server_utils_version": "$(version)", "opentrons_hardware_version": "$(version)"}')
+	$(call sync-version-file,$(host),$(ssh_key),$(ssh_opts),'$(update_dict)')
+	$(call restart-service,$(host),$(ssh_key),$(ssh_opts),"opentrons-robot-server")
+
+
+.PHONY: set-performance-metrics-ff
+set-performance-metrics-ff:
+	@curl \
+    -H "opentrons-version: *" \
+    -X POST $(host):31950/settings \
+    -H "content-type: application/json" \
+    -d '{"id": "enablePerformanceMetrics", "value": true}'
+
+.PHONY: unset-performance-metrics-ff
+unset-performance-metrics-ff:
+	@curl \
+	-H "opentrons-version: *" \
+	-X POST $(host):31950/settings \
+	-H "content-type: application/json" \
+	-d '{"id": "enablePerformanceMetrics", "value": false}'
+
 .PHONY: test
 test:
 	$(pytest) tests
--- a/performance-metrics/Makefile
+++ b/performance-metrics/Makefile
@@ -25,6 +25,9 @@ wheel_file = dist/$(call python_get_wheelname,performance-metrics,$(project_rs_d
 # Find the version of the sdist file from git using a helper script.
 sdist_file = dist/$(call python_get_sdistname,performance-metrics,$(project_rs_default),performance_metrics)
 
+# Find the branch, sha, version that will be used to update the VERSION.json file
+version_file = $(call python_get_git_version,performance-metrics,$(project_rs_default),performance_metrics)
+
 
 clean_cmd = $(SHX) rm -rf 'build' '**/*.egg-info' '**/__pycache__' **/*.pyc '.mypy_cache' '.pytest_cache'
 clean_wheel_cmd = $(clean_cmd) dist/*.whl
--- a/performance-metrics/Makefile
+++ b/performance-metrics/Makefile
@@ -81,11 +81,11 @@ push: push-no-restart
 
 .PHONY: push-no-restart-ot3
 push-no-restart-ot3: sdist
-	$(call push-python-sdist,$(host),$(ssh_key),$(ssh_opts),$(sdist_file),"/opt/opentrons-robot-server","performance_metrics",,,$(version_file))
+	$(call push-python-sdist,$(host),$(ssh_key),$(ssh_opts),$(sdist_file),/opt/opentrons-robot-server,performance_metrics,src,,$(version_file))
 
 .PHONY: push-ot3
 push-ot3: push-no-restart-ot3
-	$(call restart-server,$(host),$(ssh_key),$(ssh_opts),"opentrons-robot-server")
+	$(call restart-service,$(host),$(ssh_key),$(ssh_opts),"opentrons-robot-server")
 
 .PHONY: update-robot-version
 update-robot-version:
--- a/performance-metrics/Makefile
+++ b/performance-metrics/Makefile
@@ -1,4 +1,35 @@
 include ../scripts/python.mk
+include ../scripts/push.mk
+
+ot_project := $(OPENTRONS_PROJECT)
+project_rs_default = $(if $(ot_project),$(ot_project),robot-stack)
+project_ir_default = $(if $(ot_project),$(ot_project),ot3)
+
+SHX := npx shx
+
+# Host key location for robot
+ssh_key ?= $(default_ssh_key)
+# Other SSH args for robot
+ssh_opts ?= $(default_ssh_opts)
+# Helper to safely bundle ssh options
+ssh_helper = $(if $(ssh_key),-i $(ssh_key)) $(ssh_opts)
+
+# Defined separately than the clean target so the wheel file doesn’t have to
+# depend on a PHONY target
+
+# Find the version of the wheel from git using a helper script. We
+# use python here so we can use the same version normalization that will be
+# used to create the wheel.
+wheel_file = dist/$(call python_get_wheelname,performance-metrics,$(project_rs_default),performance_metrics,$(BUILD_NUMBER))
+
+# Find the version of the sdist file from git using a helper script.
+sdist_file = dist/$(call python_get_sdistname,performance_metrics,$(project_rs_default),performance_metrics)
+
+
+clean_cmd = $(SHX) rm -rf 'build' '**/*.egg-info' '**/__pycache__' **/*.pyc '.mypy_cache' '.pytest_cache'
+clean_wheel_cmd = $(clean_cmd) dist/*.whl
+clean_sdist_cmd = $(clean_cmd) dist/*.tar.gz
+clean_all_cmd = $(clean_cmd) dist
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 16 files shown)


performance-metrics/Makefile
```diff
@@ -20,12 +54,61 @@ teardown:
...
+	$(SHX) rm -rf build
+	$(SHX) ls dist
+
+.PHONY: push-no-restart
+push-no-restart: wheel
+	$(call push-python-package,$(host),$(ssh_key),$(ssh_opts),$(wheel_file))
+
+.PHONY: push
+push: push-no-restart
+	$(call restart-service,$(host),$(ssh_key),$(ssh_opts),"opentrons-robot-server")
```
SyntaxColoring, line 80: ~~This target is `push-no-restart` but it looks like the recipe will restart `opentrons-robot-server`?~~ nvm, I misread this.

performance-metrics/Makefile
```diff
@@ -20,12 +54,61 @@ teardown:
...
+.PHONY: push-no-restart-ot3
+push-no-restart-ot3: sdist
+	$(call push-python-sdist,$(host),$(ssh_key),$(ssh_opts),$(sdist_file),/opt/opentrons-robot-server,performance_metrics,src,,$(version_file))
+
+.PHONY: push-ot3
+push-ot3: push-no-restart-ot3
+	$(call restart-service,$(host),$(ssh_key),$(ssh_opts),"opentrons-robot-server")
+
+.PHONY: update-robot-version
+update-robot-version:
```
SyntaxColoring: Nitpick: I might call this `set-robot-version` or `mark-robot-version` to make it clearer that it's not actually "updating the robot" in the sense of, like, a normal user-facing software update from v7.2 to v7.3.

performance-metrics/setup.cfg
SyntaxColoring, line 1: Why do we have both setup.cfg and setup.py? Does something merge their contents, or is one overriding the other?

I only have a tenuous understanding of all of this, so it's definitely possible I'm missing something.

robot-server/Pipfile
```diff
@@ -36,7 +36,6 @@ sqlalchemy2-stubs = "==0.0.2a21"
 # limited by tavern
 python-box = "==6.1.0"
 types-paho-mqtt = "==1.6.0.20240106"
-performance-metrics = {file = "../performance-metrics", editable = true}
```
SyntaxColoring, line 39: Wasn't this the thing that installs performance-metrics into robot-server's dev environment, so it's available when you do `make -C robot-server dev`? How are we allowed to delete it?

robot-server/robot_server/protocols/router.py
```diff
@@ -273,48 +274,54 @@ async def create_protocol(  # noqa: C901
...
-        try:
-            analysis_summaries, _ = await _start_new_analysis_if_necessary(
-                protocol_id=cached_protocol_id,
-                analysis_id=analysis_id,
-                rtp_values=parsed_rtp,
-                force_reanalyze=False,
-                protocol_store=protocol_store,
-                analysis_store=analysis_store,
-                analyses_manager=analyses_manager,
+        @TrackingFunctions.track_getting_cached_protocol_analysis
```
SyntaxColoring, line 278: This looks like it'll work for now, but it seems like it's turning out that we want the trackers to work as context managers instead of decorators, right? In other words, we shouldn't have to turn a chunk of code into an inner function just for the sake of packaging it up and handing to `track_getting_cached_protocol_analysis()`.

performance-metrics/README.md
```diff
@@ -1,3 +1,69 @@
...
+
+Then run
+
+```bash
+make -C performance-metrics update-robot-version version=<app-version> host=<robot-ip>
+```
+
+### Enabling performance-metrics feature flag
+
+Performance metrics usage is hidden behind a feature flag. To enable it run the following command (assumes you have the `host` environment variable set):
```
SyntaxColoring: Looks like the following commands set `host` explicitly, so these instructions don't need to assume that you have it set as an environment variable?

performance-metrics/README.md
```diff
@@ -1,3 +1,69 @@
...
+```bash
+make -C performance-metrics push-no-restart-ot3 host=<flex-ip>
+make -C api push-no-restart-ot3 host=<flex-ip>
+make -C robot-server push-ot3 host=<flex-ip>
+```
+
+### Extra step when running against real robots
+
+Once this is done you might need to hack getting your robot and app to think they are on the same version.
+Go to your app -> Settings -> General and find your app version
```
SyntaxColoring, line 49: Just so I'm clear, this isn't specifically a performance-metrics thing, right? It's a thing in general when you're using a stable app to interact with a dev-tainted robot?

SyntaxColoring (changes requested):
Changes requested for ~~performance-metrics/Makefile and~~ robot-server/Pipfile (unless I'm missing something). Other than that, LGTM. Thanks!

robot-server/robot_server/protocols/router.py
SyntaxColoring: I also think you and @sfoster1 have been trying hard to avoid `performance_metrics` introducing even small overhead. I'm not sure if this is what we're doing now, but, for example, at one point you were checking the feature flag once on first import to avoid the overhead of an `if`/`else` every time the function under measurement was called. Is that still a goal?

robot-server/Pipfile
DerekMaggio: `robot-server` doesn't directly use the `performance-metrics` library. The `api` project does, and it is setup as an editable dev dependency there.

performance-metrics/README.md
DerekMaggio: Correct, I will clarify a bit. Also, "dev-tainted" made me chuckle

performance-metrics/README.md
DerekMaggio: Oh yeah, I added the host and forgot to change that. Thank you

robot-server/Pipfile
SyntaxColoring: But `robot-server`'s dev environment is separate from `api`'s dev environment. Dev dependencies of `api` are not automatically installed as dev dependencies of `robot-server`.

After doing a fresh clone and `make setup`, I just tried:

```
cd robot-server
pipenv run python -c 'import performance_metrics'
```

And I got an `ImportError` because `performance_metrics` is not in `robot-server`'s dev environment. Even though this works:

```
cd api
pipenv run python -c 'import performance_metrics'
```

robot-server/robot_server/protocols/router.py
DerekMaggio: @SyntaxColoring 

> This looks like it'll work for now, but it seems like it's turning out that we want the trackers to work as context managers instead of decorators, right? In other words, we shouldn't have to turn a chunk of code into an inner function just for the sake of packaging it up and handing to track_getting_cached_protocol_analysis().

See https://github.com/Opentrons/opentrons/pull/14834#discussion_r1556175599 for a previous discussion about this. Let me know what you think.

DerekMaggio: > I also think you and @sfoster1 have been trying hard to avoid `performance_metrics` introducing even small overhead. I'm not sure if this is what we're doing now, but, for example, at one point you were checking the feature flag once on first import to avoid the overhead of an `if`/`else` every time the function under measurement was called. Is that still a goal?

@SyntaxColoring, you're correct that it should not be checking every time. But because `_should_track` is evaluated at import time you have to restart the robot-server if you ever want to update. Should I set the ff restart_required to True maybe?

robot-server/Pipfile
DerekMaggio: But I never directly import anything from performance_metrics in `robot-server`
All the wire up logic is in `api/util/performance_helpers.py` and `robot-server` imports stuff from there.

I tested performance metrics using the dev server for both ot-2 and flex initially

robot-server/robot_server/protocols/router.py
SyntaxColoring: Thank you for the background!

So to back up, you've gone to pains in the design of `performance_metrics` to avoid introducing overhead. But then it looks like the overhead is getting added here anyway?

Specifically—and I have not profiled this, but—if every call to `create_protocol()` defines an inner function `_get_cached_protocol_analysis()`, and then decorates it, and then calls it, I'm skeptical that all of that will be cheaper than it would have been to just enter and exit a context manager. Note that this particular decoration does *not* run at file parse time the way https://github.com/Opentrons/opentrons/pull/14834#discussion_r1556175599 describes. It runs every time this block of `_get_cached_protocol_analysis()` runs.

You could avoid this by refactoring the `_get_cached_protocol_analysis()` inner function to be a module-level private function, and then the decoration really would only run once at module parse time.

But my larger point, which I've broached before and which I still believe, is: Are these pains really worth it? Are simpler approaches, like entering context managers and accessing `os.environ` "too frequently," really not good enough? I mean, we're measuring protocol analysis here. The latency of starting the background thread alone is probably going to drown out whatever microseconds we shaved by doing the decorator thing. I believe there are use cases where it could matter, but I can easily live without them, personally.

robot-server/Pipfile
SyntaxColoring: > But I never directly import anything from performance_metrics in robot-server
All the wire up logic is in api/util/performance_helpers.py and robot-server imports stuff from there.

Yeah. But the behavior of `api/util/performance_helpers.py` depends on whether `performance_metrics` is installed. And I don't think it will be if we remove this line.

> I tested performance metrics using the dev server for both ot-2 and flex initially

Uhh hm, that's weird. Have you tried that on a fresh clone in a separate directory? I'm not sure why it would work for you but not me.

performance-metrics/setup.cfg
DerekMaggio: So the bdist_wheel option makes it compatible with python 2 and 3, which we don't need.
But our wheel naming logic in scripts/python_build_utils.py outputs a name with support for both python 2 and 3.
Python packaging is really not my forte so I don't want to mess with it. Especially not in this PR.

robot-server/Pipfile
DerekMaggio: https://github.com/Opentrons/opentrons/pull/15316/commits/44907382e38df0fd1c8ef3c970cc86b25fd5df9d

performance-metrics/README.md
DerekMaggio: https://github.com/Opentrons/opentrons/pull/15316/commits/5b8d813fccc6d263a8cae9fb50c230b70fcea1f0

performance-metrics/Makefile
DerekMaggio: https://github.com/Opentrons/opentrons/pull/15316/commits/5b8d813fccc6d263a8cae9fb50c230b70fcea1f0

SyntaxColoring (approved):
Yay, thank you! Looking forward to using it.

robot-server/robot_server/protocols/router.py
sfoster1: I think that I'm in both camps. I think the default approach, and the structural approach to implementing the reusable analysis side, should be oriented towards zero-cost abstractions even if they're inconvenient. But I think that at the point of integration, i.e. right here, sometimes it's okay to say "this is not zero-cost but the cost is marginal". For instance, here it's not zero-cost - like @SyntaxColoring says, you're creating an inner function each call - but this is not a hotpath; it's the top level call that will last a long time, not something called many times in a loop. We would jump through many more hoops than this to keep a zero-cost integration in some core protocol engine function, for a counterexample. And we need to have those hoops available to jump through, which is why we designed the metrics system as a function transformer that can run at import time.

I think it is reasonable to set restart_required to true for the feature flag for the general case.

DerekMaggio: To make sure I understand. Leave the creation of the inner function as is because it is not something we call a whole lot, and even if we did it is a call that takes a while so the impact would be negligible.
If we were wrapping some logic that is called many times then the wrapping of the logic should be done in a way that is as performant as possible.

PR #15317, opened 2024-06-03
fix(ai-server): huggingface and CORS

y3rsh:
The shape of CORS is giving me fits.  This works locally.

- Changed the live-test of OPTIONS to chat/completions like the UI requests pre-flight.
- Added call to validate huggingface simulate is working.


Elyorcv (approved):
LGTM

PR #15318, opened 2024-06-03
feat(shared-data): add definition for Deep Well Adapter for Temperature Module

jbleon95:
# Overview

Closes AUTH-71.

This PR adds a labware definition for the Opentrons 96 Deep Well Adapter for Temperature Module, along with the stacking overlap for the NEST 96 Deep Well Plate 2mL. 

# Test Plan

Ran this protocol on a Flex, verifying that the pipette goes to the relative correct location with no LPC offsets, and that the gripper can both pick up the Deep Well Plate from and to the adapter.

```
metadata = {
    'protocolName': 'Deep Well Temp Adapter Test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}

def run(context):
    trash = context.load_trash_bin('B3')
    tip_rack = context.load_labware('opentrons_flex_96_tiprack_200ul', 'B2')
    pipette = context.load_instrument("flex_1channel_1000", mount="left", tip_racks=[tip_rack])
    temp_deck = context.load_module('temperatureModuleV2', 'D1')

    adapter = temp_deck.load_adapter('opentrons_96_deep_well_temp_deck_adapter')
    deep_well = adapter.load_labware('nest_96_wellplate_2ml_deep')

    pipette.pick_up_tip()

    pipette.move_to(deep_well.wells()[0].top())
    context.pause()

    pipette.drop_tip()

    context.move_labware(deep_well, new_location='C2', use_gripper=True)
    context.move_labware(deep_well, new_location=adapter, use_gripper=True)
```

# Changelog

- added definition for the `opentrons_96_deep_well_temp_deck_adapter`
- added stacking overlap for the above to `nest_96_wellplate_2ml_deep`

# Review requests

The heater-shaker has a similar adapter that was already named `opentrons_96_deep_well_adapter` with no identifier in the name that it's for the heater-shaker. Unfortunately we can't go back on that, so I decided to name this new one `opentrons_96_deep_well_temp_deck_adapter` to at least make this one clear its for the temperature module. I chose "temp_deck" since its less wordy, but I'm open to other names.

# Risk assessment

Low.


shlokamin:
@shlokamin note to self: test PD, LL and the app's labware tab

jerader (commented):
I think we need to add this labware to `latestDefs` in `shared-data/js/labware.ts` in order for it to show up in app and LL

jerader (approved):
the definition looks good to me! can't test it in the FE until it is added to the list of labware defs though.

shlokamin:
> I think we need to add this labware to latestDefs in shared-data/js/labware.ts in order for it to show up in app and LL

ugh yeah thats a thing we have to do because all of our web apps bundle shared data rather than importing shared data as a raw artifact. let's do that in this PR, ill push up a commit now

ecormany (commented):
Looking at API load names and trying to have as little oddity as possible…

You can load a Temperature Module GEN1 with `tempdeck`, but GEN2 requires `temperature module gen2`. This matches our general tendency to phase out calling modules "deck" anywhere. I think I would prefer to keep "deck" out of this adapter name too.

So `opentrons_96_deep_well_temperature_module_adapter` would best match our naming scheme. It's a little long, though.

We do have an unofficial abbreviation convention that we use in the API docs, based on code snippets like `temp_mod = protocol.load_module(…`. So we could save a few characters by calling this `opentrons_96_deep_well_temp_mod_adapter`.

Either way, people will probably copy this out of docs or LL, so it's not a huge deal.

shlokamin:
@smb2268 this affects the quick transfer flow too. would you mind giving this a look?

PR #15319, opened 2024-06-03
fix(api): rip out unneeded arguments to liquid probe

ryanthecoder:
# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Looks good as a first step.

I think we may soon (maybe now) want to make there be two pressure sensor functions: one that is for production, and one that is for testing. The production one doesn't need any filepath arguments or anything, for instance, but we'd want to preserve that for testing.

PR #15321, opened 2024-06-03
feat(api): Prepare PipetteStore to handle overpressure errors from aspirate commands

SyntaxColoring:
# Overview

This goes towards EXEC-376.

# Test Plan

Just unit tests for now, because there's not much to test yet. I'll come up with more meaningful tests in follow-up PRs.

# Changelog

* Make a new defined error, `overpressure`.
* When an `aspirate` command fails with an `overpressure` error:
    * *Do* update where Protocol Engine thinks the pipette is.
    * *Do not* update the amount of liquid that Protocol Engine thinks is in the tip.
    
      I suspect this is not quite what we will want in the long term. We probably want to set it to either however much liquid got aspirated before the error, or to some special "unknown" sentinel. I propose we keep it simple for this PR and revisit it after this is wired up to PAPI and the hardware API.

The `aspirate` command does not actually return this new error yet, so none of this code is reachable in practice yet.

# Review requests

* Do the documented semantics for the new `OverpressureError` sound good?

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/pipetting_common.py
+++ b/api/src/opentrons/protocol_engine/commands/pipetting_common.py
@@ -1,6 +1,10 @@
 """Common pipetting command base models."""
+from dataclasses import dataclass
+from opentrons_shared_data.errors import ErrorCodes
 from pydantic import BaseModel, Field
-from typing import Optional
+from typing import Literal, Optional, TypedDict
+
+from opentrons.protocol_engine.errors.error_occurrence import ErrorOccurrence
 
 from ..types import WellLocation, DeckPoint
 
@@ -117,3 +121,41 @@ class DestinationPositionResult(BaseModel):
             " after the move was completed."
         ),
     )
+
+
+class OverpressureErrorInfo(TypedDict):  # noqa: D101
+    # Documentation for this is on the field in OverpressureError
+    # so it shows up in robot-server's OpenAPI spec.
+    volume: float
+
+
+class OverpressureError(ErrorOccurrence):
+    """Returned when sensors detect an overpressure error.
+
+    This is returned by commands that move the pipette plunger. The plunger motion is
+    stopped at the point of the error. After this error, It's safe to issue subsequent
+    `aspirate`, `dispense` etc. commands to pick up from where the error happened.
+    """
+
+    isDefined: bool = True
+
+    errorType: Literal["overpressure"] = "overpressure"
+
+    errorCode: str = ErrorCodes.PIPETTE_OVERPRESSURE.value.code
+    detail: str = ErrorCodes.PIPETTE_OVERPRESSURE.value.detail
+
+    errorInfo: OverpressureErrorInfo = Field(
+        ...,
+        description=(
+            "The amount, in µL, aspirated or dispensed by this command before it"
+            " encountered the overpressure error."
+        ),
+    )
+
+
+@dataclass(frozen=True)
+class OverpressureErrorInternalData:
+    """Internal-to-ProtocolEngine data about an OverpressureError."""
+
+    position: DeckPoint
+    """Same meaning as DestinationPositionResult.position."""
--- a/api/src/opentrons/protocol_engine/commands/pipetting_common.py
+++ b/api/src/opentrons/protocol_engine/commands/pipetting_common.py
@@ -2,7 +2,7 @@
 from dataclasses import dataclass
 from opentrons_shared_data.errors import ErrorCodes
 from pydantic import BaseModel, Field
-from typing import Literal, Optional, TypedDict
+from typing import Literal, Optional
 
 from opentrons.protocol_engine.errors.error_occurrence import ErrorOccurrence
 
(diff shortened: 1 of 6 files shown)


api/src/opentrons/protocol_engine/commands/pipetting_common.py
```diff
@@ -117,3 +121,41 @@ class DestinationPositionResult(BaseModel):
             " after the move was completed."
         ),
     )
+
+
+class OverpressureErrorInfo(TypedDict):  # noqa: D101
```
sfoster1: why `TypedDict` instead of a frozen dataclass?

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -301,27 +291,37 @@ def _update_current_location(self, command: Command) -> None:
...
-                labware_id=command.params.labwareId,
-                well_name=command.params.wellName,
+                pipette_id=action.command.params.pipetteId,
+                labware_id=action.command.params.labwareId,
+                well_name=action.command.params.wellName,
             )
-
-        elif isinstance(
-            command.result,
+        elif (
```
sfoster1, line 298: This feels a little awkward, because it's verbose but also because we're spreading logic about "is this a handleable error" to both the command (where it decides whether this is a defined error or not) and to the state handler. Can we localize it? Maybe have a predicate function in the command file like `def has_valid_location(action: SucceedCommandAction | FailCommandAction) -> bool`, or something? maybe even a bool on the action object?

api/src/opentrons/protocol_engine/commands/pipetting_common.py
SyntaxColoring: Because this is overriding the `.errorInfo` field on the base `ErrorOccurrence` model, and that field is a `dict`. But now that you bring this up, there's probably a better way of doing this. Like maybe the base `.errorInfo` field can be declared as a `BaseModel` or something.

api/src/opentrons/protocol_engine/commands/pipetting_common.py
```diff
@@ -117,3 +121,41 @@ class DestinationPositionResult(BaseModel):
...
+    volume: float
+
+
+class OverpressureError(ErrorOccurrence):
+    """Returned when sensors detect an overpressure error.
+
+    This is returned by commands that move the pipette plunger. The plunger motion is
+    stopped at the point of the error. After this error, It's safe to issue subsequent
+    `aspirate`, `dispense` etc. commands to pick up from where the error happened.
+    """
```
SyntaxColoring: 1. Discussed in-person: We actually *don't* want people to be able to `aspirate` or `dispense` after an overpressure error. The valid operations should be `blowOut`, maybe `home`, and anything that doesn't touch the plunger.
2. We might want to call this something more specific to indicate that these semantics only apply to overpressure events *while plungering liquid.* We could also use the pressure sensor to detect collisions, and we would want to make that a different defined error.

api/src/opentrons/protocol_engine/commands/pipetting_common.py
```diff
@@ -117,3 +121,41 @@ class DestinationPositionResult(BaseModel):
...
+    errorCode: str = ErrorCodes.PIPETTE_OVERPRESSURE.value.code
+    detail: str = ErrorCodes.PIPETTE_OVERPRESSURE.value.detail
+
+    errorInfo: OverpressureErrorInfo = Field(
+        ...,
+        description=(
+            "The amount, in µL, aspirated or dispensed by this command before it"
+            " encountered the overpressure error."
+        ),
+    )
```
SyntaxColoring: Discussed in-person:

Getting this volume data is going to create extra work to get the hardware API to return it, and it's also not necessary if `aspirate` and `dispense` aren't valid follow-ups. We should remove this.

api/src/opentrons/protocol_engine/state/pipettes.py
SyntaxColoring: Yeah, I *think* we're bothered by the same thing.

I would like to be able to write something roughly like this:

```python
class AspirateImplementation(...):
    ...

    async def execute(
        self, params: AspirateParams
    ) -> ExecuteResult:
        ...

        deck_coords = await self._movement.move_to_well(...)

        try:
            volume = await self._pipetting.aspirate_in_place(...)
        except PipetteOverpressureError:
            volume = VOLUME_UNKNOWN

        return ExecuteResult(
            public=(
                OverpressureError(...) if volume == VOLUME_UNKNOWN
                else AspirateResult(...),
            ),
            pipette_state_update=PipetteStateUpdate(
                new_position=PositionUpdate(
                    deck_coords=deck_coords,
                    well=params.well,
                    pipette_id=params.pipette_id
                ),
                new_volume=VolumeUpdate(
                    volume=volume,
                    pipette_id=params.pipette_id,
                )
            ),
            # tip_state_update=None
            # labware_state_update=None
            # etc.
        )
```

Would that address your concerns?

Basically, breaking down high-level command information into low-level state updates is currently *mostly* the job of the state stores, and this is painful, and I think I would like to solve the pain by making the commands more responsible for it.

api/src/opentrons/protocol_engine/errors/error_occurrence.py
```diff
@@ -118,7 +118,7 @@ def from_failed(
         ),
     )
 
-    errorInfo: Dict[str, str] = Field(
+    errorInfo: Mapping[str, object] = Field(
```
sfoster1: `object` seems a little weird here. doesn't this have to be serializable?

api/src/opentrons/protocol_engine/state/pipettes.py
sfoster1: Yes that would definitely be better to me

sfoster1 (approved):
Okay, looks good. I think we're both annoyed by the same thing about the new state subhandlers and we can fix it over time, it's definitely still better than duplicating even more of the code. Nice!

api/src/opentrons/protocol_engine/errors/error_occurrence.py
SyntaxColoring: Yyyes, that is a good point. Reverting this from this PR. I'll broaden this type in some better way in a follow-up, if we need to.

PR #15323, opened 2024-06-03
feat(api,robot-server,shared-data): Specify tip overlap by version

sfoster1:
We need to have multiple versions of the tip overlap data for at least flex pipettes so that we can ship updated overlap data without inadvertently changing the behavior of protocols on stable API versions. This PR accomplishes that.

It's probably best viewed commit-by-commit:
- https://github.com/Opentrons/opentrons/commit/74a25f9f2ff6df718c38e243831a91426835d81c is the shared-data changes; it changes the jsonschema, changes the data (this was a script, so if one file looks good they all look good, and they schema match) , changes the py types, changes the js types, changes the tests, and establishes some constants
- https://github.com/Opentrons/opentrons/commit/df12f287954c8012071b8fa60dc91cf07fde8038 is the meat of the work; this changes the engine and the hardware controller to load, present, and select the tip overlaps by some version.
- https://github.com/Opentrons/opentrons/commit/63603aa801e7da9b23551da6889326765934c6f2 adapts the robot server; only the OT-2 calibration code needed to be changed.

# Details
## separate versioning
The data in shared-data has a version, but that version is separate from versions in other parts of the stack. This is on purpose, because the version has to be written in shared-data, and read in protocol engine. I think it's gross to write API versions in shared data, but we've done it before and could do it again; the real sticking point was the engine. The engine, in modern protocols, loads the tip overlap and does all that data management itself; it would need to select the appropriate version from the dict. We really don't want API versions there.

Therefore, there's a separate simple versioning scheme that's a little more like the alternate liquid handling function scheme that we haven't used in years, but with an actual string key for the version instead of implicit versioning by positioning in an array.

There will instead be a specific map of API versions to overlap versions living inside the source of the engine core for the python protocol API (that's the only place it needs to be - for why, read on).

## engine state management and data lifecycle
There were basically two ways to implement this. The first way is to save all the data - save the full versioned tip overlap dictionary, and save which version we've selected. Since I listed this first, you can guess I didn't do this. Instead, what I did was add a version to the two engine commands that load pipette data - `load_pipette` and `configure_for_volume` - and then have the functions those commands call pull the specified version out of the full config, which means that we only have to save the specific version in state and state doesn't have to change. This also means that we don't have to worry about data storage migration and compatibility.

One huge note here: **the engine argument parsing gives you the MOST RECENT tip overlap if you don't specify one**. That's in bold because it's not the way we've implemented similar things. It's necessary because we want PD protocols to get the new overlap values without having to change anything.

## python api integration
The python API, however, really wants to keep the old values (until you're on a certain version - we'll implement that when we actually add the new values). We do that in a couple ways:
- the old pipette dictionary value that has the tip overlap dict is still there, using "v0". That means that protocol API versions that use that dict, which is all those not using the engine core, get the old values consistently for free
- for protocol api versions that do use the engine core, for now we hard specify v0.

## json protocol integration

New JSON protocols, any that directly create engine commands rather than being mapped through python protocol API values, will get the new overlaps because they won't specify the parameter to the load pipette or configure for volume engine commands. Older JSON protocols that do adapt the python protocol API commands will get the python behavior.

# Testing
- This PR gets regression tested - it doesn't really contain any new features, because it doesn't really contain any new data.
  - OT-2 protocol behavior should be adequately covered by the g-code testing
  - OT-2 calibration should probably be run real quick
  - Flex behavior should be adequately tested by running LPC and a protocol - this applies to all protocols and can be easily tested.
  - Flex calibration and LPC doesn't use tips and therefore isn't affected

Closes EXEC-451


shared-data/pipette/schemas/2/pipetteLiquidPropertiesSchema.json
```diff
@@ -106,15 +106,23 @@
...
-      "properties": {
-        "default": { "type": "number" }
+      "description": "Map by version of tip overlap values with defaults",
+      "required": ["v0"],
+      "patternProperties": {
+        "^v\\d+$": {
+          "type": "object",
+          "description": "Map of tip overlap values with defaults",
+          "required": ["default"],
+          "$comment": "Other keys in here should be labware URIs",
```
SyntaxColoring, line 118: Have we ever considered YAML or JSON5 for shared-data?

robot-server/robot_server/robot/calibration/pipette_offset/user_flow.py
```diff
@@ -328,7 +328,7 @@ def _get_stored_pipette_offset_cal(
     def _get_tip_length(self) -> float:
         stored_tip_length_cal = self._get_stored_tip_length_cal()
         if stored_tip_length_cal is None or self._should_perform_tip_length:
-            tip_overlap = self._hw_pipette.tip_overlap.get(self._tip_rack.uri, 0)
+            tip_overlap = self._hw_pipette.tip_overlap["v0"].get(self._tip_rack.uri, 0)
```
SyntaxColoring, line 331: Why do we want these OT-2 calibration flows to use the oldest data instead of the newest data?

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
```diff
@@ -28,6 +28,13 @@ class ConfigureForVolumeParams(PipetteIdMixin):
         "than a pipette-specific maximum volume.",
         ge=0,
     )
+    tipOverlapNotAfterVersion: Optional[str] = Field(
```
SyntaxColoring, line 31: From a Protocol Engine caller's perspective, does it make sense for `configureForVolume` to deal with tip overlaps? Doesn't this command only affect plunger stuff?

api/src/opentrons/protocol_engine/commands/load_pipette.py
```diff
@@ -52,6 +52,13 @@ class LoadPipetteParams(BaseModel):
         description="An optional ID to assign to this pipette. If None, an ID "
         "will be generated.",
     )
+    tipOverlapNotAfterVersion: Optional[str] = Field(
+        None,
+        description="A version of tip overlap data to not exceed. The highest-versioned "
+        "tip overlap data that does not exceed this version will be used. Versions are "
+        "expressed as vN where N is an integer, counting up from v0. If None, the current "
+        "highest version will be used.",
+    )
```
SyntaxColoring, line 61: This is a nit that doesn't really matter, but since these are always integers, should we make this field an `int` instead of a `str`?

I guess we're using `"vN"` strings for readability in the underlying JSON files? I'm 50/50 on whether it's better to use a `str` here to be consistent with those, or to translate to an `int` for API niceness.

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
```diff
@@ -10,16 +11,40 @@
...
 )
 
+from ..errors.exceptions import InvalidLoadPipetteSpecsError
 from ..types import FlowRates
 from ...types import Point
 
+_TIP_OVERLAP_VERSION_RE = re.compile(r"^v\d+$")
+
+
+def validate_and_default_tip_overlap_version(version_spec: Optional[str]) -> str:
```
SyntaxColoring, line 27: I don't know if this is a good idea, but we could make a custom Pydantic type for `TipOverlapVersion`, or even `VString` if we want to be generic.

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
```diff
@@ -226,3 +259,26 @@ def get_pipette_static_config(pipette_dict: PipetteDict) -> LoadedStaticPipetteD
             front_right_offset[0], front_right_offset[1], front_right_offset[2]
         ),
     )
+
+
+def get_latest_tip_overlap_before_version(
+    overlap: Dict[str, Dict[str, float]], version: str
+) -> Dict[str, float]:
+    """Get the latest tip overlap definitions that are equal or older than the version."""
+    # TODO: make this less awful
```
SyntaxColoring, line 268: Again, I don't know if this is a good idea, but if we had a custom `TipOverlapVersion`/`VString` Pydantic type, we could overload the comparison operators on it for easier sorting.

api/src/opentrons/protocol_runner/legacy_command_mapper.py
```diff
@@ -690,7 +690,10 @@ def _map_instrument_load(
             pipette_id=pipette_id,
             serial_number=serial,
             config=pipette_data_provider.get_pipette_static_config(
-                instrument_load_info.pipette_dict
+                # Compatibility note - this is the version of tip overlap data, it stays at 0
+                # so protocol behavior does not change when you run a legacy JSON protocol
+                instrument_load_info.pipette_dict,
+                "v0",
```
SyntaxColoring, line 696: As far as I know, the stuff in here only affects how the protocol activity gets reported to the frontend. It doesn't affect actual robot behavior. So this string could be arbitrary, though I agree v0 is most correct.

SyntaxColoring (approved):
Thanks, this approach looks great. Just some more minor code-level comments.

robot-server/robot_server/robot/calibration/pipette_offset/user_flow.py
sfoster1: It's kind of a coinflip honestly - it would make it consistent with PD protocols and not older python protocols, or vice versa. Either would be unacceptable. We don't ever intend to update tip length data for OT-2 pipettes, but if we ever did we'd have to store it separately for different versions.

api/src/opentrons/protocol_engine/commands/configure_for_volume.py
sfoster1: It doesn't, really, certainly not with what's currently implemented. It is a minor sin.

api/src/opentrons/protocol_engine/commands/load_pipette.py
sfoster1: I want to keep it a string to promote "this is an arbitrary value please do not do math on it"

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
sfoster1: I'd much rather have kind of gross validation be in code we write, honestly. Maybe if we had a second use case.

api/src/opentrons/protocol_engine/resources/pipette_data_provider.py
sfoster1: Yeah but that would just wash the awfulness around. Ultimately, the only thing that would fix it is using just numbers and I want to keep them hard to do math with.

PR #15324, opened 2024-06-04
fix(opentrons-ai-client): rendering single and triple backticks

Elyorcv:
This PR attempts to render single and triple backticks correctly. To see the difference see below please: before vs after.
If a user's message contains a backtick then it is rendered as a paragraph which is a block element and creating unnecessary extra line and font size is not well aligned with respect to surrounding content.

Before
image

After
image

# Overview

# Test Plan
Click examples on the left panel and send and see in ChatDisplay

# Changelog

# Review requests

# Risk assessment
Low


y3rsh (approved):
They look good locally in my testing.
image

PR #15325, opened 2024-06-04
Add Jira Ticket Assignee

rclarke0:
# Overview

Assigns jira ticket to user during creation process

# Test Plan

Tested with robot with error.

# Changelog

reads all issues from the board
extracts all existing assignee and reporter
matches up user inputted display name with account id. 

# Review requests

# Risk assessment


sfoster1 (approved):
Nice

PR #15326, opened 2024-06-04
fix(app): filter out adapters from quick transfer labware

b-cooper:
# Overview

Because adapters cannot be pipetted into or out of directly, filter them out in the pipette compatibility list caching utility of the Quick Transfer flow

# Risk assessment
low


app/src/organisms/QuickTransferFlow/constants.ts
```diff
@@ -76,31 +76,24 @@ export const SINGLE_CHANNEL_COMPATIBLE_LABWARE = [
...
-  'opentrons/opentrons_96_well_aluminum_block/1',
   'opentrons/opentrons_96_wellplate_200ul_pcr_full_skirt/2',
-  'opentrons/opentrons_aluminum_flat_bottom_plate/1',
   'opentrons/opentrons_flex_96_filtertiprack_1000ul/1',
   'opentrons/opentrons_flex_96_filtertiprack_200ul/1',
   'opentrons/opentrons_flex_96_filtertiprack_50ul/1',
   'opentrons/opentrons_flex_96_tiprack_1000ul/1',
   'opentrons/opentrons_flex_96_tiprack_200ul/1',
   'opentrons/opentrons_flex_96_tiprack_50ul/1',
-  'opentrons/opentrons_flex_96_tiprack_adapter/1',
```
jerader, line 102: isn't this needed if the user selects a 96-channel pipette?

shlokamin (approved):
based on what i know about the quick transfer flow this makes sense to me!

app/src/organisms/QuickTransferFlow/constants.ts
b-cooper: I believe these lists are only generated for selecting source and destination labware. If a user chooses the 96 channel, we will include a 96 tip rack adapter underneath that tip rack no matter what.

PR #15327, opened 2024-06-04
style(app, protocol-designer): promote confusing void expression lint rule to error, fix all

b-cooper:
# Overview

Re-enable lint error for confusing typescript void expressions and fix all outstanding occurrences of
the error.

# Risk assessment
medium, many files changed


protocol-designer/src/components/IngredientsList/index.tsx
b-cooper, line 1: There appears to be a bug here that would make keep removeWellContents from ever being dispatched correctly. @shlokamin

shlokamin: cc @jerader for viz. thanks for fixing brian!

shlokamin (approved):
thanks for doing this! clicked around the app, pd, and ll and nothing bad seemed to happen

PR #15331, opened 2024-06-04
test: add developer validation version of smoke test protocol

DerekMaggio:
# Overview

Closes RQA-2801

Add the ability to define different configurations of our smoke test protocol to allow developers an easy-to-run preconfigured version of our test. 

All a developer needs to do is select the `Developer Validation` configuration for the `Test Configuration` runtime parameter

# Test Plan

- [ ] Get someone to run it on a Flex and help me tune the protocol

# Changelog

- Defined some dataclasses inside the test protocol to pull out the configuration properties of the protocol
- Defined a `dev_configuration` and `full_configuration`
  - dev is for devs
  - full is for QA  

# Review requests

Is there a better way to handle this configuration logic than just inline in the protocol file? This should really be some kind of importable package but that seems like a decently large lift

# Risk assessment

None

### DIFF
--- a/app-testing/files/protocols/Flex_S_v2_18_P1000_96_GRIP_HS_MB_TC_TM_Smoke.py
+++ b/app-testing/files/protocols/Flex_S_v2_18_P1000_96_GRIP_HS_MB_TC_TM_Smoke.py
@@ -346,7 +346,7 @@ def add_parameters(parameters: protocol_api.Parameters):
 
     parameters.add_bool(
         variable_name="prefer_gripper_disposal",
-        display_name="I love to refill tip racks",
+        display_name="I LOVE TO REFILL TIP RACKS",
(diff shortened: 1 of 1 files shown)


y3rsh (commented):
I think this is still way too many repeated aspirates, dispenses, and moves in the dev version. 
We should cut it down to 1 of a thing unless there is a specific reason in the dev version.
It even seems overkill for the full version unless there are specific reasons I am missing.
image

image

DerekMaggio:
@y3rsh, I agree. 
I must have messed something up because I thought it was doing way less labware movement and pipetting than this.

DerekMaggio:
Shortened normal qa protocol to ~220 steps from ~350 
Shortened dev protocol to 63 steps

y3rsh (approved):
:shipit:

PR #15332, opened 2024-06-04
feat(api): add csv parameter definitions

jbleon95:
# Overview

Closes AUTH-417.

This PR adds a new `add_csv_file` method to the existing `parameter_context` along with a new `CSVParameterDefinition` object to represent a required CSV parameter and hold data that will be passed to both analysis/run results and the protocol context. Currently basic information about display name, variable name, and description is being passed to analysis, the parameter cannot be set or used in the protocol in this PR.

Adding the new `CSVParameterDefinition` required some refactoring in the existing `ParameterDefinition` related code. Because this type of parameter is different from the more basic data types (which I'm referring to in some places as primitive runtime parameters), it didn't make sense to try and jam that into the existing `ParameterDefinition` class. Neither did it feel right to make a class that was similar but had no connection to that. Instead, a new abstract interface was created with generics in a similar way to how we've architected `protocol_api` classes between the engine and the legacy implementations.

As of this PR, the CSV definition will store the file handler as a `TextIO` type, along with the UUID that will be generated for it as part of the robot server file handling interface that will be coming in a future PR. By default, these will both be `None`, which if a run is analyzed with these values not set, will lead to a `ParameterFileRequired` (or some similar name) result for analysis. As of now, the protocol will analyze the `run` function without this being set, but the parameter cannot be accessed at all.

We are also reporting the parameter in analysis. More fields may be added as deemed necessary, but for now the basic skeleton is being reported.

# Test Plan

Tested that this protocol is able to to be analyzed and shows the parameter in analysis.

```
metadata = {
    'protocolName': 'CSV add parameter test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}


    parameters.add_int(
        display_name="Sample count",
        variable_name="sample_count",
        default=2,
        minimum=1,
        maximum=6,
    )
    parameters.add_str(
        display_name="Pipette Name",
        variable_name="pipette",
        choices=[
        ],
        default="flex_1channel_50",
    )
    parameters.add_csv_file(
        display_name="CSV Data",
        variable_name="csv_data",
        description="CSV file containing labware and volume information."
    )



def run(context):
    SAMPLE_COUNT = context.params.sample_count
    PIPETTE_NAME = context.params.pipette

    pipette = context.load_instrument(PIPETTE_NAME, mount="left", tip_racks=[tip_rack])

    for i in range(SAMPLE_COUNT):

        pipette.pick_up_tip()
        pipette.aspirate(50, source_well)
        pipette.dispense(50, destination_well)
        pipette.drop_tip()
```

relevant section of the analysis (other parameters not shown below)
```
  "runTimeParameters": [
    ...,
    {
      "displayName": "CSV Data",
      "variableName": "csv_data",
      "description": "CSV file containing labware and volume information.",
      "type": "csv_file"
    }
  ],
```

# Changelog

- Added a new `CSVParameterDefinition` that can be added via `ParameterContext.add_csv_file`
- Refactored existing `ParameterDefinition` to inherit from a new `AbstractParameterDefinition`
- Report new CSV parameter in analysis

# Review requests

Refactor-wise, we are losing a little bit of type checking in ensuring the str, bool, float and int correctness of the old parameters, but this is specific to our codebase and being enforced in code at runtime (and in unit/integration tests). I think the new abstract interface gives more flexibility in the future, but if there are other thoughts on how to organize this now is the time.

Additionally, while this feature is going to be for CSV files only for the next release, feedback on the names of the new classes/types added is welcome. There is a case to be made for making the names of the non-userfacing methods/classes more generic.

# Risk assessment

Lowish-medium, the refactor work is mostly mypy specific and functionally identical, and the CSV parameter is not touching much of the existing infrastructure in API and robot server right now.


api/src/opentrons/protocol_engine/types.py
```diff
@@ -967,7 +967,16 @@ class EnumParameter(RTPBase):
...
 
 
-RunTimeParameter = Union[NumberParameter, EnumParameter, BooleanParameter]
+class CSVParameter(RTPBase):
+    """A CSV file parameter defined in a protocol."""
+
+    id: Optional[str] = Field(
+        ...,
+        description="The UUID identifier of the file stored on the robot. For local analysis this will be empty.",
+    )
```
sanni-t: From the in-person chat: we probably want this to be something like-

```suggestion
    file: Optional[FileId] = Field(
        ...,
        description="The CSV file stored on the robot, to be used as the CSV RTP override value. For local analysis this will be empty.",
    )
```
where `FileId` is:

```
class FileId(BaseModel):
    """A CSV file UUID descriptor."""

    id: Optional[str] = Field(
        ...,
        description="The UUID identifier of the file stored on the robot. For local analysis this will be empty.",
    )
```

api/src/opentrons/protocols/parameters/csv_parameter_definition.py
```diff
@@ -0,0 +1,73 @@
...
+        Arguments:
+            display_name: The display name of the parameter as it would show up on the frontend.
+            variable_name: The variable name the parameter will be referred to in the run context.
+            description: An optional description for the parameter.
+        """
+        self._display_name = validation.ensure_display_name(display_name)
+        self._variable_name = validation.ensure_variable_name(variable_name)
+        self._description = validation.ensure_description(description)
+        self._value: Optional[TextIO] = None
+        self._id: Optional[str] = None
```
sanni-t, line 28: This is good for this PR, but we probably want something like a `CsvFileResource` that contains all metadata (id, name, createdAt, etc) instead of just saving the ID.

api/tests/opentrons/protocol_api/test_parameter_context.py
```diff
@@ -9,6 +9,7 @@
     MAX_SUPPORTED_VERSION,
 )
 from opentrons.protocols.parameters import (
+    csv_parameter_definition as mock_cvs_parameter_definition,
```
sanni-t:
```suggestion
    csv_parameter_definition as mock_csv_parameter_definition,
```
..unless you want to call it CVS definition :P

api/src/opentrons/protocol_engine/types.py
sanni-t: Also, let's add the `type` as `Literal["csv_file"]`

sanni-t (approved):
This is really good start. We might later update some typing and shape of csv data being communicated between server & API, once we discuss the shape of this data. This LGTM!

api/tests/opentrons/protocol_api/test_parameter_context.py
jbleon95: 🤦

PR #15335, opened 2024-06-05
chore(shared-data): Adds new P1000S model v3.7 to support new accuracy function

andySigler:
# Overview

Production team is almost done validating a new accuracy function for the P1000S that will improve our %D yield. This function may be ready next week.

In the mean time, we can merge in the new model number so that the new function doesn't hold up QA'ing this new model.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
LGTM

PR #15336, opened 2024-06-05
fix(opentrons-ai-client): use babel plugins in correct environments

shlokamin:
# Overview

This PR updates our build configs to use a babel plugin for styled components in the staging env, and adds some css plugins for stuff used in the components library. This *should* fix the whitescreen issue we were seeing in staging and prod.

# Risk assessment

Low


koji (approved):
thank you for fixing the white screen issue

Elyorcv (approved):

PR #14227, opened 2023-12-15
fix(app): fix buttons disabled conditions when estop is not disengaged

koji:
# Overview
Add a new hook to check estop status to control buttons' disabled on Desktop app.
This might go into `edge`.

~~Currently the target is `7.1.0` release branch but this fix was moved to 7.2.0 so the target will be switched to `edge` one `7.1.0` is merged into `edge`. Probably I will need to solve conflicts since this PR changes a lot of tests.~~
https://opentrons.slack.com/archives/C05R5390YR3/p1702938902854579

All buttons are disabled (got confirmation from Mike and Rob)
Restart robot, RobotSettings in RobotOverflowmenu , and the toggle button in Privacy tab are enabled when an e-stop button is pressed.

close RQA-2110

related thread
https://opentrons.slack.com/archives/C05R5390YR3/p1702669338799859?thread_ts=1702586313.100579&cid=C05R5390YR3

# Test Plan
### Flex
1. Go to Device Details page
2. Press the robot's e-stop button

### OT-2
The above disabled buttons are not disabled.

# Changelog

# Review requests

# Risk assessment

PR #15337, opened 2024-06-05
fix(app): Allow using update-from-file button when robot-server is down

SyntaxColoring:
# Overview

Fixes EXEC-506.

# Test Plan

* [x] See test plan in EXEC-506.

# Details

The app disables many buttons, including the update-from-file button, when the E-stop is "not disengaged." (See e.g. PR #14227.)

When robot-server is down, the endpoint to query the E-stop returns an error, of course. The app [interprets that](https://github.com/Opentrons/opentrons/blob/7631b78a4f8622cee17d17c55269401205c62f0e/app/src/resources/devices/hooks/useIsEstopNotDisengaged.ts#L24) as "not disengaged," which seems like a reasonable choice in general. But that disables the update-from-file button when robot-server is down, which is unfortunate because it makes it difficult to recover a borked robot.

My fix here is to just totally decouple the update-from-file button from the E-stop query. I leave the rest of the E-stop button-disabling logic alone.

# Review requests

* Is this the best way to solve this? Or should we treat E-stop query errors as "disengaged" in general? 
* Do we need to update any design, product, or QA resources to reflect this change, to make it less likely that we'll regress on this again in the future?

# Risk assessment

Low.

### DIFF
--- a/app/src/organisms/Devices/RobotSettings/RobotSettingsAdvanced.tsx
+++ b/app/src/organisms/Devices/RobotSettings/RobotSettingsAdvanced.tsx
@@ -207,7 +207,7 @@ export function RobotSettingsAdvanced({
         <Divider marginY={SPACING.spacing16} />
         <UpdateRobotSoftware
           robotName={robotName}
-          isRobotBusy={isRobotBusy || isEstopNotDisengaged}
+          isRobotBusy={isRobotBusy}
           onUpdateStart={() => handleUpdateBuildroot(robot)}
         />
         {isFlex ? (


mjhuff (approved):
Per in-person conversation, this makes sense. There shouldn't be a coupling between the update server and the e-stop, and adding any "unknown" e-stop state promotes that coupling. Given the context that the e-stop was added in under a tight timeframe, this was likely just an oversight.

Worst happens, we can revisit this if it poses an issue in the future. Thank you for fixing!

PR #15338, opened 2024-06-05
fix(app): Allow downloading troubleshooting logs when E-stop is engaged

SyntaxColoring:
# Overview

Fixes EXEC-507.

# Test Plan

* [x] See test plan in EXEC-507.

# Changelog

Straightforwardly, do not disable the "Download logs" button when the E-stop is engaged.

# Review requests

* Any reason not to do this?
* For similar reasons why we're doing this, should we *also* keep the "Device reset" button enabled?
* Do we need to update any design, product, or QA resources to reflect this change?

# Risk assessment

Low.


sfoster1 (approved):
Yeah this is nice. I have no opinion about `reset robot` , i could think of arguments in either direction, so not changing it is fine.

PR #15342, opened 2024-06-05
fix(shared-data): P1000S v3.7 can use latest tip-overlap values

andySigler:
# Overview

Since this pipette model hasn't been shipped to customers yet, it is safe to update the tip-overlap values in the definition to match our latest data.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Nice!

PR #15349, opened 2024-06-05
feat(api): implement multi liquid probe

ryanthecoder:
# Overview

The Z distance a liquid probe can cover is not only limited by the max_z_distance argument but also by how far the Z can move at the specified mount speed in the time it takes the pipette to travel from home->bottom or bottom->home at the specified plunger speed.

Instead of throwing a LiquidNotFoundError earlier than the max_z_distance specified, we can now handle running a single pass over and over until the entire max_z_distance is covered and only throw an error if we didn't find liquid in that entire distance.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -2626,7 +2626,6 @@ class OT3API(
         probe_start_pos = pos._replace(z=probe_settings.starting_mount_height)
         await self.move_to(mount, probe_start_pos)
         total_z_travel = probe_settings.max_z_distance
-        found = False
         z_travels = self._get_probe_distances(
             checked_mount,
             total_z_travel,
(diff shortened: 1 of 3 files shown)


api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2591,44 +2625,58 @@ async def liquid_probe(
...
             probe_settings.mount_speed,
-            (probe_settings.plunger_speed * plunger_direction),
-            probe_settings.sensor_threshold_pascals,
-            probe_settings.output_option,
-            probe_settings.data_files,
-            probe=probe if probe else InstrumentProbeType.PRIMARY,
         )
-        end_pos = await self.gantry_position(mount, refresh=True)
+        error: Optional[LiquidNotFoundError] = None
+        while len(z_travels) > 0 and not found:
```
sfoster1: this separated `while len() > 0 ...  thing.pop()` is a little unclear and unnecessary in python. Let's make this `for z_travel in z_travels:` with a `break` when we find the target

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2591,44 +2625,58 @@ async def liquid_probe(
...
+
+                height = await self._liquid_probe_pass(
+                    mount,
+                    probe_settings,
+                    probe if probe else InstrumentProbeType.PRIMARY,
+                    z_travels[0],
+                )
+                # if we made it here without an error we found the liquid
+                found = True
+            except LiquidNotFoundError as lnfe:
```
sfoster1, line 2667: if this is the only exception we care about, let's put the `try` just above the `await  self._liquid_probe_pass` call

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2591,44 +2625,58 @@ async def liquid_probe(
...
+                z_travels.pop(0)
+                if len(z_travels) == 0:
+                    # we have no more distance to go so stash the error so we can raise it after moving out of the well
+                    error = lnfe
+            except Exception as e:
+                # Unexpected error occured, raise it up
+                raise e
```
sfoster1: seems like it doesn't do anything

sfoster1 (changes requested):
Style changes

sfoster1 (approved):
Nice, thank you!

PR #15350, opened 2024-06-05
test(snapshot): major improvements

y3rsh:
# Analyses Snapshot Improvements

GitHub cannot handle the diff so review locally.

- [x] change the package name from app-testing -> analyses-snapshot-testing
- [x]  add 2.18 protocols
- [x]  add 2.19 protocols
- [x] add snapshots for the newly added protocols (2.18, 2.19, Protocol Library Protocols)
- [x] alter all snapshots now that custom labware only shows up in the snapshot if the protocol has custom labware
- [x] remove the use of .env
- [x] improve Makefile and define all environment variables explicitly there
- [x] re-write README
- [x] create a custom snapshot extension
- [x] replace parts of strings that cause unnecessary diffs like traceback line and moduleId exposed in a detail string
- [x] add back key as it should not change
- [x] replace the value of all uuid fields with the string 'UUID'
- [x] replace the value of all timestamp fields with 'TIMESTAMP'


.eslintignore
```diff
@@ -28,9 +28,9 @@ robot-server/**
 shared-data/python/**
 hardware-testing/**
 
-# app-testing don't format the json protocols
-app-testing/files
+# analyses-snapshot-testing don't format the json protocols
+analyses-snapshot-testing/files
 # app testing don't format the snapshots
```
y3rsh:
```suggestion
# don't format the snapshots
```

.github/workflows/analyses-snapshot-test.yaml
```diff
@@ -28,57 +28,57 @@ jobs:
...
     - name: Create Snapshot update Request
       id: create-pull-request
       if: failure()
       uses: peter-evans/create-pull-request@v5
       with:
-          commit-message: 'fix(app-testing): snapshot failure capture'
-          title: 'fix(app-testing): snapshot failure capture'
+          commit-message: 'fix(analyses-snapshot-testing): snapshot failure capture'
+          title: 'fix(analyses-snapshot-testing): ${{ env.ANALYSIS_REF }} snapshot failure capture'
           body: 'This PR is an automated snapshot update request. Please review the changes and merge if they are acceptable or find you bug and fix it.'
```
y3rsh:
```suggestion
          body: 'This PR is an automated snapshot update request. Please review the changes and merge if they are acceptable or find your bug and fix it.'
```

.github/workflows/analyses-snapshot-test.yaml
```diff
@@ -3,12 +3,12 @@ name: Analyses Snapshot Test
 on:
   workflow_dispatch:
     inputs:
-      TARGET:
-        description: 'Target branch or tag'
+      ANALYSIS_REF:
+        description: 'Branch or tag that provides the analysis output at test runtime'
```
DerekMaggio, line 7: nit: Add something like `what ref you want to compare snaphsot results to`

.github/workflows/analyses-snapshot-test.yaml
```diff
@@ -3,12 +3,12 @@ name: Analyses Snapshot Test
...
         required: true
         default: 'edge'
-      TEST_SOURCE:
-        description: 'Target for the test code'
+      SNAPSHOT_REF:
+        description: 'Branch or tag that provides the snapshot and test code at test runtime'
```
DerekMaggio, line 11: nit: `what ref you want to use to run code and generate snapshots with`

This is something that has always confused me every time I go to run something. So being a bit verbose would be helpful here.

If you can, you might provide a link to the README

DerekMaggio (approved):
LGTM!

PR #15351, opened 2024-06-05
feat(api): implement liquid probe protocol engine command

pmoegenburg:
# Overview

This closes EXEC-436. Pulls well dimension data from Protocol Engine StateView, which is now used for process `max_z_distance`. Also bubbles up LiquidNotFoundError.

# Test Plan

Added protocol engine command test.

# Changelog

# Review requests

# Risk assessment

Low. Largely builds off existing Aspirate protocol engine command.


sfoster1 (changes requested):
Aside from fixing what's already failing tests, please either draft this or take a pass to remove note to self comments and write some tests. They can be pretty simple for now since this command doesn't do much, but they should be present.

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -0,0 +1,87 @@
...
+
+from opentrons.types import MountType
+
+LiquidProbeCommandType = Literal["liquidProbe"]
+
+
+class LiquidProbeParams(BaseModel):
+    """Payload required to liquid probe."""
+
+    mount: MountType = Field(..., description="Instrument mount to liquid probe with.")
```
ryanthecoder: We are going to want the well we are using so we can get the well's max depth

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -0,0 +1,87 @@
...
+        # assumption: scope discludes moving to process starting position
+        # account for labware (height)?
+
+        ot3_api = ensure_ot3_hardware(self._hardware_api)
+        ot3_mount = OT3Mount.from_mount(params.mount)
+        assert (
+            ot3_mount is not OT3Mount.GRIPPER
+        ), "Expected a Pipette mount but Gripper mount was provided."
+
+        z_pos = ot3_api.liquid_probe(mount=ot3_mount)  # anything else?
```
ryanthecoder: We are going to want to pass an Instance of LiquidProbeSettings.
we can get a copy of the defaults from ot3_api.config().liquid_sense
We are going to want to change the starting mount height and the max_z_distance based on the well we are probing

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -0,0 +1,116 @@
...
+        well_def = self._movement._state_store.labware.get_well_definition(
+            labware_id, well_name
+        )
+        well_depth = well_def.depth
+        z_pos = await self._pipetting.liquid_probe_in_place(
+            pipette_id=pipette_id, max_z_dist=well_depth
+        )
+
+        return SuccessData(
+            public=LiquidProbeResult(
```
caila-marashaj: I think it might just be worth a comment somewhere that the `position` here is the position at the well where the probe starts, and `z_pos` is after the probe

caila-marashaj (approved):
Nice!

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2622,21 +2623,21 @@ async def liquid_probe(
         if not probe_settings:
             probe_settings = self.config.liquid_sense
 
-        pos = await self.gantry_position(mount, refresh=True)
+        pos = await self.gantry_position(checked_mount, refresh=True)
         probe_start_pos = pos._replace(z=probe_settings.starting_mount_height)
-        await self.move_to(mount, probe_start_pos)
-        total_z_travel = probe_settings.max_z_distance
```
ryanthecoder, line 2628: If we're not going to use this value from probe_settings. Lets remove it from that data struct so we don't confuse anything or miss any references using it in hardware-testing

api/src/opentrons/hardware_control/protocols/liquid_handler.py
```diff
@@ -171,3 +171,15 @@ async def drop_tip(
...
+
+    async def liquid_probe(
+        self,
+        mount: MountArgType,
+        max_z_dist: float,
+    ) -> float:
+        """Search for and return liquid level height using this pipette
+        at the current location.
+
+        mount : Mount.LEFT or Mount.RIGHT
```
sfoster1, line 183: should have `max_z_dist` explained too

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -0,0 +1,152 @@
...
+class LiquidProbeParams(PipetteIdMixin, WellLocationMixin):
+    """Parameters required to liquid probe a specific well."""
+
+    pass
+
+
+class LiquidProbeResult(DestinationPositionResult):
+    """Result data from the execution of a liquid-probe command."""
+
+    z_position: float = Field(..., description="Z position of the found liquid.")
```
sfoster1: worth mentioning here that
- This is the _absolute_ z position, as in from the deck, not the relative height from the bottom of the well
- It is in mm

api/src/opentrons/protocol_engine/execution/pipetting.py
```diff
@@ -245,6 +271,15 @@ async def blow_out_in_place(
...
         """Virtually blow out (no-op)."""
 
+    async def liquid_probe_in_place(
+        self,
+        pipette_id: str,
+        labware_id: str,
+        well_name: str,
+    ) -> float:
+        """Detect liquid level."""
+        return 0.0  # fix
```
sfoster1: let's either do the fix now or let's make it a better formatted TODO comment

sfoster1 (changes requested):
Couple last little changes but almost there!

sfoster1 (approved):
Very nice, love the changes, looks good to me!

PR #15352, opened 2024-06-05
feat(api): Return overpressure errors from `aspirate` and wait for their recovery

SyntaxColoring:
# Overview

Closes EXEC-376.

# Test plan

Run a protocol that `aspirate()`s. Hold the well plate against the nozzles to induce an overpressure error.

```python

requirements = {
    "apiLevel": "2.18",
    "robotType": "Flex",
}


    tip_rack = protocol.load_labware("opentrons_flex_96_tiprack_1000ul", "D2")
    well_plate = protocol.load_labware("nest_96_wellplate_100ul_pcr_full_skirt", "C3")

    protocol.load_trash_bin("A3")

    pipette = protocol.load_instrument("flex_96channel_1000")

    pipette.configure_nozzle_layout(protocol_api.COLUMN, start="A1")
    pipette.pick_up_tip(tip_rack["A12"])
    for _ in range(10):
        pipette.move_to(well_plate["A1"].bottom(-0.5))
        pipette.aspirate(1000)
        pipette.move_to(well_plate["A1"].top())
        pipette.blow_out()
    pipette.drop_tip()
```

With this PR, the run should pause and await recovery.

If you proceed to the next step, and that step is a blowout or home, the protocol *should* do that and then continue as normal. It will currently raise a `MustHomeError` for reasons that we think are related to firmware, and that we're looking into separately.

# Changelog

* Update the Protocol Engine `aspirate` command so that if it encounters an overpressure during the actual aspirate step, it swallows the exception, converts it to a defined `overpressure` error, and returns it.
* Update our error recovery policy to put the run in an `awaiting-recovery` state when this happens.
* Update PAPI's bindings to Protocol Engine to wait for any potential error recovery to complete before returning from the `aspirate()`.

# Review requests

None in particular.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/instrument.py
+++ b/api/src/opentrons/protocol_api/core/engine/instrument.py
@@ -153,7 +153,7 @@ class InstrumentCore(AbstractInstrument[WellCore]):
                 well_name=well_name,
                 well_location=well_location,
             )
-            self._engine_client.aspirate(
+            self._engine_client.aspirate_wait_for_recovery(
                 pipette_id=self._pipette_id,
                 labware_id=labware_id,
                 well_name=well_name,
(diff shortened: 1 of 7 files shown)


sfoster1 (approved):
This looks really good, exciting one to merge!

PR #15353, opened 2024-06-06
feat(robot-server): add protocol_kind to the protocols endpoint and persistance layer.

vegano1:
# Overview

A "Quick Transfer" protocol is a protocol that is generated by the on-device-display (ODD) (For now), to perform routine plate filler actions without needing to use a computer and the desktop app. This is a mostly client-side driven feature, However, we do need to make changes on the backend to support this feature. We primarily want 20 Protocol slots intended for "Quick Transfer" protocols. So we need to tell which type of protocol we are dealing with, so we can limit their count, migrate them, store them in the database, etc.

This pull request sets up the backend to handle "ProtocolKinds" sent as "protocolKind" Form data key in the `POST /protocols` endpoint. We then add this new key to the rest of the infrastructure required to consume this new key, persist, and get reported. This will let us mark protocols as "standard" or "quick_transfer" for the robot server to perform side effects on specific protocol kinds, like not storing runs for "Quick Transfer" protocols.

Closes: [PLAT-311](https://opentrons.atlassian.net/browse/PLAT-311) [PLAT-312](https://opentrons.atlassian.net/browse/PLAT-312) [PLAT-313](https://opentrons.atlassian.net/browse/PLAT-313) [PLAT-315](https://opentrons.atlassian.net/browse/PLAT-315) [PLAT-316](https://opentrons.atlassian.net/browse/PLAT-316)

# Test Plan

- [x] Make sure that a robot can start up properly after migration
- [x] Make sure adding the "protocolKind" form data key to the `POST /protocols` endpoint returns the correct kind of protocol
- [x] Make sure that the `GET /protocol` endpoint returns the correct protocolKind value
- [x] Make sure that `protocol_kind` persists through the robot server app restarts.

# Changelog
- Add the "protocolKind" form data key to the `POST /protocols` endpoint
- Add "protocolKind" to the ProtocolResource class
- Add "protocolKind" to the Protocol class
- Add the protocol_kind key to the database under schema 5
- Add a schema v4_v5 migration

# Review requests
- Make sure this makes sense and that I'm not overlooking anything too critical

# Risk assessment
- Medium, this alters the database requiring a migration

[PLAT-311]: https://opentrons.atlassian.net/browse/PLAT-311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robot-server/robot_server/protocols/router.py
```diff
@@ -186,6 +208,13 @@ async def create_protocol(
         " always trigger an analysis (for now).",
         alias="runTimeParameterValues",
     ),
+    protocol_kind: Optional[str] = Query(
+        default=None,
+        description=(
+            "Whether this is a `standard` protocol or a `quick-transfer` protocol."
+            "if ommited, the protocol will be `standard` by default."
+        ),
+    ),
```
sanni-t: What's the reason for making this a query on the request instead of the following two alternatives?-
1. make it a new form-data field, like `files` or `runTimeParameterValues` or
2. make it a part of the protocol contents itself so that the protocol reader can extract that data.

I personally think option 2 is a better way to handle this. It follows the current pattern of distinguishing between python vs. json files or OT2 vs Flex files, and it also makes sure that the protocol file itself is identifiable as a fast protocol file. With the method in this PR, you have to rely on the robot server and the database to maintain the true identity of this file, which seems like more work.

sanni-t: The fact that this also makes this function too complex is also an indicator that this protocol type specification should be handled somewhere else.

CaseyBatten: Method 2 introduces the problem of updating the JSON schema, which when discussed previously did not seem appropriate. My current understanding is that we want these to be quite similar to PD protocols, with the differentiator being how/where they are exposed to an end user. The content of the protocol itself is "promised" by the drafting client to be "quick", and the responsibility would be on the robot-server to categorize whether or not the protocol is treated as a "quick transfer" or not. The benefit with using an argument to do this is it does not matter where the quick transfer protocol originated, so long as it was indicated that it was a quick transfer upon upload. This leaves the door open for generating quick transfer protocols from other sources in the future with relative ease, such as through protocol designer.

robot-server/robot_server/protocols/protocol_models.py
```diff
@@ -109,3 +126,10 @@ class Protocol(ResourceModel):
             " See `POST /protocols`."
         ),
     )
+
+    protocolKind: Optional[str] = Field(
```
CaseyBatten: Instead of type `Optional[str]` should this be type `Optional[ProtocolKind]`? We already determine a type of ProtocolKind for the supplied str value at the POST /protocols endpoint before we execute a `Protocol.construct(...)`.

robot-server/robot_server/protocols/router.py
vegano1: @sanni-t 

We did not go with updating the schema because of the reason Casey mentioned above, as for the query arg instead of form data, the idea was to have the same interface in the POST /protocols and GET /protocols endpoint. That being said there is no functional difference on my end, so I can change the interface to form data instead of query arg for the `POST /protocols` endpoint.

This is a good call out.

CaseyBatten (approved):
Looks good overall, as discussed in person if we change out the query on the POST `/protocols` request to a form instead this will be in line with the rest of our HTTP API decisions so far. All other decisions seem sensible, your call regarding the exclusion of the `ProtocolKind` type on the Protocol constructor.

robot-server/robot_server/protocols/router.py
sanni-t: Thanks for the in-person and github explanations regarding the background on quick transfer protocols, @CaseyBatten @vegano1. Makes sense why the method 2 I mentioned above (which would involve changing the schema) does not fit the goal.

Ya, I think changing it to a form-data arg would certainly preserve the consistency for this endpoint. Thanks!

robot-server/robot_server/protocols/protocol_models.py
```diff
@@ -14,6 +16,21 @@
 from .analysis_models import AnalysisSummary
 
 
+class ProtocolKind(str, Enum):
+    """Kind of protocol, standard or quick-transfer."""
+
+    standard = "standard"
+    quick_transfer = "quick-transfer"
```
sanni-t: Nitpick- should use all caps for enum names

robot-server/robot_server/protocols/protocol_models.py
vegano1: Good call!

sanni-t (approved):
Looks good! Thank you for making all the changes! 
Good to merge once tests pass

PR #15357, opened 2024-06-06
refactor(app): Support fixit commands in Drop Tip flows

mjhuff:
Closes [EXEC-424](https://opentrons.atlassian.net/browse/EXEC-424)

# Overview

This PR effectively overhauls Drop Tip flows to facilitate its integration into Error Recovery. There are no functional changes, but a lot has changed:

- Flows now take an optional `fixitCommandTypeUtils` which is a set of overrides that Error Recovery will provide. If present, commands issue a "fixit" type as opposed to creating a maintenance run and issuing "setup" type commands.
- There's conditional rendering of the views depending on the command type, see `DropTipWizardFixitType` and `DropTipWizardSetupType`. Although there's coupling between the command type and essentially "is this error recovery or a standalone wizard flow?", this is probably an ok assumption for now. At the very least, it's hard to imagine we'd want a drop tip flow utilizing fixit commands that isn't part of Error Recovery. 
- Lots more testing and (hopefully) increased code readability.
- Drop Tip flows now utilizes routing that's similar to how it works in Error Recovery. There are three routes: Before Beginning, Drop Tip, and Blowout, and each route contains steps, with each representing a different view of the wizard.

Drop Tip flows still have room for improvement. This PR does not:
- Add testing that was missing to specific presentation components (all the "step" views, such as `ChooseLocation`, `Success`, etc). 
- Fully refactor those presentation components. There's still room to clean these up.

# Test Plan
- Fully smoke-tested flows on both desktop and ODD, including exit & step counter edge cases and error handling special-cases.

# Review requests
The changes are really all the files in the `hook` directory, `DropTipWizardHeader`, and `DropTipWizard` - all of these are effectively new components/utilities. 

# Risk assessment
Medium. This is a large overhaul. It's localized just to Drop Tip flows, though. 

[EXEC-424]: https://opentrons.atlassian.net/browse/EXEC-424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Okay, this looks really good to me. I think at some point we'll want follow on work to hoist these "what kind of run to use" hooks higher and go full dependency-injection - we could determine it at the callsite or at least at the highest hook level.

This is also exposing some grossness with the rules-of-hooks "have to be called every time" thing - I really don't love having to curry some of these params down into a react-query enable props and have everything still get called but just kind of not do anything if the maintenance run id is null. I'm not sure what the better way is though, and hiding it in some dependent hooks works well enough.

mjhuff:
> This is also exposing some grossness with the rules-of-hooks "have to be called every time" thing - I really don't love having to curry some of these params down into a react-query enable props and have everything still get called but just kind of not do anything if the maintenance run id is null. I'm not sure what the better way is though, and hiding it in some dependent hooks works well enough.

Yeah agreed. It's gross. I don't know a Reactive pattern for this without building new hooks that are effectively conditional wrappers around the raw axios requests, which is equally as gross. I'll marinate on this a bit.

PR #15360, opened 2024-06-06
fix(protocol-designer): add missing prop to getPipetteCapacity usage

jerader:
closes AUTH-489

# Overview

Somehow this didn't error before even though a usage of `getPipetteCapacity` was missing the `labwareEntities`? I guess because the `tiprack` prop is optional. But anyway glad it has been caught and fixed.

# Test Plan

# Changelog

add missing labware entites prop to util

# Review requests

# Risk assessment

low


b-cooper (approved):
Looks right to me!

PR #15368, opened 2024-06-07
refactor(app): use one-col content in attach pipette

sfoster1:
Now that we have a one-col-simple-buttons component, let's use it. A place that uses this layout open-coded is the ODD attach pipette screen. Beautifully enough, we can drop the component in place instead of the open-coded version and the tests all still pass.
## Testing
Also I don't trust at all that the tests pass because there's some weird nested stuff for the back button so let's actually try it on a machine.


mjhuff (approved):
Exciting! Thank you!

PR #15379, opened 2024-06-10
feat(app): Add Drop Tip Flows to Error Recovery

mjhuff:
Closes [EXEC-466](https://opentrons.atlassian.net/browse/EXEC-466)

# Overview

https://github.com/Opentrons/opentrons/assets/64858653/3c30a5c0-f434-4910-bf8b-0e01b9d076ba

Wires up Drop Tip Flows to Error Recovery through the "Cancel Run" option when tips are detected. Although this PR does not conclude all possible Drop Tip flow refactors, this is most of the work required for the [Error Recovery MVP ](https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=2727-77978&t=7xjW0JxCg2kc7Tpo-4) outside of a couple slight overrides that will need to be added for Overpressure ER flows.

While #15357 refactored DT logic to support fixit commands, integrating DT into ER required a bit more work, namely:

- We need to keep track of the DT flow route. This was a bit tricky - we probably don't want to tightly couple ER routing with routing in other wizard flows, so the solution here, in `useRecoveryRouting`, is to create a "subMap", which we can effectively pass to any flow that supports ER. Eventually, if we do mutli-app same view coordination, we simply append the sup map routing to the ER routing to direct users to the correct step (we'll need to tweak DT wiz a bit to support a manual "go to this particular step" override, but that will be be part of the overpressure work, anyway). Note that all of Drop Tip flows within ER are therefore in a single step, named `WIZARD`.
- We need a new component for the "before we enter" Drop Tip flows from ER, and this is `ManageTips`.
- Error Recovery needs to keep track of the previous route visited, because in the future (with overpressure), we need to route after DT ends to the previous route within ER flows. This functionality will likely be useful down the road as a general utility as well. 
- `ProceedToRoute` should now optionally support routing to a specific route _and_ step, since we actually have our first route with multiple steps, hence `ProceedToRouteAndStep`.
- Given the nature of ER, we need to track tip attachment status _after_ the flows have launched, since users may or may not issue `fixit` commands before deciding to cancel a run. We want accurate tip attachment status before prompting users about Drop Tip flows. To mitigate latency caused by checking tip status, any time a user hits the `OptionSelection` route, we prefetch tip attachment status. However, it's still possible for a user to cancel a run before this check is done (especially in longer protocols on an OT-2), so we now have a [new loading state ](https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=3202-91697&t=7xjW0JxCg2kc7Tpo-4) and related utils for calculating and managing this loading state (see `OptionSelection` changes and `CancelRun` changes).
- The presentation layer in DT flows is not ideal to say the least....I did not make any more ideal for several reasons:
  1. Design sounds like they might refactor the original DT flows, which means that will change.
  2. There are new components for Error Recovery in the works. Regardless of what we do now, we'll have to revisit this later.
  3. It's going to take several days of straight refactoring to get the presentation layer refactored well. 
  4. While messy CSS isn't fun, it's easy to verify everything works as it should just by rendering a particular view.
  5. There are bigger fish to fry, and while this is worth revisiting, it's probably not a top priority. I left TODOs and made a ticket here [EXEC-520](https://opentrons.atlassian.net/browse/EXEC-520).
- Lastly, the tip attachment logic needed a bit of refactoring, since we probably don't want to prompt users that tips are possibly attached right after they manually issued fixit commands involving tip attachment status (ie, a user hits `RunSummary`, clicks "Run again/return to dash", and sees the tip attachment modal. Getting this to work was surprisingly straightforward, and we now only re-prompt users that they may have tips attachment if A) the flex sensors tell us that's a possibility, or B) the run commands history indicates the fixit commands involving the pipette tips failed. See `getPipettesWithTipsAttached`. 

Just as a note, currently, any errors that occur within DT Wiz within ER effectively hard lock the user from doing anything without power cycling. There are currently no designs to handle this, and I didn't really want to add to the complexity of this already large PR with a stop-gap solution. I'll be sure to tackle this ASAP.

# Test Plan
- Smoke tested all the drop tip flows, ensuring ER flows and [regular DT flows](https://www.figma.com/design/0kuPeCi1t2Auu2GPMnqRb2/Primary%3A-Flex?node-id=20152-154754&t=2UGfMt2k4HpPdOMe-4) match designs. To enter DT flows within ER, fail a tip pickup command and then try to cancel the run.
- Ensured that tip attachment behavior works as expected both within ER and during Run Summary. This includes not issuing fixit commands for rectifying tip attachment within ER and ensuring that the DT modal prompt occurs during Run Summary.

# Changelog
- Added Drop Tip flows to Error Recovery.

# Risk assessment
low - this is mainly CSS work for existing dt flows, and the error recovery stuff is all behind ffs.

[EXEC-466]: https://opentrons.atlassian.net/browse/EXEC-466?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EXEC-520]: https://opentrons.atlassian.net/browse/EXEC-520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/DropTipWizardFlows/ChooseLocation.tsx
```diff
@@ -174,7 +165,7 @@ const GO_BACK_BUTTON_STYLE = css`
   @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
     font-weight: ${TYPOGRAPHY.fontWeightSemiBold};
     font-size: ${TYPOGRAPHY.fontSize22};
-    padding-left: 0rem;
+    padding-left: 0;
```
jerader, line 168: does this change do anything?

app/src/organisms/DropTipWizardFlows/DropTipWizard.tsx
```diff
@@ -222,6 +244,7 @@ export const DropTipWizardContent = (
         header={errorDetails?.header ?? t('error_dropping_tips')}
         subHeader={subHeader}
         justifyContentForOddButton={JUSTIFY_FLEX_END}
+        marginTop={`-${SPACING.spacing68}`} // See EXEC-520. This clearly isn't ideal.
```
jerader, line 247: 👍

app/src/organisms/DropTipWizardFlows/ChooseLocation.tsx
mjhuff: Nope, but it makes me feel better 😄

app/src/organisms/DropTipWizardFlows/hooks/useDropTipWithType/useDropTipCommands.ts
```diff
@@ -75,8 +74,8 @@ export function useDropTipCommands({
   const handleCleanUpAndClose = (homeOnExit: boolean = true): Promise<void> => {
     return new Promise(() => {
       if (issuedCommandsType === 'fixit') {
-        const { onCloseFlow } = fixitCommandTypeUtils as FixitCommandTypeUtils
-        onCloseFlow()
+        closeFlow()
+        return Promise.resolve()
```
jerader, line 78: why do we need to return `Promise.resolve()` here?

app/src/organisms/ErrorRecoveryFlows/RecoveryOptions/ManageTips.tsx
```diff
@@ -0,0 +1,177 @@
...
+// Builds the overrides injected into DT Wiz.
+export function useDropTipFlowUtils({
+  tipStatusUtils,
+  failedCommand,
+  previousRoute,
+  trackExternalMap,
+}: RecoveryContentProps): FixitCommandTypeUtils {
+  const { t } = useTranslation('error_recovery')
+  const { runId } = tipStatusUtils
+  const failedCommandId = failedCommand?.id ?? '' // We should have a failed command here unless the run is not in AWAITING_RECOVERY.
```
jerader, line 155: seems like this comment should be prefaced with a TODO

app/src/organisms/DropTipWizardFlows/hooks/useDropTipWithType/useDropTipCommands.ts
mjhuff: So other commands that chain off this command properly work. If you run a `.then()` after `handleCleanUpAndClose` without a `Promise.resolve()`, and that condition occurs, the `.then()` won't fire.

app/src/organisms/ErrorRecoveryFlows/RecoveryOptions/ManageTips.tsx
mjhuff: I don't believe so. It's more an explanation on why this is ok.

jerader: oh yup, i see, my bad, there is already a failed command.

app/src/organisms/DropTipWizardFlows/__tests__/DropTipWizard.test.tsx
```diff
@@ -178,15 +178,13 @@ describe('DropTipWizardContent', () => {
...
       fixitCommandTypeUtils: {
-        copyOverrides: { tipDropCompleteBtn: 'proceed_to_tip_selection' },
+        copyOverrides: { tipDropCompleteBtnCopy: 'proceed_to_tip_selection' },
       } as any,
     })
 
     expect(vi.mocked(Success)).toHaveBeenCalledWith(
-      // This "false" in front of the text is caused by how renderWithProviders
-      // implements i18n. This is hacky and should be revisited.
       expect.objectContaining({
```
jerader, line 186: oof thanks for this explanation

mjhuff: Luckily this works now that we do dependency injection here!

jerader (approved):
i didn't test on a robot and just focused on the code - but code looks good and the comments throughout make it easy to read and follow 👍 nice work!

mjhuff:
> i didn't test on a robot and just focused on the code - but code looks good and the comments throughout make it easy to read and follow 👍 nice work!

Thank you for reviewing! Much appreciated!

sfoster1 (approved):
Looks good with the caveats you mentioned!

PR #15381, opened 2024-06-11
fix(app): fix Error Recovery launching when manually cancelling a run

mjhuff:
Closes [EXEC-521](https://opentrons.atlassian.net/browse/EXEC-521)

# Overview

When manually cancelling a run, Error Recovery would launch because it recognizes `STOP_REQUESTED` as a valid state to show ER flows (correct behavior). However, we don't want to show error recovery if the FIRST run status seen is `STOP_REQUESTED`. Error Recovery should always see `AWAITING_RECOVERY` before `STOP_REQUESTED`. 

`STOP_REQUESTED` is a valid state to render ER, and we use this when cancelling a run after recovery actions to show special run cancelling components.

I think adding state to the ER hook is the best way to solve this, otherwise we'd have to pass some sort of state for each manual cancel action elsewhere. 

# Test Plan
- Run a run with ER FFs enabled, and manually cancel it. The ER splash will no longer appear.

# Changelog
- Error Recovery no longer renders when manually cancelling a run.

# Risk assessment
low

[EXEC-521]: https://opentrons.atlassian.net/browse/EXEC-521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/ErrorRecoveryFlows/__tests__/ErrorRecoveryFlows.test.tsx
+++ b/app/src/organisms/ErrorRecoveryFlows/__tests__/ErrorRecoveryFlows.test.tsx
@@ -39,19 +39,19 @@ describe('useErrorRecoveryFlows', () => {
   })
 
   it('should toggle the value of isEREnabled properly when the run status is valid', () => {
-    const { result } = renderHook(() =>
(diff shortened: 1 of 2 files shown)


sfoster1 (approved):
noice

PR #15385, opened 2024-06-11
feat(shared-data, protocol-designer): fix adapter and aluminum categorization

syao1226:
# Overview

AUTH-76: Fix categorization of adapters and aluminum blocks in Labware Library

# Test Plan

- Verified that the labware definitions are listed correctly under “adapters” in the website LL 
- Verified that the labware definitions are listed correctly under “aluminum” in the website LL 

# Changelog

- Fixed categorization of labware adapter and aluminum blocks
- Updated labware definition category for cypress tests failing issue

# Review requests

# Risk assessment

PR #15386, opened 2024-06-11
refactor(protocol-designer): check if adapter through allowedRoles

jerader:
closes AUTH-506

# Overview

This is not really a bug (yet) but will be a bug once https://github.com/Opentrons/opentrons/pull/15385 merges in. Basically, we updated the categorizations of some labware by changing the `displayCategory`. This is fine, we just can't rely on the `displayCategory` for cataloging the adapters in PD, instead we need to rely on the `allowedRole`s. This PR fixes that.

# Test Plan

Not much to test here, just check that my logic makes sense!

# Changelog

- update the `deckSetup` component to base the adapter info off of `includedRoles` instead of `displayCategory`

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/components/DeckSetup/index.tsx
+++ b/protocol-designer/src/components/DeckSetup/index.tsx
@@ -237,8 +237,9 @@ export const DeckSetupContents = (props: ContentsProps): JSX.Element => {
           moduleOnDeck.slot
         )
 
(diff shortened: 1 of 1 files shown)


shlokamin (approved):
nice, thanks for auditing this!

PR #15389, opened 2024-06-11
feat(hardware control): create logic to detect if tip has hit bottom …

aaron-kulkarni:
…of well or liquid

Creates a new file called sensor_utils.py. Introduces new error type for tip hitting bottom of the well when searching for liquid. Creates a function that detects whether the tip has hit a liquid or the bottom of the well based on pressure data.

fix EXEC-517

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/hardware/opentrons_hardware/hardware_control/sensor_utils.py
+++ b/hardware/opentrons_hardware/hardware_control/sensor_utils.py
@@ -8,27 +8,21 @@ import numpy as np
 
 
 def did_tip_hit_liquid(
-        pressure_readings: list[float],
-        liquid_solid_threshold: float
+    pressure_readings: list[float], liquid_solid_threshold: float
 ) -> bool:
-    
+
     if len(pressure_readings) < 5:
         raise LiquidNotFoundError(
-            "Liquid not found.",
-            {
-                "Not enough data to calculate pressure change."
-            }
+            "Liquid not found. Not enough data to calculate pressure change",
         )
-    pressure_difference = np.gradient(pressure_readings[-5:], 1)  
+    pressure_difference = np.gradient(pressure_readings[-5:], 1)
     end_difference = pressure_difference[-1]
     if end_difference > liquid_solid_threshold:
-        #Hit Bottom of Well
+        # Hit Bottom of Well
         raise TipHitWellBottomError(
-                "Tip hit the bottom of the well.",
-                {
-                    "Pressure change detected: " + str(end_difference)
-                },
-            )
+            "Tip hit the bottom of the well.",
+            {"Pressure change detected: ": str(end_difference)},
+        )
     else:
-        #Hit Liquid
+        # Hit Liquid
         return True
--- a/hardware/opentrons_hardware/hardware_control/sensor_utils.py
+++ b/hardware/opentrons_hardware/hardware_control/sensor_utils.py
@@ -1,3 +1,5 @@
+"""Functions aiding in Liquid Level Detection."""
+
 from opentrons_shared_data.errors.exceptions import (
     TipHitWellBottomError,
     LiquidNotFoundError,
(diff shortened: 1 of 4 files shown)


hardware/opentrons_hardware/hardware_control/sensor_utils.py
```diff
@@ -0,0 +1,27 @@
+from opentrons_shared_data.errors.exceptions import (
+    TipHitWellBottomError,
+)
+
+
+import numpy as np
+
+
+def detectLiquidOrBlockage(
```
sfoster1: our style is `snake_case` for python, so this should be `detect_liquid_or_blockage`, same with the variables (sorry!)

hardware/opentrons_hardware/hardware_control/sensor_utils.py
```diff
@@ -0,0 +1,27 @@
...
+        pressureReadings: list[float],
+        liquidSolidThreshold: float
+) -> None:
```
sfoster1: I think we want this function to return a `bool`, and probably have the function be _named_ in a way that can be answered with true or false - i.e., `did_hit_well_bottom(...) -> bool` or something. 

The reason to make it return a bool is that a function that returns a bool is more generally useful. A function that either raises or returns can't be used for anything else without doing some exception-as-control-flow mechanics we don't like doing; but if the function returns a bool, the caller can raise if the bool is false

hardware/opentrons_hardware/hardware_control/sensor_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    dx = 1
+    #take derivative of pressureReadings with respect to dx
```
sfoster1: This comment and the variables names talk about derivatives, but this is a discrete gradient and thus operates in terms of _difference_. It could be quite relevant, given that when you're living in discrete world with noise, differencing amplifies noise. Let's phrase things as differences to keep that in mind.

I also think this comment is duplicative with the next line; if you know what `dx` and a gradient are, you can understand the next line, and if you don't then this comment doesn't tell you anything new.

sfoster1 (changes requested):
Great job with the enumerated error! I think there's some things to fix up about the functions though.

sfoster1:
Also, the thing with the commit titles is 
- they want to have a type that is correct (which this is)
- but the scope, i.e. what's inside the parens, actually has to be exactly the name of a subdirectory of the monorepo 

So here it would be exactly `feat(hardware):`

hardware/opentrons_hardware/hardware_control/sensor_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+import numpy as np
+
+
+def did_tip_hit_liquid(
+        pressure_readings: list[float],
+        liquid_solid_threshold: float
+) -> bool:
+    pressure_difference = np.gradient(pressure_readings[-5:], 1)
```
sfoster1, line 20: What happens if there's fewer than 5 readings in the list?

sfoster1 (changes requested):
I think we need to handle the length of the list in some way - either gracefully degrade under 5 samples, or raise an error

hardware/opentrons_hardware/hardware_control/sensor_utils.py
ryanthecoder: there should always be a minimum of 100 in the future implementation but a check here makes sense, maybe we can raise a LiquidNotFoundError if there isn't enough data

sfoster1 (changes requested):
The code looks good to me, so now we just need to get the static analysis passing. You should autoformat the code with `make -C api format`, and see what `make -C api lint` gives you; same with `make -C shared-data/python format` and `make -C shared-data/python lint` (and a top level `make format-js` to get the errors.json file autoformatted)

sfoster1 (approved):
Looks good to me! Nice work. Future stuff will require tests but here we're making exactly one call to numpy so I think it would be a little much :)

PR #15392, opened 2024-06-12
ci: remove workaround from workflows

koji:
# Overview
remove workaround from workflows
the issue was solved already so we don't need a workaround in workflows.
https://github.com/microsoft/linux-package-repositories/issues/130

# Test Plan

# Changelog

# Review requests

# Risk assessment
low


sfoster1 (changes requested):
This isn't the first time this has happened; in fact it seems to happen about yearly. I think that if our images still work with this workaround, we should keep it.

PR #15396, opened 2024-06-12
fix(protocol-designer): append definitionId to the labwareId in migration

jerader:
closes RESC-271

# Overview

This PR addresses 2 bugs in 1. 

The first bug has to do with the 7.0.0 migration file. Lots of the logic relies on the labware id having the `definitionId` appended to the end of the uuid string. This is NOT GREAT but that's how i architectured it so the migration should handle appending it. It does in most cases but somehow missed some cases which are fixed in this PR. So basically the bug is the migration couldn't find certain labwareIds and labware definitions because it was searching via a string that didn't include the `definitionId`. 

The 2nd bug is in the 8.1.0 migration file where to populate the tiprack command, it is incorrectly searching for a tipRack key in the step form which doesn't exist yet? idk how we didn't notice this bug sooner. I guess it doesn't whitescreen but it does require the user to add their tiprack location manually instead of auto-populating upon migrating.

# Test Plan

Upload the attached protocol and see that it migrated correctly. Go to the deck map and see that there are no errors in the steps (NOTE they will have the tuberack warnings but that is fine). Export the protocol. Then, reimport the migrated protocol and see that there are no errors.

[16S Bakterien qPCR 3 Primer 5mL JoLo tubes_V6.json](https://github.com/user-attachments/files/15805665/16S.Bakterien.qPCR.3.Primer.5mL.JoLo.tubes_V6.json)

# Changelog

- make sure each labwareId instance has the definitionid appended to the end of it unless it is the fixed trash
- fix the tiprack bug in the 8.1.0 migration

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/load-file/migration/7_0_0.ts
+++ b/protocol-designer/src/load-file/migration/7_0_0.ts
@@ -248,7 +248,7 @@ export const migrateFile = (
     if (!getIsAdapter(labwareId)) {
       const definitionId = labware[labwareId].definitionId
       const labId =
-        labwareId === 'fixedTrash' || labwareId.includes(definitionId)
+        labwareId.includes(definitionId) || labwareId === 'fixedTrash'
           ? labwareId
           : `${labwareId}:${definitionId}`
       acc[labId] = labwareLocationUpdate[labwareId]
--- a/protocol-designer/src/load-file/migration/7_0_0.ts
+++ b/protocol-designer/src/load-file/migration/7_0_0.ts
@@ -79,7 +79,7 @@ export const migrateFile = (
     .filter(labwareId => getIsAdapter(labwareId))
     .reduce((acc: LabwareIdMapping, labwareId: string): LabwareIdMapping => {
       const { labwareUri, adapterUri } = getAdapterAndLabwareSplitInfo(
-        labwareId
+        labware[labwareId].definitionId
       )
       const newLabwareId = `${uuid()}:${labwareUri}`
       const newAdapterId = `${uuid()}:${adapterUri}`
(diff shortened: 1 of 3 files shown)


protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -79,7 +79,7 @@ export const migrateFile = (
     .filter(labwareId => getIsAdapter(labwareId))
     .reduce((acc: LabwareIdMapping, labwareId: string): LabwareIdMapping => {
       const { labwareUri, adapterUri } = getAdapterAndLabwareSplitInfo(
```
jerader, line 81: if you search `getAdapterAndLabwareSplitInfo` util, it relies on having the labwareDefURI info, which `definitionId` === `defURI` (i know its confusing but we removed `definitionId`s starting in 7.0.0 and started only using `defURI` or `labwareDefURI`.

protocol-designer/src/load-file/migration/7_0_0.ts
```diff
@@ -240,7 +246,12 @@ export const migrateFile = (
     labwareLocationUpdate
   ).reduce((acc: LabwareLocationUpdate, labwareId: string) => {
     if (!getIsAdapter(labwareId)) {
-      acc[labwareId] = labwareLocationUpdate[labwareId]
+      const definitionId = labware[labwareId].definitionId
+      const labId =
+        labwareId === 'fixedTrash' || labwareId.includes(definitionId)
```
koji: nit

for consistency

```suggestion
         labwareId.includes(definitionId) || labwareId === 'fixedTrash'
```

koji (approved):
the changes look good to me and the sandbox worked as expected.

e2e test failed. once the test error is fixed, this pr is good to go.

shlokamin (commented):
can we add a test case to the v7 migration file that checks for this fix? it'll serve as great documentation for what's going on here and why we're doing this

PR #15398, opened 2024-06-12
chore: deploy components on more app changes

sfoster1:
This workflow was using eslint glob syntax which isn't used here, and also we should trigger on source changes


shlokamin (approved):
oopsy thanks for fixing this!

PR #15400, opened 2024-06-12
fix(api): add runner load error handling to analyze cli

sanni-t:
Closes AUTH-486

# Overview

Fixes the issues caught in #15211 

Previously, RTP validation occurred as part of a simulating runner's run process. This run process executes the protocol inside a separate task, and any errors raised while running the task are wrapped and presented in the 'Rune result" as an analysis error. 

Now that RTP validation is happening as part of the runner load process, we have to handle the error raised in the load process in a similar way as well. This PR adds this missing error handling in `analyze` cli function.

# Test Plan

- [ ] Upload a protocol with incorrect RTP definition and verify that the in-app analysis fails with the correct error and no traceback in the error pop-up window.
- [ ] Verify that the static data of the above protocol- metadata, requirements and files are populated in the failed analysis
- [ ] Verify that any other, non-RTP-related errors raised during the runner load process are handled just like they would have in the previous stable software.

# Review requests

- does this approach make sense? Does it miss any edge cases?

# Risk assessment

Medium. Updates a critical part of in-app analysis, but is a fix and is isolated to error handling of `runner.load` only.

### DIFF
--- a/api/src/opentrons/cli/analyze.py
+++ b/api/src/opentrons/cli/analyze.py
@@ -37,6 +37,7 @@ from opentrons.protocol_runner import (
     create_simulating_runner,
     RunResult,
     PythonAndLegacyRunner,
+    JsonRunner,
 )
 from opentrons.protocol_engine import (
     Command,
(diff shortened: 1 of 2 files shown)


api/src/opentrons/cli/analyze.py
```diff
@@ -197,12 +211,75 @@ def _get_return_code(analysis: RunResult) -> int:
...
+
+
+@track_analysis
 async def _do_analyze(protocol_source: ProtocolSource) -> RunResult:
 
     runner = await create_simulating_runner(
         robot_type=protocol_source.robot_type, protocol_config=protocol_source.config
     )
-    return await runner.run(deck_configuration=[], protocol_source=protocol_source)
+    if isinstance(runner, PythonAndLegacyRunner):
```
sanni-t: Oops! Forgot to load the non-python runner. Doing that now

api/src/opentrons/cli/analyze.py
```diff
@@ -197,12 +212,78 @@ def _get_return_code(analysis: RunResult) -> int:
...
+                    yield PythonException(exc)
+
+        super().__init__(
+            code=ErrorCodes.GENERAL_ERROR,
+            message=message,
+            wrapping=[e for e in _convert_exc()],
+        )
+
+
+@track_analysis
```
SyntaxColoring: Is this `@track_analysis` addition intentional, or is it a merge conflict mistake?

api/src/opentrons/cli/analyze.py
```diff
@@ -197,12 +212,78 @@ def _get_return_code(analysis: RunResult) -> int:
...
+                wrapping=[error],
+            )
+            error_occ = ErrorOccurrence.from_failed(
+                id=err_id, createdAt=err_created_at, error=enumerated_wrapper
+            )
+        analysis = RunResult(
+            commands=[],
+            state_summary=StateSummary(
+                errors=[error_occ],
+                status=EngineStatus.IDLE,
```
SyntaxColoring, line 274: We're supplying a placeholder `EngineStatus.IDLE` because that's needed to construct the returned `RunResult`. But the returned `RunResult` is, elsewhere in this file, mostly just thrown away and translated into the `AnalyzeResults` that we actually care about.

Can we avoid this throwaway placeholder thing by directly returning the `AnalyzeResults` from this function?

SyntaxColoring (commented):
This mostly makes sense to me. The duplication between here and `ProtocolAnalyzer` is unfortunate, and so is the tedium in constructing the error shape, but both make sense given where we are today. 

A couple of minor points:

api/src/opentrons/cli/analyze.py
sanni-t: Oh! Good catch! Definitely merge conflict mistake

api/src/opentrons/cli/analyze.py
sanni-t: Trying to avoid creating this placeholder ends up mostly adding bloat code or some funny branching logic because when there's no error during runner load, the analysis result does need to be constructed from a `RunResult`. 
I don't think the small placeholder is worth the trouble.

SyntaxColoring (approved):

PR #15402, opened 2024-06-12
style(app, protocol-designer, components): promote no-unsafe-argument from warn to error, fix all

b-cooper:
# Overview

Promote the typescript eslint rule for no-unsafe-argument to error and fix all outstanding instances
in the mono repo

# Review requests

- smoke test front end apps

# Risk assessment
medium, should have no effect on code, but many files were touched


protocol-designer/src/analytics/mixpanel.ts
```diff
@@ -8,10 +8,10 @@ import type { BaseState } from '../types'
 export type AnalyticsEvent =
   | {
       name: string
-      properties: { [key: string]: unknown }
-      superProperties?: { [key: string]: unknown }
+      properties: Record<string, unknown>
+      superProperties?: Record<string, unknown>
     }
-  | { superProperties: { [key: string]: unknown } }
+  | { superProperties: Record<string, unknown> }
```
jerader, line 14: does this error with the new lint error or is this a random refactor? (asking because i 100% prefer it written this way).

jerader (approved):
this is very nice, thank you! 📝 🦖 . i smoke tested PD primarily and ~bopped around a bit in the app~ (nevermind i thought i did but it actually whitescreens) but PD looks good to me

jerader:
this is the app whitescreen error i am getting

Screenshot 2024-06-17 at 10 25 35

protocol-designer/src/analytics/mixpanel.ts
b-cooper: It is a slightly unrelated fix, and probably not necessary here.  I'm actually in favor of the index-signature syntax of the Record keyword myself.  I think we probably want to relax or invert that typescript eslint rule. I'll tackle that in the next PR.

PR #15403, opened 2024-06-12
feat(app, app-shell-odd): add function to read csv file from usb for rtp phase2

koji:
# Overview
add a function that loads csv files from an usb-drive, update protocol card & pinned protocol, and add ChooseCsvFile component (display csv files on USB).

close AUTH-460, AUTH-468, and partially AUTH-462 and AUTH-470

This PR doesn't update protocol details screen and Parametes screen. They will be implemented in a following pr since this PR's main part is displaying csv files on USB.

The following yellow background shows up when turning on the feature flag and a protocol name includes `RTP`.
The rendering condition part will be update in a following pr.
Screenshot 2024-06-18 at 7 24 19 PM

Screenshot 2024-06-18 at 7 12 26 PM

`Confirm selection` part will be implemented and a function to control radio buttons (AUTH-458)
Screenshot 2024-06-18 at 7 37 51 PM

Screenshot 2024-06-18 at 7 38 20 PM

# Test Plan
1. push this branch to a Flex
2. select a RTP protocol that has yellow background in Protocol Dashboard screen
3. do set up

# Changelog

# Review requests

# Risk assessment


app-shell-odd/src/usb.ts
```diff
@@ -164,3 +170,30 @@ export function watchForMassStorage(dispatch: Dispatch): () => void {
...
 }
+
+export function registerFilePath(
+  dispatch: Dispatch
+): (action: Action) => unknown {
+  return function handleAction(action: Action) {
+    switch (action.type) {
+      case ROBOT_MASS_STORAGE_DEVICE_ENUMERATED:
+        void enumerateMassStorage(action.payload.rootPath).then(contents => {
+          // Note (kk:06/12/2024) need to filter out resource fork files which starts ._
```
sfoster1: Could also do this inside the `enumerateMassStorage` function itself, which already does [something similar](https://github.com/Opentrons/opentrons/pull/15403/files#diff-94938be0cbfb479f2093d629201d5f82485a9606358a130d884594982af673b9R46) for os-specific trash and metadata directories

app-shell-odd/src/actions.ts
```diff
@@ -83,6 +83,7 @@ import {
   DISCOVERY_START,
   DISCOVERY_FINISH,
   SEND_READY_STATUS,
+  SEND_FILE_PATH,
```
shlokamin:
```suggestion
  SEND_FILE_PATHS,
```

app-shell-odd/src/actions.ts
```diff
@@ -99,6 +100,7 @@ import type {
   StartDiscoveryAction,
   FinishDiscoveryAction,
   RobotSystemAction,
+  SendFilePathAction,
```
shlokamin:
```suggestion
  SendFilePathsAction,
```

app-shell-odd/src/usb.ts
shlokamin: yeah i think this should be done with `enumerateMassStorage` — they are almost identical from a system interaction perspective and i don't think this warrants a new reducer that does pretty much the same thing.

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+import type { CsvFileParameter } from '@opentrons/shared-data'
+
+interface ChooseCsvFileProps {
+  handleGoBack: () => void
+  // ToDo (kk:06/18/2024) null will be removed when implemented required part
+  parameter: CsvFileParameter | null
+  setParameter: (value: boolean | string | number, variableName: string) => void
+  /** if files on robot has the same id, the file name is highlighted */
+  csvFileInfo: string
+  /** If a user chooses a file on robot, store fileId. If a user chooses a file on usb, store full file path.  */
```
shlokamin: i dont think these comments are necessary

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+  parameter,
+  setParameter,
+  csvFileInfo,
+  setCsvFileInfo,
+}: ChooseCsvFileProps): JSX.Element {
+  const { t } = useTranslation('protocol_setup')
+  const csvFilesOnUSB = useSelector(getFilePaths).payload.filePaths ?? []
+
+  // ToDo (kk:06/12/2024) get files from the endpoint: GET /protocols/{protocolId}/dataFiles/
+  // return format: https://opentrons.atlassian.net/browse/AUTH-428
```
shlokamin: don't we already have api client and react api client utils for this? may as well use them!

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const csvFilesOnRobot: any[] = []
+
+  // ToDo (06/20/2024) this will removed when working on AUTH-521
+  // const handleOnChange = (newValue: string | number | boolean): void => {
+  //   setParameter(newValue, parameter?.variableName ?? 'csvFileId')
+  // }
```
shlokamin, line 52: leftover comment?

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <Flex css={CONTAINER_STYLE}>
+            <LegacyStyledText css={HEADER_TEXT_STYLE}>
+              {t('csv_files_on_robot')}
+            </LegacyStyledText>
+            <Flex css={LIST_CONTAINER_STYLE}>
+              {csvFilesOnRobot.length !== 0 ? (
+                csvFilesOnRobot.map(csv => (
+                  <RadioButton
+                    key={csv.fileId}
+                    data-testid={`${csv.fileId}`}
```
shlokamin: do we need to add a test id? and if we do, i dont think it should be the file id. it should be something an e2e testing library should be able to create easily without having to know about system internals. for now i think we should remove it

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+          </Flex>
+          <Flex css={CONTAINER_STYLE}>
+            <LegacyStyledText css={HEADER_TEXT_STYLE}>
+              {t('csv_files_on_usb')}
+            </LegacyStyledText>
+            <Flex css={LIST_CONTAINER_STYLE}>
+              {csvFilesOnUSB.length !== 0 ? (
+                csvFilesOnUSB.map(csv => (
+                  <>
+                    {csv.length !== 0 && last(csv.split('/')) !== undefined ? (
```
shlokamin, line 104: maybe put `last(csv.split('/')` into its own variable with a semantic name?

app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -73,6 +80,15 @@ export function ProtocolSetupParameters({
     )
   )
 
+  const csvFileParameter = runTimeParameters.find(
+    (param): param is CsvFileParameter => param.type === 'csv_file'
+  )
```
shlokamin, line 89: nice use of a type guard predicate 😄

app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -73,6 +80,15 @@ export function ProtocolSetupParameters({
...
+  const initialFileInfo = csvFileParameter?.file?.id ?? ''
```
shlokamin, line 90: maybe rename to `initialFileId`? if there is no id we should fallback to `null` and handle it specially. this will avoid bugs since it will be more obvious when we're missing a file id (rather than falling back to empty string, passing it to the sever instead of catching an error, etc.)

app/src/pages/ProtocolDashboard/PinnedProtocol.tsx
```diff
@@ -96,14 +107,20 @@ export function PinnedProtocol(props: {
 
   const PUSHED_STATE_STYLE = css`
     &:active {
-      background-color: ${longpress.isLongPressed ? '' : COLORS.grey50};
+      background-color: ${longpress.isLongPressed
+        ? ''
+        : enableCsvFile && isRequiredCSV
+        ? COLORS.yellow40
+        : COLORS.grey50};
```
shlokamin: can we extract this into a fn?

app/src/pages/ProtocolDashboard/PinnedProtocol.tsx
```diff
@@ -118,15 +135,26 @@ export function PinnedProtocol(props: {
       overflowWrap={OVERFLOW_WRAP_ANYWHERE}
       padding={SPACING.spacing24}
       ref={longpress.ref}
+      data-testid={`${cardSize}_pinned_protocol_card`}
```
shlokamin, line 136: this id might not be unique right? if we're not gonna use it we should delete it

app/src/pages/ProtocolDashboard/ProtocolCard.tsx
```diff
@@ -100,6 +106,10 @@ export function ProtocolCard(props: {
         analysisForProtocolCard.result === 'not-ok')) ??
     false
 
+  // ToDo (kk:06/18/2024) this part will be updated when Nick's PR is merged into edge
+  // https://github.com/Opentrons/opentrons/pull/15440
+  const isRequiredCSV = enableCsvFile && protocolName.includes('RTP')
```
shlokamin: it is merged!

app/src/pages/ProtocolDashboard/ProtocolCard.tsx
```diff
@@ -168,14 +186,20 @@ export function ProtocolCard(props: {
         ? ''
         : isFailedAnalysis
         ? COLORS.red40
+        : isRequiredCSV
+        ? COLORS.yellow40
         : COLORS.grey50};
```
shlokamin: we should extract this too, it's a long ternary!

app/src/redux/shell/types.ts
```diff
@@ -153,6 +156,14 @@ export interface NotifySubscribeAction {
   meta: { shell: true }
 }
 
+export interface SendFilePathAction {
```
shlokamin:
```suggestion
export interface SendFilePathsAction {
```

app/src/redux/shell/types.ts
```diff
@@ -153,6 +156,14 @@ export interface NotifySubscribeAction {
...
+  type: 'shell:SEND_FILE_PATH'
```
shlokamin:
```suggestion
  type: 'shell:SEND_FILE_PATHS'
```

app/src/redux/shell/types.ts
shlokamin: or actually maybe `SendDataFilePathsAction `

app/src/redux/shell/update.ts
```diff
@@ -29,3 +33,16 @@ export const getAvailableShellUpdate: (
...
   state.available && state.info ? state.info.version : null
 )
+
+export function checkMassStorage(
+  state: State
+): RobotMassStorageDeviceEnumerated {
+  return {
+    type: 'shell: ROBOT_MASS_STORAGE_DEVICE_ENUMERATED',
+    payload: {
+      rootPath: '',
```
shlokamin, line 43: `rootPath` being empty string seems suspicious... @sfoster1 im not fully understanding how it should be used. also does this belong in the `redux/shell/update` module?

koji:
- [ ] solve conflicts

app-shell-odd/src/usb.ts
```diff
@@ -47,6 +56,18 @@ const enumerateMassStorage = (path: string): Promise<string[]> =>
     )
     .catch(() => [])
     .then(flatten)
+    .then(contents => {
+      if (filterCSV) {
+        const regex = /._\w/gm
```
shlokamin, line 61: can you add a comment for what this regex is searching for?

app-shell-odd/src/usb.ts
```diff
@@ -164,3 +185,26 @@ export function watchForMassStorage(dispatch: Dispatch): () => void {
     devWatcher.close()
   }
 }
+
+export function registerFilePath(
+  dispatch: Dispatch
+): (action: Action) => unknown {
```
shlokamin, line 191: i thought we were gonna reuse `watchForMassStorage` since it's already doing a file system watch. doesnt this sort of duplicate all of the computation?

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+export function ChooseCsvFile({
+  protocolId,
+  handleGoBack,
+  parameter,
+  setParameter,
+  csvFileInfo,
+  setCsvFileInfo,
+}: ChooseCsvFileProps): JSX.Element {
+  const { t } = useTranslation('protocol_setup')
+  const csvFilesOnUSB = useSelector(getFilePaths).payload.filePaths ?? []
```
shlokamin, line 45: is `getFilePaths` an action creator or a selector? if its meant to be a selector, we need to rework it to take in state and return file paths (rather than an action)

app/src/redux/shell/actions.ts
```diff
@@ -121,3 +125,9 @@ export const notifySubscribeAction = (
   },
   meta: { shell: true },
 })
+
+export const getFilePaths = (state: State): SendFilePathsAction => ({
+  type: SEND_FILE_PATHS,
+  payload: { filePaths: state.shell.filePaths },
```
shlokamin, line 131: is this meant to be a selector or an action?

app/src/organisms/ProtocolSetupParameters/ChooseCsvFile.tsx
koji: Actually, this is for a following PR

app-shell-odd/src/usb.ts
```diff
@@ -164,3 +185,26 @@ export function watchForMassStorage(dispatch: Dispatch): () => void {
...
+  return function handleAction(action: Action) {
+    switch (action.type) {
+      case ROBOT_MASS_STORAGE_DEVICE_ENUMERATED:
+        void enumerateMassStorage(action.payload.rootPath, true).then(
+          csvFilePaths => {
+            dispatch(sendFilePaths(csvFilePaths))
+          }
+        )
```
shlokamin, line 199: in the system update reducer, we respond to `shell:ROBOT_MASS_STORAGE_DEVICE_ENUMERATED` and filter out all of the provided file paths with ONLY relevant update files (system zips). 

in the usb case, can we do something similar? we could respond to `shell:ROBOT_MASS_STORAGE_DEVICE_ENUMERATED` (like we already are), but have a special function (similar to `getLatestMassStorageUpdateFiles` that only grabs CSV files for us. that way we can draw a distinction between getting ALL files in the usb dir vs specific files relevant to a particular redux module (system update, or csv files)

shlokamin: something like `getLatestMassStorageCsvFiles` or `getLatestMassStorageUpdateFiles`.

and that function's job is just to filter for relevant CSV files

app-shell-odd/src/usb.ts
shlokamin: maybe we can call this `registerDataFiles` or `registerCsvFiles` because this reducer will be searching specifically for protocol data files

app-shell-odd/src/usb.ts
koji: trying the above method, but action.payload.filePaths is empty  - `[]` without calling `enumerateMassStorage`.

shlokamin: `robotMassStorageDeviceEnumerated` gets called by `enumerateMassStorage`, which means that as long as you respond to an action of type `ROBOT_MASS_STORAGE_DEVICE_ENUMERATED` by calling this new `getLatestMassStorageCsvFiles` function you should be good.

is the action of type `ROBOT_MASS_STORAGE_DEVICE_ENUMERATED` getting dispatched?

koji: yes
rootPath has the path but filePaths don't have any path.

unplug the flash drive from Flex -> plug into the flash drive on Flex

```
Jun 28 21:12:10 Flex1234567890 opentrons-robot-app.sh[5458]: Mass storage device /media/sda1 removed but this was not an update source
Jun 28 21:12:16 Flex1234567890 opentrons-robot-app.sh[5458]: New mass storage device sda1 detected
Jun 28 21:12:17 Flex1234567890 opentrons-robot-app.sh[5458]: shell:ROBOT_MASS_STORAGE_DEVICE_ENUMERATED []
Jun 28 21:12:17 Flex1234567890 opentrons-robot-app.sh[5458]: no updates found in mass storage device
```

koji:
this will be taken over by #15551

PR #15406, opened 2024-06-13
feat(hardware): create new move steps in liquid_probe

aaron-kulkarni:
Create raise_z_axis and lower_plunger steps, which will aid in LLD.

fix EXEC-514

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (dismissed):
Looks good, but
- we should probably have some tests
- and let's make sure to test on a robot

aaron-kulkarni:
I think the test that is failing is unrelated to my changes? I could be wrong.

ryanthecoder (approved):
Looks good when we tested it on the robot. LGTM

PR #15407, opened 2024-06-13
fix(protocol-designer): reset the touchTip checkbox when labware changes

jerader:
closes RESC-273

# Overview

This is an interested bug. Basically, a user selected a labware where touch tip is allowed and checked the touch tip box. Then the user changed the labware to one where touch tip is not allowed. The touch tip check box was disabled but not actually de-selected. This resulted in an unwanted erroneous `touchTip` command to be generated in step-generation. 

The fix is not super straightforward though. What I changed here is when the aspirate, dispense, or mix labware changes, the touch tip checkbox field is reset. This fix is fine however if a protocol is already experiencing this error, it will not be automatically fixed upon importing their protocol.

The more involved fix (simply code-wise but has big implications) would be ensuring that a touch tip command is only generated if the touch tip checkbox is selected AND there is a height is selected. But that would mean changing the current behavior of PD which I do not think is a good idea.

# Test Plan

Upload the attached file to protocol designer in production. Navigate to the 15th step in the timeline (which is a transfer step) and open the advanced settings. change the dispense labware to `Amplicon PCR` see that the touch tip field is selected.

Now, upload the same file to protocol designer on this branch, navigate to the timeline, open the 15th step again. Open up the advanced settings and change the dispense labware to `Amplicon PCR` . See that the touch tip field is NOT selected. (this is because the setting has been properly reset).

[16S Metagenomik Amplicon PCR Clean-Up (manuelle beads zugabe) Optimiert_10-06-2024.json](https://github.com/user-attachments/files/15825137/16S.Metagenomik.Amplicon.PCR.Clean-Up.manuelle.beads.zugabe.Optimiert_10-06-2024.json)

# Changelog

- reset both the `moveLiquid` and `mix` forms touch tip checkbox fields if labware changes

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/steplist/formLevel/handleFormChange/dependentFieldsUpdateMix.ts
+++ b/protocol-designer/src/steplist/formLevel/handleFormChange/dependentFieldsUpdateMix.ts
@@ -30,7 +30,11 @@ const updatePatchOnLabwareChange = (
   const pipetteId = appliedPatch.pipette
   return {
     ...patch,
-    ...getDefaultFields('mix_mmFromBottom', 'mix_touchTip_mmFromBottom'),
+    ...getDefaultFields(
+      'mix_mmFromBottom',
+      'mix_touchTip_mmFromBottom',
+      'mix_touchTip_checkboxƒ'
+    ),
     wells: getDefaultWells({
       labwareId: appliedPatch.labware,
       pipetteId,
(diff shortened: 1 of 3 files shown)


koji:
@jerader 

>Upload the attached file to protocol designer in production.

you mean the protocol Matthew attached to the ticket?

jerader:
> you mean the protocol Matthew attached to the ticket?

@koji oops yes my bad! I just attached it here too.

koji (approved):
the changes match what you mentioned in overview.
sandbox worked as expected.

PR #15408, opened 2024-06-13
refactor(api): Prepare (almost) any PAPI command to fail recoverably

SyntaxColoring:
# Overview

This closes EXEC-339.

# Test Plan

* [x] Run some Python protocols with the `enableErrorRecoveryExperiments` feature flag disabled, and ensure they continue to fail fatally if something goes wrong.
* [x] Run some Python protocols with the `enableErrorRecoveryExperiments` feature flag *enabled,* and ensure `overpressure` and `tipPhysicallyMissing` errors are still nonfatal, and everything else is still fatal.
* [x] Run analysis snapshot tests on this PR and ensure nothing changed.
  * #15414 

# Changelog

## [Commits up to 36adc9:](https://github.com/Opentrons/opentrons/pull/15408/files/36adc941a95b9dd3e47a92c8b9e25883d5beb04a) 

Delete the layer of boilerplate where `SyncClient` has one method to duplicate every Protocol Engine command. Instead, just make the `opentrons.protocol_api` code construct the command request itself and pass it in.

This shifts verbosity into `opentrons.protocol_api`, so to mitigate that, `SyncClient` takes just the `CommandParams` and automatically maps it into the full `CommandCreate`, via runtime `Union`-unpacking hijinks.

## [Commits 36adc9..bce3f3](https://github.com/Opentrons/opentrons/pull/15408/files/36adc941a95b9dd3e47a92c8b9e25883d5beb04a..bce3f35e2090e682c49713a7566ba7d2addf2fbb):

Update `opentrons.protocol_api` to use the new `SyncClient` interface, added above, *and to use its mode that allows error recovery, wherever possible.* This means we will let the protocol thread continue if almost any random command fails—as long as `opentrons.protocol_engine` decides the error is recoverable. The holdouts are `loadLabware`/`loadModule`/`loadPipette`, which will continue to raise exceptions.

Because error recovery is still [mostly disabled down in the `opentrons.protocol_engine` layer](https://github.com/Opentrons/opentrons/blob/5ce130b4f18a041c666bf401a1c7d9fdea2461a6/api/src/opentrons/protocol_engine/error_recovery_policy.py#L56-L74), this should not cause any behavioral change yet.

These changes are almost entirely mechanical and [script-assisted](https://gist.github.com/SyntaxColoring/a09bbe6e820d4002fc356fed6cab53f5). The only real changes were that some of the Thermocycler `run_profile()` code got shifted around, and `ProtocolContext.comment()` emits real Protocol Engine `comment` commands now instead of legacy `custom` commands. The `comment` thing requires frontend followup in #15409.

# Review requests

None in particular.

# Risk assessment

Medium.

There's a risk that some of the boilerplate `SyncClient` methods had different parameter defaults than the underlying Protocol Engine types, which would cause a sneaky behavioral change. Analysis snapshot tests *should* be able to detect this.

There's also some risk that something went wrong with my scripting and I erroneously replaced `aspirate` commands with `aspirateInPlace` commands or something like that.


sfoster1 (approved):
I absolutely love this, especially the runtime type stuff!

TamarZanzouri (approved):
wow this is so clean! I love this change!

PR #15413, opened 2024-06-13
feat(shared-data): bump gripper model version and use new force polynomial table for new motor

ahiuchingau:
# Overview
The manufacturer is changing the brushes on the gripper motor, so we'll need to use a new force polynomial for the new gripper models (v1.3). In order to use the new force function, you will need to make sure the serial number is in the format of `GRPV13xxxxx`.


api/src/opentrons/config/gripper_config.py
```diff
@@ -24,7 +24,12 @@ def info_num_to_model(num: str) -> GripperModel:
     # PVT will now be 1.2
```
y3rsh, line 24: Is this still a relevant comment?

sfoster1 (approved):
Looks good to me but @ahiuchingau don't you also have to add handling for this in the serial number writer

PR #15415, opened 2024-06-13
feat(robot-server): create a new database table for data files

sanni-t:
Closes AUTH-423

# Overview

Adds new database table to store data files' metadata.
Adds it to the unreleased schema 5.

Contents of the `data_files` table will be accessed by the csv file RTP in `analysis` & `run` tables.

# Test Plan

- [x] Tested that the new table gets created in dev server

# Risk assessment

None to the code, really.
Because this is an in-between schema update without a version bump, we just need to make sure to wipe out the existing database on the robot before testing.


jbleon95 (approved):
✅

PR #15416, opened 2024-06-13
chore(ci): snapshot analyses test runs on PR to /api

y3rsh:
## Update Analyses Snapshots Workflow

- [x] Run on PRs to /api
- [x] Open a PR into the PR branch if snapshots would change
- [x] Comment on the PR if a PR is opened

### DIFF
--- a/.github/workflows/analyses-snapshot-test.yaml
+++ b/.github/workflows/analyses-snapshot-test.yaml
@@ -16,6 +16,8 @@ on:
   pull_request:
     paths:
       - 'api/**'
+      - 'shared-data/**/*'
+      - '!shared-data/js/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/analyses-snapshot-test.yaml
+++ b/.github/workflows/analyses-snapshot-test.yaml
@@ -17,6 +17,10 @@ on:
     paths:
       - 'api/**'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-and-test:
     timeout-minutes: 15
--- a/.github/workflows/analyses-snapshot-test.yaml
+++ b/.github/workflows/analyses-snapshot-test.yaml
@@ -13,14 +13,17 @@ on:
         default: 'edge'
   schedule:
     - cron:  '26 7 * * *' # 7:26 AM UTC
+  pull_request:
+    paths:
+      - 'api/**'
 
 jobs:
   build-and-test:
     timeout-minutes: 15
     runs-on: ubuntu-latest
     env:
-      TARGET: ${{ github.event.inputs.TARGET || 'edge' }}
-      TEST_SOURCE: ${{ github.event.inputs.TEST_SOURCE || 'edge' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/analyses-snapshot-test.yaml
```diff
@@ -13,14 +13,21 @@ on:
         default: 'edge'
   schedule:
     - cron:  '26 7 * * *' # 7:26 AM UTC
+  pull_request:
+    paths:
+      - 'api/**'
```
SyntaxColoring, line 18: Does this need to include `shared-data` too?

.github/workflows/analyses-snapshot-test.yaml
```diff
@@ -70,3 +78,15 @@ jobs:
           branch: 'app-testing/${{ env.TARGET }}-from-${{ env.TEST_SOURCE}}'
           base: ${{ env.TEST_SOURCE}}
 
+    - name: Comment on PR
+      if: failure() && github.event_name == 'pull_request'
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const message = 'A PR has been opened to address analyses snapshot changes. Please review the changes here: https://github.com/${{ github.repository }}/pull/${{ steps.create-pull-request.outputs.pull-request-number }}';
```
SyntaxColoring, line 88: If a PR requires snapshot changes, would we not want those snapshot changes to be in that same PR?

In other words, do we not want to treat this the same way as any other test failure?

SyntaxColoring (commented):
Cool!

.github/workflows/analyses-snapshot-test.yaml
y3rsh: I was not sure what to include.  Should I add `shared-data/python/**` ? or `shared-data/**`

.github/workflows/analyses-snapshot-test.yaml
y3rsh: We do.  The PR to update the snapshots will be opened into the  `github.head_ref` branch for the PR.
So you would look at the changes and if they make sense you merge that PR into your branch so it becomes a part of the change into edge.
This code generates a comment on your PR so you have a reminder/link to the PR in addition to the failed "test".

.github/workflows/analyses-snapshot-test.yaml
SyntaxColoring: Hm, good question. `api-test-lint-deploy.yaml` does:

```
      - 'shared-data/**/*'
      - '!shared-data/js/**'
```

Which seems reasonable to me. Let's go with that.

.github/workflows/analyses-snapshot-test.yaml
SyntaxColoring: Ohhh interesting. Will my PR get a red ❌ on it if it has an unresolved snapshot diff?

.github/workflows/analyses-snapshot-test.yaml
y3rsh, line 18:
```suggestion
      - 'api/**'
      - 'shared-data/**/*'
      - '!shared-data/js/**'
```

.github/workflows/analyses-snapshot-test.yaml
y3rsh: Yes it would.

SyntaxColoring (approved):
Yay!

PR #15418, opened 2024-06-13
chore(api): Ignore AbsorbanceReaderContext errors when building PAPI docs

SyntaxColoring:
# Overview

This silences this Sphinx warning, which is currently causing the docs build to fail:

> .../api/src/opentrons/protocol_api/protocol_context.py:docstring of opentrons.protocol_api.ProtocolContext.loaded_modules:: WARNING: py:class reference target not found: opentrons.protocol_api.module_contexts.AbsorbanceReaderContext

Sphinx wants to link to `AbsorbanceReaderContext` from the `loaded_modules` return type docs, but it can't because `AbsorbanceReaderContext` is, intentionally for now, excluded from the docs.

# Review requests

Do we need to make a ticket somewhere to undo this later?

# Risk assessment

Low.

### DIFF
--- a/api/docs/v2/conf.py
+++ b/api/docs/v2/conf.py
@@ -445,5 +445,6 @@ nitpick_ignore_regex = [
     ("py:class", r".*protocol_api\.config.*"),
     ("py:class", r".*opentrons_shared_data.*"),
     ("py:class", r".*protocol_api._parameters.Parameters.*"),
+    ("py:class", r".*AbsorbanceReaderContext"),  # shh it's a secret (for now)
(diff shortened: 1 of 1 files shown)


api/docs/v2/conf.py
```diff
@@ -445,5 +445,6 @@
     ("py:class", r".*protocol_api\.config.*"),
     ("py:class", r".*opentrons_shared_data.*"),
     ("py:class", r".*protocol_api._parameters.Parameters.*"),
+    ("py:class", r".*AbsorbanceReaderContext"),  # shh it's a secret (for now)
```
ecormany, line 448: lol perfect

ecormany (approved):
Yeah, this seems the right way to do this. It's too much risk to actually change the union of module classes now, and then have to remember to change it back later. Even if we forget to remove this, docs will continue to build and the API methods will work.

But to make sure we don't, I've ticketed it: RTC-483.

PR #15422, opened 2024-06-14
fix(protocol-designer): add missing prop to getPipetteCapacity usage

jerader:
closes AUTH-489 and RESC-272

# Overview

this is a copy of this PR https://github.com/Opentrons/opentrons/pull/15360, i'm pretty sure it is causing this escalations error https://opentrons.atlassian.net/browse/RESC-272. The error is when 2 tipracks are selected and the tipracks are on the deck, the pipetting step thinks there is no tiprack on the deck with the default tiprack. there is a work around by changing the tipracks a few times but its an annoying workaround.

# Test Plan

create a flex or ot-2 protocol and add 2 tipracks. Go to the deck map and add a transfer or mix step with the default tip selected. Save the step and see that there is no timeline error.

# Changelog

- add missing prop to util

# Review requests

see test plan

# Risk assessment

low


koji (approved):
lgtm

PR #15424, opened 2024-06-14
feat(abr-testing): Match JIRA component to ticket based off error message

rclarke0:
# Overview

Adds corresponding jira component to ticket.

# Test Plan

# Changelog

Added `match_error_to_component()` in `abr_robot_error` .
function  gets all components with corresponding project board. If the error message contains a word that matches with the components, the component is added to the ticket.

Small change - if error message is PythonException, the detail of the error is added to the ticket title instead to make it more descriptive. 

# Review requests

# Risk assessment


sfoster1 (approved):
That is extremely nice, nice work!

PR #15419, opened 2024-06-14
refactor(api): Emit proper `comment` commands from legacy protocols

SyntaxColoring:
# Overview

For historical reasons, Python Protocol API `ProtocolContext.comment("...")`s have always been reported as "legacy" `custom` commands in the analysis engine and HTTP API. https://github.com/Opentrons/opentrons/pull/15408 changes them to proper Protocol Engine `comment` commands when the Python protocol has `apiLevel` ≥2.14. This PR makes the same change for Python protocols with `apiLevel` ≤2.13, just for the sake of uniformity.

# Test Plan

I'll test this in https://github.com/Opentrons/opentrons/pull/15409.

# Review requests

None in particular.

# Risk assessment

Low.

PR #15426, opened 2024-06-14
refactor(api): Emit proper `comment` commands from legacy protocols

SyntaxColoring:
Exact same changes as PR #15419, which I accidentally closed.


TamarZanzouri (approved):
lol same comment then :-)

PR #15428, opened 2024-06-14
refactor(app): Add command component

sfoster1:
This PR adds a `Command` component that can be used to render a command from a runlog, and stories for it. This will be used in error recovery.

Storybook: https://s3-us-west-2.amazonaws.com/opentrons-components/exec-493-failed-runlog-step/index.html?path=/docs/app-molecules-command-command--docs

Figma: https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=1929-43796&t=eaSbKoVHbTKn7HS9-4

There were also two other refactors:
- Move `CommandText` to be part of the new `Command` molecule (it was a little weird that `CommandText` was an organism, since it didn't use any molecules and didn't use any react hooks or state)
- Move `CommandIcon` to be part of the new `Command` molecule because it feels like it's part of that
- Add stories for commandtext and command icon

This new molecule still exposes the two older components but also the new `Command` component, which is used to present commands with color tinting for outcome and an integrated skeleton state.

## Review requests
- Is this going to work with the ODD animation fixes from way back when? I didn't actually put the new component in there, so we can merge even if the answer is "no", but we should add some todos at least if not integrate the animation fixtures
- There's a sort of unholy mix of js and css responsiveness stuff here, mostly because of the way `CommandText` uses its own flex for thermocycler profiles that means that css in an invoking component won't actually get to the underlying text elements. Is there a better way to do this?
- There is, specifically, a fairly serious number of css misdemeanors due to the combination of that thermocycler flex and the vendor-specific css required for multiline text clipping with ellipses. `!important` is involved. I am sorry
- I require you to look at how cool the stories that let you select commands from a runlog analysis in `__fixtures__` is

When looking in storybook, keep in mind that you have to view a story (not the docs page) to switch viewports to the ODD, which will make the text styling correct and apply the two-line constraint.

Closes EXEC-493


mjhuff (approved):
Looks great! This looks ready to merge with passing CI.

- This won't interfere with the ODD CSS optimizations.
- There probably is an alternative way to do this, but I can't see it off the top of my head. Yeah, that's annoying though. Maybe we can leave a TODO here?
- The stories are actually very cool 😎

PR #15429, opened 2024-06-14
fix(step-generation): update liquid state 96-channel and reservoir

jerader:
closes RESC-278

# Overview

This fixes a gnarly edge case step-generation liquid state bug. basically when the aspirating command was updating the robot state, it didn't account for if you have a pipette aspirating multiple tips from the same well and multiple wells. For example a 96-channel aspirating from a 12 well reservoir (8 channels per reservoir). So the liquid state was not updating properly, showing incorrect liquids on the deck map and incorrect volumes. This PR fixes that

# Test Plan

Upload the attached protocol to PD in production. Click on the final deck state in the timeline and see that the 384 well plate only has some of the wells colored with a liquid (this is incorrect since the transfer was with a 96 channel).

Then upload the attached protocol to PD on this branch. Click on the final dekc state in the timeline and see that the 384 well plate has the correct 96 wells colors with a liquid.

[Nest 15mL reproduce.json](https://github.com/user-attachments/files/15842578/Nest.15mL.reproduce.json)

# Changelog

- add logic for 96 channel + reservoir, it is similar to the logic if you have an 8 channel aspirating from 1 well

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/getNextRobotStateAndWarnings/forAspirate.ts
+++ b/step-generation/src/getNextRobotStateAndWarnings/forAspirate.ts
@@ -23,6 +23,7 @@ export function forAspirate(
   const nozzles = robotState.pipettes[pipetteId].nozzles
   const pipetteSpec = invariantContext.pipetteEntities[pipetteId].spec
   const labwareDef = invariantContext.labwareEntities[labwareId].def
+  const isReservoir = labwareDef.metadata.displayCategory === 'reservoir'
   const channels = nozzles === COLUMN ? 8 : pipetteSpec.channels
 
   const { allWellsShared, wellsForTips } = getWellsForTips(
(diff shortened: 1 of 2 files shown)


koji (approved):
lgtm

step-generation/src/getNextRobotStateAndWarnings/forAspirate.ts
```diff
@@ -81,6 +82,56 @@ export function forAspirate(
...
   }
 
+  //  all wells in the reservoir are being used in this case but 8 channels per well
+  if (channels === 96 && isReservoir) {
+    //  for each well the 96 channels are aspirating into
+    wellsForTips.forEach(well => {
+      const sourceLiquidState = liquidState.labware[labwareId][well]
+      const isOveraspirate =
+        volume * 8 > getLocationTotalVolume(sourceLiquidState)
+
```
shlokamin, line 92: let's generify this to handle any reservoir with any number of columns

PR #15427, opened 2024-06-14
refactor(app): remove implicit width constraints from WellSelection

mjhuff:
Closes [EXEC-535](https://opentrons.atlassian.net/browse/EXEC-535)

# Overview

`SelectionRect` within `WellSelection` contains a `Flex` with a `width=75%`, which results in the entire `WellSelection` SVG to render at 75% width. This CSS constraint should probably be placed outside of the component itself, as it makes manipulating CSS difficult.

# Test Plan

- Verified Quick Transfer `WellSelection` CSS is consistent after changes.

Screenshot 2024-06-14 at 3 02 44 PM

# Risk assessment
low

[EXEC-535]: https://opentrons.atlassian.net/browse/EXEC-535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #15431, opened 2024-06-17
feat(app): Error Recovery: Add overpressure while aspirating route

mjhuff:
Closes [EXEC-522](https://opentrons.atlassian.net/browse/EXEC-522)

# Overview

https://github.com/Opentrons/opentrons/assets/64858653/5e89d51e-dc41-4d15-8d33-2a2f76aaf871

Adds the "Overpressure while aspirating" route, which also sets the groundwork for wiring up the other overpressure routes. Most of this PR is just building out components and utilities necessary to render out everything above, but in terms of high notes and major changes: 

### Peripherals involved in the failedCommand

All Overpressure routes require knowledge of which pipette was used in the failed liquid handling command, and this is known as the `failedCommandPipette`. The name of the labware from which the tip(s) involved in the failed command were picked up is known as the `pickUpTipLabwareName` - this is also important. A lot of the utils built out here work towards obtaining the data necessary to render relevant info (and/or feed it to the recovery deck map). Importantly, there's no great of way which pipette and labware were involved in the error without querying for all commands for the entire run up to this point (and that's one reason these utils are complex), but [EXEC-539](https://opentrons.atlassian.net/browse/EXEC-539) will solve this.

### Recovery Deck Map

There currently does not exist a deck map for handling runRecord based deck generation, so we build one here. There's some neat highlighting around the `pickUpTipLabwareName`.  This required a good number of utils, but it's honestly just a bunch of vanilla Javascript method chaining and property-checking.

### Tip Selector

Design has decided to utilize the `WellSelector` found in Quick Transfer Flows with a hard-coded deck definition, preferring one with big touch targets. There's some abstraction work that can and should be done here, but I think it's worthwhile to consult design on this one. Note that the `TipSelector` is smaller than designs intend, and this is solved by #15427.

There's a known issue with `TipSelector` (that doesn't exist in `WellSelector`...I think?), [EXEC-534](https://opentrons.atlassian.net/browse/EXEC-534). 

### Recovery Footer Buttons

Nothing that interesting here, outside of the fact that there is an optional third button now!

### `SelectedRecoveryOption`

This replaces `previousRoute`, which I developed a bit too far ahead of time...oops. The idea with `previousRoute` was that it would help us know from which route the client was at _before_ Drop Tip flows were launched, so we could route them to the next step after Drop Tip flows. We wanted something relatively implicit, but `previousRoute` is hard to make work correctly, because you have routes like "robot is moving/dropping tips/retrying failed command", etc, and all sort of other edge cases that actually set the previousRoute to something we don't care about.  After spending more time looking at designs and thinking this through, what we really want is just to know the current _recovery_ option that the user selected. If a user selects "Retry with new trips", we cache that, then we can send them off to the drop tip flows or wherever while still preserving the routing intent. This is just as implicit in practice, too. Only `SelectRecoveryOption` ever has to hold the logic containing the set functionality.

One thing that might seem somewhat odd at first (but I think makes the most sense, ultimately), is that if the first step of a flow (ex, `RetryNewTips`) is actually Drop tip flows, do NOT route the user to DT initially, but route them to `RetryNewTips`, and then have `RetryNewTips` route the user to DT. This routing is more predictable and less error prone than the alternative.

### `'ovepressure' errorType`

The BE will eventually have to delineate between types of overpressures, as we currently have this error type to represent all overpressures.

### `RetryWithNewTips`

There is a placeholder here as designs are currently borked, and this is somewhat dependent on step count work. I've made a follow up ticket for it here [EXEC-549](https://opentrons.atlassian.net/browse/EXEC-549).

# Test Plan
- Here is the protocol I used in the video. [overpressure_test_modules.py.zip](https://github.com/user-attachments/files/15866262/overpressure_test_modules.py.zip)
- Pick up a tip, then cause an overpressure while aspirating. I would stick my finger beneath the pipette to trigger it consistently. Use a PVT pipette 😃 
- Verify via video.

# Changelog
- Added "overpressure while aspirating" route to Error Recovery.

# Risk assessment
low - behind ff

[EXEC-522]: https://opentrons.atlassian.net/browse/EXEC-522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-539]: https://opentrons.atlassian.net/browse/EXEC-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/ErrorRecoveryFlows/shared/RecoveryMap.tsx
mjhuff, line 1: There are a lot of one-off design decisions in this file. The `7.04px` does not convert to any meaningful rem number, and I also refuse to convert `3px` and `4px` to rem for one-off border and box-shadow. I also don't think it makes a ton of sense to use a constant for a one-off value that coincidentally makes use of the constant. I will die on this hill!

app/src/organisms/ErrorRecoveryFlows/index.tsx
```diff
@@ -63,36 +90,41 @@ export interface ErrorRecoveryFlowsProps {
...
   }
 
   return (
     <>
       {showERWizard ? (
-        <ErrorRecoveryWizard {...props} {...recoveryUtils} />
+        <ErrorRecoveryWizard
+          {...restProps}
+          {...recoveryUtils}
+          robotType={robotType}
```
mjhuff, line 122: We pass in `robotType` here. It's not used now, but it will be useful verryyyy soon with the desktop flows.

app/src/organisms/ErrorRecoveryFlows/constants.ts
```diff
@@ -6,6 +6,7 @@ import type { StepOrder } from './types'
 
 export const ERROR_KINDS = {
   GENERAL_ERROR: 'GENERAL_ERROR',
+  OVERPERSSURE_WHILE_ASPIRATING: 'OVERPRESSURE_WHILE_ASPIRATING',
```
sfoster1:
```suggestion
  OVERPRESSURE_WHILE_ASPIRATING: 'OVERPRESSURE_WHILE_ASPIRATING',
```

app/src/organisms/ErrorRecoveryFlows/hooks/getErrorKind.ts
```diff
@@ -4,6 +4,8 @@ import type { ErrorKind } from '../types'
 
 export function getErrorKind(errorType?: string): ErrorKind {
   switch (errorType) {
+    case 'overpressure':
+      return ERROR_KINDS.OVERPERSSURE_WHILE_ASPIRATING
```
sfoster1:
```suggestion
      return ERROR_KINDS.OVERPRESSURE_WHILE_ASPIRATING
```

app/src/organisms/ErrorRecoveryFlows/hooks/useERUtils.ts
```diff
@@ -32,28 +47,75 @@ export function useERUtils({
...
-  const { recoveryMap, setRM, trackExternalMap } = useRecoveryRouting()
-  const previousRoute = usePreviousRecoveryRoute(recoveryMap.route)
-  const tipStatusUtils = useRecoveryTipStatus(runId, isFlex)
+  const { data: attachedInstruments } = useInstrumentsQuery()
+  const { data: runRecord } = useNotifyRunQuery(runId)
+  // TODO(jh, 06-04-24): Refactor the utilities that derive info
+  // from runCommands once the server yields that info directly on an existing/new endpoint.
+  const { data: runCommands } = useNotifyAllCommandsQuery(runId, {
+    cursor: 0,
+    pageLength: 999,
```
sfoster1, line 60: is this meant to be implicitly "all", or do we have a specific habit of using 999

app/src/organisms/ErrorRecoveryFlows/hooks/useErrorName.ts
```diff
@@ -1,12 +1,16 @@
...
+import { ERROR_KINDS } from '../constants'
+
 import type { ErrorKind } from '../types'
 
 // Returns the user-facing name of the errorKind.
 export function useErrorName(errorKind: ErrorKind): string {
   const { t } = useTranslation('error_recovery')
 
   switch (errorKind) {
+    case ERROR_KINDS.OVERPERSSURE_WHILE_ASPIRATING:
```
sfoster1:
```suggestion
    case ERROR_KINDS.OVERPRESSURE_WHILE_ASPIRATING:
```

sfoster1 (approved):
Looks good to me once tested... ugh fractional px

app/src/organisms/ErrorRecoveryFlows/hooks/useERUtils.ts
mjhuff: It's just a me getting a reasonably large window until we don't show text. I talked to @SyntaxColoring and he feels this is ok until we get the server-side work merged in for this.

app/src/organisms/ErrorRecoveryFlows/hooks/getErrorKind.ts
mjhuff: Is there a change here? Is Github being Github?

app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
```diff
@@ -0,0 +1,229 @@
...
+
+// Get the name of the latest labware used by the failed command's pipette to pick up tips, if any.
+export function getPickUpTipLabwareName(
+  protocolAnalysis: ErrorRecoveryFlowsProps['protocolAnalysis'],
+  recentRelevantPickUpTipCmd: Omit<PickUpTipRunTimeCommand, 'result'> | null,
+  runRecord?: Run
+): string | null {
+  const lwDefsByURI = getLoadedLabwareDefinitionsByUri(
+    protocolAnalysis?.commands ?? []
+  )
```
SyntaxColoring, line 186: I don't think we want or need to use the protocol analysis to source this, do we? We could use [`GET /runs/{id}/loaded_labware_definitions`](https://docs.opentrons.com/http/api_reference.html#tag/Run-Management/operation/get_run_loaded_labware_definitions_runs__runId__loaded_labware_definitions_get) instead.

app/src/organisms/ErrorRecoveryFlows/hooks/useERUtils.ts
SyntaxColoring: Yeah. Nitpick: I think we should leave a comment explaining that it's lower-bounded by wanting to show useful information as much as possible and upper-bounded by server performance limitations, though.

app/src/organisms/ErrorRecoveryFlows/hooks/getErrorKind.ts
SyntaxColoring: p-e-r vs. p-r-e 🙂

app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
```diff
@@ -0,0 +1,229 @@
...
+  ) {
+    return null
+  }
+
+  const failedCmdPipetteId = failedCommand.params.pipetteId
+
+  // Reverse iteration is faster as long as # recovery commands < # run commands.
+  const failedCommandIdx = runCommands.data.findLastIndex(
+    command => command.key === failedCommand.key
+  )
```
SyntaxColoring, line 98: Instead of having to scan through the commands list, this is available in `GET /runs/{id}` as `data.links.currentlyRecoveringFrom.meta.index`.

This might be moot if this whole thing will be replaced when we have EXEC-539. Your call whether it's worth optimizing.

app/src/organisms/ErrorRecoveryFlows/hooks/getErrorKind.ts
mjhuff: @SyntaxColoring thank you!

app/src/organisms/ErrorRecoveryFlows/constants.ts
mjhuff: Ohhh, thanks!

app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
mjhuff: We do, you're right! I actually created a ticket for this a couple days ago! [EXEC-525](https://opentrons.atlassian.net/browse/EXEC-525). Looks like the FE doesn't have a hook for this. Will address in a follow-up.

app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
mjhuff: Yeah, didn't realize this as we didn't have this property in shared-data. No sweat though. I think I'll hold off on optimizing since it will be moot shortly.

PR #15433, opened 2024-06-17
refactor(labware-library): increase margin between checkbox and descr…

jerader:
…iption

closes AUTH-515

# Overview

Look at the ticket and see that the margin between checkbox and description was really small. This PR increases it a bit and adds a gridGap to the module stacking offset so it matches the adapter stacking offset

Screenshot 2024-06-17 at 13 25 59

# Test Plan

Open the labware library dev env and navigate to labware creator on the top left navigation. Create a new definition for a well plate. Add 16 for the height, 12 evenly spaced columns and 8 evenly spaced rows, change the well bottom shape to v-bottom. Then you should see 2 adapters and 2 module selections under the Stacking Offsets. Click on a module checkbox and see that the description text has enough top margin

# Changelog

- adjust top margin and add grid gap

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/labware-library/src/labware-creator/components/sections/StackingOffsets.tsx
+++ b/labware-library/src/labware-creator/components/sections/StackingOffsets.tsx
@@ -241,8 +241,8 @@ export function StackingOffsets(): JSX.Element | null {
                       isChecked ? (
                         <div
                           style={{
-                            marginTop: '-1.4rem',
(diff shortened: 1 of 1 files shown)


ncdiehl11 (approved):
looks great! happy stacking made its way into LC

PR #15434, opened 2024-06-17
feat(api, robot-server): return parameter-value-required result for analyses that reference un-set csv parameters

jbleon95:
# Overview

Closes AUTH-418 and addresses part of AUTH-419

This PR adds a new `result` type to analysis for the circumstance where you reference a CSV parameter that is not set (since there are no local defaults currently). In this scenario, the result will read `parameter-value-required` instead of `ok` or `not-ok`. Since local analysis did not previously have this field, it was added to local/cli analysis as well.

In order to do this, a new error (and accordant error code) was added that is raised when a csv parameter is accessed when it has not been associated with a file. To accomplish this part, a basic CSV parameter interface was added that currently stores a file (accessible via `.file` with the parameter) and allows access to the rows (via `.rows`). Trying to use either one of these without a file will raise a `RuntimeParameterRequired` error.

This error is also re-exported to the `protocol_api` for public use in the API as `RuntimeParameterRequiredError`. This can be caught and handled, allowing a user to potentially provide their own in-protocol alternate if they wish to have local analysis without providing a file via a local dev-server.

# Test Plan

Tested that this protocol returns with a `parameter-value-required` result in the analysis.

```
metadata = {
    'protocolName': 'CSV use parameter test',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}


    parameters.add_int(
        display_name="Sample count",
        variable_name="sample_count",
        default=2,
        minimum=1,
        maximum=6,
    )
    parameters.add_str(
        display_name="Pipette Name",
        variable_name="pipette",
        choices=[
        ],
        default="flex_1channel_50",
    )
    parameters.add_csv_file(
        display_name="CSV Data",
        variable_name="csv_data",
        description="CSV file containing labware and volume information."
    )



def run(context):
    SAMPLE_COUNT = context.params.sample_count
    PIPETTE_NAME = context.params.pipette

    pipette = context.load_instrument(PIPETTE_NAME, mount="left", tip_racks=[tip_rack])
    csv_file = context.params.csv_data.file
```

and that this correctly passes analysis

```

metadata = {
    'protocolName': 'CSV use parameter test w/ try except clause',
}

requirements = {
    "robotType": "Flex",
    "apiLevel": "2.18"
}


    parameters.add_csv_file(
        display_name="CSV Data",
        variable_name="csv_data",
        description="CSV file containing labware and volume information."
    )



def run(context):
    try:
    	csv_file = context.params.csv_data.file
    except protocol_api.RuntimeParameterRequiredError:
    	pass
```

# Changelog

- added new analysis `result` type of `parameter-value-required`
- added `result` field to local CLI analysis
- added a basic CSV interface to access file and rows. If there is no file set, accessing those will raise a `RuntimeParameterRequired error.

# Review requests

Had to move some of the `code_in_error_tree` logic around to make it importable. If that belong in a better place than `protocol_engine`, I can move it, but left it there for now.

# Risk assessment

Medium-lowish, its a change to analysis (and particularly local analysis with the new field), but its adding new information and shouldn't break anything, even before official support of the `parameter-value-required` field.


api/src/opentrons/protocol_engine/protocol_engine.py
```diff
@@ -588,36 +588,35 @@ async def use_attached_modules(
...
-                return True
-            return any(
-                ProtocolEngine._code_in_error_tree(wrapped, code)
-                for wrapped in root_error.wrappedErrors
-            )
 
-        # From here we have an exception, can just check the code + recurse to wrapped errors.
-        if root_error.code == code:
+# TODO(tz, 7-12-23): move this to shared data when we dont relay on ErrorOccurrence
+def code_in_error_tree(
```
SyntaxColoring, line 593: To answer your review question: this totally works as-is, but I would probably move this function to its own `code_in_error_tree.py` file. Really a nitpick, though.

api/src/opentrons/protocols/parameters/types.py
```diff
@@ -1,8 +1,37 @@
...
+
+from .exceptions import FileParameterRequired
+
+
+class CSVParameter:
+    def __init__(self, csv_file: Optional[TextIO]) -> None:
+        self._file = csv_file
+        self._rows = []
+        if self._file is not None:
+            for row in csv.reader(self._file):
```
SyntaxColoring, line 12: A few questions about this:

* Is this the first point in the file's journey where its contents are actually parsed as CSV?
* What CSV [dialects](https://docs.python.org/3/library/csv.html#dialects-and-formatting-parameters) are we supporting? Do we want there to be one Opentrons CSV dialect that everybody needs to convert their files to, or do we want to leave it up to protocol authors?
* Is there any validation in the frontend or robot server that this CSV reader needs to agree with, with respect to CSV dialects?

api/src/opentrons/cli/analyze.py
```diff
@@ -365,6 +381,24 @@ class PythonConfig(BaseModel):
...
+        OK: No problems were found during protocol analysis.
+        NOT_OK: Problems were found during protocol analysis. Inspect
+            `analysis.errors` for error occurrences.
+        FILE_REQUIRED: A file is required for a file parameter for full
+                        analysis to be possible.
+    """
+
+    OK = "ok"
+    NOT_OK = "not-ok"
+    FILE_REQUIRED = "file-required"
```
SyntaxColoring: A naming thing, supposing we keep this fundamentally as you have it now:

Could we rename`file-required` to something like `parameter-value-required`, to anticipate being able to omit default values from other kinds of parameters, and having them run into the same problem? I don't think we want analysis's handling of CSV files to differ from its handling of other kinds of parameters.

api/src/opentrons/cli/analyze.py
```diff
@@ -381,6 +415,7 @@ class AnalyzeResults(BaseModel):
     metadata: Dict[str, Any]
 
     # Fields that should match robot-server:
+    result: AnalysisResult
```
SyntaxColoring, line 420: :O

robot-server/robot_server/protocols/analysis_models.py
```diff
@@ -34,10 +34,13 @@ class AnalysisResult(str, Enum):
         OK: No problems were found during protocol analysis.
         NOT_OK: Problems were found during protocol analysis. Inspect
             `analysis.errors` for error occurrences.
+        FILE_REQUIRED: A file is required for a file parameter for full
+                        analysis to be possible.
     """
 
     OK = "ok"
     NOT_OK = "not-ok"
+    FILE_REQUIRED = "file-required"
```
SyntaxColoring: This looks like a sneaky downgrade-breaking database change. New software can save `"file-required"`, and then you can downgrade, and then the old software will choke on it when it tries to parse a `CompletedAnalysis`.

The options I see are:

1. Chain a new `5->6` migration atop the existing ones (`0->...->5`).
    * This is the best we can do to preserve existing data, but it will make robot startup slower and add ongoing maintenance cost.
2. Delete `4->5` and replace it with a new `4->6` migration.
    * We would lose any data in `5`. But because `5` has not been released into the field yet, this would just be data on our internal robots created in the past week, so who cares.
    * Ideally, if we anticipated this ahead of time, `5` would have been named like `5-dev1` for clarity, and we would name the new one `5-dev2`. But it doesn't really matter.
3. Just mutate `5` in-place like you are now and allow downgrades to cause errors.
    * It should only be a problem for people in the office who are pushing dev branches and going back in time.

I would do option 2 or 3.

shared-data/python/opentrons_shared_data/errors/codes.py
```diff
@@ -91,6 +91,7 @@ class ErrorCodes(Enum):
     API_MISCONFIGURATION = _code_from_dict_entry("4007")
     INVALID_STORED_DATA = _code_from_dict_entry("4008")
     MISSING_CONFIGURATION_DATA = _code_from_dict_entry("4009")
+    FILE_PARAMETER_REQUIRED = _code_from_dict_entry("4010")
```
SyntaxColoring: This feels a bit too specific to me to have its own error code.

I think it would make sense to generalize this to something like `INVALID_RUNTIME_PARAMETER_VALUE`, right? What do we return currently if you pass something like an out-of-range `int`?

The nice thing about these codes is that we can freely rename them later, though, as long as we keep the same number and basic vibe.

api/src/opentrons/protocols/parameters/types.py
```diff
@@ -1,8 +1,37 @@
...
+        if self._file is None:
+            raise FileParameterRequired(
+                "CSV Parameter needs to be set to a file for full analysis."
+            )
+        return self._file
+
+    def rows(self) -> List[List[str]]:
+        if self._file is None:
+            raise FileParameterRequired(
+                "CSV Parameter needs to be set to a file for full analysis."
```
ecormany: Since these error strings are the same, should we set that somewhere and reference it here? The copy is fine, although I might lowercase `Parameter`.

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -175,7 +177,15 @@ async def update(
...
-            result = AnalysisResult.NOT_OK
+            if any(
+                code_in_error_tree(
+                    root_error=error, code=ErrorCodes.FILE_PARAMETER_REQUIRED
+                )
+                for error in errors
+            ):
+                result = AnalysisResult.FILE_REQUIRED
+            else:
+                result = AnalysisResult.NOT_OK
```
SyntaxColoring: Overall I think this works, and it looks well tested, and I'm happy for it to be merged. I'll try to recap a discussion we had in-person about this.

It doesn't feel right to me that we're reporting this via `result`. By comparison, we don't have distinct `result`s for syntax errors, divide-by-zero errors, or aspirate-without-tip errors. Distinguishing between those is the job of the `errors` field, not the `result` field.

You brought up that if we have to sift through an error tree to detect this condition, it's better to encapsulate that inside `api` instead of making the frontend TypeScript do it. I agree with that.

However, I don't think we have to sift through an error tree to detect this condition. We could instead check it by diffing the set of provided values against the set of declared parameters. That diffing would be inherently more robust against our exception-handling messes, anyway.

I'm glossing over some nuances to both sides, and some ease-of-implementation things in your favor.

SyntaxColoring (changes requested):
This generally looks good to me, thank you!

Changes requested for a `"file-required"` naming thing, and a database compatibility thing.

api/src/opentrons/protocols/parameters/types.py
jbleon95: - Yes
- There's nothing in the PRD about dialects, we can use the `csv.Sniffer` class to try and parse the dialect (though this isn't guaranteed to be 100 percent accurate in all cases). This definitely needs to be thought about, but this class is going to be fleshed out more in a future PR so that will be dealt with then.
- Right now extent of validation is the frontend checking that the file type is `text/csv` and the parameter context checking the file extension is `.csv`

robot-server/robot_server/protocols/analysis_models.py
jbleon95: After discussion in person, I believe that the chances of downgrade-breaking changes during development are low enough that we can go for option 3. The new status requires unreleased and undocumented (for now) code that is not likely to be used outside the team.

shared-data/python/opentrons_shared_data/errors/codes.py
jbleon95: I wouldn't want to raise this for an out-of-range int error, since that's an error with the parameter definition itself being invalid, rather than a value not being provided for a parameter. However, like the changes you suggest to `file-required`, I could make this more generic with something like `RUNTIME_PARAMETER_VALUE_REQUIRED`

SyntaxColoring (approved):
That github-actions snapshot PR should be merged into this before this gets merged into edge, but LGTM other than that. Thank you!

robot-server/robot_server/protocols/analysis_store.py
sanni-t: I do like that the file required pre-req is highlighted in the analysis, ~but I think calling this property 'result' gives a misleading impression about its meaning.~ (realized late that this is being added to the existing `result` field). As an alternative, can a 'file not provided' error be conveyed via something else? For example, a 'dataFileRequirementsFulfilled' boolean property. 

I can also see @SyntaxColoring 's point of checking for the missing file by diffing the parameters' values. It would actually scale better if/when we allow multiple file uploads. But I'll wait for the recap of your in-person discussion.

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -175,7 +177,15 @@ async def update(
...
-            result = AnalysisResult.NOT_OK
+            if any(
+                code_in_error_tree(
+                    root_error=error, code=ErrorCodes.RUNTIME_PARAMETER_VALUE_REQUIRED
+                )
+                for error in errors
+            ):
+                result = AnalysisResult.PARAMETER_VALUE_REQUIRED
+            else:
+                result = AnalysisResult.NOT_OK
```
sanni-t, line 188: I think we should document the behavior in comments/docs that if a required file is not specified during analysis and if the protocol never uses that file (either because it never uses the csv parameter or because it's bypassed by a try-except catch), then the result will not show the file required error.

sanni-t (approved):
LGTM! Good testing!
Will check your in-person recap notes when they get added.

PR #15436, opened 2024-06-17
fix(shared-data): Implement updated Hardware Testing pipette configuration values

CaseyBatten:
# Overview
Closes PLAT-270 and PLAT-269

Incorporate new hardware testing values for current, speed, distance and tip overlaps for the 96ch and 8ch pipettes on the Flex.

# Test Plan

# Changelog

# Review requests

# Risk assessment


SyntaxColoring (commented):
Nitpick: I think the PR title should be `fix` or `feat` instead of `perf`. `perf` is for things like compute time performance optimizations.

CaseyBatten:
> Nitpick: I think the PR title should be `fix` or `feat` instead of `perf`. `perf` is for things like compute time performance optimizations.

Makes sense. Technically these were values for mechanical/hardware performance improvements, hence the 'perf'. Fix is totally acceptable.

sfoster1 (approved):
Okay, looks good to me! Thanks for getting through this!

PR #15437, opened 2024-06-17
feat(components): added a component called Tabs

syao1226:
# Overview

RSQ-86: Create a new Tabs component to unify the TabbedButton/RoundTab components in storybook.

https://www.figma.com/design/4gRzu3rkNDZnDPkP36YxfF/Chip%2C-Tag-and-Tab?node-id=123-57775&t=zucfumEcRony14Xm-0

# Test Plan

# Changelog

# Review requests

# Risk assessment


components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,89 @@
...
+        >
+            <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing4}>
+                {buttons.map((button, index) => (
+                    <button
+                        key={index}
+                        onClick={() => {
+                            handleClick(index)
+                            button.onClick()  
+                        }}
+                        className={index === activeTab ? 'active' : ''}
```
shlokamin: what is this `className` doing? since we use styled-components instead of raw css we don't need to apply css classes do we?

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,89 @@
...
+            gridGap={SPACING.spacing16}
+            padding={SPACING.spacing16}
+        >
+            <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing4}>
+                {buttons.map((button, index) => (
+                    <button
+                        key={index}
+                        onClick={() => {
+                            handleClick(index)
+                            button.onClick()
```
shlokamin: i think having the button both update its internal state (`activeTab `) as well as call the provided `onClick` handler is a little confusing because you have two sources of truth for whether or not a button is active (this component's internal react state, as well as the provided `isActive` prop). i prefer having the parent component tell us whether or not a button is active, so i think we should get rid of the internal `activeTab` react state variable

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+export function Tabs(props: TabsProps): JSX.Element {
+
+    const { buttons } = props
+
+    const [activeTab, setActiveTab] = React.useState<number | null>(null);
+
+    const handleClick = (index: number) : void => {
+        setActiveTab(index)
+    }
```
shlokamin: see comment below, i dont think we need this!

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    <button
+                        key={index}
+                        onClick={() => {
+                            handleClick(index)
+                            button.onClick()  
+                        }}
+                        className={index === activeTab ? 'active' : ''}
+                        disabled={button.disabled}
+                        
+                        css={index === activeTab || button.isActive === true ? currentTabStyle : button.disabled === true ? disabledTabStyle : defaultTabStyle}
```
shlokamin: instead of conditionally choosing a CSS block for `disabled`, you can tell an HTML element that it is disabled by providing the `disabled` attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

this will simplify the ternary 😄

components/src/molecules/Tabs.stories.tsx
```diff
@@ -0,0 +1,75 @@
+import * as React from 'react'
+import { Tabs as TabComponent } from './Tabs'
+import type { Meta, StoryObj } from '@storybook/react'
+import { useArgs } from '@storybook/preview-api';
```
jerader: nit: we are very particular about import order, the pattern goes:
- package imports
- top level directory imports (like `opentrons/shared-data` or `opentrons/components`)
- imports from the farthest path to the shortest so `../../` should be above `./` for example
- types (type imports should also follow the above pattern)

```suggestion
import * as React from 'react'
import { useArgs } from '@storybook/preview-api';
import { Tabs as TabComponent } from './Tabs'
import type { Meta, StoryObj } from '@storybook/react'
```

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react'
+import { css } from 'styled-components'
+import { TYPOGRAPHY, SPACING } from '../ui-style-constants'
+import { COLORS, BORDERS } from '../helix-design-system'
+import { POSITION_RELATIVE, DIRECTION_COLUMN, DIRECTION_ROW } from '../styles'
+import { Flex } from '../primitives'
```
jerader: import order is good here! 👍

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+    background-color: ${COLORS.purple55};
+  }
+`
+
+export interface TabsProps {
+    buttons: Array<{
+        text: string
+        isActive?: boolean
+        disabled?: boolean
+        onClick: () => void
```
jerader: nit: optional props should go last

```suggestion
        text: string
        onClick: () => void
        isActive?: boolean
        disabled?: boolean
```

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+            padding={SPACING.spacing16}
+        >
+            <Flex flexDirection={DIRECTION_ROW} gridGap={SPACING.spacing4}>
+                {buttons.map((button, index) => (
+                    <button 
+                        key={index}
+                        onClick={() => {
+                            button.onClick()  
+                        }}
+                        css={button.isActive === true ? currentTabStyle : defaultTabStyle}
```
jerader: nit: i think this works as a shorthand
```suggestion
                        css={button.isActive ? currentTabStyle : defaultTabStyle}
```

syao1226: I've tried deleting the "=== true" statement and it gives me this problem: Unexpected nullable boolean value in conditional. Please handle the nullish case explicitly.eslint[@typescript-eslint/strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions)

jerader: i see, i guess its because isActive can be undefined. you can keep it as is!

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+        text: string
+        isActive?: boolean
+        disabled?: boolean
+        onClick: () => void
+    }>
+}
+
+export function Tabs(props: TabsProps): JSX.Element {
+
+    const { buttons } = props
```
koji: why not using interface you defined?

```ts
const = {text, onClick, isActive = false, disabled = false } =  props
```

syao1226: I have the text, onClick, isActive, and disabled wrapped inside an array in the interface because we need to show multiple buttons, so i am not able to call those props directly i think

koji: I recommend you use `tabs` instead of `buttons` since you `button` tag.

components/src/molecules/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react'
+import { css } from 'styled-components'
+import { TYPOGRAPHY, SPACING } from '../ui-style-constants'
+import { COLORS, BORDERS } from '../helix-design-system'
+import { POSITION_RELATIVE, DIRECTION_COLUMN, DIRECTION_ROW } from '../styles'
+import { Flex } from '../primitives'
+
+const defaultTabStyle = css`
```
koji: basically we use `SCREAMING_SNAKE_CASE` for a css var.

```suggestion
const DEFAULT_TAB_STYLE = css`
```

syao1226: updated

components/src/molecules/Tabs.tsx
koji: I checked a couple of ui components.
they use the following structure and I think `RoundTab` is also the same.
I guess passing a custom array to ui component is unusual.

```ts
<Tabs>
  <Tab value="1">tab1<Tab>
  <Tab value="2">tab2<Tab>
</Tabs>
```

Is there any specific reason that you would want to pass an array to this component?

syao1226: that's a really good question! based on what Shlok have told me we want the TabsProps take in a list of tabs, where each tab has text, isActive, isDisabled, and onClick. that way we keep the interface terse and generic

components/src/molecules/Tabs.tsx
koji, line 1: > unify the TabbedButton/RoundTab

This means that this new component needs to support ODD style like what Jethary does for `Tag`.

https://github.com/Opentrons/opentrons/pull/15441/files#diff-69b769c5ad20c1d688db697b1ea496fecad8e6dc0f020bba62e048b1c4d541d0

We use media-query to switch styles. The following for ODD
`@media ${RESPONSIVENESS.touchscreenMediaQuerySpecs}`

Also, could you write test for the component? 

https://opentrons.atlassian.net/wiki/spaces/PAR/pages/4086104119/UI+component+Unification+Implementation+Guide

components/src/molecules/Tabs.stories.tsx
shlokamin: thanks jethary, i forgot to explain this to shiyao

components/src/molecules/Tabs.tsx
shlokamin: lol i had no idea it was called `SCREAMING_SNAKE_CASE`

components/src/molecules/Tabs.tsx
shlokamin: @koji when you have some time would you mind pairing with Shiyao to show her how to write a test using react testing library?

components/src/molecules/Tabs.tsx
koji: ↑ any thoughts on the above?
@shlokamin @jerader @ncdiehl11 


https://github.com/reactjs/react-tabs
https://mui.com/material-ui/react-tabs/#basics
https://v2.chakra-ui.com/docs/components/tabs/usage#fitted-tabs

shlokamin (commented):
looks good so far shiyao! it looks like what's remaining is: 

1. add a test for the component (@koji or @ncdiehl11 can show you how)
2. add ODD specific style overrides for the style differences in ODD
3. replace all instances of the old chip and tag with your new component (we can do this after Mel gives us the thumbs up that the component is designed to spec)

koji:
@syao1226 
1. Create `Tab` component and bring  `TabbedButton`(for ODD)  and `RoundTab`(for desktop)  into `Tab`
TabbedButton's style should be under `@media ${RESPONSIVENESS.touchscreenMediaQuerySpecs}`
I recommend you use `Btn` component instead of `button` tag since we use that component.

we use the following structure for a component.
For Tab, the following is the preferred structure.

```shell
molecules
└── Tab
    ├── Tab.stories.tsx
    ├── __tests__
    │   └── Tab.test.tsx
    └── index.tsx
```



[samples]
https://github.com/Opentrons/opentrons/pull/15441/files#diff-69b769c5ad20c1d688db697b1ea496fecad8e6dc0f020bba62e048b1c4d541d0

https://github.com/Opentrons/opentrons/blob/edge/components/src/atoms/Chip/index.tsx

2. Create `Tabs` component that has the layout for each app(Desktop app and ODD app)
Tabs.stories.tsx should have `Tab`s

```shell
molecules
└── Tabs
    ├── Tabs.stories.tsx
    ├── __tests__
    │   └── Tabs.test.tsx
    └── index.tsx
```


3. Check `Tab` and `Tabs` in Storybook
They should work on Desktop app and ODD app.
These components shows different style by switching the screen size like below.

https://github.com/Opentrons/opentrons/assets/474225/b4ced156-c66a-47da-8110-0e33eb2d5035



4. Write a test for each component
Nick or I can show you how to write a test for a component.


5. Remove RoundTab stories and TabbledButton stories


If you have any question on this, feel free to reach out to the team/me.

components/src/molecules/Tabs.tsx
jerader: i think something we can do to make it cleaner is create a new type that includes 

```
        text: string
        isActive?: boolean
        disabled?: boolean
        onClick: () => void
```

So for example:

```
interface TabProps {
        text: string
        isActive?: boolean
        disabled?: boolean
        onClick: () => void
}
```

And then
```
export interface TabsProps {
      tabs: TabProps[]
}
```

Because this is the Tabs component that has Tab[] inside of it, i think instead of calling it buttons, it could be tabs?

components/src/molecules/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+export default meta
+
+type Story = StoryObj<typeof TabComponent>
+
+export const Tabs: Story = {
+  args: {
+    tabs: [
+      {
+        text: 'Setup',
+        isActive: false,
```
koji: when the app displays tabs, setup is selected so settings isActive: true would be designer-friendly. 

```suggestion
        isActive: true,
```

components/src/molecules/Tabs/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+import {
+  POSITION_RELATIVE,
+  DIRECTION_COLUMN,
+  DIRECTION_ROW,
+} from '../../styles'
+import { Btn, Flex } from '../../primitives'
+
+const DEFAULT_TAB_STYLE = css`
+  ${TYPOGRAPHY.pSemiBold}
+  color: ${COLORS.black90};
```
koji: the default text color is `black90` so you don't need this.

L28 in `app/src/atoms/GlobalStyle/index.ts`   


```suggestion
  
```

components/src/molecules/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react'
+import { useArgs } from '@storybook/preview-api'
+import { Tabs as TabComponent } from '.'
```
koji:
```suggestion
import { Tabs as TabsComponent } from '.'
```

components/src/molecules/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from 'react'
+import { vi, describe, it, expect, beforeEach } from 'vitest'
+import { fireEvent, screen } from '@testing-library/react'
+import { BORDERS, COLORS } from '../../../helix-design-system'
+import { SPACING } from '../../../ui-style-constants'
+import { POSITION_RELATIVE } from '../../../styles'
+import { renderWithProviders } from '../../../testing/utils'
+import { Tabs } from '..'
```
koji:
```suggestion
import { Tabs } from '../index'
```

components/src/molecules/Tabs/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export function Tabs(props: TabsProps): JSX.Element {
+  const { tabs } = props
+
+  return (
+    <Flex
+      flexDirection={DIRECTION_COLUMN}
+      gridGap={SPACING.spacing16}
+      padding={SPACING.spacing16}
```
koji: `flexDirection` and `gridGap` aren't needed since there is one element.

Could you add the Figma link you used to implement this to the pr's overview?

components/src/molecules/Tabs/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  DIRECTION_COLUMN,
+  DIRECTION_ROW,
+} from '../../styles'
+import { Btn, Flex } from '../../primitives'
+
+const DEFAULT_TAB_STYLE = css`
+  ${TYPOGRAPHY.pSemiBold}
+  color: ${COLORS.black90};
+  background-color: ${COLORS.purple30};
+  border: 0px ${BORDERS.styleSolid} ${COLORS.purple30};
```
koji: need this line?
border uses 0px which isn't visible.

components/src/molecules/Tabs/index.tsx
syao1226: oh I was using the css layout in the RoundTab component

https://www.figma.com/design/4gRzu3rkNDZnDPkP36YxfF/Chip%2C-Tag-and-Tab?node-id=123-57775&t=BqW7cEUlr7vSU9f3-0

koji: I think the style you took is for storybook's layout.
https://s3-us-west-2.amazonaws.com/opentrons-components/edge/index.html?path=/story/library-molecules-roundtab--round-tab

koji:
@shlokamin 

https://github.com/Opentrons/opentrons/pull/15437#discussion_r1644970855
Could you explain why we need to change the existing implementation to Shiyao's implementation?

shlokamin:
> [#15437 (comment)](https://github.com/Opentrons/opentrons/pull/15437#discussion_r1644970855) Could you explain why we need to change the existing implementation to Shiyao's implementation?

@koji im not sure what you mean by "existing implementation". the reason i suggested that this component take in a list of tabs is because of the description of the component in Figma:

 "Tabbed interfaces are a way of navigating between multiple panels". 

keyword "multiple"!

we will never expose a single `Tab` interface, it is meant to always be a sequence of tabs (hence the component name)

shlokamin:
ultimately i don't really think it matters too much though. having a single tab interface is fine too, but i personally don't see anything wrong with the interface this component currently accepts

koji:
> Reference in ne



> ultimately i don't really think it matters too much though. having a single tab interface is fine too, but i personally don't see anything wrong with the interface this component currently accepts

My concern is that if there is a request to change the background color of a particular tab or to display an icon, multiples would require additional changes.
I guess that would be one reason that major UI components use a single tab component and tags component to manage tab components.

shlokamin:
> My concern is that if there is a request to change the background color of a particular tab or to display an icon, multiples would require additional changes.

oh if there is ever a request to do that we would say we can't. that would be a big signal that this abstraction is not correct and that this component should not exist. cc @mmencarelli

components/src/molecules/Tabs/Tabs.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Tabs as TabsComponent } from '.'
+import type { Meta, StoryObj } from '@storybook/react'
+
+const meta: Meta<typeof TabsComponent> = {
+  title: 'Library/Molecules/Tabs',
+  component: TabsComponent,
+  parameters: VIEWPORT.touchScreenViewport,
+  argTypes: {
+    tabs: {
+      control: {
```
koji, line 13: @syao1226 Have you researched a way to control an element in array?

This isn't must-have but nice-to-have since designers would like to use a select button for disabled instead of changing the boolean directly.

syao1226: I haven't and I'll look into it now

components/src/molecules/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const tab3 = screen.getByText('module controls')
+
+    expect(tab1).toHaveStyle(`background-color: ${COLORS.purple50}`)
+    expect(tab1).toHaveStyle(`color: ${COLORS.white}`)
+
+    expect(tab2).not.toHaveStyle(`background-color: ${COLORS.purple50}`)
+    expect(tab2).not.toHaveStyle(`color: ${COLORS.white}`)
+
+    expect(tab3).not.toHaveStyle(`background-color: ${COLORS.purple50}`)
+    expect(tab3).not.toHaveStyle(`color: ${COLORS.white}`)
```
shlokamin: no need to check these since you're already asserting these same properties on lines 57-58

shlokamin (approved):
code lgtm

PR #15438, opened 2024-06-17
refactor(app,shared-data): add useRunningStepCounts hook

mjhuff:
Works towards [EXEC-550](https://opentrons.atlassian.net/browse/EXEC-550) and closes [EXEC-510](https://opentrons.atlassian.net/browse/EXEC-510)

# Overview

Since we'll be showing step counts in a few more places, during ER and also on the RunningProtocol page, let's go ahead and create a hook for getting the step count and whether or not a run is deterministic (since this helps clarify certain rendering behavior). 

This hook is added to the one place it's used, `RunProgressMeter`.

# Test Plan
- Pretty well covered by added tests here, but I also went through the office and all the protocol step counts look correct.

# Risk assessment
low

[EXEC-550]: https://opentrons.atlassian.net/browse/EXEC-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EXEC-510]: https://opentrons.atlassian.net/browse/EXEC-510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/resources/protocols/hooks/useLastRunCommandNoFixit.ts
+++ b/app/src/resources/protocols/hooks/useLastRunCommandNoFixit.ts
@@ -1,0 +1,30 @@
+import last from 'lodash/last'
+
+import { useCommandQuery } from '@opentrons/react-api-client'
+
+import type { CommandsData, RunCommandSummary } from '@opentrons/api-client'
+
+// Return the last run command is not a "fixit" command. If it is a "fixit" command,
+// return the command that failed (ie, the last run command without a fixit intent).
+export function useLastRunCommandNoFixit(
+  runId: string,
+  commandsData: CommandsData | null
+): RunCommandSummary | null {
+  const lastRunCommand = last(commandsData?.data) ?? null
+
+  const isFixitIntent =
+    lastRunCommand != null && lastRunCommand.intent === 'fixit'
+
+  // Get the failed command from the fixit command.
+  const lastRunCommandActual = useCommandQuery(
+    runId,
+    lastRunCommand?.failedCommandId ?? null,
+    {
+      enabled: isFixitIntent,
+    }
+  )
+
+  return isFixitIntent
+    ? lastRunCommandActual.data?.data ?? null
+    : lastRunCommand ?? null
+}
--- a/app/src/resources/protocols/hooks/useRunningStepCounts.ts
+++ b/app/src/resources/protocols/hooks/useRunningStepCounts.ts
@@ -1,13 +1,13 @@
 import { useMostRecentCompletedAnalysis } from '../../../organisms/LabwarePositionCheck/useMostRecentCompletedAnalysis'
 
-import { useLastRunCommandNoFixit } from './useLastRunCommandNoFixit'
+import { useLastRunProtocolCommand } from './useLastRunProtocolCommand'
 
 import type { CommandsData } from '@opentrons/api-client'
 
 export interface StepCounts {
-  /* Excludes "fixit" commands. Returns -1 if the step is not found. */
+  /* Excludes "fixit" commands. Returns null if the step is not found. */
   currentStepNumber: number | null
-  /* Returns null if the run is non-deterministic or the total command count is not found. */
+  /* Returns null if the run has diverged or the total command count is not found. */
   totalStepCount: number | null
   /* Returns whether the run has diverged from analysis. */
   hasRunDiverged: boolean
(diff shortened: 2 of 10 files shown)


app/src/resources/protocols/hooks/useRunningStepCounts.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import type { CommandsData } from '@opentrons/api-client'
+
+export interface StepCounts {
+  /* Excludes "fixit" commands. Returns -1 if the step is not found. */
+  currentStepNumber: number | null
+  /* Returns null if the run is non-deterministic or the total command count is not found. */
+  totalStepCount: number | null
+  /* Returns whether the run is deterministic. */
+  isDeterministicRun: boolean
```
SyntaxColoring: This is a really pedantic naming nitpick, but I think it would be more precise to say `!runHasDiverged` instead of `isDeterministicRun`. We can't ever answer the forward-looking question of whether a run *is* deterministic and *will* always behave the same way. We can only answer the backwards-looking question of whether the run *has* behaved differently from what analysis predicted.

app/src/resources/protocols/hooks/useRunningStepCounts.ts
```diff
@@ -0,0 +1,60 @@
+import { useMostRecentCompletedAnalysis } from '../../../organisms/LabwarePositionCheck/useMostRecentCompletedAnalysis'
+
+import { useLastRunCommandNoFixit } from './useLastRunCommandNoFixit'
+
+import type { CommandsData } from '@opentrons/api-client'
+
+export interface StepCounts {
+  /* Excludes "fixit" commands. Returns -1 if the step is not found. */
+  currentStepNumber: number | null
```
SyntaxColoring: If I'm following the logic right, I don't think this can actually be -1, can it? Should this comment say `null` instead?

SyntaxColoring (approved):
Sweet. Does this client-side solution mean EXEC-510 is no longer needed or helpful?

app/src/resources/protocols/hooks/useLastRunCommandNoFixit.ts
```diff
@@ -0,0 +1,30 @@
+import last from 'lodash/last'
+
+import { useCommandQuery } from '@opentrons/react-api-client'
+
+import type { CommandsData, RunCommandSummary } from '@opentrons/api-client'
+
+// Return the last run command is not a "fixit" command. If it is a "fixit" command,
+// return the command that failed (ie, the last run command without a fixit intent).
+export function useLastRunCommandNoFixit(
```
sfoster1: You actually would have the exact same problem if the query returned a setup command - it's just that you wouldn't care, because the default is to return 0, and setup commands only happen before protocols start so that's the right answer anyway.

That said though, it does give us a bit better of a solution for the hardest problem in computer science: let's name this `useLastRunProtocolCommand` or something so it has positive logic

app/src/resources/protocols/hooks/useLastRunCommandNoFixit.ts
```diff
@@ -0,0 +1,30 @@
...
+  runId: string,
+  commandsData: CommandsData | null
+): RunCommandSummary | null {
+  const lastRunCommand = last(commandsData?.data) ?? null
+
+  const isFixitIntent =
```
sfoster1: and this should probably check `isProtocolIntent` and later `enabled: !isProtocolintent`

sfoster1 (approved):
Structure looks good, would love to see some naming tweaks though - both the ones I suggested and the ones @SyntaxColoring suggested

mjhuff:
> Sweet. Does this client-side solution mean [EXEC-510](https://opentrons.atlassian.net/browse/EXEC-510) is no longer needed or helpful?

I do believe it does! I'll close it out with this PR 👍 

[EXEC-510]: https://opentrons.atlassian.net/browse/EXEC-510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

app/src/resources/protocols/hooks/useLastRunCommandNoFixit.ts
mjhuff: Ohhhh, yeah, forgot about that. Sounds good 👍

app/src/resources/protocols/hooks/useRunningStepCounts.ts
mjhuff: That makes sense, thanks!

app/src/resources/protocols/hooks/useRunningStepCounts.ts
mjhuff: Good catch!

PR #15439, opened 2024-06-17
refactor(app): add a one-column component

sfoster1:
provided basically for completeness. it's a box with width=100%.

Closes EXEC-527


mjhuff (approved):
Storybook looks like it's crashing (CI will probably catch this), but once that's fixed it's good to go! Thank you!

TamarZanzouri (approved):
Nice!

PR #15442, opened 2024-06-17
feat(app): add error recovery error handling

mjhuff:
Closes [EXEC-548](https://opentrons.atlassian.net/browse/EXEC-548) and [EXEC-512](https://opentrons.atlassian.net/browse/EXEC-512)

# Overview

Add error handling to errors that occur within Error Recovery, including Drop Tip Wizard. We do this by using a new route, `ERROR_WHILE_RECOVERING`. Each step represents a unique error view.

For errors that occur within Drop Tip Flows, we effectively redirect outside of the DT Flows, render the error, and then either route the user back to a specific route in DT Flows or elsewhere. Although this in-and-out of DT Flow behavior might seem weird, there are a few benefits:

- We don't special case and maintain error modals in Drop Tip Wizard that really belong to Error Recovery.
- For other overpressure routes, we will shortly need this "go to specific DT Flow" step override, because we might only want to drop tips without blowing out, for example.
- Design has indicated we might not use DT Flows in the future. Might as well not create more work for ourselves than necessary.
- When thinking about this more, it's the least hacky option.

Although design didn't slate DT Flow errors that occur besides those for blowout/drop tip, we now have "<recovery action> failed". After talking with Sue, design was ok with this.

Using specific steps for error messaging instead of full blown routes makes sense, since we one only ever have one view per error, and this helps keeps things consolidated under one route. Originally, I did not do this with "robot in motion" views, because it seemed like we might possibly show more steps per motion route. That hasn't been the case, and I added a TODO to consolidate these down to one route, many steps later. 

### PickUpTip Ignored

Until the hardware control pipette work is merged in, all pick up tip commands fail. Instead of hard blocking pretty much all of Error Recovery, I've gone ahead and turned off pickUpTip errors. I've left a TODO. We absolutely need to re-enable this before release. 

### Examples 
https://github.com/Opentrons/opentrons/assets/64858653/0a78f070-7de6-43b7-98bf-31ca3f998c14

https://github.com/Opentrons/opentrons/assets/64858653/4e9c4c8f-e96b-4dda-8afd-e3795b1d9abc

# Test Plan
- Throw random errors in ER, watch them work.
- Throw random errors in ER during drop tip flows, watch them work.

# Changelog
- Added error handling to Error Recovery.

# Risk assessment
low

[EXEC-548]: https://opentrons.atlassian.net/browse/EXEC-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-512]: https://opentrons.atlassian.net/browse/EXEC-512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (approved):
Looks great me!

PR #15444, opened 2024-06-17
feat(shared-data): add schema definitions for liquidProbeParameters

aaron-kulkarni:
The definitions of minimumHeight and minimumWellVolume are for all non tip rack labware. They will be used in LLD.

fix EXEC-518

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


shared-data/labware/schemas/2.json
```diff
@@ -173,6 +173,14 @@
         "magneticModuleEngageHeight": {
           "description": "Distance to move magnetic module magnets to engage",
           "$ref": "#/definitions/positiveNumber"
+        },
+        "minimumHeight:": {
```
sfoster1: this unfortunately needs to be more than just one number to handle labwares that have multiple different kinds of wells like the 15/50 combo tube rack. there's two options here:
1. It's part of the well datastructure, for each well
2. instead of containing a number, this is an array of objects that have a `value` member that is a number and an `appliesToWells` member that is an array of well names or something

Goes for both `minimumHeight` and `minimumWellVolume`.

sfoster1 (dismissed):
I think we need to change the structure a little to handle heterogenous-well labware like the 15/50 tuberack.

ryanthecoder:
I'd also like to see these be a joined object like
` "LiquidProbeSettings: {
            "minimumHeight": 0.5,
            "minimumWellVolume": 0,
  }`

ryanthecoder (approved):
Looks good to me

PR #15373, opened 2024-06-10
refactor(app): add modal mixed media content

sfoster1:
This is a component for InterventionModal (and eventually the rest of the kinds of modal) that presents some kind of selectable visual element
- spinner, icon, image 
- and some text.

Implements https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=1935-44819&t=pKbR40CZeBMrPLxM-4

Storybook: https://s3-us-west-2.amazonaws.com/opentrons-components/exec-489-add-mixed-content/index.html?path=/docs/app-molecules-interventionmodal-modalcontentmixed--docs

Closes EXEC-489

PR #15446, opened 2024-06-17
refactor(app): InterventionModal Description component

sfoster1:
Implements the Description modal content component for ODD and Desktop.

Also updates `InlineNotification` to have corresponding desktop styles using the new `StyledText`, which means it doesn't have to be ODD only anymore so pop its stories over to `App`.

Note - can't be merged until #15373 is merged because it needs the StyledText fixes from there.

### DIFF
--- a/app/src/atoms/InlineNotification/index.tsx
+++ b/app/src/atoms/InlineNotification/index.tsx
@@ -1,7 +1,9 @@
 import * as React from 'react'
+import { css } from 'styled-components'
 import {
   ALIGN_CENTER,
   BORDERS,
+  Box,
   Btn,
   COLORS,
   DIRECTION_ROW,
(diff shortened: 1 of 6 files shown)


app/src/atoms/InlineNotification/index.tsx
```diff
@@ -58,52 +66,136 @@ const INLINE_NOTIFICATION_PROPS_BY_TYPE: Record<
...
+              desktopStyle="bodyDefaultRegular"
+              textDecoration="underline"
+            >
+              {linkText}{' '}
+            </StyledText>
+          </Link>
+        )}
+        {onCloseClick && (
+          <Btn
+            data-testid="InlineNotification_close-button"
```
TamarZanzouri, line 171: for what are we using this? just wondering why we have this arg

TamarZanzouri (approved):
LGTM! Thank you for the in-person explanation

PR #15449, opened 2024-06-18
refactor(app): rename error recovery "hooks" dir to "utils"

mjhuff:
# Overview

A lot of the Error Recovery hooks used as utils are dependent on non-ER-hooks used as utils, and instead of making some weird cross-dependency between hooks that act like utils in a hooks directory and utils that are only used by hooks in a utils directory, I think it makes sense to keep everything in a "utils" directory and call it a day.

I'm open to thoughts/comments/concerns. There are a good number of utils in "utils", but if it's any consolation, we are most likely done adding any utils for MVP at this point (well, maybe there will be a couple more for good measure, I don't know yet).

# Test Plan
- CI passes.

# Risk assessment
none


sfoster1 (changes requested):
I dunno, I do think it's nice to have hooks in `hooks/` and stuff that's not hooks in `utils/`. I appreciate that they might be used in different ways, and that the hooks do have standard names, but I think it's just nice and neat to keep them in separate directories also.

mjhuff:
Per convo, we'll go with separate directories for now with the intention of breaking the hooks directory into data vs. component-utilizing-hooks later (all the ER hooks are data hooks, currently).

sfoster1 (approved):
Nice, thanks!

PR #15450, opened 2024-06-18
docs(api): Corrected slicing and well names in Multiple Air Gaps example

Kuhjuice:
# Overview

Fixed a mistake in the documentation.

In the Example for [Multiple Air Gaps](https://docs.opentrons.com/v2/new_examples.html#multiple-air-gaps) there is the code:
``` python
# aspirate from the first 5 wells
    for well in reservoir.wells()[:5]
```

In the text below the code it says:

> [...] (in the code sample as **[:4]**) lets us select the first five wells of the well plate only. [...]

For the first five wells the slicing of the code **[:5]** is correct and should be corrected in the text.

- The slice `[:5]` means "from the start up to but not including index 5".
- So, we get the elements at indices 0, 1, 2, 3, and 4.

# Test Plan
Think about it and accept it.

# Changelog

- Minor error correction in the documentation

# Review requests

# Risk assessment

### DIFF
--- a/api/docs/v2/new_examples.rst
+++ b/api/docs/v2/new_examples.rst
@@ -397,7 +397,7 @@ Opentrons electronic pipettes can do some things that a human cannot do with a p
 
                 p300.return_tip()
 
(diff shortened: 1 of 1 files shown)


ecormany:
Thanks for spotting this! I've taken a closer look at this explanatory text and it has another small error: the wells are A1–A5, not A1–E1 (because the labware is a reservoir, which only has an A row). I'll work on adding another commit to this branch.

ecormany (approved):
Thanks again for directing our attention to this. We love docs contributions, no matter how small!

FYI: This won't appear on the live docs site immediately after it's merged. Our next docs deploy should happen on or around June 26.

PR #15452, opened 2024-06-18
fix(step-generation): add nozzle props to consolidate and distribute

jerader:
closes RESC-283

# Overview

Oof, this is not a great bug. Basically, distribute and consolidate are currently broken in production when using a 96-channel because the not required `nozzles` prop was not extended for those components 🫠 I need to work with the QA team to add testing consolidate and distribute for single, 8 and 96 channels to the test plan. I'm pretty sure we only ever tested single and 8 channel 😬 

Anyway, this PR adds those props in and solves the issue the user was having.

# Test Plan

Upload the attached protocol to PD in production and see that it errors on step 9. Upload it to PD on this branch and see that it does not error

[Flex LDH 96 well kit (2) (1).json](https://github.com/user-attachments/files/15888095/Flex.LDH.96.well.kit.2.1.json)

# Changelog

- add nozzles prop to the replace tip in consolidate and distribute

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/step-generation/src/commandCreators/compound/consolidate.ts
+++ b/step-generation/src/commandCreators/compound/consolidate.ts
@@ -292,6 +292,7 @@ export const consolidate: CommandCreator<ConsolidateArgs> = (
             pipette: args.pipette,
             dropTipLocation,
             tipRack: args.tipRack,
+            nozzles: args.nozzles ?? undefined,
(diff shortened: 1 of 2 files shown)


koji (approved):
lgtm

PR #15453, opened 2024-06-18
refactor(app): update CommandText

mjhuff:
# Overview

`CommandText` should take a typography override. Let's use `as` with a default of `p`, since it's a consistent method for utilizing design-approved typography without generating our own one-off CSS styles. This adds the style override to the only place it's used, `StepInfo`.

Screenshot 2024-06-18 at 10 58 58 AM

# Risk assessment
low


sfoster1 (approved):
Okay, seems fine!

PR #15459, opened 2024-06-18
feat(components): add helix product typography

sfoster1:
This PR implements the new helix product typography system. You can find it here: https://www.figma.com/design/1ot18My22DALIcjdLl5LJh/Helix-Design-System?node-id=3571-46578&t=B4pPbtL9mapki2Jg-4

The Product type system is distinct from at least the Web type system, which will have different styles for readability in a different context. It is currently only defined for places other than the ODD.

This ended up being a big PR because our usage of the `StyledText` atom had to be changed somewhat.

The change was necessary because the helix product styles do not and are not intended to align with the semantic styling presented by the `StyledText` component - there is no direct alignment that "when you're making a `<p>`, you use this style". Instead, the intent is that higher level components will use the style by the name it has in helix, without an explicit semantic link outside of the designs. That makes the way the old styled text works kind of nonsense.

In addition, there are no helix product styles for the ODD, and there is no current map and possibly no intention to have a future map between the semantic styles of the ODD and Desktop. That means that having a `StyledText` component that maps from a single semantic element to a set of appropriate styles just doesn't really make sense.

So, `StyledText` is renamed to `LegacyStyledText` including at all callsites. The new `StyledText` requires you to specify styles separately for ODD and Desktop, though it gives you a handy type system integrated hint towards which styles should be allowed. Well, not quite required, more like you have no way to specify both in one prop.

 You can view it commit by commit:

- https://github.com/Opentrons/opentrons/pull/15459/commits/07dae4b338d5b0966fc186fa27a944005a2e3fda adds the actual data for the helix product typography styling, and stories for viewing them as design components (and for viewing the old typography styles). This is more or less the only commit there was during the initial review, though it didn't then have the stories for the old typography
- https://github.com/Opentrons/opentrons/pull/15459/commits/a36f50178c93d95f9cf86610fe8313ae643afdaa is a repo-wide find and replace to change `StyledText` to `LegacyStyledText`
- https://github.com/Opentrons/opentrons/pull/15459/commits/acec9340bcc955921c5007eb388ffec06ce6ac3c implements the new `StyledText`.

Stories here: https://s3-us-west-2.amazonaws.com/opentrons-components/EXEC-566-add-helix-product-typography/index.html?path=/docs/design-tokens-typography--docs

Closes EXEC-566


components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,137 @@
...
+// Display styles
+const fontFamilyDisplay = fontFamily
+const fontSizeDisplay = '2.4375rem' // 39px
+const lineHeightDisplay = '3rem' // 48px
+
+// Display-Bold
+export const fontSizeDisplayBold = fontSizeDisplay
+export const lineHeightDisplayBold = lineHeightDisplay
+export const fontFamilyDisplayBold = fontFamilyDisplay
+export const fontWeightDisplayBold = '700'
```
koji: maybe define font-weight ?

```ts
const regular = 400'
const semiBold = '600'
const bold = '700'
```

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,137 @@
...
+// Caption
+const fontSizeCaption = '0.8125rem' // 13px
+const lineHeightCaption = '1rem' // 16px
+const fontFamilyCaption = 'Public Sans'
+
+// Caption-Medium
+export const fontSizeCaptionMedium = fontSizeCaption
+export const lineHeightCaptionMedium = lineHeightCaption
+export const fontFamilyCaptionMedium = fontFamilyCaption
+export const fontWeightCaptionMedium = '500px'
```
koji: 500?

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,137 @@
...
+export const fontStyleCaptionMedium = `${fontWeightCaptionMedium} ${fontSizeCaptionMedium}/${lineHeightCaptionMedium} ${fontFamilyCaptionMedium}`
+
+// Caption-Regular
+export const fontSizeCaptionRegular = fontSizeCaption
+export const lineHeightCaptionRegular = lineHeightCaption
+export const fontFamilyCaptionRegular = fontFamilyCaption
+export const fontWeightCaptionRegular = '400px'
```
koji: 400?

components/src/helix-design-system/product/typography.ts
sfoster1: yup, direct per @mmencarelli . Caption's faded slightly.

components/src/helix-design-system/product/typography.ts
sfoster1: same as above, direct from design

sfoster1: Whoops

components/src/helix-design-system/product/typography.ts
sfoster1: whoops

koji (approved):
lgtm!
thank you for adding this!

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,141 @@
+/**
+ * Note: these values are only currently correct for desktop and should not be used on ODD.
+ */
+
+// Valid for most typography styles
+export const fontFamily = 'Public Sans'
```
koji, line 6: can we use this in `app/src/atoms/GlobalStyle/index.ts`?

components/src/atoms/StyledText/index.tsx
koji, line 1: use `.ts` itead of `.tsx`?

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,141 @@
...
+export const fontSizeCaptionRegular = fontSizeCaption
+export const lineHeightCaptionRegular = lineHeightCaption
+export const fontFamilyCaptionRegular = fontFamilyCaption
+export const fontWeightCaptionRegular = '400'
+export const fontStyleCaptionRegular = `${fontWeightCaptionRegular} ${fontSizeCaptionRegular}/${lineHeightCaptionRegular} ${fontFamilyCaptionRegular}`
+
+// Code
+const fontSizeCode = '0.75rem' // 12px
+const lineHeightCode = '1.25rem' // 20px
+const fontFamilyCode = 'Reddit Mono'
```
koji, line 135: can we use `Reddit Mono` without installing a package?

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,141 @@
...
+export const fontSizeBodyDefaultRegular = fontSizeBodyDefault
+export const lineHeightBodyDefaultRegular = lineHeightBodyDefault
+export const fontFamilyBodyDefaultRegular = fontFamilyBodyDefault
+export const fontWeightBodyDefaultRegular = fontWeightRegular
+export const fontStyleBodyDefaultRegular = `${fontWeightBodyDefaultRegular} ${fontSizeBodyDefaultRegular}/${lineHeightBodyDefaultRegular} ${fontFamilyBodyDefaultRegular}`
+
+// Caption
+const fontSizeCaption = '0.8125rem' // 13px
+const lineHeightCaption = '1rem' // 16px
+const fontFamilyCaption = 'Public Sans'
```
koji:
```suggestion
const fontFamilyCaption = 'Public Sans'
```
```suggestion
const fontFamilyCaption = fontFamily
```

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,141 @@
...
+export const lineHeightCaptionMedium = lineHeightCaption
+export const fontFamilyCaptionMedium = fontFamilyCaption
+export const fontWeightCaptionMedium = '500'
+export const fontStyleCaptionMedium = `${fontWeightCaptionMedium} ${fontSizeCaptionMedium}/${lineHeightCaptionMedium} ${fontFamilyCaptionMedium}`
+
+// Caption-Regular
+export const fontSizeCaptionRegular = fontSizeCaption
+export const lineHeightCaptionRegular = lineHeightCaption
+export const fontFamilyCaptionRegular = fontFamilyCaption
+export const fontWeightCaptionRegular = '400'
```
koji:
```suggestion
export const fontWeightCaptionRegular = fontWeightRegular
```

components/src/helix-design-system/product/typography.ts
```diff
@@ -0,0 +1,141 @@
...
+// Caption
+const fontSizeCaption = '0.8125rem' // 13px
+const lineHeightCaption = '1rem' // 16px
+const fontFamilyCaption = 'Public Sans'
+
+// Caption-Medium
+export const fontSizeCaptionMedium = fontSizeCaption
+export const lineHeightCaptionMedium = lineHeightCaption
+export const fontFamilyCaptionMedium = fontFamilyCaption
+export const fontWeightCaptionMedium = '500'
```
koji: nit

maybe define as fontWeightMedium and add it to the top fontWeight group?

then

```ts
export const fontWeightCaptionMedium = fontWeightMedium
```

components/src/helix-design-system/product/typography.ts
sfoster1: Definitely not, or at least without installing a font however we install a font - if you look at storybook, this shows up as whatever the browser fallback is. I really haven't investigated how/whether we can isntall this, so I think that if we don't want to keep it as a placeholder I'd rather delete it for now.

components/src/helix-design-system/product/typography.ts
sfoster1: It's not though, `fontWeightMedium` is 600

sfoster1: well, `fontWeightSemibold` anyway

sfoster1: and we got a dqa update to make it 600, so i did this

jerader (approved):
very nice 🤩

components/src/atoms/StyledText/StyledText.tsx
```diff
@@ -0,0 +1,275 @@
...
+      }
+    `,
+  },
+} as const
+
+const ODDStyleMap = {
+  level1Header: {
+    as: 'h1',
+    style: css`
+      @media ${RESPONSIVENESS.touchscreenMediaQuerySpecs} {
```
jerader, line 120: question, do we need these media queries if these are exclusively for the ODD?

components/src/helix-design-system/product/typography.ts
koji: For Public Sans, `app` is using `@fontsource` and fontsource supports `reddit mono`.

https://github.com/fontsource/font-files/blob/025ea945ff916c7bfe4fe5aaef1f3f57a523a1c6/fonts/google/reddit-mono/README.md?plain=1

components/src/atoms/StyledText/StyledText.tsx
sfoster1: Yes because the idea is that you will be passing _both_ a `desktopStyle` and `oddStyle` to the component, so we want them to work together nicely

PR #15468, opened 2024-06-20
fix(app): fix multichannel 384 well selection

brenthagen:
# Overview

selects only the intial well of a multichannel selection instead of all wells, so the correct commands are generated

# Test Plan

verified multichannel quick transfers pass analysis

# Changelog

 - Fixes multichannel 384 well selection

# Review requests

 - check 8- and 96-channel 384 well selection for expected behavior
 - generate the quick transfer and initiate setup to verify analysis passes

# Risk assessment

low


jerader (commented):
i tested 8- and 96-channel pipetting into a 384 well plate and looks good. However, a single channel column selection selects 2 at a time which i'm not sure is expected?

https://github.com/Opentrons/opentrons/assets/66035149/0e61d1d6-5f60-4563-aaee-9e652a47f867

brenthagen:
> i tested 8- and 96-channel pipetting into a 384 well plate and looks good. However, a single channel column selection selects 2 at a time which i'm not sure is expected?
> 
>  Screen.Recording.2024-06-21.at.11.57.35.mov

yep, that's a reversion, added back the single channel column select. thanks for catching

smb2268 (approved):
Tested this out with all 3 pipette configurations and it works as expected! I'll follow up with this ticket https://opentrons.atlassian.net/browse/PLAT-344 to adapt logic now that the length of the `sourceWell` and `destWell` arrays are no longer equal to the number of wells actually selected.

PR #15469, opened 2024-06-20
refactor(api): wire protocol engine pick_up_tip to new hardware control function

caila-marashaj:
## Overview
Hardware control now has a `tip_pickup_moves` function, which is just the most immediate movement required to pick up a tip, and unlike `pick_up_tip`, does not cache a tip in the pipette handler or prepare the pipette for aspiration. 

This code changes the protocol engine's `pick_up_tip` command to use this new function if a flex is executing the command, and does not change the behavior for an ot2.

## Test Plan

- [x] Test tip pickup on an ot2 and make sure behavior hasn't changed
- [x] Verify that a flex can now retry tip pickup multiple times in error recovery, and then proceed with a protocol once tips are present

### DIFF
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -1152,7 +1152,7 @@ class API(
                 mount, back_left_nozzle, front_right_nozzle, starting_nozzle
             )
 
-    async def tip_pickup_moves(
+    async def _tip_pickup_moves(
         self,
         mount: top_types.Mount,
         spec: PickUpTipSpec,
@@ -1200,7 +1200,7 @@ class API(
             home_flagged_axes=False,
         )
 
-        await self.tip_pickup_moves(mount, spec)
+        await self._tip_pickup_moves(mount, spec)
         _add_tip_to_instrs()
 
         if prep_after:
--- a/api/src/opentrons/hardware_control/api.py
+++ b/api/src/opentrons/hardware_control/api.py
@@ -65,7 +65,7 @@ from .robot_calibration import (
     RobotCalibration,
 )
 from .protocols import HardwareControlInterface
-from .instruments.ot2.pipette_handler import PipetteHandlerProvider, PickUpTipSpec
+from .instruments.ot2.pipette_handler import PipetteHandlerProvider
 from .instruments.ot2.instrument_calibration import load_pipette_offset
 from .motion_utilities import (
     target_position_from_absolute,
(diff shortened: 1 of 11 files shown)


sfoster1 (changes requested):
Tests tests tests! But it lookks good to me otherwise.

api/tests/opentrons/protocol_engine/execution/test_tip_handler.py
```diff
@@ -76,6 +78,76 @@ async def test_create_tip_handler(
...
+async def test_flex_pick_up_tip_state(
+    decoy: Decoy,
+    mock_state_view: StateView,
+    mock_labware_data_provider: LabwareDataProvider,
+    tip_rack_definition: LabwareDefinition,
+    tip_state: TipStateType,
+) -> None:
+    """Test the protocol engine's pick_up_tip logic."""
+    try:
+        from opentrons.hardware_control.ot3api import OT3API
```
DerekMaggio: You only need to wrap the import statement in the try. The rest of the test logic can be outside the try-except.
Also, is this how we normally run Flex-only tests? Or should the `ot3_only` mark be used and it can be expected the import will always work?

api/src/opentrons/protocol_engine/execution/tip_handler.py
```diff
@@ -172,13 +172,29 @@ async def pick_up_tip(
...
         )
 
-        await self._hardware_api.pick_up_tip(
-            mount=hw_mount,
-            tip_length=actual_tip_length,
-            presses=None,
-            increment=None,
-        )
-        await self.verify_tip_presence(pipette_id, TipPresenceStatus.PRESENT)
+        if self._state_view.config.robot_type == "OT-3 Standard":
```
SyntaxColoring: Could you explain why we need to do this `if`/`else` on `robot_type`? Does the OT-2 HWAPI not behave the same way as the OT-3 HWAPI?

api/src/opentrons/hardware_control/api.py
```diff
@@ -1152,7 +1152,7 @@ async def update_nozzle_configuration_for_mount(
                 mount, back_left_nozzle, front_right_nozzle, starting_nozzle
             )
 
-    async def tip_pickup_moves(
+    async def _tip_pickup_moves(
```
sfoster1: I think this can be a public function, no reason not to really

sfoster1: does this have a `cache_tip` already also? if not, should add that too

api/tests/opentrons/protocol_engine/execution/test_tip_handler.py
sfoster1: try it with the `ot3_only` mark, yeah

api/src/opentrons/protocol_engine/execution/tip_handler.py
caila-marashaj: no, we don't want to run `verify_tip_presence` for an ot2. We pretty much just want to run the old `pick_up_tip` if we have an ot2, and then unconditionally cache the tip.

sfoster1 (changes requested):
I think we can make the ot-2 stuff more similar to the flex stuff which will reduce complexity

api/src/opentrons/protocol_engine/execution/tip_handler.py
SyntaxColoring: But this code *is* calling `verify_tip_presence()` on an OT-2, isn't it? Down on line 197?

caila-marashaj: yeah you're right I think this actually is compatible with ot2

sfoster1 (approved):
Looks good to me, nice!

PR #15466, opened 2024-06-20
feat(app): Add Error Recovery Routes

mjhuff:
Closes [EXEC-562](https://opentrons.atlassian.net/browse/EXEC-562), [EXEC-563](https://opentrons.atlassian.net/browse/EXEC-563), [EXEC-564](https://opentrons.atlassian.net/browse/EXEC-564), [EXEC-471](https://opentrons.atlassian.net/browse/EXEC-471), [EXEC-549](https://opentrons.atlassian.net/browse/EXEC-549), [EXEC-465](https://opentrons.atlassian.net/browse/EXEC-465), and [EXEC-463](https://opentrons.atlassian.net/browse/EXEC-463)

# Overview

This PR adds in the remaining Error Recovery routes and designs for ODD (no liquid detected, overpressure while prepare to aspirate, and overpressure while dispensing), updating copy and components along the while. I apologize in advance for the size of this one - given time constraints, it would seem impractical to do 8 rounds of PR reviews for one blocked PR after another. 

This PR is best viewed by commit, however, note that I do a lot of clean up in the last commit, so if something looks off, I'd first reference the final git diff:
e291821a6b5e21152fd2c09f838c6cf85f10b82d - The initial wiring up of the aforementioned routes.
9df270911364def221cfeebfd5e57d231d7e0c62 - This is largely a naming refactor with just one new fn, `getRelevantFailedLabwareCmdFrom`. For MVP, there are two instances we care about labware. First, in "no liquid detected" errors, we care about the well reservoir. Second, in all overpressure errors, we care about the pickUpTipLabware (the old behavior). 
7b763a6b948424a15b392e9b158a6feff822a389 - [Figma](https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=3496-65568&t=m04Js8AjZeIlUoar-4).
a0d764cefb8a2c24e8a0d092fa042a5f5e905ac9- I added a new `FailedStepNextStep` component, which is used in pretty much every flow.
b350440fa7a6fc2c2bdb45d70e6fa07de0f8e828 - Just a wrapper component.
506571c2205376a3357404fe798753c1c92b6fae - [Figma](https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery?node-id=3552-28404&t=m04Js8AjZeIlUoar-4)
[...]

PR #15471, opened 2024-06-20
refactor(app): update current step behavior

mjhuff:
[EXEC-572](https://opentrons.atlassian.net/browse/EXEC-572)

# Overview

Harmonizes the return value logic of `currentStepNumber` with `totalStepCount`. 

Prior to 8.0, we’d fallback to deriving the current step count from the active run endpoint if we couldn’t get it from analysis (the run had diverged). Now, because we have fixit commands, we can’t do this, since that fallback step count will include fixit commands. This isn’t a functional change, just an "enforcing consistency change". We were already not using the run commands endpoint for “total steps” as a fallback. Realistically, we can always find the current step number from protocol analysis, since we filter out “not protocol” commands, so if `currentStepNumber` is ever `null`, there's an issue with a fetch, the protocol analysis itself, or most likely, the current command has diverged from analysis.

# Test Plan
- If you want to test, throw `edge` robot-server and cherry-pick this commit on top of #15466 before pushing the ODD. Open the error details modal in the upper right during error recovery, look at the step count, then issue some fixit commands. Re-open the error details modal and confirm the step counts haven't changed.
-  Smoked test app to confirm no regression. Verified adding fixit commands does not cause the step count to change. 

# Changelog

# Review requests

# Risk assessment

[EXEC-572]: https://opentrons.atlassian.net/browse/EXEC-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/src/organisms/RunProgressMeter/__tests__/RunProgressMeter.test.tsx
+++ b/app/src/organisms/RunProgressMeter/__tests__/RunProgressMeter.test.tsx
@@ -100,7 +100,7 @@ describe('RunProgressMeter', () => {
   it('should show only the total count of commands in run and not show the meter when protocol is non-deterministic', () => {
     vi.mocked(useCommandQuery).mockReturnValue({ data: null } as any)
     render(props)
-    expect(screen.getByText('Current Step 42/?')).toBeTruthy()
+    expect(screen.getByText('Current Step ?/?')).toBeTruthy()
     expect(screen.queryByText('MOCK PROGRESS BAR')).toBeFalsy()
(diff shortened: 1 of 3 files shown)


sfoster1 (approved):
Nice

mjhuff:
The lint failure doesn't look particularly related to this PR. Will merge once rest of CI completes.

PR #15481, opened 2024-06-20
feat(robot-server): add a new endpoint to upload csv data files

sanni-t:
Closes AUTH-426

# Overview

Adds a new router for uploading and managing data files. The new endpoint can accept a single data file, either as a file upload or a path to the file on the robot.

The endpoint saves the uploaded file to the persistent directory on the robot, adds the file's info to the data_files database, and returns the file ID and other deets in the response.

Duplicate file uploads use the previously saved file & database entry instead of creating a new one.

# Test Plan

- On a robot, upload a data file using the `POST /dataFiles` endpoint and providing a file in the request. Verify that-
    - [ ] the file is uploaded to the persistent storage at location- `../opentrons-robot-server/data_files/<file_id>/`
    - [ ] the file's info is added to the database
    - [ ] the file's ID is returned in the response
- [ ] On a robot, send a file to `POST /dataFiles` by providing a `file_path` to a data file on the robot (the file should be outside the persistent directory) and verify all the above
- [ ] Send an already existing file to the server and verify that no new file is saved and no new entry is added to database. Verify that the response contains the previously assigned file ID

# Changelog

- added new router, models and data files store for the new endpoint
- added unit & integration tests

# Review requests

- check that the implementation meets client's expectations
- any suggestions on placement of file_reader_writer and file_hasher classes/functions? They have been so far associated with only protocols and hence were part of the protocols/ api packages but I think it's time to move it to a common utility
- any other code suggestions

# Risk assessment

None to the existing system. New, isolated feature.


robot-server/robot_server/data_files/router.py
```diff
@@ -0,0 +1,143 @@
...
+    if file is None and file_path is None:
+        raise NoDataFileSourceProvided(
+            detail="You must provide either a file or a file_path in the request."
+        ).as_error(status.HTTP_422_UNPROCESSABLE_ENTITY)
+    try:
+        [buffered_file] = await file_reader_writer.read(files=[file or Path(file_path)])  # type: ignore[arg-type, list-item]
+    except FileNotFoundError as e:
+        raise FileNotFound(detail=str(e)).as_error(status.HTTP_404_NOT_FOUND) from e
+    # TODO (spp, 2024-06-18): probably also validate CSV file *contents*
+    if not buffered_file.name.endswith(".csv"):
```
jbleon95, line 105: I agree that we should be doing some validation on files before storing them on the robot and adding them to the dataFiles database, but only allowing CSVs to this endpoint seems limiting for now. We can of course change this to make it more permissive later, but maybe instead of limiting the types of files, we can either do something like:
- blacklist certain file types
- limit the size of files
- whitelist file paths on the robot to prevent any security risks

jbleon95 (approved):
Looks great! Only one concern about how we'll be limiting the POST /dataFiles endpoint but this is good to go, those can always be addressed in a future conversation and more discussion with the team and other developers.

robot-server/robot_server/data_files/router.py
sanni-t: I'll prefer to have a whitelist of select allowed file types rather than having a blacklist of disallowed ones so that we can be absolutely sure of the file types that get added, but ya I like these suggestions. Let's talk to the team and finalize allowed types, sizes and file paths.

PR #15482, opened 2024-06-20
feat(api): add liquid presence detection to load_instrument()

pmoegenburg:
# Overview

This PR adds a boolean argument to the load_instrument ProtocolContext method that enables liquid presence detection for the loaded pipette via an additional pipette state variable.

Closes EXEC-541.

# Test Plan

Added tests to ensure this bool arg can only be populated in API versions 2.20 onward and when using a Flex robot.

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -897,6 +898,7 @@ def load_instrument(
                              control <advanced-control>` applications. You cannot
                              replace an instrument in the middle of a protocol being run
                              from the Opentrons App or touchscreen.
+        :param bool liquid_presence_detection: If ``True``, enable liquid presence detection for instrument.
```
sfoster1: This should be noted as only being acceptable if the API version is above some level. It should also default to `None`, so that we can later raise an exception if the API level is below that value and the argument isn't None

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -931,6 +933,7 @@ def load_instrument(
         instrument_core = self._core.load_instrument(
             instrument_name=checked_instrument_name,
             mount=checked_mount,
+            liquid_presence_detection=liquid_presence_detection,
```
sfoster1: Before passing it in, we should check that if we're below the API level we added it in then the value is `None`. if it is, though, then `None` should become `False` when we pass to lower levels.

api/tests/opentrons/protocol_engine/state/test_pipette_view.py
```diff
@@ -62,6 +62,7 @@ def get_pipette_view(
     static_config_by_id: Optional[Dict[str, StaticPipetteConfig]] = None,
     flow_rates_by_id: Optional[Dict[str, FlowRates]] = None,
     nozzle_layout_by_id: Optional[Dict[str, Optional[NozzleMap]]] = None,
+    liquid_presence_detection_by_id: Optional[Dict[str, Optional[bool]]] = None,
```
sfoster1: why `Optional[bool]` rather than just `bool`?

sfoster1 (changes requested):
Couple changes needed to support the API level distinctions (and we should test them also)

api/tests/opentrons/protocol_engine/state/test_pipette_view.py
pmoegenburg: The LoadPipetteParams liquidPresenceDetection is Optional

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -928,9 +930,22 @@ def load_instrument(
             f"Loading {checked_instrument_name} on {checked_mount.name.lower()} mount"
         )
 
+        if (
+            self._core.robot_type != "OT-3 Standard"
+            and liquid_presence_detection is not None
+        ):
+            raise ValueError("Liquid presence detection only available on Flex robot.")
+        if (
+            self._api_version < APIVersion(2, 19)
```
sfoster1: oh, this will be 2.20 - 2.19 is being added in 7.3.1

api/src/opentrons/protocol_engine/state/pipettes.py
```diff
@@ -133,6 +133,7 @@ class PipetteState:
     static_config_by_id: Dict[str, StaticPipetteConfig]
     flow_rates_by_id: Dict[str, FlowRates]
     nozzle_configuration_by_id: Dict[str, Optional[NozzleMap]]
+    liquid_presence_detection_by_id: Dict[str, Optional[bool]]
```
sfoster1: let's make this non-optional. The argument to the `LoadPipette` command is optional, but let's reify that into a bool the same way we do in protocol context to make this just a bool

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -928,9 +930,22 @@ def load_instrument(
             f"Loading {checked_instrument_name} on {checked_mount.name.lower()} mount"
         )
 
+        if (
+            self._core.robot_type != "OT-3 Standard"
+            and liquid_presence_detection is not None
+        ):
+            raise ValueError("Liquid presence detection only available on Flex robot.")
```
sfoster1: This should be an enumerated error of some kind, not just a plain value error

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -928,9 +930,22 @@ def load_instrument(
...
+        if (
+            self._api_version < APIVersion(2, 19)
+            and liquid_presence_detection is not None
+        ):
+            raise APIVersionError(
```
sfoster1: This check should happen _before_ the flex-only check, because we want a protocol running on an ot-2 on 2.18 to raise the version error, while an ot-2 on 2.20 would raise the robot error

sfoster1 (changes requested):
Couple other nits

api/src/opentrons/protocol_api/protocol_context.py
```diff
@@ -928,9 +931,24 @@ def load_instrument(
             f"Loading {checked_instrument_name} on {checked_mount.name.lower()} mount"
         )
 
+        if (
+            self._api_version < APIVersion(2, 20)
+            and liquid_presence_detection is not None
+        ):
+            raise APIVersionError(
+                "Liquid Presence Detection is only supported on Flex protocols in API Version 2.20 and above."
```
sfoster1:
```suggestion
                "Liquid Presence Detection is only supported on Flex protocols in API Version 2.20 and above."
```
```suggestion
                "Liquid Presence Detection is only supported in API Version 2.20 and above."
```

api/src/opentrons/protocols/api_support/util.py
```diff
@@ -37,6 +37,12 @@
 MODULE_LOG = logging.getLogger(__name__)
 
 
+class RobotTypeError(Exception):
```
sfoster1: Looks like we've found some exceptions that need to be `EnumeratedError`s and aren't yet. We should make this exception inherit from [UnsupportedHardwareError](https://github.com/Opentrons/opentrons/blob/217268511518e3074ce4ca661739bc192099dd5e/shared-data/python/opentrons_shared_data/errors/exceptions.py#L968)

api/src/opentrons/protocols/api_support/util.py
```diff
@@ -37,6 +37,12 @@
...
+    """Error raised when a protocol attempts to access behavior not available to the robot type in use."""
+
+    pass
+
+
```
sfoster1, line 48: we should probably add a new api version error with a new code in shared-data  for APIVersionError, and make `UnsupportedAPIError` use [ApiRemoved](https://github.com/Opentrons/opentrons/blob/217268511518e3074ce4ca661739bc192099dd5e/shared-data/python/opentrons_shared_data/errors/exceptions.py#L900) or something (but look at how `UnsupportedApiError` is actually used, might be inappropriate). We can do this in a followup though.

sfoster1 (changes requested):
Couple last small things but looking good

sfoster1 (dismissed):
Last thing - API verison 2.20 doesn't exist yet. We'll need to add it, which will be pretty easy, but https://github.com/Opentrons/opentrons/pull/15507 should probably get merged first. So we can treat this as done for now and fix it up once that happens in the next day or two.

ryanthecoder (approved):
lgtm

PR #15488, opened 2024-06-21
feat(api): change liquid_probe to respect min well height

aaron-kulkarni:
Add logic to liquid_probe_in_place that will not go to the very bottom of the well, but stop
slightly above it(as specified by the well's minimum height)

fix EXEC-576

# Overview

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -413,6 +413,37 @@ def get_should_center_pipette_on_target_well(self, labware_id: str) -> bool:
             or len(self.get_definition(labware_id).wells) >= 96
         )
 
+    def get_well_min_height(self, labware_id: str, well_name: str) -> float:
+        """Get's the minimum distance that a liquid probe must stop away from the bottom of a well.
```
sfoster1:
```suggestion
        """Gets the minimum distance that a liquid probe must stop away from the bottom of a well.
```

api/src/opentrons/protocol_engine/state/labware.py
```diff
@@ -413,6 +413,37 @@ def get_should_center_pipette_on_target_well(self, labware_id: str) -> bool:
...
+        if len(height_reqs) == 0:
+            return default_val
+        if len(height_reqs) == 1:
+            return float(height_reqs[0]["value"])
+        for entry in height_reqs:
+            if well_name in entry["applicableWells"]:
+                return float(entry["value"])
+            if (
+                entry["applicableWells"] == []
+            ):  # A "custom" default value will have "applicableWells" set to []
```
sfoster1: will it? or will it have `applicableWells` set to a list of all the wells?

sfoster1 (dismissed):
Couple small issues, but the big one is: this needs testing. The state getter should have tests that make sure the right thing is returned in all 5 of the optional cases (no info, key but no values, one height requirement, full height requirement).

ryanthecoder (approved):
I Think this looks good

PR #15494, opened 2024-06-21
feat(hardware): Implement moving baseline readings for sensor-move

ryanthecoder:
# Overview

We are changing the pressure sensor timing during moves so that we can baseline the pressure value as the plunger is in motion, this lets us account for variations in tips more accurately. We also added a delay in turning off the pressure sensor readings after movement has finished so that we can attempt to tell better if we have hit liquid or labware bottom
# Test Plan

# Changelog

# Review requests

# Risk assessment


hardware/opentrons_hardware/hardware_control/tool_sensors.py
```diff
@@ -173,6 +173,20 @@ async def run_sync_buffer_to_csv(
     """Runs the sensor pass move group and creates a csv file with the results."""
     sensor_metadata = [0, 0, mount_speed, plunger_speed, threshold]
     positions = await move_group.run(can_messenger=messenger)
+    # wait a little to see the dropoff curve
+    await asyncio.sleep(0.15)
```
sfoster1, line 177: let's make this a passed-in configuration value please. also if this timing is critical we really should not be ensuring it in python in linux userspace with an asyncio sleep, those times are not guaranteed at all. could have a `request_data_after_delay` can message or something

hardware/opentrons_hardware/sensors/sensor_driver.py
```diff
@@ -250,14 +250,20 @@ async def __aexit__(self, *args: Any) -> None:
...
         """Wait for the data to stop, only use this with a send_accumulated_data_request."""
         self.event = asyncio.Event()
-        recieving = True
-        while recieving:
-            await asyncio.sleep(wait_time)
-            recieving = self.event.is_set()
-            self.event.clear()
+        self.expected_ack = message_index
+        start = time.time()
+        while True:
```
sfoster1:
```suggestion
         try:
             await asyncio.wait_for(event.wait(), wait_time)
         except asyncio.TimeoutError:
             pass
         self.event.clear()
```
?

sfoster1 (approved):
Some nits about waiting and about where some timing should live.

PR #15496, opened 2024-06-24
chore(hardware-testing): Gravimetric test supports ethanol, glycerol, and hexane

andySigler:
# Overview

Adds a few different liquids to the gravimetric test, specifically defining the relative humidity of each liquid type so that we can measure how many uL were aspirated/dispensed. This liquid type is set through an argument, and the choices are:

- water (default)
- ethanol
- glycerol
- hexane

All are considered to be 100% (not diluted). We could probably add something later to specify the percentage of water in each, and then linear scale the relative density based on that concentration.

# Test Plan

# Changelog

# Review requests

# Risk assessment


ryanthecoder (approved):
excited to see this

PR #15500, opened 2024-06-24
fix(app-shell-odd): quit the app when the render process is killed.

vegano1:
# Overview

While investigating an issue with the ODD, we noticed that the ODD electron app (opentrons-robot-app.service) continues to run even when the render process is killed for whatever reason. If we ever get into this state, we should quit the application so that the service exits and systemd can restart the service. This pull request goes in conjunction with this oe-core pr [Opentrons/oe-core#152](https://github.com/Opentrons/oe-core/pull/152).

# Test Plan

- [x] Make sure that the application exits when the render process is killed
- [x] Make sure that the `opentrons-robot-app.service` exits and restarts when the renderer is killed

# Changelog

- Quit the electron app when we get a `render-process-gone` event

# Review requests

- Makes sense?
- Any other times the renderer can be killed that we don't want this to happen, ie we have another window that gets closed or something?

# Risk assessment

Low, this should only happen when the renderer is killed.

### DIFF
--- a/app-shell-odd/src/main.ts
+++ b/app-shell-odd/src/main.ts
@@ -61,8 +61,8 @@ let rendererLogger: Logger
 app.prependOnceListener('ready', startUp)
 if (config.devtools) app.once('ready', installDevtools)
 
-app.once('window-all-closed', () => {
-  log.debug('all windows closed, quitting the app')
+function quitApplication(): void {
+  app.quit()
   closeBrokerConnection()
     .then(() => {
       app.quit()
@@ -71,6 +71,16 @@ app.once('window-all-closed', () => {
       log.warn('Failed to properly close MQTT connections:', error)
       app.quit()
     })
+}
+
+app.once('window-all-closed', () => {
+    log.debug('all windows closed, quitting the app')
+    quitApplication()
+})
+
+app.once('render-process-gone', () => {
+    log.debug('Renderer process has died, quitting the app')
+    quitApplication()
 })
 
 function startUp(): void {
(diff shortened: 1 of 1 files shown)


app-shell-odd/src/main.ts
```diff
@@ -71,6 +71,16 @@ app.once('window-all-closed', () => {
...
     })
+}
+
+app.once('window-all-closed', () => {
+    log.debug('all windows closed, quitting the app')
+    quitApplication()
+})
+
+app.once('render-process-gone', () => {
+    log.debug('Renderer process has died, quitting the app')
```
SyntaxColoring: Should this be a `log.error()`?

SyntaxColoring (approved):
Looks like you need to `make format-js`, but otherwise LGTM!

sfoster1 (approved):
Very nice! This also might make restarting the app with systemd a lot faster.

vegano1:
> Very nice! This also might make restarting the app with systemd a lot faster.

Not by itself, we needed to add 'TimeoutStopSec=10' in [this](https://github.com/Opentrons/oe-core/pull/152) oe-core pr to achieve that.

shlokamin (approved):
thanks @vegano1!

can we also log the cause of the render process crashing? from the PR: 

```js
* `event` Event
* `details` Object
  * `reason` String - The reason the render process is gone.  Possible values:
    * `clean-exit` - Process exited with an exit code of zero
    * `abnormal-exit` - Process exited with a non-zero exit code
    * `killed` - Process was sent a SIGTERM or otherwise killed externally
    * `crashed` - Process crashed
    * `oom` - Process ran out of memory
    * `launch-failure` - Process never successfully launched
    * `integrity-failure` - Windows code integrity checks failed
```

https://github.com/electron/electron/pull/23096/files

PR #15501, opened 2024-06-24
refactor(protocol-engine): Minor CommandStore/CommandView refactors

SyntaxColoring:
# Overview

This was originally going to fix EXEC-581, but it turns out that that behavior is intentional. So now this just adds a test for the intended behavior and makes a couple of other very minor refactors.

# Test plan

None needed.

# Review requests

See comments below.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/commands.py
+++ b/api/src/opentrons/protocol_engine/state/commands.py
@@ -380,15 +380,22 @@ class CommandStore(HasState[CommandState], HandlesActions):
 
         elif isinstance(action, StopAction):
             if not self._state.run_result:
-                if self._state.queue_status == QueueStatus.AWAITING_RECOVERY:
-                    self._state.recovery_target_command_id = None
+                self._state.recovery_target_command_id = None
 
-                self._state.queue_status = QueueStatus.PAUSED
                 if action.from_estop:
                     self._state.stopped_by_estop = True
                     self._state.run_result = RunResult.FAILED
+                elif self._state.queue_status in (
+                    QueueStatus.AWAITING_RECOVERY,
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_engine/state/commands.py
```diff
@@ -380,16 +380,23 @@ def handle_action(self, action: Action) -> None:  # noqa: C901
 
         elif isinstance(action, StopAction):
             if not self._state.run_result:
-                if self._state.queue_status == QueueStatus.AWAITING_RECOVERY:
-                    self._state.recovery_target_command_id = None
+                self._state.recovery_target_command_id = None
```
SyntaxColoring, line 383: @TamarZanzouri Do you remember if there was any particular reason this was conditional on `if self._state.queue_status == QueueStatus.AWAITING_RECOVERY`? I think it's safe to remove the `if` statement for simplicity, but am I missing anything?

https://github.com/Opentrons/opentrons/pull/15066

api/tests/opentrons/protocol_engine/state/test_command_state.py
```diff
@@ -770,3 +770,69 @@ def test_final_state_after_stop() -> None:
...
+        )
+    )
+    subject.handle_action(
+        actions.HardwareStoppedAction(
+            completed_at=sentinel.hardware_stopped_action_completed_at,
+            finish_error_details=None,
+        )
+    )
+    assert subject_view.get_status() == EngineStatus.FAILED
+    assert subject_view.get_error() is None
```
SyntaxColoring, line 804: Oops, I need to change this to return the underlying command error.

sfoster1 (approved):
Looks good!

PR #15503, opened 2024-06-25
fix(protocol-designer): dropTip location auto-generates upon import

jerader:
closes RESC-290

# Overview

This has existed in production for at least since 8.1.0. Basically, the drop tip location is being auto-generated incorrectly for an OT-2 protocol because the reducer was not finding a trash bin step. So it was auto-generating a trash bin entity with a `trashId` that did not match the current `trashId`. So the drop tip location from the imported protocol wouldn't populate correctly in the pipetting form because there was a new drop tip location generated.

# Test Plan

Upload the attached protocol and see that there is no timeline error. Upload the same protocol in PD in production and see that there is a timeline error.

[MakeBeads_v1.0_2X_SingleCycleTest10_MockReagents.json](https://github.com/user-attachments/files/15971111/MakeBeads_v1.0_2X_SingleCycleTest10_MockReagents.json)

# Changelog

- clean up the trash bin logic for if a trash bin is being used via dispense, blowout and dropTip for mix and move liquid. NOTE: i removed the aspirate because you can't aspirate into a trash bin anymore

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/src/step-forms/reducers/index.ts
+++ b/protocol-designer/src/step-forms/reducers/index.ts
@@ -1585,7 +1585,7 @@ export const additionalEquipmentInvariantProperties = handleActions<NormalizedAd
               ),
         },
       }
-      console.log('trashBin', trashBin)
+
       if (isFlex) {
         if (trashBin != null) {
           return {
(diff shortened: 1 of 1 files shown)


ncdiehl11 (approved):
code LGTM and test works as expected

PR #15506, opened 2024-06-25
feat(api): more robust CSV parsing and usage in run context for CSV parameters

jbleon95:
# Overview

Closes AUTH-419.

This PR fleshes out some of the work done by #15434 to add a `contents` property to the CSV parameter interface as used in the run context (e.g. a parameter called `csv_data` as would be accessed via `protocol_context.params.csv_data`), as well as more robustly parse the CSV file contents for usage with the `rows()` function.

CSV files can come in a number of different formats, called `dialects` by the Python CSV library. This determines things like the delimiter, how quotes are escaped, whether initial spaces are skipped etc. Because there is a lack of strict standardization in the format, any combination of those properties is still a valid CSV file. Fortunately the python CSV library provides a `Sniffer()` class that can try and detect which dialect is being used by being given a sample of the CSV file contents. For example, look at this CSV file example (all examples in this description taken from https://people.sc.fsu.edu/~jburkardt/data/csv/csv.html):

```
"Year", "Mileage (thousands)",  "Price"
1998,  27,    9991 
1997,  17,    9925 
1998,  28,   10491 
1998,   5,   10990 
1997,  38,    9493 
1997,  36,    9991 
1997,  24,   10490 
1997,  37,    9491 
1997,  38,    9491 
1997,  30,    9990 
1997,  38,    9491 
1997,  25,    9990 
1997,  39,    9990 
1997,  22,    9390 
1997,  24,    9990 
1997,  37,    9990 
1997,  29,    9990 
1997,  70,    8990 
1996,  29,    7990 
1995,  72,    5994 
1993,  72,    5994 
1994,  61,    5500 
1998,   7,   11000 

```

A pretty straightforward CSV file with the header strings in quotes and then three columns of numeric data. Yet without dialect parsing, the header and the first three rows look as such:

```
['Year', ' "Mileage (thousands)"', '  "Price"'],
['1998', '  27', '    9991 '], 
['1997', '  17', '    9925 '], 
['1998', '  28', '   10491 '],
```

Notice how the initial space is being taken as input and the quotes are then being included in the element for the header, and how the initial variable spacing before the second two elements of the data fields also include initial spaces. By providing the contents of the file to the sniffer, the output instead looks like:

```
['Year', 'Mileage (thousands)', 'Price'],
['1998', '27', '9991 '],
['1997', '17', '9925 '],
['1998', '28', '10491 ']
```

Again, not perfect with the trailing space at the end of the third elements of the data fields, but much more accurate to what the writer of this CSV file intended.

Unfortunately this sniffer functionality is not entirely perfect in determining the accurate dialect. For another example consider this next headerless and slightly deranged CSV file:

```
John,Doe,120 jefferson st.,Riverside, NJ, 08075
Jack,McGinnis,220 hobo Av.,Phila, PA,09119
"John ""Da Man""",Repici,120 Jefferson St.,Riverside, NJ,08075
Stephen,Tyler,"7452 Terrace ""At the Plaza"" road",SomeTown,SD, 91234
,Blankman,,SomeTown, SD, 00298
"Joan ""the bone"", Anne",Jet,"9th, at Terrace plc",Desert City,CO,00123

```

Without any dialect sniffing, the CSV reader parses this mostly correctly barring some spacing issues:
```
['John', 'Doe', '120 jefferson st.', 'Riverside', ' NJ', ' 08075'],
['Jack', 'McGinnis', '220 hobo Av.', 'Phila', ' PA', '09119'],
['John "Da Man"', 'Repici', '120 Jefferson St.', 'Riverside', ' NJ', '08075'],
['Stephen', 'Tyler', '7452 Terrace "At the Plaza" road', 'SomeTown', 'SD', ' 91234'],
['', 'Blankman', '', 'SomeTown', ' SD', ' 00298'], 
['Joan "the bone", Anne', 'Jet', '9th, at Terrace plc', 'Desert City', 'CO', '00123']
```

Yet with dialect sniffing it determines, due to the inconsistent spacing in the CSV file, that the delimiter is a literal space character and then parses it incredibly incorrectly like so:
```
['John,Doe,120', 'jefferson', 'st.,Riverside,', 'NJ,', '08075'],
['Jack,McGinnis,220', 'hobo', 'Av.,Phila,', 'PA,09119'],
['John "Da', 'Man""",Repici,120', 'Jefferson', 'St.,Riverside,', 'NJ,08075'],
['Stephen,Tyler,"7452', 'Terrace', 'At', 'the', 'Plaza""', 'road",SomeTown,SD,', '91234'],
[',Blankman,,SomeTown,', 'SD,', '00298'],
['Joan "the', 'bone"",', 'Anne",Jet,"9th,', 'at', 'Terrace', 'plc",Desert', 'City,CO,00123']
```

In order to provide the best balance of usability and user choice, we will default to trying to sniff the dialect while allowing any optional keyword arguments to be passed to the csv reader, whether `detect_dialect` is set or not.

TL;DR: CSV files are super inconsistent and we'll try our hardest to parse it but no guarantees if your data is super funky.

# Test Plan

Not of ton of end to end testing since the data file endpoint and setting parameters is not fully implemented, but tested the parsing on the above example files as well as others, plus unit testing.

# Changelog

- Added `contents` property to `CSVParameter` that returns one string of the entire contents of the file
- More robust CSV file content parsing for `CSVParameter.rows()`

# Review requests

If we can't parse the CSV rows, should we fail the protocol or should we just disallow access to `rows`? If so, how do we best communicate it.

# Risk assessment

Low.


api/src/opentrons/protocols/parameters/types.py
```diff
@@ -1,27 +1,61 @@
...
         self._rows = []
+        self._contents = ""
         if self._file is not None:
-            for row in csv.reader(self._file):
-                self._rows.append(row)
+            self._file.seek(0)
+            self._contents = self._file.read()
+            self._file.seek(0)
+            try:
+                dialect = csv.Sniffer().sniff(self._file.read(1024))
```
SyntaxColoring: > TL;DR: CSV files are super inconsistent and we'll try our hardest to parse it but no guarantees if your data is super funky.

> If we can't parse the CSV rows, should we fail the protocol or should we just disallow access to rows? If so, how do we best communicate it.

I think we should solve this by *not* mandatorily autodetecting the dialect. "In the face of ambiguity, refuse the temptation to guess." The same goes for the file encoding, which *something* must be assuming somewhere, for it to be a `TextIO` by the time it reaches here.

I think both should be left up to the protocol, because the protocol author potentially knows more than we and Python do about where their CSV files are coming from.

 I'm imagining an interface like:

```python
def run(protocol):
    csv_param: FileParam = protocol.params.csv

    # Open it as a Python file-like object:
    with csv_param.open_file(
        # t/b mode and encoding may have defaults
        mode="t", encoding="utf-8"
    ) as f:
        # ...

    # Or slurp all contents at once if you know the file is small:
    as_str = csv_param.read_str(encoding="utf-8")  # encoding may have a default
    as_bytes = csv_param.read_bytes()

    # Or parse the whole thing as CSV and return rows.
    # This is basically just a shortcut to feeding .as_file() into a csv.reader yourself.
    # This could have a bunch of configurable parameters but it could also still
    # default to Sniffer(), just as long as there's some way of overriding it.
    rows = csv_param.parse_as_csv(...)
```

Critlcally, we would *not* do any of this reading, decoding and parsing eagerly during protocol initialization, like we are now. We would wait until it's requested inside `run()`, so we'd know as much as we can about how to do it.

sanni-t: I too agree that we should allow the user to interact with the file as is, without assuming any dialects or formatting. The provision to access rows should be an additional feature that users can optionally take advantage of. In which case, we can enforce some rules for csv file format if one wants to use this feature.

I like @SyntaxColoring 's example suggestion. It definitely would be better to not have to sniff or even extract file contents during initialization. The protocol author might not even use them, in which case we would be performing these steps unnecessarily.

SyntaxColoring (approved):
Thanks, the latest changes make sense!

We discussed this in-person and you mentioned how although this moves the parsing to be in the protocol author's hands, we still assume UTF-8 text encoding. I think we would ideally also leave the text encoding in the author's hands (because why not, basically). But the interfaces you've added here can easily be extended to do that later on, when/if it ever becomes worth worrying about.

sanni-t (approved):
Nice! Changes look good!

PR #15516, opened 2024-06-25
fix(api): inherit protocol api errors from enumerated errors

pmoegenburg:
# Overview

This PR changes Protocol API APIVersionError to inherit from enumerated error IncorrectAPIVersion and Protocol API UnsupportedAPIError to inherit from enumerated error APIRemoved.

# Test Plan

# Changelog

# Review requests

# Risk assessment


api/src/opentrons/protocols/api_support/util.py
```diff
@@ -37,15 +41,15 @@
...
+class APIVersionError(IncorrectAPIVersion):
     """
     Error raised when a protocol attempts to access behavior not implemented
     """
 
     pass
 
 
-class UnsupportedAPIError(Exception):
+class UnsupportedAPIError(APIRemoved):
```
SyntaxColoring, line 57: It looks like this exception is currently unused, so it doesn't really matter, but note that "API removed" is not the only thing that could be meant by "a protocol attempts to use an unsupported API." It could also mean that the protocol attempts to use an API that's too new, like if you tried to use `"apiLevel": "2.19"` on robot software v5.0. (We're currently [just raising `ValueError` for that](https://github.com/Opentrons/opentrons/blob/fe022795bf2a071d30c04dd7fccf07ecc3d1a45e/api/src/opentrons/protocol_api/create_protocol_context.py#L86), but we probably shouldn't be.)

I wonder if we want just a generic code and exception for any protocol API version error. I'm not sure it's worth having 3 separate things for "too old," "too new," and "supported but invalid."

SyntaxColoring (approved):

sfoster1 (changes requested):
Looks good but you'll have to change the way you create the exception - take a look at the exception definition in shared-data for how to do that. You'll also have to add the code to the underlying json file in `shared-data/errors`.

api/src/opentrons/protocols/api_support/util.py
ecormany: Three is probably too many, yes. Two is probably good, breaking down into:
- Something is wrong with the combination of your code and API version.
- Your code used to be good but isn't anymore and furthermore that feature is never coming back ever ever so don't get your hopes up.

SyntaxColoring: But I think I'm trying to say that that specific breakdown of 2 is not sufficient to cover the case where your code *is* good, you just need to update your robot. That breakdown is covering problems like `protocol_apilevel < MIN_SUPPORTED_VERSION` and `not feature_available_in(protocol_apilevel)`, but is not covering the problem of `protocol_apilevel > MAX_SUPPORTED_VERSION`.

shared-data/python/opentrons_shared_data/errors/exceptions.py
```diff
@@ -921,6 +921,39 @@ def __init__(
...
+        message: Optional[str] = None,
+        detail: Optional[Dict[str, str]] = None,
+        wrapping: Optional[Sequence[EnumeratedError]] = None,
+    ) -> None:
+        """Build an IncorrectAPIVersion error."""
+        checked_detail: Dict[str, Any] = detail or {}
+        checked_detail["identifier"] = api_element
+        checked_detail["until_version"] = until_version
+        checked_detail["current_version"] = current_version
+        # make this cleaner?!
```
sfoster1: Remove or make a proper todo comment (I think the code is fine as is though)

sfoster1 (approved):
Looks good once you clean up dev comments

api/src/opentrons/protocol_api/deck.py
```diff
@@ -107,8 +107,9 @@ def __delitem__(self, key: DeckLocation) -> None:
             # * PAPIv2.14 (Protocol Engine): No
             # * PAPIv2.15 (Protocol Engine): Yes
             raise APIVersionError(
-                f"Deleting deck elements is not supported with apiLevel {self._api_version}."
-                f" Try increasing your apiLevel to {APIVersion(2, 15)}."
+                api_element="Deleting deck elements",
+                until_version="2.15",
```
ecormany, line 111: Wait, was `del protocol.deck[1]` actually broken in 2.14? I don't think we ever documented that.

sfoster1: Yes, definitely

ecormany: Yeah, all we have is this:
> Changed in version 2.14: Includes the Thermocycler in all of the slots it occupies.
>
> Changed in version 2.15: `del` sets the corresponding labware’s location to `OFF_DECK`.

We could update the versionchanged notice…or not.

sfoster1 (approved):
Nice work!

PR #15526, opened 2024-06-26
feat(abr-testing): Download Weston log during error recording

rclarke0:
# Overview

Uses scp command to download weston.log during error recording.

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Nice, thank you!

PR #15527, opened 2024-06-26
feat(app, api-client, react-api-client): send CSV RTP file to robot before creating protocol and run

ncdiehl11:
# Overview

The new RTP paradigm for utilizing CSV files as runtime parameters requires sending the ID of the file to be used as the value in the `runTimeParameterValues` object sent with `createProtocol` and `createRun`. Here, we send any selected RTP files to the new `/dataFiles` endpoint and await the response containing an ID for each file. We then associate the returned ID with its respective runtime parameter variable name, and send that key:value pair along with the other value runtime parameters.

# Test Plan

This is a little tricky to test right now because the `/dataFiles` endpoint is not yet wired up, and the `/protocols` and `/runs` endpoint only accepts `<variableName>: <value>` pairs for `runTimeParameterValues`, not the new format of `<variableName>: {file_id: <id>}` for CSV parameters.

#### Protocol card and detail
- upload protocol that requires CSV file parameter like this:
[rtp_tests.py.zip](https://github.com/user-attachments/files/16018648/rtp_tests.py.zip)

- verify that `CSV file required` banner renders on protocol card
- select protocol card
- verify that `CSV file required` banner renders on protocol details

# Changelog

- post CSV files from RTPs when confirming values on both `ChooseProtocolSlideout` and `ChooseRobotToRunProtocolSlideout`, and use returned IDs in `runTimeParametersValues` sent with `createRunFromProtocolSource`
- fix MiniCard warning text margins on above slideouts
- update `getAnalysisStatus` util to return new `parameterRequired` status when analysis result is `parameter-value-required` 
- refactor `useCreateProtocolMutation` to accept `runTimeParameterValues` argument in its returned mutation's callback rather than in the hook's invocation itself
- update tests

# Review requests

Go through flow of uploading a CSV parameter protocol according to test plan above. Expected behavior is for a `runCreationError` to be raised when confirming value because the robot server does not yet know what to do with CSV parameter runTimeParameterValues overrides.

# Risk assessment

medium

### DIFF
--- a/api-client/src/dataFiles/uploadCsvFile.ts
+++ b/api-client/src/dataFiles/uploadCsvFile.ts
@@ -27,9 +27,11 @@ export function uploadCsvFile(
   const fileId = uuidv4()
   const stub = {
     data: {
-      id: fileId,
-      createdAt: '2024-06-07T19:19:56.268029+00:00',
-      name: 'rtp_mock_file.csv',
+      data: {
+        id: fileId,
+        createdAt: '2024-06-07T19:19:56.268029+00:00',
+        name: 'rtp_mock_file.csv',
+      },
     },
   }
   return Promise.resolve(stub)
--- a/api-client/src/runs/types.ts
+++ b/api-client/src/runs/types.ts
@@ -132,7 +132,7 @@ export interface LabwareOffsetCreateData {
   vector: VectorOffset
 }
 
-type RunTimeParameterValueType = string | number | boolean | { file_id: string }
+type RunTimeParameterValueType = string | number | boolean | { fileId: string }
 export type RunTimeParameterCreateData = Record<
   string,
   RunTimeParameterValueType
--- a/api-client/src/runs/types.ts
+++ b/api-client/src/runs/types.ts
@@ -132,13 +132,10 @@ export interface LabwareOffsetCreateData {
   vector: VectorOffset
 }
 
-type FileRunTimeParameterCreateData = Record<string, string | number | boolean>
-
-type ValueRunTimeParameterCreateData = Record<string, { id: string }>
-
-export type RunTimeParameterCreateData =
-  | FileRunTimeParameterCreateData
-  | ValueRunTimeParameterCreateData
+export type RunTimeParameterCreateData = Record<
+  string,
+  string | number | boolean | { file_id: string }
+>
 
 export interface CommandData {
   data: RunTimeCommand
--- a/api-client/src/runs/types.ts
+++ b/api-client/src/runs/types.ts
@@ -132,9 +132,10 @@ export interface LabwareOffsetCreateData {
   vector: VectorOffset
 }
 
+type RunTimeParameterValueType = string | number | boolean | { file_id: string }
 export type RunTimeParameterCreateData = Record<
   string,
-  string | number | boolean | { file_id: string }
+  RunTimeParameterValueType
 >
 
 export interface CommandData {
--- a/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
@@ -164,7 +164,7 @@ export function ChooseRobotToRunProtocolSlideoutComponent(
             const [response, varName] = responseTuple
             return {
               ...acc,
-              [varName]: { file_id: response.data.id },
+              [varName]: { fileId: response.data.id },
             }
           },
           runTimeParameterValues
--- a/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
@@ -168,7 +168,7 @@ export function ChooseRobotToRunProtocolSlideoutComponent(
       createRunFromProtocolSource({
         files: srcFileObjects,
         protocolKey,
-        runTimeParameterValues: runTimeParameterValuesWithFiles,
+        runTimeParameterValues, // : runTimeParameterValuesWithFiles,
       })
     })
   }
--- a/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
+++ b/app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
@@ -18,6 +18,7 @@ import { useUploadCsvFileMutation } from '@opentrons/react-api-client'
 
 import { Tooltip } from '../../atoms/Tooltip'
 import { getRobotUpdateDisplayInfo } from '../../redux/robot-update'
+import { useFeatureFlag } from '../../redux/config'
 import { OPENTRONS_USB } from '../../redux/discovery'
 import { appShellRequestor } from '../../redux/shell/remote'
 import { useTrackCreateProtocolRunEvent } from '../Devices/hooks'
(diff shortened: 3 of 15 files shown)


api-client/src/runs/types.ts
```diff
@@ -132,13 +132,10 @@ export interface LabwareOffsetCreateData {
...
-type FileRunTimeParameterCreateData = Record<string, string | number | boolean>
-
-type ValueRunTimeParameterCreateData = Record<string, { id: string }>
-
-export type RunTimeParameterCreateData =
-  | FileRunTimeParameterCreateData
-  | ValueRunTimeParameterCreateData
+export type RunTimeParameterCreateData = Record<
+  string,
+  string | number | boolean | { file_id: string }
```
jerader: i think it'd read easier if `string | number | boolean | { file_id: string }` was split up into its own type, just because there are so many types that the RTP can be

ncdiehl11: Are you thinking something like:
```
type StringRunTimeParameterValueType = Record<string, string>
type NumberRunTimeParameterValueType = Record<string, number>
type BooleanRunTimeParameterValueType = Record<string, boolean>
interface FileRunTimeParameterValueType {
  file_id: string
}
export type RunTimeParameterCreateData = Record<
  string,
  | StringRunTimeParameterValueType
  | NumberRunTimeParameterValueType
  | BooleanRunTimeParameterValueType
  | FileRunTimeParameterValueType
>

```

ncdiehl11: Are you thinking something like
```
type StringRunTimeParameterValueType = string
type NumberRunTimeParameterValueType = number
type BooleanRunTimeParameterValueType = boolean
interface FileRunTimeParameterValueType {
  file_id: string
}
export type RunTimeParameterCreateData = Record<
  string,
  | StringRunTimeParameterValueType
  | NumberRunTimeParameterValueType
  | BooleanRunTimeParameterValueType
  | FileRunTimeParameterValueType
>
```

api-client/src/dataFiles/uploadCsvFile.ts
```diff
@@ -27,9 +27,11 @@ export function uploadCsvFile(
   const fileId = uuidv4()
   const stub = {
     data: {
-      id: fileId,
-      createdAt: '2024-06-07T19:19:56.268029+00:00',
-      name: 'rtp_mock_file.csv',
+      data: {
```
koji, line 30: 👍

api-client/src/runs/types.ts
koji: can we use camel case instead of snake case for file id?

ncdiehl11: something like:
```
type RunTimeParameterValueType = string | number | boolean | { file_id: string }
export type RunTimeParameterCreateData = Record<
  string,
  RunTimeParameterValueType
>
```

ncdiehl11: @koji I will confirm with robot server. I saw `file_id` initially

app/src/organisms/ChooseRobotToRunProtocolSlideout/index.tsx
koji, line 1: Can we use `useFeatureFlag`?

app/src/organisms/ProtocolDetails/index.tsx
```diff
@@ -434,7 +435,7 @@ export function ProtocolDetails(
               ) : null}
               {analysisStatus !== 'loading' &&
```
koji: do we need this line when the condition has `analysisStatus === 'error'`?

app/src/organisms/ProtocolsLanding/ProtocolCard.tsx
```diff
@@ -217,7 +224,7 @@ function AnalysisInfo(props: AnalysisInfoProps): JSX.Element {
       >
         {/* error and protocol name section */}
         <Flex flexDirection={DIRECTION_COLUMN} gridGap={SPACING.spacing8}>
-          {mostRecentAnalysis?.result === 'parameter-value-required' ? (
+          {analysisStatus === 'parameterRequired' ? (
```
koji, line 227: 👍

app/src/organisms/ProtocolsLanding/utils.ts
```diff
@@ -16,7 +22,13 @@ export function getAnalysisStatus(
...
+    if (analysis.result === 'parameter-value-required') {
+      return 'parameterRequired'
+    } else if (analysis.errors.length > 0) {
+      return 'error'
+    } else {
+      return 'complete'
+    }
   } else {
     return 'missing'
   }
```
koji, line 22: nit

```ts
  if (isAnalyzing) {
    return 'loading';
  }

  if (analysis == null) {
    return 'missing';
  }

  if (analysis.liquids == null || analysis.runTimeParameters == null) {
    return 'stale';
  }

  if (analysis.result === 'parameter-value-required') {
    return 'parameterRequired';
  }

  if (analysis.errors.length > 0) {
    return 'error';
  }

  return 'complete';
```

react-api-client/src/dataFiles/__tests__/useUploadCsvFileMutation.test.tsx
```diff
@@ -42,29 +42,23 @@ describe('useUploadCsvFileMutation', () => {
...
-    )
+    const { result } = renderHook(() => useUploadCsvFileMutation(), {
+      wrapper,
+    })
     expect(result.current.data).toBeUndefined()
-    result.current.uploadCsvFile(mockFilePath)
+    result.current.uploadCsvFile(mockFilePath).catch(_ => {})
     await waitFor(() => {
-      expect(result.current.data).toBeUndefined()
+      expect(result.current.error).toBe('oh no')
```
koji, line 65: nit
maybe check both?

```ts
expect(result.current.data).toBeUndefined()
expect(result.current.error).toBe('oh no')
```

react-api-client/src/dataFiles/__tests__/useUploadCsvFileMutation.test.tsx
koji, line 1: can we add a success test case for file (upload a csv file)?

app/src/organisms/ProtocolDetails/index.tsx
ncdiehl11: Nope, good catch

app/src/organisms/ProtocolsLanding/utils.ts
ncdiehl11: this is a cleaner refactor. thanks

koji:
@ncdiehl11 

The design doesn't have info icon next to `CSV file`. Maybe we don't need to display that for CSV file?

Screenshot 2024-06-27 at 12 37 45 PM

ncdiehl11:
> @ncdiehl11
> 
> The design doesn't have info icon next to `CSV file`. Maybe we don't need to display that for CSV file?
> 
> Screenshot 2024-06-27 at 12 37 45 PM

@koji Do you mean for the description?


I mean literary the icon.
Your branch displays icon next to `CSV Data`. But in the design there is no icon (the above screenshot).
I'm not 100% sure that removing the icon is intentional or not.

Screenshot 2024-06-27 at 12 43 14 PM


Another is `Name` of csv file.
The design says `CSV File` which I'm also not 100% sure this is the fixed text or `displayName` since csv file row is the unique can be the fixed text or `displayName`.
Probably we would need to ask about this to Rob/Mel.

koji (approved):
left a comment about Parameter Table's csv file part, but the changes look good to me.

PR #15530, opened 2024-06-26
feat(ai-server): use env variables for the cpu and memory of the task

y3rsh:
## Use variables for cpu and memory

> [!NOTE]
> It is easy to increase the number of tasks running in a given cluster with the AWS console so we won't automate that part now.

- [x] Add cpu and memory to settings and use the value at task definition creation
- [x] Update dev env variables secret
- [x] Update prod env variables secret

### Once merged 
1. see that staging has the correct values.
2. tag and see the prod deployment has the correct values.


shlokamin (approved):
lgtm

PR #15532, opened 2024-06-26
feat(api): Update robot deck extents to check the full pipette bounds

Laura-Danielle:
# Overview

Previously, we were only checking for a very specific set of bounds for the robot. Now, we should be able to check the exact "absolute" value that a pipette type can move to and throw and error if we exceed those bounds.

## Questions

1. I have one failing test, I'm not sure if this means the extents I'm using aren't 100% correct or if the older ones were overly generous
2. Should I bump the deck version?
3. In AddressableState there are a lot of free floating functions that access the deck definition components directly, I wasn't sure why we were doing this but if you want me to keep the same format I can.
4. I will need assistance at some point to get my JS env up and running again as I cannot run make format

# Test Plan

- Put a bunch of protocols through an app and see whether the software correctly determines if something is out of bounds

# Changelog

- Added two new components to the deck definition for 'extents' and 'mount offsets'
- Added accessors for that in the protocol engine
- Modified `_is_within_pipette_extents` function to check the absolute bound per pipette type per mount without any offsets

# Review requests

Pls check my math, I'm almost sure I did one or two things wrong

# Risk assessment

Medium. If we don't get this right it could throw erroneous deck bound errors.

### DIFF
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -110,8 +110,8 @@ class GeometryView:
         right_offset = self._addressable_areas.mount_offsets["right"]
 
         front_left_abs = {
-            MountType.LEFT: left_offset,
-            MountType.RIGHT: right_offset,
+            MountType.LEFT: Point(left_offset.x, -1 * left_offset.y, left_offset.z),
+            MountType.RIGHT: Point(right_offset.x, -1 * right_offset.y, right_offset.z),
         }
         back_right_abs = {
             MountType.LEFT: self._addressable_areas.deck_extents + left_offset,
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -95,6 +95,10 @@ class GeometryView:
         self._addressable_areas = addressable_area_view
         self._last_drop_tip_location_spot: Dict[str, _TipDropSection] = {}
 
+    @property
+    def deck_extents(self) -> Point:
+        return self._addressable_areas.deck_extents
+
     def get_labware_highest_z(self, labware_id: str) -> float:
         """Get the highest Z-point of a labware."""
         labware_data = self._labware.get(labware_id)
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -108,16 +108,14 @@ class GeometryView:
         """The absolute deck extents for a given robot deck."""
         left_offset = self._addressable_areas.mount_offsets["left"]
         right_offset = self._addressable_areas.mount_offsets["right"]
-        left_changed_x_component = Point(-1 * left_offset.x, left_offset.y, left_offset.z)
-        right_changed_x_component = Point(-1 * right_offset.x, right_offset.y, right_offset.z)
 
-        back_right_abs = {
-            MountType.LEFT: left_changed_x_component,
-            MountType.RIGHT: right_changed_x_component,
-        }
         front_left_abs = {
-            MountType.LEFT: self._addressable_areas.deck_extents + left_changed_x_component,
-            MountType.RIGHT: self._addressable_areas.deck_extents + right_changed_x_component,
+            MountType.LEFT: left_offset,
+            MountType.RIGHT: right_offset,
+        }
+        back_right_abs = {
+            MountType.LEFT: self._addressable_areas.deck_extents + left_offset,
+            MountType.RIGHT: self._addressable_areas.deck_extents + right_offset,
         }
         return _AbsoluteRobotExtents(
             front_left=front_left_abs, back_right=back_right_abs
--- a/api/src/opentrons/protocol_engine/state/geometry.py
+++ b/api/src/opentrons/protocol_engine/state/geometry.py
@@ -3,6 +3,8 @@ import enum
 from numpy import array, dot, double as npdouble
 from numpy.typing import NDArray
 from typing import Optional, List, Tuple, Union, cast, TypeVar, Dict
+from dataclasses import dataclass
+from functools import cached_property
 
 from opentrons.types import Point, DeckSlotName, StagingSlotName, MountType
 
@@ -71,6 +73,12 @@ class _GripperMoveType(enum.Enum):
     DROP_LABWARE = enum.auto()
 
 
+@dataclass
+class _AbsoluteRobotExtents:
+    front_left: Dict[MountType, Point]
+    back_right: Dict[MountType, Point]
+
+
 _LabwareLocation = TypeVar("_LabwareLocation", bound=LabwareLocation)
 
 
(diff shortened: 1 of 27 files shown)


shared-data/python/opentrons_shared_data/deck/dev_types.py
```diff
@@ -26,9 +26,15 @@ class Metadata(TypedDict, total=False):
     displayName: str
     tags: List[str]
 
+class mountOffset(TypedDict):
+    left: List[float]
+    right: List[float]
+    gripper: Optional[List[float]]
```
SyntaxColoring: I think this needs to be `typing_extensions.NotRequired` instead of `Optional`. The way the schema is defined right now, the `"gripper"` key can be missing entirely, not `null`/`None`.

shared-data/deck/schemas/5.json
```diff
@@ -98,12 +98,35 @@
...
           "description": "Model of the robot",
           "type": "string",
           "enum": ["OT-2 Standard", "OT-3 Standard"]
+        },
+        "extents": {
+          "description": "The maximum addressable coordinates",
+          "$ref": "#/definitions/xyzArray"
+        },
+        "mountOffsets": {
+          "description": "The physical mount offsets of the robot",
```
SyntaxColoring: Are there more details that make sense to provide here? Like, for example, `left` is `[-13.5, -60.5, 255.675]`. But `[-13.5, -60.5, 255.675]` is the vector from where to where? Is there any specific physical reference point?

api/tests/opentrons/protocol_api_integration/test_pipette_movement_deck_conflicts.py
```diff
@@ -82,6 +82,7 @@ def test_deck_conflicts_for_96_ch_a12_column_configuration() -> None:
 
     # Will NOT raise error since first column of TC labware is accessible
     # (it is just a few mm away from the left bound)
+    # TODO Confirm if this location is valid after bounds updates
```
SyntaxColoring: Hm, concerning if this is failing. @sanni-t Any thoughts?

shared-data/deck/definitions/5/ot2_standard.json
SyntaxColoring, line 1: I see we have [older code](https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/hardware_control/util.py#L67) that claims to do bounds checking. Where does it get its data from? Does it need to be deduplicated with this, or have `# todo` comments describing how to do the deduplication in the future?

shared-data/deck/schemas/5.json
```diff
@@ -98,12 +98,35 @@
...
-      "required": ["model"],
+      "required": ["model", "extents", "mountOffsets"],
       "properties": {
         "model": {
           "description": "Model of the robot",
           "type": "string",
           "enum": ["OT-2 Standard", "OT-3 Standard"]
+        },
+        "extents": {
+          "description": "The maximum addressable coordinates",
```
SyntaxColoring: In other words, are these the maximum coordinates with just the gantry itself, and no instruments attached?

SyntaxColoring (commented):
Yay! Some questions:

shared-data/deck/definitions/5/ot2_standard.json
sanni-t: Good question. Would like to know too

api/tests/opentrons/protocol_api_integration/test_pipette_movement_deck_conflicts.py
SyntaxColoring: Oh and we have some analysis snapshots failing, too. https://github.com/Opentrons/opentrons/pull/15533

api/src/opentrons/protocol_engine/state/geometry.py
```diff
@@ -95,6 +103,26 @@ def __init__(
         self._addressable_areas = addressable_area_view
         self._last_drop_tip_location_spot: Dict[str, _TipDropSection] = {}
 
+    @cached_property
+    def absolute_deck_extents(self) -> _AbsoluteRobotExtents:
```
sfoster1, line 107: I really don't like this, because I don't get the coordinate space it lives in.

I would expect the value this returns to define a map from mount type to a region in deck coordinates, and documented as such. But that doesn't seem to be what this does - instead, the region returned here appears to be defined such that (0,0) is in the back right corner and the front left corner is (+ve, +ve). 

I'm also confused about the way the entire region, both corners, is offset for each mount. It feels like it should be the case that the following regions have a strictly diminishing area:
- the `deck_extents` from the definition
- the area accessible to the left mount = the area accessible to the right mount
- the area accessible by a given pipette on a given mount


But here, it seems like the mount-offset-adjusted regions are the same size as the overall extents because both corners are being shifted. It seems like for the right mount, the left corner would be shifted in and the right corner left alone; and for the left mount, the right corner would be shifted in and the left corner left alone.

shared-data/deck/definitions/5/ot3_standard.json
```diff
@@ -8,7 +8,13 @@
     "tags": ["ot3", "12 slots", "standard"]
   },
   "robot": {
-    "model": "OT-3 Standard"
+    "model": "OT-3 Standard",
+    "extents": [477.20, 493.8, 0.0],
+    "mountOffsets": {
```
sfoster1: I don't like the mount offsets being here, and I don't love the extends though I'm fundamentally ok with it. Note that we do actually have [robot definitions](https://github.com/Opentrons/opentrons/blob/edge/shared-data/robot/definitions/1/ot3.json) in shared data - could we maybe put these things there instead?

sfoster1 (changes requested):
Have some inline comments about some details of this, but it has a more fundamental problem: it doesn't take mount offset or pipette offset calibration into account. This may be the cause of the test failure, but it will definitely be the cause of weird errors around interactions near the edge of the deck. I think we need to get at least some of the mount data from the hardware controller so it can fold in the calibration.

shared-data/deck/schemas/5.json
Laura-Danielle: Will add more details!

shared-data/deck/definitions/5/ot2_standard.json
Laura-Danielle: Will check it out and try to summon my brain power to remember

shared-data/deck/definitions/5/ot3_standard.json
Laura-Danielle: Sure that works for me as well.

api/src/opentrons/protocol_engine/state/geometry.py
Laura-Danielle: This function should not be dealing with specific pipette offsets at all, purely the data coming from deck map and/or robot info (wherever it lands in shared data) from the perspective of the mounts. It should return the max extents of the left pip and right pip mount. The reason for the X component direction change is because we want the mount positions in the front left corner based on that being the "maximum" extents, but the original values are from the relative position of the middle of the carriage.

I can maybe change the name to 'MountExtents' if that makes more sense?

sfoster1: If it's going to be in protocol engine then it needs to be in deck coordinates. Nothing above the hardware controller should know about machine coordinates. That means that even if these are displacements from the home position, they should be negative not positive; and if they're absolute coordinates, (0, 0) needs to be in the front left.

shared-data/deck/schemas/5.json
Laura-Danielle: That's correct, I can specify that in the description.

api/tests/opentrons/protocol_api_integration/test_pipette_movement_deck_conflicts.py
Laura-Danielle: This test is actually super off and failing correctly. We have a column 12 configuration moving to the first column of a plate all the way on the east side of the robot.

CaseyBatten (approved):
Based on changes made and testing executed I believe this satisfies our need to prevent the robot from exceeding the deck extents, particularly in regards to actions involving partial tips commands. The following behavior was tested on a Flex and performed as expected:
- Attempting Single tip pickup using Nozzle A1 of an 8ch pipette on a tiprack in D3 failed analysis
- Attempting Single tip pickup using Nozzles A1 and A12 as well as Row A1 of a 96ch on a tiprack in D3 failed analysis
- Attempting Single tip pickup using Nozzle H1 of a 96ch on a tiprack in A3 failed analysis
- Attempting Single tip pickup using Nozzle H12 of a 96ch on a tiprack in A1 failed analysis
- Attempting a Row A1 pickup on Well E of a labware in D3 passed analysis and executed correctly on the robot

With each of these cases performing as expected, I believe we can safely move on to QA testing of the deck extents. Thanks Laura!

shared-data/deck/definitions/5/ot2_standard.json
Laura-Danielle: So that check is determining whether the target position is outside of the bounds of the robot but it is only a point and doesn't actually check for the pipette bounding box.

I think it's OK to leave in since it's checking a different thing.

sfoster1 (approved):
This is excellent, thank you for taking the reviews into account! I'd love to see a followup that takes mount offset values from these deck defs rather than python constants but that definitely doesn't have to be part of this PR.

PR #15535, opened 2024-06-27
feat(protocol-designer): multiple h-s and magnetic block support

jerader:
closes RAUT-1048

# Overview

Support for multiple heater-shaker and magnetic blocks for Flex only in protocol-designer

# Test Plan

Should be able to add up to 10 magnetic blocks and 7 heater-shakers. The wizard modal and edit modules should work as expected. If you add commands for the heater-shakers, it should do so as expected and properly differentiate between the 2.

# Changelog

- make moam logic generic for other modules
- fix tests

# Review requests

see test plan

# Risk assessment

low


protocol-designer/src/components/modals/CreateFileWizard/utils.ts
```diff
@@ -76,15 +76,18 @@ export const getNumSlotsAvailable = (
...
   let filteredModuleLength = modules != null ? Object.keys(modules).length : 0
   if (hasTC) {
     filteredModuleLength = filteredModuleLength + 1
   }
-  if (hasMagneticBlock) {
-    filteredModuleLength = filteredModuleLength - 1
+  if (magneticBlocks.length > 0) {
+    //  once blocks exceed 4, then we dont' want to subtract the amount available
+    //  because block can go into the center slots where all other modules/trashes can not
+    const numBlocks = magneticBlocks.length > 4 ? 4 : magneticBlocks.length
```
koji, line 89: nit
define a new const for `4`?

koji (approved):
lgtm
tested the sandbox and it worked as expected.  (tried something similar to what Derek did before)
nice work!

PR #15537, opened 2024-06-27
feat(api): add getter and setter for instrument lld setting

aaron-kulkarni:
# Overview

fix #[EXEC-542](https://opentrons.atlassian.net/browse/EXEC-542) and fix #[EXEC-543](https://opentrons.atlassian.net/browse/EXEC-543)

These change implement a getter and setter for a pipette's lld setting. Since they do not have to be called at the time you load the instrument, they allow for finer control of when LLD is active.
# Test Plan

Added Unit Tests to test_instrument_context.py to ensure that the functions properly interacted with the underlying InstrumentCore.

# Changelog

# Review requests

# Risk assessment

[EXEC-542]: https://opentrons.atlassian.net/browse/EXEC-542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EXEC-543]: https://opentrons.atlassian.net/browse/EXEC-543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1669,6 +1669,24 @@ class InstrumentContext(publisher.CommandPublisher):
     @tip_racks.setter
     def tip_racks(self, racks: List[labware.Labware]) -> None:
         self._tip_racks = racks
+    
+    @property
+    @requires_version(2, 0)
+    def liquid_detection(self) -> bool:
+        """
+        Gets the global setting for liquid level detection.
+        
+        When True, `liquid_probe` will be called before
+        aspirates and dispenses to bring the tip to the liquid level. 
+        
+        The default value is False.
+        """
+        return self._core.get_liquid_presence_detection_configuration()
+        
+    @liquid_detection.setter
+    def configure_liquid_detection(self, enable: bool) -> None:
+        self._core.set_liquid_presence_detection_configuration(enable)
+        
 
     @property
     @requires_version(2, 0)
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1671,7 +1671,7 @@ class InstrumentContext(publisher.CommandPublisher):
         self._tip_racks = racks
 
     @property
-    @requires_version(2, 2)
+    @requires_version(2, 20)
     def liquid_detection(self) -> bool:
         """
         Gets the global setting for liquid level detection.
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -1684,6 +1684,7 @@ class InstrumentContext(publisher.CommandPublisher):
         return self._core.get_liquid_presence_detection()
 
     @liquid_detection.setter
+    @requires_version(2, 20)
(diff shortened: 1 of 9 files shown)


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1670,6 +1670,23 @@ def tip_racks(self) -> List[labware.Labware]:
     def tip_racks(self, racks: List[labware.Labware]) -> None:
         self._tip_racks = racks
 
+    @property
+    @requires_version(2, 2)
```
sfoster1:
```suggestion
    @requires_version(2, 20)
```

sfoster1 (changes requested):
Really great start! Let's add some quick tests that will serve as regressions and make sure we didn't forget an implementation, and please add a nice descriptive PR description, but solid!

api/tests/opentrons/protocol_api/test_instrument_context.py
```diff
@@ -1059,6 +1059,15 @@ def test_flow_rate(
     assert result == flow_rates
 
 
+def test_liquid_presence_detection(
+    decoy: Decoy, mock_instrument_core: InstrumentCore, subject: InstrumentContext
+) -> None:
+    """It should have a default liquid presence detection boolean set to False."""
+    decoy.when(mock_instrument_core.get_liquid_presence_detection()).then_return(False)
+    assert subject.liquid_detection is False
+    decoy.when(mock_instrument_core.set_liquid_presence_detection(True)).then_return()
```
sfoster1: maybe set the underlying value to True and then check that the property returns True now

sfoster1 (changes requested):
One more tiny test improvement and then I think we're good

api/tests/opentrons/protocol_api/test_instrument_context.py
```diff
@@ -1059,6 +1059,17 @@ def test_flow_rate(
...
 
 
+def test_liquid_presence_detection(
+    decoy: Decoy, mock_instrument_core: InstrumentCore, subject: InstrumentContext
+) -> None:
+    """It should have a default liquid presence detection boolean set to False."""
+    decoy.when(mock_instrument_core.get_liquid_presence_detection()).then_return(False)
+    assert subject.liquid_detection is False
+    subject.liquid_detection = True
+    decoy.when(mock_instrument_core.get_liquid_presence_detection()).then_return(True)
```
sfoster1: we should also add a `decoy.verify(mock_instrument_core.set_liquid_presence_detection(False)` to make sure we called it

sfoster1: When writing tests, it's usually a good idea to doublecheck that the test is write by breaking the implementation (i.e., change line 1688 of `instrument_context.py` to `pass`, or to always set liquid presence detection to `False`) and make sure that the test fails in that case

sfoster1 (changes requested):
Let's make sure that the test fails appropriately if the code is wrong

sfoster1 (approved):
Awesome, nice work!

PR #15538, opened 2024-06-27
feat(app): add quick transfer advanced settings tab + aspirate/dispense flowrate, and pipette path options.

vegano1:
# Overview

The "Quick Transfer™" flow will let you set advanced settings before you create the protocol, this pull request sets up the Advanced Settings tab and the first few advanced settings options (aspirate flowrate, dispense flowrate, pipette path). A follow-up PR will add the rest of the advanced settings and interactions.

Closes: [PLAT-221](https://opentrons.atlassian.net/browse/PLAT-221)

# Test Plan

- [x] Push to Flex robot and make sure we can create a Quick Transfer™ protocol
- [x] Create a Quick Transfer™ protocol, change the aspirate flow rate to a non-default value, and ensure the protocol engine adheres to the change.
- [x] Create a Quick Transfer™ protocol, change the pipette path from single to multi-dispense, and make sure we use one tip for multiple-dispenses.
- [x] Create a Quick Transfer™ protocol, change the pipette path from single to multi-aspirate, and make sure we use one tip for multi-aspirates
- [x] Make sure the "pipette path" option is not set for single transfers

# Changelog

- Add Quick Transfer™ Advanced Settings tab with aspirate flowrate, dispense flowrate, and pipette path options

# Review requests

- Does this make sense?
- styling ok
- tests?

# Risk assessment

Low, unreleased

[PLAT-221]: https://opentrons.atlassian.net/browse/PLAT-221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/QuickTransferFlow/utils/getFlowRateRange.ts
```diff
@@ -0,0 +1,28 @@
...
+  let pipetteName = state.pipette.model
+  if (state.pipette.channels === 1) {
+    pipetteName = pipetteName + `_single_flex`
+  } else if (state.pipette.channels === 8) {
+    pipetteName = pipetteName + `_multi_flex`
+  } else {
+    pipetteName = pipetteName + `_96`
+  }
+
+  const spec = getPipetteNameSpecs(pipetteName as PipetteName)
```
brenthagen: is there a way to do this without string manipulation to get the pipette name? we have a `getPipetteModelSpecs` helper that might work

app/src/organisms/QuickTransferFlow/AdvancedSettings/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  QuickTransferSummaryAction,
+  QuickTransferSummaryState,
+} from '../types'
+
+interface AdvancedSettingsProps {
+  state: QuickTransferSummaryState
+  dispatch: React.Dispatch<QuickTransferSummaryAction>
+}
+
+export function AdvancedSettings(
```
brenthagen: could consider renaming this file/function `QuickTransferAdvancedSettings` or similar simply because we have several other components and pages with the same name

app/src/organisms/QuickTransferFlow/AdvancedSettings/BaseSettings.tsx
```diff
@@ -0,0 +1,134 @@
...
+                    <Icon name="more" size={SIZE_2} />
+                  ) : null}
+                </Flex>
+              </Flex>
+            </ListItem>
+          ))
+        : null}
+      {selectedSetting === 'aspirate_flow_rate' ? (
+        <FlowRateEntry
+          kind={'aspirate'}
```
brenthagen: style nit, prefer `kind="aspirate"`

app/src/organisms/QuickTransferFlow/AdvancedSettings/FlowRate.tsx
```diff
@@ -0,0 +1,130 @@
...
+  kind: FlowRateKind
+}
+
+export function FlowRateEntry(props: FlowRateEntryProps): JSX.Element {
+  const { onBack, state, dispatch, kind } = props
+  const { i18n, t } = useTranslation(['quick_transfer', 'shared'])
+  const keyboardRef = React.useRef(null)
+
+  const [flowRate, setFlowRate] = React.useState<string>(
+    state.aspirateFlowRate ? state.aspirateFlowRate.toString() : ''
```
brenthagen: this needs to default to `dispenseFlowRate` when setting dispense flow rate, right now the input defaults to aspirate flow rate in all cases

brenthagen: also should do an explicit null check `state.aspirateFlowRate != null`

brenthagen (commented):
looks good! main comment is the dispense flow rate default

app/src/organisms/QuickTransferFlow/utils/getFlowRateRange.ts
```diff
@@ -0,0 +1,28 @@
...
+  let defaultFlowRate
+  if (kind === 'aspirate') defaultFlowRate = spec?.defaultAspirateFlowRate
+  else if (kind === 'dispense') defaultFlowRate = spec?.defaultDispenseFlowRate
+  else if (kind === 'blowout') defaultFlowRate = spec?.defaultBlowOutFlowRate
+  const min: number =
+    defaultFlowRate?.min !== undefined ? defaultFlowRate.min : 0
```
brenthagen: oh and we typically nullish check instead of checking for undefined: `defaultFlowRate?.min != null`

app/src/organisms/QuickTransferFlow/AdvancedSettings/FlowRate.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { getFlowRateRange } from '../utils'
+
+import type {
+  QuickTransferSummaryState,
+  QuickTransferSummaryAction,
+  FlowRateKind,
+} from '../types'
+
+import { ACTIONS } from '../constants'
+import { createPortal } from 'react-dom'
```
jerader: can this import go above `opentrons/components`? and the the ACTIONS import go above the types? (sorry if you already know this but we are picky and like to import things in order of packages, opentrons modules, files from the farthest away to the least farthest away from the component, types)

app/src/organisms/QuickTransferFlow/AdvancedSettings/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+  props: AdvancedSettingsProps
+): JSX.Element | null {
+  const { state, dispatch } = props
+
+  return (
+    <Flex
+      gridGap={SPACING.spacing8}
+      flexDirection={DIRECTION_COLUMN}
+      marginTop="192px"
```
jerader:
```suggestion
      marginTop="12rem"
```

app/src/organisms/QuickTransferFlow/utils/getFlowRateRange.ts
jerader: i agree with @brenthagen - grabbing the pipetteName from `getPipetteModelSpecs` is less brittle. i beleive the `getPipetteModelSpecs`'s `name` is the `pipetteName`

app/src/organisms/QuickTransferFlow/utils/getFlowRateRange.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (state.pipette.channels === 1) {
+    pipetteName = pipetteName + `_single_flex`
+  } else if (state.pipette.channels === 8) {
+    pipetteName = pipetteName + `_multi_flex`
+  } else {
+    pipetteName = pipetteName + `_96`
+  }
+
+  const spec = getPipetteNameSpecs(pipetteName as PipetteName)
+  let defaultFlowRate
```
jerader: nit: can you initialize this to 0 and type it as number maybe?

```suggestion
  let defaultFlowRate: number = 0
```

jerader (commented):
nice work! overall the code is pretty clean and looks good 🖌️ just left a few comments

app/src/organisms/QuickTransferFlow/utils/getFlowRateRange.ts
vegano1: This is not a number, it's an object like so `spec?.defaultAspirateFlowRate`
what is the best way to type this interface?

app/src/organisms/QuickTransferFlow/AdvancedSettings/FlowRate.tsx
vegano1: got catch

app/src/organisms/QuickTransferFlow/AdvancedSettings/FlowRate.tsx
vegano1: noted, thanks for pointing it out.

app/src/organisms/QuickTransferFlow/utils/getFlowRateRange.ts
jerader: oh i see, i recommend making a new type interface at the top of the file and using that. For example:

```
interface DefaultFlowRate {
    default: number
    valuesByApiLevel: Record<string, number>
}
```

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { useTranslation } from 'react-i18next'
+import {
+  Flex,
+  SPACING,
+  DIRECTION_COLUMN,
+  ALIGN_CENTER,
+  POSITION_FIXED,
+  COLORS,
+} from '@opentrons/components'
+import type { SupportedTip } from '@opentrons/shared-data'
```
jerader: can you move this import type down below the import ACTIONS?

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const tipType = getTipTypeFromTipRackDefinition(state.tipRack)
+  const flowRatesForSupportedTip: SupportedTip | undefined =
+    state.volume < 5
+      ? state.pipette.liquids.lowVolumeDefaults.supportedTips[tipType]
+      : state.pipette.liquids.default.supportedTips[tipType]
+  const minFlowRate = 0.1
+  const maxFlowRate =
+    flowRatesForSupportedTip?.uiMaxFlowRate !== undefined
+      ? flowRatesForSupportedTip?.uiMaxFlowRate
+      : Infinity
```
jerader: nit: the `in` operator is a bit cleaner imo

```suggestion
  const maxFlowRate =
    flowRatesForSupportedTip != null &&  'uiMaxFlowRate' in flowRatesForSupportedTip
      ? flowRatesForSupportedTip.uiMaxFlowRate
      : Infinity
```

jerader: or actually, maybe this is cleaner:

`const maxFlowRate = flowRatesForSupportedTip?.uiMaxFlowRate ?? Infinity`

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
```diff
@@ -0,0 +1,145 @@
...
+      ? state.dispenseFlowRate
+      : null
+  )
+
+  // use lowVolume for volumes lower than 5ml
+  const tipType = getTipTypeFromTipRackDefinition(state.tipRack)
+  const flowRatesForSupportedTip: SupportedTip | undefined =
+    state.volume < 5
+      ? state.pipette.liquids.lowVolumeDefaults.supportedTips[tipType]
+      : state.pipette.liquids.default.supportedTips[tipType]
```
jerader: i think it is `lowVolumeDefault` instead of `lowVolumeDefaults`? Additionally, we need to make sure the key exists. So maybe for extra protection we can make sure the pipette supports low volume and the key exists in the object.

```suggestion

const liquidSpecs = state.pipette.liquids

  const flowRatesForSupportedTip: SupportedTip | undefined =
    state.volume < 5 && `lowVolumeDefault` in liquidSpecs && LOW_VOLUME_PIPETTES.includes(getPipetteModelSpecs(state.pipette.model).name)
      ? liquidSpecs.lowVolumeDefault.supportedTips[tipType]
      : liquidSpecs.default.supportedTips[tipType]
```

jerader: `LOW_VOLUME_PIPETTES` is a constant that can be imported from `shared-data`!

jerader (approved):
Left a few comments about the flowRate logic but once those are addressed, this lgtm!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
smb2268: Should we use a safer alternative than `Infinity` in case this value doesn't exist? Or are we confident that it always will? If for some reason we get into a situation where the tiprack we've selected doesn't have this object populated for this pipette (which should never happen) I think putting an arbitrary value that we know are pipettes can handle would be safer.

smb2268 (approved):
This looks great, Brayan! Congrats on making your first FE PR, this was a complicated ticket to tackle.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
jerader: the value will not exist for GEN1 pipettes. I spoke with Andy sigler and Seth about this when i was refining PD flowRates and Andy said GEN1s don't have a max flow rate so infinity was ok to set.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
vegano1: This is super helpful, thank you

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/FlowRate.tsx
smb2268: Got it, that makes sense! since quick transfer is only used for Flex, only GEN3 pipettes will be relevant here. I bet if we get into a situation where we have a pipette/tip mismatch causing the flowRatesForSupportedTip object to be null, analysis would likely fail for incompatible tip anyway.

jerader: oh that is a good point, i forgot that only GEN3 pipettes will be used for this. in that case, we can use `0` instead? Either way.

PR #15541, opened 2024-06-27
fix(CI): nightly only snapshot test and no PR

y3rsh:
# Too much noise from snapshot test

> [!IMPORTANT]
> These are a very important tool but the information presentation and dev experience must be improved.

## Going back to baseline and running the test nightly but not generating a PR.

- Diffs are so big the PR is not helpful, have to do it locally
- Too many PRs being generated when happening on PR

### DIFF
--- a/.github/workflows/analyses-snapshot-test.yaml
+++ b/.github/workflows/analyses-snapshot-test.yaml
@@ -3,39 +3,30 @@ name: Analyses Snapshot Test
 on:
   workflow_dispatch:
     inputs:
-      TARGET:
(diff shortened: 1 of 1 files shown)


SyntaxColoring:
I'm not sure about this. The only trouble I've had with these snapshot tests is when one diff is covering lots of PRs at once and it's difficult to see what changes are my fault and what changes are other people's faults. I think we should keep it PR-based.

y3rsh:
Made a lot of progress and cleanup closing since #15350

PR #15542, opened 2024-06-27
feat: capture system resources

DerekMaggio:
# Overview

Closes https://opentrons.atlassian.net/browse/EXEC-370

Create a service to capture system resource usage. 

The service uses the [psutil](https://psutil.readthedocs.io/en/latest/#) library to query running processes, filters them to ones we care about, polls CPU and memory usage metrics, and stores them as csv on the robot. 

In future PRs the service will be implemented as a systemd service on the robot. Configuration will be set in the API project by setting environment variables.

# Test Plan

- [/] Unit testing
- [x] Manual testing
  - [x] Push performance_metrics to robot
  - [x] `cd /opt/opentrons-robot-server/`
  - [x] Run `python3 -m performance_metrics.system_resource_tracker` for like 30 seconds.
  - [x] Check `/data/performance_metrics_data/system_resource_data_headers` and `/data/performance_metrics_data/system_resource_data` to make sure they both have data that makes sense

# Changelog

- Add _SystemResourceTracker which provides an API into retrieving system metrics. Class is intentionally private as __main__.py should be the only thing that uses _SystemResourceTracker when running the tracker as a systemd service
- Define a journal logging config so logs can be retrieved the same way as other projects
- Add __main__.py script which runs _SystemResourceTracker as a polling service 
- Add ProcessResourceUsageSnapshot dataclass to define storage shape for captured system metrics
- Add tests

# Review requests

None

# Risk assessment

Low, in order to run any of this you have to manually push performance_metrics and trigger the service script


DerekMaggio:
Would it make sense to implement this as a systemd service running on the robot?

performance-metrics/tests/performance_metrics/test_process_resource_tracker.py
```diff
@@ -0,0 +1,36 @@
...
+def test_process_filtering():
+    tracker = ProcessResourceTracker(process_filters=["*my_script.py", "*another_script*"])
+    
+    tracker._refresh_processes()
+    filtered_processes = tracker._processes
+    
+    assert len(filtered_processes) == 3
+    assert filtered_processes[0].pid == 1
+    assert filtered_processes[1].pid == 2
+    assert filtered_processes[2].pid == 3
```
SyntaxColoring: Can this be written in terms of the public `ProcessResourceUsageSnapshot` return type so we're not coupling these tests to private implementation details of the subject?

performance-metrics/src/performance_metrics/util.py
```diff
@@ -0,0 +1,23 @@
+import typing
+import platform
+from functools import partial
+
+def format_command(cmd_list: typing.List[str]) -> str:
+    """Format the command line for the given process."""
+    return " ".join(cmd_list).strip()
```
SyntaxColoring: I don't think this matters for your purposes, but FYI this will not account for arguments that need shell quoting.

If I run `some-command "foo bar" baz`, this will format it as `some-command foo bar baz`, which is not necessarily the same thing.

performance-metrics/src/performance_metrics/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Opentrons performance metrics library."""
 
 from .robot_context_tracker import RobotContextTracker
-from .types import RobotContextState, SupportsTracking
+from .dev_types import RobotContextState, SupportsTracking
```
SyntaxColoring: This is a small thing, but some notes on module organization.

First, the `dev_types` thing is an old convention, from when we didn't have `typing_extensions` installed on the robot. The idea was that these modules could only be imported in dev environments for type-checking (hence "dev" types), not in production environments at runtime. You were supposed to guard each `dev_types` import under `if TYPE_CHECKING`. Eventually, we got `typing_extensions` installed in production environments, so `dev_types` is obsolete today (EXEC-220).

Second, this is not new to this PR, but re-exporting modules through `__init__.py` hierarchies can be [problematic](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/3997401136/Imports+and+module+structure#Don%E2%80%99t-re-export-things-from-__init__.py) and we should be careful about it in new code.

I would:

* Leave it as `types.py` instead of `dev_types.py`.
* And, either:
    * If you want callers to be able to selectively import just the types or just the `RobotContextTracker`: Remove all re-exports from this `__init__.py`.
    * Or, if you want `performance_metrics` to be monolithic: Keep the re-exports here, but rename `types.py` to `_types.py` and `robot_context_tracker.py` to `_robot_context_tracker.py`, so there is One True Way for callers to import things from `performance_metrics`.

performance-metrics/Pipfile
```diff
@@ -6,6 +6,7 @@ name        = "pypi"
 [packages]
 opentrons-shared-data = {file = "../shared-data/python", editable = true}
 performance-metrics = {file = ".", editable = true}
+psutil = "6.0.0"
```
SyntaxColoring: Is the plan to add this dependency to https://github.com/Opentrons/buildroot, or to just not run performance-metrics on OT-2s?

performance-metrics/src/performance_metrics/process_resource_tracker.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class ProcessResourceUsageSnapshot:
+    """Data for a tracked process."""
+
+    query_time: int
+    command: str
+    running_since: float
+    cpu_percent: float
```
SyntaxColoring: I feel like Linux and/or systemd probably also have accounting for, like, "how many CPU-seconds has this process consumed since it started." That might be more useful because then we're not relying on a snapshot being taken at the exact right moment to capture a spike.

SyntaxColoring (commented):
All looks good so far!

Yes, I think it makes sense to run this periodically as a systemd service. What frequency were you thinking?

performance-metrics/Pipfile
DerekMaggio: Either is fine with me. Not sure if we want performance-metrics on OT-2s or not

DerekMaggio:
> Yes, I think it makes sense to run this periodically as a systemd service. What frequency were you thinking?

I was going to try every second and see what the performance impact was. But with your suggestion of capturing the CPU seconds, it likely doesn't need to be that frequent.

performance-metrics/tests/performance_metrics/test_process_resource_tracker.py
DerekMaggio: 92f39978aa56f017be00e83eae454808d7e3f5cb

performance-metrics/src/performance_metrics/util.py
DerekMaggio: 928ad4e2f25f9e7c0c13625d161d149b1b82020f

I didn't encounter this issue. But I added a test. Let me know if I am misunderstanding

performance-metrics/src/performance_metrics/__init__.py
DerekMaggio: 57ebd57d47bacdb47e99b63de8cd301809093415

Is this what you mean?

performance-metrics/Pipfile
DerekMaggio: Posed the question to the execution team in Slack

performance-metrics/src/performance_metrics/__init__.py
SyntaxColoring: Almost. That's prefixing the submodules with underscores *and* removing the re-exports. I think you want to do just one or the other.

Option 1 is to keep the re-exports and prefix the submodules with underscores. Then, callers (`api`, `robot-server`, etc.) would do, e.g., `from performance_metrics import RobotContextTracker`. The fact that `RobotContextTracker` happens to live in its own file would be a private implementation detail of the `performance_metrics` package.

Option 2 is to remove the re-exports and not prefix the submodules with underscores. Then, callers (`api`, `robot-server`, etc.) would do, e.g., `from performance_metrics.robot_context_tracker import RobotContextTracker`. The submodule organization—the fact that `RobotContextTracker` lives in a `robot_context_tracker` submodule—*would* be part of the `performance_metrics` package's public interface.

DerekMaggio: ah I see. Thank you. Will fix

performance-metrics/src/performance_metrics/util.py
SyntaxColoring: Yeah, what I have in mind could theoretically be represented with this test case:

```python
(["some-command", "foo bar", "baz"], 'some-command "foo bar" baz')
```

SyntaxColoring: Given the way that you're using this (globbing), I think this is probably just an academic thing and not worth worrying too much about.

performance-metrics/src/performance_metrics/process_resource_tracker.py
DerekMaggio: https://github.com/Opentrons/opentrons/pull/15542/commits/aea102e84704b3f15f7c87a638650529a01523e3

performance-metrics/Pipfile
DerekMaggio: Going to ticket adding this to buildroot but make it low priority

caila-marashaj (approved):
very cool !

sfoster1 (approved):
Nice, looks good to me. making "private classes" by starting their name with an underscore is not really something we do but I don't really mind if you want to.

performance-metrics/src/performance_metrics/system_resource_tracker/__main__.py
```diff
@@ -0,0 +1,29 @@
+"""System resource tracker."""
+
+import logging
+import time
+from pathlib import Path
+from ._logging_config import log_init
+from ._system_resource_tracker import _SystemResourceTracker
```
SyntaxColoring: Nitpick, but for future reference, I think the `_SystemResourceTracker` class should not be underscore-prefixed if it's used like this.

`performance_metrics._system_resource_tracker`, the module, is private to things outside of `performance_metrics`. This is `performance_metrics.__main__`, a sibling of the private module, so it's allowed to import it. All of this makes sense so far.

But I read the `_SystemResourceTracker` class as being private to the `_system_resource_tracker` module. This is a different module so it should not be able to access it.

performance-metrics/src/performance_metrics/system_resource_tracker/__main__.py
```diff
@@ -0,0 +1,29 @@
...
+        should_track=True,
+        refresh_interval=5,
+    )
+
+    try:
+        while True:
+            tracker.get_and_store_system_data_snapshots()
+            time.sleep(tracker.refresh_interval)
+    except Exception as e:
+        logger.error("Exception occurred: %s", str(e))
```
SyntaxColoring: Instead of doing `str(e)`, do `logger.error("Exception occurred:", exc_info=True)` or, equivalently, `logger.exception("Exception occurred:")`.

`str(e)` is not a complete representation of `e` and it's frequently insufficient for debugging. For example:

```python
try:
    d = {}
    d["blah"]
except Exception as e:
    print(str(e))  # Prints "blah", not "KeyError: blah" or a line number or anything like that.
```

`logger.error(..., exc_info=True)` or `logger.exception(...)` includes the stack trace in its full glory.

SyntaxColoring (commented):
One pet peeve about logging, but this otherwise LGTM. Thank you!

skowalski08 (commented):
When I manually tested, I got:
`2024-07-03 18:57:16,258 - __main__ - INFO - Starting system resource tracker...
^C2024-07-03 18:57:33,090 - __main__ - INFO - System resource tracker is stopping.
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/opt/opentrons-robot-server/performance_metrics/system_resource_tracker/__main__.py", line 25, in <module>
    time.sleep(tracker.refresh_interval)
KeyboardInterrupt`

is the traceback expected?

SyntaxColoring:
If that happened on its own, something's definitely going wrong and we need to look into it.

If that happened when you hit Ctrl+C to stop the program, it's harmless but we should probably do something to make it less confusing. I guess the normal way to handle that is to catch `KeyboardInterrupt` at the top level.

performance-metrics/src/performance_metrics/system_resource_tracker/__main__.py
DerekMaggio: Okay. That makes total sense. 
How we underscore everything is slightly different than standard Python so I always get messed up. But the way you described it made it completely clear. I will update. Thanks Max!

DerekMaggio: https://github.com/Opentrons/opentrons/pull/15542/commits/d657d3a1bc1bf9220c9df1ff29f6fd9bc11d2dc6

DerekMaggio:
> When I manually tested, I got: `2024-07-03 18:57:16,258 - __main__ - INFO - Starting system resource tracker... ^C2024-07-03 18:57:33,090 - __main__ - INFO - System resource tracker is stopping. Traceback (most recent call last): File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main return _run_code(code, main_globals, None, File "/usr/lib/python3.10/runpy.py", line 86, in _run_code exec(code, run_globals) File "/opt/opentrons-robot-server/performance_metrics/system_resource_tracker/__main__.py", line 25, in <module> time.sleep(tracker.refresh_interval) KeyboardInterrupt`
> 
> is the traceback expected?

@skowalski08, fixed with https://github.com/Opentrons/opentrons/pull/15542/commits/89c687eaa957c1758686d6e45930569893000d1f

skowalski08 (approved):

performance-metrics/src/performance_metrics/system_resource_tracker/__main__.py
DerekMaggio: https://github.com/Opentrons/opentrons/commit/89c687eaa957c1758686d6e45930569893000d1f

PR #15543, opened 2024-06-27
fix(app): link to `api` release notes from robot settings, not `app-shell`

ecormany:
# Overview

Updates the link target in robot settings. Previously there was no way to get to the API release notes from within the app — you'd have to go directly to GitHub.

# Test Plan

- `make test-js`
- run dev app and dev robot, go to robot settings, click link

# Changelog

update URL constant in component and test.

# Risk assessment

v low


koji (approved):
thank you for fixing this!

PR #15545, opened 2024-06-27
fix(protocol-designer): append definitionid to end of liquid location…

jerader:
…s key

closes RESC-294

# Overview

This is part 2 of the duplication bug error fixed in PD 8.1.1, basically the duplicated labware from prior to 7.0.0 migration does not include the `definitionId` at the end of labware id string. 

# Test Plan

Open the attached protocol and see that the 2nd `loadLiquid` command has a labware id that is just a uuid. Upload it to the branch and export and see that the exported file's 2nd `loadLiquid` command has a labware id that includes the definition id (the `labwareDefURI`) at the end
[testingColors.json](https://github.com/user-attachments/files/16020354/testingColors.json)

# Changelog

- make sure definition id is appended to the end of labware id

# Review requests

see test plan

# Risk assessment

low

### DIFF
--- a/protocol-designer/fixtures/protocol/8/example_1_1_0MigratedToV8.json
+++ b/protocol-designer/fixtures/protocol/8/example_1_1_0MigratedToV8.json
@@ -6,7 +6,7 @@
     "author": "Author name",
     "description": "Description here",
     "created": 1560957631666,
-    "lastModified": 1719517327666,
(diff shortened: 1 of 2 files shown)


koji (approved):
tested the sandbox worked as expected.
the changes aligned with the description.

PR #15553, opened 2024-07-01
feat(api): add liquid state property

caila-marashaj:
## Overview
This adds a `liquid_presence_state` property to the protocol api's instrument context, along with an enum for the possible states.

The value here is `Optional` and defaulted to `None`, which I think makes sense for the ot2 case.

## Changelog

- add getter and setter for `liquid_presence_state`
- add `LiquidPresenceState` enum in `opentrons.types` 

## Review Requests
Is there any further error handling/prevention I should do for ot2 instruments?

### DIFF
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -111,6 +111,7 @@ class InstrumentContext(publisher.CommandPublisher):
             labware.Labware, TrashBin, WasteChute, None
         ] = trash
         self.requested_as = requested_as
+        self._liquid_presence_state: Optional[types.LiquidPresenceState] = None
 
     @property
     @requires_version(2, 0)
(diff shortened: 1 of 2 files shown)


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1642,6 +1643,16 @@ def flow_rate(self) -> "FlowRates":
         """
         return self._core.get_flow_rate()
 
+    @property
+    @requires_version(2, 0)
```
ecormany: 2.20, no?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -1642,6 +1643,16 @@ def flow_rate(self) -> "FlowRates":
...
+    def liquid_presence_state(self) -> Optional[types.LiquidPresenceState]:
+        return self._liquid_presence_state
+
+    @liquid_presence_state.setter
```
ecormany, line 1651: Would we want a public setter, or should this always be driven by sensor measurements?

ecormany (commented):
Couple comments. I may be confused as to what distinguishes this from `InstrumentContext.liquid_detection`, so ignore me if I'm off base.

api/src/opentrons/protocol_api/instrument_context.py
caila-marashaj: yep ty

caila-marashaj:
> Couple comments. I may be confused as to what distinguishes this from `InstrumentContext.liquid_detection`, so ignore me if I'm off base.

`InstrumentContext.liquid_detection` is a bool that would enable liquid probing before an aspirate, and this new property will hold the current state of liquid presence in a well. So after a liquid probe, we would update this new state to be `PRESENT`, `ABSENT`, or `UNKNOWN`. 

But you raise a good point in that now that we're adding to this it would be good to update that name to `enable_liquid_detection` or something.

caila-marashaj:
whoopsie duplicate ticket

PR #15555, opened 2024-07-01
feat(api): create instr ctx methods for checking liquid presence

aaron-kulkarni:
Added detect_liquid and require_liquid to instrument_context. The two methods both check for liquid in a specific well, but one returns true/false and the other raises exceptions.

Also adds measure_liquid_level to instrument context. This method probes the well to find the height of the liquid.
close EXEC-544, EXEC-545, EXEC-546

# Overview

# Test Plan

# Changelog

# Review requests

In find_liquid_level in the instrument core, I couldn't find a method that would execute the command with recovery and also return the result of the command. ~~So I settled for execute_command_without_recovery but I'm not sure if that's the right way to go about it.~~ I created a new method called execute_command_with_result to get around this.

# Risk assessment


aaron-kulkarni:
still need to test

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2047,49 @@ def configure_nozzle_layout(
...
         self._tip_racks = tip_racks or []
+
+    @requires_version(2, 20)
+    def detect_liquid(self, well: labware.Well) -> bool:
+        """Check if there is liquid in a well.
+
+        :returns: A boolean.
+        """
+        if well is None:
+            raise WellDoesNotExistError
```
ryanthecoder: You're returning the TYPE WellDoesNotExistError here, not an instance. you should call it like
`raise WellDoesNotExistError()`

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2047,49 @@ def configure_nozzle_layout(
...
+
+        :returns: A float representing the height of the liquid.
+        """
+        if well is None:
+            raise WellDoesNotExistError
+
+        try:
+            height = self._core.find_liquid_level(well._core)
+            return float(height)
+        except Exception:
```
ryanthecoder: I think we should only explicity catch LiquidNotFoundException if we're going to handle them. If there is some other issue that results in an error this could let that go undetected

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -843,3 +844,27 @@ def retract(self) -> None:
...
+            cmd.LiquidProbeParams(
+                labwareId=labware_id,
+                wellName=well_name,
+                wellLocation=well_location,
+                pipetteId=self.pipette_id,
+            )
+        )
+        if (
+            result is None
+        ):  # this should probably only happen in testing with mock components
```
sfoster1: No, we shouldn't do this. If we're using mock components, those mock components should return data in the same way the real things do even if they get it from somewhere else - this is why we have mocks.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+            return False
+
+    @requires_version(2, 20)
+    def require_liquid(self, well: labware.Well) -> None:
+        """If there is no liquid in a well, raise an error.
+
+        :returns: None.
+        """
+        if well is None:
+            raise WellDoesNotExistError("Well type was none.")
```
sfoster1:
```suggestion
            raise WellDoesNotExistError("You must provide a well to check.")
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+        except PipetteLiquidNotFoundError:
+            return False
+
+    @requires_version(2, 20)
+    def require_liquid(self, well: labware.Well) -> None:
+        """If there is no liquid in a well, raise an error.
+
+        :returns: None.
+        """
+        if well is None:
```
sfoster1:
```suggestion
        if not isinstance(well, labware.Well):
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+        except PipetteLiquidNotFoundError as e:
+            raise e
+
+    @requires_version(2, 20)
+    def get_liquid_height(self, well: labware.Well) -> float:
+        """Check the height of the liquid within a well.
+
+        :returns: A float representing the height of the liquid.
+        """
+        if well is None:
```
sfoster1:
```suggestion
        if not isinstance(well, labware.Well):
```

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+            raise WellDoesNotExistError()
```
sfoster1: this error should take the same form as the same check in the other methods

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+
+        try:
+            height = self._core.find_liquid_level(well._core)
+            return float(height)
+        except PipetteLiquidNotFoundError:
+            return 0
```
sfoster1: is this really an appropriate return value? Remember, this function returns a height in absolute deck coordinates, in which z=0 is the level of the deck. So this function can sometimes return, without indicating an error, that the liquid level is at the height of the deck. That doesn't make much sense to me.

I think this function should just let the `PipetteLiquidNotFoundError` through.

sfoster1 (changes requested):
Couple changes we should make to the API. Also, if you merge edge into this PR it should get rid of the snapshot problem.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -843,3 +844,27 @@ def retract(self) -> None:
...
         z_axis = self._engine_client.state.pipettes.get_z_axis(self._pipette_id)
         self._engine_client.execute_command(cmd.HomeParams(axes=[z_axis]))
+
+    def find_liquid_level(self, well_core: WellCore) -> float:
+        labware_id = well_core.labware_id
+        well_name = well_core.get_name()
+        well_location = WellLocation(
+            origin=WellOrigin.TOP, offset=WellOffset(x=0, y=0, z=0)
+        )
+        result = self._engine_client.execute_command_without_recovery(
```
sfoster1:
```suggestion
        result = self._engine_client.execute_command(
```
we definitely want to recover from this command

sfoster1 (dismissed):

api/src/opentrons/protocol_api/instrument_context.py
ryanthecoder: I agree we should pass this through. The only api call that shouldn't raise errors is detect_liquid

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+        try:
+            height = self._core.find_liquid_level(well._core)
+            if height > 0:
+                return True
+            return False  # it should never get here
+        except PipetteLiquidNotFoundError:
+            return False
+
+    @requires_version(2, 20)
+    def require_liquid(self, well: labware.Well) -> None:
```
SyntaxColoring: The name `require_liquid` matches what the ticket specifies, but is the ticket correct? [The proposal](https://opentrons.atlassian.net/wiki/spaces/PER/pages/4210491403/Proposal+Liquid+Presence+Verification+Syntax+and+Semantics) says `require_liquid_presence`.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
         )
         # TODO (spp, 2023-12-05): verify that tipracks are on adapters for only full 96 channel config
         self._tip_racks = tip_racks or []
+
+    @requires_version(2, 20)
+    def detect_liquid(self, well: labware.Well) -> bool:
```
SyntaxColoring: The name `detect_liquid` matches what the ticket specifies, but is the ticket correct? [The proposal](https://opentrons.atlassian.net/wiki/spaces/PER/pages/4210491403/Proposal+Liquid+Presence+Verification+Syntax+and+Semantics) says `detect_liquid_presence`.

api/src/opentrons/protocol_api/core/engine/instrument.py
SyntaxColoring: I think the problem here is that one of this method's callers, `InstrumentContext.get_liquid_height()`, is returning the probed height. And if the `liquidProbe` command fails and we let the frontend drive error recovery, there *isn't* a probed height in the general case. The client can choose to skip to the next command without retrying the failed `liquidProbe`.

We can certainly solve this for *`InstrumentContext.require_liquid()`*. Just, yeah, call `execute_command()` (with recovery) and discard the `liquidProbe` result. A successful `liquidProbe`, or a failed-but-recovered `liquidProbe`, will return from `execute_command()`. A failed-and-not-recovered `liquidProbe` will raise from `execute_command()`, and we should let that propagate up.

But for `InstrumentContext.get_liquid_height()` and `InstrumentContext.detect_liquid_presence()`, it's unclear to me what we would even want to happen.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+        try:
+            self._core.find_liquid_level(well._core)
+        except PipetteLiquidNotFoundError as e:
+            raise e
+
+    @requires_version(2, 20)
+    def get_liquid_height(self, well: labware.Well) -> float:
+        """Check the height of the liquid within a well.
+
+        :returns: A float representing the height of the liquid.
```
SyntaxColoring: A guideline on documentation in general:

We don't need to say that this returns a float, because that should be self-evident from the method signature. (Our documentation generator automatically includes method signatures, e.g. [like this](https://docs.opentrons.com/v2/new_protocol_api.html#opentrons.protocol_api.Well.diameter)).

What we *do* need to say is how to interpret the float. What units? (mm.) From what origin? (The bottom of the well, or the bottom of the deck?)

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2048,49 @@ def configure_nozzle_layout(
...
+        if well is None:
+            raise WellDoesNotExistError("Well type was none.")
+
+        try:
+            self._core.find_liquid_level(well._core)
+        except PipetteLiquidNotFoundError as e:
+            raise e
+
+    @requires_version(2, 20)
+    def get_liquid_height(self, well: labware.Well) -> float:
```
SyntaxColoring: The name `get_liquid_height` is what the ticket specifies, but I would consider naming this something like `measure_liquid_height()` or `probe_liquid_height()` to emphasize that it will *actually make the robot move*—it's not a typical getter/setter that merely returns some preexisting variable.

Ignore me if the name `get_liquid_height` has already been workshopped to death.

SyntaxColoring (commented):
Thanks, this is coming along! Agreed with the existing comments.

api/src/opentrons/protocol_api/instrument_context.py
SyntaxColoring: `raise Exception` is in fact valid Python, equivalent to `raise Exception()`, but it is discouraged so your point still stands.

api/src/opentrons/protocol_api/instrument_context.py
aaron-kulkarni: Different parts of the proposal say different things. At the top, under the "design" header, it says:

A dedicated raise-if-not-present command: pipette.require_liquid(well: Well)

But below that in the more detailed sections it does say: 
pipette.require_liquid_presence(well: Well).

It seems like require_liquid was the one-off error, though, and the correct version is require_liquid_presence. I'll change it but we might want to also edit the proposal to correct that error.

api/src/opentrons/protocol_api/instrument_context.py
aaron-kulkarni: Same as above.

api/src/opentrons/protocol_api/instrument_context.py
aaron-kulkarni: I agree that measure_liquid_height probably is better. I'll change it for now but if anyone has strong opinions I don't mind changing it back.

api/src/opentrons/protocol_api/instrument_context.py
SyntaxColoring: TY, proposal edited.

api/src/opentrons/protocol_engine/clients/sync_client.py
```diff
@@ -89,6 +90,26 @@ def execute_command_without_recovery(
...
+        result = self._transport.execute_command_wait_for_recovery(create_request)
+        if result.error is None:
+            return result.result
+        if isinstance(result.error, BaseException):  # necessary to pass lint
+            raise result.error
+        raise ProtocolCommandFailedError(
+            original_error=result.error,
+            message=f"{result.error.errorType}: {result.error.detail}",
+        )
+
```
SyntaxColoring, line 118: I think we should remove this new `execute_command_with_result()` method for now. It seems like it's contributing to confusion about the tradeoff between returning a result and allowing for error recovery.

Thankfully, we shouldn't need it for the behavior that we want to implement right now. See my other comment.

aaron-kulkarni:
I implemented the changes that we discussed in standup as I understood them. Probably forgot something or misunderstood it so if I did just let me know

Changes include:
In the bool only method, change to execute_command_without_recovery
In the bool or exception method, changed nothing
In the float or exception method, change to execute_command_without_recovery

Also edited the tests to reflect this

api/src/opentrons/protocol_api/core/engine/instrument.py
SyntaxColoring: Recap of in-person discussion. Looks like you've already done some of this by the time I'm posting this comment.

## `InstrumentContext.require_liquid_presence()`

We want this method to give the client the opportunity for error recovery, but raise an exception if that ultimately fails. That's what `SyncClient.execute_command()`, so under the hood, this should be implemented using that. If that raises an exception, we should just let it propagate.

## `InstrumentContext.detect_liquid_presence()`

We want this method to straight-up return `True`/`False` *without* giving the client the opportunity for error recovery. So, under the hood:

* We should implement it via `SyncClient.execute_command_without_recovery()`.
* If the probe detects no liquid, `execute_command_without_recovery()` will raise that as an exception. We should look for that and translate it into a `False` return.

  We can catch `ProtocolCommandFailedError`, and then check if `exception.original_error` is an instance of [`LiquidNotFoundError`](https://github.com/Opentrons/opentrons/blob/6ba97194fc65349a91a24b41b5c244c69bb8e6ed/api/src/opentrons/protocol_engine/commands/pipetting_common.py#L150). If so, we should translate it into a `False` return. Otherwise, we should re-raise the exception to let it propagate.

I think this will expose a bug where the robot will mistakenly go into error recovery mode (because "the command failed") *and* try to keep running the protocol (because we're catching the exception, translating it to `False`, and letting the protocol continue). We can take that on in a separate PR: EXEC-598.

## `InstrumentContext.measure_liquid_height()`

We want to keep this internal and experimental for now and keep our options open for the future. So:

* We should say in its docstring that it's internal and experimental, and exclude it from the public docs.opentrons.com website. [Here is an example of how to exclude it from docs.opentrons.com.](https://github.com/Opentrons/opentrons/blob/6ba97194fc65349a91a24b41b5c244c69bb8e6ed/api/src/opentrons/protocol_api/disposal_locations.py#L207-L214)
* We should implement this via `SyncClient.execute_command_without_recovery()`. So if no liquid is detected, it will raise an exception. We should just let that propagate.

The reason we want to keep this internal and experimental is that we haven't really fleshed out how to support error recovery on PAPI methods that return results. Does the UI flow need to have a step to retry the failed probe? Should we just raise some documented exception for the protocol author to catch and deal with how they please?

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2046,3 +2047,48 @@ def configure_nozzle_layout(
...
+        if not isinstance(well, labware.Well):
+            raise WellDoesNotExistError("You must provide a valid well to check.")
+
+        self._core.find_liquid_level(well_core=well._core, error_recovery=True)
+
+    @requires_version(2, 20)
+    def measure_liquid_height(self, well: labware.Well) -> float:
+        """Check the height of the liquid within a well.
+
+        :returns: The height, in mm, of the liquid from the deck.
```
SyntaxColoring, line 2086: For API usability, I think we would ideally have this return the height from the bottom of the well, instead of from the deck. That seems more relevant because it wouldn't be coupled to the labware offsets that the operator happened to choose at the beginning of the run.

This doesn't need to happen right now as long as we're marking this method internal and experimental.

ryanthecoder (approved):
lgtm

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -843,3 +845,35 @@ def retract(self) -> None:
...
+                    pipetteId=self.pipette_id,
+                )
+            )
+
+        if result is not None and isinstance(result, LiquidProbeResult):
+            return result.z_position
+        # should never get here
+        raise PipetteLiquidNotFoundError(
+            "Error while trying to find liquid level.",
+        )
```
SyntaxColoring, line 879: I think this gets a lot cleaner if you split this into two methods:

1. One that does `SyncClient.execute_command()`.
2. One that does `SyncClient.execute_command_without_recovery()`.

`SyncClient` is drawing a distinction those two fundamentally modes of operation, representing a tradeoff between having a command result and allowing for error recovery, and representing that tradeoff with two different method signatures. Those map cleanly to the public `InstrumentContext` behaviors that we want to implement, if we maintain the two-method distinction through `InstrumentCore`.

PR #15559, opened 2024-07-01
ci(app-shell): update macos github runner version

shlokamin:
# Overview

Github actions using macos-11 (which we use for app backend tests) are stalled because github depcrted macos-11 at the end of June: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

This PR updates the app workflow to use the latest macos version

### DIFF
--- a/app-shell/Makefile
+++ b/app-shell/Makefile
@@ -145,8 +145,8 @@ dist-ot3: package-deps
 .PHONY: dist-macos-latest
 dist-macos-latest: dist-osx
 
-.PHONY: dist-macos-11
-dist-macos-11: dist-osx
+.PHONY: dist-macos-12
+dist-macos-12: dist-osx
 
 .PHONY: dist-ubuntu-latest
 dist-ubuntu-latest: dist-linux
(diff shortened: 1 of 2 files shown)


ncdiehl11 (approved):

app-shell/Makefile
```diff
@@ -145,8 +145,8 @@ dist-ot3: package-deps
 .PHONY: dist-macos-latest
 dist-macos-latest: dist-osx
 
-.PHONY: dist-macos-11
-dist-macos-11: dist-osx
+.PHONY: dist-macos-12
+dist-macos-12: dist-osx
```
y3rsh: os-latest currently resolves to 14
why would we pin to 12 here and what is this alias for?

I don't see the `dist-macos-11` used anywhere in the repo.

### Another thought

latest (14) is an M1 processor.  wonder if that might have any effects on the build?

shlokamin: good catch, dont think this is being used ill delete it.

the other one is for when we build the app in the app build test deploy workflow:
```
make -C app-shell dist-${{ matrix.os }}
```

this evaluates to `make -C app-shell dist-macos-latest`

sfoster1 (approved):

PR #15561, opened 2024-07-01
fix(ai-server): dependency upgrades for python 3.12.4

y3rsh:
### dependency upgrades

> moving to 3.12.4 we needed to update pydantic.

- [x] added the ability to build and run from container and to shell into the running container to inspect what is built
- [x] tested locally both inside the container and with a local run


shlokamin (approved):

opentrons-ai-server/Pipfile
```diff
@@ -7,11 +7,11 @@ name = "pypi"
 openai = "==1.25.1"
 python-dotenv = "==1.0.1"
 httpx = "==0.27.0"
-llama-index = "==0.10.24"
-pydantic = "==2.7.1"
+llama-index = "==0.10.51"
```
koji, line 10: @Elyorcv 
I feel that llama-index's updates aren't really a patch update since they have added new features.
Do you have any idea to update llama-index safely?

y3rsh: I think we need some tests to validate our llama-index usage 
probably both unit and and end to end so that we can upgrade and see if they pass.

I am going to make a ticket for this.

Am merging this change so that it deploys to staging.

Elyorcv: @koji I dont know how to do it safely, but adding tests might work for now.

PR #15564, opened 2024-07-01
refactor(hardware control): start liquid probe 2mm higher, refactor z moves

caila-marashaj:
## Overview
Quick refactor for liquid probe. This just breaks up the main function a little bit in `ot3api` to keep readability, and takes moves that aren't directly related to the liquid probing movement out of `tool_sensors` to keep that modular. 

## Changelog

- take the z axis move up at the beginning of liquid probe out of `tool_sensors`, and move it to ot3api
- instead of moving to `liquid_settings.probe_start_position`, move 2mm up from wherever the pipette is at the time of the call
- move some of the motion calculation to an outside helper function


y3rsh:
> A PR has been opened to address analyses snapshot changes. Please review the changes here: #15565

FYI this change is in edge so if you merge in edge this test will pass 😊

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2635,56 +2623,64 @@ async def liquid_probe(
...
+            # safe distance so we don't accidentally aspirate liquid if we're already close to liquid
+            safe_plunger_pos = pos._replace(z=(pos.z - 2))
+            # overlap amount we want to use between passes
+            pass_start_pos = pos._replace(z=(pos.z - 0.5))
+
+            # Prep the plunger
+            await self.move_to(checked_mount, safe_plunger_pos)
             if probe_settings.aspirate_while_sensing:
-                await self._move_to_plunger_bottom(checked_mount, rate=1.0)
+                await self._move_to_plunger_bottom(checked_mount, rate=p_prep_speed)
```
sfoster1, line 2647: do we still need this switch to flip? would it actually work if we flipped it at this point?

hardware/tests/opentrons_hardware/hardware_control/test_tool_sensors.py
```diff
@@ -230,11 +209,12 @@ def move_responder(
 
     message_send_loopback.add_responder(move_responder)
 
+    # problem is with tool_sensors somewhere
```
sfoster1: ?

sfoster1 (changes requested):
Couple little things but looking good overall

api/src/opentrons/hardware_control/ot3api.py
ryanthecoder: Should still work, don't know if we want the ability to use it in the future or not

sfoster1: can we add a TODO to remove it at some point

sfoster1 (approved):
Looks good, please make sure the eventual commit message has a more in-depth description that mentions the change to specifying moves by the plunger distance though

PR #15568, opened 2024-07-02
refactor(app): Fill in error type placeholders

SyntaxColoring:
# Overview

The frontend has some code to decide, based on the failed command and its error, which error recovery flow to run. This updates that code to follow the interface that the server will use in the upcoming release.

Closes EXEC-594.

# Test plan

* [x] Induce a run error and make sure it still starts the correct recovery flow.

# Changelog

* Unify `OVERPRESSURE_WHILE_ASPIRATING` and `OVERPRESSURE_WHILE_DISPENSING`. My plan is for the server to use `errorType: 'overpressure'` for both, and the client will use `commandType` to disambiguate. This PR makes the client do that. The client currently just handles `aspirate` and `dispense`—see EXEC-593 for `aspirateInPlace` and `dispenseInPlace`.
* Fill in the `NO_LIQUID_DETECTED` placeholder.
* Remove, without filling in, the `PIPETTE_COLLISION` placeholder. [The server will not emit this error type in the upcoming release](https://opentrons.slack.com/archives/C06M9F26YGH/p1719863593167419?thread_ts=1719429843.314929&cid=C06M9F26YGH), and there is not an HTTP API for it yet.
* Some other minor adjustments—see inline notes.

# Review requests

See inline notes.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/protocol_engine/commands/liquid_probe.py
+++ b/api/src/opentrons/protocol_engine/commands/liquid_probe.py
@@ -43,6 +43,7 @@ class LiquidProbeResult(DestinationPositionResult):
     z_position: float = Field(
         ..., description="The Z coordinate, in mm, of the found liquid in deck space."
     )
+    # New fields should use camelCase. z_position is snake_case for historical reasons.
 
 
 _ExecuteReturn = Union[
--- a/api/src/opentrons/protocol_engine/commands/pipetting_common.py
+++ b/api/src/opentrons/protocol_engine/commands/pipetting_common.py
@@ -155,7 +155,7 @@ class LiquidNotFoundError(ErrorOccurrence):
 
     isDefined: bool = True
 
-    errorType: Literal["LiquidNotFound"] = "LiquidNotFound"
+    errorType: Literal["liquidNotFound"] = "liquidNotFound"
 
     errorCode: str = ErrorCodes.PIPETTE_LIQUID_NOT_FOUND.value.code
     detail: str = ErrorCodes.PIPETTE_LIQUID_NOT_FOUND.value.detail
--- a/app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
+++ b/app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
@@ -18,6 +18,7 @@ import type {
   PickUpTipRunTimeCommand,
   AspirateRunTimeCommand,
   DispenseRunTimeCommand,
+  LiquidProbeRunTimeCommand,
 } from '@opentrons/shared-data'
 import type { ErrorRecoveryFlowsProps } from '..'
 
@@ -88,8 +89,9 @@ export function useFailedLabwareUtils({
 
 type FailedCommandRelevantLabware =
   | Omit<AspirateRunTimeCommand, 'result'>
-  | Omit<PickUpTipRunTimeCommand, 'result'>
   | Omit<DispenseRunTimeCommand, 'result'>
+  | Omit<LiquidProbeRunTimeCommand, 'result'>
+  | Omit<PickUpTipRunTimeCommand, 'result'>
   | null
 
 interface RelevantFailedLabwareCmd {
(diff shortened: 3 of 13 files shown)


app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
```diff
@@ -102,11 +104,11 @@ export function getRelevantFailedLabwareCmdFrom({
   failedCommand,
   runCommands,
 }: RelevantFailedLabwareCmd): FailedCommandRelevantLabware {
-  const errorKind = getErrorKind(failedCommand?.error?.errorType)
+  const errorKind = getErrorKind(failedCommand)
 
   switch (errorKind) {
     case ERROR_KINDS.NO_LIQUID_DETECTED:
-      return failedCommand as Omit<AspirateRunTimeCommand, 'result'>
+      return failedCommand as LiquidProbeRunTimeCommand
```
SyntaxColoring, line 111: This `as` strikes me as a bit dangerous—it's coupled to the implementation of `getErrorKind()`—but I'm having trouble following the intent of this "relevant labware" stuff, so I didn't want to mess with it too much.

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -43,6 +43,7 @@ class LiquidProbeResult(DestinationPositionResult):
     z_position: float = Field(
         ..., description="The Z coordinate, in mm, of the found liquid in deck space."
     )
+    # New fields should use camelCase. z_position is snake_case for historical reasons.
```
SyntaxColoring, line 46: It's possible we could get away with just changing this to `zPosition`, but it didn't seem worth the potential confusion if it were to break run or analysis storage on one of our internal robots.

api/src/opentrons/protocol_engine/commands/pipetting_common.py
```diff
@@ -155,7 +155,7 @@ class LiquidNotFoundError(ErrorOccurrence):
 
     isDefined: bool = True
 
-    errorType: Literal["LiquidNotFound"] = "LiquidNotFound"
+    errorType: Literal["liquidNotFound"] = "liquidNotFound"
```
SyntaxColoring, line 158: `errorType` values were historically `PascalCase` because they were, verbatim, the names of Python exceptions. Now that they're part of a more deliberate HTTP API, we're using our `camelCase` [convention](https://opentrons.atlassian.net/wiki/spaces/RPDO/pages/575078506/HTTP+API+Style+Guide).

This change should be safe because when the sever parses a stored error that's old and "wrong", it should be caught by [this catch-all](https://github.com/Opentrons/opentrons/blob/0e7effd1b2b5f82cd9a482cba7e73c46fb57d868/api/src/opentrons/protocol_engine/commands/command.py#L176-L177), so it should not raise an invalid schema error.

app/src/organisms/ErrorRecoveryFlows/utils/getErrorKind.ts
```diff
@@ -1,19 +1,34 @@
...
+/**
+ * Given server-side information about a failed command,
+ * decide which UI flow to present to recover from it.
+ */
+export function getErrorKind(failedCommand: RunTimeCommand | null): ErrorKind {
+  const commandType = failedCommand?.commandType
+  const errorIsDefined = failedCommand?.error?.isDefined ?? false
+  const errorType = failedCommand?.error?.errorType
+
+  if (errorIsDefined) {
```
SyntaxColoring, line 15: Always checking `isDefined`—not just `errorType`—is the Technically Correct thing to do, given [the way that we've designed this API](https://github.com/Opentrons/opentrons/blame/79dabd304486ea5ce0ceadc159852e43a338984c/api/src/opentrons/protocol_engine/errors/error_occurrence.py#L74-L92). This is pedantic, but it's probably a good habit for us to get into now before it becomes a pain to fix.

app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
mjhuff: Note that most of this code (actually pretty much all of it) is irrelevant once the server returns the "relevant labware" (see the comments on what this means near the top of the file). It is confusing, agreed, but that's because it is complicated for the client to do this, and it's is a temp workaround until those server changes are made. The intention was just to get something working for now.

mjhuff (approved):
The FE changes are correct, and the backend changes seem reasonable. As long as the test plan works as expected, this LGTM!

app/src/organisms/ErrorRecoveryFlows/utils/getErrorKind.ts
mjhuff: Thanks for the refactor here!

TamarZanzouri (approved):
Nice work! LGTM!

app/src/organisms/ErrorRecoveryFlows/hooks/useFailedLabwareUtils.ts
SyntaxColoring: Aha, thanks for the pointer to that comment.

PR #15570, opened 2024-07-02
refactor(app): add and use deck map component in interventionmodal

sfoster1:
Adds DeckMapContent to intervention modal. This is a small wrapper around one of two different kinds of deck map:
- A labware-rendering map that is also capable of rendering highlights around some labware, used in some selection screens in error recovery
- A deck config style block map that supports clicking on some of the blocks to select them, used in some selection screens in the drop tip wizard (and thus in error recovery)

Stories for the deckmap are [on storybook](https://s3-us-west-2.amazonaws.com/opentrons-components/exec-501-deckmap-content/index.html?path=/docs/app-molecules-interventionmodal-deckmapcontent--docs), where it's rendered side-by-side with the standin since it is only destined to be used in a two-column layout.

Extra fun changes:
- Use the ODD text size args for _all_ slot labels in `BaseDeck`, because otherwise they are completely unreadable once the deckmap gets to small. I think this looks a lot better everywhere, and is actually readable when the deckmap is small, so let's go with it; may want to come back and make these something more specific on desktop
- Error recovery has a `RecoveryMap`, a large data structure that is core to the wizard flow and defines how users move between steps and screens. Error recovery also had a `RecoveryMap`, which was a component for rendering a deckmap, and a `useRecoveryMapUtils`, a hook for getting that component's arguments. Now it uses the `DeckMapContent` component above directly, and `useDeckMapUtils`, a hook for getting that component's arguments
- There's something about the deck config style map that isn't quite right and I think that it's because of this: https://opentrons.atlassian.net/browse/EXEC-513

Review requests:
- i'm pretty sure what I did to drop tip will work but it might be reasonably called somewhat gross. this is all somewhat gross though, so c'est la vie
- the deckmap slot labels do not render right in firefox at these sizes. if you're looking at storybook, you're going to have to use chrome
- fix an issue that was preventing ER wizard from being displayed on ODD
Testing, todo:
- [ ] error recovery
- [ ] drop tip in error recovery
- [ ] drop tip outside of error recovery

Closes EXEC-501


app/src/molecules/InterventionModal/DeckMapContent.stories.tsx
```diff
@@ -0,0 +1,157 @@
...
+    labwareLocation: { slotName: 'C2' },
+    definition: fixture96Plate as LabwareDefinition2,
+  },
+  {
+    labwareLocation: { slotName: 'C3' },
+    definition: fixtureTiprack1000ul as LabwareDefinition2,
+  },
+]
+
+const CONSOLE_LOG_ON_SELECT = (location: ModuleLocation): void => {
```
TamarZanzouri, line 61: nice!

app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
```diff
@@ -44,7 +44,7 @@ export function useRunPausedSplash(
   showERWizard: boolean
 ): boolean {
   // Don't show the splash when desktop ER wizard is active.
-  return !(!isOnDevice && showERWizard)
+  return isOnDevice && !showERWizard
```
TamarZanzouri, line 47: Thank you!

app/src/organisms/ErrorRecoveryFlows/hooks/useDeckMapUtils.ts
```diff
@@ -21,29 +19,33 @@ import type {
...
 
-export interface UseRecoveryMapUtilsResult {
+export interface UseDeckMapUtilsResult {
   deckConfig: CutoutConfigProtocolSpec[]
-  runCurrentModules: RunCurrentModulesOnDeck[]
-  runCurrentLabware: RunCurrentLabwareOnDeck[]
+  modulesOnDeck: RunCurrentModulesOnDeck[]
+  labwareOnDeck: RunCurrentLabwareOnDeck[]
+  highlightLabwareEventuallyIn: string[]
+  kind: 'intervention'
```
TamarZanzouri, line 39: just want to make sure we always want the kind to be intervention?

TamarZanzouri (approved):
code changes look really good! so much cleaner! I posted a question but besides that looks great

app/src/organisms/ErrorRecoveryFlows/hooks/useDeckMapUtils.ts
sfoster1: yeah. in this case we do because this is only picked up by error recovery use cases which always use that style of deck map. the deck config style is only used by the drop tip wiz

PR #15574, opened 2024-07-03
docs(api): fix stringified runtime parameter value

ecormany:
# Overview

RTP value was inside quotes. That won't work.

This is a docs hotfix. Once approved and tests pass, the commit from this branch will be tagged, docs deployed, and then this branch will be regular-merged to `edge`.

# Test Plan

[Sandbox](http://sandbox.docs.opentrons.com/docs-hotfix-stringified-rtp/v2/parameters/using_values.html#limitations) – make sure no references to 2.20 here

### DIFF
--- a/api/docs/v2/parameters/using_values.rst
+++ b/api/docs/v2/parameters/using_values.rst
@@ -88,7 +88,7 @@ An example of such an action is applying labware offset data. Say you have a par
 
     # within run()
     plate = protocol.load_labware(
-        load_name="protocol.params.plate_type", location="D2"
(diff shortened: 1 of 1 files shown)


ncdiehl11 (commented):
🚀

ncdiehl11 (approved):

y3rsh (approved):
Please ignore the analyses tests.  They are triggered because of where the changed file lives but I think because you don't have latest edge in your branch, the newly located and refactored tests cause the test to fail at an unexpected step and the comment gets added and should not.

PR #15576, opened 2024-07-03
fix(ci): analyses snapshot test trigger and step if logic improvements

y3rsh:
### Update analyses snapshot workflow so that:

- [x] doc or test changes alone do not trigger the test
- [x] only run the snapshot-update if the tests actually ran and failed
- [x] only run the PR creator if the snapshot update actually ran successfully
- [x] only comment on the PR if the PR creator was successful
- [x] update to v6 of PR creator


SyntaxColoring (approved):
TY!

PR #15578, opened 2024-07-03
refactor(api): better names for liquid probe commands

aaron-kulkarni:
# Overview

Edit detect_liquid_presence, require_liquid_presence, and measure_liquid_height as per the standup today.

Change instrument_context tests to check for the right error(LiquidNotFoundError)

Split find_liquid_level into two functions: liquid_probe_with_recovery and liquid_probe_without_recovery

# Test Plan

Added unit tests for liquid_probe_with_recovery and liquid_probe_without recovery to test_instrument_core.py
Added unit tests for detect_liquid_presence, require_liquid_presence, and measure_liquid_height to test_instrument_context.py

After the PR is merged, these API calls will be tested on the robot.

# Changelog

# Review requests

Accidentally merged #15555 too soon, this is a continuation of that PR.

# Risk assessment

Low: Since all the work was done in newly created functions, the risk that it breaks existing code is minimal.
However, the new API calls still do need to be tested on a robot before they can be used in other parts of the system.

### DIFF
--- a/api/src/opentrons/protocol_api/core/engine/instrument.py
+++ b/api/src/opentrons/protocol_api/core/engine/instrument.py
@@ -31,8 +31,6 @@ from opentrons.protocol_engine.types import (
 from opentrons.protocol_engine.errors.exceptions import TipNotAttachedError
 from opentrons.protocol_engine.clients import SyncClient as EngineClient
 from opentrons.protocols.api_support.definitions import MAX_SUPPORTED_VERSION
-
-from opentrons_shared_data.errors.exceptions import PipetteLiquidNotFoundError
 from opentrons_shared_data.pipette.dev_types import PipetteNameType
 from opentrons.protocol_api._nozzle_layout import NozzleLayout
 from opentrons.hardware_control.nozzle_manager import NozzleConfigurationType
@@ -880,8 +878,3 @@ class InstrumentCore(AbstractInstrument[WellCore]):
 
         if result is not None and isinstance(result, LiquidProbeResult):
             return result.z_position
-        # should never get here
-        print("I got here")
-        raise PipetteLiquidNotFoundError(
-            "Error while trying to find liquid level.",
-        )
--- a/api/src/opentrons/protocol_api/core/engine/instrument.py
+++ b/api/src/opentrons/protocol_api/core/engine/instrument.py
@@ -846,30 +846,37 @@ class InstrumentCore(AbstractInstrument[WellCore]):
         z_axis = self._engine_client.state.pipettes.get_z_axis(self._pipette_id)
         self._engine_client.execute_command(cmd.HomeParams(axes=[z_axis]))
 
-    def find_liquid_level(self, well_core: WellCore, error_recovery: bool) -> float:
+    def find_liquid_presence_with_recovery(self, well_core: WellCore) -> None:
         labware_id = well_core.labware_id
         well_name = well_core.get_name()
         well_location = WellLocation(
             origin=WellOrigin.TOP, offset=WellOffset(x=0, y=0, z=0)
         )
-        if error_recovery:
-            result = self._engine_client.execute_command_with_result(
-                cmd.LiquidProbeParams(
-                    labwareId=labware_id,
-                    wellName=well_name,
-                    wellLocation=well_location,
-                    pipetteId=self.pipette_id,
-                )
+
+        self._engine_client.execute_command(
+            cmd.LiquidProbeParams(
+                labwareId=labware_id,
+                wellName=well_name,
+                wellLocation=well_location,
+                pipetteId=self.pipette_id,
             )
-        else:
-            result = self._engine_client.execute_command_without_recovery(
-                cmd.LiquidProbeParams(
-                    labwareId=labware_id,
-                    wellName=well_name,
-                    wellLocation=well_location,
-                    pipetteId=self.pipette_id,
-                )
+        )
+
+    def find_liquid_level_without_recovery(self, well_core: WellCore) -> float:
+        labware_id = well_core.labware_id
+        well_name = well_core.get_name()
+        well_location = WellLocation(
+            origin=WellOrigin.TOP, offset=WellOffset(x=0, y=0, z=0)
+        )
+
+        result = self._engine_client.execute_command_without_recovery(
+            cmd.LiquidProbeParams(
+                labwareId=labware_id,
+                wellName=well_name,
+                wellLocation=well_location,
+                pipetteId=self.pipette_id,
             )
+        )
 
         if result is not None and isinstance(result, LiquidProbeResult):
             return result.z_position
--- a/api/src/opentrons/protocol_api/core/engine/instrument.py
+++ b/api/src/opentrons/protocol_api/core/engine/instrument.py
@@ -881,6 +881,7 @@ class InstrumentCore(AbstractInstrument[WellCore]):
         if result is not None and isinstance(result, LiquidProbeResult):
             return result.z_position
         # should never get here
+        print("I got here")
         raise PipetteLiquidNotFoundError(
             "Error while trying to find liquid level.",
         )
--- a/api/src/opentrons/protocol_api/core/engine/instrument.py
+++ b/api/src/opentrons/protocol_api/core/engine/instrument.py
@@ -846,7 +846,7 @@ class InstrumentCore(AbstractInstrument[WellCore]):
         z_axis = self._engine_client.state.pipettes.get_z_axis(self._pipette_id)
         self._engine_client.execute_command(cmd.HomeParams(axes=[z_axis]))
 
-    def find_liquid_presence_with_recovery(self, well_core: WellCore) -> None:
+    def liquid_probe_with_recovery(self, well_core: WellCore) -> None:
         labware_id = well_core.labware_id
         well_name = well_core.get_name()
         well_location = WellLocation(
@@ -862,7 +862,7 @@ class InstrumentCore(AbstractInstrument[WellCore]):
             )
         )
 
-    def find_liquid_level_without_recovery(self, well_core: WellCore) -> float:
+    def liquid_probe_without_recovery(self, well_core: WellCore) -> float:
         labware_id = well_core.labware_id
         well_name = well_core.get_name()
         well_location = WellLocation(
--- a/api/src/opentrons/protocol_api/instrument_context.py
+++ b/api/src/opentrons/protocol_api/instrument_context.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 import logging
 from contextlib import ExitStack
 from typing import Any, List, Optional, Sequence, Union, cast, Dict
+from opentrons.protocol_engine.errors.error_occurrence import ProtocolCommandFailedError
 from opentrons_shared_data.errors.exceptions import (
     CommandPreconditionViolated,
     CommandParameterLimitViolated,
@@ -2057,14 +2058,14 @@ class InstrumentContext(publisher.CommandPublisher):
         if not isinstance(well, labware.Well):
             raise WellDoesNotExistError("You must provide a valid well to check.")
         try:
-            height = self._core.find_liquid_level(
-                well_core=well._core, error_recovery=False
-            )
-            if height > 0:
-                return True
-            return False  # it should never get here
+            self._core.find_liquid_level_without_recovery(well._core)
+            return True
         except PipetteLiquidNotFoundError:
             return False
+        except ProtocolCommandFailedError as e:
+            if isinstance(e.original_error, PipetteLiquidNotFoundError):
+                return False
+            raise e
         except Exception as e:
             raise e
 
@@ -2077,18 +2078,20 @@ class InstrumentContext(publisher.CommandPublisher):
         if not isinstance(well, labware.Well):
             raise WellDoesNotExistError("You must provide a valid well to check.")
 
-        self._core.find_liquid_level(well_core=well._core, error_recovery=True)
+        self._core.find_liquid_presence_with_recovery(well._core)
 
     @requires_version(2, 20)
     def measure_liquid_height(self, well: labware.Well) -> float:
         """Check the height of the liquid within a well.
 
         :returns: The height, in mm, of the liquid from the deck.
+
+        :meta private:
+
+        This is intended for Opentrons internal use only and is not a guaranteed API.
         """
         if not isinstance(well, labware.Well):
             raise WellDoesNotExistError("You must provide a valid well to check.")
 
-        height = self._core.find_liquid_level(
-            well_core=well._core, error_recovery=False
-        )
+        height = self._core.find_liquid_level_without_recovery(well._core)
         return float(height)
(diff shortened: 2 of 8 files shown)


api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2057,14 +2058,14 @@ def detect_liquid_presence(self, well: labware.Well) -> bool:
...
-            if height > 0:
-                return True
-            return False  # it should never get here
+            self._core.find_liquid_level_without_recovery(well._core)
+            return True
         except PipetteLiquidNotFoundError:
             return False
+        except ProtocolCommandFailedError as e:
+            if isinstance(e.original_error, PipetteLiquidNotFoundError):
+                return False
```
SyntaxColoring: This `isinstance()` check needs to be against [`opentrons.protocol_engine.commands.pipetting_common.LiquidNotFoundError`](https://github.com/Opentrons/opentrons/blob/34d92b78bf3cb2ac4961853fe54ad222e651c316/api/src/opentrons/protocol_engine/commands/pipetting_common.py#L150), not `opentrons_shared_data.errors.exceptions.PipetteLiquidNotFoundError`. The latter is a lower-level thing that gets wrapped into the former by the time it gets here.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2057,14 +2058,14 @@ def detect_liquid_presence(self, well: labware.Well) -> bool:
...
-            height = self._core.find_liquid_level(
-                well_core=well._core, error_recovery=False
-            )
-            if height > 0:
-                return True
-            return False  # it should never get here
+            self._core.find_liquid_level_without_recovery(well._core)
+            return True
         except PipetteLiquidNotFoundError:
             return False
```
SyntaxColoring: I don't think this `except PipetteLiquidNotFoundError` will ever run—that's not something that `SyncClient` will currently raise. I think we can delete this `except` block, unless there's something I'm missing.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2057,14 +2058,14 @@ def detect_liquid_presence(self, well: labware.Well) -> bool:
...
+        except ProtocolCommandFailedError as e:
+            if isinstance(e.original_error, PipetteLiquidNotFoundError):
+                return False
+            raise e
         except Exception as e:
             raise e
```
SyntaxColoring: Catching an `Exception` and then re-raising it is effectively a no-op. We can delete this.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2057,14 +2058,14 @@ def detect_liquid_presence(self, well: labware.Well) -> bool:
...
             raise WellDoesNotExistError("You must provide a valid well to check.")
         try:
-            height = self._core.find_liquid_level(
-                well_core=well._core, error_recovery=False
-            )
-            if height > 0:
-                return True
-            return False  # it should never get here
+            self._core.find_liquid_level_without_recovery(well._core)
+            return True
```
SyntaxColoring: Instead of:

```python
try:
    do_something()
    return True
except SomeError:
    return False
```

It can sometimes be clearer to write:

```python
try:
    do_something()
except SomeError:
    return False
else:
    return True
```

Because it tightens the `try` block and puts both returns in parallel branches of the control flow tree, if that makes sense.

100% a stylistic thing, totally up to you.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -846,30 +846,37 @@ def retract(self) -> None:
         z_axis = self._engine_client.state.pipettes.get_z_axis(self._pipette_id)
         self._engine_client.execute_command(cmd.HomeParams(axes=[z_axis]))
 
-    def find_liquid_level(self, well_core: WellCore, error_recovery: bool) -> float:
+    def find_liquid_presence_with_recovery(self, well_core: WellCore) -> None:
```
SyntaxColoring: Totally optional nitpick, but I might name these `liquid_probe_with_recovery()` and `liquid_probe_without_recovery()` to avoid introducing a new phrase, "find liquid," whose meaning might not be clear to the reader.

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2077,18 +2078,20 @@ def require_liquid_presence(self, well: labware.Well) -> None:
...
+        This is intended for Opentrons internal use only and is not a guaranteed API.
         """
         if not isinstance(well, labware.Well):
             raise WellDoesNotExistError("You must provide a valid well to check.")
 
-        height = self._core.find_liquid_level(
-            well_core=well._core, error_recovery=False
-        )
+        height = self._core.find_liquid_level_without_recovery(well._core)
         return float(height)
```
SyntaxColoring: This is a nitpick and doesn't have to do with this PR, but `height` is already a float, so I think we can remove this `float()` cast.

api/tests/opentrons/protocol_api/test_instrument_context.py
```diff
@@ -1280,7 +1280,7 @@ def test_detect_liquid_presence(
     """It should only return booleans. Not raise an exception."""
     mock_well = decoy.mock(cls=Well)
     decoy.when(
-        mock_instrument_core.find_liquid_level(mock_well._core, False)
+        mock_instrument_core.find_liquid_level_without_recovery(mock_well._core)
     ).then_raise(PipetteLiquidNotFoundError())
```
SyntaxColoring: You're running into a gotcha with our mocking library here: it's letting you set up the mock to raise a `PipetteLiquidNotFoundError()`, but that's not quite what the real code will actually raise. (See my other comment.)

I think you'll have to change this to something roughly like:

```python
.then_raise(
    # Full import paths spelled out for clarity. Probably don't actually write it this verbosely.
    opentrons.protocol_engine.errors.ProtocolCommandFailedError(
       original_error=opentrons.protocol_engine.commands.pipetting_common.LiquidNotFoundError(
            ...
        )
    )
)
```

SyntaxColoring (changes requested):
This looks great. Thanks for the quick follow-up. A few adjustments:

sfoster1 (commented):
Before we merge, let's
- Change the subject line of the PR to be an indication of what the changes _do_, rather than what the changes _are_ (for instance, `refactor(api): better names for liquid probe commands`, or `refactor(api): handle errors in liquid probe commands`)
- Since the subject line is too small to have enough information, use the description to go into more depth, even if that depth is just a bullet list about changing names and handling errors

api/tests/opentrons/protocol_api/test_instrument_context.py
aaron-kulkarni: I'm trying to implement this, and so far I have the code below:
```
errorToRaise = ProtocolCommandFailedError(
        original_error=LiquidNotFoundError,
        message=f"{LiquidNotFoundError.errorType}: {LiquidNotFoundError.detail}" 
)
decoy.when(mock_instrument_core.liquid_probe_with_recovery(mock_well._core))
subject.require_liquid_presence(mock_well)
decoy.when(
         mock_instrument_core.liquid_probe_with_recovery(mock_well._core)
).then_raise(errorToRaise)
```

But when running `make -C api lint`, I'm getting:
```
Argument "original_error" to "ProtocolCommandFailedError" has incompatible type "type[LiquidNotFoundError]"; expected "ErrorOccurrence | None"
```

Not sure how to cast from LiquidNotFoundError to ErrorOccurrence

SyntaxColoring: You're passing the class `LiquidNotFoundError` itself as the argument to `original_error`, but we want to instantiate `LiquidNotFoundError` and pass the resulting object as the argument.

The mypy error message is a bit confusing if you haven't seen it before, but that's what it means by `type[LiquidNotFoundError]`. In other words, in `x=123`, `x` is of type `int`; whereas in `x=int`, `x` is of type `type[int]`.

aaron-kulkarni: ah got it thanks

api/tests/opentrons/protocol_api/core/engine/test_instrument_core.py
```diff
@@ -1319,12 +1319,26 @@ def test_find_liquid_level(
...
+) -> None:
+    """It should raise an exception on an empty well."""
+    well_core = WellCore(
+        name="my cool well", labware_id="123abc", engine_client=mock_engine_client
+    )
     try:
-        subject.find_liquid_level(well_core=well_core, error_recovery=False)
+        subject.liquid_probe_with_recovery(well_core=well_core)
     except PipetteLiquidNotFoundError:
         assert True
```
SyntaxColoring: Oh, sorry for not noticing this in the last PR—you're trying to make sure `subject.liquid_probe_with_recovery()` raises an exception on an empty well, which is correct, but this code does not do that.

If the subject does not raise any exception, the test will erroneously pass. The potentially surprising reason is that to fail a test, an exception has to propagate out of it. If a test runs and it just doesn't `assert` anything, that does not, on its own, constitute a test failure.

The correct way to test that the subject raises an exception is with [`pytest.raises`](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions), which it looks like you're already using in some of the other tests.

Ditto for `test_liquid_probe_without_recovery()`.

api/tests/opentrons/protocol_api/test_instrument_context.py
```diff
@@ -1279,9 +1283,14 @@ def test_detect_liquid_presence(
...
+        original_error=lnfe,
+        message=f"{lnfe.errorType}: {lnfe.detail}",
+    )
     decoy.when(
-        mock_instrument_core.find_liquid_level(mock_well._core, False)
-    ).then_raise(PipetteLiquidNotFoundError())
+        mock_instrument_core.liquid_probe_without_recovery(mock_well._core)
+    ).then_raise(errorToRaise)
     result = subject.detect_liquid_presence(mock_well)
     assert isinstance(result, bool)
```
SyntaxColoring, line 1295: I think we should go a little further here and also do `assert not result`, i.e. assert that the subject returned `False`, not just any `bool`. That's how I view the work that this subject is doing: it's receiving an exception (possibly) and translating it into a `True` or `False` result, and we want to test that translation.

api/tests/opentrons/protocol_api/test_instrument_context.py
```diff
@@ -1295,13 +1304,20 @@ def test_require_liquid_presence(
...
     decoy.when(
-        mock_instrument_core.find_liquid_level(mock_well._core, True)
-    ).then_raise(PipetteLiquidNotFoundError())
-    with pytest.raises(PipetteLiquidNotFoundError):
+        mock_instrument_core.liquid_probe_with_recovery(mock_well._core)
+    ).then_raise(errorToRaise)
+    with pytest.raises(ProtocolCommandFailedError) as pcfe:
         subject.require_liquid_presence(mock_well)
+    assert pcfe.value.original_error is not None
+    assert pcfe.value.original_error.errorType == "liquidNotFound"
```
SyntaxColoring: Optional nitpick:

I think we could make this a little more concise and clear by doing `assert pcfe.value is errorToRaise` instead. We want to test that if the inner function call raises an exception, the subject raises that same exception untouched.

Ditto for `test_measure_liquid_height()`.

SyntaxColoring (changes requested):
* Here are a couple of test fixups.
* Let's try to fill in the "test plan" and "risk assessment" sections of the PR description.

  I think I overheard you and @ryanthecoder talk about writing some Python protocols to play around with it. That sounds like a good idea, and you can describe that as the test plan. Just make it clear whether you're going to do that before or after merging this.

Other than that, looks great.

P.S. Here is the bug that I'm predicting will be exposed by this, which I alluded to in the last PR. EXEC-598. (It's my fault, not yours. 🙂)

api/tests/opentrons/protocol_api/core/engine/test_instrument_core.py
aaron-kulkarni: Actually, I thought about it some more and shouldn't `liquid_probe_with_recovery` not error? Because presumably it'll get a PipetteLiquidNotFoundError and then recover from that, at which point it will not raise the error.

But definitely yes for `liquid_probe_without_recovery`. I've edited that one to use pytest.raises

SyntaxColoring: Uhh maybe. Let me think about that and get back to you.

SyntaxColoring: OK, yes, I think you're on to the right thing.

`liquid_probe_with_recovery()` should just call `SyncClient.execute()` without doing anything special to convert its result.  If `SyncClient.execute()` raises an exception, `liquid_probe_with_recovery()` should propagate it as-is.

So, the way to test that, then, is just the `decoy.verify()` that you already have. We *could* get more sophisticated with Decoy's `then.raise()` and Pytest's `pytest.raises`, but we generally don't consider that worthwhile in cases like this.

So we can get rid of the `try`/`except` blocks in `test_liquid_probe_with_recovery()` and just leave the `decoy.verify()`.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -881,6 +881,7 @@ def liquid_probe_without_recovery(self, well_core: WellCore) -> float:
         if result is not None and isinstance(result, LiquidProbeResult):
             return result.z_position
         # should never get here
+        print("I got here")
```
SyntaxColoring: Print statement left over from debugging? Let's remove this before merging.

```suggestion
```

api/tests/opentrons/protocol_api/core/engine/test_instrument_core.py
```diff
@@ -1303,10 +1303,8 @@ def test_liquid_probe_without_recovery(
     well_core = WellCore(
         name="my cool well", labware_id="123abc", engine_client=mock_engine_client
     )
-    try:
+    with pytest.raises(PipetteLiquidNotFoundError):
```
SyntaxColoring: This `pytest.raises` is covering a `raise` that we're saying should never happen (your `# should never get here` comment), and which is happening only because this test didn't set up the mock `self._engine_client.execute_command_without_recovery()` call to return a realistic result.

I think this test should:

1. Rehearse (using `decoy.when()`) `mock_engine_client.execute_command_without_recovery()` to return a realistic result. Like [this outline in the Decoy documentation](https://michael.cousins.io/decoy/usage/when/#configuring-a-stub).
2. Call `subject.liquid_probe_without_recovery()`
3. Assert that the return value matches what we expect: that the subject has pulled the proper data out of the realistic result that we fed it
4. *Not* do `decoy.verify()`, because it's redundant if we're doing `decoy.when()`, so [Decoy doesn't want us to do that](https://michael.cousins.io/decoy/usage/errors-and-warnings/#redundantverifywarning).

Sorry for the continued back-and-forth on this. I could've thought this through better in my earlier reviews. I think we can move on for now and continue working on this later.

SyntaxColoring (approved):
This looks good to me to merge once the `print()` statement is removed.

One note about `test_liquid_probe_without_recovery()` for future reference, but I think we can move on for now.

api/tests/opentrons/protocol_api/core/engine/test_instrument_core.py
aaron-kulkarni: After discussion in slack, test has been edited to cover both error and success cases. Also, the raise error at the end has been deleted since it's unnecessary.

PR #15580, opened 2024-07-03
feat(app, components): LiquidIcon component

ncdiehl11:
# Overview

creation and initial implementation of `LiquidIcon` component

# Test Plan

- upload protocol using liquids ([example](https://github.com/user-attachments/files/16090142/15dc4c.py.zip))
- select protocol on Desktop and navigate to 'Liquids' tab
- verify that new liquid icon matches designs

# Changelog

- add new LiquidIcon component with stories and tests
- implement in ProtocolDetails on Desktop

# Review requests

see Test Plan and storybook

# Risk assessment

low

### DIFF
--- a/components/src/molecules/LiquidIcon/LiquidIcon.stories.tsx
+++ b/components/src/molecules/LiquidIcon/LiquidIcon.stories.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react'
 import { Flex } from '../../primitives'
 import { SPACING } from '../../ui-style-constants'
-import { GlobalStyle } from '../../../../app/src/atoms/GlobalStyle'
 import { customViewports } from '../../../../.storybook/preview'
 import { LiquidIcon } from '.'
 import type { Meta, StoryObj } from '@storybook/react'
@@ -31,7 +30,6 @@ const meta: Meta<typeof LiquidIcon> = {
   decorators: [
     Story => (
       <Flex padding={SPACING.spacing16} width="15rem" height="5rem">
-        <GlobalStyle />
         <Story />
       </Flex>
     ),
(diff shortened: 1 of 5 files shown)


components/src/molecules/LiquidIcon/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+  width: max-content;
+  background-color: ${COLORS.white};
+  border-style: ${BORDERS.styleSolid};
+  border-width: 1px;
+  border-color: ${COLORS.grey30};
+  border-radius: ${BORDERS.borderRadius4};
+`
+
+export function LiquidIcon(props: LiquidIconProps): JSX.Element {
+  const { color, size } = props
```
koji: nit

if medium is used more than small or small is used more than medium, we can set the default size and make it optional.

components/src/molecules/LiquidIcon/LiquidIcon.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+  parameters: {
+    viewport: {
+      viewports: customViewports,
+      defaultViewport: 'desktopSmall',
+    },
+  },
+  decorators: [
+    Story => (
+      <Flex padding={SPACING.spacing16} width="15rem" height="5rem">
+        <GlobalStyle />
```
koji: I guess we may not need this since Jethary add `<GlobalStyle /> to `preview.jsx` when she worked on introducing public sans to Desktop.

```suggestion

```

koji (approved):
left a couple of comments but the code looks good to me.
thank you for adding this component!

components/src/molecules/LiquidIcon/index.tsx
ncdiehl11: Good point, maybe we can default to 'small' for now

PR #15581, opened 2024-07-03
feat(app): add aria-disabled for displaying snack bar

koji:
# Overview
introduce `aria-disabled` to display a snack bar when the ODD displays `aria-disabled=true` button.

[design]
https://www.figma.com/design/MlZjOO8aZXsw3HfTBoyw9U/Release%3A-CSV?node-id=57-27179&t=ZOByYuCNeGTl4u5a-4

close AUTH-543

# Test Plan
1. send the attached protocol to a robot-server/Flex
2. setup a protocol
3. tap confirm values 
check the ODD shows the snack bar.

# Changelog

# Review requests

# Risk assessment
medium since this PR updates SmallButton


app/src/atoms/buttons/SmallButton.tsx
```diff
@@ -133,6 +135,10 @@ export function SmallButton(props: SmallButtonProps): JSX.Element {
         .disabledBackgroundColor};
       color: ${SMALL_BUTTON_PROPS_BY_TYPE[buttonType].disabledColor};
     }
+
+    &[aria-disabled='true'] {
+      background-color: red;
+    }
```
koji: this is for demo.
if we go with `aria-disabled`, we will add actual disabled style.

app/src/atoms/buttons/SmallButton.tsx
```diff
@@ -34,6 +34,7 @@ interface SmallButtonProps extends StyleProps {
   iconName?: IconName | null
   buttonCategory?: ButtonCategory // if not specified, it will be 'default'
   disabled?: boolean
+  ariaDisabled: boolean
```
koji: this should be optional

koji:
ToDo
- [x] test screens that use `SmallButton`

app/src/organisms/ProtocolSetupParameters/index.tsx
```diff
@@ -149,18 +151,27 @@ export function ProtocolSetupParameters({
...
-      labwareOffsets,
-      runTimeParameterValues: getRunTimeParameterValuesForRun(
-        runTimeParametersOverrides
-      ),
-    })
+    if (
+      enableCsvFile &&
+      mostRecentAnalysis?.result === 'parameter-value-required'
+    ) {
+      console.log('tapped')
```
jerader:
```suggestion
```

app/src/atoms/buttons/SmallButton.tsx
```diff
@@ -34,6 +34,8 @@ interface SmallButtonProps extends StyleProps {
   iconName?: IconName | null
   buttonCategory?: ButtonCategory // if not specified, it will be 'default'
   disabled?: boolean
+  /** aria-disabled for displaying snack bar */
+  ariaDisabled?: boolean
```
jerader: can you make the comment a bit more descriptive? something like this: 


```suggestion
  /** aria-disabled for displaying snack bar, used for ODD only at this time. */
  ariaDisabled?: boolean
```

it would be good to mention that this is only for ODD components at the moment, which i know `Smallbutton` is only used for ODD but it is still good to give additional details.

app/src/atoms/buttons/SmallButton.tsx
```diff
@@ -133,18 +140,25 @@ export function SmallButton(props: SmallButtonProps): JSX.Element {
...
+        .disabledBackgroundColor};
+      color: ${SMALL_BUTTON_PROPS_BY_TYPE[buttonType].disabledColor};
+    }
   `
 
   return (
     <Btn
       css={SMALL_BUTTON_STYLE}
-      disabled={disabled}
+      disabled={ariaDisabled ? false : disabled}
```
jerader, line 154: so if i user adds `ariaDisabled` and `disabled`, only the `ariaDisabled` will work, right? Is there a way to change the props easily so you can only add one or the other? If it requires too much of a refactor then you can skip doing that.

koji: This change is to keep the existing disabled case.
I think refactoring would need some time.

jerader (approved):
thank you for researching `aria-disabled`! This looks great

PR #15560, opened 2024-07-01
refactor(app): Wire up Error Recovery flows on desktop

TamarZanzouri:
# Overview

closes https://opentrons.atlassian.net/browse/EXEC-552.
@mjhuff 's implementation :-) I just fixed the test and put up the PR. 

# Test Plan

- make sure the error recovery flow matches Figma designs and logic on desktop app.

# Risk assessment

low.

PR #15583, opened 2024-07-05
fix(app): fix a logic error that prevented recovery from running

sfoster1:
#15560 introduced a logic bug that prevented ER from launching on ODD, this fixes it.


SyntaxColoring (approved):
TY!

PR #15586, opened 2024-07-05
feat(app): CategorizedStepContent component

sfoster1:
This is a component that implements https://www.figma.com/design/8dMeu8MuPfXoORtOV6cACO/Feature%3A-Error-Recovery-August-Release?node-id=4113-73992&t=oJBTLKrTkARxEUyD-4 (https://opentrons.atlassian.net/browse/EXEC-499). It's used in error recovery.

~There are some open questions and todos around the step indices which is why this is a draft. Specifically,~
- ~What's the sizing behavior of the step numbers in the desktop version? As-implemented, they'll be at least 3 characters wide and then all increase size in lockstep to the longest one~ that's fine
- ~I'm guessing at the font on desktop for the step indices~ I guessed right, it was caption regular

~Also, I'm going to implement this in the currently-existing places that use it before undrafting.~ conveniently there was just one

# to do:
- [x] ODD layout test
- [x] make sure nothing is more broken on desktop (you can't get to this in desktop anyway)

Storybook: https://s3-us-west-2.amazonaws.com/opentrons-components/exec-499-two-step-component/index.html?path=/story/app-molecules-interventionmodal-categorizedstepcontent--categorized-step

Closes EXEC-499


app/src/molecules/Command/Command.tsx
```diff
@@ -106,31 +127,55 @@ export function CenteredCommand(
...
-            }
+            ${props.forceTwoLineClip === true
+              ? TEXT_CLIP_STYLE
+              : ODD_ONLY_TEXT_CLIP_STYLE}
             @media not (${RESPONSIVENESS.touchscreenMediaQuerySpecs}) {
               max-height: 240px;
               overflow: auto;
             } ;
           `}
+          modernStyledTextDefaults
```
TamarZanzouri, line 176: sorry, I cant find where this is coming from?

app/src/molecules/InterventionModal/CategorizedStepContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ]
+}
+
+type MappedState =
+  | {
+      command: RunTimeCommand
+      state: NonSkeletonCommandState
+      commandTextData: CommandTextData
+    }
+  | { command: null; state: 'loading'; commandTextData: null }
```
TamarZanzouri: not a big deal but can we define this and reuse it?

app/src/molecules/InterventionModal/CategorizedStepContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const commandAndState = (
+  command: CommandWithIndex | null,
+  state: NonSkeletonCommandState,
+  commandTextData: CommandTextData | null
+): MappedState =>
+  command == null || commandTextData == null
+    ? { state: 'loading', command: null, commandTextData: null }
```
TamarZanzouri: reuse here

app/src/molecules/InterventionModal/CategorizedStepContent.tsx
```diff
@@ -0,0 +1,146 @@
...
+        }
+      `}
+      gap={SPACING.spacing24}
+    >
+      <Flex
+        flexDirection={DIRECTION_COLUMN}
+        justifyContent={JUSTIFY_FLEX_START}
+        gap={SPACING.spacing4}
+      >
+        <StyledText oddStyle="bodyTextSemiBold" desktopStyle="captionSemiBold">
```
TamarZanzouri, line 85: is this an ODD component only?

app/src/organisms/ErrorRecoveryFlows/shared/TwoColTextAndFailedStepNextStep.tsx
```diff
@@ -36,33 +40,83 @@ export function TwoColTextAndFailedStepNextStep({
...
-        <TwoColumn>
-          <Flex gridGap={SPACING.spacing8} flexDirection={DIRECTION_COLUMN}>
-            <LegacyStyledText as="h4SemiBold">{leftColTitle}</LegacyStyledText>
+  const { t } = useTranslation('error_recovery')
+  const nthStepAfter = (n: number): number | undefined =>
+    stepCounts.currentStepNumber == null
+      ? undefined
+      : stepCounts.currentStepNumber + n
+  const commandsAfter = [
+    commandsAfterFailedCommand == null
```
TamarZanzouri: can we maybe split this? its very complicated and unreadable

TamarZanzouri (changes requested):
looks great! a few suggestions but not big deal if we wont do them!

app/src/molecules/InterventionModal/CategorizedStepContent.tsx
sfoster1: Nope, it's for both

app/src/molecules/Command/Command.tsx
sfoster1: that is defined here: https://github.com/Opentrons/opentrons/pull/15586/files#diff-56eb32b51bfe9e226cf8f10718147ffcb6d38ff31363a271be656ab0f4bdf239R56

If you mean "what the hell is this for and what is it doing", the answer is that it's used here: https://github.com/Opentrons/opentrons/pull/15586/files#diff-56eb32b51bfe9e226cf8f10718147ffcb6d38ff31363a271be656ab0f4bdf239R445-R475 and is fundamentally about my cowardice. The command text needs to be stylistically correct on desktop and ODD, but it's also used in a whole bunch of different places and I don't want to uniformly change it everywhere at once - this is how we display commands in the runlog, for instance. Having this awkward prop and the CommandStyledText means that we can choose whether we want the modern styles or the legacy styles. Command always wants the modern styles because Command is only used in error recovery right now (the other uses only use CommandText)

app/src/molecules/InterventionModal/CategorizedStepContent.tsx
sfoster1: ~can't reuse it here, this is a value here and above it's a type~ talked abotu it in person, defined a constant for use here and used `typeof` to set a type

app/src/molecules/InterventionModal/CategorizedStepContent.tsx
TamarZanzouri: woops missed the `desktopStyle`

TamarZanzouri (approved):
Thank you! This is great!

PR #15587, opened 2024-07-05
feat(api): Enable error recovery in JSON protocol runners

SyntaxColoring:
# Overview

Closes EXEC-402.

# Test Plan

[8x transfer.json.zip](https://github.com/user-attachments/files/16113098/8x.transfer.json.zip)

* [x] Error recovery behaves as specified by the designs when a JSON protocol encounters an error.
* [x] Cancelling a JSON protocol run in the middle of a hardware motion still works.
* [x] Cancelling a JSON protocol run in the middle of a `pause` or `delay` command still works.

# Changelog

When executing commands from JSON protocols, leave room for the commands to undergo error recovery. The implementation almost exactly mirrors [how we do it for Python protocols](https://github.com/Opentrons/opentrons/blob/432f72724354966d55fd9d7ebb36d94c97600102/api/src/opentrons/protocol_engine/clients/transports.py#L76-L99).

# Review requests

See inline comments.

# Risk assessment

Probably at least medium, if not high. This sort of code has historically been bug-prone. Also, until we complete EXEC-417, this code is going to be confusing because it's halfway through an architectural refactor.


api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -371,14 +372,32 @@ async def run(  # noqa: D102
...
-        for command in self._queued_commands:
-            result = await self._protocol_engine.add_and_execute_command(command)
-            if result and result.error:
-                raise ProtocolCommandFailedError(
-                    original_error=result.error,
-                    message=f"{result.error.errorType}: {result.error.detail}",
+    async def _add_and_execute_commands(self) -> None:
+        for command_request in self._queued_commands:
+            # todo(mm, 2024-07-05): This logic to handle the various command execution
+            # outcomes mirrors PAPI's ChildThreadTransport. Simplify or deduplicate it.
```
SyntaxColoring, line 378: I could try harder to improve this if we think it's important right now. I'm being lazy about it because I don't think we'll stick with this separation of responsibilities.

For one thing: I suspect we'll need to lift `get_error_recovery_type()` from being a `ProtocolEngine` decision to being a `RunOrchestrator`+runner decision, to help with EXEC-598.

Another thing: Right now, `ProtocolEngine` is "pulling" commands from higher-level code, on its own time, via `command_generator`. This is where the weird `CommandStatus.QUEUED` case comes from. If we simplified things so the higher-level code "pushed" commands down to `ProtocolEngine` one at a time, we could avoid that.

api/tests/opentrons/protocol_runner/test_protocol_runner.py
```diff
@@ -497,19 +501,67 @@ async def test_load_json_runner(
...
     await run_func()
     decoy.verify(
-        await protocol_engine.add_and_execute_command(
+        # todo(mm, 2024-07-08): This triggers decoy.RedundantVerifyWarning.
+        # Above, we're rehearsing each add_and_execute_command() call because the
+        # subject needs the return values. But the subject doesn't return anything
+        # based on on those return values, so here, we need to "redundantly" verify
+        # each call to make sure the subject actually did some work.
+        # It doesn't seem like Decoy wants subjects to be like this, so it's unclear
+        # how to fix this.
```
SyntaxColoring, line 553: Same deal as my other comment. I'm being lazy about these tests because I'm more concerned about improving the overall layering.

I'm actually looking at these tests and wondering if we'd lose any meaningful coverage if we just deleted them.

sfoster1 (approved):
This looks good to me, nice! I think the awkwardness from being mid-architecture-transition is responsible for whatever's ugly here and I'm happy to leave it til then.

PR #15588, opened 2024-07-05
feat(api): Enable error recovery without the feature flag

SyntaxColoring:
# Overview

Closes EXEC-547.

# Test Plan

* [x] Flex protocols still enter recovery mode when there's a tip pickup or overpressure error, and still fail for other errors.
* [x] `opentrons_simulate`, local in-app analysis, and robot-side analysis still work as expected when the protocol has an error.
* [x] OT-2 protocols still fail, without entering recovery mode, when there's an error.

# Changelog

* Stop gating error recovery behind the `enableErrorRecoveryExperiments` feature flag.

  The feature flag doesn't do anything anymore. I'm leaving it in-place in case we want to use it for the ongoing work in liquid level detection.

* For things other than real robot-server protocol runs (including analysis, `opentrons_execute`, maintenance runs, etc.), explicitly choose a policy of "never recover."

  We don't ever want to enter recovery mode for these things, because nothing could bring it *out* of recovery mode, and it would hang forever.
  
* Explicitly disable error recovery on the OT-2. None of the designs for the upcoming release make sense on an OT-2, because the OT-2 lacks the hardware feedback to detect the errors in the first place.

# Review requests

* I looked around for a frontend feature flag but I couldn't find one. Have we already removed it?
* Anyone have thoughts on what error recovery policy is appropriate for stateless commands (the "default orchestrator", in this code)?

# Risk assessment

Medium.

### DIFF
--- a/api/src/opentrons/config/advanced_settings.py
+++ b/api/src/opentrons/config/advanced_settings.py
@@ -211,11 +211,8 @@ settings = [
         title="Enable error recovery experiments",
         description=(
             "Do not enable."
-            " This is an Opentrons internal setting to experiment with"
+            " This is an Opentrons internal setting to enable additional,"
             " in-development error recovery features."
-            " This will interfere with your protocol runs,"
-            " corrupt your robot's storage,"
-            " bring misfortune and pestilence upon you and your livestock, etc."
         ),
         robot_type=[RobotTypeEnum.FLEX],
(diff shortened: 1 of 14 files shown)


sfoster1 (approved):
Code looks good to me, haven't tested though. I think `never_recover` is appropriate for stateleess command execution.

PR #15591, opened 2024-07-07
feat(robot-server): Exclude quick-transfer protocols from getting auto-deleted by the ProtocolAutoDeleter.

vegano1:
# Overview

When a Quick Transfer protocol is created the user can save it for later or execute it now, in both cases, the protocol will be uploaded to the robot via the POST /protocol endpoint. When saved, the protocol will have no run history, be marked as “unused“ and can be deleted by the “ProtocolAutoDeleter“. We should not auto-delete quick transfer protocols since their total count is limited by the “max quick transfer protocols“ robot server setting.

Closes: [PLAT-318](https://opentrons.atlassian.net/browse/PLAT-318)

# Test Plan

- [x] Manually set the "max_unused_protocols" to a smaller number of protocols to force auto-deletion
- [x] Make sure that `quick-transfer` protocols are not automatically deleted since they are "unused".

# Changelog

- add a `ProtocolAutoDeleter` instance for quick-transfer protocols to limit the number based on the `max_quick_transfer_protocols` setting.

# Review requests

- does this make sense? Is there anything I'm overlooking?

# Risk assessment

low, does not affect existing functionality.

[PLAT-318]: https://opentrons.atlassian.net/browse/PLAT-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robot-server/robot_server/protocols/protocol_auto_deleter.py
```diff
@@ -19,8 +21,24 @@ def __init__(
         self._protocol_store = protocol_store
         self._deletion_planner = deletion_planner
 
-    def make_room_for_new_protocol(self) -> None:  # noqa: D102
-        protocol_run_usage_info = self._protocol_store.get_usage_info()
+    def make_room_for_new_protocol(
+        self, exclude_kind: Optional[ProtocolKind] = None
```
sfoster1: any reason this has to be optional? i think previous work makes every protocol have a kind, right?

vegano1: The idea here is that `exclude_kind` tells the auto-deleter to ignore the kind of protocol passed in, so by default, we want to include all protocols stored when determining if there's anything we should delete. This keeps the functionality the same as today, if you want to specify a kind of protocol to ignore then you can supply the `exclude_kind` arg.

If we can do without keeping the existing behavior, then making `exclude_kind` a required arg makes sense.

robot-server/robot_server/deletion_planner.py
```diff
@@ -70,6 +70,8 @@ def plan_for_new_protocol(
         Args:
             existing_protocols: Information about all protocols that currently exist.
                 Must be in order from oldest first!
+            exclude_kind: Exclude these kids  of protocols when taking into account
```
CaseyBatten:
```suggestion
            exclude_kind: Exclude these kinds of protocols when taking into account
```

Typo comment correction.

robot-server/robot_server/protocols/protocol_auto_deleter.py
```diff
@@ -19,8 +21,24 @@ def __init__(
...
+    ) -> None:
+        """Finds unused protocols and deletes them.
+
+        arg: exclude_kind: Excludes the this ProtocolKind from being deleted
```
CaseyBatten:
```suggestion
        arg: exclude_kind: Excludes this ProtocolKind from being deleted
```

Typo correction

CaseyBatten (approved):
Looks good outside of some minor comment typos, filters the protocols appropriately. Agree that `exclude_kind` can be an optional argument so as not to interrupt current behavior expectations.

vegano1:
@sfoster1 

I added another instance of the `ProtocolAutoDeleter` to handle quick-transfer protocols, this should make things cleaner.
Let me know what you think.

sfoster1 (approved):
Absolutely love this. Really happy it ended up being this easy to use the auto deleter.

PR #15592, opened 2024-07-07
feat(robot-server): exclude quick-transfer protocol runs when making room for new runs.

vegano1:
# Overview

Quick-transfer protocol runs are short-lived, the client deletes them once the run is complete, because of this we need to make sure that quick-transfer runs don't count when auto-deleting runs.

Closes: [PLAT-319](https://opentrons.atlassian.net/browse/PLAT-319)

# Test Plan

- [x] Set the max number of runs to a low number like 5
- [x] Create 5 standard protocol runs
- [x] Create 1 quick-transfer protocol run and make sure only unused runs are deleted
- [x] Create 2 quick-transfer protocol runs and make sure one is deleted.

# Changelog

- Add another instance of the `RunAutoDeleter` to handle quick transfer run cleanup.

# Review requests

- make sure logic makes sense

# Risk assessment
Low, unreleased feature

[PLAT-319]: https://opentrons.atlassian.net/browse/PLAT-319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sfoster1 (commented):
Reading this and the equivalent for protocol runs, it feels a little bug-prone to carve out both these positive and negative exceptions for the storage managers - don't care about quick transfers here, only care about quick transfers there. 

What if we just make another instance of the data managers for quick transfers?

robot-server/robot_server/runs/engine_store.py
```diff
@@ -256,7 +256,9 @@ async def create(
             assert (
                 protocol is not None
             ), "A JSON protocol should have a protocol source file."
+            _log.warning(f"LOAD JSON: {protocol}")
```
sfoster1: this seems like a mistaken inclusion

robot-server/robot_server/runs/router/base_router.py
```diff
@@ -180,7 +190,13 @@ async def create_run(
     # TODO(mc, 2022-05-13): move inside `RunDataManager` or return data
     # to pass to `RunDataManager.create`. Right now, runs may be deleted
     # even if a new create is unable to succeed due to a conflict
-    run_auto_deleter.make_room_for_new_run()
+    run_deleter: RunAutoDeleter = run_auto_deleter
```
sfoster1, line 193: even though we only use it once, it might be worth making a utility function for this, which would let the variable be defined only once

sfoster1 (approved):
Looks good aside from that log!

robot-server/robot_server/runs/router/base_router.py
vegano1: Yeah, the naming is not the best, `run_deleter` is defined once and reassigned to either `run_auto_deleter` or `quick_transfer_run_auto_deleter`. I will add a todo comment to clean this up in an upcoming pr.

PR #15593, opened 2024-07-07
feat(app): Add Aspirate and Dispense for Quick Transfer advanced settings.

vegano1:
# Overview

Adds the Aspirate and Dispense settings to the Quick Transfer Advanced Settings tab, which lets you change the specific behavior of the quick transfer protocol for precise control of the protocol generated.

Closes: [PLAT-222](https://opentrons.atlassian.net/browse/PLAT-222) [PLAT-223](https://opentrons.atlassian.net/browse/PLAT-223)

# Test Plan

- [x] Make sure that the Quick Transfer protocol file generated reflects the advanced settings chosen when it was created.
- [x] Run the Quick Transfer protocol and make sure that the "Pipette path" option is respected based on the transfer type.
- [x] Run the Quick Transfer protocol and make sure that the "Blowout" option is respected.
- [x] Make sure that the input limits are being followed for all the various options.

# Changelog

- Refactored the Quick Transfer Advanced Settings tab so the components are on one page instead of having "BaseSettings", "AspirateSettings", and "DispenseSettings" components as this led to component rendering issues.
- Added "TipPosition", "Blowout", "FlowRate", "AirGap", "TouchTip", "PipettePath", "Mix", and "Delay" components.
- The above components are shared between aspirate and dispense settings, as they render differently based on the kind.

# Review requests

- Any styling issues?
- Does this make sense?
- Please take a look at the range mins and max as right now they are placeholders, where can I find this data?
- Please also take a look at the options and point out which options would invalidate another.

# Risk assessment

Low, unreleased feature.

# TODO:
- [x] Add min/max ranges for various options, need to find data.
- [x] Add disabling of options based on the combination of settings, I'm not sure what all the combinations are yet.
- [x] Add snackbar for disabled options, need to find data of when options are disabled

[PLAT-222]: https://opentrons.atlassian.net/browse/PLAT-222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLAT-223]: https://opentrons.atlassian.net/browse/PLAT-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TouchTip.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { NumericalKeyboard } from '../../../atoms/SoftwareKeyboard'
+
+interface TouchTipProps {
+  onBack: () => void
+  state: QuickTransferSummaryState
+  dispatch: React.Dispatch<QuickTransferSummaryAction>
+  kind: FlowRateKind
+}
+
+export function TouchTip(props: TouchTipProps): JSX.Element {
```
jerader, line 33: touch tip is only allowed if the labware does not have `touchTipDisabled` included in the labware's quirks.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/AirGap.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const setSaveOrContinueButtonText = (): string => {
+    return t(
+      selectedOption === 'Enabled' && currentStep < flowSteps.length - 1
+        ? 'shared:continue'
+        : 'shared:save'
+    )
+  }
+
+  // TODO: find the actual min and max for these values.
+  const volumeRange = { min: 1, max: 100 }
```
jerader: the max here should depend on the volume left in the tip after aspirating.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
```diff
@@ -0,0 +1,270 @@
...
+  const delayRange = { min: 1, max: 100 }
+
+  const delayError =
+    delayDuration !== null &&
+    (delayDuration < delayRange.min || delayDuration > delayRange.max)
+      ? t(`value_out_of_range`, {
+          min: delayRange.min,
+          max: delayRange.max,
+        })
+      : null
```
jerader: these errors seems like a lot of repeated logic, i recommend pulling it out into a util that each component can import and use.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Mix.tsx
```diff
@@ -0,0 +1,255 @@
...
+    return t(
+      selectedOption === 'Enabled' && currentStep < flowSteps.length - 1
+        ? 'shared:continue'
+        : 'shared:save'
+    )
+  }
+
+  // TODO: find the actual min and max for these values.
+  const minVolume = 1
+  const maxVolume = 100
```
jerader: these min and max volumes should match the tip's min and max volumes

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Mix.tsx
```diff
@@ -0,0 +1,255 @@
...
+  const minReps = 1
+  const maxReps = 100
```
jerader: do we need a max here? 😅

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+  const positionRange = { min: 1, max: 100 }
```
jerader: this range should be based off of the height of the labware i think? like we don't want the user to smash into the labware if its too low or move too high above the labware

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TipPosition.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const keyboardRef = React.useRef(null)
+
+  const [tipPosition, setTipPosition] = React.useState<number | null>(
+    kind === 'aspirate'
+      ? state.tipPositionAspirate
+      : kind === 'dispense'
+      ? state.tipPositionDispense
+      : null
+  )
+  const tipPositionRange = { min: 1, max: 2 } // TODO: set this based on range
```
jerader: range should be based off the height of the labware i think.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TouchTip.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const setSaveOrContinueButtonText = (): string => {
+    return t(
+      selectedOption === 'Enabled' && currentStep < flowSteps.length - 1
+        ? 'shared:continue'
+        : 'shared:save'
+    )
+  }
+
+  // TODO: find the actual min and max for these values.
+  const positionRange = { min: 1, max: 100 }
```
jerader: same here, should be based off of the height of the labware

jerader (commented):
i did a quick read through and it looks good so far! nice work! i also did some smoke testing (only with n:n transfers so far) and from what i can tell, the commands are generated as expected 🎉

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
jerader: thinking about this value more and digging into step-generation, I don't think delay position is actually wired up in step-generation?

jerader: whooops nevermind it totally is! my bad

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/AirGap.tsx
```diff
@@ -0,0 +1,196 @@
...
+  ]
+
+  const flowSteps: string[] = ['enable_air_gap', 'select_volume']
+
+  const handleClickBackOrExit = (): void => {
+    currentStep > 0 ? setCurrentStep(currentStep - 1) : onBack()
+  }
+
+  const handleClickSaveOrContinue = (): void => {
+    if (selectedOption === 'Enabled') {
```
smb2268: I think instead of having a string "enabled" as the option, we should have this be a boolean that maps `true` to "enabled" text and `false` to "disabled"

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/AirGap.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const [volume, setVolume] = React.useState<number | null>(
+    kind === 'aspirate'
+      ? state.airGapAspirate ?? null
+      : kind === 'dispense'
+      ? state.airGapDispense ?? null
+      : null
+  )
+
+  let headerCopy: string = ''
+  let Action:
```
smb2268: action here should be lowercase - we use camel case for all variable names and capitalize the first letter for typescript types!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/AirGap.tsx
smb2268: yep! I think you'll need to replicate some of the logic in `getVolumeRange` here, but we should only need to grab the pipette and tip capacity, and take the lesser of the two. The and then if `state.path` is single the max would be ((lesser of pipette and tip capacity)- state.volume), if it's multi we'd have to do ((lesser of pipette and tip capacity)- (state.volume * 2))

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/BlowOut.tsx
```diff
@@ -0,0 +1,158 @@
...
+  ]
+
+  const flowSteps: string[] = ['enable_blowout', 'select_location']
+
+  const handleClickBackOrExit = (): void => {
+    currentStep > 0 ? setCurrentStep(currentStep - 1) : onBack()
+  }
+
+  const handleClickSaveOrContinue = (): void => {
+    if (selectedOption === 'Enabled') {
```
smb2268: same comments as above apply throughout for making the option a boolean and Action lowercase!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/BlowOut.tsx
```diff
@@ -0,0 +1,158 @@
...
+      />
+      {flowSteps[currentStep] === 'enable_blowout' ? (
+        <Flex
+          marginTop={SPACING.spacing120}
+          flexDirection={DIRECTION_COLUMN}
+          padding={`${SPACING.spacing16} ${SPACING.spacing60} ${SPACING.spacing40} ${SPACING.spacing60}`}
+          gridGap={SPACING.spacing4}
+          width="100%"
+        >
+          {displayItems.map(option => (
```
smb2268:
```suggestion
          {displayItems.map(displayItem => (
```
would make this a little more readable so we don't have `option.option` below!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/BlowOut.tsx
```diff
@@ -0,0 +1,158 @@
...
+    {
+      option: 'Disabled',
+      value: t('option_disabled'),
+      onClick: () => {
+        setSelectedOption('option_disabled')
+      },
+    },
+  ]
+
+  const blowOutLocationItems = [
```
smb2268: We need to grab the trash bin and waste chute location items from the user's deck config -- you can see this done in the `TipDropLocation` options as a reference! That way we can display to the user a _specific_ trash bin or waste chute instead of trying to find one later on. The type for this may need to change to `source_well` | `dest_well` | CutoutConfig

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
```diff
@@ -0,0 +1,270 @@
...
+  const setSaveOrContinueButtonText = (): string => {
+    return t(
+      selectedOption === 'Enabled' && currentStep < flowSteps.length - 1
+        ? 'shared:continue'
+        : 'shared:save'
+    )
+  }
+
+  // TODO: find the actual min and max for these values.
+  const delayRange = { min: 1, max: 100 }
```
smb2268: I don't know that we need a specific range for these values, I would imagine we let them delay for however long they want.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
smb2268: Yeah it looks like the default and min is 1 mm from bottom of the well, and we can allow them to enter a value 2x the height of the well which can be grabbed from the labware def!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Mix.tsx
smb2268: I think the max here should be the lesser of the max value from `getMinVolumeRange` and `state.volume` and then min can be the same as the min from `getMinVolumeRange`

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Mix.tsx
smb2268: looks like PD does not have a max!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Mix.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+  const error =
+    // TODO: This error does not take into account min/maxReps
+    (mixVolume !== null && (mixVolume < minVolume || mixVolume > maxVolume)) ||
```
smb2268: Lets separate this out so that it's based on the currentStep -- 
if (currentStep === select_volume) error =  mixVolume !== null && (mixVolume < minVolume || mixVolume > maxVolume)

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/PipettePath.tsx
```diff
@@ -27,6 +33,16 @@ interface PipettePathProps {
...
   const { onBack, state, dispatch } = props
   const { t } = useTranslation('quick_transfer')
+  const keyboardRef = React.useRef(null)
+
+  const [selectedOption, setSelectedOption] = React.useState<PathOption>(
+    state.path
+  )
+  const [currentStep, setCurrentStep] = React.useState<number>(0)
+  const [blowOutLocation, setBlowOutLocation] = React.useState<string>(
+    state.blowOut ?? 'trashBin'
```
smb2268: Same comment applies here about grabbing a `trashBin` or `wasteChute` if there is one configured in the deck config

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/PipettePath.tsx
```diff
@@ -55,44 +67,153 @@ export function PipettePath(props: PipettePathProps): JSX.Element {
...
+  const handleClickBackOrExit = (): void => {
+    currentStep > 0 ? setCurrentStep(currentStep - 1) : onBack()
+  }
+
+  const handleClickSaveOrContinue = (): void => {
+    if (selectedOption === 'multiAspirate') {
+      if (currentStep < flowSteps.length - 1) {
+        setCurrentStep(currentStep + 1)
+      } else {
+        console.log({ selectedOption, volume, blowOutLocation })
```
smb2268: leftover console statement!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/PipettePath.tsx
```diff
@@ -55,44 +67,153 @@ export function PipettePath(props: PipettePathProps): JSX.Element {
...
+  const setSaveOrContinueButtonText = (): string => {
+    return t(
+      selectedOption === 'multiAspirate' && currentStep < flowSteps.length - 1
+        ? 'shared:continue'
+        : 'shared:save'
+    )
+  }
+
+  // TODO: find the actual min and max for these values.
+  const volumeRange = { min: 1, max: 100 }
```
smb2268: Comments from `BlowOut.tsx` about volume range apply here too!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/PipettePath.tsx
```diff
@@ -27,6 +33,16 @@ interface PipettePathProps {
...
+    state.path
+  )
+  const [currentStep, setCurrentStep] = React.useState<number>(0)
+  const [blowOutLocation, setBlowOutLocation] = React.useState<string>(
+    state.blowOut ?? 'trashBin'
+  )
+  const [volume, setVolume] = React.useState<number>(state.volume)
 
   const allowedPipettePathOptions: PathOption[] = ['single']
   if (state.sourceWells.length === 1 && state.destinationWells.length > 1) {
```
smb2268: we also need to take in consideration the volume here -- I think this logic should be:

```suggestion
    if (
    state.transferType === 'consolidate' &&
    volumeLimits.max >= state.volume * 2
  ) {
    allowedPipettePathOptions.push('multiDispense')
  } else if (
    state.transferType === 'distribute' &&
    volumeLimits.max >= state.volume * 2
  ) {
    allowedPipettePathOptions.push('multiAspirate')
  }
```

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/AirGap.tsx
smb2268: actually for multiAspirate we'd do volume* 2 but for multiDispense it'd be volume*3

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/PipettePath.tsx
```diff
@@ -55,44 +67,153 @@ export function PipettePath(props: PipettePathProps): JSX.Element {
...
+        onBack()
+      }
+    } else {
+      dispatch({ type: ACTIONS.SET_PIPETTE_PATH, path: selectedOption })
+      onBack()
     }
-    onBack()
   }
+
+  const setSaveOrContinueButtonText = (): string => {
```
smb2268:
```suggestion
  const setSaveOrContinueButtonText =  selectedOption === 'multiAspirate' && currentStep < flowSteps.length - 1
        ? t( 'shared:continue')
        : t('shared:save')
```

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TipPosition.tsx
smb2268: yep looks like for PD the logic here is that the min and default are both 1 and the max is 2x the height of the well

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TouchTip.tsx
smb2268: yep here it looks like the default should be (height of well - 1) and the range should be (half height - height)

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
```diff
@@ -1,11 +1,35 @@
...
+  TEXT_ALIGN_LEFT,
+} from '@opentrons/components'
 
 import type {
   QuickTransferSummaryAction,
   QuickTransferSummaryState,
 } from '../types'
+import { ACTIONS } from '../constants'
+import { useTranslation } from 'react-i18next'
+import { useToaster } from '../../../organisms/ToasterOven'
```
smb2268, line 23: this import can be removed

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
```diff
@@ -1,11 +1,35 @@
...
+  SIZE_2,
+  TEXT_ALIGN_LEFT,
+} from '@opentrons/components'
 
 import type {
   QuickTransferSummaryAction,
   QuickTransferSummaryState,
 } from '../types'
+import { ACTIONS } from '../constants'
+import { useTranslation } from 'react-i18next'
```
smb2268: this import should be moved to the top of the list!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
```diff
@@ -16,14 +40,485 @@ export function QuickTransferAdvancedSettings(
...
+          : '',
+      enabled: true,
+      onClick: () => {
+        setSelectedSetting('aspirate_tip_position')
+      },
+    },
+    {
+      option: t('pre_wet_tip'),
+      value: state.preWetTip ? t('option_enabled') : '',
+      enabled: true,
```
smb2268, line 141: Your logic below uses this `enabled` field to determine whether to show the arrow icon, so for this field we'll need to have an exception. Maybe we keep this as true but below change the logic so the icon doesn't appear for this line

app/src/organisms/QuickTransferFlow/reducers.ts
```diff
@@ -199,5 +199,11 @@ export function quickTransferSummaryReducer(
         dropTipLocation: action.location,
       }
     }
+    case 'SET_VOLUME': {
```
smb2268: What was this case added for? we have a `SET_VOLUME` above

smb2268 (changes requested):
The UX here looks great! This has been a ton of work, thanks for digging into all of it. I added in comments with what I think the logic should be at each step. It may be tricky because many are based on the height of the well which means they won't be consistent between aspirate and dispense - let me know if you want to go over any of this!

smb2268:
Situations in which we should disable certain settings from felix:
If transfer type = distribute, disable mix and blowout, add disposal volume
If transfer type = consolidate, disable mix
If labware = reservoir, disable touch tip

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
```diff
@@ -112,15 +112,15 @@ export function Delay(props: DelayProps): JSX.Element {
   if (kind === 'aspirate') {
     wellHeight = Math.max(
       ...state.sourceWells.map(well =>
-        state.source ? state.source.wells[well].depth : 0
+        state.source !== null ? state.source.wells[well].depth : 0
```
smb2268: This should be != null, which checks for both `null` and `undefined`

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/BlowOut.tsx
```diff
@@ -12,7 +12,7 @@ import {
   WASTE_CHUTE_FIXTURES,
   FLEX_SINGLE_SLOT_BY_CUTOUT_ID,
   TRASH_BIN_ADAPTER_FIXTURE,
-  DeckConfiguration,
+  type DeckConfiguration,
```
smb2268: let's separate this import out to a separate line at the bottom of the import list

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/Delay.tsx
```diff
@@ -112,15 +112,15 @@ export function Delay(props: DelayProps): JSX.Element {
...
       )
     )
   } else if (kind === 'dispense') {
     const destLabwareDefinition =
       state.destination === 'source' ? state.source : state.destination
     wellHeight = Math.max(
       ...state.destinationWells.map(well =>
-        destLabwareDefinition ? destLabwareDefinition.wells[well].depth : 0
+        destLabwareDefinition !== null ? destLabwareDefinition.wells[well].depth : 0
```
smb2268: This should also be != instead of !==

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TipPosition.tsx
```diff
@@ -42,15 +42,15 @@ export function TipPositionEntry(props: TipPositionEntryProps): JSX.Element {
   if (kind === 'aspirate') {
     wellHeight = Math.max(
       ...state.sourceWells.map(well =>
-        state.source ? state.source.wells[well].depth : 0
+        state.source !== null ? state.source.wells[well].depth : 0
```
smb2268: same here and below!

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/TouchTip.tsx
```diff
@@ -96,15 +96,15 @@ export function TouchTip(props: TouchTipProps): JSX.Element {
   if (kind === 'aspirate') {
     wellHeight = Math.max(
       ...state.sourceWells.map(well =>
-        state.source ? state.source.wells[well].depth : 0
+        state.source !== null ? state.source.wells[well].depth : 0
```
smb2268, line 99: Same here

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
```diff
@@ -162,18 +174,25 @@ export function QuickTransferAdvancedSettings(
...
     {
-      option: t('touch_tip'),
+      option: 'aspirate_touch_tip',
+      copy: t('touch_tip'),
       value:
         state.touchTipAspirate !== undefined
           ? t('touch_tip_value', { position: state.touchTipAspirate })
           : '',
-      enabled: true,
+      enabled: state.sourceWells.length > 1,
```
smb2268: This check isn't sufficient to determine whether a labware is a reservoir since a user can also select just 1 well in a well plate and there are reservoirs with more than 1 well. Instead we should use `state.source.metadata.displayCategory !== 'reservoir'`

and for destination we'll need to do the same thing but make sure we have the correct labware definition:
`const destinationLabwareDef = state.destination === 'source' ? state.source | state.destination`
`destinationLabwareDef.metadata.displayCategory !== 'reservoir'`

app/src/organisms/QuickTransferFlow/reducers.ts
```diff
@@ -199,5 +209,11 @@ export function quickTransferSummaryReducer(
...
```
smb2268: This can be removed! I added to the SET_PIPETTE_PATH action to accept the disposal volume and blowout location

app/src/organisms/QuickTransferFlow/types.ts
```diff
@@ -61,7 +62,8 @@ export interface QuickTransferSummaryState {
     positionFromBottom: number
   }
   touchTipDispense?: number
-  blowOut?: string // trashBin or wasteChute or 'SOURCE_WELL' or 'DEST_WELL'
+  disposalVolume?: number
+  blowOut?: BlowOutLocation // trashBin or wasteChute or 'SOURCE_WELL' or 'DEST_WELL'
```
smb2268: let's remove this comment since it's no longer accurate

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
```diff
@@ -16,14 +44,548 @@ export function QuickTransferAdvancedSettings(
...
+      },
+    },
+    {
+      option: 'aspirate_air_gap',
+      copy: t('air_gap'),
+      value:
+        state.airGapAspirate !== undefined
+          ? t('air_gap_value', { volume: state.airGapAspirate })
+          : '',
+      enabled: true,
```
brenthagen, line 207: should probably disable if max available capacity is 0:

Screen Shot 2024-07-11 at 1 52 38 PM

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
```diff
@@ -8,6 +8,7 @@ import {
   getTipTypeFromTipRackDefinition,
   TRASH_BIN_ADAPTER_FIXTURE,
   WASTE_CHUTE_FIXTURES,
+  CutoutConfig,
```
smb2268: this needs to be imported as a type on a separate line at the bottom of the imports

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
smb2268: That's a good call out - I can tackle this in a follow on since it'd involve separating out some of the volume range logic into a util

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
```diff
@@ -180,10 +186,10 @@ export function QuickTransferAdvancedSettings(
         state.touchTipAspirate !== undefined
           ? t('touch_tip_value', { position: state.touchTipAspirate })
           : '',
-      enabled: state.sourceWells.length > 1,
+      enabled: !isReservoir,
```
smb2268:
```suggestion
      enabled: state.source.metadata.displayCategory !== 'reservoir',
```

smb2268: We should only disable touchTip aspirate if the source labware is a reservoir and touchTip dispense if the destination labware is a reservoir!

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
```diff
@@ -320,8 +320,8 @@ export function generateQuickTransferArgs(
   const commonFields = {
     pipette: pipetteEntity.id,
     volume: quickTransferState.volume,
-    sourceLabware: sourceLabwareEntity?.id as string,
-    destLabware: destLabwareEntity?.id as string,
+    sourceLabware: sourceLabwareEntity?.id!,
```
smb2268: I'm not familiar with the `!` at the end - is that a replacement for type casting?

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
vegano1: The `isReservoir` const checks if both the source and destination are reservoir which should be sufficient here no?

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
smb2268: ah I see, cool - https://stackoverflow.com/questions/42273853/in-typescript-what-is-the-exclamation-mark-bang-operator-when-dereferenci

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
smb2268: This would disable `touchTip` for _both_ aspirate and dispense if _either_ source or destination labware is a reservoir. We should only disable it for the action that is directly involving the reservoir! So that if the source labware is a well plate and dest labware is a reservoir, `touchTip` will be enabled for aspirate but disabled for dispense

smb2268: so you'll want that change and then on line 267 you should have 
` enabled: destinationLabwareDef.metadata.displayCategory !== 'reservoir',`

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
vegano1: The non-null assertion operator states that even if the variable can be `undefined`, it should never be undefined by this point so tell the compiler to ignore it. It seems like this was added by auto-fixing with eslint. Regardless I think I still need to cast to string here.

app/src/organisms/QuickTransferFlow/QuickTransferAdvancedSettings/index.tsx
vegano1: gotcha! that makes sense

smb2268 (approved):
once the reservoir change is in and tests are passing this is good to merge 🚀

app/src/organisms/QuickTransferFlow/types.ts
```diff
@@ -99,6 +101,7 @@ export type QuickTransferSummaryAction =
   | SetAirGapDispense
   | SetChangeTip
   | SetDropTipLocation
+  | SetVolumeAction
```
smb2268: If you remove this unused `SetVolumeAction` type the ts error in the `reducers` file should go away!

app/src/organisms/QuickTransferFlow/utils/generateQuickTransferArgs.ts
smb2268: Yeah looks like our linter has `Forbidden non-null assertion` errors for these so if you switch them back tests should pass!

PR #15601, opened 2024-07-09
fix(app): ERSplash doesn't care about device

sfoster1:
I don't know where in my brain this idea came from but both the ODD and the desktop want to show this screen.

## testing
- [x] enter ER on the desktop app and it shows a splash screen and a wizard when you click the "enter error recovery button"

### DIFF
--- a/app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
+++ b/app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
@@ -40,7 +40,7 @@ import type { ERUtilsResults } from './hooks'
 import { useHost } from '@opentrons/react-api-client'
 
 export function useRunPausedSplash(showERWizard: boolean): boolean {
-  // Don't show the splash when desktop ER wizard is active.
+  // Don't show the splash when the ER wizard is active.
   return !showERWizard
 }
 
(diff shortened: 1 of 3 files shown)


app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
```diff
@@ -39,12 +39,9 @@ import type { ErrorRecoveryFlowsProps } from '.'
 import type { ERUtilsResults } from './hooks'
 import { useHost } from '@opentrons/react-api-client'
 
-export function useRunPausedSplash(
-  isOnDevice: boolean,
-  showERWizard: boolean
-): boolean {
+export function useRunPausedSplash(showERWizard: boolean): boolean {
   // Don't show the splash when desktop ER wizard is active.
```
SyntaxColoring: Does this comment still match what's going on? `showERWizard` doesn't mean specifically the *desktop* ER wizard, does it?

SyntaxColoring (approved):

app/src/organisms/ErrorRecoveryFlows/RunPausedSplash.tsx
sfoster1: Good catch, it did not. Fixed up

PR #15602, opened 2024-07-09
fix(protocol-designer): fix logic in safePipetteMovements collision detection

ncdiehl11:
Closes RAUT-1046

# Overview

Several errors in `safePipetteMovements` collision logic were causing errors to be raised in `transfer` command creator calls using a 96-channel pipette with COLUMN configuration. This PR addresses these errors by
1. passing `wellLocationPoint` to `getPipetteBoundsAtSpecifiedMoveToPosition` to get correct pipette bounds
2. passing the _surrounding_ labware ID rather than the target labware ID to `getHighestZInSlot`, and
3. not early-returning if the highest Z in the surrounding slot does not exceed to the pipette bounds at that slot.

# Test Plan

- launch dev PD off this branch
- upload the following PD protocol that was incorrectly raising a collision error previously and verify that no warning is raised
- using the same protocol, modify the starting deck state to include a high labware like a 15ml tuberack to the slot left-adjacent to the destination wellplate (slot D1)
- verify that a collision error is properly raised

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/step-generation/src/utils/safePipetteMovements.ts
+++ b/step-generation/src/utils/safePipetteMovements.ts
@@ -244,12 +244,21 @@ const getSlotHasPotentialCollidingObject = (
       ) &&
       pipetteBounds[0].z != null
     ) {
-      const highestZInSlot = getHighestZInSlot(
-        robotState,
-        invariantContext,
-        labwareId
-      )
-      return highestZInSlot >= pipetteBounds[0]?.z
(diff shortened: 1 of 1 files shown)


ncdiehl11:
out of date and fixed by https://github.com/Opentrons/opentrons/pull/15725

PR #15606, opened 2024-07-09
chore(release): internal release notes 1.5.0 to 2.0.0-alpha.1

y3rsh:
## Internal release notes catch up 

image

### Merge to edge before cutting `2.0.0-alpha.1`

> [!NOTE]
> These feel perfunctory on most of our internal releases but there may one day be an internal release where they are not.


ecormany (approved):
These should probably say something more than boilerplate, but
- we haven't done that for like 8 versions
- it's not worth blocking cutting an alpha

so :shipit:

PR #15607, opened 2024-07-09
fix(ci): need always in workflow conditional for analyses snapshot fa…

y3rsh:
## always() left out

These steps won't run unless `always()` is included in the step conditional if the workflow had a previous step fail.


y3rsh:
Used this branch as the source of the workflow file for this run of the tests to prove this fix.

<https://github.com/Opentrons/opentrons/actions/runs/9863715749>

SyntaxColoring (approved):
Gotta love YAML control flow.

PR #15609, opened 2024-07-10
refactor(api): Split deck configuration out of ProtocolEngine.play()

SyntaxColoring:
# Overview

This goes towards EXEC-418.

Broadly, we want to lift run-level concerns out of `opentrons.protocol_engine` and into `opentrons.protocol_runner`. Play/pause/stop, for example. It would be nice if those methods existed elsewhere than `ProtocolEngine`.

One of many small things that's getting in the way of that is that `ProtocolEngine.play()` does two things. It does the expected run-level thing of starting/resuming the command queue, but it also happens to set the engine's deck configuration, which is totally unrelated. As described in #13971, this was a quick hack.

This splits the deck configuration part into its own thing so it can be left in `ProtocolEngine` while the play/pause/stop stuff gets extracted in EXEC-418.

# Test Plan

* [x] `opentrons_simulate` still works (we've had deck-configuration–related hanging bugs before)
* [x] Playing a protocol that needs a waste chute, without a waste chute in the deck configuration, causes a runtime error
* [x] Playing a protocol that needs a waste chute, with a waste chute in the deck configuration, succeeds

# Changelog

* Add a basic unit test for the existing functionality. It seemed to be completely uncovered.
*  Remove the `deck_configuration` argument from `ProtocolEngine.play()` and `PlayAction`. Replace it with `ProtocolEngine.set_deck_configuration()` and `SetDeckConfigurationAction`, which are new. I did my best to preserve the exact existing behavior.

# Review requests

* The semantics of the old `deck_configuration` arg were undocumented and I'm inferring intent from code and memory. Does the docstring on the new `set_deck_configuration()` method correctly describe how this has always worked?
* ~~Any ideas for testing this? Automated test coverage is apparently nonexistent.~~

# Risk assessment

Low.


api/src/opentrons/protocol_runner/protocol_runner.py
```diff
@@ -100,7 +100,11 @@ def was_started(self) -> bool:
 
     def play(self, deck_configuration: Optional[DeckConfigurationType] = None) -> None:
         """Start or resume the run."""
-        self._protocol_engine.play(deck_configuration=deck_configuration)
+        # todo(mm, 2024-07-09): The deck configuration is set at the same time here for
+        # historical reasons. It's unsafe to change the deck configuration mid-run
+        # and we're relying on the caller to not do that.
```
sfoster1, line 105: could we maybe raise if the engine has been played and the passed deck config isn't none

sfoster1 (approved):
Nice minimal refactor. I think going farther down this path is going to put some stress on the "simulated deck config" dichotomy and it might be worth in the background rethinking how to represent that mode

api/src/opentrons/protocol_runner/protocol_runner.py
SyntaxColoring: Yeahhh. I'm also wondering if the deck configuration can be part of the engine's init-time immutable `Config` object.

That means we would need to defer creating the engine until the run is played for the first time. But maybe that's an inherently safer thing to do anyway, given the way maintenance runs can overlap with regular runs? It bugs me that we have multiple engines live at the same time, even if we try to only keep at most one of them "active."

PR #15614, opened 2024-07-10
refactor(api): Change error message for `opentrons_simulate` reading new JSON protocols

SyntaxColoring:
# Overview

Because of PLAT-94, `opentrons_simulate` and `opentrons_execute` can't deal with modern JSON protocols. We were catching this case explicitly and trying to print out a slightly more helpful error message, but for some reason, I wrote that catch to look for exactly `schema == 6` instead of `schema >= 6`. This fixes that.

This also updates the ticket reference, since things on Jira have shuffled around a bit.

# Test Plan

* [x] Run `opentrons_simulate` with a modern JSON protocol and make sure it prints this message.

# Risk assessment

Low.

### DIFF
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -288,7 +288,7 @@ def get_arguments(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
     return parser
 
 
-def execute(  # noqa: C901
+def execute(
     protocol_file: Union[BinaryIO, TextIO],
     protocol_name: str,
     propagate_logs: bool = False,
--- a/api/src/opentrons/execute.py
+++ b/api/src/opentrons/execute.py
@@ -394,7 +394,9 @@ def execute(
             extra_data=extra_data,
         )
     except parse.JSONSchemaVersionTooNewError as e:
-        # See https://opentrons.atlassian.net/browse/PLAT-94
+        # opentrons.protocols.parse() doesn't support new JSON protocols.
+        # The code to do that should be moved from opentrons.protocol_reader.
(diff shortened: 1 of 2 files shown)


api/src/opentrons/execute.py
```diff
@@ -394,11 +394,8 @@ def execute(  # noqa: C901
             extra_data=extra_data,
         )
     except parse.JSONSchemaVersionTooNewError as e:
-        if e.attempted_schema_version == 6:
-            # See Jira RCORE-535.
-            raise NotImplementedError(_JSON_TOO_NEW_MESSAGE) from e
-        else:
-            raise
+        # See https://opentrons.atlassian.net/browse/PLAT-94
```
sfoster1: Unfortunately, this is open-source code but this jira isn't publically accessible. We can leave the link as a note to self, but can we also put like two sentences in here about the required changes explicitly?

api/src/opentrons/simulate.py
```diff
@@ -533,11 +533,8 @@ def simulate(
             extra_data=extra_data,
         )
     except parse.JSONSchemaVersionTooNewError as e:
-        if e.attempted_schema_version == 6:
-            # See Jira RCORE-535.
-            raise NotImplementedError(_JSON_TOO_NEW_MESSAGE) from e
-        else:
-            raise
+        # See https://opentrons.atlassian.net/browse/PLAT-94.
```
sfoster1, line 538: see other comment

sfoster1 (approved):
Looks good but I think we should put some more explanation of issues in there

PR #15619, opened 2024-07-10
fix(hardware_control): Fixing liquid probe bugs found in end to end testing

ryanthecoder:
# Overview

These changes came from testing the end to end implementation of liquid level detection on the ot3 and fixing the bugs that came up.

# Test Plan

Changed the unit tests to match new functionality.
Tested all new methods for liquid level detection on the ot3. 

# Changelog

Changes include:
* Deleting unnecessary checks in ot3controller.py
* Editing ot3api.py to issue the correct probe movements
* Standardizing the attribute names for liquid_presence_detection
* Conditionally adding a require_liquid_presence command before aspirates
* Redoing movement in liquid_probe so that it actually takes a sensible path
* Editing tests accordingly

# Review requests

We are aware of a bug that even though we handle LLD exceptions at the protocol_context level, protocol_engine commands will still be marked as failed, so all future commands will fail. As far as I know for this PR, this only affects detect_liquid_presence on a well with no liquid. Another PR will be coming soon that addresses this.

# Risk assessment

Medium? I'm not sure.


api/lldfunctionexamples.py
```diff
@@ -0,0 +1,93 @@
+from opentrons import protocol_api, types
```
sfoster1: We shouldn't check this in, I think. I don't think this is a good location for it.

api/src/opentrons/hardware_control/backends/ot3controller.py
```diff
@@ -1363,16 +1362,16 @@ async def liquid_probe(
...
-            if (
-                self._subsystem_manager.device_info[
-                    SubSystem.of_mount(mount)
-                ].revision.tertiary
-                != "1"
-            ):
-                raise UnsupportedHardwareCommand(
-                    "Liquid Probe not supported on this pipette firmware"
-                )
+        # if output_option == OutputOptions.sync_buffer_to_csv:
```
sfoster1: Delete rather than checking in commented otu code

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -851,6 +851,16 @@ def liquid_probe_with_recovery(self, well_core: WellCore) -> None:
             origin=WellOrigin.TOP, offset=WellOffset(x=0, y=0, z=0)
         )
 
+        loc = Location(well_core.get_top(10), "Well")
```
sfoster1: let's get this value from some config instead of it being a specific number literal

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -867,6 +877,16 @@ def liquid_probe_without_recovery(self, well_core: WellCore) -> float:
...
```
sfoster1: same here, but actually we can just reuse the `loc` value, it's the same well correct?

sfoster1 (dismissed):
Some style and consistency stuff to fix up. Also, let's get a nicer description in here.

api/src/opentrons/protocol_api/core/engine/instrument.py
```diff
@@ -851,6 +851,16 @@ def liquid_probe_with_recovery(self, well_core: WellCore) -> None:
...
+
+        self.move_to(
+            location=loc,
+            well_core=well_core,
+            force_direct=False,
+            minimum_z_height=None,
+            speed=None,
+        )
```
SyntaxColoring: Why do we need to do this? Doesn't the `liquidProbe` Protocol Engine command already move to the target well?

SyntaxColoring: We probably do need to update PAPI's logical "current well" bookkeeping, though. Like how `aspirate()` and `dispense()` do `self._protocol_core.set_last_location()`.

api/src/opentrons/protocol_api/core/engine/instrument.py
SyntaxColoring: I'm also just noticing that our public `InstrumentContext` methods don't have a way of specifying an X/Y/Z offset, but the `liquidProbe` Protocol Engine command does (although [it's a work in progress](https://opentrons.slack.com/archives/C06M9F26YGH/p1719937084438019)).

Is the difference intentional?

CC @pmoegenburg since you were working on the Protocol Engine side.

SyntaxColoring: `Location(well_core.get_top(10), "Well")` doesn't look like a valid way to construct a `Location`. It's trying to set the `.labware` attribute to a deck slot named `"Well"`. You might have intended to provide an actual instance of the `Well` class, or another value that's valid [per the types of this union](https://github.com/Opentrons/opentrons/blob/edge/api/src/opentrons/types.py#L111-L117).

Also keep in mind that `InstrumentContext` can be the thing that calls `InstrumentCore.move_to()`. I suspect that will be easier because you'll have better access to the things you'll need to build a `location` argument.

SyntaxColoring (changes requested):

api/src/opentrons/protocol_api/core/engine/instrument.py
SyntaxColoring: Discussed in-person. This is working around a weird problem with path planning, which we're looking into.

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -77,8 +77,9 @@ async def execute(self, params: LiquidProbeParams) -> _ExecuteReturn:
         Return the z-position of the found liquid.
 
         Raises:
-            TipNotAttachedError: if there is not tip attached to the pipette
+            TipNotAttachedError: if there is no tip attached to the pipette
             MustHomeError: if the plunger is not in a valid position
+            TipNotEmptyError: if the tip starts with liquid in it(AKA it's wet and unusable for LLD)
```
ryanthecoder: So a tip can theoretically be wet and also empty, if we've used it before and then emptied it. So lets leave out that AKA part

api/src/opentrons/protocol_api/core/engine/instrument.py
aaron-kulkarni: Fixed, we were setting CurrentWell incorrectly in liquidProbe

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -2055,12 +2062,14 @@ def detect_liquid_presence(self, well: labware.Well) -> bool:
...
-        if not isinstance(well, labware.Well):
-            raise WellDoesNotExistError("You must provide a valid well to check.")
+        loc = well.top()
         try:
-            self._core.liquid_probe_without_recovery(well._core)
+            self._core.liquid_probe_without_recovery(well._core, loc)
         except ProtocolCommandFailedError as e:
-            if isinstance(e.original_error, LiquidNotFoundError):
+            # if we handle the error, we change the protocl state from error to valid
+            if isinstance(e.original_error, LiquidNotFoundError) or isinstance(
```
ryanthecoder: I'm not sure we should catch Stall's here, that's something that needs to be recovered from and if we've caused a stall then we've done something else wrong

api/src/opentrons/protocol_api/instrument_context.py
```diff
@@ -261,6 +260,14 @@ def aspirate(
             c_vol = self._core.get_available_volume() if not volume else volume
         flow_rate = self._core.get_aspirate_flow_rate(rate)
 
+        if (
+            self.api_version >= APIVersion(2, 20)
+            and well is not None
+            and self.liquid_presence_detection
+        ):
+            self.require_liquid_presence(well=well)
+            self.prepare_to_aspirate()
```
SyntaxColoring, line 268: Do we need this `self.prepare_to_aspirate()` call? It looks like this is [currently being handed in the `opentrons.protocol_engine` layer?](https://github.com/Opentrons/opentrons/blob/8b88dda01afa364464de3f1506763b1ec08adc6d/api/src/opentrons/protocol_engine/commands/aspirate.py#L104) Did you find that this was necessary for it to work?

api/src/opentrons/protocol_engine/commands/liquid_probe.py
```diff
@@ -99,19 +100,12 @@ async def execute(self, params: LiquidProbeParams) -> _ExecuteReturn:
                 message="Current position of pipette is invalid. Please home."
             )
 
-        current_well = CurrentWell(
-            pipette_id=pipette_id,
-            labware_id=labware_id,
-            well_name=well_name,
-        )
-
```
SyntaxColoring, line 107: Ah, yep, that'd do it. Nice catch.

api/src/opentrons/protocol_engine/execution/pipetting.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, state_view: StateView, hardware_api: HardwareControlAPI) -> N
 
     def get_is_empty(self, pipette_id: str) -> bool:
         """Get whether a pipette has a working volume equal to 0."""
-        return self._state_view.pipettes.get_working_volume(pipette_id) == 0
+        return self._state_view.pipettes.get_aspirated_volume(pipette_id) == 0
```
SyntaxColoring: Does the docstring here need to be updated?

api/src/opentrons/protocol_engine/execution/pipetting.py
```diff
@@ -235,7 +235,7 @@ def __init__(
...
```
SyntaxColoring: Does the docstring here need to be updated?

SyntaxColoring (approved):
This all looks good to me assuming others' pending comments have been resolved. Some small final notes:

api/src/opentrons/protocol_api/instrument_context.py
aaron-kulkarni: Yes, I thought it should be unnecessary too. But if I didn't have this prepare_to_aspirate, it was not aspirating any liquid at all after liquid probing.

api/src/opentrons/protocol_engine/execution/pipetting.py
aaron-kulkarni: Yes, thanks for noticing that.

api/src/opentrons/protocol_api/instrument_context.py
SyntaxColoring: OK, let's keep it for now to unblock testing, but let's keep looking into it. Might be worth leaving a `# todo` comment.

PR #15624, opened 2024-07-11
RQA-2830 fix mac dmg to be x64 binary and use x64 bundled python to pip install

y3rsh:
## Overview

- <https://opentrons.atlassian.net/browse/RQA-2830>
- `macos-11` GitHub runner was intel and now we use macos-latest that is arm [doc](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)
- specify --x64 argument to electron-builder
- ~~per [doc](https://github.com/actions/setup-python?tab=readme-ov-file#supported-architectures) x64 python should be installed so the pip install will be the right platform.  I'm not confident in this.~~
- see comment in #15618
  - we need to use the python we download on mac to do the pip install so the installed packages match the platform correctly

## Test Plan

- [x] build locally and see that it works
- [x] watch CI and install the built package
- [x] validate that once the package dmg is x64 that analysis works (python has packages with the right arch)

# Review requests

- This the right approach?

# Risk assessment

High because our .dmg currently cannot analyze protocols 2.0.0-alpha.0 internal release build.

### DIFF
--- a/app-shell/Makefile
+++ b/app-shell/Makefile
@@ -123,7 +123,7 @@ dist-posix: package-deps
 
 .PHONY: dist-osx
 dist-osx: package-deps
-	$(builder) --mac
+	$(builder) --mac --x64
 	$(MAKE) _dist-collect-artifacts
 
 .PHONY: dist-linux
--- a/app-shell/electron-builder.config.js
+++ b/app-shell/electron-builder.config.js
@@ -1,4 +1,5 @@
 'use strict'
+const { arch } = require('os')
 const path = require('path')
 
 const {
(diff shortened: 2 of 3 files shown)


shlokamin (approved):
lgtm, validated this works locally. after the build generated by CI clears let's make sure that works too before merging

sfoster1 (approved):
Just installed the CI build, all archs are as expected (aka x64) and analysis works!

y3rsh:
validated https://builds.opentrons.com/app/Opentrons-v7.3.1-mac-b44980-RQA-2830-mac-python-package-app-build.dmg as well. merging.

PR #15631, opened 2024-07-11
feat(app-shell): update user license agreement for desktop app installers

shlokamin:
# Overview

Update the EULA text to latest rev from legal.

# Risk assessment

Low


koji (approved):
lgtm

once quick question, will we need to update this text file, when we update Desktop?

shlokamin:
@koji nope, we just need to update this file when there are updates to the end user license agreement (which there should rarely be)

PR #15632, opened 2024-07-11
fix(components, app): update useSwipe to improve swipe action detection

koji:
# Overview
Update useSwipe to improve swipe action detection and update the package, interactjs.

https://github.com/user-attachments/assets/03b74661-8a72-4761-b4f4-844c53027e52

close RQA-2825

# Test Plan
- make teardonw-js && make setup-js 
- setup a run
- start a run
- swipe left and right to switch the screen content

# Changelog

# Review requests

# Risk assessment
low


mjhuff (approved):
Thank you for fixing this! Any idea what was causing the problem initially?

koji:
> Thank you for fixing this! Any idea what was causing the problem initially?

still trying to understand that. So far I haven't seen anything that causing this issue...
but the priority is highest that is the reason why I refactored the hook and Brian recommended I improve that before.

PR #15633, opened 2024-07-11
chore(monorepo): use --cache option when running eslint and prettier.

vegano1:
# Overview 

Speed up local development by using the cache option for eslint and prettier, which will only check changed files by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ watch ?= false
 cover ?= true
 updateSnapshot ?= false
 quiet ?= false
+cache ?= true
 
 FORMAT_FILE_GLOB = ".*.@(js|ts|tsx|yml)" "**/*.@(ts|tsx|js|json|md|yml)"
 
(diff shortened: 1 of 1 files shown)


sfoster1 (commented):
I don't know if this is generally true but i've been running 
```
while true ; do yarn eslint --cache --quiet=true --ignore-pattern "no_modules/" '.*.@(js|ts|tsx)' '**/*.@(js|ts|tsx)'  ; sleep 10 ; done
```
in a terminal, and it gives a lot of false results that go away when i stop caching

vegano1:
@sfoster1 

Yeah, I see the same issue when running that snippet, I can't reproduce it when just running `make lint-js` manually though. We can do without thorough if we have reason to believe this will cause some issues.

y3rsh:
I looked at #15402 when I made the changes in 
https://github.com/Opentrons/opentrons/pull/15350/commits/bac3633432cb17e216ea888bda2beea35bfabff8
but I don't see in Brian's PR why he had to remove cache?

sfoster1:
I bet he was probably running into errors that kept mysteriously persisting, since he was focused on removing them. I don't think it works very well.

koji:
> I looked at #15402 when I made the changes in [bac3633](https://github.com/Opentrons/opentrons/commit/bac3633432cb17e216ea888bda2beea35bfabff8) but I don't see in Brian's PR why he had to remove cache?

I think Brian removed cache option because of [this](https://opentrons.slack.com/archives/CSCLVUW3C/p1719508822081029)

vegano1:
Sounds good, will close!

PR #15636, opened 2024-07-11
refactor(app): Recovery fixups

sfoster1:
This is a couple fixups and a couple removals of hidden rendering for error recovery. Notably,

- Remove all the places where we just didn't render anything on desktop. Now we render the wrong thing, which doesn't seem like a step forward but actually does let us see those screens now, which is good
- In a couple places get some incredibly basic components in; for instance, the footers now use desktop components on desktop
- Remove a lot of `isOnDevice` checks
- Remove `BeforeBeginning`, which wasn't used, didn't have its content in the translation files, and isn't in design
- Fix a couple of the grosser mis-stylings; the content is vaguely the right size and shape and has fewer (not none) scroll bars, and things that should be full width are


app/src/organisms/ErrorRecoveryFlows/shared/RecoveryFooterButtons.tsx
```diff
@@ -97,15 +88,57 @@ function RecoveryPrimaryBtn({
...
+        disabled={tertiaryBtnDisabled}
+        css={ODD_ONLY_BUTTON}
+      />
+      <SecondaryButton
+        onClick={tertiaryBtnOnClick ?? tertiaryBtnDefaultOnClick}
+        disabled={tertiaryBtnDisabled}
+        css={DESKTOP_ONLY_BUTTON}
+      >
+        {tertiaryBtnText}
+      </SecondaryButton>
```
TamarZanzouri, line 147: Nice!

TamarZanzouri (approved):
its a race but this time you win :-) ill fix my branch but fix the tests before

PR #15637, opened 2024-07-11
feat(abr-testing): Compare LPC of errored labware to LPC data

rclarke0:
# Overview

Compares LPC of labware in error step to average LPC coordinates of historical runs.

# Test Plan

Tested on robots with errors including and excluding labware in their error step.

# Changelog

- Gets LPC coordinates of labware from errored protocol step
- Connects to ABR LPC sheet
- Filters data to match robot, labware type, adaptor, module, and slot location
- Finds average of each coordinate that was not from a run with an error
- Creates string that is posted in jira ticket description.

# Review requests

# Risk assessment


sfoster1 (approved):
This is absolutely brilliant, thank you!

PR #15640, opened 2024-07-12
feat(app): add inline notification case to ChildNavigation storybook

koji:
# Overview
add inline notification case to ChildNavigation

When checked the latest CSV design, I noticed that we were missing inline notification case in Storybook.

# Test Plan

# Changelog

# Review requests

# Risk assessment
low

### DIFF
--- a/app/src/organisms/ChildNavigation/ChildNavigation.stories.tsx
+++ b/app/src/organisms/ChildNavigation/ChildNavigation.stories.tsx
@@ -87,3 +87,14 @@ export const TitleWithTwoButtonsDisabled: Story = {
     ariaDisabled: true,
   },
 }
+
+export const TitleWithInlineNotification: Story = {
+  args: {
+    header: 'Header',
+    onClickBack: () => {},
+    inlineNotification: {
+      type: 'neutral',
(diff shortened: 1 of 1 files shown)


app/src/organisms/ChildNavigation/ChildNavigation.stories.tsx
```diff
@@ -87,3 +87,14 @@ export const TitleWithTwoButtonsDisabled: Story = {
...
   },
 }
+
+export const TitleWithInlineNotification: Story = {
+  args: {
+    header: 'Header',
+    onClickBack: () => {},
+    inlineNotification: {
+      type: 'neutral',
+      heading: 'Inline notification',
```
syao1226, line 97: for my task, i added the inline notification text in the design to protocol_setup.json and putted the key as the heading of the inline notification inside ChildNavigation

koji: This is stories.tsx and basically we don't use i18n text this type of use. That is the reason why I just passed text.

jerader (approved):
lgtm!

PR #15645, opened 2024-07-12
fix(hardware-testing): fix call to capacitive probe

caila-marashaj:
Now that the call to `tool_sensors::capacitive_probe` has changed, just update the calls in the hardware testing repo


ryanthecoder (approved):
Thank you

PR #15649, opened 2024-07-15
fix(app): odd: keep bullets on runningprotocol low

sfoster1:
These little bullets that establish the ability to swipe left and right need to be relatively above the rest of the runningprotocol screen yes but they shouldn't be above anything else. So create an explicit z-stacking context for the content part of running protocol.

I'm not quite sure why this is now necessary and wasn't previously, but the rules around what creates an implicit stacking context are complex so we probably changed something 8 components away that removed whatever stacking context existed previously.

## Testing
On ODD:
- Note that the bullets look correct on running protocol
- Note that the bullets are not displayed on top of error recovery anymore
- Note that the other running protocol modals (or "modals", since most of them don't actually use the modal root) still display correctly. This is the cancel-confirmation modal and the intervention modal.


mjhuff (approved):
Noice

TamarZanzouri (approved):
LGTM

PR #15650, opened 2024-07-15
refactor(robot-server): move run time params to their own DB table

sanni-t:
Closes AUTH-527

# Overview

This PR moves primitive RTPs to their own table and refactors the analysis-generation process. This entire refactor makes it easier to add on CSV-parameter handling.

Architecture and database changes:

- Created a new table to store all primitive (non-CSV) run time parameter values used in the analysis.
- Previously, we used to compare only the RTP values provided by the client with the full set of RTP values used in the last analysis. Because the values provided by the client cannot be assumed to be the full set of values of run time parameters in a protocol (parameters that don't have a client-specified value get assigned a default one), we had to make additional comparison of previous RTP values and default values.
- This PR simplifies that comparison by extracting the full set of RTP values for the current analysis beforehand and then comparing it with the full set of previous RTP values.
- This extraction of full RTP values earlier in the process meant that we needed to re-architect the flow of analysis.

Here's what the new analysis flow looks like on a high level-

```mermaid
  flowchart TD
  %% Nodes
      A1(Protocol)
      A2(Run time param values from client)
      B1(Upload protocol endpoint)
      C1("Initialize analyzer: \n Validate and set run time parameters \n using definitions in protocol \nand values passed by client")
      D1("Create a not-ok analysis result. \nSave result to DB. \nReturn analysis completed result.")
      D2("Compare the newly validated RTP values \nwith those of the last analysis (if any) \nof this protocol.")
      E1("No analysis required. Use last analysis in analyses list.")
      E2("In a background task, initiate new analysis using the new RTP values.")
      F1("Pending analysis is created")
      F2("Return pending analysis summary response")
  
  %% Edge connections between nodes
  
      A1 ---> B1
      A2 ---> B1
      B1 ---> C1
      C1 -- Initialization/RTP validation failed --> D1
      C1 -- RTP values validated successfully --> D2
      D2 -- Matching RTPs --> E1
      D2 -- Non-matching RTPs --> E2 --> F1 --> F2
```

```mermaid
flowchart TD
%% Nodes
    EF("Analyze task")
    G1("Pending analysis is updated to a \n'Completed analysis' with an error.")
    G2("Pending analysis is updated to a \n successful 'Completed analysis'.")
    End(" \nAnalysis and RTP values are saved to DB.")

%% Edge connections between nodes

    EF -- Analysis Error --> G1 --> End
    EF -- Sucessful analysis --> G2 --> End
```

# Test Plan

- Since this is only a refactor, existing integration and snapshot tests passing should be a good enough check for whether this is accidentally changing anything it shouldn't be
- This changes the DB by removing a column from a table in schema 4 so we should make sure that a robot on DB v4, that has protocols (both RTP & non-RTP), gets migrated to v5 successfully and there are no issues handling their analyses.

# Changelog

- added the new primitive RTP table and removed the 'rtp values and default' column from analysis table
- updated DB v4 -> v5 migration
- refactored `AnalysesManager`, `ProtocolAnalyzer`, `AnalysisStore` and protocols `router` to use the new flow as shown above.
- updated tests 

# Review requests

- @SyntaxColoring and I had discussed most of these changes pretty extensively. I was initially hesitant towards moving the RTP validation and extraction earlier in the process because of the amount of refactoring that would result in, but as i was working towards adding CSV parameter handling, the early validation method deemed quite useful, especially since CSV parameters will be relying on external files to set their 'values'. (so thanks for the suggestion Max!)
- Does the new flow make sense? Anything that isn't clear?

# Risk assessment

Medium-high. It changes quite a critical part of the protocol upload and analysis process. But the fact that all integration tests and analyses snapshots are passing makes me feel quite confident that the existing behavior of these endpoints has not changed by this work.


robot-server/robot_server/persistence/_files_and_directories.py
```diff
@@ -0,0 +1,6 @@
+"""Files and directories in persistent storage."""
```
SyntaxColoring, line 1: Good idea, thanks.

robot-server/robot_server/persistence/tables/schema_5.py
```diff
@@ -49,11 +49,35 @@
...
+        primary_key=True,
+    ),
+    sqlalchemy.Column(
+        "analysis_id",
+        sqlalchemy.ForeignKey("analysis.id"),
+        nullable=False,
+    ),
     sqlalchemy.Column(
-        "run_time_parameter_values_and_defaults",
+        "parameter_variable_name",
```
SyntaxColoring, line 78: ~~It might make sense to throw in [`unique=True`](https://docs.sqlalchemy.org/en/20/core/constraints.html#unique-constraint) on the `parameter_variable_name` column.~~

Whoops, no, sorry, definitely not just on the `parameter_variable_name` column. It would need to be on the `"analysis_id", "parameter_variable_name"` combination.

This is just a nice-to-have.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
```diff
@@ -4,50 +4,129 @@
 
 - Adds a new "protocol_kind" column to protocols table
 - Adds a new "data_files" table
+- Removes the "run_time_parameter_values_and_defaults" column of analysis_table and
+  creates a separate analysis_primitive_type_rtp_table instead. The migration does not
+  port the data from run_time_parameter_values_and_defaults into analysis_primitive_type_rtp_table.
+  The consequence of which is that any checks for previous matching analysis (for protocols with RTPs only)
+  will fail and a new analysis will be triggered. This new analysis will then save its
+  RTP data to the new table. RTP data belonging to previous analyses will still be available
+  as part of the completed analysis blob.
```
SyntaxColoring, line 13: I think schema 5 is widely distributed enough, and these changes are fundamental enough to what schema 5 "is," that we should move to a fresh directory instead of continuing to use `5/`.

If any of the bots that are currently on schema 5 (SW dev bots, QA bots, maybe ABR?) upgrade to these changes, things will break because `analysis_primitive_type_rtp_table` will be missing, and the server won't know that it needs to run a migration.

I think we could still call this "schema 5" and just change the directory from `5/` to, say, `5.1/`:

https://github.com/Opentrons/opentrons/blob/2ff941786feff0b20d9df3d475db96d7f9a88c92/robot-server/robot_server/persistence/persistence_directory.py#L54

So in other words, we'd be replacing `3/ -> 4/ -> 5/` with `3/ -> 4/ -> 5.1/`.

This would pull everyone's run and protocol history back in time to what it was in schema 4, which is unfortunate and which we'll have to communicate, but it's better than the alternatives, I think.

robot-server/robot_server/persistence/tables/schema_5.py
SyntaxColoring, line 1: <strike>This is completely unrelated to your work, but @vegano1 and I noticed that `protocol_table` is missing an index on its `protocol_kind` column, now that the HTTP API is querying it. If you end up making a schema break anyway, it might make sense to throw that in with your changes.

[Edit] And I remember now that the `protocol_kind` column also does not have to be nullable, and it can be [an enum](https://github.com/Opentrons/opentrons/pull/15650#discussion_r1679579117).

But don't worry about it if it complicates anything.</strike>

Off-topic for this PR.

robot-server/robot_server/persistence/tables/schema_5.py
```diff
@@ -49,11 +49,35 @@
...
         sqlalchemy.String,
-        nullable=True,
+        nullable=False,
+    ),
+    sqlalchemy.Column(
+        "parameter_type",
+        sqlalchemy.String,
```
SyntaxColoring: Can we use a [`sqlalchemy.Enum`](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Enum) for `parameter_type` instead of a plain `sqlalchemy.String`, to make this more self-documenting and to make it harder to store incorrect values in the future?

We haven't used `sqlalchemy.Enum` before, so we'll need to set the pattern. I suggest:

* We should use a new `Enum` class that's just for the database, and not reused from the HTTP API, and explicitly map back and forth. This is a pretty soft opinion, so definitely feel free to share one enum if you think that turns out better. I guess I'm worried about entangling Pydantic+FastAPI with SQLAlchemy any more than necessary.
* We should not inherit the enum class from `str`, but we should still give each enum value a string representation, like `FOO = "bar"`.
* We should use [`values_callable`](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Enum.params.values_callable) to get SQLAlchemy to store the enum values (`"bar"`) instead of their names (`FOO`). I don't know about you, but it's pretty surprising to me that it would store the names by default.
* We should use [`create_constraint=True`](https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.Enum.params.create_constraint) because I'm pretty sure SQLite will not do that natively, and having a constraint seems like a good idea.

SyntaxColoring (changes requested):
Working my way through this, looks great, thanks! Here are some requests.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
```diff
@@ -4,50 +4,129 @@
...
+            source_dir / PROTOCOLS_DIRECTORY, dest_dir / PROTOCOLS_DIRECTORY
+        )
+
+        source_db_file = source_dir / DB_FILE
+        dest_db_file = dest_dir / DB_FILE
 
         # Append the new column to existing protocols in v4 database
         with ExitStack() as exit_stack:
+            source_engine = exit_stack.enter_context(sql_engine_ctx(source_db_file))
+            schema_4.metadata.create_all(source_engine)
```
SyntaxColoring: Nitpick: This `create_all()` call should no-op, because the source should be provided to us already complete and conformant to schema v4. Can we remove it for clarity?

This might have been copied from `up_to_3.py`? It was needed there basically because it was *up to 3* as opposed to *2 to 3.* It's confusing and I should have fixed it, sorry.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
```diff
@@ -4,50 +4,129 @@
...
+        source_transaction,
+        dest_transaction,
+        order_by_rowid=True,
+    )
+    copy_rows_unmodified(
+        schema_4.run_command_table,
+        schema_5.run_command_table,
+        source_transaction,
+        dest_transaction,
+        order_by_rowid=True,
```
SyntaxColoring, line 91: Nitpick: I think this is harmless, but `order_by_rowid` can be `False` for `run_command_table` because it has explicitly-declared `row_id` and `index_in_run` columns.

I wonder if `order_by_rowid=True` is actually any slower or more memory-hungry than `=False`. Maybe it should just always be `True` to simplify things.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
```diff
@@ -4,50 +4,129 @@
...
+    )
+    insert_new_analyses = sqlalchemy.insert(schema_5.analysis_table)
+    for old_row in source_transaction.execute(select_old_analyses).all():
+        dest_transaction.execute(
+            insert_new_analyses,
+            id=old_row.id,
+            protocol_id=old_row.protocol_id,
+            analyzer_version=old_row.analyzer_version,
+            completed_analysis=old_row.completed_analysis,
+            # run_time_parameter_values_and_defaults column is omitted
```
SyntaxColoring, line 130: It looks like we're dropping the old column here and not copying its data into the new table. Is that correct?

If so, is that safe? e.g. from a caching perspective, won't it look like a protocol has been successfully analyzed without any parameter values, whereas in fact it *did* need parameter values, and they were just lost in the migration?

If it's a pain to migrate the actual RTP values, we could consider outright dropping *all* of the analysis data. It's supposed to be repeatable, anyway, so the worst case is that the robot just does it anew the next time the user tries to run the protocol, right?

robot-server/robot_server/protocols/rtp_resources.py
```diff
@@ -0,0 +1,68 @@
...
+                converted_num = int(value)
+            except ValueError:
+                return float(value)
+            return converted_num
+
+        param_types: Dict[str, Callable[[str], Union[str, float, bool]]] = {
+            "str": str,
+            "int": _int_converter,
+            "float": float,
+            "bool": bool,
```
SyntaxColoring: I don't think this serialization/deserialization strategy for `bool` will work.

`to_sql_values()` returns `str(self.parameter_value)`, e.g. `str(False)`, which is `"False"`.

But then, going the other way, `from_sql_row()` will do `bool(parameter_value)`, e.g. `bool("False")`, which is `True`.

robot-server/robot_server/protocols/rtp_resources.py
```diff
@@ -0,0 +1,68 @@
...
+
+        parameter_val_str = sql_row.parameter_value
+        assert isinstance(parameter_val_str, str)
+
+        def _int_converter(value: str) -> float:
+            try:
+                converted_num = int(value)
+            except ValueError:
+                return float(value)
+            return converted_num
```
SyntaxColoring: Could you clarify what this is doing? I can't see why `int(value)` would fail.

robot-server/robot_server/protocols/rtp_resources.py
SyntaxColoring: If we still want every primitive RTP type to share one table, I would consider using `json.dumps()`/`json.loads()` just for the sake of not having to roll our own serialization/deserialization. We would use JSON for *just `parameter_value`*, basically replacing this callable dict, not for `parameter_variable_name` or `parameter_type`. As long as we stick to JSON scalars, it won't introduce formatting or field ordering nondeterminism concerns.

But if you want to keep the DIY serialization/deserialization and special-case `bool`, that works for me too.

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -34,45 +37,57 @@ def __init__(
...
+        """Return the protocol resource."""
+        return self._protocol_resource
+
+    def get_verified_run_time_parameters(self) -> List[RunTimeParameter]:
+        """Get the extracted RTPs."""
+        assert self._orchestrator is not None
+        return self._orchestrator.get_run_time_parameters()
 
-    async def load_runner(
+    async def load_orchestrator(
```
SyntaxColoring, line 52: Thanks for this rename.

robot-server/robot_server/protocols/protocol_store.py
```diff
@@ -347,6 +359,12 @@ def _sql_get_all_from_engine(
         return [_convert_sql_row_to_dataclass(sql_row=row) for row in all_rows]
 
     def _sql_remove(self, protocol_id: str) -> None:
+        analyses_using_protocol = self._get_referencing_analysis_ids(protocol_id)
```
SyntaxColoring: This is opening a transaction to get some data, then computing things to delete, then opening a separate transaction to do the deletion. For atomicity and isolation, can we bring this under one transaction?

There are a few ways to do this. Here's one.

Right now, in this:

```python
        delete_analysis_rtps_statement = sqlalchemy.delete(
            analysis_primitive_type_rtp_table
        ).where(
            analysis_primitive_type_rtp_table.c.analysis_id.in_(analyses_using_protocol)
        )
```

You're passing the separately-computed Python list, `analyses_using_protocol`, as the argument. You can actually pass a `sqlalchemy.select()` construct instead and keep all of the logic inside SQL. Something like:

```python
        delete_analysis_rtps_statement = sqlalchemy.delete(
            analysis_primitive_type_rtp_table
        ).where(
            analysis_primitive_type_rtp_table.c.analysis_id.in_(
                sqlalchemy.select(analysis_table.c.id).where(analysis_table.c.protocol_id == protocol_id)
            )
        )
```

You could also do this with some kind of join. Whatever you think is clearer.

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -34,45 +37,57 @@ def __init__(
...
         self._protocol_resource = protocol_resource
+        self._orchestrator: Optional[RunOrchestrator] = None
+
+    @property
+    def protocol_resource(self) -> ProtocolResource:
+        """Return the protocol resource."""
+        return self._protocol_resource
+
+    def get_verified_run_time_parameters(self) -> List[RunTimeParameter]:
+        """Get the extracted RTPs."""
```
SyntaxColoring: Could you elaborate a little here on why we're calling these "verified" run time parameters?

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -207,16 +207,49 @@ async def update(
...
+        )
+        completed_analysis_resource = CompletedAnalysisResource(
+            id=completed_analysis.id,
+            protocol_id=protocol_id,
+            analyzer_version=_CURRENT_ANALYZER_VERSION,
+            completed_analysis=completed_analysis,
+        )
+        await self._completed_store.make_room_and_add(
+            completed_analysis_resource=completed_analysis_resource,
+            primitive_rtp_resources=[],
```
SyntaxColoring, line 251: Is `primitive_rtp_resources=[]` right? If an analysis has one of these errors, shouldn't we store that the analysis was attempted with the RTP values that it was actually attempted with? If we don't, I'm a little worried about the implications of this for caching.

Maybe I'm getting confused about whether this column represents the *actual fully-resolved* RTP values or merely the *client-supplied* RTP values.

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -258,9 +291,6 @@ def get_summaries_by_protocol(self, protocol_id: str) -> List[AnalysisSummary]:
         completed_analysis_ids = self._completed_store.get_ids_by_protocol(
             protocol_id=protocol_id
         )
-        # TODO (spp, 2024-06-05): populate runTimeParameters in the completed analysis summaries once
-        #  we start saving RTPs to their own table. Currently, fetching RTPs from a
-        #  completed analysis requires de-serializing the full analysis resource.
         completed_analysis_summaries = [
             AnalysisSummary.construct(id=analysis_id, status=AnalysisStatus.COMPLETED)
             for analysis_id in completed_analysis_ids
```
SyntaxColoring, line 297: Just making sure, are you intentionally removing this `# todo` comment without populating `AnalysisSummary`'s `runTimeParameters` field here?

robot-server/robot_server/protocols/router.py
```diff
@@ -414,30 +428,44 @@ async def _start_new_analysis_if_necessary(
...
+    else:
+        if (
+            force_analyze
+            or
+            # Unexpected situations, like powering off the robot after a protocol upload
+            # but before the analysis is complete, can leave the protocol resource
+            # without an associated analysis.
+            len(analyses) == 0
+            or
+            # The most recent analysis was done using different RTP values
```
SyntaxColoring, line 456: Hypothetically, is it correct that the data is accessible enough now that we could efficiently check *all* analyses, not just the most recent one, if we wanted to?

SyntaxColoring (commented):
Final set of comments.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
sanni-t: I think that's a good idea

robot-server/robot_server/protocols/router.py
sanni-t: Yep!

robot-server/robot_server/protocols/rtp_resources.py
sanni-t: Darn! Good catch!! 
Hmm, alright, going to see if using json would be better than special-casing booleans

robot-server/robot_server/protocols/rtp_resources.py
sanni-t: `int("5.0")` fails. We allow having `5.0` as a valid integer param value.

robot-server/robot_server/protocols/analysis_store.py
sanni-t: > Maybe I'm getting confused about whether this column represents the actual fully-resolved RTP values or merely the client-supplied RTP values.

It represents the fully-resolved RTP values.

> Is primitive_rtp_resources=[] right? If an analysis has one of these errors, shouldn't we store that the analysis was attempted with the RTP values that it was actually attempted with? If we don't, I'm a little worried about the implications of this for caching.

I have been thinking about this, although more from a troubleshooting perspective. Let me write down the different data that gets saved when analysis fails at different stages and we can discuss if/ how to improve the rtp error handling.

robot-server/robot_server/protocols/analysis_store.py
sanni-t: er.. I am confused at my own comment

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
sanni-t: > won't it look like a protocol has been successfully analyzed without any parameter values, whereas in fact it did need parameter values, and they were just lost in the migration?

The parameters also get saved as part of the full analysis and that's what the app uses for getting parameters' info. The `run_time_parameter_values_and_defaults` column was only used internally to check if the last analysis' parameters match the ones in the new request. So what will happen here is that the migration will still keep info of the RTPs in each of the persisted analyses but when the internal code tries to check for previous RTP values, it will just not find any matching values and will initiate a new analysis (*see note below). And it sounds like that's the behavior you were proposing too.

Removing analyses for uploaded protocols will be tougher to handle, especially because the ODD relies on analysis info to show the protocols on the screen. Without any analyses, it will just show all protocols with errors.

(*as I was typing this, I realized I have added a check that verifies that parameters defined in last analysis are same as in current analysis. As in, if the last analysis has three RTPs, then the new analysis should also contain the same three parameters, which could have different values obviously. In order for my above described behavior, I will need to make an exception for when there were no RTPs found corresponding to the last analysis, and I think that's acceptable)

robot-server/robot_server/protocols/protocol_analyzer.py
sanni-t: Because they are the full, validated params, as opposed to the RTP key-value pairs received from a client.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
sanni-t: Talked with the team and we've decided to put these changes inside a new version so that the internal, v5 changes aren't lost. That means for someone on v4, there will be a transient v5 folder introduced and the migration will take slightly longer because of having to do 2 migrations, but v4 to v5 only involves addition of a new column ('protocol kind') while rest of the DB is copied as is, so it shouldn't be too time consuming. What do you think?

sanni-t (commented):
Notes from meeting:
- Update the persistence snapshot tests to include RTP checks.

robot-server/robot_server/protocols/rtp_resources.py
```diff
@@ -0,0 +1,73 @@
...
+from opentrons.protocols.parameters.types import PrimitiveAllowedTypes
+
+
+@dataclass
+class PrimitiveParameterResource:
+    """A primitive runtime parameter from a completed analysis, storable in a SQL database."""
+
+    analysis_id: str
+    parameter_variable_name: str
+    parameter_type: str
```
SyntaxColoring, line 20: Does `parameter_type` need to change, or can it change, to an `Enum`, now that the underlying column is a SQLAlchemy enum?

I would expect SQLAlchemy to give us the enum, not the str, when we do `results = transaction.execute(statement).all()`. But maybe I'm misunderstanding something.

robot-server/robot_server/protocols/rtp_resources.py
SyntaxColoring: I think I'm confused, sorry.

I understand why you'd want to treat `5.0` as a valid integer param for the HTTP API and any other JSON documents, because JavaScript and JSON don't distinguish between ints and floats.

But are you saying that we're also constructing instances of this class where `parameter_type` is `"int"` and `parameter_value` is `5.0`, and then storing `("int", "5.0")` in SQL? If so, can we not do that? I'd find it pretty surprising.

robot-server/robot_server/protocols/protocol_store.py
```diff
@@ -347,6 +348,18 @@ def _sql_get_all_from_engine(
         return [_convert_sql_row_to_dataclass(sql_row=row) for row in all_rows]
 
     def _sql_remove(self, protocol_id: str) -> None:
+        select_referencing_analysis_ids = (
+            sqlalchemy.select(analysis_table.c.id)
+            .where(analysis_table.c.protocol_id == protocol_id)
+            .order_by(sqlite_rowid)
```
SyntaxColoring: Nitpick: no need for this `.order_by` since we're just testing that things exist in this set, not returning it as an ordered list.

```suggestion
```

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
SyntaxColoring: OK. So it sounds like that leaves us with the same data stored in two different places, and they won't always match? I think that's going to be error-prone, and your "as I was typing this..." note is an example of why.

I think ideally, we should migrate the old representation to the new representation. I'm not sure whether we can do that performantly, i.e. without the migration taking 25 minutes. I suspect the answer is "yes, with some care."

If we end up *not* doing that migration, I guess it's not a dealbreaker because we can do it later. But can we leave a comment here with your explanation of why it's okay to skip?

sanni-t: > So it sounds like that leaves us with the same data stored in two different places, and they won't always match?

Not really. The only two possibilities are- 
- the analysis blob contains the full RTP params list and there's no corresponding RTP values in the RTP table (will happen for analyses migrated from v4)
- the analysis blob contains the full RTP params list and the same RTP values are in the RTP table for that analysis (will happen for any analyses generated since v6)

robot-server/robot_server/protocols/protocol_store.py
sanni-t: I thought I removed this. Good catch

robot-server/robot_server/protocols/rtp_resources.py
sanni-t: Yep, it returns the enum, and we use the string value of it when creating the ParameterResource object from the db row. This is more favourable because the parameter types declared in the engine are string literals. Also the enum is more of a database concern and I don't want to leak that into rest of the server code.

robot-server/robot_server/persistence/_migrations/v4_to_v5.py
SyntaxColoring: > * the analysis blob contains the full RTP params list and there's no corresponding RTP values in the RTP table (will happen for analyses migrated from v4)

Right, to clarify, that's exactly what I mean by "the same data stored in two different places, and they won't always match."

sanni-t: Talked about this in person. The RTP table is the record of all run time parameters that were successfully validated with the new analyzer version. The RTPs in the table will differ from those in the completed analysis if either the RTPs in the analysis had failed validation OR if they were not analyzed with the updated version. 
We agree that this is an acceptable scenario.

robot-server/robot_server/protocols/analysis_store.py
sanni-t: Talked about this in person. As mentioned above, we will not be saving any entries in RTP table for validation-failed parameters. The partially-validated RTPs will still live in the completed analysis blob for inspecting analyses retrospectively.

robot-server/robot_server/protocols/analysis_store.py
sanni-t: Alright, I remembered what this was about. And yes, it's an intentional removal. We decided to not save the full RTPs (with definitions) to their own table(s) so the only way to fetch the full list of RTPs with definitions is by de-serializing the completed analysis resource, and we don't want to do that for `AnalysisSummary` since its intention was to provide a glimpse of the analysis without having to make the expensive de-serializing step.
So I don't see a good way of adding RTPs to completed summaries and nor is it needed by anything.

robot-server/robot_server/protocols/rtp_resources.py
sanni-t: Yea, it can do that. I believe it's because we accept integers defined with a `.0` and don't truncate the trailing zero. We'll take a look at it again if it's causing confusion or any problems but I don't think it has to be in this PR.

robot-server/robot_server/protocols/rtp_resources.py
SyntaxColoring: OK, I see now that you're doing the `str`<->`Enum` conversion here instead of in `CompletedAnalysisStore`. That works for me.

I wouldn't have considered `PrimitiveParamSQLEnum` to be too low-level to appear in `CompletedAnalysisStore`, for what it's worth.

SyntaxColoring (approved):
Great, thanks!

robot-server/robot_server/protocols/analyses_manager.py
```diff
@@ -11,6 +14,15 @@
 from robot_server.protocols import protocol_analyzer
 from robot_server.protocols.protocol_store import ProtocolResource
 from robot_server.service.task_runner import TaskRunner
+import robot_server.errors.error_mappers as em
+
+
+class FailedToInitializeAnalyzer(Exception):
```
sfoster1, line 20: Should this be an enumerated error so it can carry more context?

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -342,48 +360,32 @@ async def matching_rtp_values_in_analysis(
...
+        if len(primitive_rtps_in_last_analysis) == 0:
+            # Protocols migrated from v4 will not have any entries in RTP table,
+            # this is fine and we should just trigger a new analysis and have
+            # the new values be stored in the RTP table.
             return False
-        for (
-            parameter,
-            prev_value_and_default,
-        ) in rtp_values_and_defaults_in_last_analysis.items():
+        assert set(param.variableName for param in new_parameters) == set(
```
sfoster1, line 381: Are we sure about this one? I really don't like using `assert` in this way in production code. If we're going to be paranoid, why not have this function (1) log something big and then (2) just return `False`, since if these parameter names don't match then they're definitionally different?

robot-server/robot_server/protocols/analysis_store.py
```diff
@@ -407,22 +409,22 @@ def add(
         self,
         protocol_id: str,
         analysis_id: str,
-    ) -> PendingAnalysis:
+        run_time_parameters: List[RunTimeParameter],
+    ) -> None:
```
sfoster1, line 413: why change this? feels like a nice convenience thing to return the newly created object, and it's a smaller diff

robot-server/robot_server/protocols/completed_analysis_store.py
```diff
@@ -336,8 +267,28 @@ def get_ids_by_protocol(self, protocol_id: str) -> List[str]:
...
+        """Get the saved primitive RTP values from database."""
+        statement = (
+            sqlalchemy.select(analysis_primitive_type_rtp_table)
+            .where(analysis_primitive_type_rtp_table.c.analysis_id == analysis_id)
+            .order_by(sqlite_rowid)
+        )
+        with self._sql_engine.begin() as transaction:
+            results = transaction.execute(statement).all()
+
+        rtps: Dict[str, PrimitiveAllowedTypes] = {}
```
sfoster1, line 282: nit:
```
resources = [PrimitiveParameterResource.from_sql_row(row) for row in results]
rtps = {param. parameter_variable_name: param.parameter_value for param in resources}
```
? Means you don't have to explicitly type `rtps`

robot-server/robot_server/protocols/protocol_analyzer.py
```diff
@@ -34,45 +37,57 @@ def __init__(
         """Initialize the analyzer and its dependencies."""
         self._analysis_store = analysis_store
         self._protocol_resource = protocol_resource
+        self._orchestrator: Optional[RunOrchestrator] = None
```
sfoster1, line 40: Looking around the code, it seems like the first thing you always do with a protocol analyzer is call `load_runner` (now `load_orchestrator`) with the RTPs, including [in this pr](https://github.com/Opentrons/opentrons/pull/15650/files#diff-f1a7ff907d0ad851ce0c41eed2fed7cf46d31aa7e4f8df7669fd59cb6e4cd641R58). Given that, instead of having `load_orchestrator` can we
- Require an orchestrator instance as an initializer param, meaning it's not-null by construction
- make `create_protocol_analyzer` take the optional `run_time_param` values and build this there

That way we don't need all these asserts and we make it impossible to call `get_verified_run_time_parameters` in an illegal way.

sfoster1 (approved):
Nice, I'm always for breaking stuff into more steps to better mirror semantics. I have a couple things I'd love to see changed but if stuff depends on this getting merged it's ok:
- Remove some null checks by slightly reorganizing
- Make remaining checks not just `assert`s but actual errors
- Just saying, `Scalar` feels like a better name than `Primitive`

sanni-t:
Thanks for the feedback @sfoster1. I'll address them in a follow-up PR since this one's quite overdue for merging.

robot-server/robot_server/protocols/analysis_store.py
sanni-t: @sfoster1 can you explain why you don't like using `assert` in this way? 
I prefer assert rather than silently logging because it makes sure that the error is actually brought to attention. It's a practice we've followed in most of robot server and engine code. If we only log the error and continue running (by returning `False` in this case), then the server might run into some other error down the line because failing this assertion means that there's a bug in our code somewhere else. So it's better to fail right away and know why it failed rather than failing a little later with a possibly unclear reason.

PR #15652, opened 2024-07-15
feat(hardware-testing): add testing teams feature requests

ryanthecoder:
# Overview

- Makes the script copy the exact way that Protocol engine calls the ot3api liquid probe
- Remove the Radwag vial as a default, since multiprobe with an empty vial can result in a collision and they're fragile
- Adds a flag that does a "pre-wet" of the tip to test using the probe with wet tips
- adds a flag to override the plunger's solo move time to test that part of the noise reduction
- adds my implementation of "blockage" detection to print that out in the results csv

# Test Plan

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):

api/src/opentrons/protocol_engine/execution/pipetting.py
```diff
@@ -188,7 +188,7 @@ async def liquid_probe_in_place(
             pipette_id=pipette_id
         )
         z_pos = await self._hardware_api.liquid_probe(
-            mount=hw_pipette.mount, max_z_dist=well_depth - lld_min_height
+            mount=hw_pipette.mount, max_z_dist=well_depth - lld_min_height + 2
```
pmoegenburg: Can we make this method take a well_location and use the well_location.offset.z instead of the magic number 2 here?

api/src/opentrons/hardware_control/ot3api.py
```diff
@@ -2666,6 +2666,8 @@ async def liquid_probe(
             except PipetteLiquidNotFoundError as lnfe:
                 error = lnfe
             pos = await self.gantry_position(checked_mount, refresh=True)
+        await self.move_to(checked_mount, probe_start_pos + top_types.Point(z=2))
```
pmoegenburg, line 2669: This can just be a move to probe_start_pos, and then line 2671 can be removed

pmoegenburg (dismissed):
Looks good! Just one change that I think would clean things up a bit

api/src/opentrons/hardware_control/ot3api.py
ryanthecoder: Cant actually physically, this 2 is a different magic 2. Will ticket in our tech debt epic

PR #15654, opened 2024-07-15
fix(app): odd: fix ER fundamental layout issues

sfoster1:
I love Box. I don't know why. I always want to use it and every single time it breaks something. This is one of those times

- Remove Box as the container for intervention modal because it makes all of our common tools for layout not function
- Make the ODD containers 100% instead of explicit size everywhere because if you are using the right kind of container they won't overflow anyway
- 3am going back for more <Box />: inRecovery FooterButtons, add a small dummy element to the button row when there is no "go back" button so the primary button is still on the right

## Testing
all on ODD:
- There shouldn't be scrollbars in the error recovery components anymore
- If there's no "go back" button, the continue button should still be on the right

### DIFF
--- a/app/src/organisms/ErrorRecoveryFlows/shared/RecoveryFooterButtons.tsx
+++ b/app/src/organisms/ErrorRecoveryFlows/shared/RecoveryFooterButtons.tsx
@@ -3,7 +3,8 @@ import { useTranslation } from 'react-i18next'
 import { css } from 'styled-components'
 
 import {
-  ALIGN_CENTER,
+  ALIGN_FLEX_END,
+  Box,
   Flex,
   JUSTIFY_SPACE_BETWEEN,
   SPACING,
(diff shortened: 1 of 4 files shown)


TamarZanzouri (approved):
changes LGTM! need to test on the robot :-)

app/src/organisms/ErrorRecoveryFlows/shared/RecoveryFooterButtons.tsx
```diff
@@ -62,7 +63,9 @@ function RecoveryGoBackButton({
         {t('go_back')}
       </SecondaryButton>
     </Flex>
-  ) : null
+  ) : (
+    <Box />
```
mjhuff, line 67: I'm glad you at least got one `Box` in.

mjhuff (approved):
Tested and lgtm! In the end, we are all forced to `Flex`.

app/src/organisms/ErrorRecoveryFlows/shared/RecoveryFooterButtons.tsx
sfoster1: Gotta have it

PR #15655, opened 2024-07-15
Add AVG historical temp and RH conditions to error ticket 

rclarke0:
# Overview

Add AVG historical temp and RH conditions to error ticket 

# Test Plan

- Connects to ABR sheet
- Connects to ABR Ambient conditions sheet
- Filters out data between time stamps and robot and same protocol
- Adds a string that compares found averages to the ticket description.

# Changelog

# Review requests

# Risk assessment


sfoster1 (approved):
Nice

PR #15657, opened 2024-07-15
feat(protocol-designer): create redesign ff and scaffolding for new nav

jerader:
closes AUTH-555

# Overview

Creates the top level PD redesign ff and scaffolding using `react-router-dom` in PD in preparation for the PD redesign.

# Test Plan

Build your dependencies, i added the latest version of `react-router-dom`!

Enable dev FFs and go to the setting page and click "Enable redesign". Click confirm on the modal and you should get the following page:

Screenshot 2024-07-15 at 13 30 39

That is the landing page. Click on "create new" and it should enable the nav bar with 5 options: 'create new protocol', 'liquids', 'protocol overview', 'protocol steps', 'starting deck state'. If you click on them, it should render the components, which right now is only the text. If you click on the top button "turn off redesign", then you should see the normal PD

here are the low-fi PD figma designs. Since they are low-fi, i didn't do anything crazy with adding designs yet. This is simply a scaffolding PR. https://www.figma.com/design/WbkiUyU8VhtKz0JSuIFA45/Feature%3A-Protocol-Designer-Phase-1?node-id=3234-153628&t=VICg8I3ywHwrcRPg-0 

# Changelog

- add react-router-dom
- create pages directory with a folder for each page in the redesign
- create a nav bar
- create simple components for each page and a todo test
- wire into `ProtocolEditor`
- create feature flag for the Pd redesign

# Review requests

see test plan

# Risk assessment

low behind ff


protocol-designer/package.json
```diff
@@ -54,6 +54,7 @@
     "redux": "4.0.5",
     "redux-actions": "2.2.1",
     "react-popper": "1.0.0",
+    "react-router-dom": "6.24.1",
```
jerader, line 57: note: this is a higher version than the app uses (app uses version 5). I think we should have a chore ticket to update the app version.

protocol-designer/src/components/Navbar.tsx
```diff
@@ -0,0 +1,64 @@
...
+      justifyContent={JUSTIFY_SPACE_BETWEEN}
+      alignItems={ALIGN_CENTER}
+    >
+      <Flex
+        flexDirection="row"
+        flex={FLEX_NONE}
+        alignItems={ALIGN_FLEX_START}
+        alignSelf={ALIGN_STRETCH}
+      >
+        {navRoutes.map(({ name, navLinkTo }: RouteProps) =>
```
jerader: note: the figma designs show `import` and `export` as buttons in the nav bar, those will be added later since they aren't actual pages, just CTAs!

protocol-designer/src/feature-flags/selectors.ts
```diff
@@ -2,10 +2,16 @@ import { createSelector } from 'reselect'
...
+
+const selectFeatureFlags = (state: BaseState): Flags => state.featureFlags.flags
+
+export const getFeatureFlagData: Selector<Flags> = createSelector(
+  [selectFeatureFlags, getFlagsFromQueryParams],
+  (flags, queryParamsFlags) => ({
+    ...flags,
+    ...queryParamsFlags,
+  })
+)
```
jerader: refactored this to fix the following warning: `Selector getFeatureFlagData returned a different result when called with the same parameters. This can lead to unnecessary rerenders.
Selectors that return a new reference (such as an object or an array) should be memoized`

protocol-designer/package.json
koji: Brian and I talked about `react-router-dom` since v6's best practices are very different from v5 ones.
At that time, we didn't have to rush to update it.
https://reactrouter.com/en/main/upgrading/v5
Probably need to let this know all frontend folks.

protocol-designer/src/components/Navbar.tsx
```diff
@@ -0,0 +1,64 @@
...
+import type { RouteProps } from './types'
+
+const NavbarLink = styled(NavLink)`
+  color: ${COLORS.black90};
+  text-decoration: none;
+  align-self: ${ALIGN_STRETCH};
+  &:hover {
+    color: ${COLORS.black70};
+  }
+`
```
koji: nit

recently we put styled tag in the bottom of the component.
Maybe PD may be a good model case to make coding rules.

protocol-designer/src/components/Navbar.tsx
```diff
@@ -0,0 +1,64 @@
...
+        flexDirection="row"
+        flex={FLEX_NONE}
+        alignItems={ALIGN_FLEX_START}
+        alignSelf={ALIGN_STRETCH}
+      >
+        {navRoutes.map(({ name, navLinkTo }: RouteProps) =>
+          //  NOTE: '/' is the initial landing page which is not in the navBar
+          navLinkTo === '/' ? null : (
+            <NavbarLink key={name} to={navLinkTo as string}>
+              <LegacyStyledText
```
koji: question

Does the new PD design use the new `StyledText` in Figma?

jerader: i think the new styled text is only for the app! But good question, we can ask mel during the design meeting. none of the designs in figma for PD are finalized yet so this component will have to be updated anyway.

protocol-designer/src/components/Navbar.tsx
jerader: I didn't realize we made that rule but good idea. i agree it is a good time to make coding rules in PD. I'll move it to the bottom.

protocol-designer/src/components/types.ts
koji, line 1: can we make a new folder something like `types`?
I guess `components` could be misleading in our usage since we follow `atomic-design`.

protocol-designer/src/pages/Landing/index.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { NavLink } from 'react-router-dom'
+import { Flex, SPACING } from '@opentrons/components'
+
+export function Landing(): JSX.Element {
+  return (
+    <Flex flexDirection="column" margin={SPACING.spacing24}>
```
koji:
```suggestion
    <Flex flexDirection={DIRECTION_COLUMN} margin={SPACING.spacing24}>
```

protocol-designer/src/pages/Landing/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+      Create Opentrons protocol
+      <Flex
+        alignItems="center"
```
koji:
```suggestion
        alignItems={ALIGN_CENTER}
```

protocol-designer/package.json
shlokamin: hmm, what are the consequences of having different versions of react router dom in the monorepo? i know with react and some of these other core dependencies they pretty much have to be versioned the same or it makes configuration really challenging. 

until now we've tried realllly hard to keep the core dependencies between the app, pd, and ll the same so building/deploying/developing these apps in the monorepo is as easy as possible. it makes me nervous to break that pattern, but im not necessarily against it 

to @koji's point, i think it's worth migrating up the app's version of react-router-dom. i can get some sustaining resources to make this happen

protocol-designer/src/components/Navbar.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <Flex
+        flexDirection="row"
+        flex={FLEX_NONE}
+        alignItems={ALIGN_FLEX_START}
+        alignSelf={ALIGN_STRETCH}
+      >
+        {navRoutes.map(({ name, navLinkTo }: RouteProps) =>
+          //  NOTE: '/' is the initial landing page which is not in the navBar
+          navLinkTo === '/' ? null : (
+            <NavbarLink key={name} to={navLinkTo as string}>
```
shlokamin: interesting, is this the standard way to handle a landing page in react router dom? i would expect them to have some nice pattern to deal with this

protocol-designer/src/feature-flags/selectors.ts
```diff
@@ -23,3 +29,7 @@ export const getEnableAbsorbanceReader: Selector<boolean> = createSelector(
   getFeatureFlagData,
   flags => flags.OT_PD_ENABLE_ABSORBANCE_READER ?? false
 )
+export const getRedesign: Selector<boolean> = createSelector(
+  getFeatureFlagData,
+  flags => flags.OT_PD_REDESIGN ?? false
```
shlokamin: nit: this breaks the naming convention we have for other FF selectors

protocol-designer/src/components/types.ts
jerader: this is a good call, PD right now has the top level `App` component rendered in `src/components` 🤔 so i can move that out of `src/components` and move `protocolEditor` out of components as well and then make a top level `types` file. Let me see.

protocol-designer/src/components/ProtocolRoutes.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { Liquids } from '../pages/Liquids'
+import { StartingDeckState } from '../pages/StartingDeckState'
+import { ProtocolSteps } from '../pages/ProtocolSteps'
+import { CreateNewProtocol } from '../pages/CreateNewProtocol'
+import { Navbar } from './Navbar'
+
+import type { RouteProps } from './types'
+
+const pdRoutes: RouteProps[] = [
+  {
```
koji: Does the design team think of adding 404 page?

jerader: no idea! I will flag this to Felix though. thanks for calling it out!

protocol-designer/src/components/types.ts
```diff
@@ -0,0 +1,15 @@
...
+   * drop developed components into slots held by placeholder div components
+   * */
+  Component: React.FC
+  /** a route/page name to render in the nav bar
+   */
+  name: string
+  /** the path for navigation linking, for example to push to a default tab
+   */
+  path: string
+  navLinkTo?: string
```
koji: maybe remove optional since PD won't have `undefined` in a path, will it?
if  `?` can be removed, you don't need to use `as string` in Navbar component. 

```suggestion
  navLinkTo: string
```

protocol-designer/src/components/types.ts
jerader: hmm this is actually giving me more trouble than i thought. i'll keep digging into it. maybe i can fix it in a follow up

protocol-designer/src/components/ProtocolRoutes.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const location = useLocation()
+  const currentPath = location.pathname
+  const landingPage: RouteProps = {
+    Component: Landing,
+    name: 'Landing',
+    navLinkTo: '/',
+    path: '/',
+  }
+  const allRoutes = [...pdRoutes, landingPage]
+  const landingRoute = '/'
```
shlokamin: nit: since this is a string const we should rename to `LANDING_ROUTE`

protocol-designer/src/components/Navbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+      alignItems={ALIGN_CENTER}
+    >
+      <Flex
+        flexDirection="row"
+        flex={FLEX_NONE}
+        alignItems={ALIGN_FLEX_START}
+        alignSelf={ALIGN_STRETCH}
+      >
+        {navRoutes.map(({ name, navLinkTo }: RouteProps) => (
+          <NavbarLink key={name} to={navLinkTo as string}>
```
shlokamin: can we type guard in your `routes.filter` on like 22 so we don't have to do the string type assertion here? this could be dangerous since this is a route and is meant to me a string

shlokamin: aaand we should do the same thing in the desktop app lol. i think the awkwardness of this is coming from the `routes` list having some routes meant to be used in the nav bar, and some that aren't. might be cleaner to separate this into two different lists with two different interfaces. they can share the same variables but reading this (and the app code) it's not immediately clear why there's both a `navLinkTo` and `path` that seem to always hold the same value. splitting this into two different lists would fix that

protocol-designer/src/components/ProtocolEditor.tsx
```diff
@@ -14,43 +25,65 @@ import { PortalRoot as TopPortalRoot } from './portals/TopPortal'
...
-        <ConnectedNav />
-        <Sidebar />
-        <div className={styles.main_page_wrapper}>
-          <ConnectedTitleBar />
+      {enableRedesign ? (
+        <Flex flexDirection={DIRECTION_COLUMN}>
+          <Flex padding={SPACING.spacing12} flexDirection={DIRECTION_ROW}>
+            <PrimaryButton
+              onClick={() => {
+                dispatch(setFeatureFlags({ OT_PD_ENABLE_REDESIGN: null }))
```
shlokamin: more explicit to set this to `false` right?

protocol-designer/src/components/ProtocolEditor.tsx
```diff
@@ -14,43 +25,65 @@ import { PortalRoot as TopPortalRoot } from './portals/TopPortal'
...
       {showGateModal ? <GateModal /> : null}
       <PrereleaseModeIndicator />
-      <div className={styles.wrapper}>
-        <ConnectedNav />
-        <Sidebar />
-        <div className={styles.main_page_wrapper}>
-          <ConnectedTitleBar />
+      {enableRedesign ? (
+        <Flex flexDirection={DIRECTION_COLUMN}>
+          <Flex padding={SPACING.spacing12} flexDirection={DIRECTION_ROW}>
```
shlokamin: we might need to refactor this component, but i think this feature flag branching should go even higher than this (like way up at the `App` level). it's okay for old PD and new PD to share providers and middleware, but that should really be it. we want the UI to be completely new, and there's a bunch of old stuff in here (i can see css modules!).

protocol-designer/src/components/ProtocolRoutes.tsx
shlokamin: there's no need! PD is not talking to a server, if there is a route that doesn't exist we should just take them to the landing page

protocol-designer/src/feature-flags/selectors.ts
```diff
@@ -2,10 +2,16 @@ import { createSelector } from 'reselect'
 import { getFlagsFromQueryParams } from './utils'
 import type { BaseState, Selector } from '../types'
 import type { Flags } from './types'
-export const getFeatureFlagData = (state: BaseState): Flags => ({
-  ...state.featureFlags.flags,
-  ...getFlagsFromQueryParams(),
-})
+
+const selectFeatureFlags = (state: BaseState): Flags => state.featureFlags.flags
```
shlokamin: this makes it sound like it's a setter. it looks like this is just getting all FFs in redux. maybe rename to `getFeatureFlags`?

protocol-designer/src/pages/CreateNewProtocol/__tests__/CreateNewProtocol.test.tsx
```diff
@@ -0,0 +1,3 @@
+import { it } from 'vitest'
+
+it.todo('write test fore CreateNewProtocol')
```
shlokamin: `fore` 😛 

up to you, but i dont really like adding `it.todo`s cuz they add cruft. i think we should either just write a test or not make a test file

protocol-designer/src/pages/Landing/__tests__/Landing.test.tsx
```diff
@@ -0,0 +1,3 @@
+import { it } from 'vitest'
+
+it.todo('write test for Landing page')
```
shlokamin, line 3: same thing here re: it.todo (unless youre planning on adding one like rly rly soon)

protocol-designer/src/pages/Landing/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+export function Landing(): JSX.Element {
+  return (
+    <Flex flexDirection={DIRECTION_COLUMN} margin={SPACING.spacing24}>
+      Create Opentrons protocol
+      <Flex
+        alignItems={ALIGN_CENTER}
+        marginTop={SPACING.spacing24}
+        gridGap={SPACING.spacing16}
+      >
+        <NavLink to={'/createNewProtocol'}>Create new</NavLink>
```
shlokamin: i18n?

protocol-designer/src/pages/Landing/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+  Flex,
+  SPACING,
+  DIRECTION_COLUMN,
+  ALIGN_CENTER,
+} from '@opentrons/components'
+
+export function Landing(): JSX.Element {
+  return (
+    <Flex flexDirection={DIRECTION_COLUMN} margin={SPACING.spacing24}>
+      Create Opentrons protocol
```
shlokamin: i18n?

protocol-designer/src/pages/Landing/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <Flex
+        alignItems={ALIGN_CENTER}
+        marginTop={SPACING.spacing24}
+        gridGap={SPACING.spacing16}
+      >
+        <NavLink to={'/createNewProtocol'}>Create new</NavLink>
+        Import
```
shlokamin: i18n?

protocol-designer/src/components/ProtocolEditor.tsx
jerader: good point, there are a few classNames that shoudln't be in the redesign, i will refactor

protocol-designer/src/pages/CreateNewProtocol/__tests__/CreateNewProtocol.test.tsx
jerader: hmmm i like having the todos because it reminds us to go back and fix them, i can write a test but figured everything is still changing 😬

protocol-designer/src/pages/Landing/index.tsx
jerader: i chose not to add these to i18n because  i wanted to account for if the text isn't finalized. But since its the nav bar text it is probably final

protocol-designer/src/components/Navbar.tsx
jerader: i think i updated the code here, i am unsure what you are referring to?

protocol-designer/src/components/types.ts
koji: import issue?

protocol-designer/package.json
shlokamin: same question as last review about react router dom. since we rely on stuff like portals and other shared components from CL that need react router dom, is this going to cause us issues?

protocol-designer/src/components/types.ts
jerader: ah ya, i tried to move them again and got it working this time 👍

protocol-designer/src/Navbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+    >
+      <Flex
+        flexDirection="row"
+        flex={FLEX_NONE}
+        alignItems={ALIGN_FLEX_START}
+        alignSelf={ALIGN_STRETCH}
+      >
+        {navRoutes.map(({ name, navLinkTo }: RouteProps) => (
+          <NavbarLink key={name} to={navLinkTo as string}>
+            <LegacyStyledText
```
shlokamin: same comments here as my last review : 
https://github.com/Opentrons/opentrons/pull/15657#discussion_r1678360378
https://github.com/Opentrons/opentrons/pull/15657#discussion_r1678371785

we shouldn't be type asserting via `as` for something as important as routes. i know we do this in the app, but i dont think we should copy and paste this pattern.

i think we should either type guard in the `navRoutes` filter statement above, or split out nav bar routes and all routes into two separate lists with different interfaces

protocol-designer/src/ProtocolRoutes.tsx
```diff
@@ -0,0 +1,72 @@
...
+    name: 'Protocol steps',
+    navLinkTo: '/protocolSteps',
+    path: '/protocolSteps',
+  },
+  {
+    Component: CreateNewProtocol,
+    name: 'Create new protocol',
+    navLinkTo: '/createNewProtocol',
+    path: '/createNewProtocol',
+  },
```
shlokamin: nit: since these routes will show up in the browser's url i think they should be as terse as possible. `liquids` and `startingDeckState` seems like good names to me, but i feel like we can drop the `protocol` prefix/postifx in the rest of these routes

protocol-designer/src/localization/en/feature_flags.json
```diff
@@ -15,5 +15,9 @@
   "OT_PD_ENABLE_ABSORBANCE_READER": {
     "title": "Enable absorbance plate reader",
     "description": "Enable absorbance plate reader support."
+  },
+  "OT_PD_ENABLE_REDESIGN": {
+    "title": "Enable redesign",
+    "description": "A whole new world."
```
shlokamin, line 21: lol

protocol-designer/src/Navbar.tsx
jerader: oh i see. ya navLinkTo is actually always a string here so i can remove the type guard

protocol-designer/package.json
jerader: ok so there are 3 components in CL that use `react-router-dom`, 2 of them are not in use anywhere in the repo. the other one is used in PD; `NavTab` which is used in `ConnectedNavTab`. I smoke tested it and it works correctly still. 

I think we should keep this version of react-router-dom and update the app and components version? And delete the unused components.

protocol-designer/src/Navbar.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return (
+    <Flex
+      css={TYPOGRAPHY.h3Regular}
+      flexDirection={DIRECTION_COLUMN}
+      flex={FLEX_NONE}
+      justifyContent={JUSTIFY_SPACE_BETWEEN}
+      alignItems={ALIGN_CENTER}
+    >
+      <Flex
+        flexDirection="row"
```
koji: I think you can remove this or use `DIRECTION_ROW`

koji:
I think `OT_PD_REDESIGN.title` and `OT_PD_REDESIGN.description` shouldn't be displayed.

Screenshot 2024-07-16 at 5 57 40 PM

jerader:
> I think `OT_PD_REDESIGN.title` and `OT_PD_REDESIGN.description` shouldn't be displayed.
> 
> Screenshot 2024-07-16 at 5 57 40 PM

@koji yeah, i am pretty sure this is a cache issue. i've noticed it in the past with other FFs that were there but then removed. OT_PD_REDESIGN doesn't exist anymore.

protocol-designer/package.json
jerader: ok i thought about this and here is what i _think_ makes sense to do: 

- remove the 2 components that are no longer in use in Components Library that use `react-router-dom`
- move the `NavTab` into PD or even combine it with `ConnectedNavTab`
- remove `react-router-dom` as a dependency in Components `package.json`
- in a follow up - update App's `react-router-dom` version and Labware-library's `react-router-dom` version

What do you think @shlokamin @koji

protocol-designer/src/ProtocolEditor.tsx
koji, line 1: one question
for redesign, we will use `styled-components` instead of `css modules` , won't we?

protocol-designer/package.json
koji: that sounds good to me.

protocol-designer/package.json
koji, line 1: when we work on redesign, will we update the packages that are used only for PD?
I'm asking this because when we worked on v6.0, I guess we didn't update the packages used only for app.

koji (approved):
The changes look good to me.
I look forward to seeing the new PD!

protocol-designer/package.json
shlokamin: that sounds good to me! i think the last bullet is very important — and not just for code cleanliness. we need our dependencies to play nicely together, and we value the freedom to be able to share large pieces of infra/logic (that need those shared dependencies) across all of our frontend repos. this is why we want to keep all of our dependencies in lock step. so let's ticket that work out and make sure it's assigned. y'all have the 👍🏽 from me to do this whenever y'all please

shlokamin (approved):
the future is now

protocol-designer/package.json
jerader: great! i'll do that immediately next: https://opentrons.atlassian.net/browse/AUTH-564

protocol-designer/package.json
jerader: PD designs will not be finalized until around September. So there is time and i think during down time, it would be good to take care of some tech debt and update packages. I updated a bunch earlier this year in PD (like react dnd) so i think if there are other packages only used in Pd that are out of date, it would be great to migrate them.

PR #15658, opened 2024-07-15
fix(hardware-testing): fix ABR asair sensor time adjustment

rclarke0:
# Overview

Change asair time adjustment from 5 hrs to 4 hrs to account for daylight savings.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/abr-testing/abr_testing/data_collection/abr_robot_error.py
+++ b/abr-testing/abr_testing/data_collection/abr_robot_error.py
@@ -50,8 +50,8 @@ def compare_current_trh_to_average(
         avg_rh = round(mean(relevant_temp_rhs["Relative Humidity (%)"]), 2)
     except StatisticsError:
         # If there is one value assign it as the average.
-        if len(relevant_temp_rhs["Temp oC"]) == 1:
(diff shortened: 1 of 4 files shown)


sfoster1 (changes requested):
Ooh, let's not do this this way (unless you like having this exact same PR twice a year). Instead, either:
- If this code runs on a robot, move adjusting the timestamp to someplace that is not on a robot; this possibly includes "google sheets", making it known in the data that the time is in UTC and doing something gross like this: https://davidkeen.com/blog/2017/01/time-zone-conversion-in-google-sheets/ because google doesn't do timezone conversion for some reason (?)
- If this code runs on a laptop, which would indeed be happy and is what it looks like, then instead of doing `timestamp = datetime.datetime.now()` you can do `timestamp = datetime.datetime.now(tz=tzinfo.utcoffset(0))` ([tzinfo.utcoffset](https://docs.python.org/3/library/datetime.html#datetime.tzinfo.utcoffset) will generate the appropriate time zone info for the conversion), which also will work even if you run this script with a locale and timezone set to somewhere other than new york for whatever reason.

sfoster1 (approved):
Nice!

PR #15659, opened 2024-07-15
fix(app): disallow 8-channel 384 well selection if no starting well selected

brenthagen:
# Overview

disables the plus button for 8-channel 384 well selection when no starting well selected. also fixes a 384 well edge case when navigating back through the quick transfer flow - deselects all wells on 384 well selection mount to keep the the last selected index pointer in sync with the selected wells.

closes RQA-2854

Screen Shot 2024-07-15 at 12 13 13 PM

# Test Plan

manually verified the fix

# Changelog

 - Disallows 8-channel 384 well selection if no starting well selected

# Review requests

is this the behavior we want

# Risk assessment

low


smb2268 (approved):
looks good!

PR #15661, opened 2024-07-15
fix (abr-testing): Checks that most recent run is completed in abr_scale script

rclarke0:
# Overview

Checks that run being pulled is completed. Adds error handling if no new runs need to be added.

# Test Plan

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/abr-testing/abr_testing/data_collection/abr_google_drive.py
+++ b/abr-testing/abr_testing/data_collection/abr_google_drive.py
@@ -39,6 +39,8 @@ def create_data_dictionary(
     """Pull data from run files and format into a dictionary."""
     runs_and_robots: List[Any] = []
     runs_and_lpc: List[Dict[str, Any]] = []
+    headers: List[str] = []
+    headers_lpc: List[str] = []
     for filename in os.listdir(storage_directory):
         file_path = os.path.join(storage_directory, filename)
         if file_path.endswith(".json"):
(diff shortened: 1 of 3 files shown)


abr-testing/abr_testing/data_collection/abr_google_drive.py
```diff
@@ -141,6 +149,11 @@ def create_data_dictionary(
                 continue
     transposed_runs_and_robots = list(map(list, zip(*runs_and_robots)))
     transposed_runs_and_lpc = list(map(list, zip(*runs_and_lpc)))
+    try:
+        num_of_runs_added = len(runs_and_robots)
+        print(f"{num_of_runs_added} run(s) added.")
+    except UnboundLocalError:
```
sfoster1: this is usually not an exception you want to be catching. Maybe make `runs_and_robots` optionally `None` instead and always exist?

sfoster1 (changes requested):
Spidey-sense tingles at that UnboundLocalError, I think that should be optionally `None` instead

abr-testing/abr_testing/data_collection/abr_google_drive.py
rclarke0: runs_and_robots is defined at the beginning so i think i can get rid of the error handling

sfoster1 (approved):
Looks good!

PR #15663, opened 2024-07-15
fix(api): prevent stalls on PickUpTip Protocol API command

pmoegenburg:
# Overview

Recently in ABR testing, we've seen significant stalls surface from the head boards. One possible culprit is `_move_to_plunger_bottom()` now occurring after `tip_pickup_moves()`. This PR moves `_move_to_plunger_bottom()` back to before `tip_pickup_moves()`.

Closes RABR-431.

# Test Plan

This branch can now be run on ABR3 to determine if the method call order is indeed the cause of the stall issue.

# Changelog

# Review requests

# Risk assessment

### DIFF
--- a/api/src/opentrons/hardware_control/ot3api.py
+++ b/api/src/opentrons/hardware_control/ot3api.py
@@ -1808,6 +1808,8 @@ class OT3API(
         realmount = OT3Mount.from_mount(mount)
         instrument = self._pipette_handler.get_pipette(realmount)
 
+        self._move_to_plunger_bottom(realmount, rate=1.0)
+
         if (
             self.gantry_load == GantryLoad.HIGH_THROUGHPUT
             and instrument.nozzle_manager.current_configuration.configuration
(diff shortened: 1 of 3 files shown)


sfoster1 (changes requested):
If it works I suppose I'll change my tune, but (a) I really don't understand a physical mechanism for this fixing that problem - how'd you come up with it? and (b) contra the pr description it doesn't seem like this call moved it seems like it disappeared, and now reappeared. Is there somewhere else that we have to remove it from? Are we moving twice now?

pmoegenburg:
This proposal comes from logic not physical first principles, i.e. stalls were occurring less with `_move_to_plunger_bottom()` prior to `tip_pickup_moves()` and stalls are occurring more without `_move_to_plunger_bottom()` prior to `tip_pickup_movs()`. The recent refactor mimics the previous code process (with `prepare_for_aspirate()` (which calls `_move_to_plunger_bottom()`) occurring after `tip_pickup_moves()`) with the exception of the first `_move_to_plunger_bottom()`, which was deleted.

pmoegenburg:
Did not fix stall issue in ABR testing

PR #15666, opened 2024-07-15
feat(api, shared-data): build out functionality for tracking liquid height

aaron-kulkarni:
# Overview

Added public getter and private setter for tracking liquid level heights of wells.

re EXEC-603

# Test Plan

Still need to write a ton of unit tests.

# Changelog

* Added new exception: NoLiquidHeightDataError. This will only be for internal use and a user should never receive it.
* `get_well_last_measured_height` starts in `protocol_api/labware`, calls down to `protocol_api/core/labware`, then finally down to `protocol_engine/state/labware`
* `set_well_last_measured_height` is only called in `liquid_probe_in_place` in `protocol_engine/execution/pipetting.py`
* Adds a new attribute to labware definitoin called WellState. WellState has attribute `lastMeasuredLiquidHeight`. I decided to make the WellState attribute for future cases where we might want more detailed well state tracking.

# Review requests

Not sure if the way I'm setting it is correct. I couldn't find a convenient and seemingly correct way to edit labware state so I'm just doing it in `liquid_probe_in_place`.

I also ended up having to add wellStates to the labware schema which seems wrong but I don't know how else I would do it.

# Risk assessment

Pretty low.


aaron-kulkarni:
I think this is the wrong way to go about this. I'm going to start a new PR that implements WellState in a better way.

aaron-kulkarni:
Here is the other PR that takes a different approach to this problem: https://github.com/Opentrons/opentrons/pull/15681

